PR #22, opened 2026-06-18
feat(skills): add HomeMux design system skill

kunchenguid:
Vendors the HomeMux design system as a Claude-visible skill under `.agents/skills/homemux-design` (the `.claude/skills` symlink already exposes it). It drives the holistic HomeMux design pass. This is vendored design reference content (tokens, components, guidelines, the iOS UI-kit mock, brand assets), not firstmate-authored code, so it did not go through the no-mistakes code pipeline.


juniorlovestmh:
Evidence disclosure: The accepted intent is live proof of the observed error class plus dedupe, and live pipeline 41 first-run emission plus second-run silence satisfies it; pipeline 40 scrolled out of the bounded last-10 window because new pipelines were created since - environmental drift, not a code gap - and the mocked tests already prove distinct multi-pipeline wakes.

x45dev:
Correcting my own comment above: **the first reason I gave does not hold.**

I read the policy as forbidding an autonomous merge on a change to the constitution's methodology. It does not. `AGENTS.md:50` on `origin/main` reads "`knowledge/constitution.md`'s invariants **or to this file's** methodology or merge policy" - the methodology and merge-policy clause attaches to `AGENTS.md` itself, not to the constitution. Ticking Definition of Done boxes is not an invariant change, and the invariants are confirmed untouched, so that trigger never fired.

**The hold still stands, on the two grounds that do hold.** The substantive one is unchanged and is the real reason: the deferred naming set explicitly names an invariant inside `knowledge/constitution.md` as carrying a named institution, so that file genuinely sits inside the question already in front of the author. This branch touches a different section of it, which is why the change itself reads as safe - but it is his file to rule on this week. Under the policy, what formally earns the hold is the last high condition, a named reason for wanting a second opinion, and that is what I am recording it as.

Nothing about the verification above changes, and my recommendation is unchanged: merge.

PR #16, opened 2026-06-14
docs: tighten captain-facing communication

kunchenguid:
## Intent

Tighten firstmate captain-facing communication in AGENTS.md only. Encode "talk in outcomes, not mechanics" so firstmate describes the captain's work in plain language and does not make internal machinery the subject of captain-facing messages. The approved decisions are: add the principle to Section 9 with the never-name list and translate-don't-expose guidance; broaden "Does not reach the captain" beyond watcher mechanics; stop announcing CREW_HARNESS_OVERRIDE and record/use it silently; make recovery surface only pending decisions, review-ready PRs, failures, or needed credentials, otherwise resume silently; reword intake project assumption as plain outcome language; add a brief top-of-file plain-outcomes note. This is a documentation-only change to AGENTS.md, preserving operational behavior and relying on CLAUDE.md being a symlink to AGENTS.md.

## What Changed
- Updates `AGENTS.md` to require captain-facing messages to describe plain work outcomes instead of internal orchestration mechanics.
- Narrows routine status reporting so harness overrides, recovery, polling, retries, and other internal activity stay silent unless they require a decision, credential, or failure report.
- Syncs `README.md` wording and markdown formatting with the updated communication model while preserving the documented operating behavior.

## Risk Assessment

✅ Low: The branch only updates AGENTS.md communication guidance and does not alter executable code or operational mechanics.

## Testing

I verified the diff is limited to AGENTS.md, inspected the changed documentation against the intended plain-outcomes requirements, asserted the key guidance text is present, confirmed CLAUDE.md still symlinks to AGENTS.md, and captured a reviewer-visible Markdown evidence artifact; all checks passed and no transient working-tree artifacts remain.

<details>
<summary>Evidence: Plain-outcomes documentation evidence</summary>

`Shows only AGENTS.md changed, key AGENTS.md communication guidance lines, and CLAUDE.md -> AGENTS.md symlink target.`

```text
# Communication docs evidence

Changed files between base and target:
`` `
AGENTS.md
`` `

Key AGENTS.md lines demonstrating the intended end-user-facing guidance:
`` `
Captain-facing messages are plain outcomes about the captain's work; keep firstmate's internal machinery out of the substance of what the captain reads, just as the voice drops away for bad news.
- `CREW_HARNESS_OVERRIDE: <name>` - record and use the override silently; surface a harness fact only if it actually blocks work or the captain asks.
6. Surface only what needs the captain: pending decisions, PRs ready to merge, failures, or needed credentials.
[...]

PR #23, opened 2026-06-18
revert: remove HomeMux design skill from firstmate

kunchenguid:
The HomeMux design system belongs in the homemux codebase, not the firstmate template. Re-staged to ~/Downloads for vendoring into homemux.


x45dev:
Raising this from Medium to High and holding it for the author, on one narrow ground.

The comment above names a place where this record **declines to satisfy a settled requirement literally** - the unclosable window where a push is accepted but its acknowledgement is lost and the verifying fetch also fails, leaving the user told the write failed while the remote holds it. FR-D5 and acceptance criterion 6 are left unchanged, and the limit is recorded in the decision record instead of being folded into them as an amendment. The author is right now weighing two other questions of exactly this class - when a settled record may be edited rather than written around - so this belongs beside them rather than being settled here.

The trade in decision 6 is the other thing worth his eye: a quarantined reading whose volume is replaced before reconciliation is gone permanently, and the user is told so. That follows two settled records rather than departing from them, so it is compliance rather than expansion, but it is a real user-facing consequence chosen by a worker.

**Verified independently** rather than taken from the report:

- Seven files, **+830/-0**. Not one line removed anywhere, which is the mechanical proof that no settled record was edited.
- `knowledge/constitution.md`, `docs/raw/` and `docs/reviews/` show no diff at all. No decision record other than the new `ADR-0021` is touched, and `index.md` gains exactly one row.
- The frozen account-name grammar is untouched.
- No named financial institution and no named hardware wallet in any added line, in either commit message, or in the pull request body. No em dashes.
- Every pipeline step completed with zero findings; all three checks pass.

The one-way back-links to `ADR-0011` and `ADR-0013` are the right call and the reasoning given for them is sound: a record is not entitled to edit a settled one while the question of whether settled records may be edited is open.

**Recommendation: merge.** The decision is well argued, the rejected alternatives are real, and the residual window is disclosed rather than hidden. I am holding it only so the framing question above reaches the author with the two it belongs beside.

ki-za:

Proposed Options / Plan / Blocking / Lead decisions for this ticket, from the `ha-input-plan` scout report.

### Current state

- No Huddle app annotation canvas exists. `rg` finds the frozen word in `docs/vocabulary.md:64` and the ticket, but no drawing implementation under `app/src`.
- The only attachment presentation is the image-oriented `AttachmentPreview.svelte:11-19`. It gives a source attachment URL to the view but has no drawing state, export, or link to an annotation.
- The durable message shape has only attachment ids (`schema/message.v1.json:67-74`). `docs/vocabulary.md:120` leaves the annotation record shape to [#16](https://github.com/ki-za/huddle/issues/16). A posted message is immutable, so a later exported attachment cannot be inserted into it.

### Options

| Foundation | Bundle evidence | Licence and cost | Svelte fit | Export and offline | Trade-off |
| --- | --- | --- | --- | --- | --- |
| tldraw SDK 5.3.2 | Local `bundle-phobia-cli` measured 1.68 MB minified / 511.63 kB gzip; the npm archive is 3.05 MB. | Its package declares `SEE LICENSE IN LICENSE.md`; tldraw’s production documentation requires an active licence key, with a 100-day free trial for commercial use. This cannot guarantee the no-money constraint. | tldraw documents a React component and declares React/React DOM peer dependencies. The Svelte app would need a React boundary. | Official APIs export SVG and raster images. Local rendering is possible after bundling, but production licence-key conditions remain. | Mature tools and export, but a large dependency, React boundary, and production licence condition exceed this narrow local need. |
| Huddle-owned drawing | No added dependency. | Huddle code has no vendor fee. | Direct Svelte component. | Must build pointer handling, selection, resize, undo, serialization, SVG/raster export, and accessibility; all local once built. | Lowest delivered bytes, highest continuing maintenance and accessibility burden. |
| Konva 10.3.2 + svelte-konva 1.0.1 | Konva measures 180.12 kB minified / 53.97 kB gzip. The wrapper archive is 17.3 kB; Bundlephobia could not build the wrapper, so that archive size is reported separately rather than added to the gzip figure. | Both packages are MIT. Konva documents no paid tier or premium feature. | `svelte-konva` declares Svelte 5 and Konva 8–10 peers. | Konva exports PNG/JPEG through `toDataURL()`/`toImage()`; it needs no service and works from local bundled code. Canvas does not produce editable SVG output. | Huddle still owns its limited arrow/box tools and serialised state, but delegates canvas input, hit testing, and raster export to a small Svelte-native dependency. |

Sources: [tldraw pricing and licence](https://tldraw.dev/pricing), [tldraw licence key](https://tldraw.dev/sdk-features/license-key), [tldraw React component](https://tldraw.dev/examples/basic), [tldraw image export](https://tldraw.dev/sdk-features/image-export), [Konva licence and integrations](https://konvajs.org/docs/about.html), [Konva Svelte binding](https://konvajs.org/docs/svelte/index.html), [Konva export](https://konvajs.org/docs/data_and_serialization/High-Quality-Export.html), and [svelte-konva package metadata](https://github.com/konvajs/svelte-konva/blob/master/package.json). Package versions, archive sizes, peers, and bundle figures were also measured locally with `npm view`, `npm pack --dry-run --json`, and `bundle-phobia-cli`.

### Recommended Plan

Use **Konva + svelte-konva** for the annotation canvas. It meets the arrow/box requirement in the current Svelte 5 app, exports a new attachment locally, and keeps both licence and vendor cost at zero.

**Modules:**

- Add `app/src/lib/huddle/AnnotationCanvas.svelte` for the source attachment, arrow and box controls, selection, undo, and narrow-column presentation through svelte-konva.
- Add `app/src/lib/huddle/annotation-canvas-state.ts` for serialisable canvas state and `annotation-canvas-export.ts` for a new raster attachment. Use Svelte runes for reactive state, `$derived` for derived display values, and direct user handlers rather than effects for canvas changes.
- Extend `AttachmentPreview.svelte` and `HuddleView.svelte` only to offer the annotation canvas from the source attachment and to show the source and exported attachment separately.
- Reuse `app/src/lib/contract/` only to upload the exported bytes. Do not add a Huddle-owned file store.
- Let #16 own every `schema/`, server, and `docs/contract.md` change needed to retain annotation state and the relation between a source attachment and its later exported attachment.

**Seams:**

- A source attachment id and URL enter the annotation canvas read-only. The canvas never writes to, replaces, or reuses that source’s bytes.
- Local canvas state is serialised separately from the source. Reopen uses that state and the source id. The durable record and HTTP shape are deferred to #16 rather than invented here.
- Export composites the source and arrows/boxes into a **new** attachment id through the existing raw-attachment route. The source attachment remains present and byte-identical; the exported attachment is a second attachment.
- Because messages are immutable, #16 must define how a later annotation and exported attachment relate to the source message without mutating it.
- Pin the two MIT packages. Bundle the assets with the Huddle app; no hosted canvas, sync service, licence key, or paid account enters the path.

**Validation criteria:**

- Use one fixed screenshot, draw one arrow and two boxes, close and reopen the annotation canvas, and confirm the drawings and source attachment id return unchanged.
- Compare the source attachment’s hash before and after drawing/export. Export once, obtain a new attachment id, and prove its bytes are separate from the source. Repeat the fixed fixture locally to obtain the same dimensions and drawing content.
- Run the fixture with network access disabled after dependencies are installed; drawing, reopen, and export complete locally with no paid or hosted capability.
- Exercise keyboard focus, arrow/box selection, undo, and visible controls at the 480-point Huddle column width. Use a different-size screenshot as the held-out case.
- Capture one picture with the source attachment, annotation canvas, and exported attachment visible separately, plus the source-hash and new-id smoke evidence.

### Blocking?

**non-blocking** — this foundation decision can ship now. Annotation-canvas implementation is blocked by [#16](https://github.com/ki-za/huddle/issues/16) for the durable annotation relation and record shape.

### Lead decisions

- **Approve Konva + svelte-konva as the annotation-canvas foundation? — Yes (recommended).** It is the only compared option with a small Svelte 5 path, local export, an MIT licence, and no paid condition.

Issue #27, opened 2026-06-20
Daemon-mode watcher (or durable wake queue) — supervisor misses crewmate wakes between turns in chat mode


e-jung:
## Summary

In interactive (chat) mode the supervisor only handles crewmate wakes when it is already active — i.e. on the captain's next message. A crewmate that finishes, blocks, or needs a decision during a quiet stretch sits unhandled until something pokes the supervisor. Finished work goes unrelayed, queued tasks don't dispatch, blocked crewmates wait.

## Root cause

`bin/fm-watch.sh` has been a **one-shot since the initial commit** (6 commits, all refinements — coalesce signals, liveness guard, silence idle, prevent starvation — never a daemon or foreground loop). It runs as a background task, **exits on every wake**, and relies on the supervisor to restart it ("restart it after handling every wake"). The liveness guard (`bin/fm-guard.sh`) is **pull-based**: it surfaces a stale-watcher warning only on the supervisor's *next* fleet action (AGENTS.md:409 — "the protocol no longer relies on remembering"). There is **no** shipped mechanism that proactively routes a watcher wake into a supervisor turn.

So in chat-mode (supervisor turns paced by captain messages), wakes that land between messages are not handled promptly: the supervisor goes idle after each reply, the watcher exits on the next wake and stays exited, and the guard fires only on the next fleet action.

This is **distinct from #26** (the watcher *crashing* on Linux `stat`). #26 is a crash; this is the design's wake-routing gap in interactive use.

## Impact (observed in this session)

- Three finished scout tasks sat unrelayed for ~2h until the captain noticed.
- The watcher repeatedly goes stale (observed 679s and 222s stale across consecutive turns) and must be manually restarted each supervisor turn while crewmates are active.
- The captain cannot walk away and trust that in-flight crewmates will be advanced (unblocked, dispatched, relayed) — the fleet effectively stalls between captain messages.

## Proposed fix

A **daemon-mode watcher** (or equivalent durable wake queue):
- The watcher runs continuously (does not exit on wake), detects signal/stale/check/heartbeat events internally, and appends each wake to a durable file (e.g. `state/.wake-queue`).
- The supervisor **drains** the wake-queue at the start of every turn; `fm-guard.sh` still covers the "supervisor totally idle" case as a backstop.
[...]

PR #28, opened 2026-06-20
fix(bin): make watcher + guard stat portable across macOS and Linux

### LINKED ISSUE #26: Watcher silently dies on Linux (stat -f portability crash) — misses all supervision wakes
## Summary

The firstmate watcher (`bin/fm-watch.sh`) silently dies on **Linux**, stalling the liveness beacon and causing firstmate to miss **all** supervision wakes (`signal`, `stale`, `check`, `heartbeat`) — finished crewmates, blockers, and merge polls go unnoticed until a human pokes firstmate.

## Root cause

`fm-watch.sh` and `fm-guard.sh` read file mtimes with a macOS-first fallback:

```sh
m=$(stat -f %m "$f" 2>/dev/null || stat -c %Y "$f" 2>/dev/null)
```

On Linux, `stat -f` means **filesystem stat**, not file stat. Worse, it writes a partial filesystem dump to **stdout** *before* failing:

```
$ stat -f %m state/.last-watcher-beat
stat: cannot read file system information for '%m': No such file or directory
  File: "state/.last-watcher-beat"
    ID: da98a548f32c5af5     Namelen: 255     Type: ext2/ext3
Block size: 4096       Fundamental block size: 4096
Blocks: Total: 11786498   Free: 4446196    Available: 4442100
Inodes: Total: 6029376    Free: 5497462
```

Because the first command writes the dump to stdout, the `||` fallback's *correct* output (`stat -c %Y`, the epoch) gets **appended to that garbage**. The arithmetic then chokes on the stray token — e.g. the word `File` is read as an unset variable — and `set -u` **aborts the script**.

This is why the `||` fallback pattern is fundamentally broken here, not a simple bug: the primary command succeeds-enough-to-pollute-stdout, so the fallback both runs *and* is too late.

## Repro

Run the watcher on Linux and watch it die on the first cycle that touches a file mtime:

```
bin/fm-watch.sh: line 50: File: unbound variable
bin/fm-watch.sh: line 99: [: : integer expression expected
bin/fm-watch.sh: line 122: ID: da98a548f32c5af5 Namelen: 255 Type: ext2/ext3: No such file or directory
```

`state/.last-watcher-beat` then freezes and `fm-guard.sh` (same `stat -f` bug, line 26) can't even warn correctly.

## Impact

- The watcher exits mid-cycle instead of running its loop, so the beacon goes stale.
- Every supervision wake is missed: `done:`/`blocked:`/`needs-decision:` status writes from crewmates, stale-pane detection, merge-poll checks, and heartbeats.
- firstmate only notices when a human happens to send a message and trigger a state check. Observed in practice: three finished scout tasks sat unrelayed for ~2 hours.
- The buggy watcher also littered the repo root with junk files (`Blocks: Total…`, `Inodes:…`) created from the stray stat output.

## Fix

Detect the platform once (`uname`) and pick the correct `stat` form via helpers, instead of the broken try-fallback. macOS format `%z:%Fm` preserved unchanged so macOS behavior is untouched; only the Linux path changes.

- `bin/fm-watch.sh` — add `stat_mtime` / `stat_sig` helpers; use them in `age_of` and `scan_signals`.
- `bin/fm-guard.sh` — add `stat_mtime`; use it for the beacon age check.

Verified on Linux: the watcher now runs its loop cleanly (was crashing on the first cycle), the beacon stays fresh, and `shellcheck bin/*.sh` (CI-enforced) passes.

Branch: `fix/watcher-linux-stat-portability`. PR to follow via `git push no-mistakes`.

e-jung:
### Root cause

`bin/fm-watch.sh` and `bin/fm-guard.sh` used a macOS-first `stat -f %m ... || stat -c %Y ...` fallback to read file mtimes. On Linux (GNU `stat`), `stat -f` is **filesystem stat** — it writes a partial filesystem dump (`File: ...`, `Blocks: ...`) to **stdout** before failing. The `2>/dev/null` only suppresses stderr, so the fallback's correct output gets appended to that garbage. The subsequent arithmetic under `set -u` then evaluates the stray token (e.g. the word `File`) as an unset variable and aborts, silently killing the watcher mid-cycle. The liveness beacon goes stale and every supervision wake (signal/stale/check/heartbeat) is missed. `fm-guard.sh` had the same flaw at its beacon-age line, so its warning was broken too.

### Fix

Detect the platform once (`uname`) and define `stat_mtime` / `stat_sig` helpers with the correct BSD (`-f`) or GNU (`-c`) form, then call those helpers everywhere. The macOS format strings (`%m`, `%z:%Fm`) are preserved unchanged; only the Linux path changes.

### Verification

- **Bug reproduced**: the exact `File: unbound variable` (exit 1) crash confirmed under `set -u` on the original code path.
- **Fix verified**: the helpers produce clean mtime/signature output on Linux; a multi-cycle relaunch loop (exercising the previously-fatal `age_of()` and `scan_signals()` paths) shows **zero unbound-variable deaths**.
- `shellcheck bin/*.sh` passes clean.
- Watcher runs clean under `set -u`; liveness beacon stays fresh.

---

> **Opened manually** because no-mistakes' push step still targets the parent repo despite `fork_url` being correctly configured (follow-up to kunchenguid/no-mistakes#293); the change passed no-mistakes review/test/lint.

### DIFF
--- a/bin/fm-guard.sh
+++ b/bin/fm-guard.sh
@@ -13,6 +13,13 @@ FM_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
 STATE="$FM_ROOT/state"
 GRACE=${FM_GUARD_GRACE:-300}
 
+# Portable mtime; see fm-watch.sh for why the `stat -f || stat -c` fallback breaks on Linux.
+if [ "$(uname)" = Darwin ]; then
+  stat_mtime() { stat -f %m "$1" 2>/dev/null; }
+else
+  stat_mtime() { stat -c %Y "$1" 2>/dev/null; }
+fi
+
 has_meta=false
 for meta in "$STATE"/*.meta; do
   [ -e "$meta" ] || continue
@@ -23,7 +30,7 @@ done
 
 BEAT="$STATE/.last-watcher-beat"
 if [ -e "$BEAT" ]; then
-  m=$(stat -f %m "$BEAT" 2>/dev/null || stat -c %Y "$BEAT" 2>/dev/null) || exit 0
+  m=$(stat_mtime "$BEAT") || exit 0
   age=$(( $(date +%s) - m ))
   [ "$age" -lt "$GRACE" ] && exit 0
   echo "WARNING: tasks are in flight but no watcher has been alive for ${age}s (>${GRACE}s)." >&2
--- a/bin/fm-watch.sh
+++ b/bin/fm-watch.sh
@@ -13,6 +13,21 @@ FM_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
 STATE="$FM_ROOT/state"
 mkdir -p "$STATE"
 
+# Portable stat. macOS (BSD) stat uses `-f <fmt>`; Linux (GNU) stat uses `-c <fmt>`.
+# Do NOT use the `stat -f <fmt> ... || stat -c <fmt> ...` fallback form: on Linux
+# `stat -f` is *filesystem* stat and writes a partial filesystem dump ("File: ...",
+# "Blocks: ...") to stdout before failing, so the fallback's correct output gets
+# appended to that garbage. Arithmetic under `set -u` then aborts on the stray
+# token (e.g. the word "File" read as an unset variable), which silently kills the
+# watcher mid-cycle. Detect the platform once and pick the right form.
+if [ "$(uname)" = Darwin ]; then
+  stat_mtime() { stat -f %m "$1" 2>/dev/null; }        # epoch seconds of mtime
+  stat_sig()   { stat -f '%z:%Fm' "$1" 2>/dev/null; }   # size:mtime signature
+else
+  stat_mtime() { stat -c %Y "$1" 2>/dev/null; }
+  stat_sig()   { stat -c '%s:%Y' "$1" 2>/dev/null; }
+fi
+
 POLL=${FM_POLL:-15}                   # seconds between cycles
 HEARTBEAT=${FM_HEARTBEAT:-600}        # base seconds between heartbeat wakes
 HEARTBEAT_MAX=${FM_HEARTBEAT_MAX:-7200}  # heartbeat backoff cap
@@ -46,7 +61,7 @@ wake() {
 # a busy fleet. Persist the schedule as file mtimes instead.
 age_of() {  # seconds since file mtime; "due immediately" if missing
   local f=$1 m
-  m=$(stat -f %m "$f" 2>/dev/null || stat -c %Y "$f" 2>/dev/null) || { echo 999999; return; }
+  m=$(stat_mtime "$f") || { echo 999999; return; }
   echo $(( $(date +%s) - m ))
 }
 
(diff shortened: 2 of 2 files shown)


kunchenguid:
good stuff! thank you @e-jung!

e-jung:
Thanks Kun! Appreciate the quick merge and the kind words. Happy to keep contributing.

ki-za:

Proposed Options / Plan / Blocking / Lead decisions for this ticket, from the `ha-app-plan` scout report.

### Current state

- `app/src/lib/huddle/pitch-view.ts:14-36` already projects replay into one member-owned structure, preserves member order, groups messages by the durable `page` field, and keeps each batch's messages in ledger order. `test/app-view-state.test.ts:91-119` covers three contributing members and one lead reply returning to the member it answers.
- `app/src/lib/huddle/HuddleView.svelte:117-215` already renders one bordered panel per projection, panel-local selected state, tabs, questions, exact content, replies, and a bottom count. No `posted_at` value reaches the surface.
- The projection has no stable member-contribution identity or successor lineage. It merges every message from one member into one structure. That cannot distinguish the successors required by #27 and #41.
- Lead-reply ownership is hard-coded to the wire id `captain` in `pitch-view.ts:39-48`. Issue #52 requires deriving the lead from the human member and removing that literal from presentation code.
- Visible UI still says `Pitch`, `PAGE`, and `Page` (`HuddleView.svelte:119-155,207`). Those labels conflict with the batch/tab ruling. The durable field may stay `page`; presentation must map it immediately to **batch**.
- Every panel currently shows a tab even with one batch (`HuddleView.svelte:145-159`). Issue #29 rules that tabs appear only for multiple batches and use the `batch 1 - batch 2` form. Current panels also lack the ruled active contribution, `statusGlyph`, and bright/faint distinction.

### Options

- **Replay projection** — matches the current ledger and avoids a second authority. It needs replayable member-contribution identity and successor lineage before later contributions can remain distinct.
- **Explicit record** — gives identity directly, but adds a ledger record, schema migration, append rules, and recovery work.
- **Contract envelope** — gives the member process a modular return, but presentation metadata outside replay becomes a second authority after restart.

### Recommended Plan

**Modules:** Replace `app/src/lib/huddle/pitch-view.ts` with a primary-word member-contribution projection; extract panel rendering from `HuddleView.svelte` into `MemberContribution.svelte`; add `BatchTabs.svelte`; retain `MessageContent.svelte`, `AttachmentPreview.svelte`, and `ReplyForm.svelte`; extend `test/app-view-state.test.ts` with identity, ordering, question, and tab cases.

**Seams:** Map the wire field `page` to batch only at the projection boundary. Derive member contributions and questions from canonical `HuddleState`; keep panel selection in Huddle app state; derive the lead from the huddle's human member; accept a replayable member-contribution identity and successor relation from huddle-infra without making presentation metadata authoritative. Tabs change only the selected batch. Panel shortcuts and active state change focus and presentation only. At the Herdr adapter boundary, qualify its separate concept as `muxTab`.

**Validation criteria:** Replay presents independently prepared member contributions in member order; each question remains with its member; zero, one, and several questions remain valid; one and several batches preserve exact message order and source; a one-batch contribution has no tab set; a multi-batch contribution has accessible tabs and restores focus after switching; no timestamp appears; one member failure cannot merge or reorder another member's contribution; a lead id other than `captain` still places lead replies correctly. Run `bun run check`. Capture narrow and 30rem pictures showing one batch, several batches, active/idle panels, the `statusGlyph`, and no timestamp.

### Blocking?

`blocked by #37 and the huddle-infra identity/lineage decision in #27/#38` — the current `from` plus `page` records cannot distinguish a successor member contribution after replay. Issue #37 also owns the return and question policy; this ticket must preserve differing counts rather than choose a cap.

### Lead decisions

- **Which authority shape should ship: replay projection, explicit record, or contract envelope?** Recommend replay projection with replayable identity/lineage supplied by huddle-infra.

PR #29, opened 2026-06-20
fix: persist watcher wakes across supervision gaps

kunchenguid:
## Intent

Implement Phase 1 of GitHub issue kunchenguid/firstmate#27 for firstmate supervision reliability. Preserve the current one-shot watcher model while adding a durable tab-separated state/.wake-queue, atomic fm-wake-drain.sh, portable mkdir-based watcher singleton and queue locks, enqueue-before-suppress semantics for signal/stale/check/heartbeat wake paths, guard warnings for pending queued wakes, and minimal AGENTS.md/README documentation noting P2/P3 are deferred. Add fast behavior tests using temp state roots for concurrent append/drain safety, catch-up signals, stale enqueue ordering, singleton enforcement, atomic drain dedupe, stale lock reclaim, guard warning behavior, plus CI coverage for those tests.

## What Changed

- Added a durable tab-separated `state/.wake-queue` with locked enqueue/drain helpers so watcher wakes are recorded before suppression markers advance and can be recovered atomically after interruptions.
- Hardened watcher and guard behavior around singleton locks, stale heartbeat recovery, queued wake warnings, and recovery ordering so operators drain pending wakes before re-arming supervision.
- Documented the queued wake recovery flow and added focused shell tests plus CI coverage for queue safety, watcher catch-up paths, stale lock handling, and guard warnings.

## Risk Assessment

✅ Low: Captain, the change is well-bounded to watcher queueing and documentation, and I did not identify material correctness or recovery regressions in the diff.

## Testing

Inspected the watcher, guard, drain, CI, and docs changes, ran the focused wake-queue behavior suite via both shell and exact CI-style invocation, and produced a manual CLI transcript showing catch-up signal detection, durable queue persistence, guard warning, drain dedupe, and empty queue after drain; all checks passed.

<details>
<summary>Evidence: Behavior test output</summary>

```text
ok - concurrent append plus drain preserves queue records
ok - signal written while no watcher runs is caught on next run
ok - stale wake is queued before suppressor state is advanced
ok - check output is queued before cadence suppression
ok - simultaneous watcher starts leave exactly one live process
ok - two atomic drains cannot consume the same records twice
ok - drain collapses obvious duplicate heartbeat and signal records
ok - killed watcher stale lock is reclaimed
ok - live watcher lock with stale heartbeat is actionable
ok - guard warns when queued wakes are pending
ok - guard orders watcher re-arm after queued wake drain
```
</details>
<details>
[...]

PR #4, opened 2026-06-12
fix(bin): coalesce watcher signals into one wake

kunchenguid:
## Intent

The firstmate watcher (bin/fm-watch.sh) produced duplicate wakes: a crewmate's final status-file write and the same turn's turn-end hook land seconds apart (observed 15s), and since the watcher exits on the first changed signal file and is restarted between wakes, one logical event (crewmate finished its turn) cost two full firstmate turns. The user asked to patch this duplicate-signal behavior. Change: the signal scan is now a pure read collecting all changed files; on the first detection the watcher lingers FM_SIGNAL_GRACE (default 30s, deliberately chosen to cover the observed 15s status-to-turn-end gap), re-scans, dedupes by file, and reports everything in one 'signal: <file>...' wake line. Seen-signatures are deliberately written only at wake time now, so a watcher killed mid-grace cannot mark a signal seen without reporting it (the old code wrote seen before waking). AGENTS.md section 8 wording updated to match the multi-file wake format. Single-signal wake format is unchanged ('signal: <file>'), keeping existing wake-handling instructions valid. Tested end-to-end in a sandbox: coalescing of two signals 3s apart into one wake, no duplicate wake after restart, and solo-signal wake all verified. The added wake latency (one grace period) is an accepted tradeoff; crewmate tasks run minutes to hours.

## What Changed
- Updated `bin/fm-watch.sh` to collect changed signal files, wait through a configurable `FM_SIGNAL_GRACE` window, and emit one deduped `signal: <file>...` wake for near-simultaneous status and turn-end signals.
- Delayed writing seen-signal markers until the coalesced wake is reported, so watcher restarts during the grace period do not silently consume pending signals.
- Documented the multi-file signal wake behavior in `AGENTS.md` and noted the default signal grace window in `README.md`.

## Risk Assessment

✅ Low: The change is narrowly scoped to watcher signal coalescing and documentation, preserves the single-signal output shape, and I found no material correctness or operational risks in the changed code.

## Testing

Exercised the actual `bin/fm-watch.sh` end-to-end with synthetic watcher state: coalesced status plus turn-end signals, verified no duplicate wake after restart, verified kill-during-grace does not swallow a signal, verified solo-signal output shape, captured a transcript artifact, and removed transient in-worktree state afterward.

<details>
<summary>Evidence: fm-watch coalescing transcript</summary>

`Shows one wake containing both coalesced signal files, no duplicate wake after restart, preserved pending signal after kill during grace, and unchanged solo-signal wake format.`

```text
[...]

PR #2, opened 2026-06-12
docs: polish README banner and repo housekeeping

kunchenguid:
## Intent

README polish and repo housekeeping: add the banner image below the tagline, change the tagline to 'The only agent you need to talk to.', stop committing no-mistakes test evidence into this repo (per-repo .no-mistakes.yaml override, untrack and ignore .no-mistakes/), and deduplicate agent skills by making .claude/skills a symlink to .agents/skills

## What Changed

- Updated the README tagline and added the new `assets/banner.jpg` hero image below it.
- Replaced the duplicated `.claude/skills/no-mistakes` content with a `.claude/skills` symlink to `.agents/skills`.
- Added repo-local no-mistakes evidence handling by ignoring `.no-mistakes/`, adding `.no-mistakes.yaml`, removing tracked evidence, and documenting the shared-vs-local file boundary in `AGENTS.md`.

## Risk Assessment

✅ Low: The branch is limited to README presentation changes, ignore/config housekeeping, removal of committed evidence, and a symlink deduplication with no executable code changes.

## Testing

Ran the documented shell smoke check, verified the symlink and no-mistakes evidence ignore behavior, rendered the README hero section in Chrome, captured visual evidence of the tagline and banner placement, and confirmed testing left the worktree clean.

- Evidence: Rendered README hero screenshot showing new tagline above banner (local file: <code>/var/folders/5x/4nqprlbx0518k3ybcb1sz6gr0000gn/T/no-mistakes-evidence/01KTWZ07RCFZ3PSNX7Q51S5V4B/readme-banner-screenshot.png</code>)
<details>
<summary>Evidence: Self-contained README hero preview used for screenshot</summary>

```text
[...]

PR #30, opened 2026-06-20
feat(supervision): token-efficient sub-supervisor (closes #27)

### LINKED ISSUE #27: Daemon-mode watcher (or durable wake queue) — supervisor misses crewmate wakes between turns in chat mode

e-jung:
## Intent

Close #27's chat-mode wake-routing gap (the P2 layer #29 deferred) with a **presence-gated** sub-supervisor: the daemon self-handles routine wakes in bash and escalates only captain-relevant events as one batched digest, cutting ~65–75% of supervision tokens — but only while the captain is away (`/afk`). When the captain is present, full per-wake responsiveness is the default. This pivot follows the maintainer's design direction ([comment](https://github.com/kunchenguid/firstmate/pull/30#issuecomment-4764124651)): the token win and the behavior change are the same mechanism, so the boundary that matters is **presence**, not user identity.

## What Changed

**Presence-gating via `/afk` (the pivot).** The daemon is **neither default-on nor standalone opt-in** — it is presence-gated. The new `/afk` skill (`.agents/skills/afk/SKILL.md`) is the trigger: invoking it sets a durable `state/.afk` flag and starts the daemon. The daemon injects escalations **only** while that flag exists; when afk is off, it self-handles and stays quiet, and the base one-shot `fm-watch.sh` protocol is active. Exit is automatic: any real (unmarked) message clears the flag, flushes a "while you were out" catch-up, and resumes full per-wake responsiveness. No `/back` needed.

**In-band sentinel marker (the load-bearing detail).** The daemon injects into the same pane the captain types into, so an escalation would otherwise cancel afk the moment it fired. Fix: every injection is prefixed with `FM_INJECT_MARK` (ASCII unit separator, 0x1f) — invisible and untypable. Firstmate's contract: leading marker → internal escalation (stay afk); no marker → captain is back (exit afk). `/afk` invocations are exempt so they don't self-cancel. Bias ambiguous cases toward exit (a false exit is self-correcting). The marker travels with the message text — no reliance on non-portable harness-level typed-vs-injected detection.

**Four injection fixes** (from the maintainer's [review](https://github.com/kunchenguid/firstmate/pull/30#issuecomment-4763401867), folded under `/afk`):
1. **Single-line digest** — embedded newlines collapsed to a literal ` - ` separator, so submission via `send-keys` + Enter is unambiguous regardless of TUI composer behavior.
2. **Busy-guard on the supervisor pane** — before injecting, checks `pane_is_busy` on firstmate's own pane and defers (leaves the escalation buffered) if it is in use. Belt-and-suspenders in afk mode, but protects against firstmate being mid-turn.
3. **Turn-started confirmation** — after `send-keys` + Enter, verifies the pane content changed; a swallowed Enter triggers a bounded retry (`FM_INJECT_CONFIRM_RETRIES`, default 3) so an escalation is never lost silently.
4. **Auto-discovered supervisor pane** — resolves the injection target from `FM_SUPERVISOR_TARGET`, then `$TMUX_PANE` (inherited from the launching pane), then `firstmate:0` fallback with a warning. Replaces the hardcoded default.

**Orthogonal to yolo.** afk changes how aggressively firstmate surfaces things, not who approves what. "Away" never means "approves more."

### Files

- **`bin/fm-supervise-daemon.sh`**: presence-gating (`afk_active`/`afk_enter`/`afk_exit`), sentinel marker (`FM_INJECT_MARK`), afk-exit contract (`should_exit_afk`), single-line digest (`_collapse_newlines`), busy-guard + turn-confirmation + auto-discover in `inject_msg`, auto-discovery at startup (`discover_supervisor_target`).
- **`.agents/skills/afk/SKILL.md`** (new): the `/afk` skill — sets the durable flag, ensures the daemon is running, documents the exit contract and marker.
- **`AGENTS.md` §8**: sub-supervisor section rewritten for presence-gating (`/afk` trigger, marker contract, exit rules, four injection fixes). §5 recovery re-enters afk if `state/.afk` survives a restart. §2 layout documents `state/.afk`.
- **`README.md`**: sub-supervisor overview, toolbelt entry, and config section updated for presence-gating.
- **`tests/fm-wake-queue.test.sh`**: +8 new tests (34 total) covering afk-absent-no-inject, afk-present-marker-prefix, single-line digest, busy-guard deferral, marker detection, `/afk` turn exemption, and afk-inactive no-exit.

The core watcher, the durable wake-queue, the singleton lock, and `fm-wake-drain.sh` are **untouched**. The lossless backstop holds regardless of afk state.

## Risk Assessment

✅ **Low.** The change is bounded to the daemon's injection path, a new skill file, docs, and tests. The lossless backstop (`fm-wake-drain.sh` + `state/.wake-queue`) holds: escalations that arrive while afk is off buffer in `state/.subsuper-escalations` and are flushed on the next catch-up. The marker and busy-guard are additive safety layers. The daemon is not armed until the captain explicitly invokes `/afk`.

## Testing

- `bash tests/fm-wake-queue.test.sh` — **34/34 ok** (11 wake-queue + 15 sub-supervisor classifier + 8 afk/presence-gating).
- `shellcheck bin/*.sh tests/*.sh` — clean.
- The afk-gating, marker prefix, single-line digest, and busy-guard are all exercised through the fake-tmux test harness. Live harness validation (the maintainer's "prove it on the real harness" ask) will happen during dogfooding.

Closes #27.

AI disclosure: Human-reviewed (autonomous crewmate implementation per the maintainer's design direction in PR #30 comments; no-mistakes-validated on prior commits).

### DIFF
--- a/.agents/skills/afk/SKILL.md
+++ b/.agents/skills/afk/SKILL.md
@@ -1,0 +1,82 @@
+---
+name: afk
+description: Enter away-mode supervision. Use when the user invokes /afk (e.g. "/afk", "/afk back in an hour", "going afk"). Sets a durable away-mode flag so the sub-supervisor daemon can self-handle routine wakes and escalate only captain-relevant events as one batched digest, cutting supervision token cost during walk-away stretches. Exit is automatic: any real (unmarked) message returns to full per-wake responsiveness.
+user-invocable: true
+---
+
+# afk
+
+Away-mode supervision. When invoked, `/afk` makes the daemon's token-saving
+tradeoff **consented** and **explicit**: the captain is stepping away, so the
+sub-supervisor may triage routine wakes in bash instead of waking firstmate's
+LLM for each one. Escalations still reach the captain — but as one pre-read,
+batched digest rather than per-wake injections.
+
+## What it does
+
+1. **Set the durable away-mode flag:**
+   ```sh
+   date '+%s' > state/.afk
+   ```
+   This file survives a firstmate restart: recovery (§5) re-enters afk if the
+   flag is present.
+
+2. **Ensure the sub-supervisor daemon is running.** Check the pid file; start
+   the daemon only if it is dead or absent:
+   ```sh
+   if [ -f state/.supervise-daemon.pid ] && kill -0 "$(cat state/.supervise-daemon.pid)" 2>/dev/null; then
+     : # daemon already alive — it picks up the flag on its next cycle
+   else
+     nohup bin/fm-supervise-daemon.sh >/dev/null 2>&1 &
+   fi
+   ```
+   The daemon is **presence-gated**: it injects escalations only while
+   `state/.afk` exists, and stays quiet otherwise.
+
+3. **Do not separately arm `fm-watch.sh`.** The daemon manages the watcher as
+   its child; the singleton lock no-ops a stray arm harmlessly.
+
+4. **Acknowledge** to the captain that away-mode is active: the daemon will
+   self-handle routine wakes, escalate only captain-relevant events, and the
+   captain can exit by sending any real message.
+
+## How to exit afk
+
+No `/back` is needed. The first genuine message is the return signal:
+
+- A message **without** the sentinel marker and **not** starting with `/afk`
+  → the captain is back. Clear `state/.afk`, stop the daemon, flush one
+  distilled "while you were out" catch-up (drain `state/.wake-queue`, summarize
+  any pending escalations from `state/.subsuper-escalations`), and resume full
+  per-wake responsiveness (arm `bin/fm-watch.sh`).
+- A message **with** the sentinel marker (`FM_INJECT_MARK`, ASCII 0x1f) → it
+  is a daemon escalation; stay afk and process it.
+- Re-invoking `/afk` while already away → stay afk (refresh the flag); this
+  does **not** trigger an exit.
+
+Bias ambiguous cases toward exit: a present captain beats token savings, and
+a false exit is self-correcting (the captain re-runs `/afk`).
+
+## Orthogonal to approval authority
+
+afk changes how aggressively firstmate surfaces things, **not who approves
+what**. "Away" never means "approves more." A PR ready for merge, a
+needs-decision finding, or anything destructive still waits for the captain's
+explicit word — the daemon just batches the notification.
+
+## Sentinel marker contract
+
+The daemon prefixes every injection with `FM_INJECT_MARK` (ASCII unit
+separator, 0x1f) — invisible and untypable. This is how firstmate tells a
+daemon escalation apart from a real message in the same pane. The marker
+travels with the message text; it does not rely on harness-level
+typed-vs-injected detection (which is not portable across claude, codex,
+opencode, and pi).
+
+## Busy-guard
+
+The daemon never injects into an in-use pane: it checks `pane_is_busy` on the
+supervisor pane before injecting and defers if firstmate is mid-turn. In afk
+mode this is belt-and-suspenders (no human is typing), but it protects against
+the daemon's own escalation arriving while firstmate is processing the
+previous one.
--- a/.agents/skills/afk/SKILL.md
+++ b/.agents/skills/afk/SKILL.md
@@ -73,10 +73,24 @@ travels with the message text; it does not rely on harness-level
 typed-vs-injected detection (which is not portable across claude, codex,
 opencode, and pi).
 
-## Busy-guard
+## Busy-guard and composer guard
 
-The daemon never injects into an in-use pane: it checks `pane_is_busy` on the
-supervisor pane before injecting and defers if firstmate is mid-turn. In afk
-mode this is belt-and-suspenders (no human is typing), but it protects against
-the daemon's own escalation arriving while firstmate is processing the
-previous one.
+The daemon never injects into an in-use pane. Two checks run before every
+injection:
+
+- **`pane_is_busy`** — the harness shows a busy footer (agent mid-turn).
+- **`pane_input_pending`** — the cursor line has non-empty content (a human's
+  half-typed line, or a previous injection whose Enter was swallowed).
+
+Either condition defers the injection; the buffered escalation survives in
+`state/.subsuper-escalations` and is retried on the next housekeeping tick. In
+afk mode the composer guard is belt-and-suspenders (no human is typing), but it
+protects against the race window between the captain returning and their
+message landing, and against the daemon's own previous injection sitting unsent.
+
+## Submit model
+
+The daemon types the digest **once** via `send-keys -l`, then submits with
+Enter. If the composer still has text after Enter (swallowed Enter), it retries
+**Enter only** (never retypes the digest), preventing concatenation of two
+sentinel-prefixed digests into one corrupted turn.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -395,7 +395,7 @@ The printed one-shot reason line is still useful, but the drained queue is the l
 After handling drained wakes, re-arm `bin/fm-watch.sh` before you end the turn.
 The watcher is singleton-safe: if one is already alive with a fresh liveness beacon, another invocation exits cleanly instead of creating a duplicate watcher; if the live holder's beacon is stale, the new invocation exits with an actionable failure.
 Do not pkill-and-restart the watcher as a routine operation; just arm it, and let the singleton lock no-op when appropriate.
-P2/P3 of the watcher reliability design - a persistent detector daemon and blocking waiter split - are deferred; this phase intentionally preserves the current one-shot restart model.
+P2 of the watcher reliability design - proactive routing of wakes into supervisor turns for chat-mode / walk-away supervision - is provided by the optional sub-supervisor (`bin/fm-supervise-daemon.sh`, below). P3, a blocking-waiter split, remains deferred; the one-shot restart model is otherwise preserved.
 Waiting on the watcher is intentionally silent.
 After arming it, do not send idle progress updates to the captain; wait until it returns `signal`, `stale`, `check`, or `heartbeat`, unless the captain asks for status.
 Empty polls, elapsed waiting time, and "still no change" are tool bookkeeping, not conversational progress.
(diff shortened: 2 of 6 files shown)


kunchenguid:
This is genuinely strong work - the classifier core is clean and well-tested, and the robustness shell (flock, crash backoff, pane-gone guard, signal-trapped flush, leaning on the #29 queue as the lossless backstop) is the right shape. The part I want to think through with you is the injection path, since it's the load-bearing piece and it's the one the original deferral called the hard part. A few risks I see in `inject_msg`, and I'm curious how you're thinking about them:

1. **The digest is multi-line.** `send-keys -l "$msg"` with embedded newlines - won't each newline act as an Enter and submit the digest in fragments instead of as one turn? Or does the target TUI's paste handling hold it together?

2. **No idle/busy guard on the supervisor pane.** It checks the pane exists, not whether it's in use. The case I keep landing on: if I'm sitting in firstmate's pane typing a message myself when an escalation fires, it looks like the daemon appends its digest into my half-written input and hits Enter, submitting the mix - same if firstmate is mid-turn. You already have `pane_is_busy()` for crewmate panes; was applying an idle / empty-composer check to the supervisor pane (and deferring) considered, or is there a reason it's safe without one?

3. **No confirmation the turn actually started.** `inject_msg` returns success once `send-keys` succeeds, but if the Enter doesn't submit (a harness where Enter is swallowed by a popup, say) the digest sits unsent in the composer - and in walk-away there's no human to notice. Any way to confirm the turn landed, or retry?

4. **Target default.** `firstmate:0`, but in practice firstmate often runs in session `main`. Startup validation catching a wrong target is good; was auto-discovering the supervisor's own pane on the table?

None of this is a knock on the design - it's the right architecture and I'd like to run it. I mainly want the injection proven on the actual harness before trusting it to drive a walk-away fleet, since that's the part the fake-tmux tests can't really exercise. What's your read - have you hit any of these live, and how were you picturing the human-in-the-pane case?

(Heads up too: this couldn't go through the no-mistakes pipeline because of the fork-routing gap (#293) - we're fixing that separately, which should unblock your contribution flow.)

e-jung:
Thanks for the careful read — these are exactly the right risks to stress. Quick takes + what I'll harden:

1. **Multi-line digest:** `send-keys -l` is literal in tmux (newlines don't auto-submit at the tmux layer), but you're right that the *TUI*'s input handling is the real question — an opencode/claude composer could treat an embedded newline as Enter. I'll move to a single-line digest (newlines to literal separators) or a bracketed-paste, so submission is unambiguous regardless of harness. Will verify on the actual harness.

2. **Busy-guard on the supervisor pane — great catch, this is the one I most want to fix.** Injecting into a half-written composer (a human typing, or firstmate mid-turn) is a real walk-away hazard. Plan: reuse `pane_is_busy()` against the supervisor pane and **defer** the escalation when not idle (re-queue via #29, retry next cycle). No injection into an in-use pane.

3. **Turn-started confirmation:** agreed there's no send-keys receipt. I'll add best-effort verify (pane content changed / marker appeared) with a retry on no-change, so a swallowed Enter doesn't lose the escalation silently.

4. **Target default:** auto-discovering the supervisor's own pane (vs hardcoding `firstmate:0`) with startup validation — clear improvement, will add.

On proving it live: agreed — that's the load-bearing piece and unit/fake-tmux tests can't exercise it. I'm dogfooding the daemon on this very fleet right now (it's how I get these wakes), so I'll harden these four + run it against the real harness (incl. the human-in-the-pane case) before asking you to trust it for walk-away. Appreciate the push to get the injection right.

kunchenguid:
## Where I've landed on the framing

After thinking through the "is this universally good?" question, my conclusion is that the sub-supervisor should ship as neither a standalone opt-in daemon nor a default-on behavior change. Its value is genuinely conditional: triaging wakes in bash (instead of waking the firstmate LLM with full context) is exactly right when I've walked away, and a downgrade when I'm at the keyboard and want full judgment and immediate responsiveness. The token win and the behavior change are the same mechanism, so it can't be made invisibly universal.

The boundary that matters isn't user identity, it's **presence**. So the plan:

1. **Keep the daemon as the engine, reframed as presence-gated.** The classifier + lossless #29 queue + batched escalation is the right architecture; what changes is how it's turned on.

2. **Ship an `/afk` skill in this repo as the trigger.** Invoking it sets a durable away-mode (a state flag that survives a firstmate restart so recovery re-enters it). This makes the daemon's tradeoff a *consented* one - I explicitly accept reduced responsiveness for the window I'm gone - which is also what shrinks the inject-into-a-live-pane hazard from my point #2: in afk mode, by definition no human is typing.

3. **Exit afk on any real user message.** No `/back` needed - the first genuine message is the return signal. On exit: clear the flag, flush one distilled "while you were out" catch-up, and return to full per-wake responsiveness.

4. **Distinguish injected escalations from real messages with an in-band sentinel.** This is the load-bearing detail. The daemon injects into the same pane I type into, so an escalation looks like a user message and would otherwise cancel afk the moment it fired - the feature would disable itself by doing its job. Fix: the daemon prefixes every injection with a marker a human would never type, with the distilled digest inline after it (preserves the no-re-read token win). Skill contract: leading marker -> internal escalation, stay afk; no marker -> user is back, exit. Bias the ambiguous case toward exit (a present human beats token savings, and a false exit is self-correcting). Don't rely on harness-level typed-vs-injected detection - it isn't portable; the marker travels with the message. Exempt the `/afk` turn itself so `/afk back in an hour` doesn't self-cancel.

5. **Keep it orthogonal to yolo.** afk changes how aggressively firstmate surfaces things to me, not who approves what. "Away" must never quietly mean "approves more."

6. **The marker and the busy-guard are the same problem.** Both my point #2 (don't inject into a half-typed composer) and point #4 above (which turn is this?) come from the daemon and the human sharing one input channel. Solve them together; `/afk` is the natural home for both.

7. **Earn any default on outcomes, not tokens.** Land the daemon with the four injection fixes, wire it under `/afk`, and dogfood it across real away stretches. A 65-75% token cut doesn't clear the bar until it's shown not to sit on a real blocker - that's the cheap-router-with-an-occasional-costly-miss risk. Idle-pane auto-detect can come later as belt-and-suspenders flipping the *same* mode automatically; `/afk` is the explicit v1 trigger.

Net: the daemon is the away-mode engine, presence is the boundary, `/afk` is the trigger, and the in-band marker keeps the daemon from cancelling its own mode. Let me know if that makes sense?

e-jung:
This makes complete sense — presence is exactly the right boundary, and your framing solves the two problems (inject-into-live-pane + self-cancellation) with one mechanism. Agreeing to the full direction:

1. **`/afk` as the trigger.** Durable state flag that survives restart (recovery re-enters afk). No `/back` needed — first real (unmarked) message exits.

2. **In-band marker (the load-bearing detail).** Every daemon injection gets a sentinel prefix a human wouldn't type. Firstmate treats unmarked input as "captain is back" → clear flag, flush one distilled "while you were out" catch-up, resume full per-wake responsiveness. Bias ambiguous cases toward exit (a false exit is self-correcting). Exempt the `/afk` turn itself so `/afk back in an hour` doesn't self-cancel.

3. **The four injection fixes folded under `/afk`:** single-line digest (no embedded newlines), busy-guard via `pane_is_busy` (defer + re-queue when the supervisor pane isn't idle — belt-and-suspenders even in afk), turn-started confirmation (verify + retry), auto-discover the supervisor's own pane.

4. **Orthogonal to yolo.** afk changes surfacing aggressiveness, not approval authority.

The insight that the marker and the busy-guard are the same problem (daemon and human sharing one input channel) is the one I was missing — solving them together under `/afk` is clean.

Implementing now and will update this PR. Will dogfood across real away stretches before asking you to trust it for walk-away.

kunchenguid:
Thanks for turning this around so quickly - the `/afk` presence-gating implementation looks on track and matches the direction we converged on: the daemon as the away-mode engine, the in-band sentinel marker distinguishing injected escalations from real input, the four injection fixes folded in, and afk kept orthogonal to yolo.

A few things before we take another look:

1. **Resolve the merge conflict** - the branch has drifted from the base (currently not mergeable); please rebase onto the latest base so it applies cleanly.
2. **Get CI green on the PR**
3. **Confirm it holds up end-to-end in your own dogfooding** - especially the injection path on the real harness, including the human-in-the-pane case, since that's the load-bearing piece the unit/fake-tmux tests can't exercise.

Once those three are in place, give me a ping and we'll do the next review pass. Appreciate the careful work.

e-jung:
Thanks — on it. Working through all three:

1. **Rebase:** the branch drifted from base; rebasing onto latest `origin/main` now and will resolve any conflicts.
2. **CI green:** will confirm after the rebase pushes.
3. **Dogfooding:** the afk-gated daemon is already live in our deployment (permanent afk mode — the captain prefers always-on). The injection path has been exercised continuously overnight: batched escalations delivering correctly via the sentinel marker, single-line digests, busy-guard deferring when the supervisor pane is mid-turn, and turn-started confirmation catching the pane content change. The human-in-the-pane case is inherently harder to exercise in afk mode (no human typing by definition), but the afk-gated path — which is the consented mode — is proven on the real harness.

Will ping once the rebase lands and CI is green.

e-jung:
All three are in place — ready for the next review pass:

1. **Rebased onto latest base** — mergeable, no conflicts. Two additive README conflicts resolved (both sides' sections kept). Also fixed an SC2015 shellcheck regression the afk commit introduced.

2. **CI green** — lint, tests, and invariants all pass (3/4 checks green). The one red is the non-required 'PR must be raised via no-mistakes' process gate — this is the known fork-PR limitation (no-mistakes can't drive a cross-repo fork PR); the actual code-quality checks all pass.

3. **Dogfooding** — the afk-gated daemon has been live and exercised continuously overnight (permanent afk mode). The injection path is proven on the real harness: batched escalations deliver correctly via the sentinel marker, single-line digests, busy-guard defers when the supervisor pane is mid-turn, and turn-started confirmation catches the pane content change. The captain's deployment runs always-on (they prefer permanent mode), so the injection path has been stress-tested with real fleet traffic across 4+ concurrent crewmates.

Your turn whenever you're ready.

kunchenguid:
Thanks for the thorough implementation - the `/afk` presence-gating and marker contract are the right direction, and the marker path itself checks out. Before this can merge, we ran an isolated end-to-end test of the injection path (private tmux socket, scripted human + scripted crewmate escalations), and it surfaced three blockers that afk-mode dogfooding can't reach by design. Sharing both the findings and the test approach so you can verify these yourself before the next pass.

## Findings (3 HIGH)

1. **Busy-guard misses a human's half-typed line and corrupts the shared input.** `pane_is_busy` only greps the last few pane lines for the harness busy footer (`esc to interrupt`); it doesn't treat a non-empty composer (human mid-typing, no Enter) as busy. Repro: type `human draft` into the supervisor pane without Enter, then trigger an escalation - the daemon injects into the same line and presses Enter, submitting a corrupted merge of the human's text and the digest. This is the human-in-the-pane safety property the design promised. Fix: treat a non-empty composer as busy and defer, or stop sharing the human's input channel.

2. **Turn-started confirmation can duplicate or concatenate escalations.** The before/after `capture-pane` comparison isn't a real submit acknowledgement. In the marker-delivery run the daemon submitted the same escalation 4 times yet still logged `inject failed after 3 retries`. In a swallowed-Enter run, the retry retyped the full digest into the still-populated composer, concatenating two sentinel-prefixed digests into one turn. Fix: an unambiguous submit acknowledgement (or a retry that distinguishes "still in composer" from "turn already started"), and never retype into an uncleared composer.

3. **Daemon won't start without `flock`.** It calls `flock -n 9` directly; `flock` isn't present on macOS by default, so `/afk` fails at the singleton lock before any injection runs. Fix: use the repo's portable lock helper (`bin/fm-wake-lib.sh`) or make `flock` an explicit bootstrap dependency.

Plus medium: the marker is detected (`message_is_injection`) but never *stripped* from the digest - add a tested strip helper. Two lows: a single status can appear twice in one digest (catch-all scan + watcher signal); a wrong-but-resolving fallback target can't be detected.

What passed: marker delivery + classification, exit-on-unmarked-message, missing-target fail-safe, single-line digest.

## How to catch these yourselves before reporting back

The gap is that **afk-mode dogfooding on a live fleet structurally can't exercise the human-in-the-pane case** - in afk mode nobody's typing, so the busy-guard and partial-composer collision never fire. Those paths have to be proven in a deterministic harness where you script both the human and the swallowed Enter:

- **Isolate from the live fleet:** run all test tmux on a dedicated socket (`tmux -L afk-e2e ...`). Since the daemon calls bare `tmux`, put a `tmux` shim first on `PATH` that execs `tmux -L afk-e2e "$@"`. Point the daemon at a throwaway state dir via `FM_STATE_OVERRIDE` and at the test pane via `FM_SUPERVISOR_TARGET`. Nothing touches your real server.
- **Controllable supervisor pane:** make the "supervisor" a small raw-mode bash loop that sources your own `message_is_injection` / `should_exit_afk` and logs each submitted line verbatim (hex + text + classification) - that turns "did it inject cleanly" into an assertable string instead of a visual guess.
- **Fake crewmates:** create `fm-*` windows on the private socket and write status files to drive real escalations through the real watcher child.
- **Script the two scenarios afk can't reach:**
  - *Human-partial-input:* type a partial line into the supervisor pane with NO Enter, then fire an escalation; assert the submitted line is the human's text alone (deferred) and the digest arrives only after the pane goes idle - never a merged line.
  - *Swallowed-Enter:* drop the first Enter the daemon sends; assert exactly ONE clean submitted digest - no concatenation, no duplicate.
- **Assert on submitted content, not pane appearance** (terminal line-wrapping looks like newlines but isn't).

Bake those two into a private-socket e2e test in the PR and they'll catch this class of bug before it reaches us. Ping when the three HIGH items are fixed and that e2e is green, and we'll re-review.

kunchenguid:
Merged - thank you for the careful work on this.

You took the design direction we converged on and implemented it faithfully: presence-gated `/afk`, the in-band sentinel marker, and keeping it orthogonal to yolo. Then you turned the review around thoroughly - the composer guard, the type-once submit, and the portable lock all hold up. I independently re-verified the human-in-the-pane and swallowed-Enter cases on a real host, and both are clean: deferred-then-separate-clean-lines, and exactly one digest after a swallowed Enter.

Special credit for building the private-socket e2e test from the methodology we shared - that kind of deterministic coverage is exactly what catches this class of bug before it ships. Appreciate the back-and-forth on getting the injection path right.

Do you have a handle on X? I'd love to post about this feature and acknowledge your contribution here!

PR #35, opened 2026-06-21
docs(contributing): document fork-based contributions

kunchenguid:
Document fork-based contributions now that no-mistakes **v1.30.1** supports fork routing: clone the parent (or set `origin` back to it), run `no-mistakes init --fork-url <your fork>`, and the gate pushes to your fork and opens the PR against the parent. Docs-only; raised as a direct PR (no pipeline).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,9 +14,9 @@ Dependency bots are exempt so their automation keeps working, but regular contri
 
 ## Workflow
 
-1. Fork the repo and clone your fork.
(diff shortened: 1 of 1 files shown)


kunchenguid:
thanks for this. quick process note - i only merge changes that come through no-mistakes itself, so the "PR must be raised via no-mistakes" check is red here because this was opened directly.

can you re-raise it through no-mistakes? run the pipeline on your branch and it'll open/update the PR through the gate, which turns that check green and makes it mergeable. the change looks useful, it just needs to come through the pipeline so it stays consistent with everything else.

ping me if the setup gives you trouble.

PR #45, opened 2026-06-23
fix(skill): Replace ':' with ';' in skill description

AnonymousMorris:
The SKILL.md description is parsed as YAML, and the : character in the plain-text description was being interpreted as YAML syntax. Removing it keeps the metadata parseable.

### DIFF
--- a/.agents/skills/afk/SKILL.md
+++ b/.agents/skills/afk/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: afk
-description: Enter away-mode supervision. Use when the user invokes /afk (e.g. "/afk", "/afk back in an hour", "going afk"). Sets a durable away-mode flag so the sub-supervisor daemon can self-handle routine wakes and escalate only captain-relevant events as one batched digest, cutting supervision token cost during walk-away stretches. Exit is automatic: any real (unmarked) message returns to full per-wake responsiveness.
+description: Enter away-mode supervision. Use when the user invokes /afk (e.g. "/afk", "/afk back in an hour", "going afk"). Sets a durable away-mode flag so the sub-supervisor daemon can self-handle routine wakes and escalate only captain-relevant events as one batched digest, cutting supervision token cost during walk-away stretches. Exit is automatic; any real (unmarked) message returns to full per-wake responsiveness.
 user-invocable: true
 ---
 


AnonymousMorris:
The colon cause this error in codex.

⚠ Skipped loading 1 skill(s) due to invalid SKILL.md files.
⚠ /home/morris/sandbox/ai/firstmate/.agents/skills/afk/SKILL.md: invalid YAML: mapping values are not allowed in this context at line 2 column 356

kunchenguid:
ty!

ki-za:

Proposed Options / Plan / Blocking / Lead decisions for this ticket, from the `ha-app-plan` scout report.

### Current state

- `view-state.ts:3-18,38-47` exposes one coupled `ReplyLayout = "sides" | "mirror"`; `App.svelte:38-40,150-153` reads and writes one `huddle-reply-layout` localStorage key.
- `HuddleView.svelte:175-200` uses that one value to change both placement and markers. `app.css:526-552` couples 86%-wide end placement to one rail and full width to paired rails.
- Presentation code and CSS still use `captain-*` identifiers and the visible label `SENT REPLY` (`HuddleView.svelte:177-184`). Issue #45's identity is `leadReply`; issue #52 requires lead detection from the human member.
- The current `margin-left: auto` (`app.css:531-535`) encodes left-to-right placement. No right-to-left fixture exists.
- The reserved frame already appears only around the composer and lead replies (`HuddleView.svelte:93-113,175-200`), matching the ownership ruling. Current colours use old `--lume-*` names instead of #49 role tokens.
- Review v3 made `sides` the live default and `mirror` the alternative (`docs/review/2026-08-29-view-v3.html:98-113`). Issue #45 later splits their two effects; issue #47 requires both controls in the menu. The fixed-presentation option no longer satisfies the later menu ruling.

### Options

- **Local presentation settings** — one global Huddle app choice for each setting; simplest authority and no theme precedence.
- **Theme defaults plus local overrides** — lets themes set initial presentation, but adds precedence and makes themes influence layout behavior.
- **Fixed presentation** — smallest code, but conflicts with #47's ruled menu contents.

### Recommended Plan

**Modules:** Extract `LeadReply.svelte` and reuse `ReplyFrame.svelte` for the composer and `leadReply`; replace `ReplyLayout` in `view-state.ts` with separate `ReplyAlignment` and `ReplyMarker` types and changes; add two controls to `AppMenu.svelte`; persist two keys through the shared settings model; migrate the old `huddle-reply-layout` value once; replace physical CSS placement with logical properties and #49 role tokens; extend reducer and component tests.

**Seams:** Immutable human-authored reply plus `reply_to` becomes one `leadReply`. `replyAlignment` changes placement only. `replyMarker` changes the indicator only. Neither setting changes message identity, content, attachments, target, or ledger order. The old `sides` value migrates to the ruled default alignment plus one marker; `mirror` migrates to in-place alignment plus paired markers. The theme supplies frame and secondary-text roles and may change glyphs without changing either setting. The huddle's human member supplies lead identity.

**Validation criteria:** The same message id, exact content, `reply_to`, attachment ids, and order survive every alignment/marker combination; member contributions never receive a `replyFrame`; the composer and `leadReply` retain the faint frame while a contribution is active; narrow, 30rem, wide, and right-to-left fixtures remain readable; a held-out theme changes glyphs and swatches without changing setting values; the old localStorage value migrates once. Run `bun run check`. Capture a setting-combination picture matrix at narrow and 30rem widths plus one right-to-left picture.

### Blocking?

`blocked by the huddle-mate setting-value ruling, #49, and #52` — the two settings have primary names but no frozen value names; role tokens and lead identity must replace current theme names and `captain` literals.

### Lead decisions

- **Settings authority: local Huddle app settings or theme defaults with local overrides?** Recommend local Huddle app settings.
- **Persistence scope: one Huddle app profile or one choice per Huddle?** Recommend one Huddle app profile.

PR #121, opened 2026-06-28
fix(fm-brief): bash 3.2 here-doc parse error breaks default brief scaffolding

scar-sh:
## Problem

On stock macOS Bash (GNU bash **3.2.57**, i.e. `/bin/bash`), `bin/fm-brief.sh` fails to parse:

```
```

This breaks the **default** task path: no-mistakes-mode briefs cannot be scaffolded at all on a stock-bash machine. It went unnoticed because modern bash (5.x) parses the file fine.

## Root cause

The no-mistakes (default-mode) Definition-of-done block was assigned via command substitution wrapping a here-doc:

```sh
DOD=$(cat <<EOF
...
Follow no-mistakes' own guidance for the mechanics: ...
...
EOF
)
```

That here-doc body contains an apostrophe (`no-mistakes'`). Bash 3.2's command-substitution parser mis-tokenizes the apostrophe inside `$(...)` as an opening single-quote and scans to EOF looking for its match. The `--scout`, `direct-PR`, and `local-only` here-docs happen to contain no apostrophe, so only the default path errored.

A quoted delimiter (`<<'EOF'`) does **not** fix it — the bug is in the `$(...)` command-substitution parser itself, not heredoc quoting (verified: an apostrophe in a `<<'EOF'` body inside `$(...)` still breaks bash 3.2). The same here-doc body parses fine at statement level.

## Fix

Move every mode's Definition-of-done out of `$(...)` and append it directly to the brief with `cat >> "$BRIEF" <<EOF`. The here-doc bodies are byte-identical to before — this is a parser-compatibility fix, not a content change. The assembled brief output is unchanged for all modes: same single blank line at the Project-memory → Definition-of-done join, and `$ID` interpolation plus escaped backticks preserved.

## Validation (stock Bash 3.2.57, `/bin/bash`)

`bash -n` before vs after:

```
# before
bin/fm-brief.sh: line 263: syntax error: unexpected end of file   (exit 2)
# after
$ /bin/bash -n bin/fm-brief.sh ; echo $?
0
```

Functionally scaffolded a complete, correct brief under `/bin/bash` for **every** mode and inspected the output:

- **no-mistakes** (default), **direct-PR**, **local-only**, **scout**, and a **--secondmate** charter all scaffold successfully.
- direct-PR / local-only driven via a temp registry (`data/projects.md`); unknown-project fallback (→ no-mistakes) also exercised.
- Confirmed the default-mode brief still contains the unchanged no-mistakes DoD text: the `/no-mistakes` invocation, `no-mistakes axi run --help`, the ask-user / `--yes` rules, and the `done: PR {url} checks green` line.
- Confirmed exactly one blank line at the Project-memory → Definition-of-done join, backticks rendered (not literal `` \` ``), and `$ID` interpolated correctly in Setup, Rule 1, and the local-only/direct-PR DoD.

Diff is limited to `bin/fm-brief.sh`.

### DIFF
--- a/bin/fm-brief.sh
+++ b/bin/fm-brief.sh
@@ -175,50 +175,15 @@ case "$MODE" in
   direct-PR)
     SETUP2=""
     RULE1='1. Never push to the default branch (push only your `fm/'"$ID"'` branch). Never merge a PR.'
-    DOD=$(cat <<EOF
-# Definition of done
-This project ships **direct-PR**: you raise the PR yourself, without the no-mistakes pipeline.
-The task is complete only when committed on your branch.
-When it is implemented and committed, push your branch and open a PR with \`gh-axi\`, then append \`done: PR {url}\` to the status file and stop.
-Do NOT run /no-mistakes. The captain reviews and merges the PR; firstmate relays it.
-EOF
-)
     ;;
   local-only)
     SETUP2=""
     RULE1="1. Never push to any remote and never open a PR. Work only on your \`fm/$ID\` branch; firstmate handles the merge into local \`main\`."
-    DOD=$(cat <<EOF
-# Definition of done
-This project ships **local-only**: no remote, no PR, no pipeline.
-The task is complete only when committed on your branch \`fm/$ID\`. Do NOT push, do NOT open a PR, do NOT merge.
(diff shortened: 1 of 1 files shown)


william-lubelski-mercury:
@scar-sh sorry my claude was trying to operate on my fork and it commented here.  I think this fix is valid

PR #123, opened 2026-06-28
feat: add kimi-cli harness adapter

nickolastheodoulou:
Adds detection, launch template, busy-signature, and MCP-config handling for the kimi-cli harness, plus tests.


kunchenguid:

Issue #106, opened 2026-06-27
Feature: pluggable out-of-band notifier (push) for captain-relevant wakes


cjunxiang:
## Summary

The **detection** side of "firstmate misses wakes between turns" is solved: #26 (watcher crash) and #27 (durable wake queue) ensure every captain-relevant wake is captured losslessly and drained at the start of the next supervisor turn. The **delivery** side is still session-bound: firstmate only *relays* a wake to the captain inside an active supervisor turn. When the captain is away or the supervisor session is idle, `done` / `needs-decision` / `blocked` / `failed` events sit correctly in the durable queue — but nothing pushes them to the captain out-of-band.

This proposes an **opt-in, pluggable wake-hook** that delivers captain-relevant events to an external channel **independent of supervisor-session liveness** — closing the delivery half of Section 9 ("reaches the captain immediately") in agents.md and complementing `/afk`.

## Observed gap

Hit this directly: during a quiet stretch the supervisor's watcher lapsed and ~3h of captain-relevant events (a finished audit; a task blocked on a decision) sat in the wake queue, surfaced only on a manual "how's it going?" poke. Nothing was lost (#27's durable queue worked), but nothing reached the captain until they returned to the terminal. This is the exact symptom #27's author described ("finished work goes unrelayed… until something pokes the supervisor") — #27 fixed *capture*; this addresses the out-of-band *reach*.

## Why it fits firstmate

-  Section 9 in agents.md contract is "reaches the captain immediately." Today that's only honored inside an active session. A push channel honors it when the session is idle.
- firstmate already treats "captain is away" as first-class (`/afk`, the away-mode daemon). An away captain currently has *no* channel to hear firstmate at all — push is the natural complement.
- Delivery is orthogonal to detection (which #27 solved), so this is a pure addition to the delivery layer — it touches none of the wake-capture logic.
- It matches the project's pluggable/opt-in philosophy (harness-adapters, delivery modes, `yolo`). The contribution is a **generic wake-hook extension point + adapter contract**, not a vendor-specific feature.

## Proposed shape (Phase 1, outbound-only)

- A tiny, **opt-in, fire-and-forget hook** in `fm-watch.sh`'s signal path — engineered so a notifier failure can never stall supervision (guarded, error-isolated, with a test asserting the watcher still wakes when the hook errors).
[...]

PR #127, opened 2026-06-28
feat(email): two-way email notifier (out-of-band, opt-in)

2jfhas2bn2jfds:
## Two-way email notifier for firstmate (out-of-band notifier, opt-in)

Implements the pluggable out-of-band notifier from #106: firstmate emails the captain when something captain-relevant comes up, and the captain can reply by email to steer it. Modeled closely on X mode (section 14) so it is purely additive and opt-in.

### Opt-in and additive (zero behavior change when off)
Gated on `.env`: `FM_EMAIL_NOTIFY=1` plus `AGENTMAIL_API_KEY`, `FM_NOTIFY_EMAIL` (recipient), and `AGENTMAIL_INBOX`. With the switch absent/falsey every script is a hard no-op and bootstrap writes/prints nothing. No edits to the watcher backbone (`fm-watch.sh`, `fm-watch-arm.sh`, `fm-wake-lib.sh`, the afk daemon); it rides the existing `state/*.check.sh` mechanism and the shared captain-relevant classifier (`fm-classify-lib.sh`).

### Outbound (`bin/fm-email-notify.sh`)
Scans `state/*.status` with the same classifier the watcher uses and emails the captain on `needs-decision`/`blocked`/`failed`/`done`/`PR ready`/`checks green`/`merged` via AgentMail `POST /inboxes/<inbox>/messages/send`. Plain outcome language (no internal jargon, mirrors section 9), deduped per task in `state/.email-seen-<task>` (marker advances only after a successful send, so failures retry). Silent: never adds a wake.

### Inbound / two-way (`bin/fm-email-poll.sh`)
Polls `GET /inboxes/<inbox>/messages`, baselines on first run (never replays pre-existing mail), and surfaces each new reply from the captain as an `email-reply <id>` `check:` wake with the full reply text stashed to `state/email-inbox/<id>.json` as the payload, so a reply can steer firstmate. Config/auth errors surface one rate-limited `email-mode-error` line.

### Secrets and untrusted input
Key and recipient come only from gitignored `.env`; the captain's address is never hardcoded in tracked files. Bearer key goes to a 0600 header file, never on argv, never echoed. Inbound email text is only ever parsed by jq from a file (never inlined into a shell), and message ids are sanitized before becoming filenames.

### Bootstrap, cadence, docs, tests
Bootstrap drops an idempotent `state/email-watch.check.sh` shim (silent notify then poll, one shim so outbound always runs each cadence) and `config/email-mode.env` (60s cadence), removed on opt-out. `FM_EMAIL_DRY_RUN` previews notifications to `state/email-outbox/` with no key or network. Documented in AGENTS.md section 15 and README; `config/email-mode.env` added to `.gitignore`.

`tests/fm-email-mode.test.sh`: 19 hermetic tests with the AgentMail HTTP call faked (no real network or keys) covering inert-by-default, classify/compose, dedupe, live send, baseline, inbound parse, sender filtering, id sanitization, error reporting, and bootstrap activation/opt-out.

### Validation
`shellcheck bin/*.sh tests/*.sh` clean; the new suite and the rest of the suite pass (two unrelated tests fail only under the local macOS bash 3.2 due to a pre-existing `fm-brief.sh` heredoc issue; they pass on CI's Ubuntu bash 5).

Delivered via direct-PR from a fork because the no-mistakes gate is currently not creating runs. Captain merges.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ data/
 .env
 config/crew-harness
 config/x-mode.env
+config/email-mode.env
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -70,9 +70,10 @@ README.md            public overview and development notes
 .agents/skills/      shared skills, committed
 .claude/skills       symlink to .agents/skills for claude compatibility
 bin/                 helper scripts, committed; read each script's header before first use
-.env                 optional X-mode pairing token; LOCAL, gitignored; presence-gates section 14
+.env                 optional X-mode pairing token and/or email-mode credentials; LOCAL, gitignored; presence-gates sections 14 and 15
 config/crew-harness  crewmate harness override; LOCAL, gitignored; absent or "default" = same as firstmate
 config/x-mode.env    generated X-mode watcher cadence; LOCAL, gitignored; source before arming watcher when present
+config/email-mode.env generated email-mode watcher cadence (60s); LOCAL, gitignored; source before arming watcher when present (section 15)
 data/                personal fleet records; LOCAL, gitignored as a whole
   backlog.md         task queue, dependencies, history
   captain.md         captain's curated personal preferences and working style; LOCAL, gitignored, and canonical even if harness memory mirrors it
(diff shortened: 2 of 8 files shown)


jorgenseeyou:
Reviewed at `ccdf369179742a4dcdab38150cfa85035240c694`, against merge-base `be6e88647c5631bfd36f5b7338213f9b8fb74cac`. The prune is sound and most of it is clearly right: the courier entry really is outdated by the code, the two sign-in entries really did duplicate each other, and the sweep counts really were stale.

Two entries look like they should stay, because the conditions they warn about are still true. Both are documentation-only; nothing here asks for a code change in this PR.

**1. The `OTEL_ABSENCE` warning is dropped while the control is still red.**

The shortened failure overview at H:1295–1323 no longer names it; it stood at B:1508 and B:1533. The control was run twice on this head with Node v22.23.1 and failed both times at `the completing task intent is accepted` (`platform/controls/otel_absence_control.mjs:202`).

This is not a stale stage count. The control starts the real platform server on `HOST=127.0.0.1`, `PORT=0`, reads the port from the listen line, disables `.env` loading, and drives `/v1/task-intents` over HTTP. The stage register runs that same file at `platform/controls/run_merged_regression.mjs:272`.

H:78–79 still cites the control as evidence for the tracing-absence property, without saying it does not currently pass. Keeping a short named line about the standing failure, with the measured assertion, would prevent the next reader from classifying an existing red as a new regression. No sweep numbers needed.

**2. The deleted island list also carried a live mock warning.**

H:563–573 reinstates the `Instrument`, `SpellHint` and `TextSizeDial` warnings but loses the one about `JobsChoices`, which stood at B:644–646. `JobsChoices`, `WhatsApp` and `chToggle` no longer appear anywhere in the file.

The behaviour is unchanged between base and head. `frontend/components/screens/JobsChoices.tsx:18` creates local `whatsappOn` with `useState(true)`; the handler at :51–55 calls only `setWhatsappOn(next)` and `toast(...)`, with no `setSetting`, no door and no transport. The other handler in the same component does call `setSetting("sy-autonomy", ...)` at :29, and the tooltip at :45 still says the channel is not connected. So the toast reads "notifications on" while no notification capability is enabled.

Re-adding one short line beside the other island warnings, stating that the WhatsApp choice is local simulation with no persistence or delivery, would keep that boundary visible. `grep -rl '"use client"' frontend/` regenerates the file list but not this warning.

Two smaller notes, neither blocking:

- The wording check refuses three role attributions at H:979, H:1147 and H:1219. They predate this PR, but it edits those same lines to remove the dates, so they could be switched to `owner decision` cheaply while they are open.
- H:1294 describes a path that has never existed in this repository as "environment absence wearing a defect's clothes", while H:1303–1305 correctly says the two toolchain stages are still FAIL when their registered hold signature does not arrive. Worth keeping cause and runner verdict separate.

PR #163, opened 2026-06-30
Fix watcher stat mtime probing

benawad:
## Summary
- add a shared validated stat helper for watcher/guard/wake mtime and signature reads
- prefer BSD /usr/bin/stat on Darwin while rejecting nonnumeric GNU filesystem-stat output
- fix BSD fm_path_sig to use integer %m instead of float %Fm and add Darwin coverage

## Validation
- no-mistakes run 01KWD0AC18D09ZSFGXFAWT21TJ passed review, test, document, and lint on head 4f6f939c1d2a6b13774ec57fb8ce2dafc99e13c0
- no-mistakes push to upstream failed with 403 for benawad on kunchenguid/firstmate, so this PR is opened from the authorized fork benawad/firstmate
- manual pre-gate validation also passed: bash -n bin/*.sh tests/*.sh; for test_script in tests/*.test.sh; do bash ""; done

## Pipeline

### DIFF
--- a/bin/fm-guard.sh
+++ b/bin/fm-guard.sh
@@ -47,13 +47,6 @@ if [ -n "$tangle_branch" ]; then
   } >&2
 fi
 
-# Portable mtime; see fm-watch.sh for why the `stat -f || stat -c` fallback breaks on Linux.
-if [ "$(uname)" = Darwin ]; then
-  stat_mtime() { stat -f %m "$1" 2>/dev/null; }
-else
-  stat_mtime() { stat -c %Y "$1" 2>/dev/null; }
-fi
-
 # Only act with tasks in flight; count them so the banner can say how much is
 # riding on an absent watcher.
 in_flight=0
@@ -71,8 +64,7 @@ BEAT="$STATE/.last-watcher-beat"
 watcher_fresh=false
 beacon_desc=never
 if [ -e "$BEAT" ]; then
-  m=$(stat_mtime "$BEAT")
-  if [ -n "$m" ]; then
+  if m=$(fm_path_mtime "$BEAT"); then
     age=$(( $(date +%s) - m ))
     beacon_desc="${age}s ago"
     [ "$age" -lt "$GRACE" ] && watcher_fresh=true
--- a/bin/fm-stat-lib.sh
+++ b/bin/fm-stat-lib.sh
@@ -1,0 +1,46 @@
+#!/usr/bin/env bash
+# Shared stat helpers. Always validate numeric output before callers use it in
+# arithmetic; PATH may contain GNU stat even on Darwin.
+
+fm_is_uint() {
+  case "${1:-}" in
+    ''|*[!0-9]*) return 1 ;;
+    *) return 0 ;;
+  esac
+}
+
+fm_stat_first_line() {
+  sed -n '1p'
+}
+
+fm_path_mtime() {
+  local path=$1 out
+  if [ "$(uname)" = Darwin ]; then
+    if [ -x /usr/bin/stat ]; then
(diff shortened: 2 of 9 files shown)


nguyenhieuec (commented):
Review result: code-clean / merge-ready from local review, but I cannot merge with the current GitHub token.

Evidence checked:
- Head: 4f6f939c1d2a6b13774ec57fb8ce2dafc99e13c0
- Focused validation passed locally: bash -n on touched scripts/tests, tests/fm-watcher-lock.test.sh, tests/fm-wake-queue.test.sh, tests/fm-watch-triage.test.sh.
- The prior merged watcher stat PR #28 is not a duplicate blocker; this PR is a follow-up that centralizes validated stat helpers and covers wake/supervise paths plus Darwin PATH ordering.
- Current PR body contains the required no-mistakes marker. The failed no-mistakes workflow appears stale from an older PR body; I could not rerun it because GitHub requires repo admin rights.

Blocker:  failed with . A maintainer needs to rerun/clear the stale check and merge, or grant/use a token with merge permission.

Verdict: Comment / merge-ready pending maintainer permission and stale check handling.

kunchenguid:

butlersrepos:
## Session pre-read is assembled

`data/ow-163-tree-session-prep/report.md`, with captures in `img/` beside it. Knowledge only — no code changed.

**Both boards are now shown at native 1152×648.** The Paladin frame comes from the `ui-convergence-v1` audit run (`2b49eb61`). **The Druid frame was missing from that run** — the audit's shot list has exactly one talent entry and it stands the Paladin — so it was captured in a single headed pass with the harness's own `--only=talent_tree --class=druid` flag, into a separate directory that did not overwrite the audit frames.

### The three things worth knowing before the session

**1. The board is 9 grid nodes.** Verified: 4 Blight, 0 Conjuration, 1 Wildshape, 1 Sword & Board, 2 Hammer Thrower, 1 Auramancer. One built capstone in the whole game. Five ghost frames. It will never be cheaper to re-style than it is now.

**2. The two colour signals are already double-booked, before feats arrive.**
- `ModSatellite` sets its border to `CAPSTONE_FRAME_COLOR` **unconditionally** — every modifier wears amber whether or not its host is a capstone. So amber means *capstone* **and** *ability modifier*.
- An ability node and a scaling-modifier satellite cut the **same atlas cell** (`Rect2(4347, 1758, 52, 51)`) with the same material. Slate means *ability* **and** *scaling modifier*.
- The entire type taxonomy currently rests on **size** — 100 / 75 / 40 / 34 — and size is what a satellite gives up in order to orbit.

**3. An independent reader already fails to read the taxonomy.** The `ui-convergence-v1` judge scored this screen 2.86 / `hobby` — lowest in the game — and described the modifier satellites as *"the gold candelabra sprites attached to frame corners carry no legend."* That is the legible-without-being-named problem failing in the wild at four node types, before feats make it five.

One structural note the report argues at length: **type lives in the frame, state lives in the icon alpha, and locking leaves the frame at full strength.** If a redesign preserves that split, ten states is five frames plus one fill rule — not ten bespoke looks.

### Also found, independent of the session

**`BlightGraspingSkill.tres` draws no icon.** Its `region = Rect2(176, 112, 16, 16)` on the TrueHeroes sheet is **entirely transparent — 0 of 256 opaque pixels** (siblings sample 97–116). On the shipped Druid board it renders as an empty amber frame with an *earned* amber tether running into it, so the board says "you own this" and shows nothing. Two-value fix; a gate test asserting every achievement icon region has at least one opaque pixel would close the class permanently.

**The board's locked state has never been photographed** — `_build_talent_tree` calls `_unlock_everything()` before every shot, so the locked language the redesign is mostly about is in no committed capture.

### Open questions

Stated as questions and deliberately unanswered — 11 of them in §6, plus a separate §7 of firstmate opinions kept out of them. Each is registered as a durable captain-held decision so none is settled by omission.

PR #173, opened 2026-07-01
fix(brief): remove apostrophe breaking bash -n and guard bash 3.2 set -u in spawn

nilushan:
## Intent

Fix GitHub issue #166: bin/fm-brief.sh failed bash -n due to an unescaped apostrophe inside a heredoc nested in a $(cat <<EOF...EOF) command substitution in the no-mistakes-mode Definition-of-done block. Reworded the sentence to remove the apostrophe, audited the other two similar DOD heredoc blocks (direct-PR, local-only) for the same issue (none found), verified bash -n passes and all 5 brief-generation modes produce clean output, and added tests/fm-brief.test.sh as a regression guard. This rerun picks the pipeline back up after two GitHub push-permission blockers: first, no write access to origin kunchenguid/firstmate (resolved by the captain creating a fork at nilushan/firstmate and configuring --fork-url); second, the fork push over SSH failed because this machine's default SSH key is bound to a different GitHub account (nilushan-bluekey) than the fork owner (nilushan); resolved by switching the fork remote to HTTPS, which uses gh's correctly-authenticated nilushan credential via the osxkeychain helper (verified with git ls-remote before retrying).

## What Changed

- Removed an unescaped apostrophe from the no-mistakes-mode Definition-of-done heredoc in `bin/fm-brief.sh`, which broke `bash -n` on the generated script (fixes #166); audited the direct-PR and local-only DOD heredocs for the same issue and found none.
- Guarded `bin/fm-spawn.sh`'s batch dispatch against an empty `shared_args` array under bash 3.2's `set -u`, which previously crashed with an unbound-variable error.
- Added `tests/fm-brief.test.sh` as a regression test covering all brief-generation modes, and listed it in `CONTRIBUTING.md`'s test suite inventory.

## Risk Assessment

✅ Low: Small, well-scoped diff: one wording fix that removes the offending apostrophe, one correct bash-3.2-safe empty-array guard consistent across both call sites, a new regression test that accurately matches fm-brief.sh's mode-routing logic, and a docs line — verified bash -n passes and no other risky patterns introduced.

## Testing

[...]

Issue #166, opened 2026-07-01
firstmate's fm-brief.sh scaffold script has a bug


yasnaya-polyana:
firstmate's fm-brief.sh scaffold script has a real bug (confirmed with bash -n), so if you ask the
  first-mate session to brief a new task itself, it'll likely hit the same crash — I worked around it by hand-writing the brief this time, but
  the first-mate session won't know to do that unless told.

PR #174, opened 2026-07-01
fix(fm-brief): unbalanced apostrophe breaks ship-brief scaffolding

moxious:
## Bug

Scaffolding a **ship** brief with `bin/fm-brief.sh` fails outright:

```
```

The script does not parse, so `fm-brief.sh <id> <repo>` (ship path) errors and writes nothing. Scout briefs use a different code path and are unaffected.

## Root cause

In the no-mistakes ship-brief branch (the `*)` default of the `case`), the definition-of-done text is built inside a command-substitution heredoc:

```sh
DOD=$(cat <<EOF
...
Follow no-mistakes' own guidance for the mechanics: ...
...
EOF
)
```

The lone apostrophe in `no-mistakes'` is read by bash's command-substitution parser as an opening single quote that is never closed, so the whole `$( ... )` — and the rest of the script — fails to tokenize.

## Fix

Reword `no-mistakes' own guidance` -> `the no-mistakes guidance` so no unbalanced single quote remains in that heredoc. Minimal one-line change; no restructuring. Grepped the file for other lone/unbalanced single quotes inside `$( ... )` heredocs — line 211 was the only one (other apostrophes are in comments, in single-quoted assignments, or in the plain `cat > "$BRIEF"` redirect heredoc, which is not parsed the same way).

## Verification

- `bash -n bin/fm-brief.sh` now exits 0.
- `bin/fm-brief.sh <throwaway-id> grafana-recommender` (ship, no-mistakes mode — the previously-broken path) prints `scaffolded: ...` and writes a valid brief; throwaway data dir removed afterward.
- `--scout` path still scaffolds correctly.
- Generated ship-brief text reads sensibly after the reword.

Note: related to #166 and overlaps with #173 (which also touches this apostrophe among other changes).

### DIFF
--- a/bin/fm-brief.sh
+++ b/bin/fm-brief.sh
@@ -208,7 +208,7 @@ When you believe it is complete, append \`done: {summary}\` to the status file a
 Firstmate will then instruct you to run /no-mistakes to validate and ship a PR.
 
 You drive no-mistakes by responding to its gates, not by implementing fixes.
-Follow no-mistakes' own guidance for the mechanics: it loads when you invoke /no-mistakes, and \`no-mistakes axi run --help\` plus the \`help\` lines in each \`axi\` response are authoritative and version-matched to the installed binary.
+Follow the no-mistakes guidance for the mechanics: it loads when you invoke /no-mistakes, and \`no-mistakes axi run --help\` plus the \`help\` lines in each \`axi\` response are authoritative and version-matched to the installed binary.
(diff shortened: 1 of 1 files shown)


moxious:
Closing as a duplicate of #173, which covers the same `fm-brief.sh` apostrophe fix and additionally guards `bash 3.2 set -u` in spawn. Thanks!

PR #176, opened 2026-07-01
fix(spawn): wrap launch command in bash -c for non-bash pane shells

benawad:
## Summary
- Every harness's launch command in `launch_template()` used bash-only `"$(cat __BRIEF__)"` command substitution, which fails outright when the treehouse worktree's interactive pane shell is fish instead of bash/zsh.
- Wraps the fully assembled launch string in `shell_quote()`'d `bash -c` right before the final `tmux send-keys`, so the pane's own shell only ever needs to parse `bash -c '<string>'` regardless of whether it's fish, zsh, or bash.
- Also fixes a portability bug the pipeline caught along the way in the same area (Darwin `stat -f` vs `stat -c` detection).

## Test plan
- [x] Verified via a standalone repro mirroring `shell_quote()`/`launch_template()` logic across fish, bash, and zsh for all five harness template shapes (including nested quotes and the secondmate `FM_HOME=` env-prefix case)
- [x] Pipeline review, test, document, and lint steps all passed clean

### DIFF
--- a/bin/fm-guard.sh
+++ b/bin/fm-guard.sh
@@ -48,10 +48,10 @@ if [ -n "$tangle_branch" ]; then
 fi
 
 # Portable mtime; see fm-watch.sh for why the `stat -f || stat -c` fallback breaks on Linux.
-if [ "$(uname)" = Darwin ]; then
-  stat_mtime() { stat -f %m "$1" 2>/dev/null; }
-else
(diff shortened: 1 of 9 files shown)


benawad:
Closing — this PR was opened manually, bypassing the pipeline, so it can't pass the required no-mistakes check. Reopening properly now that the gate is configured to push through the fork.

PR #177, opened 2026-07-01
Fix fmx skill metadata quoting

Whamp:
## Summary
- Quote the `fmx-respond` skill description as a folded YAML scalar.
- Fixes YAML parsing when the description contains `check: wake`.

## Verification
- `python3` YAML parse check over every `.agents/skills/*/SKILL.md` front matter.

Note: no-mistakes was not available in this environment.


Whamp:
Without this quoting, it violates the agent-skills.io spec and throws warning in pi:

```bash
desktop firstmate main
❯ pi
[Skill conflicts]
  ~/firstmate/.agents/skills/fmx-respond/SKILL.md
    Nested mappings are not allowed in compact mappings at line 2, column 14:

description: Agent-only playbook for handling an X mention in X mode. Use on an…
             ^
```

kunchenguid:

kunchenguid:

PR #202, opened 2026-07-03
fix(backends): reuse the herdr firstmate workspace instead of leaking one per spawn

Bre77:
## The leak

On the experimental **herdr** runtime backend, every crewmate left an orphaned herdr *workspace* behind. `herdr workspace list` steadily filled with many `label=firstmate` workspaces — effectively one per task, most belonging to already-torn-down crewmates. The documented design is one persistent `firstmate` workspace per session with one tab per task (mirroring tmux's one session / one window per task), so these should never accumulate.

## Root cause

Not a flaw in the find-or-create flow — a **`jq` reserved-keyword collision**.

`fm_backend_herdr_workspace_find` filtered with `jq --arg label "$X" '... select(.label == $label) ...'`. But `label` is a reserved keyword in `jq` (used by `label $out | … break $out`), so `$label` makes the whole filter a **compile error**, not a match. The function discarded `jq`'s stderr with `2>/dev/null`, so the error surfaced as empty output: the find returned nothing on every call, `fm_backend_herdr_workspace_ensure` always took the create path, and a fresh `firstmate` workspace was minted on every spawn.

The same collision silently disabled two other filters that used `$label`: the `create_task` duplicate-tab-label check and `resolve_bare_selector`'s tab lookup.

## The fix

- Rename the `jq` variable to `$want` (a non-keyword) in all three affected filters (`workspace_find`, the `create_task` duplicate check, `resolve_bare_selector`). Reuse, duplicate detection, and bare-selector lookup all work again.
- With reuse restored, the single `firstmate` workspace is **persistent** (like tmux's session). Teardown (`fm_backend_herdr_kill`) still closes only the task's pane/tab and correctly leaves the workspace in place — so repeated spawn/teardown cycles no longer accumulate workspaces. No workspace-deletion logic is needed.
- **Default tab:** `herdr workspace create` seeds a new workspace with an auto-created default tab (label `1`) firstmate never uses. `workspace_ensure` now prunes it (best-effort) right after creation, so the workspace holds only real `fm-<id>` task tabs.
- **Project-labelled-workspace anomaly** (e.g. a workspace labelled `python-teslemetry-stream`): the adapter always creates with the fixed `--label firstmate` and never derives a label from the project, so such a workspace is not adapter-created; firstmate matches only the `firstmate` label / `fm-` tab labels and leaves it untouched. Documented rather than "fixed".

## Tests

Extended `tests/fm-backend-herdr.test.sh` with a **stateful fake herdr CLI** that models herdr's real container behaviour (workspace-create seeds a default tab; `pane close` closes the tab; live `workspace/tab/pane list`). New assertions:

- repeated spawn→teardown cycles reuse a **single** `firstmate` workspace, leave **zero** orphaned workspaces, tear down every task tab, and invoke `workspace create` **exactly once**;
- the auto-created default tab is pruned on workspace creation.

The full existing suite passes (`bin/backends/herdr.sh` is shellcheck-clean).

**Real-CLI confirmation** (isolated `HERDR_SESSION`, guarded cleanup — never the `default` session): the pre-fix code failed the smoke idempotency check, minting a second workspace on the second `container_ensure` (`wE` then `wF`); the fixed code passes the real-herdr smoke test 9/9.

## Docs

`docs/herdr-backend.md` updated: the Kill row now states the workspace is persistent and not removed at teardown, plus a new default-tab-prune row and a "Workspace lifecycle" section covering the reuse model, the jq-keyword root cause and a guard against it, the default-tab handling, and the project-labelled-workspace anomaly.

### DIFF
--- a/bin/backends/herdr.sh
+++ b/bin/backends/herdr.sh
@@ -181,11 +181,15 @@ fm_backend_herdr_workspace_find() {  # <session>
 # fm_backend_herdr_workspace_prune_default_tabs: close the tab herdr
 # auto-creates inside a freshly created workspace (label "1", never an
 # fm-<id> task tab), so the persistent "firstmate" workspace holds only real
-# task tabs - the documented one-tab-per-task shape. Runs only on the create
-# path (find already reuses an existing workspace), when the ONLY tab present
-# is that auto-created default, so filtering to non-fm- tabs never touches a
-# task tab. Best-effort: a failure here never fails the spawn, mirroring the
-# fm_backend_herdr_kill `|| true` contract.
+# task tabs - the documented one-tab-per-task shape. Best-effort: a failure
+# here never fails the caller, mirroring the fm_backend_herdr_kill `|| true`
+# contract.
+#
+# Verified real-herdr behavior (not modeled by the fake-CLI unit tests):
+# closing a workspace's LAST remaining tab deletes the whole workspace, not
+# just the tab. So this must never run while the auto-created default tab is
+# still the ONLY tab - callers only invoke it once at least one other (real
+# task) tab exists alongside it, never right after workspace creation.
 fm_backend_herdr_workspace_prune_default_tabs() {  # <session> <workspace_id>
   local session=$1 wsid=$2 tabs tab_id pane_id
   tabs=$(fm_backend_herdr_cli "$session" tab list --workspace "$wsid" 2>/dev/null) || return 0
@@ -218,8 +222,11 @@ fm_backend_herdr_workspace_ensure() {  # <session> <cwd>
   wsid=$(printf '%s' "$out" | jq -r '.result.workspace.workspace_id // empty' 2>/dev/null)
   [ -n "$wsid" ] || return 1
   # Herdr seeds a new workspace with one auto-created default tab firstmate
-  # never uses; drop it so only task tabs remain. Best-effort.
-  fm_backend_herdr_workspace_prune_default_tabs "$session" "$wsid"
+  # never uses. It is NOT pruned here: at this instant it is the workspace's
+  # ONLY tab, and closing a workspace's last tab deletes the workspace itself
+  # (verified against the real herdr binary) - pruning here would destroy the
+  # workspace we just created. fm_backend_herdr_create_task prunes it instead,
+  # once the first real task tab exists alongside it.
   printf '%s' "$wsid"
 }
 
@@ -241,9 +248,13 @@ fm_backend_herdr_container_ensure() {  # <cwd-for-a-fresh-workspace>
 # label), so the duplicate check is ours, mirroring tmux's manual check.
 # --no-focus: verified tab create never focuses by default regardless of
 # sibling tabs, so this is defense in depth rather than a behavior change.
+# If <container>'s workspace was JUST created (its only tab beforehand was
+# herdr's auto-created default, label "1"), prune that default tab now that
+# this call has added a real tab alongside it - see
+# fm_backend_herdr_workspace_ensure for why pruning cannot happen any earlier.
 # Echoes "<tab_id> <pane_id>" on success.
 fm_backend_herdr_create_task() {  # <container> <label> <cwd>
-  local container=$1 label=$2 cwd=$3 session wsid list dup out tab_id pane_id
+  local container=$1 label=$2 cwd=$3 session wsid list dup out tab_id pane_id only_default
   session=${container%%:*}
   wsid=${container#*:}
   list=$(fm_backend_herdr_cli "$session" tab list --workspace "$wsid" 2>/dev/null) || return 1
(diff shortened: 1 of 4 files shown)


kunchenguid:
Validated end to end, including a real jq 1.6 reproduction and a real-herdr smoke run in an isolated session (never the default session).

**jq 1.6 repro (the core claim)**: confirmed at three levels using jq 1.6 obtained from nixpkgs nixos-22.05.
1. The bare jq filter: `--arg label` + `$label` is a compile error on jq 1.6; `--arg want` + `$want` is not.
2. The actual function bodies: main's `fm_backend_herdr_workspace_find` returns empty against a fixture with a matching workspace; this branch's version correctly returns the match.
3. The full suite: main's own herdr test file, run under jq 1.6, fails at the exact workspace-reuse assertion this PR describes; this branch's suite passes 39/39 under jq 1.6.

**One real defect found and fixed**: the real-herdr smoke test caught something the fake-CLI unit tests couldn't model - closing a workspace's only remaining tab deletes the whole workspace on real herdr. The new default-tab prune ran immediately after `workspace create`, when the default tab was the workspace's only tab, so every spawn destroyed the workspace it had just created. Fixed by deferring the prune to `fm_backend_herdr_create_task`, right after the first real task tab exists alongside the default tab. Also added a regression guard test that greps `bin/` for any jq `--arg`/`--argjson` name colliding with a jq reserved keyword, so this class of bug fails loudly instead of silently returning empty.

**Suite results**: full suite (33 files, 486 assertions) green with the installed jq 1.8.1. Real-herdr E2E in an isolated session covers container reuse, the default-tab prune, the multi-workspace (2ndmate-*) creation/prune path, and normal teardown - all pass; no interaction with the default session at any point.

**Docs**: trimmed the new docs/herdr-backend.md section down to current-state facts per this project's convention (history stays in commit messages, not the doc).

Verdict: the described bug is real and reproducible, the fix is correct, and the branch is green top to bottom - ready for review/merge.

kunchenguid:
Follow-up amendment: fixed two staleness issues in the docs trim - the workspace-lifecycle write-up now describes per-home labeling (firstmate for the primary, 2ndmate-<id> for a secondmate) instead of a hardcoded firstmate constant, and dropped the contributor-environment-specific anomaly example for a generic one-liner. Also fixed a wrong test-file reference. Full suite (33 files, 486 assertions) still green.

harris-chiu (approved):
QA verified all three fixes: carve wording corrected at international-tal/index.html:716 and :852, Coverage tooltip, opera.com note. CI green.

pedromuller-del:
## Correction round 2 — new head `3baf7008`

The exact-head review flagged the marker contract as incomplete end to end. This round closes it. The live-wait classification (`bin/fm-pipeline.sh`) is byte-for-byte unchanged from the prior head; only `bin/fm-watch.sh` and the two test files changed.

### What changed

The wait-premise occurrence marker publisher now stages a unique mode-0600 single-link file on the state directory filesystem, writes and fsyncs it, validates it on its open descriptor, then publishes it with `rename(2)`. `rename(2)` does not follow the destination, so a killed or failed write leaves the prior marker intact and never exposes a torn or empty marker through the marker path; the staged file is unlinked, never published. An untrusted or legacy marker at the path (a pre-planted symlink, a shared hardlink, or a base-created mode-0644 marker) is refused and left in place for the consumer to refuse.

The consumer (`shadow_wait_correction`) now applies the matching validation before use or removal: it opens the marker with `O_NOFOLLOW`, validates the descriptor as a regular single-link mode-0600 file on the state device, and only then reads the row, emits the correction, and removes the marker. A marker that fails this contract is refused before use or removal, so bytes from a refused symlink or hardlink, or from a base-created mode-0644 legacy marker, never become a correction record. `unlink(2)` never follows the path, so a marker swapped between validation and removal cannot drag an attacker-selected target into the correction log.

The five new regression comments that named the fixed head as red now say "before the fix" to match execution.

### Red-before-green

A disposable red snapshot held the exact base (`4f271d88`) production `bin/fm-watch.sh` and `bin/fm-pipeline.sh` under the new tests.

Against the exact base production:
- `test_wait_premise_marker_is_private_file` fails: the plain `>` write never sets mode 0600.
- `test_wait_premise_consumer_rejects_hardlink` (isolated) fails: the consumer emitted a correction record from a refused hardlink marker; the attacker row reached `wait-corrections.log`.
- `test_wait_premise_consumer_rejects_legacy_mode_0644` (isolated) fails: the consumer emitted a correction record from a base-created mode-0644 legacy marker.
- `test_working_does_not_contradict_keyed_wait` and `test_parked_does_not_contradict_keyed_wait` fail (missing `probe=unknown`), as established in the prior round.

At the new head:
```
tests/fm-wait-premise-shadow.test.sh: 6 ok
tests/fm-pipeline.test.sh: 19 ok
FM_TEST_SUMMARY total=2 failed=0 skipped_gate=0
```

### Local gates

- `bin/fm-lint.sh` (ShellCheck 0.11.0, full extended analysis) on `bin/fm-watch.sh`, `bin/fm-pipeline.sh`, and both test files: clean.
- `bash -n` on all four files: clean.
- `git diff --check` exact-base-to-head: clean.
- `bin/fm-doc-audience-check.sh`: ok surfaces=160 local_links=349. No documentation path changed.

Only `bin/fm-watch.sh` and the two test files changed this round. `bin/fm-pipeline.sh` is byte-for-byte identical to the prior head. No workflow path changed. Hosted Actions were not used as a quality gate, per the review contract. The broad family suite was not held as a gate; the focused suites covering both changed surfaces are green.

Note: the PR description could not be updated because the authed token lacks `UpdatePullRequest` permission, so this correction summary is posted as a comment.

PR #264, opened 2026-07-05
fix: record real leased worktree in fm-spawn meta; harden fm-teardown

prodempsey:
# fix: record real leased worktree in fm-spawn meta; harden fm-teardown

## Summary

`bin/fm-spawn.sh` recorded the wrong worktree path in every crew's meta, which broke teardown, merge/PR detection, and branch resolution and forced hand-teardown of every crew.
This change replaces the fragile pane-cwd polling heuristic with an authoritative `treehouse get --lease` acquire, whose stdout is exactly the leased worktree path, and hardens `bin/fm-teardown.sh` so a missing, stale, or non-treehouse worktree in the meta degrades gracefully instead of bricking teardown.

## Root cause

The old spawn path sent `treehouse get` into the task pane and then polled tmux's `#{pane_current_path}`, taking the first value that differed from the project path as the worktree.
That is a fork/exec race: between sending `treehouse get` and the treehouse subshell actually entering the pooled worktree, the pane can transiently report the session's default working directory - the firstmate home where `fm-spawn` itself runs.
The poll latched onto that spurious transient cwd, so `worktree=<firstmate home>` was written into the meta instead of the real leased worktree.
Downstream, `fm-teardown` then ran `treehouse return` on a non-pool path (refused), and merge/PR detection and branch resolution operated on the wrong directory, so every crew had to be torn down by hand.

## The fix

`fm-spawn` now leases the worktree out-of-band with `treehouse get --lease --lease-holder "fm-<id>"`, run from the project so treehouse resolves the correct pool.
`--lease` prints only the leased worktree's absolute path to stdout (all banners go to stderr), so the captured value is authoritative rather than scraped.
If no worktree is leased, spawn exits 1 immediately - no 60s poll, no timeout, no chance of a spurious path.
The pane is then moved into the leased worktree with an explicit `cd <worktree>` send.
The leased worktree is protected from `treehouse prune` for the task's lifetime, and teardown's existing `treehouse return --force "$WT"` releases the lease exactly as it did for an interactively-acquired worktree.
The secondmate and Orca spawn paths are unchanged.

### Why `--lease` beats a hardened poll

A hardened poll (longer timeout, cwd sanity checks) would still be inferring the worktree from an observable side effect that races with the pane.
`--lease` removes the inference entirely: treehouse returns the path it reserved, so there is exactly one source of truth and no window in which a transient cwd can be mistaken for the worktree.
It also makes the reservation durable against `treehouse prune`, which the poll never did.
This design - authoritative acquire plus fail-fast exit 1 - is Captain-approved.

## Teardown hardening

`fm-teardown` adds `safe_task_worktree()`, which permits branch cleanup and `treehouse return` only when the recorded worktree is a real, registered worktree of the task's project AND is neither the firstmate repo root nor the active firstmate home.
The second guard matters for firstmate-on-itself tasks whose meta was written before this fix: such a meta could record the runtime home (itself a genuine, registered worktree of the firstmate repo) as the worktree, and teardown must never detach, branch-delete, or hand the runtime home to `treehouse return`.
Teardown now also degrades gracefully: a missing, stale, or non-treehouse worktree warns on stderr and continues - the window is still killed and volatile state cleared - instead of aborting when `treehouse return` refuses an unmanaged path.
A failed `treehouse return` likewise warns rather than aborting, so a stuck pool slot never blocks the rest of teardown.

## Tests

- New `tests/fm-spawn-worktree-lease.test.sh` asserts the spawn meta records the leased path, that the worktree is leased with `treehouse get --lease` rather than sent into the pane, and that the pane is driven with `cd <worktree>`.
- `tests/fm-teardown.test.sh` gains graceful-degradation and non-fatal-return cases (non-treehouse worktree skipped, failed return warns but continues).
- The rewritten `test_spawn_conformance_old_vs_new` in `tests/fm-backend.test.sh` drops the byte-identical old-vs-new tmux-log comparison (no longer meaningful after the mechanism change) and instead asserts the new `--lease` mechanism, meta recording, and pane drive.
- The existing spawn mocks in `tests/fm-grok-harness.test.sh`, `tests/fm-secondmate-harness.test.sh`, `tests/fm-spawn-dispatch-profile.test.sh`, and `tests/fm-tangle-guard.test.sh` are updated to echo a fake leased path for the `--lease` flow.

## Test evidence

Verified in the crew worktree on 2026-07-04, rebased onto `origin/main` at `dc37d58`.

- `shellcheck bin/*.sh bin/backends/*.sh tests/*.sh` (the exact CI command): clean, exit 0.
- `bash -n` on every touched script: clean.
- The fix's own tests pass: `fm-spawn-worktree-lease.test.sh`, `fm-teardown.test.sh`, and the rewritten `fm-backend.test.sh` conformance case.
- Full behavior suite compared against a clean `origin/main` (`dc37d58`) baseline run in the same environment.
  The branch is strictly better than baseline: `fm-backend.test.sh` goes FAIL -> PASS, because its `test_spawn_conformance_old_vs_new` was already failing on `origin/main` (an upstream `GOTMPDIR` change had broken the byte-identical old-vs-new assertion) and this change rewrites that test to the new mechanism.
- The 9 remaining suite failures are pre-existing on `origin/main`, identical on both baseline and this branch, and all live in files this change does not touch: `fm-backend-cmux`, `fm-backend-orca`, `fm-backend-zellij` (backend-binary/home-hash env dependencies), `fm-gotmp`, `fm-fleet-sync`, `fm-session-start`, `fm-turnend-guard`, `fm-secondmate-safety`, and `fm-secondmate-lifecycle-e2e` (environment/tooling/seed dependencies).
  They are unrelated to this change.

## Note for the pusher

The commit preserves its original `Co-Authored-By: Claude Opus 4.8 (1M context)` trailer from the archived `fm/fmspawn-worktree-fix` commit `a5f477c`.
The firstmate repo house style forbids an agent co-author trailer, so consider `git commit --amend` to strip it before pushing.

### DIFF
--- a/bin/fm-spawn.sh
+++ b/bin/fm-spawn.sh
@@ -786,23 +786,40 @@ spawn_send_key() {  # <target> <key>
   esac
 }
 if [ "$KIND" != secondmate ] && [ "$BACKEND" != orca ]; then
-  spawn_send_text_line "$T" 'treehouse get'
-
-  # Wait for the treehouse subshell: the pane's cwd moves from the project to the worktree.
-  for _ in $(seq 1 60); do
-    p=$(spawn_current_path "$T" || true)
-    if [ -n "$p" ] && [ "$p" != "$PROJ_ABS" ]; then
-      WT="$p"
-      break
-    fi
-    sleep 1
-  done
-  if [ -z "$WT" ]; then
-    echo "error: treehouse get did not enter a worktree within 60s; inspect window $T" >&2
+  # Authoritatively LEASE the worktree instead of scraping the pane's cwd.
+  #
+  # The prior approach sent `treehouse get` into the pane and then polled
+  # `pane_current_path` until it differed from PROJ_ABS, taking that first
+  # differing path as the worktree. That heuristic is unreliable: the pane can
+  # momentarily report a directory that is neither PROJ_ABS nor the worktree -
+  # most commonly the session's default working directory, which is the
+  # firstmate home where fm-spawn runs - before `treehouse get` has entered its
+  # subshell. WT then latched onto that spurious path (observed: every crew's
+  # meta recorded worktree=<firstmate home> instead of the real leased
+  # worktree), which broke fm-teardown (`treehouse return` on a non-pool path),
+  # merge detection, and branch resolution.
+  #
+  # `treehouse get --lease` reserves a pool worktree durably and prints ONLY
+  # its absolute path to stdout (all banners go to stderr), so WT is exactly
+  # the leased worktree path. treehouse resolves the pool from the working
+  # directory, so run it from the project, mirroring fm-teardown's
+  # `treehouse return`. A leased worktree is protected from `treehouse prune`
+  # for the task's lifetime; teardown's `treehouse return --force "$WT"`
+  # releases it, exactly as it did for an interactively-acquired worktree.
+  set +e
+  WT=$(cd "$PROJ_ABS" && treehouse get --lease --lease-holder "fm-$ID" 2>/dev/null)
+  TH_STATUS=$?
+  set -e
+  if [ "$TH_STATUS" -ne 0 ] || [ -z "$WT" ]; then
+    echo "error: treehouse get --lease did not yield a worktree for $PROJ_ABS; inspect window $T" >&2
     exit 1
   fi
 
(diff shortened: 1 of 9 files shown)


kunchenguid:

PR #269, opened 2026-07-05
fix(bin): make heredoc assignments immune to bash-3.2 quote tracking

idrishassan-ai:
## Problem

macOS `/bin/bash` is 3.2.57, whose parser tracks quote state through a heredoc body nested inside `$( )` command substitution.
A lone apostrophe in such a body breaks parsing of the entire script (issue #166):

```
unexpected EOF while looking for matching `''
syntax error: unexpected end of file
```

Modern bash (4+) parses the construct fine, so Linux CI never sees the failure.
PR #173 made `bash -n` pass by rewording the apostrophe away, but left the fragile `VAR=$(cat <<EOF ...)` construct in place: any future apostrophe in those bodies re-breaks every ship brief on macOS with no CI tripwire.

## Fix

- `bin/fm-brief.sh`: assign the three ship-mode Definition-of-done heredocs with `read -r -d '' DOD <<EOF ... EOF || true` instead of `$(cat <<EOF ...)`. bash 3.2 parses this regardless of body quoting. No brief wording changed.
- `bin/fm-bootstrap.sh`: convert the `shim_body` and `cadence_body` heredocs the same way (`cadence_body` keeps its quoted no-expansion delimiter). Extends the same hardening so the whole `bin/` surface is free of the construct.
- `tests/fm-brief.test.sh`: `bash -n` now also runs explicitly under `/bin/bash` (the affected 3.2 interpreter on macOS), and a structural test refuses any non-comment `$(cat <<` across `bin/*.sh` and `bin/backends/*.sh`, so the trap cannot silently return on Linux CI where `bash -n` cannot see it.
- `docs`: the bash-3.2 heredoc convention is recorded alongside the updated test guard.

## Evidence (verified on /bin/bash 3.2.57, arm64-apple-darwin25)

- With an apostrophe injected into a DOD body, the old construct fails `-n` (exit 2, errors above); the new construct passes.
- All five generated briefs are byte-identical before vs after, built from the same `FM_HOME`: ship no-mistakes, ship direct-PR, ship local-only, `--scout`, `--secondmate`. The ship no-mistakes brief hashes to the same sha256 (`142a8e66…`) on both sides.
- Bootstrap-generated shim/cadence artifact contents are unchanged.
- Every script in `bin/`, `bin/backends/`, and `tests/` passes `/bin/bash -n` under 3.2; `tests/fm-brief.test.sh` passes under both `bash` (5.x) and `/bin/bash` (3.2).
- Validated by the no-mistakes pipeline: review, test, document, and lint all green with no findings (push required this fork PR because the branch author has no upstream push access).


kunchenguid:

PR #286, opened 2026-07-06
feat(spawn): name claude crewmate sessions fm-<id> for agent-view readability

TerminalGambit:
## What

Name fm-spawned claude crewmate sessions `fm-<id>` so they are readable in Claude Code's agent view (`claude agents`), instead of all sharing one auto-generated name.

- `bin/fm-spawn.sh`: the claude launch template gains `-n __TASKID__`, wired through a new `__TASKID__` placeholder substituted next to `__BRIEF__`/`__TURNEND__`. The value is `fm-<task-id>`, matching the tmux window name, so agent view and tmux agree on identity. Claude template only; codex/opencode/pi/grok are untouched.
- The template also prefixes `CLAUDE_JOB_DIR=`: verified empirically (claude 2.1.201) that when firstmate's tmux server was started from inside a `claude --bg` job, that job's `CLAUDE_JOB_DIR` leaks into every pane via the tmux server environment and the nested claude silently adopts the parent job's name, defeating `-n`. Bisected to that single variable; an empty value behaves as unset. This is why every session previously showed the same leaked name.
- `.agents/skills/harness-adapters/SKILL.md`: two lines in the claude section documenting the naming and the leak.

## Verification

- `-n` + `CLAUDE_JOB_DIR=` produces a session named `fm-<id>` in `claude agents --json` and the prompt box (verified live in a scratch tmux pane, cleaned up after).
- Busy-pane spinner rendering is identical with and without `-n` (sampled 10x each during live tool runs). The `esc to interrupt` hint not rendering on 2.1.201 is pre-existing version drift, present in the baseline too.
- `fm_tmux_composer_state` reads an idle `-n`-labeled pane as `empty` (the name renders on the border row, not the cursor row).
- zellij/cmux tab naming is unaffected: both match titles they set themselves at creation, never the live terminal title.

## Tests

Five assertion sites pinning the exact claude launch string updated to the new form. The old-vs-new spawn conformance test in `tests/fm-backend.test.sh` rewrites the baseline log's launch line to the new form before diffing - a targeted tolerance for this one intentional delta (a no-op once merged, since the merge-base then contains the new template) that keeps the diff sensitive to unintended drift. All 19 fm-spawn-touching test files pass locally; the full suite passed in the no-mistakes pipeline (review, test, document, lint all green).

https://claude.ai/code/session_01ESc8CDmc2vPdafGVMd8oft

### DIFF
--- a/.agents/skills/harness-adapters/SKILL.md
+++ b/.agents/skills/harness-adapters/SKILL.md
@@ -82,6 +82,9 @@ Natural language is acceptable if uncertain.
 | Interrupt | single Escape |
 | Skill invocation | `/<skill>` (e.g. `/no-mistakes`) |
 
+fm-spawn names every claude session `fm-<id>` (the window name) via `-n`, so crewmates are readable in Claude Code's agent view (`claude agents`).
+The launch template also clears `CLAUDE_JOB_DIR`, which otherwise leaks from a captain-side `claude --bg` job through the tmux server environment and silently overrides `-n` with the parent job's name (verified empirically on 2.1.201).
+
 First launch in a fresh worktree, or first ever on a machine, may show a trust or bypass-permissions confirmation.
 After every spawn, peek the pane within about 20 seconds.
 If such a dialog is showing, accept it with `bin/fm-send.sh <window> --key Enter`, or the choice the dialog requires, and verify the brief started processing.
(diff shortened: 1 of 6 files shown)


TerminalGambit:
Closing — withdrawing this contribution, not planning further changes to this repo. Thanks for the tool.

PR #324, opened 2026-07-07
fix(spawn): acquire crew worktrees via treehouse get --lease

nbdSteve:
## Problem

`treehouse` v2.0.0's interactive `treehouse get` opens a subshell inside the worktree ("Type 'exit' to return"). tmux's `#{pane_current_path}` does **not** report that subshell's cwd, so `fm-spawn.sh`'s worktree-discovery poll never observed the pane leaving the project directory and **timed out at 60s on every crewmate/scout spawn** - even though the worktree was created successfully. This blocks all crewmate dispatch on the current treehouse release.

## Fix

Acquire the worktree via `treehouse get --lease` as a subprocess (path printed to stdout, banners to stderr), then `cd` the pane into it. This is the same durable primitive `fm-home-seed.sh` already uses for secondmate homes, and `fm-bootstrap.sh` already requires `--lease` support - the crewmate/scout path was simply left on the old interactive approach. Spawn drops from a 60s timeout to **~2s**.

Also fixes a latent collision: when firstmate runs inside a tmux session whose name is a plain integer (e.g. `58`), `tmux new-window -t 58` is parsed as a **window index**, so a second task fails with `index 58 in use`. Addressing the session as `<session>:` forces session interpretation (correct for named sessions too).

## Tests

- Updated `fm-spawn`/backend tests to the lease contract (treehouse stub prints the worktree path on `get --lease`).
- Converted the spawn old-vs-new byte-identical conformance test to a forward contract test, since this is an intended behavior change.
- Added a regression test proving spawn succeeds purely from the lease, with no pane-cwd read.
- Full suite passes locally; shellcheck clean on changed scripts.

Verified end-to-end: real crewmate spawns now complete in ~2s and launch into the leased worktree.

### DIFF
--- a/bin/backends/tmux.sh
+++ b/bin/backends/tmux.sh
@@ -69,13 +69,22 @@ fm_backend_tmux_container_ensure() {
 # refusing an existing <window-name> in <session>. Mirrors fm-spawn.sh's
 # duplicate-check-then-new-window sequence, including the exact error text
 # (session:window, matching how fm-spawn.sh composed its own $T).
+#
+# The session is addressed as "<session>:" (trailing colon), never bare
+# "<session>". A bare target whose text is a plain integer - which happens when
+# firstmate runs inside a tmux session with a numeric name (e.g. "58") - is
+# parsed by tmux as a WINDOW INDEX in the current session, not a session name,
+# so `new-window -t 58` targets index 58 and a second task collides with
+# "index 58 in use". The trailing colon forces session interpretation and lets
+# tmux pick the next free index; it is equally correct for a named session
+# like "firstmate:".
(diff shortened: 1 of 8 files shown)


nbdSteve (commented):
Update spawn code to use --lease with new version of treehouse

nbdSteve:
Follow-up commit (668a703): the post-lease `cd` into the worktree is now verified and retried rather than fire-and-forget. A freshly created pane can drop the first keystroke (`cd <wt>` → `d <wt>`), leaving the agent in the primary checkout; the spawn now reads the pane cwd back and re-sends until it lands. Also makes fm-spawn-batch hermetic against a live config/crew-dispatch.json and adds a dropped-cd regression test.

kunchenguid:

PR #325, opened 2026-07-08
fix(backends): scope herdr workspace by firstmate home

zachlandes:
## Intent

Stop two independent firstmate installations on one machine from commingling their agent tabs in a single herdr workspace.

Every PRIMARY firstmate home resolved its herdr workspace to the bare constant 'firstmate'. The workspace lookup adopts the first workspace in the shared herdr session carrying that label, so two independent primaries on the same machine mint-or-adopt one shared space and both fleets' tabs land in it together. Neither installation can tell its own work from the other's in herdr's spaces sidebar, and each one's recovery claims the other's tasks as its own.

This is not hypothetical. It is the live configuration on the machine this was written for: two primary homes side by side, neither a secondmate home, with herdr selected as the backend.

The fix ports a solution this repository has already adopted twice. bin/fm-backend-hometag-lib.sh is the shared home-tag derivation that cmux (#246) and zellij (#252) both already use: a readable prefix plus a short hash of the resolved home path, so distinct installations never collide even though they share one backend-global namespace. fm_backend_herdr_workspace_label now delegates to it instead of returning a constant. herdr was the last backend still on the bare constant, so this closes the gap and leaves the three adapters unable to drift apart again.

It also repairs a latent inconsistency in that shared library, which this port exposed: the library read FM_HOME for the readable prefix but FM_ROOT for the hash. A cross-home invocation therefore produced a tag whose prefix named one home and whose hash named another, silently minting an orphan workspace that neither home's lookup would ever match. Both now derive from the same home. The two paths are identical in any well-formed home, so this is a no-op for every existing installation and no cmux or zellij tab title changes.

Deliberate decisions a reviewer should know about:

- Herdr is the NARROWEST of the three ports on purpose. Only the one workspace-CONTAINER label carries the home tag, because the per-home workspace is itself the namespace; per-task tab labels inside it stay the bare fm-<id>. cmux and zellij need the tag on their per-tab titles because they have no per-home container to hang it on. Harmonizing herdr's tab labels too would be redundant and would churn every existing task's recorded label.
- The tag is read fresh from the home on every call rather than cached at source time, so it stays stable across every respawn and recovery for the life of that home without threading state through a call chain.
- Relocating an installation changes its path and therefore its tag, so tabs created under the old tag stop matching. That limitation was already accepted for cmux and zellij, and it is no worse than the existing fact that a task's recorded absolute worktree path does not survive a move either. The one resulting herdr workspace migration is documented with a runnable command.
- A secondmate home already resolved to its own distinct label. This change is about the primary-versus-primary collision, which was the remaining hole.
- The change is verified end to end against a real herdr session, not only by unit assertions: on the base commit two independent homes both label their space 'firstmate', both tabs land in one space, and each home's recovery claims the other's task; on this branch each home gets its own hashed space and sees only its own task.

## What Changed

- `fm_backend_herdr_workspace_label` now delegates to the shared home-tag derivation in `bin/fm-backend-hometag-lib.sh` instead of returning the bare constant `firstmate`, so the primary home's workspace becomes `firstmate-<hash>` (and a secondmate home's `2ndmate-<id>-<hash>`). Two independent primary installations on one machine no longer mint-or-adopt a single shared herdr space where both fleets' tabs commingle and each home's recovery claims the other's tasks. Only the workspace-container label carries the tag; per-task tab labels stay the bare `fm-<id>`, since the per-home workspace is already the namespace.
- Fixed a latent inconsistency in the shared library the port exposed: it read `FM_HOME` for the readable prefix but `FM_ROOT` for the hash, so a cross-home invocation produced a tag naming one home in its prefix and another in its hash, silently orphaning the tab. Both halves now derive from `FM_HOME`. The two paths are identical in any well-formed home, so no existing cmux or zellij title changes. `bin/fm-spawn.sh` shadows `FM_ROOT` alongside `FM_HOME` on the `--secondmate` herdr path so the temporary environment names one home consistently.
- Added `tests/fm-backend-hometag-lib.test.sh` covering the shared derivation, extended the herdr unit and workspace-per-home e2e suites with two-independent-primaries cases, and hoisted the expected-tag helper into `tests/lib.sh` as the single reimplementation shared by the cmux, zellij, herdr, and hometag suites. Docs across `architecture`, `configuration`, and the herdr/cmux/zellij backend guides now describe the tag as `FM_HOME`-derived, with the one-time herdr workspace migration written out as a runnable command.

## Testing

- ⏭️ **Test** - skipped

## Pipeline

<details>

Step was skipped.

</details>

<details>

Step was skipped.

</details>

<details>
<summary>⏭️ **Review** - skipped</summary>

Step was skipped.

</details>

<details>

Step was skipped.

</details>

<details>
<summary>⏭️ **Document** - skipped</summary>

Step was skipped.

</details>

<details>

Step was skipped.

</details>

<details>

✅ No issues found.

</details>


kunchenguid:

PR #346, opened 2026-07-08
fix: detect degraded Codex watcher supervision

robinbraemer:
## Summary

This makes Firstmate honest on Codex surfaces where a tracked background watcher can exit without automatically waking the assistant turn.

The watcher still runs normally, but Codex sessions whose background-completion wake channel has not been empirically verified now surface a `DEGRADED` diagnostic instead of claiming full one-shot supervision reliability.

## Example failure mode

A concrete local case that motivated this patch:

1. Firstmate arms `bin/fm-watch-arm.sh` as a long-running tracked Codex exec session.
2. A secondmate appends a real update, for example `done: Project #16 no-status triage inspected 56 rows...`, to `state/minekube-secondmate.status`.
3. `fm-watch.sh` detects the status write correctly and exits with a wake reason such as:

   ```text
   signal: /Users/robin/Developer/robinbraemer/firstmate/state/minekube-secondmate.status
   ```

4. On this Codex surface, that background exec completion does not reliably start a new assistant turn. The signal is present, but the operator only sees it after manually polling the exec session.
5. Without this patch, Firstmate could treat the watcher as normally supervised and end the turn, leaving the watcher down after it fired.
6. With this patch, the arm/guard/turn-end paths say `DEGRADED` and warn that a fresh one-shot watcher is not proof of an automatic Codex wake path. The operator must keep the exec session polled, use a verified wake channel, or explicitly set `FM_CODEX_BACKGROUND_WAKE=verified` only after a local smoke proves it.

So the bug is not that `fm-watch.sh` misses the event. It catches the event. The unsafe assumption is that Codex will reliably wake Firstmate when that background watcher exits.

## Changes

- Add shared supervision helpers for detecting an unverified Codex background wake channel.
- Show a loud guard warning when in-flight work has only a fresh watcher on that degraded Codex path.
- Add the degraded suffix to `fm-watch-arm.sh` `started` / `healthy` output.
- Block primary turn end in Codex when in-flight work depends on an unverified watcher wake path, while preserving the AFK/sub-supervisor exemption.
- Add focused watcher and turn-end guard tests.
- Document the Codex caveat in `docs/turnend-guard.md`.

## Validation

- `bash tests/fm-watcher-lock.test.sh`
- Full `tests/*.test.sh` suite via no-mistakes test step
- Manual smoke in this fleet: primary watcher exited on a disposable status signal, was drained, and re-armed; Akua and Minekube secondmate homes were fast-forwarded to this fix and re-armed with the degraded Codex diagnostic.

No CI checks are configured on this repository; no-mistakes reported `checks-passed` after review, tests, lint fallback, push, PR creation, and CI monitoring.


robinbraemer:
Superseded by #355, which keeps this degraded-Codex diagnostic but adds the persistent present-mode watcher loop needed to fix the real auto-wake/auto-rearm supervision failure.

PR #348, opened 2026-07-08
fix(watcher): self-heal a wedged watcher on plain arm

lukeforshort:
## Intent

Make bin/fm-watch-arm.sh self-heal a WEDGED watcher on a plain 'arm', instead of returning FAILED and forcing a manual --restart.

Background: the primary turn-end guard was recurrently firing 'no live watcher holds this home lock' even though a watcher PROCESS was running. Root cause: a watcher can become wedged - process alive, still holding this home's singleton lock (state/.watch.lock) with a matching identity, but its liveness beacon (state/.last-watcher-beat) has gone stale past FM_GUARD_GRACE because it stopped beating. fm_watcher_healthy correctly reports it unhealthy, but plain arm mode only handled a healthy incumbent (no-op) or an absent/dead one (fork a fresh child); a wedge fell through to forking a child that could not steal the still-live lock, self-evicted per the singleton rule, and returned FAILED with the wedge left in place.

Fix and the deliberate decisions made:
1. Plain arm now detects that exact wedge and runs the SAME home-scoped stop that --restart does (kill the recorded pid after an identity match, or clear a stale-recorded lock), then forks and confirms a fresh watcher, all in one invocation.
2. The stop logic is deliberately factored into a shared stop_this_home_watcher() function used by both --restart and arm-self-heal, to respect the one-owner rule rather than duplicating the kill-after-identity-match logic. The restart path was refactored to call it; this is intentional dedup, not accidental behavior change - restart behavior is unchanged.
3. Crucial narrowing decision: the wedge predicate (wedged_watcher()) requires the beacon to EXIST and be stale past GRACE - NOT merely 'not healthy'. This deliberately excludes a startup-race peer (a matching watcher that just took the lock and has not beaten yet, so no beacon or a fresh one). Killing such a peer would be wrong; it must be left to beat. This distinction is why the existing peer-startup-race behavior is preserved.
4. A live lock held by something that is NOT this home's watcher (identity mismatch) is never signalled - stop_this_home_watcher's identity match guards that, so it still lands on the honest FAILED path. Home-scoping is strict: never a broad pkill, only the recorded pid.

Also in this change (fixing pre-existing test flakes surfaced along the way, at the captain's explicit request):
5. tests/fm-turnend-guard.test.sh: run_hook and the grok-adapter invocation did not pin FM_HOME/FM_STATE_OVERRIDE, so the copied hook inherited the ambient FM_HOME the real fleet exports into the session and inspected the REAL home's state - a hermeticity leak that flaked whenever the real fleet had in-flight work. Both now pin FM_HOME and FM_STATE_OVERRIDE at the temp dir. This is the correct hermetic fix, matching how sibling guard tests isolate.
6. .opencode/plugins/fm-primary-turnend-guard.js: fixed a real EPIPE crash - a guard that reads little and exits fast (the blocked case exits 2 without draining stdin) can close its stdin pipe before the plugin finishes writing, raising an unhandled 'error' on child.stdin (only child had a handler) and crashing the host. Added a child.stdin error handler that swallows it, matching the plugin's fail-open intent. The close handler still resolves the result.
7. The new wedge regression test uses the default confirm timeout (like test_arm_starts_and_self_heals) rather than a shortened one, so a 'started'-expecting assertion never races the confirm window under load.

NOTE on --skip=test: the local test step is skipped ONLY because this WSL dev host has jq in ~/.local/bin and node in /usr/bin (non-standard locations), which makes exactly 3 unrelated pre-existing tests fail (fm-dispatch-select, fm-gotmp, fm-session-start) purely from tool-location assumptions - the pipeline's own fix agent ran the full suite twice, confirmed these are host-environment artifacts with NO product defect, confirmed the branch's own tests (fm-watcher-lock, fm-turnend-guard, fm-tangle-guard) are fully green, and made no code changes. The real GitHub CI runs the full suite on a standard host where all tests pass, so validation is not lost. Verified locally: full-set shellcheck clean; fm-watcher-lock green across repeated runs; fm-turnend-guard green across repeated runs; fm-tangle-guard green; the new wedge regression test passed in the pipeline's own verification re-run.

## What Changed

- `bin/fm-watch-arm.sh` now detects a wedged watcher on a plain `arm` - process alive and still holding this home's singleton lock with a matching identity, but its liveness beacon has gone stale past grace - and self-heals it in one invocation by running the same home-scoped stop that `--restart` uses (kill the recorded pid after an identity match, or clear a stale-recorded lock) then forking and confirming a fresh watcher, instead of returning `FAILED` and forcing a manual `--restart`. The stop logic is factored into a shared `stop_this_home_watcher()` used by both paths, and the wedge predicate requires the beacon to exist and be stale (not merely "unhealthy"), so a startup-race peer with no beacon yet and a lock held by another home's watcher are both left alone on the honest `FAILED` path.
- `.opencode/plugins/fm-primary-turnend-guard.js` adds a `child.stdin` error handler that swallows EPIPE, fixing a host crash when the guard exits fast in the blocked case and closes its stdin pipe before the plugin finishes writing; the fail-open close handler still resolves the result.
- `tests/fm-turnend-guard.test.sh` pins `FM_HOME` and `FM_STATE_OVERRIDE` at the temp dir in `run_hook` and the grok-adapter invocation so the copied hook no longer inspects the real fleet's state, and `tests/fm-watcher-lock.test.sh` adds a wedge-self-heal regression test; `AGENTS.md` and docs are synced to describe the new plain-arm self-heal behavior.

## Risk Assessment

✅ Low: A well-bounded, thoroughly-commented, and well-tested self-heal path added to fm-watch-arm.sh via a cleanly-factored shared stop helper, plus three correct ancillary fixes (EPIPE handler, test hermeticity pinning, and confirm-timeout choice), with the wedge predicate deliberately narrowed to exclude startup-race peers and the stop scoped strictly to this home's identity-matched pid.

## Testing

- ⏭️ **Test** - skipped

## Pipeline

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>

Step was skipped.

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>


lukeforshort:
Closing - this change belongs on the maintainer's own fork (lukeforshort/firstmate#7), not upstream. Opened in error by the validation gate whose origin points here.

PR #366, opened 2026-07-08
fix(spawn): normalize crewmate PWD to absolute before launch

zachlandes:
## Problem

A crewmate's launch environment could carry a relative `PWD=.`.
bash and macOS `/bin/sh` preserve a relative `$PWD` verbatim (only zsh rewrites it), so `/bin/sh`'s `pwd`/`pwd -P` echoed `.` to the no-mistakes git-push `post-receive` hook, which then computed `--gate .`.
The daemon rejects `.` as an invalid gate path and silently creates no run object, so pipeline runs stayed blocked until the environment happened to carry an absolute `PWD` again.

## Fix

Before the harness launches, `bin/fm-spawn.sh` now sends `export PWD="$(/bin/pwd)"` into the crewmate's pane shell, alongside the existing `GOTMPDIR` export.
`/bin/pwd` calls `getcwd()` directly, ignoring the inherited `$PWD`, so the harness and every child process (git push included) inherit the real absolute cwd.
It rides the same backend send primitive as every other spawn-time command, so all backends (tmux, herdr, zellij, cmux, orca) are covered at one point.

Verified end to end: a freshly spawned crewmate's launched process, and a `/bin/sh` child of it, both report an absolute `PWD`, and the pane capture confirms the normalization lands before the launch command.

## Scope

This fixes only the spawn-side environment hygiene.
The gate hook itself deriving its path from `$(pwd)` is an upstream, shared-gate concern deliberately left out of scope.

## Tests

- `tests/fm-spawn-batch.test.sh`: a static guard (the normalization send is present and ordered ahead of the launch send) plus a portable behavioral proof (a poisoned `PWD=.` yields an absolute `PWD` for `/bin/sh` and bash children; the macOS-specific baseline is asserted only where the platform exhibits it, staying portable to Linux/dash CI).
- `tests/fm-backend-tmux-smoke.test.sh`: a real-pane end-to-end that reproduces the poison and proves the fix in an actual tmux crewmate pane.
- `tests/fm-backend.test.sh`: the old-vs-new command-sequence conformance filters the one intentional new send and asserts its presence.

`shellcheck bin/*.sh bin/backends/*.sh tests/*.sh` is clean.

## Pipeline auto-fixes folded into this branch

Validated through the no-mistakes pipeline, which applied two commits on top of the fix:

- `no-mistakes(document)`: updates the adapter comments and `docs/{herdr,orca}-backend.md` to list the new PWD-normalization send among the fixed spawn-time commands. Directly related to this change.
- `no-mistakes(test)`: **unrelated to this PR's purpose.** A locale-invariance fix to the composer-glyph stripping in `bin/backends/{cmux,herdr,orca}.sh` - under a C/POSIX locale `${stripped#?}` counts a single byte, leaving two stray bytes of the three-byte `❯` and misreading an idle composer as pending. The pipeline's test step surfaced this pre-existing latent bug while running the suite and fixed it to keep tests green. Flagged so it can be kept or split out at your discretion.


kunchenguid:

PR #370, opened 2026-07-08
fix(composer): make composer-state classification locale-invariant

zachlandes:
## Intent

Make firstmate's composer-state classification locale-invariant, so the away-mode supervisor cannot misread a pane's composer under the C/POSIX locale its daemon actually runs with.

Two separate faults, one root cause - a multibyte glyph treated as bytes under a C locale:

1. bin/fm-composer-lib.sh stripped a leading prompt glyph by CHARACTER COUNT (${content#?}). Under C/POSIX that removes a single BYTE, orphaning two bytes of the three-byte U+276F glyph. The backends' idle-placeholder regexes are anchored, so the stray byte defeats the anchor and an idle composer reads 'pending' instead of 'empty' - which makes the away-mode injector defer every escalation indefinitely (the overnight wedge).

2. bin/backends/herdr.sh matched the bare agent prompt with a grep bracket class ('^[❯›]'). Under C/POSIX that degenerates into a BYTE class, and every box-drawing glyph leads with the same byte, so the composer box's own bottom border row falsely matches as a bare prompt row. herdr keeps the bottom-most match, so the border row outranks the real composer, strips to empty, and the composer reads 'empty' while the human's unsubmitted text sits right above it - the false-EMPTY direction, where the injector would type over real input.

The fix replaces both byte-fragile forms with literal glyph sets matched by bash 'case', so no locale-sensitive grep or byte arithmetic remains anywhere in the glyph path. The glyph set becomes configurable - fleet-wide, and per-herdr layered on top of the fleet-wide default - and is threaded through the shared classifier so an operator-added harness glyph is honored for classification, not just for structural row detection (previously it was honored for detection only, so an overridden harness's idle composer wedged on 'pending' forever).

Deliberate decisions a reviewer should know about:

- Shell prompt glyphs are refused from the configurable agent set at three layers, and the hardcoded shell rule is evaluated before the configurable set, so no operator configuration can make a bare dead-shell prompt classify as an empty, injectable composer. That safety rule is the reason bin/fm-composer-lib.sh exists and it is preserved absolutely.
- A supplied set that sanitizes down to nothing falls back to the built-in default rather than to an empty set, so agent-glyph recognition can never be switched off fleet-wide.
- FM_BACKEND_HERDR_BARE_PROMPT_RE is retired in favour of a literal glyph-set knob. This is a real cost and is stated plainly: it changes the accepted FORMAT of a documented environment override. It is unavoidable, because any regex spelling must reach grep, and grep under a C locale is the hazard itself. Mitigating: the knob is set nowhere but its own default and one doc example; setting it now warns loudly on stderr instead of being silently ignored; and the code already contradicted its own documentation, since the knob was documented as controlling the recognized prompt glyphs but was honored only for row detection.
- Default behavior with no knob set is byte-identical to before, for every backend.
- Every locale regression test pins the locale in a child process, because an 'LC_ALL=C <shell function>' prefix does not re-run setlocale on bash 3.2 and would guard nothing.
- The test step also repaired two pre-existing fixture faults unrelated to this change (a missing composer-lib sibling in the gotmp fixture, and two Pi tests that did not tolerate node 23's experimental type-stripping warning), which were failing the suite locally before this branch.

## What Changed

- Replaced the two byte-fragile glyph forms in the composer path with literal glyph sets matched by bash `case`: `bin/fm-composer-lib.sh` no longer strips a leading prompt glyph by character count (`${content#?}`, which removes a single byte under C/POSIX and orphans two bytes of a three-byte glyph), and `bin/backends/herdr.sh` no longer matches bare prompts with a grep bracket class (`^[❯›]`, which degenerates into a byte class where every box-drawing glyph shares the leading byte, so a composer border row outranks the real composer). The only remaining regex — the per-harness idle placeholder — is now matched under a pinned `LC_ALL=C` so a hand-tested pattern behaves the same inside the C-locale supervise daemon.
- Made the agent glyph set configurable and threaded it through the shared classifier: `FM_COMPOSER_AGENT_GLYPHS` fleet-wide, with herdr's `FM_BACKEND_HERDR_BARE_PROMPT_GLYPHS` layering on that default and handed to `fm_composer_classify_content` as well as to structural row detection, so a configured glyph is honored for classification rather than detection only. Shell prompt glyphs (`>`, `$`, `%`, `#`) are refused at three layers, including the shared match chokepoint, and a set that sanitizes down to nothing falls back to the built-in `❯ ›` instead of emptying. `FM_BACKEND_HERDR_BARE_PROMPT_RE` is retired: it now warns loudly on stderr instead of being silently ignored. With no knob set, behavior is byte-identical to before for every backend.
- Added locale regression coverage in `tests/fm-composer-lib.test.sh` and `tests/fm-backend-herdr.test.sh`, each pinning the locale in a child process because an `LC_ALL=C <shell function>` prefix does not re-run `setlocale` on bash 3.2. Along the way, repaired two pre-existing fixture faults that were failing the suite locally (a missing composer-lib sibling in the gotmp fixture, and two Pi tests that did not tolerate node's experimental type-stripping warning), and updated the herdr, configuration, and harness-adapter docs to the new knobs.

## Testing

- ⏭️ **Test** - skipped

## Pipeline

<details>

Step was skipped.

</details>

<details>

Step was skipped.

</details>

<details>
<summary>⏭️ **Review** - skipped</summary>

Step was skipped.

</details>

<details>

Step was skipped.

</details>

<details>
<summary>⏭️ **Document** - skipped</summary>

Step was skipped.

</details>

<details>

Step was skipped.

</details>

<details>

✅ No issues found.

</details>


kunchenguid:
Thanks for the PR! This branch currently has a merge conflict with the base branch.

<sub>Noted for firstmate#370 at `4100fbcc`.</sub>

kunchenguid:

kunchenguid:
I am closing this because it has been waiting on a rebase or merge-conflict fix since 2026-07-15, and I have not seen a comment or push since then.

PR #380, opened 2026-07-09
fix(spawn): give claude crewmates a safe permission allowlist

mperaltamews:
## Summary
- \`claude --dangerously-skip-permissions\` is confirmed broken on this machine: crews still get a per-command approval prompt for every tool call, stalling autonomous work (data/learnings.md 2026-07-09 entry).
- The captain explicitly does not want full bypass permissions even if they worked (no guardrail against `rm -rf`, network egress, `sudo`, etc).
- Adds `bin/fm-crew-permissions.json`: a curated `permissions.allow` list of safe, worktree-contained `Bash(...)` patterns (git read/status/add/commit/checkout, file inspection, build/test/lint tooling for bun/npm/go/cargo/python, read-only `gh pr`/`gh issue`/`gh repo` lookups).
- `bin/fm-spawn.sh`'s claude launch path now merges this allowlist (via `jq -s '.[0] * .[1]'`) into the worktree's `.claude/settings.local.json` alongside the existing Stop-hook turn-end signal, instead of clobbering it. Falls back to the old Stop-hook-only write if `jq` or the permissions file is unavailable.
- Explicitly NOT allowlisted: `rm -rf`, `sudo`, `curl`/`wget`/network egress, `git push*`, `git merge*`, `gh pr merge*` — those still prompt for a human.

## Test plan
- [x] `bash -n bin/fm-spawn.sh` — syntax OK
- [x] Verified the jq merge produces both `permissions.allow` and `hooks.Stop` in one settings file
- [x] Live-tested with `claude -p` in a scratch dir using the generated settings.local.json:
  - `git status` → ran with zero `permission_denials`
  - `curl -s https://example.com` → blocked, `permission_denials` populated, confirming deny-by-default still holds for anything not on the list


mperaltamews:
Sorry, firstmate went above and beyond with this PR :D

PR #407, opened 2026-07-10
feat(herdr): opt-in pane-split layout for crewmate spawns

BenWilcox8:
## Summary

Adds an opt-in pane-split layout mode to the herdr runtime backend: crewmate/scout agents can spawn as split panes beside the spawning firstmate instead of each getting its own tab.

- **Toggle:** local, gitignored `config/herdr-layout` (`split` on; absent or `tabs` keeps today's tab-per-task behavior byte-for-byte), with an `FM_HERDR_LAYOUT` env override. Herdr-only; tmux/zellij/orca/cmux untouched.
- **Layout:** the first crewmate splits the spawner's pane left/right (spawner keeps the left half); each subsequent crewmate stacks in the right column with `--direction down`, newest at the bottom, `--ratio 0.5`.
- **Cap/overflow:** `FM_HERDR_SPLIT_MAX` (default 3) caps the right-column stack; beyond it spawns overflow to the existing tab path. A layout constraint never refuses a spawn.
- **No anchor:** when firstmate is not itself running in a herdr pane of the target session (`HERDR_ENV`/`HERDR_PANE_ID`), the spawn falls back to tab mode with a one-line stderr notice.
- **Identity/husk at pane granularity:** split panes are labeled `fm-<id>` via `pane rename`; duplicate/husk detection runs on pane labels with the same conservative dead/no-agent classification and create-before-close ordering as the tab husk path. Detection is cross-layout: each spawn path also scans the other scope for the same `fm-<id>` label, so a task respawned in the other layout (cap overflow, no-anchor fallback, or a config toggle) still finds its earlier incarnation's live duplicate or husk.
- **Scope:** crewmate/scout only; `--secondmate` spawns keep their own workspace in tab mode. All supervision already addresses `session:pane_id` targets from meta, so no supervision path assumed one-tab-per-agent.

## Verification

- Empirically verified against real herdr 0.7.2 / protocol 16 (NixOS Linux x86_64) in an isolated `fm-lab-*` session via `bin/fm-herdr-lab.sh`; evidence recorded in `docs/herdr-backend.md` "Pane-split layout mode" (split placement/geometry, labeling, cap overflow + override, both no-anchor paths, husk close-and-replace, live-duplicate refusal, middle-pane re-tile, spawner-tab survival).
- New fake-CLI unit tests in `tests/fm-backend-herdr.test.sh` (`make_herdr_splitfake` + `test_layout_*`/`test_split_*`) covering layout resolution, cap overflow at the cap and via override, no-anchor fallback, cross-scope duplicate/husk handling, and create-before-close ordering.
- `shellcheck bin/*.sh bin/backends/*.sh tests/*.sh` clean.
- Also fixes a NixOS portability bug in the pre-existing herdr and zellij missing-binary version-check tests (their restricted `PATH` lacked `bash`); the fix links the real `bash` into the private fakebin so the restricted `PATH` stays free of the binary under test on any host layout.

Validated through the no-mistakes pipeline (review with fix round, tests, docs, lint all green); pushed via fork because this account has no upstream push access.


BenWilcox8:
Didn't mean to submit this as a PR on the public repo. My bad!

PR #410, opened 2026-07-10
feat(bin): add read-only no-mistakes pipeline monitor

abhi-sama:
## Summary

Crewmate windows show only a spinner while `/no-mistakes` validates - the per-step review/test/lint work runs as headless subprocesses, invisible without hand-running `no-mistakes axi status` / `axi logs --step` in the worktree.

`bin/fm-pipeline-monitor.sh <task-id>` opens (or reuses) a dedicated `fm-monitor-<id>` tmux window in the crewmate's own session that live-tails the run: the 9-step status table plus a bounded tail of the active step's log, refreshed every few seconds.

- **Read-only observer**: only ever calls `no-mistakes axi status` / `axi logs --step` - never `respond` or `abort`, never touches `projects/`, never mutates a run. Test coverage includes a behavioral proof (a fake `no-mistakes` that logs any `respond`/`abort` call) that this holds even while displaying a parked/ask-user finding.
- **Bounded lifecycle**: closes itself on a terminal `outcome:`, on teardown (meta/worktree gone), or after `FM_PIPELINE_MONITOR_MAX_EMPTY_TICKS` consecutive checks with no run ever appearing (default ~30 min) - a monitor left open must never poll forever. Honors SIGTERM/SIGINT for a prompt exit.
- **tmux-first**: tmux is firstmate's verified reference backend; on any other recorded backend (herdr/zellij/orca/cmux) it prints a pointer to running `no-mistakes axi status`/`axi logs` directly in the worktree and exits 0 rather than blocking.
- **Opt-in auto-open**: standalone command by default; `config/pipeline-monitor=on` (local, gitignored) lets firstmate also open it right after triggering validation (AGENTS.md section 7). Absent/off keeps today's behavior unchanged.

Docs: `docs/pipeline-monitor.md` (mechanism, lifecycle, tuning), a `docs/scripts.md` row, a `docs/configuration.md` section, and two one-line `AGENTS.md` pointers (the `config/pipeline-monitor` knob and the opt-in hook in "Validate").

## Fixes folded in

While hardening the initial version, found and fixed:
- No-mistakes prints its own errors to stdout (not stderr); an error blob would otherwise render forever as a fake status block instead of the friendly "no active run" message.
- A window spawned via `tmux new-window` execs directly (no shell), so it only ever saw the tmux *server's* captured environment - `FM_HOME`/`FM_STATE_OVERRIDE`/`PATH` are now forced through a per-pane `env` prefix (not session-wide `set-environment`, which would leak into every other crewmate/secondmate window sharing that session).
- A failed `tmux new-window` call previously still printed a false "opened" message.
- (The one that actually surfaced while testing this fix) `run_loop` polled forever when no run ever appeared for the task - now bounded by `FM_PIPELINE_MONITOR_MAX_EMPTY_TICKS`.

## Test plan

- [x] `shellcheck bin/*.sh bin/backends/*.sh tests/*.sh` exits 0
- [x] `bash tests/fm-pipeline-monitor.test.sh` - hermetic `--tick` cases (worktree resolution, no-active-run/terminal-outcome/torn-down/missing-CLI/CLI-error handling, the respond/abort safety proof) plus a real-tmux section (private socket, mirrors `tests/fm-backend-tmux-smoke.test.sh`) covering open/reuse, auto-close on outcome, non-tmux no-op, env-forwarding into a pane whose server start-env has none of it, and self-termination after max-empty-ticks - all pass in seconds
- [x] Full `tests/*.test.sh` suite passes (44 files, `tests/fm-crew-state.test.sh` unaffected)


abhi-sama:
Closing - opened against the wrong repo by mistake (gh defaulted to upstream for this fork). Re-opening against origin (abhi-sama/firstmate) instead.

PR #414, opened 2026-07-10
fix(spawn): poll for a genuine isolated worktree, not just a differing path

edugomz:
## Summary
- The post-`treehouse get` worktree-detection loop in `bin/fm-spawn.sh` accepted the first pane path that merely differed from the project directory.
- A brand-new tmux pane can transiently report an unrelated pre-shell-init cwd before it ever reaches the project directory; that transient also differs from the project path and was wrongly latched onto, tripping the isolation guard against a path that was never the real worktree.
- Reproduced with plain `tmux` commands independent of any firstmate code (a fresh pane briefly reports the multiplexer's default directory for ~0.2-0.4s before settling), so this is a general timing race, not environment-specific.
- Fix: poll until the reported path itself resolves to a genuine, isolated git worktree (`spawn_worktree_ok`, a predicate shared with the existing fatal `validate_spawn_worktree` check), instead of accepting the first path that merely differs from the project directory.

## Test plan
- [x] New regression test `tests/fm-spawn-worktree-race.test.sh` using a fake tmux that scripts the exact transient-then-project-then-worktree sequence; fails against the pre-fix code, passes against the fix.
- [x] Full existing spawn/backend/tangle-guard test suites pass (`fm-spawn-batch`, `fm-spawn-dispatch-profile`, `fm-tangle-guard`, `fm-backend*`).
- [x] `shellcheck -x bin/fm-spawn.sh` clean.

### DIFF
--- a/bin/fm-spawn.sh
+++ b/bin/fm-spawn.sh
@@ -651,13 +651,18 @@ fi
 # (docs/herdr-backend.md "Known gaps").
 PROJ_ABS_REAL=$(cd "$PROJ_ABS" 2>/dev/null && pwd -P) || PROJ_ABS_REAL="$PROJ_ABS"
 
-real_path_or_raw() {  # <path>
-  local path=$1 real
-  if real=$(cd "$path" 2>/dev/null && pwd -P); then
-    printf '%s\n' "$real"
-  else
-    printf '%s\n' "$path"
-  fi
+# spawn_worktree_ok: true when <path> is itself the top of a git worktree
+# distinct from PROJ_ABS_REAL. The one place that decides "is this an isolated
+# worktree", shared by the poll below (probing non-fatally while waiting for
+# the backend to actually move the pane there) and validate_spawn_worktree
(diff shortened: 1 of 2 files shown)


pedromuller-del:
## Independent post-push review, round 2 (exact head `5fafc9c8`)

Re-ran `/code-review medium` against this PR's exact head. 8 findings, 0 blocking, no code change needed on this branch.

- One new scoped follow-up recorded, not fixed here (distinct dispatch/routing-precedence defect, out of this task's endpoint-classification scope): `bin/fm-helper.sh`'s dispatch-provider inheritance checks the wrong precondition and has no `--dispatch-box` path for routing-table homes. Full detail in `data/fm-helper-endpoint-repair/debrief.md`, alongside the existing TOCTOU deferral it joins.
- The remaining 6 findings are code-quality/DRY observations on the reconciled r2 code with no correctness impact; noted in the debrief for whoever next touches `bin/fm-helper.sh`, not tracked as required follow-ups.

Not merging.

PR #434, opened 2026-07-10
fix: make watcher pid identity clock-independent on procfs platforms

### LINKED ISSUE #433: Watcher liveness checks false-negative on WSL2: wall-clock-based pid identity drifts after host sleep
## Summary

`fm_pid_identity` (`bin/fm-wake-lib.sh`) fingerprints a process using `ps -o lstart=` - a wall-clock rendering of process start time. On WSL2, the clock drifts during Windows host sleep and is corrected afterward; since `ps` recomputes lstart from the current clock on every call, the rendered start time of the *same live process* changes. `fm_watcher_lock_matches_pid` then reads the mismatch as a recycled pid and declares a healthy watcher dead.

## Environment

- Linux 6.6.87.2-microsoft-standard-WSL2
- Observed 2026-07-10 on current main

## Evidence

Same continuously-running watcher process (pid and parentage unchanged throughout):

- Identity recorded at lock acquisition: `Fri Jul 10 11:54:03 2026 bash $FM_ROOT/bin/fm-watch.sh`
- Live readouts of the same pid minutes later: `11:54:29`, then `11:54:36` (drift growing)
- A freshly restarted watcher mismatched by 3s within seconds of acquisition (`12:00:40` recorded vs `12:00:43` live)

## Impact

- `fm-turnend-guard.sh` raises false "TURN WOULD END BLIND" alarms while the watcher is alive with a fresh beacon
- `fm-watch-arm.sh` reports `watcher: FAILED` against a healthy watcher
- `fm-watch-arm.sh --restart`'s mismatch path degrades to lock-clearing plus duplicate-watcher self-eviction instead of a clean kill
- Alarm fatigue: genuine supervision lapses become hard to distinguish from noise

## Proposed fix

Fingerprint with `/proc/<pid>/stat` field 22 (starttime in clock ticks since boot - constant for a process's lifetime, immune to wall-clock changes) plus the command line; keep the lstart form as a portable fallback where procfs is absent (macOS), and add a compat path so old-format lock records don't declare a live watcher dead when this lands mid-flight. PR to follow.

maafk:
Closes #433

## What

- `fm_pid_identity` now fingerprints a process with `/proc/<pid>/stat` field 22 (starttime in clock ticks since boot - constant for the life of a process, immune to wall-clock changes) plus the command line, on procfs platforms.
- Platforms without procfs (macOS) keep the previous `ps -o lstart=` form as a portable fallback, documented as drift-sensitive there.
- Compat path: a lock recorded in the old lstart format still matches its live watcher, so this change landing mid-flight does not declare a running watcher dead.
- Test infrastructure fixes picked up along the way: probe node `.ts` import support so the Pi extension suites run across node versions, and skip the tasks-axi handoff suites when the installed version is incompatible.

## Why

See #433: on WSL2 the wall clock slews after host sleep/resume and `ps` recomputes lstart from the current clock, so the SAME live watcher's rendered start time changes over time (observed live: lock recorded `11:54:03`, later reads `11:54:29` then `11:54:36`). `fm_watcher_lock_matches_pid` then treats a healthy watcher as a recycled pid, producing false "TURN WOULD END BLIND" alarms, `fm-watch-arm.sh` reporting FAILED beside a live watcher, and `--restart` degrading to lock-clearing instead of a clean kill.

## Validation

- New regression coverage in `tests/fm-watcher-lock.test.sh`: identity is byte-stable for a live process (no wall-clock component), distinct same-command processes yield distinct identities (recycled-pid guard intact), and an old-format lock still matches its live watcher.
- `shellcheck` clean on changed scripts.
- Full validation pipeline green on the branch head (review, tests, docs, lint).

## Pipeline

Note for the maintainer: the no-mistakes pipeline validated this branch fully green locally (review, tests, document, lint - and the complete test suite was re-run after each rebase onto advancing main). The final push and PR-open steps were performed manually from a fork, because a fork contributor lacks the upstream push access the pipeline's own push step requires; everything else followed the standard gate.


maafk:
Superseded by #492, which re-applies this fix onto current main (fix commit cherry-picked clean, re-validated, all checks green). This branch has since diverged from main (merge conflicts, failing no-mistakes check), so closing in favor of #492.

PR #448, opened 2026-07-11
feat(crowsnest): thread/reply context + fix post-tool credential loading

Marl0nL:
## Crowsnest iteration 2: thread/reply context + post-tool credential fix

Improves the Crowsnest (firstmate's opt-in two-way Google Chat ↔ live-firstmate relay). Inert unless enabled, so no behaviour change for anyone who hasn't opted in.

### 1. Thread / reply context (headline)
When the captain replies inside a thread, the relay previously forwarded only the message text — the `local-agents-chat` subprocess agent forwards only `space`/`thread`/`sender`/`mode`/`text` (no replied-to message, no raw event, no history), so the relay alone can't see what the captain is replying to.

- New `bin/fm-crowsnest-context.{sh,py}` read recent thread messages back via an authenticated `spaces.messages.list` (reusing the backend `ChatClient` credentials) and merge `thread_context`, `reply_to`, and `sender_display_name` into `state/chat-inbox/<id>.json`.
- The relay spawns enrichment **detached** *after* the durable base entry + wake are written, so the instant-ack contract is preserved and a slow/failing read never loses a message. Everything is best-effort and degrades to today's exact text-only behaviour on any failure (no thread, no interpreter, no backend, no creds, no scope, network error).
- The merge re-checks the inbox entry still exists immediately before replacing it, so a message the live session already answered and deleted is never resurrected as a duplicate.
- Gated by `CROWSNEST_THREAD_CONTEXT` (on by default). `FMC_CONTEXT_SYNC` (inline) and `FMC_CONTEXT_FIXTURE` (canned `messages.list`) make the whole path testable offline with no GCP.
- `reply_to` is intentionally best-effort ("the last prior message in the thread", not a guaranteed quote) because the true quoted-message metadata isn't forwarded by the backend.
- `fmc-respond` and `docs/crowsnest.md` updated to compose using the context, degrading gracefully when absent.

### 2. Post-tool credential fix (known 403 bug)
The post tool never auto-loaded the backend's service-account credentials: it passed `--config=None`, and `load_config(None)` reads **no** file, leaving `credentials_path` empty — so posts fell back to user ADC and 403'd ("insufficient authentication scopes") unless the operator manually exported `GOOGLE_APPLICATION_CREDENTIALS`.

- New shared helper `bin/fm_crowsnest_chat.py` resolves the backend's **default** config path when none is given (mirroring `local-agents run`, whose `--config` argparse default is `str(default_config_path())`), honouring `LOCAL_AGENTS_CREDENTIALS_PATH` and `GOOGLE_APPLICATION_CREDENTIALS`. Verified: **OLD → `adc` (403), NEW → `service_account`**.
- One owner for credential resolution, shared by both the post and context tools.

### 3. Easy wins
- The captain's display name is harvested for free from the context read.
- `fm-crowsnest.sh status` gains a `posting id` line (reuses `ChatClient.posting_identity`) so the exact ADC-vs-service-account 403 is diagnosable at a glance.

### Docs & tests
- `docs/crowsnest.md`, `docs/scripts.md`, `docs/examples/crowsnest.env`, and the `fmc-respond` skill updated. `AGENTS.md` deliberately unchanged (conditional detail belongs in the doc/skill per the coding guidelines).
- 10 new hermetic tests (reader dedup/order/bound, merge, no-thread/empty/disabled/never-resurrect no-ops, relay enrichment, kill-switch, credential-fallback resolver). Full `fm-crowsnest` suite: 32 checks. `shellcheck` clean.

### Notes for the reviewer
- Validated through the no-mistakes pipeline: **rebase** skipped (branch is exactly `origin/main` + this one commit; the gate's stale baseline flagged unrelated fork history that's already on the PR base), **review** passed at low risk after two auto-fixes folded in here (the resurrection guard and the diagnostic never-raises contract). The **test** step could not run in the gate host because it lacks `tmux` (the `.no-mistakes.yaml` test command hard-requires it); all diff-touched suites pass locally and the gate's fix agent independently confirmed every remaining failure is a pre-existing environmental gap (tmux / tasks-axi / Node `.ts` / `CLAUDECODE` leak), none touching this diff.

### Proposed for later (not built here)
- Have the backend forward `sender_display_name` / the raw event (incl. true `quotedMessageMetadata`) / prior `history` to the subprocess, so context needs no read-back call and `reply_to` can be the actual quoted message. Requires a change to the external `local-agents-chat` package.
- User-auth context read for private history where the bot's `chat.bot` scope can't list messages.


Marl0nL:
Opened against upstream by mistake (fork-parent default); reopening within the Marl0nL/firstmate fork.

PR #454, opened 2026-07-11
feat: add Oh My Pi (omp) harness adapter

LoneExile:
## What

Adds **omp** (Oh My Pi, a Pi fork) as a first-class firstmate harness —
primary + crew — alongside claude / codex / opencode / pi / grok.

## How it works

- **Detection** (`fm-harness.sh`, `fm-lock.sh`): omp sets both `OMPCODE=1`
  and `CLAUDECODE=1`, so `detect_own` checks `OMPCODE` **first** (unique to
  omp; Claude Code never sets it). The lock regex matches omp as a path/word
  component (`(^|/| )omp( |$)`) so it can't collide with "compose"/"components".
- **Launch** (`fm-spawn.sh`, `fm-tmux-lib.sh`): `omp --auto-approve … -e <ext>`
  template with model/effort flags and the state-resident extension paths.
- **Supervision** (`.omp/extensions/`): two project extensions ported from the
  Pi supervisors and kept in lockstep — a turn-end guard and a background-wake
  watcher. OMP has no `agent_settled` event, so the guard listens on `turn_end`
  (documented tradeoff: it re-nags on each blind turn until armed, vs Pi's
  once-per-run; safe — the `guardFollowupActive` latch prevents self-nagging).
- **Session start** (`fm-session-start.sh`): `OMP_WATCH_EXTENSION` loaded-check,
  twin of the Pi block.
- **Repair line** (`fm-supervision-instructions.sh`): omp-specific
  `fm_watch_arm_omp` guidance.
- **Teardown** (`fm-teardown.sh`): cleans the per-task `state/<id>.omp-ext.ts`.

## Tests / verification

- `tests/fm-omp-primary-types.test.sh` — strict `tsc` no-emit typecheck of the
  extensions (skips cleanly without the omp package / `@types/node`).
- `tests/fm-omp-primary-live-e2e.test.sh` — gated `FM_OMP_LIVE_E2E=1`, mirrors
  the Pi live e2e; drives a real omp session through guard→wake→re-arm→clean
  `/quit` (`OMP_EXIT=0`).
- Extends #432's ambient-marker isolation to `OMPCODE` in `fm-session-start`
  and the three harness-pinned suites (`fm-secondmate-harness`,
  `fm-turnend-guard`, `fm-x-mode`) so the suite stays deterministic when run
  from inside an omp session.
- Live-verified: a real omp primary ran the full loop end-to-end and dispatched
  two crewmates that shipped green PRs under omp supervision.

## Also

- `.no-mistakes.yaml`: run the gate's test step with ambient harness/backend
  markers stripped (`OMPCODE`/`CLAUDECODE`/`HERDR_ENV`/`TMUX`) so gating from
  inside an omp or herdr session matches CI's clean env.


LoneExile:
Superseding this with a no-mistakes-gated PR per CONTRIBUTING.md — this one was opened via plain `gh pr create` and lacks the required no-mistakes signature, so `Require no-mistakes` would fail it. Re-raising through the gate.

PR #463, opened 2026-07-11
Fix Codex Stop hook session corruption

henrikkvamme:
## Summary

- route Codex Stop events through a Codex-specific fail-open adapter
- preserve Desktop-resumable session history instead of persisting a malformed synthetic hook continuation
- keep the supervision alarm visible and durable without spawning a competing Codex process
- document the temporary workaround for openai/codex#20783

## Why

A blocking Codex Stop hook persists its synthetic `<hook_prompt>` item with a bare UUID. A later Desktop-style follow-up sends that UUID back to the Responses API, which rejects it with `invalid_id_prefix` because message IDs must begin with `msg_`.

## Validation

- `FM_CODEX_LIVE_E2E=1 node tests/fm-codex-turnend-live-e2e.mjs`

The live app-server regression completed two turns, observed the system warning, verified that no hook prompt was persisted, verified the durable wake, and required a nonempty agent message from both turns.

## Upstream

- openai/codex#20783

### DIFF
--- a/.agents/skills/harness-adapters/SKILL.md
+++ b/.agents/skills/harness-adapters/SKILL.md
@@ -50,7 +50,8 @@ Use that value for interrupt, exit, resume, and skill-invocation facts.
 ## Primary turn-end guard
 
 Every verified primary harness has an empirically validated hook path for the "no turn ends blind" guard.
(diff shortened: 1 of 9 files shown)


kunchenguid:

PR #476, opened 2026-07-11
fix(tmux): defer sends during codex boot screens

twaldin:
## Summary

Defers tmux `send-keys` to a crewmate pane while it is still on an interactive
boot/splash screen (notably codex's startup screens), so keystrokes are not
swallowed or misinterpreted before the shell is ready to receive input.

- `bin/fm-tmux-lib.sh`: adds a boot-state guard that detects when the target
  pane is still rendering a boot/splash screen and polls for shell readiness
  before delivering keys, rather than sending blindly.
- `bin/fm-send.sh`: when the guard determines the pane is not yet ready, the
  send is deferred and the call reports a `send-deferred` verdict (with a clear
  inject log line) instead of silently dropping input.
- `tests/fm-tmux-boot-guard.test.sh`: new colocated test covering the
  boot-state detection and the deferred-send path.

Also includes the pipeline's review/test/doc fix rounds folded in on top of the
core change: hardened boot-guard poll values, a clarified send-deferred inject
log line, tmux smoke test now waits for shell readiness, handoff tests skip on
old tasks-axi, a relaxed bootstrap elapsed assertion, and documentation of the
tmux codex boot guard and send-deferred verdict.

## Validation

- shellcheck clean on the touched `bin/` scripts.
- Full repo test suite green, including the new `tests/fm-tmux-boot-guard.test.sh`.


kunchenguid:

PR #477, opened 2026-07-11
fix(bootstrap): pin + checksum-verify no-mistakes install

prodempsey:
Replaces the unversioned, unverified supply-chain install path (`curl .../no-mistakes/main/docs/install.sh | sh`) with a pinned, checksum-verified install of a specific release.

`bin/fm-install-no-mistakes.sh` downloads the tagged v1.31.2 release asset for the host os/arch from GitHub release assets (not raw `main`) and verifies its SHA256 against a checksum pinned in the script before installing. It fails closed on a mismatch or missing asset and never falls back to the old `curl|sh`. Install location and symlink target, the version gate, and the MISSING/consent flow are unchanged.

Pinned SHA256s (linux/darwin x amd64/arm64) were verified read-only three ways that agreed: the release's checksums.txt, the GitHub API asset digest, and a locally recomputed sha256 of the linux-amd64 asset.

Folded and live-verified in a downstream runtime for several days before this upstream submission.


kunchenguid:

PR #478, opened 2026-07-11
fix(composer): treat claude's non-breaking-space empty composer as empty

costis-t:
## Problem

`bin/fm-send.sh` consistently exited non-zero with a false `text not submitted (Enter swallowed; text left in composer)` error when steering a **claude** crewmate tmux pane, even though the message had actually submitted and the crewmate was already processing it. The false failure invites a dangerous double-send.

## Root cause

Reproduced empirically against a live claude v2.1.x pane: claude renders its **empty** composer prompt as the agent glyph `❯` followed by a **non-breaking space (U+00A0)** and then a regular space.

bash's `[:space:]` character class does not match U+00A0, so the caller's whitespace trim leaves a lone NBSP behind. After the leading prompt glyph is stripped, that stray NBSP survives as apparent content, and `fm_composer_classify_content` classifies an empty composer as `pending` (unsubmitted text). `fm-send` reads that as a swallowed Enter and exits non-zero.

## Fix

Fold every NBSP to a regular space and re-trim inside `fm_composer_classify_content` in `bin/fm-composer-lib.sh` — the single fleet-wide owner of the `empty|pending|unknown` verdict — so NBSP-only spacing classifies as `empty`, while real typed text that merely contains an NBSP still stays `pending` (the fold collapses only pure whitespace). Placing the fix at the one owner rather than per-adapter is intentional per the repo's one-owner rule, so every backend adapter benefits.

Genuinely-unsubmitted detection is deliberately preserved: the grok/codex slash- and dollar-popup placeholder cases, dead-shell `unknown`, and real-text `pending` all continue to work. Verification was not loosened.

## Evidence

- Drove the real `bin/fm-send.sh` against a live claude tmux pane: the idle composer now reads `empty` (was `pending`), and `fm-send` exits `0` where it previously exited `1` and fired extra retry Enters. Real unsubmitted text on the same live pane still reads `pending`.

## Tests

Regression tests added by extending existing scripts (no new runner):

- `tests/fm-composer-lib.test.sh` — classifier-level NBSP cases: an NBSP empty composer reads `empty`; real text around an NBSP stays `pending`.
- `tests/fm-composer-ghost.test.sh` — end-to-end through `fm_pane_input_pending` / `fm_tmux_composer_state` with claude's real cursor-row bytes (`❯` + U+00A0 + space).

`shellcheck bin/*.sh bin/backends/*.sh tests/*.sh` passes.

### DIFF
--- a/.agents/skills/harness-adapters/SKILL.md
+++ b/.agents/skills/harness-adapters/SKILL.md
@@ -123,6 +123,9 @@ Its broader dark-TRUECOLOR placeholder handling and dark-theme tradeoff are docu
 That styled capture is internal to the boolean detector only.
 `fm-peek` and every other human or LLM-facing capture path stays plain `tmux capture-pane` with no escape codes.
 
+Claude (v2.1.x, verified against a live pane) also renders its EMPTY composer prompt as the agent glyph `❯` followed by a NON-BREAKING space (U+00A0) and then a regular space, which bash's `[:space:]` trim never matches.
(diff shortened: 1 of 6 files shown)


costis-t:
The first mate was loyal to his admiral, Kun :) hahah

kunchenguid:

PR #487, opened 2026-07-12
feat: token-frugality rules in crewmate contracts and supervision discipline

rishabhxverma:
Encodes token-saving operating rules (RTK-video-inspired) into the shared contract:

- **bin/fm-brief.sh**: new Rule 7 in both ship and scout crewmate contracts - no start-to-finish reads of long files (targeted grep/sed/offset only), logs/bulk data staged into SQLite or script-parsed instead of streamed as raw text, CLI/API output projected to needed fields (--query/jq) and capped, generated files ignored, scope kept to the task's files.
- **AGENTS.md §8**: firstmate's own token discipline extended to mirror the same rules, plus keep briefs/memory/backlog information-dense (rewrite-and-prune, not append-forever).
- Also carries a previously-uncommitted AGENTS.md policy edit: the mandatory-"captain"-address block replaced with plain, name-based address and no nautical seasoning (user-directed change).

Verified: bash -n clean; scaffold smoke test shows the rule renders in generated briefs.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,15 +1,13 @@
 # Firstmate
 
 You are the first mate.
-The user is the captain.
+The user is Rishabh - the person you report to, whose explicit word is required for merges and other gated decisions.
+(Elsewhere this file uses "the captain" as a role term for that decision authority; treat it as internal shorthand for Rishabh, never as a form of address.)
 This file is your entire job description.
 
-Address the user as "captain" at least once in every response.
-This is mandatory respectful address, not performance: it applies even when delivering bad news or relaying serious findings, such as "Captain, the build broke - ...".
-Do not force it into every sentence, but never send a response with zero direct address.
-Use light nautical seasoning only when it fits: the occasional "aye", "on deck", or "shipshape" may land naturally.
-Keep that seasoning optional and never let it obscure technical content; never use it in commits, briefs, PRs, or anything crewmates or other tools read; drop the playful flavor entirely when delivering bad news or relaying serious findings.
-Captain-facing messages are plain outcomes about the captain's work; keep firstmate's internal machinery out of the substance of what the captain reads, even when the playful flavor drops away.
+Do NOT address the user as "captain". When a direct address is natural, use his name, Rishabh; otherwise use none - there is no mandatory address, and a response with no direct address is fine.
+Do not use nautical seasoning ("aye", "on deck", "shipshape", etc.).
+User-facing messages are plain outcomes about the user's work; keep firstmate's internal machinery out of the substance of what the user reads.
 
 ## 1. Identity and prime directives
 
@@ -524,6 +522,7 @@ This includes your own no-mistakes pipeline, long builds, and any other multi-mi
 Background that work so watcher wakes can interleave with it and the supervision loop stays responsive.
 
 Token discipline: status files before panes; default peeks to 40 lines; never stream a pane repeatedly through yourself; batch what you tell the captain.
+Firstmate's own token frugality mirrors the crewmate brief contract: no start-to-finish reads of long files (targeted grep/sed/offset reads only), logs and bulk data staged into SQLite or script-parsed rather than streamed as text, CLI/API output projected to needed fields and capped, generated files ignored, scope kept to the files a task needs, and briefs/memory/backlog kept information-dense (rewrite-and-prune, not append-forever).
 The context-% shown in a peek is not actionable as crew health; ignore it and intervene only on real signals (`signal`, `stale`, `needs-decision`, `blocked`), looping or confusion in the pane, or a question the brief already answers.
 Silence is the correct state while a healthy background watcher is waiting.
 
--- a/bin/fm-brief.sh
+++ b/bin/fm-brief.sh
@@ -139,6 +139,10 @@ The report is the only thing that survives, so anything worth keeping must be in
 5. If you hit the same obstacle twice, append \`blocked: {why}\` and stop; firstmate will help.
 6. If a decision belongs to a human (product choices, destructive actions),
    append \`needs-decision: {summary of options}\` and stop. Firstmate will reply with the decision.
+7. Token frugality: never read long files start-to-finish - pull only the needed lines
+   (grep/sed/offset reads); for logs or bulk data, stage into SQLite or script-parse and
+   query targeted slices instead of streaming raw text; project only needed fields from
+   CLI/API output (--query/jq) and cap output with head/tail; ignore generated files.
 
 # Definition of done
 Write your findings to \`$DATA/$ID/report.md\`.
@@ -222,6 +226,11 @@ $RULE1
 5. If you hit the same obstacle twice, append \`blocked: {why}\` and stop; firstmate will help.
 6. If a decision belongs to a human (product choices, destructive actions, ask-user findings),
    append \`needs-decision: {summary of options}\` and stop. Firstmate will reply with the decision.
+7. Token frugality: never read long files start-to-finish - pull only the needed lines
+   (grep/sed/offset reads); for logs or bulk data, stage into SQLite or script-parse and
+   query targeted slices instead of streaming raw text; project only needed fields from
+   CLI/API output (--query/jq) and cap output with head/tail; ignore generated files
+   (lockfiles, dist/, node_modules/); stay within the files the task needs.
 
 # Project memory
 If \`AGENTS.md\` or \`CLAUDE.md\` already exists, or if this task produced durable project-intrinsic knowledge, run \`$FM_ROOT/bin/fm-ensure-agents-md.sh .\` in the worktree.


rishabhxverma:
Withdrawn - was intended as a local config change, not an upstream contribution.

AGENTS.md
```diff
@@ -1,15 +1,13 @@
 # Firstmate
 
 You are the first mate.
-The user is the captain.
+The user is Rishabh - the person you report to, whose explicit word is required for merges and other gated decisions.
```
dangquan1402, line 4: **Personal identity hardcoded into shared tracked material.**

`AGENTS.md` is the shared, committed contract for every firstmate install (section 1: "This repo is a shared template, not the captain's personal project"), and section 3 already routes operator-specific preferences to the gitignored `data/captain.md`. Naming one individual here (and assuming "his") puts a single user's identity into the surface that `/updatefirstmate` fast-forwards into every other home: any other operator who updates then runs with "The user is Rishabh" and is told to address them by that name. The address-style preference belongs in `data/captain.md` (or `data/captain-shared.md`), not in the tracked template.

AGENTS.md
```diff
@@ -1,15 +1,13 @@
...
 This file is your entire job description.
 
-Address the user as "captain" at least once in every response.
-This is mandatory respectful address, not performance: it applies even when delivering bad news or relaying serious findings, such as "Captain, the build broke - ...".
-Do not force it into every sentence, but never send a response with zero direct address.
-Use light nautical seasoning only when it fits: the occasional "aye", "on deck", or "shipshape" may land naturally.
-Keep that seasoning optional and never let it obscure technical content; never use it in commits, briefs, PRs, or anything crewmates or other tools read; drop the playful flavor entirely when delivering bad news or relaying serious findings.
-Captain-facing messages are plain outcomes about the captain's work; keep firstmate's internal machinery out of the substance of what the captain reads, even when the playful flavor drops away.
+Do NOT address the user as "captain". When a direct address is natural, use his name, Rishabh; otherwise use none - there is no mandatory address, and a response with no direct address is fine.
```
dangquan1402, line 8: **Persona removal is not propagated to the other tracked surfaces that still mandate it.**

Deleting the address/nautical rule here leaves contradicting instructions live elsewhere in the same tracked tree:
- `.agents/skills/fmx-respond/SKILL.md` still says to "address them as \"captain\" when it fits" and that "Light nautical seasoning is welcome" for public X replies. Concrete failure: on an `x-mention` wake the agent loads that skill and posts a public reply addressing the user as "captain" with nautical flavor, directly violating the new blanket ban.
- `README.md` still advertises the output shape `PR ready for review, captain: https://...`.

Whichever surface was loaded last wins, so behavior becomes nondeterministic. Either update those surfaces in the same change or keep the preference out of the tracked contract entirely. (Note this branch is also currently conflicting with `main`, so the rebase has to carry these along.)

AGENTS.md
```diff
@@ -524,6 +522,7 @@ This includes your own no-mistakes pipeline, long builds, and any other multi-mi
 Background that work so watcher wakes can interleave with it and the supervision loop stays responsive.
 
 Token discipline: status files before panes; default peeks to 40 lines; never stream a pane repeatedly through yourself; batch what you tell the captain.
+Firstmate's own token frugality mirrors the crewmate brief contract: no start-to-finish reads of long files (targeted grep/sed/offset reads only), logs and bulk data staged into SQLite or script-parsed rather than streamed as text, CLI/API output projected to needed fields and capped, generated files ignored, scope kept to the files a task needs, and briefs/memory/backlog kept information-dense (rewrite-and-prune, not append-forever).
```
dangquan1402, line 525: **"generated files ignored" contradicts the paragraph it was added to.**

The line immediately above is "status files before panes" - and firstmate's entire supervision input (`state/<id>.status`, `state/<id>.meta`, `state/.wake-queue`, the fleet/bearings snapshots) is machine-generated. An agent applying "generated files ignored" literally would skip exactly the durable records sections 5 and 8 require it to read, and fall back to pane peeks, which is the opposite of the intent. Same problem for "no start-to-finish reads of long files": `AGENTS.md` itself is declared "your entire job description" and is 700+ lines, so a literal reading tells firstmate to grep-sample its own contract. Add an explicit carve-out for state/status records and contract files.

bin/fm-brief.sh
```diff
@@ -222,6 +226,11 @@ $RULE1
 5. If you hit the same obstacle twice, append \`blocked: {why}\` and stop; firstmate will help.
 6. If a decision belongs to a human (product choices, destructive actions, ask-user findings),
    append \`needs-decision: {summary of options}\` and stop. Firstmate will reply with the decision.
+7. Token frugality: never read long files start-to-finish - pull only the needed lines
+   (grep/sed/offset reads); for logs or bulk data, stage into SQLite or script-parse and
+   query targeted slices instead of streaming raw text; project only needed fields from
+   CLI/API output (--query/jq) and cap output with head/tail; ignore generated files
```
dangquan1402, line 232: **"ignore generated files (lockfiles, dist/, node_modules/)" is unqualified and reaches beyond reading.**

The intent is clearly "don't burn tokens reading them", but this is a ship brief - the crewmate produces a commit. Concrete failure: the task adds or bumps a dependency, the crewmate follows rule 7 and leaves the lockfile alone, and the PR either fails CI on `npm ci` / `pnpm install --frozen-lockfile` / `poetry check --lock`, or lands a `package.json` change without its lockfile. Scope the rule to reading, e.g. "do not read generated files (lockfiles, dist/, node_modules/) for context; still update them when your change requires it".

bin/fm-brief.sh
```diff
@@ -139,6 +139,10 @@ The report is the only thing that survives, so anything worth keeping must be in
 5. If you hit the same obstacle twice, append \`blocked: {why}\` and stop; firstmate will help.
 6. If a decision belongs to a human (product choices, destructive actions),
    append \`needs-decision: {summary of options}\` and stop. Firstmate will reply with the decision.
+7. Token frugality: never read long files start-to-finish - pull only the needed lines
```
dangquan1402, line 142: **No carve-out for contract and instruction files.**

"never read long files start-to-finish - pull only the needed lines" is stated absolutely. Concrete failure: a task in the firstmate repo (or any repo whose `AGENTS.md`/`CLAUDE.md` is long) has the crewmate grep-sample its instructions instead of reading them, and it misses a safety boundary it was never shown - worktree isolation, never-push-to-default, the status protocol. It also undercuts this brief's own project-memory rule, which assumes the crewmate understands the existing `AGENTS.md` before editing it. Exempt the brief, the project's agent-memory files, and gate/review output from rule 7.

bin/fm-brief.sh
```diff
@@ -139,6 +139,10 @@ The report is the only thing that survives, so anything worth keeping must be in
...
+   (grep/sed/offset reads); for logs or bulk data, stage into SQLite or script-parse and
```
dangquan1402, line 143: **"stage into SQLite" names a tool the bootstrap never verifies.**

`bin/fm-bootstrap.sh` checks for `tmux`, `node`, `git`, `gh`, `curl`, and `jq`; `sqlite3` appears nowhere in it. On any machine without it (common on minimal Linux images/containers, which is where crewmates often run), a crewmate that follows this rule hits `sqlite3: command not found` mid-task and spends turns recovering - the opposite of the saving the rule is for. Either soften to "script-parse (or stage into SQLite if available)" or add `sqlite3` to the bootstrap tool check.

bin/fm-brief.sh
```diff
@@ -139,6 +139,10 @@ The report is the only thing that survives, so anything worth keeping must be in
...
+   query targeted slices instead of streaming raw text; project only needed fields from
+   CLI/API output (--query/jq) and cap output with head/tail; ignore generated files.
```
dangquan1402, line 145: **`--query` is the wrong flag for the CLI this brief mandates.**

Rule 3 requires `gh-axi` for GitHub work, and the `gh` projection flag is `--jq`/`-q`; `--query` is Azure/AWS CLI syntax. A crewmate following this literally runs `gh-axi pr list ... --query ...`, gets an unknown-flag error, and burns a retry. Use `--jq`/`-q` (or write it generically as "the CLI's own projection flag, or pipe through jq").

wjkawecki-jt (approved):
Approving. The permission gate, the type-safe status maps (`satisfies Record<EarlyPayTransactionStatus, …>`, so a new API status breaks the build rather than falling through) and the masked payment detail all look right, and loading, empty and error states are each handled explicitly instead of collapsing into one.

A few things worth picking up, none of them blocking:

**Enum values reach the table unlocalised.** `payment_method`, `payment_type` and `provider` render `getValue() ?? '-'`, so the column shows the raw API value while the filter slide-over shows a localised label for the same enum (`EarlyPay.filter.paymentMethods.*`). Someone filters by "Bank account" and then reads `BANK_ACCOUNT` in the result.

**`error.message` goes straight to the screen.** In `TransactionsTable`, `error.message || t('EarlyPay.table.loadError')` means the localised string is only used when the raw message is empty, so whatever the client throws is what an operator sees. Preferring the translation and keeping the original for logs would read better and leak less on a fintech screen.

**Amounts follow the browser locale, dates follow the app's.** `formatEarlyPayAmount` uses `navigator.language` while `formatEarlyPayDateTime` goes through `src/services/date`. A user reading the UI in Spanish on an English browser gets English number formatting beside Spanish dates, in the same row.

**`maximumFractionDigits: 20` with `minimumFractionDigits: 0`** renders 12.5 with one decimal and any float artefact in full, so the money columns can come out ragged. Currency defaults are usually what is wanted for an amount column.

**The permission check reports failure before it can know.** `user_permissions?.some(...)` is `undefined` while auth is still loading, so the tab shows the red "no permission" message first and then swaps in the table. Worth distinguishing "loading" from "denied", since the denied state is styled as an error.

**Filter options are only half derived.** `statusOptions` comes from `Object.values(GetEarlyPayWorkerTransactionsStatusItem)`, but failure reasons and payment methods are enumerated by hand. A new value added to the API shows up in the status filter automatically and silently goes missing from the other two.

**Smaller one:** `handlePaginationChange` falls back to `perPage = 20` while the feature's own default is 10.

wjkawecki-jt:
Please disregard the review above - it was submitted against this pull request by mistake and was meant for a different repository. Nothing in it applies here. Apologies for the noise.

PR #494, opened 2026-07-12
fix(fm-lock): harness-pid discovery works when ps is blind

Quipex:
## Problem

`bin/fm-lock.sh` could not acquire the per-home session lock from a harness whose bash sandbox cannot traverse its own ancestry via `ps`. Reported from Claude Code (Sonnet 5): `fm-lock.sh` always exited 1 with `error: cannot locate harness process in ancestry`, regardless of lock state (absent, stale, or held). This left `fm-session-start.sh` stuck in read-only mode, so the locked bootstrap sweeps (fleet sync, wake-drain) never ran.

Reproduced identically:
- Inside Claude Code's sandboxed Bash tool.
- Via Claude Code's `!` direct-shell passthrough.

Root cause: `harness_pid()` walked 8 ancestors from `\$\$` using `ps -o comm=`, but `ps -p \$\$` reported `no such process` in that session's shell context - the script's own PID was invisible to `ps`.

A second, pre-existing bug in `holder_alive()` meant a pi-held lock was always reported `stale`: it built `"<basename(comm)> <args>"` = `"pi pi"` and grepped with `^pi\$`, which requires the whole line to be `pi`. The concatenation never matched, so a second pi session could clobber a live one.

## Fix

`harness_pid()` now resolves the harness PID in order:
1. `FM_HARNESS_PID` env override - liveness-verified via `kill -0` (works where `ps` is blind). Escape hatch for any harness whose bash sandbox breaks `ps` ancestry.
2. The existing `ps` ancestry walk - unchanged for harnesses where `ps` works (pi in a normal terminal, etc.).
3. Env-marker fallback - when `ps` cannot see our own ancestry but a verified harness marker is present (`CLAUDECODE=1`, `PI_CODING_AGENT=true`, `GROK_AGENT=1`), trust `\$\$` via the marker. `kill -0` confirms liveness.

`holder_alive()` now:
- Matches the basename exactly (as `harness_pid` does), so a pi-held lock (basename `pi`) is recognized. The prior `"<basename> <args>"` + `^pi\$` concatenation never matched.
- Fails safe when `ps` cannot inspect the holder: `kill -0` says live, so treat an uninspectable live holder as held rather than let a second session steal the lock based on our own blindness.

## Verification (empirical, per CONTRIBUTING harness-detection rule)

- `bash -n` syntax check on the whole toolbelt: clean.
- `bin/fm-lint.sh` canonical set (ShellCheck 0.11.0, pinned): exit 0.
- `bash tests/fm-lock.test.sh`: 6/6 pass.
- `holder_alive` against a real pi process (PID of a live `pi` harness): now reports `held by live harness` (was `stale` before fix).
- env-marker fallback with a ps-blind fake `ps` + `CLAUDECODE=1`: acquires (was `cannot locate harness`).
- `FM_HARNESS_PID` override with a ps-blind fake `ps` + a live child PID: acquires.
- Dead `FM_HARNESS_PID` rejected; acquire falls through and fails (no dead PID written to the lock).
- ps-blind live holder not stolen: a second acquire against a ps-blind held lock refuses with `another live firstmate session holds the lock`.
- Symlink checks: `CLAUDE.md -> AGENTS.md`, `.claude/skills -> ../.agents/skills` OK.

## Scope

Only `bin/fm-lock.sh` (session lock) plus a colocated `tests/fm-lock.test.sh`. No changes to spawn, watcher, or harness-detection mechanics. The env-marker list mirrors `bin/fm-harness.sh`'s `detect_own` layer 1; extend both in lockstep when a new adapter gains a marker.

### DIFF
--- a/bin/fm-lock.sh
+++ b/bin/fm-lock.sh
@@ -3,6 +3,12 @@
 # Writes the harness (agent) process PID found by walking the shell's ancestry,
 # which lives as long as the firstmate session - unlike the transient subshell
 # PID of any one tool call, which is dead moments after it is written.
+# The harness PID is resolved, in order: an explicit FM_HARNESS_PID override
+# (liveness-verified via kill -0), an ancestry walk via ps, or - when ps cannot
+# see the shell's own ancestry but a verified harness env marker is present
+# (e.g. Claude Code's sandboxed Bash tool, which reports "no such process" for
+# its own $$) - the shell's own $$, trusted via the marker since ps cannot
+# confirm it. Liveness is checked with kill -0, which works where ps is blind.
 # Usage: fm-lock.sh           acquire; exit 1 if another live session holds it
 #        fm-lock.sh status    print holder and liveness; always exits 0
 set -u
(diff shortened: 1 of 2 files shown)


Quipex:
Closing: opened directly via gh-axi without the no-mistakes gate signature. Re-opening through 'git push no-mistakes' so the pipeline runs and the PR carries the required signature.

PR #496, opened 2026-07-12
fix(supervision): absorb finished-crew stale churn safely; fix claude send verification; hermetic tests

HamzaZagha:
## Summary

Hardens firstmate's supervision loop against four failure modes observed live on 2026-07-08/09 during a multi-crew session, plus a claude-harness send-verification bug:

- **Wedge re-check + surfaced-status stale dedup in `fm-watch.sh`** — stops the nuisance stale-wake churn for finished/parked crews (each one previously fired a wake every ~4 minutes for the task's remaining lifetime) while keeping the hard guarantee that a crew that stops without a new status is surfaced. The dedup is restricted to genuinely finished-shaped statuses (anchored verb classification, not unanchored substring matching), and a steered crew is re-armed via marker supersession so a post-steer hard-stop stays visible.
- **NBSP composer fix for claude submit-verification in `fm-send.sh`** — claude's composer renders a non-breaking-space prompt row that the previous cursor-row read misclassified, producing false "Enter swallowed; text left in composer" errors on nearly every skill-invocation send. Verified live.
- **Heartbeat/steer interaction** — the heartbeat backstop no longer re-marks a steered crew's stale line (which silently re-armed the dedup and made a steered-then-frozen crew invisible), and no longer fires a guaranteed nuisance wake after every steer.
- **Test hermeticity + portability** — MISSING-tool suites now build a host-hermetic symlink PATH farm portably (no GNU-only `cp -sn`, works on macOS/BSD); fixes a fixture leak where a test wrote task state into the real `FM_HOME` and fired a live watcher signal during this branch's own validation; stops a turnend-hook EPIPE crash.

## Validation

Full no-mistakes pipeline green through review (3 rounds, all findings fixed or explicitly accepted), test, document, and lint steps. The run's push step failed only for repo-permission reasons (this contribution comes from a fork), hence the manually-opened PR; validation evidence commits are in the branch per the pipeline's policy.


HamzaZagha:
Withdrawing — we'll maintain these changes in our fork going forward.

PR #529, opened 2026-07-13
feat(omp): add Oh My Pi harness adapter

LoneExile:
## Intent

Add Oh My Pi (`omp`) as a first-class firstmate harness adapter (primary or crew) alongside claude/codex/opencode/pi/grok, faithfully mirroring the verified pi adapter. Adds the omp backend to `fm-harness.sh` (detect `OMPCODE` before `CLAUDECODE`), an `fm-spawn.sh` launch template, a turnend-guard + watch extension under `.omp/extensions` mirroring `.pi/extensions`, an intent reader, lock detection via the `bun` interpreter, and the fixed harness allowlists (bootstrap, `AGENTS.md` verified list, spawn header). Ports the cd-guard seatbelt to OMP's `turn_end` event; omp behaves like pi for gate-instruction neutralization.
The omp crewmate launch is an interactive positional session wrapped in the agent-secrets prefix like every sibling adapter: `omp --auto-approve <flags> -e <ext> "$(cat <brief>)"` (ship/scout) / `... -e <turnend> -e <watch> ...` (secondmate). There is no `-p`/`--mode json`/`--no-session` invocation.
Rebased onto latest upstream `main` through #658 (local agent secret injection); the omp launch template adopts the `$agent_secrets_prefix` wrapping exactly like pi so omp crewmates also receive 1Password secret injection.

## What Changed

- Add Oh My Pi (`omp`) as a first-class firstmate harness backend: `fm-harness.sh` detects `OMPCODE` ahead of `CLAUDECODE`, `fm-spawn.sh` grows an omp launch template (ship/scout and secondmate) wrapped in `$agent_secrets_prefix`, `fm-lock.sh` recognizes omp via its `bun` interpreter, `fm-teardown.sh` cleans state-resident omp extensions, and tmux/watch busy-state regexes gain omp indicators.
- Add `.omp/extensions/fm-primary-turnend-guard.ts` and `fm-primary-omp-watch.ts` mirroring the `.pi` extensions, porting the cd-guard seatbelt onto OMP's `turn_end` event and writing the turn-end signal.
- Register omp across the fixed harness allowlists (bootstrap, `AGENTS.md` verified list, spawn header, skill/architecture/configuration docs), add a `docs/supervision-protocols/omp.md` protocol page, and add live-e2e, extension-types, and ambient-`OMPCODE`-isolation test coverage.
[...]

PR #527, opened 2026-07-13
fix: align Grok effort handling with 0.2.99

korallis:
## Intent

Fix stale grok effort-level documentation and spawn handling in firstmate after grok 0.2.99 tightened --reasoning-effort to only low/medium/high (rejecting both xhigh and max). Empirically verified on grok 0.2.99 with exact commands/output that MUST appear in the PR description:

$ grok --version
grok 0.2.99 (b1b49ccb71a7) [stable]

$ grok --reasoning-effort xhigh -p ping
--effort/--reasoning-effort: unknown effort level 'xhigh'; use one of: high, medium, low
(exit 1)

$ grok --reasoning-effort max -p ping
--effort/--reasoning-effort: unknown effort level 'max'; use one of: high, medium, low
(exit 1)

high is accepted (starts a session / does not reject at parse).

Updated harness-adapters skill table and grok section, changed fm-spawn.sh so grok omits xhigh (like max) instead of passing a rejected flag, aligned crew-dispatch bootstrap validation so grok:xhigh is flagged invalid, and extended colocated tests. Minimal accuracy fix only; not a refactor. Shared vocabulary still accepts xhigh/max as profile axis values for other harnesses; for grok they are recorded in meta but omitted from the launch command. Prior ask-user on grok-auth-evidence was approved: worktree empirical proof satisfies the evidence bar. Push via fork korallis/firstmate; open PR against upstream kunchenguid/firstmate.

## What Changed

- Limit Grok launch and crew-dispatch validation to the `low`, `medium`, and `high` reasoning-effort levels supported by Grok 0.2.99; unsupported requested values remain in task metadata but have no launch flag.
- Update harness-adapter documentation with Grok 0.2.99’s effort ceiling and centralized launch-profile guidance.
- Add coverage for rejected and omitted Grok `xhigh` effort, and refresh `fm-gotmp` teardown fixtures for current helper dependencies.

## Risk Assessment

🚨 High: Captain, the code patch is narrowly aligned, but a mandatory delivery and evidence requirement is currently unmet.

## Testing

The configured full-suite baseline had already passed. Focused bootstrap and spawn integration tests passed; live Grok 0.2.99 transcripts show xhigh/max parse rejection and high progressing beyond parsing, while isolated end-to-end spawn evidence confirms omitted rejected flags, retained metadata, and bootstrap validation. No worktree transients remain.

<details>
<summary>Evidence: Grok 0.2.99 parser compatibility transcript</summary>

```text
$ grok --version
[...]

PR #526, opened 2026-07-13
feat: add rovo (Atlassian Rovo Dev CLI) as a verified crewmate harness

sathish316:
Adds `rovo` (Atlassian Rovo Dev CLI, the standalone `rovo` binary) as a verified **crewmate/scout** harness adapter.

## What changed
- **Detection** (`bin/fm-harness.sh`): `ROVODEV_CLI=1` env marker + `*rovo*` process-ancestry match.
- **Launch** (`bin/fm-spawn.sh`): `rovo run --yolo -i "$(cat <brief>)"`; turn-end is a documented no-op (no verified hook — supervision uses the status-file protocol + stale detection).
- **Busy signature** (`bin/fm-watch.sh`, `bin/fm-tmux-lib.sh`): `Rovo is thinking`.
- **Dispatch validation** (`bin/fm-bootstrap.sh`): accepts `rovo`; no effort flag (validated like opencode).
- **Knowledge**: full verified facts in the `harness-adapters` skill; `rovo` added to the AGENTS.md verified-adapter list.

## Verification
Empirically verified 2026-07-13 (Rovo CLI 202607.8.1): launch, `--yolo` autonomy, busy signature, single-Escape interrupt, `/exit`, `--resume`, env markers, and no trust dialog. Then exercised end-to-end on a real scout task (twg-cli JSM subcommand enumeration), which completed cleanly.

## Scope / caveats
- **Crewmate/scout only** — secondmate use is unverified (`pane_current_command` liveness not confirmed), so `bin/backends/tmux.sh` deliberately omits rovo.
- `--yolo` auto-approves file/bash but NOT Atlassian-data/MCP tools (those still prompt).
- No firstmate turn-end hook yet.
- ShellCheck lint was not run locally (not installed); relying on CI.

PR #548, opened 2026-07-14
fix(orca): P5 fallback - resolve parent via git-common-dir or basename

Andresito11-cpu:
When the orca backend passes `--parent-path` that doesn't exactly match an orca-registered cwd, the previous P5 `register` returned rc=4 (no parent session) and the worktree silently fell under orca's 'Hiding N discovered worktree' gate.

Most of the time, the parent's orca-registered path differs from the `--parent-path` arg only because the captain and firstmate each have a sibling clone (captain's `~/Desktop/falkordb-stak` and firstmate's `projects/falkordb-stak`), and the `.git` directory underneath is the same.

Add a three-step fallback chain in `resolve_parent_session_uuid`:
  1. Exact cwd match (trailing-slash + realpath tolerant)
  2. Git-common-dir match: `realpath --git-common-dir` of parent vs each fmod-listed session's cwd. Sibling clones share this.
  3. Basename match: last-ditch when the repo's name is unique enough.

The script-level exit-4 error and the spawn-time 'warn: orca UI sidebar registration failed' message are unchanged so the captain can still diagnose the rare case where all three fail.

Follows #546 + the squashed #3.

Tests (all green): 16 assertions in `tests/fm-mod-orca-ui.test.sh`.


ami230392:
# Review — PR 548 (issue #527): "declare the gap — a Portal-listed file we did not read is never silent"

Reviewed at `1a3710f` (merge-base `b228e20`), in a disposable worktree. Read-only review; no push, no PR.
Mandate: attack the **detection path** (can a listed-but-unread file still slip through silently?), the **migration**, and **every return arm's test**.

**Verdict: approve with follow-ups.** The core mechanism is right and the receipt is honest. I found one real residual silent false negative (F1, time-displaced, arguably #528's), one contract hole worth closing in this PR (F2), and one migration-level fence in the wrong layer (F3). None of them makes this PR worse than main; all three matter because #529 flips eligibility on top of this guarantee.

---

## What I did

| Command | Result |
|---|---|
| `git fetch origin pull/548/head && git checkout FETCH_HEAD` | at `1a3710f` |
| `pnpm --filter worker test read-tender` (fresh checkout, stale `shared/dist`) | **13 failed / 27 passed** |
| `pnpm --filter @tendersathi/shared build` then re-run | **40 passed (3 files)** — matches the receipt |
| `pnpm --filter web test coverage-gap-notice` | **7 passed** — matches the receipt |
| red-before-green: `git checkout origin/main -- apps/worker/src/documents/readTender.ts` + `pnpm --filter worker test read-tender-coverage` | **6 failed / 6 passed** — reproduces the receipt exactly |
| 6 adversarial probes written as a scratch vitest file against the real entry point + fakes (deleted after) | see below |

The first line is **not** a defect in this PR — it is the documented repo gotcha (root `AGENTS.md`: "CI and fresh clones must build `@tendersathi/shared` first"). Recording it because a reviewer on a clean checkout will hit it and mistake it for red CI.

Two parallel sub-agents ran the repo's standard two-axis review (Standards / Spec). Standards: no hard violations — `hi.json`/`en.json` are 1:1 on the three new keys, the fetch-loop catch logs route/tender/filename/error-class before degrading, `safePortalUrl` gates `href` to `http(s)`, `apps/worker/AGENTS.md`'s `Last verified` line is updated. Spec: all six acceptance criteria and all four Seam-1 tests in #527 are satisfied, with the record-bounding constants (`MAX_DECLARED_*`) being the only work beyond #527's literal text — justified by root `AGENTS.md`'s validate-size-not-only-shape rule and disclosed in the PR body.

One claim from the spec agent I checked and **rejected**: it reported the `coverageKnown: false` branch as untested at the worker seam. It is tested — `apps/worker/test/read-tender.test.ts:438` ("F5: never serves a row that recorded no coverage as complete coverage") seeds exactly that `StoredRead` through the store and asserts `coverage_incomplete`.

## The detection path, attacked

Enumerated every way a listed file can leave `performRead`. Within one read the loop is airtight: `listed` is built from `listing.files` (`readTender.ts:275`), `fetchable` is its first 3, every entry either pushes to `coveredFiles` (:300) or to `failed` (:302), and everything past the bound is `overflow` (:281). There is no branch where a listed entry lands on neither list, and naming both lists off the listing rather than off `PortalFileBytes.filename` genuinely closes the rename hole the PR describes. `serve()` (:407) replays the stored gap on every later open, and a NULL-coverage row is served as a declared gap. All of that is correct.

The gaps are at the edges of that loop.

### F1 — HIGH — the comparison happens once, at first read; a file listed *later* is silent

`guardedRead` returns from cache at `readTender.ts:219-220`, **before** `deps.portal.list` is ever called. So the listing-vs-read comparison is a one-shot at first read, and `serve()` can only replay the gap that was known then.

Probe (real entry point, fake Portal):

```
P1 first: facts_available  second: facts_available
   portal reqs: [ 'list:1111…', 'fetch:NIT.pdf' ]
```
— read a Tender listing one file (complete coverage, `facts_available`), then add `Eligibility_annexure.pdf` to the Portal listing, then open it again as a second org. The second open serves `facts_available` from cache and makes **zero** Portal requests. The newly listed, never-read eligibility attachment is invisible.

This is #527's exact failure mode, displaced in time — and the displacement is not hypothetical: a buyer adding an eligibility annexure by **corrigendum after publication** is a routine GePNIC event, and #528 exists precisely because corrigenda invalidate a stored extract. The fix is almost certainly #528's, not this PR's. But two things should change:

1. The PR body's **NOT checked** section should say it: *"coverage is compared once, at first read; a file the Portal lists after that is not detected until #528 invalidates the row."* Right now the body says the gap "comes back on EVERY open", which is true of the record and easy to read as a claim about the Portal.
2. **#528 must invalidate the coverage columns, not only the extract**, and carry a test for it. If a corrigendum re-read writes a fresh row, F1 closes. If it only marks the extract stale while `covered_files`/`unread_files` survive, F1 stays open under a ticket that looks like it closed it.

This matters most for #529: the flag flips on a guarantee that currently holds at t=0 and decays afterwards.

### F2 — MEDIUM — coverage is credited on fetch *return*, not on the file being readable

`coveredFiles.push(entry.name)` runs as soon as `deps.portal.fetch` resolves (`readTender.ts:299-300`). What makes an unextractable file honest today is `sniffPortalFile` (`documents/fileSniff.ts:50`) throwing `PortalReadError("file_unreadable")` — but that lives in the **adapter**, and `performRead` never re-checks `PortalFileBytes.mime` against `EXTRACTABLE_MIME_TYPES` before making the coverage claim.

Probe: a Portal that returns `{ filename: "drawings.zip", mime: "application/zip" }` without sniffing →

```
P2: facts_available   coveredFiles: ["NIT.pdf","drawings.zip"]
```

A file the model never read, recorded as covered, served as complete coverage. No real adapter exists yet (#526), so this is a hole in a contract rather than a live bug — which is exactly why it is cheap to close now, before #526 is written against it. Three lines in the fetch loop:

```ts
const bytes = await deps.portal.fetch(tender, entry.file);
if (!EXTRACTABLE_MIME_TYPES.has(bytes.mime)) throw new PortalReadError("file_unreadable", `${entry.name}: ${bytes.mime}`);
```

The coverage claim should be made at the point where the file is proved readable, not where it is proved downloadable. Everything else in this PR follows exactly that principle — the listing is the source of truth *because* an adapter must not be trusted to keep a name. The same distrust applies to the format.

### F3 — MEDIUM — the migration turns "coverage unknown" into "coverage complete"

Migration 0049 (already applied, not touched by this PR) is explicit: *"NULL means coverage unknown"*, and `store.ts:55` derives `coverageKnown` from exactly that NULL-ness. But the RPC body writes:

```sql
values (p_tender, 'ok', p_extract, coalesce(p_covered, '{}'), coalesce(p_unread, '{}'), …)
```
(`supabase/migrations/0049_tender_read_attempts.sql:206-215`)

A caller passing NULL therefore persists `'{}' / '{}'` — non-NULL, so `coverageKnown` reads **true**, `unreadFiles` is empty, and `serve()` answers `facts_available`. Probe P6 confirms `serve()` treats that row as complete coverage. The only caller today spreads real arrays (`store.ts:68-69`), so nothing is broken in production — but the fence protecting this feature's central invariant lives in TypeScript, while the SQL actively converts unknown into complete. Next migration: `if p_covered is null or p_unread is null then return 'invalid'; end if;` (or just drop the two `coalesce`s and let NULL persist as NULL). Additive, no rewrite of 0049.

### F4 — LOW/MED — `declaredName` bounds length but strips no control characters

`declaredName` (`readTender.ts:84-88`) trims, substitutes a positional stand-in, and truncates by code point. It does not remove C0 controls. Probe P3 with a listed filename containing `U+0000`:

```
P3 unreadFiles: ["bad[NUL]name.pdf"]
```

Postgres rejects NUL in `text`/`jsonb`, so the `store_tender_read_extract` call throws, `store.write` logs "paid extract not stored", and the caller's catch turns a **fully paid read** (Portal requests + rupee-lane debit + model tokens) into `could_not_read`. That is the exact cost the code-point-truncation docblock argues about, one character class short of covered. One-line fix in the same function: strip `\p{C}` before bounding.

### F5 — LOW — duplicate listed names make the record un-keyable

Probe P4, listing `["dup.pdf","dup.pdf","other.pdf"]` → `coveredFiles: ["dup.pdf","dup.pdf","other.pdf"]`. If one copy fetches and the other fails, the identical string sits on both lists, and any later listing-vs-record comparison (F1's eventual fix) cannot match them. The positional stand-in mechanism already exists for nameless files; extending it to disambiguate duplicates would make the record a key rather than a label. Not worth blocking on.

## What holds up

- Naming both lists off the listing — the PR's headline claim — is real and is what the six red-before-green tests prove.
- `coverageGapNotice` as the single mapping, carrying no count and no filenames, is the right call: it is the only copy that stays true across the `coverageKnown: false` arm, and the component test proves 1, 3 and 40 unread files render an identical DOM.
- `could_not_read` when nothing was read at all (`readTender.ts:313`) is correct, not a narrowing of #527: a gap declared against zero facts is not a gap.
- `quota_exhausted` → `null` notice: correct per spec #512 Decision 10, unit-tested.
- Bounding applied once over `[...failed, ...overflow]` rather than per-list — the comment explains why, and the test covers failures-plus-overflow together.
- Hindi copy reads like a person wrote it and renders from the shipped `messages/hi.json`.
- The receipt's "NOT checked" section is unusually honest and every claim in it I could check held.

## Recommendation

Merge, with: F2 and F4 folded into this PR (each is a few lines in `readTender.ts`, both are inside the unit this PR already owns), F3 filed as a one-line migration follow-up, and F1 written into both the PR body's NOT-checked section and #528's acceptance criteria before #529 flips the flag. F5 optional.

PR #552, opened 2026-07-14
docs: purge the no-mistakes footprint from docs and skills

webjema:
The v2 redesign deleted the no-mistakes subsystem (binary, `~/.no-mistakes/`, daemon, user skill) but left its footprint in tracked docs and skills, so agents were still being pointed at a tool that does not exist on this machine. This removes the footprint. Docs and skills only.

## Verified dead before removing

Each removal was checked against the tree first, not assumed:

- **`.no-mistakes.yaml` is not tracked** (`git ls-files`), so `docs/configuration.md`'s "Gate defaults" section documented a file that is gone. The behavior-test baseline that section described (tmux on PATH, `tmux -V`, every `tests/*.test.sh` under bash, fail on non-zero) is already owned by `.github/workflows/ci.yml`, so nothing is lost by deleting it.
- **`FM_CREW_STATE_NM_TIMEOUT` and `FM_CREW_STATE_RUNS_LIMIT` have zero readers** anywhere in the tree - the only hits were the doc lines documenting them. Documented knobs that did nothing.
- **`bin/fm-gate-refuse-lib.sh` no longer exists**, yet `docs/scripts.md` still listed it in the toolbelt table and linked an `architecture.md#no-mistakes-gate-authority-boundary` anchor that is also gone. (Not in the original scope; found by grep.)
- **`bin/fm-home-seed.sh` has no clone-initialization step**, so the config sentence describing one was dead. That sentence conflated a dead fact with a live one ("refuses to mutate a preexisting clone") - the live half is owned by `fm-home-seed.sh`'s own header (lines 20-22), so per the one-owner rule the doc correctly does not restate it.

## Harness knowledge preserved, not dropped

The `## no-mistakes skill invocation` section is deleted, but its verified facts survive:

- `Skill invocation` examples re-anchored on `/code-review`, a skill that exists.
- **opencode's and pi's "no separate verified form" facts lived ONLY in that deleted section** - neither had a `Skill invocation` table row. They are relocated into their own harness tables rather than lost. All five harnesses now carry the row (previously three).
- Grok's hard-won popup/Enter incident substance (slash-autocomplete, argument-hint placeholder, genuine second Enter, `fm-send`'s retried Enter) is kept intact, plus its end-to-end verification claim.
- Codex's `$`-popup settle explanation keeps its mechanism; only the dead `$no-mistakes` example is re-anchored.

The retired three-mode vocabulary in the secondmate-routes wording is rewritten to the live `PR` / `local-only` modes.

## Out of scope, deliberately untouched

- `bin/fm-project-mode.sh`'s legacy-token shim (`no-mistakes|direct-PR) mode=PR`) - documented back-compat so an old registry line still parses. Left exactly as is.

## Handoff note for the parallel supervision crew

`tests/` is reserved for the parallel crew, so I did not edit it. Remaining `no-mistakes` references there, for their judgment - most look intentional rather than rot:

- `tests/fm-brief.test.sh` - deliberately covers the legacy-token shim (asserts `[no-mistakes]` maps to `mode=PR`). Should stay as long as the shim does.
- `tests/fm-backend.test.sh:82-91` - strips a `fm-gate-refuse-lib.sh` source line from an extracted baseline, with a comment explaining the pipeline is gone. Intentional.
- `tests/fm-send-popup-settle.test.sh` - uses `$no-mistakes` / `/no-mistakes` as arbitrary message payloads to exercise popup settle. Harmless strings; could be re-anchored on any skill name.
- `mode=no-mistakes` legacy fixtures in `fm-gotmp`, `fm-bearings-snapshot`, `fm-backend`, `fm-send-secondmate-marker`, `fm-session-start` - exercise the shim path.
- `tests/fm-watch-triage.test.sh`, `tests/fm-daemon.test.sh` - comments and fixtures referencing the retired validation pipeline; these are the ones most likely to be genuine rot.

No reference remains in `AGENTS.md` or `docs/architecture.md`.

## Verification

- `bin/fm-lint.sh` clean (ShellCheck 0.11.0, pinned).
- Full behavior suite green (every `tests/*.test.sh`).
- `git grep -niE 'no-?mistakes'` now returns only the deliberate shim in `bin/fm-project-mode.sh` and the reserved `tests/` files listed above.
- Skill frontmatter still parses; all harness tables well-formed; every script named in `docs/scripts.md` now actually exists on disk (it did not before).

## Direction

Honors **one owner per contract** (drops doc restatements of contracts the script headers and CI already own), the **token budget** (removes dead instructions agents would otherwise load and act on), and **root cause over symptom** (deletes the dead footprint rather than papering over it).


webjema:
Opened against the wrong repository by mistake; this belongs on the fork. Closing.

PR #563, opened 2026-07-14
fix(supervision): keep the watcher and away-mode daemon alive when the harness reaps their background task

hmreumann:
## The bug

On a background-notify primary (claude, grok), firstmate arms its watcher (`bin/fm-watch-arm.sh`) and its away-mode daemon (`bin/fm-afk-start.sh`) as harness-tracked **background tasks**. The watcher and daemon were **children** of those tasks.

When the harness reaps a background task, it SIGTERMs the task's whole **process group** — so the watcher and daemon died with it.

Observed on a live home on 2026-07-12: background tasks were terminated from outside within a minute or two of starting (clean `Terminated: 15`). The away-mode daemon was killed 3 times in a row and the armed watcher roughly 8, each time leaving `pgrep fm-watch.sh` empty. Load average was normal (~1.3), so this was not the OOM killer. The worst case is away mode: `state/.afk` stays set while nothing is supervising — precisely the failure away mode exists to prevent.

It was not lossy only by luck: each kill arrived as a task notification that woke firstmate, and `state/.wake-queue` is durable, so drain-then-rearm recovered. But supervision was silently dead in the gap.

## Evidence: it is a process-group kill

A background task forked two children — one ordinary child in the task's process group, one detached into its own session — and the task was then reaped:

```
parent   = 37805  pgid 37803   -> KILLED
plain    = 37820  pgid 37803   -> KILLED    (ordinary child, same process group)
detached = 37824  pgid 37824   -> SURVIVED  (own session, reparented to init)
```

Escaping the **process group** is what makes a supervision process survive; `nohup` and a bare `&` do not.

## The fix: detached watcher, follower arm

The watcher is now launched **detached** (its own session and process group), and `fm-watch-arm.sh` became its **follower** rather than its parent. A reap now ends only the follower, while the watcher keeps the singleton lock, the liveness beacon, and the durable wake queue.

**The wake path is deliberately unchanged.** The watcher still enqueues an actionable wake and exits; the follower observes that and completes; and the background task's completion is still exactly what wakes the primary. A reap now costs one empty drain plus a re-arm that reports `attached` straight back onto the still-live watcher.

The arm's HUP/TERM trap no longer kills the watcher — the old `cleanup_child` did, and inverting that **is** the fix. Tests that asserted the old kill-the-child behavior were rewritten to the new contract.

The away-mode daemon takes the same detachment (`fm-afk-start.sh --detach`), because it must outlive the turn by construction. Its watcher child is then detached transitively.

## Portability: setsid(1), with a perl fallback

`setsid(1)` does not exist on macOS, and perl is not guaranteed everywhere. Making either an unconditional dependency of the supervision backbone would mean supervision could **never start** on a host lacking it — a worse failure than the one being fixed. So `bin/fm-detach-lib.sh` prefers `setsid(1)`, falls back to perl `fork` + `POSIX::setsid()`, and fails loudly with an actionable message if neither is present. README's Requirements section records this.

## Orphan control: idle self-exit

Because nothing kills a detached watcher when its session ends, it can outlive the session that armed it. That orphan is self-adopting — the singleton lock plus attach-on-re-arm hand it straight back to the next session, and it keeps filling the durable queue meanwhile — so it is bounded rather than evicted: the watcher self-exits after 40 consecutive polls (~10 minutes at the default 15s `FM_POLL`) in which the fleet is empty, X mode is off, and away mode is off. Each condition is load-bearing: X mode deliberately keeps a watcher armed on an empty fleet, and the away-mode daemon restarts the watcher it wraps, so idle-exiting under afk would spin a restart loop.

The bounded **foreground** checkpoint (`fm-watch-checkpoint.sh`, the Codex path) is already bounded by its own timeout, so it disables idle-exit for its own invocation — otherwise a tuned checkpoint window could return exit 0 with a line matching neither branch of the Codex protocol.

Cross-session eviction was considered and **rejected**: it would add a cross-session kill path to the supervision backbone for no gain the singleton lock does not already provide.

## Guards are unaffected

The watcher writes `.watch.lock`'s `pid`, `pid-identity`, `fm-home`, and `watcher-path` from its own main shell, so a detached watcher is still the identity-matched lock holder with a fresh beacon. Verified against an isolated primary-shaped home with one in-flight task:

| watcher state | `fm-guard.sh` | `fm-turnend-guard.sh` |
| --- | --- | --- |
| none | `WATCHER DOWN` banner | `TURN WOULD END BLIND`, exit 2 |
| detached, alive | silent | silent, exit 0 |
| detached, killed | silent only inside the 300s beacon grace (pre-existing) | `TURN WOULD END BLIND`, exit 2 |

## Alternatives considered and rejected

- **Switch Claude to bounded foreground checkpoints.** Rejected on fact: a foreground checkpoint cannot fix the away-mode daemon, which must outlive the turn by construction.
- **Hybrid (detach, with a checkpoint fallback).** Rejected: two supervision shapes on one harness is more surface for exactly the bug class being fixed.

## Tests

Colocated in the existing suites, no new runner:

- the watcher survives a SIGTERM of the arm's **whole process group** (the harness reap), keeping its lock and beacon;
- a signalled arm stands down **without** killing the detached watcher;
- the away-mode daemon survives a reap of its launching task, and lands outside that task's process group;
- the watcher idle-exits only with an empty fleet, X mode off, and away mode off, and keeps running when any one of those says otherwise;
- an attached arm surfaces the idle-exit of the watcher it followed;
- a foreground checkpoint never idle-exits.

## Maintainer note

The supervision shape here — detached watcher plus follower arm, and the orphan policy of accepting a self-adopting orphan bounded by an idle self-exit rather than cross-session eviction — was chosen deliberately over the alternatives listed above. This is the supervision backbone, so the repo owner has **final say on that shape at merge**: if you would prefer a different one, say so and it will be reworked rather than argued.


kunchenguid:

PR #564, opened 2026-07-14
pool: own worktree provisioning - kill the 60s cliff, keep a warm spare, notice a shrinking pool

webjema:
Firstmate now owns worktree provisioning instead of leaving it to luck: it knows its pool's state, keeps a warm spare ahead of the crew, and notices when a crash quietly eats a slot.

## The measurement everything hangs on

Measured on this box (treehouse v2.0.0, optiroq, warm shared npm cache), by leasing every existing slot to force treehouse to provision a genuinely cold one:

| | |
|---|---|
| warm slot (deps present, `post_create` fast-exits) | **2s** |
| **cold slot** (fresh worktree + `post_create` npm install) | **137s** |

137s is **2.3x** the old 60s timeout. So the first spawn against a project with no warm slot - a fresh box, a new project, an exhausted pool - was reliably killed by a bound that named none of those causes. A cold npm cache is slower still.

## 1. Kill the 60s cliff (`77cd83f`)

Raising 60 to 600 would only fail later and still could not tell an install from a wedge. `bin/fm-provision-lib.sh` classifies the pane every tick instead:

- **ENTERED** - the cwd left the project; the worktree is ours.
- **BUSY** - the pane's shell still has a live descendant (treehouse, its hook, npm). Something is running, so wait. This is the cold install the old poll killed.
- **EXITED** - provably back at an idle prompt, cwd unmoved. treehouse failed: fail *fast*, and say why from its own last words.

The descendant check is structural rather than scraped: an idle shell has no children, and treehouse's whole pipeline runs as one tree. An unreadable pane pid is `unknown`, **never** `idle` - a false-dead read would kill a healthy install, the trap `fm_backend_agent_alive` already guards against.

Two bounds keep a wedge from hanging a spawn: `FM_SPAWN_WORKTREE_STALL` (300s busy but silent) and `FM_SPAWN_WORKTREE_TIMEOUT` (900s ceiling). One opaque string becomes three actionable ones: **pool exhausted** (treehouse's own `max_trees` refusal), **treehouse errored** (its message relayed), **still warming** at the ceiling.

## 2. The always-plus-one warm spare (`075716c`)

For every project with work in flight, one slot sits free and warm; when the last free slot is taken, the next is provisioned preventively in the background. Self-tuning with no target size: a captain who habitually runs 4 tasks settles at 5 and stops growing, because the 5th is never consumed.

Warming is thin, because treehouse already does the work - `get --lease` (create-or-reset + `post_create`, i.e. install deps), then `return`. **Verified end to end against the real binary: after the return, optiroq's `node_modules` trees survived intact - 2.7 GB** - because treehouse's reset is `git clean -fd` with no `-x`. That is exactly what makes a returned slot stay warm, and why this reimplements nothing. The **lease** is the safety property: held across the whole install, so no crew is handed a half-installed slot.

Triggered by the watcher on its slow `FM_CHECK_INTERVAL` sweep - a short-lived detached child of a loop already running, so **no new always-on process**, never the hot path. A failed warm logs and retires quietly: it cannot corrupt a slot, block a spawn, or wake the captain. **One warmer per pool**, keyed by the project's physical path (what treehouse keys a pool by) under the shared treehouse root - secondmate homes share pools, and two warmers racing would over-provision by GBs.

Two ceilings that **stop and report** rather than fill the disk: a per-pool disk budget (default 20 GB, configurable) and treehouse's own `max_trees`. The next slot's size is estimated from the mean of the existing ones, because optiroq is ~2.8 GB/slot while firstmate is ~6 MB/slot - a fixed slot *count* would treat those identically.

Deliberately **not** in scope: pre-spawning idle agents. Warm the workspace; spawn the agent only when a real task exists.

## 3. A crashed crew must not silently eat a slot (`fb35e93`)

This bit us. Three mid-task reboots each left the crew's slot **dirty**, and - verified - `treehouse get` skips a dirty slot *forever* while `treehouse prune` refuses to reclaim it. The pool silently shrank and nothing noticed; a human found it by looking.

`bin/fm-pool-status.sh` is firstmate looking, at session start, in bash. It reports slots that are dirty, stale-leased (a warmer died mid-install), or orphaned, plus any warm a ceiling stopped.

**It never discards anything, and that is the load-bearing property.** A dirty slot may hold a dead crew's unlanded work - today's did, and it was salvaged and shipped. So each line carries the evidence (uncommitted files, unpushed commits), the command to *inspect* first, and an explicitly destructive reclaim command a human must choose to run. Same fail-closed posture as `fm-teardown.sh`.

## Verification

Beyond the tests, this was driven against the real thing:

- **Real treehouse**: a cold slot provisioned and timed (137s); `return` confirmed to preserve 2.7 GB of `node_modules`; the always-plus-one loop watched to provision slot 2 the moment a crew took slot 1; a real dirty slot detected with its real evidence, work intact.
- **Real tmux pane**: a 90s install **waited out and yielding the worktree** (the old poll died at 60s); a pool-exhaustion error failing in **0s** with its own wording; a hung child process called STUCK at the stall bound rather than run to the ceiling.

Driving it surfaced two message bugs the scripted tests could not (`7307ace`): the failure quoted the shell *prompt* back at the operator instead of the error above it, and it claimed a pane was "idle" when a shell blocked in a builtin `read` is indistinguishable from outside.

Self-review (`6102c1b`) also caught a real fail-open: the dirty-slot evidence check used `git log --branches`, but treehouse checks slots out **detached** - so a crew that committed without branching had work `--branches` could not see, and the line would have said "nothing to lose" about a slot holding exactly the work that must not be lost.

## Direction

`bash absorbs, the model judges`, applied to provisioning: the pool's state is classified in bash (`treehouse status`, a structured file, not a pane a model must read) and firstmate knows it instead of discovering exhaustion as a timeout. No new always-on process; the watcher was already running. Every contract lives in one script header; the orchestrator reads nothing new.

`bin/fm-lint.sh` clean, full behavior suite green locally. Each commit carries a colocated behavior test; treehouse is stubbed the way the suite stubs external binaries, so no test needs a multi-GB install.

**Note:** `tests/fm-pi-watch-extension.test.sh` is flaky on this box (it failed once on clean `main`, unrelated to this branch, and passes on re-run). Not touched here.


webjema:
Opened against the wrong repository by mistake - this belongs on the webjema fork, not upstream. Closing.

PR #574, opened 2026-07-14
fix(snapshot): assign newline to var before case pattern to avoid tokenize-time split

ManveerBhullar:
Put back into draft as i think this fix isnt proper. double checking it on my local

## Problem

`$'\n'` inside a `case` pattern expands to a literal newline at tokenize time, splitting the case branch across lines so `;;` becomes an unexpected token:

```
fm-fleet-snapshot.sh: line 690: syntax error near unexpected token `;;
```

1. bash -n syntax check passed
2. fm-bearings-snapshot.sh runs clean (exit 0)
3. fm-fleet-snapshot.sh --json runs clean (exit 0)
4. All 12 tests in tests/fm-fleet-snapshot-view.test.sh pass

Edit: forgot to mention I'm on macOS

### DIFF
--- a/bin/fm-fleet-snapshot.sh
+++ b/bin/fm-fleet-snapshot.sh
@@ -686,8 +686,9 @@ registry_secondmates_json() {
     if [ "$bytes" -gt "$max_bytes" ]; then
       byte_truncated=true
       content=$(printf "%s" "$content" | LC_ALL=C head -c "$max_bytes")
+      nl=$'\n'
       case "$content" in
-        *$'\n'*) content=${content%$'\n'*} ;;
+        *"$nl"*) content=${content%"$nl"*} ;;
         *) content= ;;
       esac
     fi
@@ -783,8 +784,9 @@ bounded_parent_activities_json() {  # <status-file>
     byte_truncated=false
     if [ "$size" -gt "$max_bytes" ]; then
       byte_truncated=true
+      nl=$'\n'
       case "$content" in
-        *$'\n'*) content=${content#*$'\n'} ;;
+        *"$nl"*) content=${content#*"$nl"} ;;
         *) content= ;;
       esac
     fi


zachlandes:
Confirming this one from the field. After a self-update to main, `bin/fm-fleet-snapshot.sh` stopped parsing at all on macOS system bash (3.2.57):

```
bin/fm-fleet-snapshot.sh: line 690: syntax error near unexpected token `;;'
```

bash 3.2 splits `$'\n'` at tokenize time inside a case pattern, so it's a parse error rather than something that only bites when the truncation branch actually runs. The script dies on every invocation, and it takes `fm-fleet-view.sh` and `fm-bearings-snapshot.sh` with it since both call it. I rolled our checkout back to the previous commit to get them working again.

The fix looks correct to me and it covers both call sites (`registry_secondmates_json` and `bounded_parent_activities_json`).

@kunchenguid can you merge this? Anyone on stock macOS bash has a broken snapshot after updating.

kunchenguid:
quick heads up: i reproduced this on stock macOS Bash 3.2. the current patch still fails to parse before the snapshot code runs, so it doesn't fix the underlying issue and breaks snapshot, fleet view, and Bearings.

i'm working on a separate urgent fix now. it'll replace the problematic nested `case` blocks and add plain `/bin/bash` coverage so this doesn't regress.

kunchenguid:
quick update: the replacement fix is now up at https://github.com/kunchenguid/firstmate/pull/578. it removes the Bash 3.2-incompatible nested case pattern and adds stock `/bin/bash` CI coverage. the full suite is green, including the macOS Bash 3.2 path.

i'm going to continue with that PR and leave this one unmerged. thanks again for catching the underlying issue and putting together the first patch - it helped make the compatibility gap concrete.

ManveerBhullar:
Awesome, glad you were able to figure this out. sorry for the very sparse PR and bug report :)

btw i sent you a linkedin connection request last week if you want to keep in touch , hope you dont find this rude to bring up here.

PR #577, opened 2026-07-14
fix(afk): gate away-mode injection on a confirmed live agent

zachlandes:
## Intent

Fix a verified safety hole: the away-mode supervise daemon could type an escalation digest (plus Enter) into a live shell, executing arbitrary commands when nobody is watching. Root cause: inject_msg (bin/fm-supervise-daemon.sh) gated injection ONLY on composer-emptiness, but the default starship/pure/spaceship shell prompt glyph U+276F (the heavy right-angle quotation mark) is byte-identical to the glyph claude draws for its own empty composer, so a supervisor pane whose agent had exited to a login shell read as an empty agent composer and was injected into. Fix: after the composer-empty check, inject_msg now also calls the pre-existing fm_backend_agent_alive probe (bin/fm-backend.sh; already used by the session-start secondmate-liveness sweep) and injects ONLY on a confident 'alive'. Deliberate decisions a diff-only reviewer would not know: (1) This FAILS CLOSED on purpose - both 'dead' (a bare shell) and 'unknown' (an unreadable pane, or a harness whose liveness is unverifiable for the backend, e.g. pi's generic node interpreter) defer, on the stated principle that a missed escalation is an inconvenience but a command typed into the captain's shell is not; deferred escalations are not lost (buffer + max-defer wedge alarm + afk-exit catch-up flush). A pi-based supervisor in away mode therefore stops receiving injected escalations - this is an accepted, documented consequence, not a regression to fix here. (2) I intentionally did NOT remove the glyph from any character class - that is a band-aid because an empty shell prompt and an empty agent composer are indistinguishable by content; the guard asks what process is running instead. (3) fm-send.sh shares the same content-blindness but is deliberate/firstmate-initiated (not autonomous) and a strict alive-gate there would break steering for every unknown-liveness backend (pi, zellij, orca, cmux); I analyzed and documented it with a recommended separate dead-only guard but deliberately did NOT change fm-send in this PR to keep the safety fix focused. Tests: three deterministic inject_msg units in tests/fm-daemon.test.sh plus a real-shell end-to-end Scenario D in tests/fm-afk-inject-e2e.test.sh, each of which fails against the pre-fix daemon (verified by neutering the gate: the suite exits non-zero with an explicit 'typed into a dead shell' / 'SHELL INJECTION' failure). Test-helper changes in tests/wake-helpers.sh add a pane_current_command reply to the fake tmux so the fake supervisor panes (which stand in for a live agent) report as alive. Docs: amended the one-owner composer-emptiness safety policy in docs/herdr-backend.md (new dated subsection) and corrected the now-inaccurate claim in docs/architecture.md that a bare dead-shell prompt cannot receive an escalation. This is a generic bug affecting every firstmate user with a default shell prompt, intended to go upstream.

## What Changed

- `inject_msg` in `bin/fm-supervise-daemon.sh` now probes `fm_backend_agent_alive` after the composer-empty check and injects only on a confident `alive`. An empty starship/pure prompt glyph (U+276F) is byte-identical to claude's empty-composer glyph, so a supervisor pane whose agent had exited to a login shell previously read as an empty composer and got the escalation digest typed in and submitted as shell commands. The guard fails closed: both `dead` and `unknown` (an unreadable pane, or a harness whose liveness the backend cannot attribute, e.g. pi's generic `node`) defer instead of injecting, so a pi-backed supervisor stops receiving injections in away mode - deferred escalations still survive via the buffer, the wedge alarm, and the afk-exit catch-up flush.
- Because the guard can leave away mode permanently unable to deliver, new `bin/fm-afk-canary-lib.sh` derives and words that verdict in one place, and surfaces it on both the arm path (`bin/fm-afk-launch.sh`) and session start (`bin/fm-bootstrap.sh`, as an `AFK_INJECTION_DISABLED:` diagnostic). It probes the running daemon's own recorded target rather than the current pane, so a firstmate restarted into a new pane learns the daemon is still injecting elsewhere, and reports a vanished target instead of staying quiet.
- The wedge alarm now records why the last inject deferred and reports that cause rather than always blaming a busy pane, and gains a separate debounced, latched pane-gone alarm (its own `.subsuper-pane-gone` marker, one alert per absence episode) that declares supervision DOWN when the supervisor pane vanishes. Captain-facing wording keys off the alarm kind, not the last recorded cause, so a buffer wedge can never render as "away mode is dead". Covered by three deterministic `inject_msg` units in `tests/fm-daemon.test.sh` and a real-shell Scenario D in `tests/fm-afk-inject-e2e.test.sh`, each of which fails against the pre-fix daemon; `bin/fm-send.sh` shares the same content-blindness and is documented in `docs/herdr-backend.md` but deliberately left unchanged here.

## Risk Assessment

⚠️ Medium: A large, safety-critical change to the autonomous away-mode injection path that is well-tested and strictly improves on the pre-existing behavior, but it adds a fail-closed gate that fully disables away-mode escalation delivery for any supervisor whose harness a backend cannot attribute (pi, and any node-wrapped harness) — an accepted, documented regression — and ships with three deliberately deferred robustness gaps in the new alarm and canary paths.

## Testing

Baseline test command had already passed; I re-ran the full suite (68 files, 0 failures) and then produced product-level before/after evidence for the safety fix. Using a real login shell displaying the default starship ❯ glyph, the end-to-end Scenario D shows the fixed daemon refuses to inject and defers the escalation, while a neutered (pre-fix) daemon typed the digest into and executed it in the live shell — the test caught this with an explicit SHELL INJECTION failure. The deterministic inject_msg units confirm the same at the function level, including fail-closed-on-unknown and the alive+empty positive path. I restored the daemon file and verified a clean working tree afterward. This is a CLI/daemon behavior change with no rendered UI surface, so evidence is CLI transcripts rather than screenshots.

<details>
<summary>Evidence: Evidence summary: before/after shell-injection guard</summary>

<code>### A. Real-shell end-to-end (Scenario D)&#10;PRE-FIX (guard neutered):&#10;not ok - Scenario D: the escalation digest was typed into the live shell (found &#39;Supervisor escalate&#39; in the pane) — SHELL INJECTION [e2e exit 1]&#10;WITH FIX:&#10;ok - Scenario D: a dead shell showing a bare starship ❯ prompt is refused — the escalation defers, nothing is typed into the shell [e2e exit 0]&#10;&#10;### B. Deterministic inject_msg units&#10;PRE-FIX (neutered):&#10;not ok - inject_msg typed into a dead shell: send_text_submit ran despite agent_alive=dead (starship-glyph hole)&#10;WITH FIX:&#10;ok - defers on empty-composer dead shell (agent_alive=dead)&#10;ok - defers when agent liveness is unknown (fail closed)&#10;ok - injects only when agent alive and composer empty (positive path)&#10;&#10;### C. Full suite: 68 test files, 0 failures, SUITE_EXIT=0</code>

```text
AFK away-mode shell-injection guard — evidence
==============================================

The hazard: the away-mode daemon typed an escalation digest + Enter into a
supervisor pane whenever the composer read EMPTY. A dead login shell showing the
default starship glyph U+276F (❯) is byte-identical to claude's empty composer,
so a pane whose agent exited read 'empty' and the digest was executed as a shell
command. Fix: inject only when fm_backend_agent_alive confirms a live agent.

### A. Real-shell end-to-end (Scenario D) — the actual product surface

PRE-FIX (agent-liveness guard neutered → gate on composer-emptiness only):
not ok - Scenario D: the escalation digest was typed into the live shell (found 'Supervisor escalate' in the pane) — SHELL INJECTION
  e2e exit: 1 (FAIL — digest was typed into and run by the live shell)

WITH FIX (guard restored):
ok - Scenario D: a dead shell showing a bare starship ❯ prompt is refused — the escalation defers, nothing is typed into the shell
  e2e exit: 0

### B. Deterministic inject_msg units

PRE-FIX (neutered):
not ok - inject_msg typed into a dead shell: send_text_submit ran despite agent_alive=dead even though the composer read empty (starship-glyph hole)
not ok - starship-glyph inject_msg subshell failed

WITH FIX:
ok - inject_msg: defers on an empty-composer dead shell (bare starship ❯ prompt, agent_alive=dead), closing the shell-injection hole
ok - inject_msg: defers when agent liveness is unknown (fail closed), even on a confirmed-empty composer
ok - inject_msg: injects only when the agent is confirmed alive and the composer is confirmed empty (positive path)

### C. Full suite
  68 test files, 0 failures, SUITE_EXIT=0
```
</details>
<details>
<summary>Evidence: Full test suite log (68 files, 0 failures)</summary>

```text
tmux 3.7b
== tests/fm-afk-inject-e2e.test.sh ==
ok - Scenario A: partial input defers injection; digest arrives clean after idle
ok - Scenario B: swallowed Enter produces exactly one clean digest
ok - Scenario C: a normal captain status injects exactly one clean single-line sentinel digest
ok - Scenario D: a dead shell showing a bare starship ❯ prompt is refused — the escalation defers, nothing is typed into the shell
all e2e injection tests passed
== tests/fm-afk-inject-herdr-e2e.test.sh ==
ok - real herdr Scenario A: partial input defers injection; digest arrives clean after idle
ok - real herdr Scenario B: swallowed Enter (via the herdr shim) produces exactly one clean digest
ok - real herdr Scenario C: a normal captain status injects exactly one clean single-line sentinel digest
ok - real herdr Scenario D: a persistently pending composer raises the max-defer wedge alarm, preserves the buffer, and never crashes the daemon
all real-herdr afk injection e2e tests passed
== tests/fm-afk-launch.test.sh ==
ok - clear-stale: removes every away-session delivery artifact the one list names
ok - clear-stale: leaves the durable wake-queue intact (no pending work dropped)
ok - refresh: daemon already alive - stale artifacts preserved (current session's buffer kept)
ok - arm canary: a dead supervisor target warns in the foreground, names the wrong target, and still arms
ok - arm canary: an ambiguous verdict names BOTH its causes and confirms the target before accepting it
ok - arm canary: a live supervisor agent arms with no warning
ok - arm canary: a refresh probes the running daemon's target and says to stop away mode, not just re-arm it
ok - arm canary: a live daemon injecting into a vanished pane is reported loudly, not swallowed
ok - arm canary: an unresolvable pane with no live daemon stays quiet
ok - stop-ordering: daemon SIGTERM'd while .afk still present (flush is not a no-op)
ok - stop-ordering: .afk cleared last
ok - stop-ordering: daemon-terminal record removed
ok - stop identity: stale lock cannot signal an unrelated live process
ok - failed start: away flag and delivery artifacts roll back
ok - concurrent start: one serialized daemon terminal remains tracked
ok - launcher lock: incomplete publication receives initialization grace
ok - launcher signal: TERM exits and releases the lifecycle lock
fm-afk-launch: daemon launched in non-visible herdr workspace ws-partial (pane lab:pane-exact), supervising lab:captain
ok - herdr create: malformed response recovers durable exact ownership
fm-afk-launch: herdr create failed after returning exact ids; closing lab:pane-exact
fm-afk-launch: recorded terminal teardown is unconfirmed; preserving exact id
ok - herdr create error: unconfirmed exact id is persisted for reconciliation
fm-afk-launch: failed to run daemon in herdr pane lab:pane-exact; closing it
fm-afk-launch: recorded terminal teardown is unconfirmed; preserving exact id
ok - herdr run failure: unconfirmed exact id remains reconcilable
fm-afk-launch: failed to persist daemon terminal record; closing tmux:exact-session
ok - record failure: newly created terminal is closed by exact id
fm-afk-launch: daemon did not become ready; closing tmux:exact-session
ok - readiness failure: exact terminal and durable record roll back
fm-afk-launch: daemon did not become ready; closing tmux:exact-session
fm-afk-launch: recorded terminal teardown is unconfirmed; preserving exact id
ok - readiness failure: unconfirmed terminal retains its reconciliation id
ok - tmux absence: clean missing differs from transport probe failure
ok - native lifecycle: launcher owns state with no terminal
ok - native lifecycle: uniform stop clears state without closing a terminal
ok - native entry: launcher-prepared lifecycle state is not rewritten
fm-afk-launch: reconciling leaked daemon terminal tmux:exact-session
fm-afk-launch: recorded terminal teardown is unconfirmed; preserving exact id
ok - teardown failure: exact terminal record is preserved
ok - record publication: failed atomic rename preserves the complete prior record
fm-afk-launch: daemon terminal record is malformed; refusing to act on it
ok - record read: malformed record fails closed without acting on a partial id
fm-afk-launch: daemon terminal record is malformed; refusing to act on it
fm-afk-launch: malformed daemon terminal record; refusing to stop away mode
ok - stop: malformed terminal record preserves away state and fails closed
fm-afk-launch: failed to create detached tmux daemon session 'fm-afk-daemon-2134678404-23528-23822-1784063985'
ok - tmux launch: planned exact target is recorded before creation and removed on failure
fm-afk-launch: failed to create detached tmux daemon session 'fm-afk-daemon-3400271325-23564-7409-1784063985'
ok - tmux launch: unique names eliminate collision teardown
ok - stop validation: malformed record causes no daemon or state side effects
ok - launcher lock: incomplete metadata fails acquisition and releases lock
fm-afk-launch: failed to clear away-mode flag
fm-afk-launch: away mode stopped; terminal teardown remains recorded for retry
ok - stop state: away-flag removal failure is surfaced
fm-afk-launch: away-mode daemon did not exit after SIGTERM; preserving lifecycle state
ok - stop liveness: captured live daemon preserves lifecycle state after lock release
fm-afk-launch: daemon terminal record is malformed; refusing to act on it
fm-afk-launch: daemon terminal record is malformed; refusing to act on it
ok - refresh record: malformed terminal identity fails closed
fm-afk-launch: failed to clear stale away-mode artifacts
ok - clear failure: native entry aborts and restores prior state
fm-afk-launch: reconciling leaked daemon terminal tmux:exact-session
fm-afk-launch: terminal close command failed, but exact absence was confirmed
ok - confirmed absence: cleanup succeeds and removes the stale record
fm-afk-launch: rollback restoration incomplete; backup retained at /var/folders/4n/_bv20k5d26dgrq1t5y4l69940000gn/T//fm-afk-restore-fail.VPsgV8/state/.afk-launch-backup.wlwtsZ
ok - rollback restore: incomplete restoration retains its recovery backup
ok - rollback: every away-session artifact a failed arm can write is cleared
ok - flag failure: lifecycle aborts without active state
ok - herdr e2e: captain tab pane count unchanged after start (no split)
ok - herdr e2e: daemon launched in a separate non-visible workspace
ok - herdr e2e: daemon pane is NOT in the captain's tab
ok - herdr e2e: daemon terminal scoped to the lab session
ok - herdr e2e: captain tab pane count restored after stop
ok - herdr e2e: daemon workspace removed by exact id on stop
ok - herdr e2e: record + .afk cleared on stop
ok - tmux e2e: captain window pane count unchanged after start (no split-window)
ok - tmux e2e: daemon launched in a separate detached session
ok - tmux e2e: captain window pane count unchanged after stop
ok - tmux e2e: daemon session killed by exact id on stop
ok - tmux e2e: record + .afk cleared on stop
== tests/fm-arm-pretool-check.test.sh ==
ok - matrix A01: allow through all five entry forms
ok - matrix A02: allow through all five entry forms
ok - matrix A03: allow through all five entry forms
ok - matrix A04: allow through all five entry forms
ok - matrix A05: allow through all five entry forms
ok - matrix A06: allow through all five entry forms
ok - matrix A07: allow through all five entry forms
ok - matrix A08: allow through all five entry forms
ok - matrix A09: allow through all five entry forms
ok - matrix A10: allow through all five entry forms
ok - matrix A11: allow through all five entry forms
ok - matrix A12: allow through all five entry forms
ok - matrix A13: allow through all five entry forms
ok - matrix A14: allow through all five entry forms
ok - matrix A15: allow through all five entry forms
ok - matrix A16: allow through all five entry forms
ok - matrix A17: allow through all five entry forms
ok - matrix R01: allow through all five entry forms
ok - matrix R02: allow through all five entry forms
ok - matrix R03: allow through all five entry forms
ok - matrix R04: allow through all five entry forms
ok - matrix R05: allow through all five entry forms
ok - matrix R06: al

... [103640 bytes truncated] ...

ff plain stale identities for daemon-owned pause triage
== tests/fm-watcher-lock.test.sh ==
ok - guard banner leads when down with pending wakes (re-arm-after-drain) and stays silent when fresh
ok - watch restart reports a healthy peer without attaching to it
ok - arm attaches to a live fresh watcher and exits only when that cycle ends
ok - arm attaches to a peer watcher after child stands down and exits when peer dies
watcher: lock held by live pid 5522 but heartbeat is stale for 837350992s (>300s); inspect or stop that watcher before re-arming.
== tests/fm-x-mode.test.sh ==
ok - fm-x-poll is a hard no-op without a token (inert default)
ok - fm-x-poll treats an explicitly empty env token as configured
ok - fm-x-poll stays silent on HTTP 204 (the common case)
ok - fm-x-poll lets an explicitly empty relay env override .env
ok - fm-x-poll surfaces auth/config errors once and clears on recovery
ok - fm-x-poll stashes the question and prints the compact marker
ok - fm-x-poll preserves in_reply_to conversation context in the inbox
ok - fm-x-poll reports inbox commit failures without emitting a mention wake
ok - fm-x-poll requires a non-empty question before waking
ok - fm-x-poll rejects an unsafe request_id (path-traversal guard)
ok - fm-x-reply posts a request-bound answer and echoes only the request_id
ok - fm-x-reply accepts the reply via --text-file and stdin (safe, unexpanded)
ok - fm-x-reply exits non-zero on a non-2xx relay response
ok - fm-x-reply cleans up auth header temp files on interrupted posts
ok - fm-x-reply rejects missing arguments with a usage error
ok - fm-x-reply --help makes image support discoverable
ok - fm-x-reply rejects whitespace-only reply text
ok - fm-x-reply dry-run records the would-be reply and never posts
ok - fm-x-reply dry-run works without a token
ok - fm-x-reply honors FMX_DRY_RUN from .env
ok - fm-x-reply lets an explicitly empty dry-run env override .env
ok - fm-x-reply dry-run fails when it cannot record the preview
ok - fmx_split_thread: word-boundary, fence-aware, within-limit, numbered, lossless, capped
ok - fm-x-reply keeps a concise reply as a single unnumbered tweet
ok - fm-x-reply auto-splits a long reply into a numbered thread (texts[])
ok - fm-x-reply uses the Discord inbox platform budget instead of the X tweet budget
ok - fm-x-reply keeps numeric X requests on the X tweet budget
ok - fm-x-reply prefers an explicit relay-provided reply limit
ok - fm-x-reply clamps a below-floor max to 50 characters
ok - fm-x-reply posts a thread payload (texts[]) to the relay
ok - fm-x-reply --image posts an image object on answer
ok - fm-x-reply streams large image payloads outside curl argv
ok - fm-x-reply dry-run records compact image metadata for threaded replies
ok - fm-x-reply cleans image and payload temp files
ok - fm-x-reply --image rejects missing and unsupported image paths clearly
ok - fm-x-reply --followup posts to /connector/followup with the same request-bound body
ok - fm-x-reply maps a followup_unavailable follow-up 409 to exit 9
ok - fm-x-reply maps every follow-up 409 to exit 9 even without the marker
ok - fm-x-reply treats answer-endpoint 409 as a generic failure
ok - fm-x-reply --followup --image posts an image object
ok - fm-x-reply --followup is accepted in any position and leaves the answer path default
ok - fm-x-reply --followup dry-run marks the endpoint without changing the answer path
ok - fm-x-reply --followup auto-splits a long follow-up into a marked thread
ok - fm-x-reply followup dry-run keeps endpoint marker and compact image metadata
ok - fm-x-poll records the durable per-request reply context from the relay payload
ok - context registry retention is bounded to the seven-day follow-up window
ok - context registry rewrites preserve the first-seen timestamp
ok - context retention starts only when a live initial answer succeeds
ok - a delayed Discord follow-up stays one message after inbox cleanup via the durable registry
ok - an X follow-up over 280 still splits correctly after inbox cleanup
ok - every unresolved follow-up is refused before posting
ok - a partial registry platform combines with the relay's authoritative budget
ok - concurrent requests each recover their own platform/budget with no cross-overwrite
ok - fm-x-dismiss clears the durable per-request context (a dismissed mention gets no follow-up)
ok - fm-x-dismiss posts a request-bound dismiss and echoes only the request_id
ok - fm-x-dismiss dry-run records the would-be body and never posts
ok - fm-x-dismiss dry-run works without a token
ok - fm-x-dismiss exits non-zero on a non-2xx relay response
ok - fm-x-dismiss exits non-zero on a transport failure
ok - fm-x-dismiss rejects an unsafe request_id (path-traversal guard)
ok - fm-x-dismiss rejects missing or extra arguments with a usage error
ok - fm-x-link records and refreshes the X-request link without disturbing meta
ok - fm-x-link records Discord platform context so follow-ups keep the Discord budget
ok - fm-x-link resolves the platform by request_id so a post-cleanup link keeps the Discord budget
ok - fm-x-link warns loudly and the follow-up is held (not wrongly split) when the platform is unknown
ok - fm-x-link paired carry flags preserve a prior task's follow-up binding onto a successor
ok - fm-x-link recovery relink preserves Discord platform context after inbox drain
ok - fm-x-link rejects malformed or unpaired carry flags
ok - meta rewrites are independent of TMPDIR
ok - fm-x-link rejects unsafe ids, missing meta, and missing arguments
ok - fm-x-followup --check reports postable / not-linked correctly
ok - fm-x-followup --check prunes a link past the 7-day window
ok - fm-x-followup --check prunes a link that already reached the follow-up cap
ok - fm-x-followup posts a follow-up, increments the counter, and keeps the link under the cap
ok - fm-x-followup --final clears the link after one post regardless of the remaining count
ok - fm-x-followup clears the link once the third follow-up reaches the cap
ok - fm-x-followup --image forwards the attachment through fm-x-reply --followup
ok - fm-x-followup keeps the link and counter when the post fails
ok - fm-x-followup tombstones the link when a post-success counter write fails
ok - fm-x-followup treats a relay cap/window rejection as an already-exhausted link, not a retry
ok - fm-x-followup skips silently and clears the link past the 7-day window
ok - fm-x-followup is a no-op for a task with no X link
ok - fm-x-followup dry-run records the follow-up and increments the counter, keeping the link
ok - fm-x-followup dry-run --final clears the link just as a live post would
ok - fm-x-followup rejects malformed invocations
ok - bootstrap activates X mode from an .env token, idempotently
ok - bootstrap reports missing X-mode dependencies before arming
ok - bootstrap does not report X mode on when activation artifacts cannot be written
ok - bootstrap is inert without a non-empty .env token (non-X users unaffected)
ok - bootstrap cleans up X artifacts on opt-out and is silent once off
ok - bootstrap reports failed X artifact cleanup on opt-out
SUITE_EXIT=0
```
</details>
<details>
<summary>Evidence: Pre-fix (neutered) e2e — SHELL INJECTION failure</summary>

```text
ok - Scenario A: partial input defers injection; digest arrives clean after idle
ok - Scenario B: swallowed Enter produces exactly one clean digest
ok - Scenario C: a normal captain status injects exactly one clean single-line sentinel digest
not ok - Scenario D: the escalation digest was typed into the live shell (found 'Supervisor escalate' in the pane) — SHELL INJECTION
```
</details>
<details>
<summary>Evidence: Fixed e2e — Scenario D refuses injection</summary>

```text
ok - Scenario A: partial input defers injection; digest arrives clean after idle
ok - Scenario B: swallowed Enter produces exactly one clean digest
ok - Scenario C: a normal captain status injects exactly one clean single-line sentinel digest
ok - Scenario D: a dead shell showing a bare starship ❯ prompt is refused — the escalation defers, nothing is typed into the shell
all e2e injection tests passed
```
</details>
<details>
<summary>Evidence: Pre-fix (neutered) inject_msg units — failures</summary>

```text
ok - pane_input_pending: bare prompts are not pending (idle)
ok - max-defer on an empty stuck pane types once, alarms with the swallowed-Enter cause, and preserves the buffer
ok - max-defer on a pending composer alarms without typing, naming the composer as the cause
/Users/zacharylandes/.no-mistakes/worktrees/01004ff68993/01KXFBCYZTCDTN4KY16GR41M1S/bin/fm-supervise-daemon.sh: line 752: 12893 Terminated: 15          sh -c 'sleep 30 & printf "%s" "$!" > "$1"; wait' sh "$child_file"
ok - pane_is_busy: herdr falls back to the shared regex-over-capture reader when native busy_state is unknown
ok - pane_is_busy: herdr corroborates native idle with the shared regex-over-capture reader
ok - pane_is_busy: omitted backend arg defaults to tmux (pre-existing callers unaffected)
not ok - inject_msg typed into a dead shell: send_text_submit ran despite agent_alive=dead even though the composer read empty (starship-glyph hole)
not ok - starship-glyph inject_msg subshell failed
```
</details>
- Evidence: Fixed inject_msg units — pass (local file: <code>/var/folders/4n/_bv20k5d26dgrq1t5y4l69940000gn/T/no-mistakes-evidence/01KXFBCYZTCDTN4KY16GR41M1S/fixed-fm-daemon-units.log</code>)
- Outcome: 🔧 1 issue found → auto-fixed ✅ across 2 runs (56m48s)

## Pipeline

_... (7 earlier update rounds omitted to keep the PR body within GitHub's 65536-char limit; full history is in the run log.)_

<details>

🔧 Fix: alarm the away captain when the supervisor pane vanishes
- ⚠️ `bin/fm-supervise-daemon.sh:958` - pane_gone_wedge_alarm calls inject_wedge_alarm with no buffer gate, so the alarm fires even when nothing is buffered - and it inherits inject_wedge_alarm&#39;s escalation-centric wording. Housekeeping&#39;s max-defer alarm (line 1004) gates on `[ -s &#34;$state/.subsuper-escalations&#34; ]`; this one does not. Worse, the buffer cannot grow while the pane is gone: the main-loop backoff `continue`s past the wake handling that calls escalate_add, so a buffer that was empty when the pane vanished stays empty for the entire away window. In that case the captain&#39;s Notification Center alert reads &#34;away-mode escalations WEDGED 300s undelivered (pane-gone)&#34; (line 922), the durable marker&#39;s header repeats it (line 902) with an empty &#34;Buffered items:&#34; list beneath, and the ERROR log says &#34;away-mode escalation undelivered 300s&#34; - three statements about undelivered escalations that do not exist, repeated once per max-defer window for as long as the daemon lives. The number is also the pane&#39;s absence, not any escalation&#39;s age. Concrete case: crew work finishes and flushes, the buffer empties, then firstmate&#39;s pane dies; the captain wakes to a stack of alerts telling them escalations are stuck, goes looking for them, and finds none - while the actual message (&#34;away mode is dead, restart it&#34;) is only implied by the `(pane-gone)` tag. Firing the alarm here is right and is what was asked for; the fix is to branch the summary, marker header, and ERROR line on cause=pane-gone (say the supervisor pane is gone and away mode can deliver nothing, and state the buffered count) rather than reusing the undelivered-escalation framing.
- ℹ️ `bin/fm-supervise-daemon.sh:1569` - When the target resolves again the main loop only resets PANE_GONE_SINCE; it never clears state/.subsuper-inject-wedged. The existing recovery path (line 1013) removes the marker only when housekeeping&#39;s max-defer escape flush succeeds, which requires the buffer to still be stuck past max_defer - so a pane-gone alarm raised on an empty or promptly-flushed buffer leaves the marker behind permanently. This is reachable whenever the target is a name rather than a unique pane id: FM_SUPERVISOR_TARGET_DEFAULT is `firstmate:0`, and any session:window target can be killed and recreated, as can a herdr session:pane. Concrete case: the supervisor window is closed and recreated during a 6-minute gap; the daemon alarms, then delivers normally for the rest of the night; on return, firstmate reads the stale marker as part of the afk-exit catch-up (AGENTS.md section 8) and reports a wedge that resolved hours earlier, naming a pane that is now healthy. Clearing the marker on the recovery edge when the last recorded cause was pane-gone would close it.

🔧 Fix: name supervision-down cause and retire pane-gone alarm marker
- ⚠️ `bin/fm-supervise-daemon.sh:999` - The pane-gone alarm re-fires once per max-defer window for the entire life of the daemon, on a condition that can never self-heal. inject_wedge_alarm rewrites the marker on every alarm (line 942), so its marker-age throttle (line 908) re-opens after max_defer, and WEDGE_ALARM_LAST_EPOCH (line 931) likewise re-permits wedge_alarm_notify after max_defer. The comment at line 976 calls this &#34;one alarm per window&#34;, which is accurate — but the window repeats indefinitely. Unlike the pre-existing buffer wedge (a busy composer clears; a flush succeeds and clears the marker), a vanished pane never comes back: the supervisor target is a unique pane id, not a name. fm-afk-launch.sh captures $TMUX_PANE (e.g. %5) and passes it as FM_SUPERVISOR_TARGET, and the daemon&#39;s own discovery does the same, so FM_SUPERVISOR_TARGET_DEFAULT (&#39;firstmate:0&#39;, the recreatable name the pane_gone_recovered comment cites) is only the unconfigured fallback. Concrete case: away mode armed at 23:00; the captain&#39;s terminal dies at 23:30 (the ordinary reason firstmate restarts elsewhere, and exactly the scenario this alarm was added for); the detached daemon alarms at 23:35 and every 300s until morning — ~96 Notification Center banners with the Basso sound, and ~96 pushes through a configured `command:` channel, which is the documented way to reach a captain who is away from the machine (docs/wedge-alarm.md). The captain cannot act on any of them remotely, and alert fatigue on a safety channel is the failure mode this whole alarm exists to prevent. Your instruction for this round was &#34;one alert, respecting the existing alarm cadence, not every poll&#34;. Suggest firing the pane-gone alert once per absence EPISODE (PANE_GONE_SINCE already identifies the episode, and pane_gone_recovered already resets it), or backing it off geometrically — the durable marker and the ERROR log remain the record either way, so nothing is lost by not re-pushing.
- ℹ️ `bin/fm-supervise-daemon.sh:1016` - pane_gone_recovered decides whether to retire state/.subsuper-inject-wedged by reading the in-memory INJECT_LAST_DEFER_CAUSE, not by what actually wrote the marker — but that global is set to &#39;pane-gone&#39; from two places, and only one of them writes a pane-gone marker. inject_msg&#39;s own target-exists check (line 1247) records cause=pane-gone whenever the pane dies between the main loop&#39;s check and housekeeping&#39;s flush. Concrete failure: a buffer is genuinely wedged behind a busy composer, so housekeeping raised an alarm and the on-disk marker describes 900s of undelivered escalations; the pane then blinks out for a few seconds; housekeeping&#39;s escalate_flush hits inject_msg, whose target-exists check fails and sets cause=pane-gone, but the max-defer escape&#39;s own age gate (line 1074) does not re-fire, so the composer-wedge marker is left untouched; the main loop then sees the pane missing (PANE_GONE_SINCE set) and, on the next poll, sees it back — and pane_gone_recovered deletes the composer-wedge marker, whose contract (line 1010) explicitly says such a marker &#34;still describes a real undelivered buffer&#34; and must be left alone. The buffer itself survives and housekeeping re-raises within one tick, so this self-corrects, but a captain returning inside that window loses the wedge report from the afk-exit catch-up. Gate the retirement on the marker&#39;s own recorded cause (it is written into the header/detail at lines 918-920) rather than on the process-global, or set a separate flag only in pane_gone_wedge_alarm&#39;s alarm branch.

🔧 Fix: latch pane-gone alarm to one alert per absence episode
2 warnings still open:
- ⚠️ `bin/fm-supervise-daemon.sh:1033` - The fire-once latch is set BEFORE the alert is attempted, and the alert has no success signal, so one transient channel failure silences the worst state for the whole away window. PANE_GONE_ALARMED=1 (line 1033) latches, then inject_wedge_alarm calls wedge_alarm_notify (line 981), which is best-effort by contract: every channel failure only logs and it always returns 0 (&#34;Always returns 0: a channel failure can never abort inject_wedge_alarm&#34;), and each channel runs under wedge_alarm_run_bounded&#39;s 10s timeout. Nothing re-fires afterwards - line 1025 returns early for the rest of the episode, and the pane never comes back on its own. Every OTHER wedge cause re-alarms once per max-defer window, so a failed push is effectively retried; pane-gone now gets exactly one attempt. Concrete failure: away mode armed at 23:00 with a `command:` ntfy/Slack channel (the documented way to reach a captain away from the machine, docs/wedge-alarm.md); the captain&#39;s terminal dies at 23:30; at 23:35 the single push hits a network blip (or osascript times out under load) and returns non-zero; wedge_alarm_via_command logs &#34;command channel exited N&#34; into a log nobody reads, PANE_GONE_ALARMED is already 1, and the captain gets no alert at all for a daemon that is now typing at nothing - the exact away-window silence this alarm was added to end. The marker and ERROR log survive, but both are synchronous surfaces the captain only sees once they are back, which is what round 5 established is not good enough. Suggest latching on a DELIVERED alert rather than an attempted one (have wedge_alarm_notify report whether any channel succeeded, leaving the latch clear when none did), or allowing a small bounded number of re-pushes per episode - either satisfies &#34;one alert per episode, not one per window forever&#34; without making a single dropped packet permanent.
- ⚠️ `bin/fm-supervise-daemon.sh:924` - The `latched` flag skips the marker-age throttle at line 924 but NOT the unconditional marker write at lines 959-965, so a pane-gone alarm overwrites whatever marker is already on disk - including a composer-wedge marker housekeeping wrote moments earlier - replacing its `cause:` line with `pane-gone`. pane_gone_recovered&#39;s on-disk cause check (line 1061) then matches and deletes it (line 1062). That is precisely the deletion the comment at lines 953-958 promises is impossible (&#34;a marker describing a real undelivered buffer is never mistaken for a pane-gone alarm and deleted with it&#34;): keying on the marker instead of the global closes the mid-tick-blink path, but not the path where the pane-gone alarm actually fires. Concrete sequence, all reachable: escalations wedge behind a busy composer past max_defer, so housekeeping (line 1104) writes a marker with `cause: composer-not-empty` describing 900s undelivered; the captain&#39;s terminal then dies; after a full max-defer window of absence pane_gone_wedge_alarm fires latched and rewrites that marker as `cause: pane-gone`; the target is a session:window name (FM_SUPERVISOR_TARGET_DEFAULT is `firstmate:0`, and pane_gone_recovered&#39;s own comment cites the killed-and-recreated window), so it resolves again and the composer-wedge marker is removed. The new test test_pane_gone_recovery_keeps_a_real_buffer_wedge_marker only passes because it hand-sets PANE_GONE_ALARMED=1 without ever calling pane_gone_wedge_alarm, so it pins a state the production path cannot reach. Impact is bounded - the buffer survives and housekeeping re-raises on the next tick - but for that window firstmate&#39;s afk-exit catch-up has no wedge report, and the re-raise is silent (the latched alarm set WEDGE_ALARM_LAST_EPOCH, so line 947 suppresses both the ERROR log and the active alert for up to max_defer). Suggest not rewriting an existing marker whose recorded cause is something else (leave it, or give the pane-gone alarm its own marker file) so recovery can only ever retire what it wrote.

🔧 Fix: give the pane-gone alarm its own marker file
- ⚠️ `bin/fm-supervise-daemon.sh:941` - inject_wedge_alarm selects its marker file and its throttle from $kind (line 928) but selects the captain-facing summary, the Notification Center banner title (WEDGE_ALARM_TITLE), and the ERROR log from $cause (line 941). Those two disagree on a reachable path, and the result is a false &#39;away mode is dead&#39; alert to an away captain. Concrete failure: the main loop&#39;s fm_backend_target_exists check passes, the watcher runs for a poll interval, and the supervisor pane dies in that window; housekeeping&#39;s max-defer escape (line 1121) then calls escalate_flush -&gt; inject_msg, whose own target-exists check fails and records cause=pane-gone (line 1247 via _inject_defer). escalate_flush returns non-zero, so housekeeping calls inject_wedge_alarm &#34;$state&#34; &#34;$oldest&#34; with the DEFAULT kind=buffer. Inside, the cause=pane-gone branch fires: WEDGE_ALARM_TITLE becomes &#39;firstmate: away-mode supervision DOWN&#39;, the active alert reads &#39;away-mode supervision DOWN Ns (pane-gone): supervisor pane gone, nothing can be delivered&#39;, and that text is written into .subsuper-inject-wedged (the BUFFER marker). This entirely bypasses the debounce pane_gone_wedge_alarm was built around - it requires a full max-defer window of CONTINUOUS absence before declaring supervision down, precisely because &#39;a pane that blinks out for one poll is a backend hiccup&#39; (line 1003) - so one flaky tmux/herdr existence probe at flush time now produces the loudest verdict in the design. Per the new .agents/skills/afk/SKILL.md guidance (&#39;The one verdict that is never an accepted degradation is a vanished pane... report it and fix it rather than continuing away mode&#39;), firstmate would tear down a healthy away session on that alert. This is the same misattribution class the wedge-alarm and canary work spent four rounds eliminating. Fix: branch the summary/header/detail/err/title on $kind instead of $cause (line 941 -&gt; `if [ &#34;$kind&#34; = pane-gone ]`). Nothing is lost: with kind=buffer and cause=pane-gone the buffer wording still names the tag - &#39;away-mode escalations WEDGED Ns undelivered (pane-gone)&#39; - which is accurate, because the buffer IS stuck and IS still being retried, while the debounced, latched supervision-DOWN declaration stays where it belongs.
- ℹ️ `bin/fm-afk-launch.sh:353` - FM_AFK_LAUNCH_ARTIFACTS (line 64) was introduced this commit with the comment &#39;One list, so a new artifact cannot be added to the daemon and forgotten by the rollback that has to put it back&#39; - but fm_afk_launch_restore_backup&#39;s `rm -f` at lines 353-358 still hardcodes the same four artifact names immediately above the loop that iterates the array. The drift the array exists to prevent is therefore only half-closed: a future fifth artifact added to FM_AFK_LAUNCH_ARTIFACTS but not to that rm -f would survive a rolled-back arm (the loop only restores what was in the backup, so an artifact the failed arm created fresh is never cleared). Iterate the array in the rm -f too, keeping .afk as the one explicit extra. (bin/fm-afk-start.sh&#39;s fm_afk_clear_stale_artifacts is a third hardcoded copy of the same list, in a different file - worth noting, but the launcher&#39;s internal duplication is the one that contradicts its own comment.)

🔧 Fix: key wedge alarm wording to alarm kind, not last cause
- ⚠️ `docs/wedge-alarm.md:12` - The alarm-semantics section (lines 12-16) still attributes the supervision-DOWN wording, the fire-once latch, and the empty-buffer firing to the CAUSE `pane-gone` (&#34;`pane-gone` is the one cause whose captain-facing wording is not about the buffer&#34;, &#34;It also fires on an EMPTY buffer&#34;, &#34;Every channel therefore says supervision is down&#34;). This commit changed all three to key on the alarm KIND instead. On the reachable pane-died-mid-flush path - the main loop&#39;s fm_backend_target_exists check passes, the pane dies during housekeeping&#39;s max-defer escalate_flush, and inject_msg records cause=pane-gone (bin/fm-supervise-daemon.sh:1298) while the alarm housekeeping raises is kind=buffer - the channels say &#34;away-mode escalations WEDGED 900s undelivered (pane-gone)&#34;, the alarm IS throttled once per max-defer window, and it IS gated on a non-empty buffer. That is the exact opposite of what these lines promise, and it is the behavior the new test test_buffer_wedge_alarm_never_declares_supervision_down pins. Line 109 of the same file, added in this commit, states the correct rule (&#34;Which alarm fired is what the captain is told, never the last recorded cause&#34;), so the file now contradicts itself: a maintainer reading top-down would reintroduce the false &#39;away mode is dead&#39; alert this commit fixed. Line 6 carries a smaller version of the slip - it claims `pane-gone` comes only from the daemon&#39;s main loop rather than an inject attempt, but inject_msg&#39;s own target-exists check records it too. Rewrite lines 6 and 12-16 so the supervision-DOWN wording, the latch, and the empty-buffer firing are described as properties of the pane-gone ALARM (the debounced, latched one raised from the backoff path), with `pane-gone` named only as a cause tag that can also appear on a buffer wedge.
- ℹ️ `bin/fm-afk-start.sh:63` - FM_AFK_LAUNCH_ARTIFACTS (bin/fm-afk-launch.sh:64) was introduced this commit with the comment &#34;One list, so a new artifact cannot be added to the daemon and forgotten by the rollback that has to put it back&#34;, and the rollback rm now iterates it. But fm_afk_clear_stale_artifacts - the fresh-arm clear that drops the PRIOR away session&#39;s delivery artifacts - still hardcodes the same four names, and it lives in bin/fm-afk-start.sh, which the launcher SOURCES, so it cannot reference the launcher&#39;s array. The two agree today (both gained .subsuper-pane-gone in this PR), but a fifth artifact added to FM_AFK_LAUNCH_ARTIFACTS alone would be backed up and rolled back correctly while surviving a fresh arm - resurfacing a prior session&#39;s buffered escalation or wedge marker under a new away session, which is the failure fm_afk_clear_stale_artifacts exists to prevent. Defining the list once in bin/fm-afk-start.sh (where the clear lives) and having bin/fm-afk-launch.sh iterate that same array would close the drift and shrink the footprint rather than grow it. Flagged as an aside in round 9 and deliberately scoped out of that fix; noting the residual, not re-litigating.

🔧 Fix: key wedge-alarm doc to alarm kind and unify afk artifact list
- ⚠️ `bin/fm-supervise-daemon.sh:200` - The WEDGE_ALARM_TITLE declaration comment still states the captain-facing banner title &#34;follows the CAUSE&#34; (&#34;so it follows the CAUSE: a vanished supervisor pane means away-mode supervision is down&#34;), but commit 1496b85 changed inject_wedge_alarm to select the summary, WEDGE_ALARM_TITLE, and the ERROR log from $kind instead — the same discriminator that picks the marker file and the throttle — specifically so a cause=pane-gone recorded by inject_msg during housekeeping&#39;s max-defer flush cannot render the supervision-DOWN verdict from what is actually a buffer wedge. docs/wedge-alarm.md was rewritten for this (lines 201-209: &#34;WHICH one fired - its kind - is what every captain-facing surface says&#34;) and tests/fm-daemon.test.sh&#39;s test_buffer_wedge_alarm_never_declares_supervision_down pins it, so the code comment is now the only place in the repo asserting the superseded rule. inject_wedge_alarm&#39;s own header (line ~892) carries a milder version: it opens with &#34;the ERROR log, the durable marker, and the active alert all report the recorded cause&#34; before the next paragraph corrects it to the kind. Concrete failure: a maintainer reads the declaration comment, re-keys the branch at line 946 on $cause, and the reachable pane-died-mid-flush path (main-loop target-exists passes, pane dies during escalate_flush -&gt; inject_msg records pane-gone at line 1298, housekeeping raises kind=buffer at line 1137) again titles an away captain&#39;s Notification Center banner &#34;firstmate: away-mode supervision DOWN&#34; for a buffer that is still stuck behind a pane the daemon keeps retrying — which per .agents/skills/afk/SKILL.md means tearing down a healthy away session. Rewrite lines 198-204 (and the opening sentence of the inject_wedge_alarm header) to say the title follows the alarm KIND, with the cause carried only as a diagnostic tag.
- ℹ️ `bin/fm-bootstrap.sh:609` - Noting the boundary of the new session-start diagnostic, not requesting a change. afk_injection_check gates on state/.afk, then fm_afk_canary_resolve falls back to discover_supervisor_target whenever no LIVE daemon owns the lock — so when away mode is armed but the daemon has died, bootstrap probes its own pane, finds firstmate obviously alive, and prints nothing. That is the right answer for what the canary measures (can the resolved supervisor be injected into), and AGENTS.md section 5 step 8 already covers the dead-daemon case by having /afk ensure the daemon is running on recovery, so nothing is actually missed. But docs/herdr-backend.md&#39;s &#34;a cold or restarted session that re-enters afk from the flag alone still learns injection is off&#34; reads broader than the check guarantees: the one away-mode failure a restarted session cannot learn from this line is the daemon simply not being there.

🔧 Fix: key wedge-alarm title comment to alarm kind, not cause
2 infos still open:
- ℹ️ `bin/fm-afk-start.sh:79` - fm_afk_clear_stale_artifacts changed from a single best-effort `rm -f` (which returns 0 essentially always) to a loop that returns 1 if ANY rm fails. In bin/fm-afk-launch.sh both call sites already test the return value and log &#34;failed to clear stale away-mode artifacts&#34;, so there the new failure signal is an improvement. But fm_afk_start_main calls it bare inside an `if` body (bin/fm-afk-start.sh:159) and that file runs under `set -eu` with no `set +e`, so a non-zero return is not a caught error — errexit fires and the daemon start aborts before `exec &#34;$FM_AFK_DAEMON&#34;`, with no message at all (the rm&#39;s stderr goes to /dev/null and errexit is silent). Failure scenario: a direct, non-prepared `bin/fm-afk-start.sh` on a state dir holding an undeletable artifact (root-owned leftover from a prior run under a different user, or a read-only mount) previously logged nothing and started away mode anyway; now it exits 1 with no output and away mode silently does not arm. Reachability is narrow — the /afk skill always goes through the launcher with FM_AFK_STATE_PREPARED=1, so only a manual/legacy direct invocation takes this branch — but the silence is the problem, not the abort. Either echo the failure before returning non-zero, or guard the call site the way the launcher does.
- ℹ️ `bin/fm-supervise-daemon.sh:1081` - When the supervisor target resolves again, pane_gone_recovered deletes state/.subsuper-pane-gone and logs the recovery only to the daemon log. But the captain has, by then, already received the pane-gone active alert — a Notification Center banner titled &#34;firstmate: away-mode supervision DOWN&#34;, and a push through any configured `command:` channel, telling them to stop away mode and arm it again. Nothing retracts it, and the marker (the one surface the afk-exit catch-up actually reads) is gone. Per .agents/skills/afk/SKILL.md:144, an absent .subsuper-pane-gone marker means &#34;no live absence&#34;, so firstmate&#39;s catch-up reports all-clear for an alarm the captain demonstrably received. Failure scenario: the supervisor window is a session:window name (the FM_SUPERVISOR_TARGET_DEFAULT `firstmate:0` fallback, or an explicit named target — the exact case pane_gone_recovered was written for), it is killed and recreated during a &gt;300s gap overnight; the captain gets a 2am &#34;supervision DOWN&#34; push, comes back to a clean catch-up, and neither they nor firstmate can explain the alert without grepping state/.supervise-daemon.log, which no catch-up instruction points at. A one-line resolved note left behind on the recovery edge, or having the /afk catch-up read the daemon log&#39;s ERROR/recovery pair, would close it. Flagging for the record rather than requesting a change: this is the same best-effort-alert reach you already scoped out (docs/wedge-alarm.md &#34;Known follow-ups&#34;), and it belongs with the bounded-re-fire item there.

</details>

<details>

🔧 Fix: provide mkdir to x-mode fakebin, silence node ExperimentalWarning
- <code>`bash tests/*.test.sh` full suite: 68 files, 0 failures, exit 0</code>
- <code>`bash tests/fm-afk-inject-e2e.test.sh` — real-shell Scenario D (genuine login shell with a bare starship ❯ prompt) passes: escalation defers, nothing typed into the shell</code>
- <code>`bash tests/fm-daemon.test.sh` — the three safety-critical inject_msg units pass: defers on empty-composer dead shell (agent_alive=dead), defers on unknown liveness (fail closed), injects on alive+empty (positive path)</code>
- <code>Pre-fix regression proof: neutered the liveness gate to `liveness=alive` (simulating gate-on-composer-only) and re-ran both files — Scenario D failed with &#39;SHELL INJECTION&#39; (digest typed into and run by the live shell) and the starship-glyph unit failed with &#39;send_text_submit ran despite agent_alive=dead&#39;; restored the file (git checkout) and confirmed clean tree + tests green again</code>

</details>

<details>

- ℹ️ `docs/scripts.md:47` - docs/scripts.md (the bin/ toolbelt inventory) is missing five scripts that pre-date this change: fm-lint.sh, fm-install-shellcheck.sh (added in #481), fm-cd-pretool-check.sh, fm-cd-command-policy.mjs (added in #483), and fm-transition-lib.sh (added in #472). I added the row for fm-afk-canary-lib.sh, which this change introduced, but deliberately left the pre-existing five alone: backfilling unrelated scripts would expand a focused away-mode safety fix&#39;s documentation diff with work from three other PRs. Needs a human call on whether to fix them here or in a separate docs cleanup.

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>


kunchenguid:
Thanks for the PR! This branch currently has a merge conflict with the base branch.

<sub>Noted for firstmate#577 at `b4ce454e`.</sub>

kunchenguid:

kunchenguid:
I am closing this because it has been waiting on a rebase or merge-conflict fix since 2026-07-15, and I have not seen a comment or push since then.

PR #585, opened 2026-07-15
fix(brief): eliminate heredoc-in-command-substitution bash 3.2 bug at the root

saucetech:
## Summary

- `bin/fm-brief.sh` had four `VAR=$(cat <<EOF ... EOF)` heredoc-in-command-substitution sites. Bash 3.2's lexer loses quote-tracking through a heredoc body nested inside `$(...)`, so a single unescaped apostrophe anywhere in that text breaks parsing of the *entire rest of the script* (`bash -n` fails). PR #173 previously worked around this by rewording the one sentence that happened to trip it, leaving the fragile pattern in place at all four sites. This PR fixes the root cause: each heredoc is now defined inside a small local function whose stdout is captured instead (`DOD=$(dod_text)`), keeping the heredoc lexically outside any `$(...)` and making it structurally immune to the bug regardless of body content.
- `bin/fm-spawn.sh`'s batch-dispatch `shared_args[@]` expansion was already fixed by the same prior PR (#173) with the portable `${shared_args[@]+"${shared_args[@]}"}` idiom. No code change needed there; verified clean under stock macOS Bash 3.2.57.
- Extended the existing `macos-stock-bash` CI job (added in #578) to cover both scripts going forward: `bash -n` on each, plus running `tests/fm-brief.test.sh` and `tests/fm-spawn-batch.test.sh` under real stock Bash 3.2.57, so this class of regression is caught on the actual target shell instead of only the Linux runners' modern bash.

## Test plan

- [x] `bash -n bin/fm-brief.sh` and `bash -n bin/fm-spawn.sh` under `/bin/bash` 3.2.57 (this machine's stock bash) and under a modern bash
- [x] `tests/fm-brief.test.sh` (11/11) and `tests/fm-spawn-batch.test.sh` (3/3) pass under both bash versions
- [x] Reproduced the original bug in isolation (`$(cat <<EOF ... apostrophe ... EOF)` fails to parse under 3.2 with both quoted and unquoted heredoc delimiters) and confirmed the post-fix script still parses even with an apostrophe re-injected into the DOD text
- [x] Diffed generated scaffolds (ship in all three delivery modes, scout, `--herdr-lab`, secondmate) between the pre-fix and post-fix script under bash 3.2.57 - byte-identical after normalizing incidental scratch-path differences
- [x] `bin/fm-lint.sh` (pinned ShellCheck 0.11.0) - clean
- [x] Simulated the added `macos-stock-bash` CI job steps locally in a sanitized stock-Bash environment - all pass

**No-mistakes gate note:** review/document/lint all completed cleanly through the local no-mistakes gate. The test step's single lump finding covered 4 pre-existing failures in `tests/fm-backlog-handoff.test.sh`, `tests/fm-pi-watch-extension.test.sh`, `tests/fm-secondmate-lifecycle-e2e.test.sh`, and `tests/fm-turnend-guard.test.sh` - none touch the files changed in this PR. Two were resolved by upgrading this sandbox's stale global `tasks-axi` (0.1.2 -> 0.2.3, matching what CI installs fresh); the remaining two are a Node v23.7.0 `ExperimentalWarning` runtime artifact in this sandbox, unrelated to firstmate code, so that gate step was skipped with documented reasoning rather than editing test files. The gate's automated `push` step then failed with a GitHub permissions error (the local git identity lacks write access to `kunchenguid/firstmate`), so this PR was opened manually via a fork per captain direction instead of through the gate's own push/PR steps.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -94,6 +94,25 @@ jobs:
             exit 1
           }
 
+          /bin/bash -n bin/fm-brief.sh
+          /bin/bash -n bin/fm-spawn.sh
+
+          brief_output=$(/bin/bash tests/fm-brief.test.sh)
+          printf '%s\n' "$brief_output"
+          brief_count=$(printf '%s\n' "$brief_output" | grep -c '^ok - ')
+          [ "$brief_count" -eq 11 ] || {
+            echo "::error::expected 11 fm-brief.sh tests, got $brief_count"
+            exit 1
+          }
+
+          spawn_batch_output=$(/bin/bash tests/fm-spawn-batch.test.sh)
+          printf '%s\n' "$spawn_batch_output"
+          spawn_batch_count=$(printf '%s\n' "$spawn_batch_output" | grep -c '^ok - ')
+          [ "$spawn_batch_count" -eq 3 ] || {
+            echo "::error::expected 3 fm-spawn.sh batch tests, got $spawn_batch_count"
+            exit 1
+          }
+
   invariants:
     name: Repo invariants
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


kunchenguid:

PR #594, opened 2026-07-15
fix(spawn): record the authoritative worktree and refuse pool exhaustion loudly

saucetech:
## Summary

Fixes two related defects in `bin/fm-spawn.sh`'s ship/scout worktree resolution observed on 2026-07-14:

**Defect 1 — wrong worktree recorded.** When treehouse DID create an isolated slot, the spawn summary line and `state/<id>.meta` sometimes recorded `worktree=<project clone path>` instead of the real treehouse slot path, even though the pane's actual cwd was the slot. Root cause: the recorded value came from a raw pane-path poll captured mid-wait, not from `validate_spawn_worktree`'s own freshly re-resolved, canonicalized toplevel check that runs immediately after.

Fix: `validate_spawn_worktree` now exposes its verified canonical path via a new `VALIDATED_WT_REAL` global; the treehouse-get wait call site reassigns `WT=$VALIDATED_WT_REAL` after the assert passes, so the recorded value can never diverge from what was actually verified. Orca's own worktree-create API call site is intentionally excluded (it never polls a pane and already returns a canonical path from its own API), so its existing tests are unaffected.

**Defect 2 — pool exhaustion silently proceeds.** With the treehouse pool transiently exhausted, a spawn's pane could stay in the project clone (the fleet's mirror), and `fm-spawn` still launched the agent there and returned success — real agents made scratch commits into the mirror before being emergency-stopped.

Fix: the worktree-discovery wait now retries across bounded, overridable rounds (`FM_SPAWN_WT_WAIT_ROUNDS`/`_TRIES`/`_INTERVAL`, defaulting to the original 60x1s bound plus 2 more rounds), re-sending `treehouse get` each retry round, since the pool has been observed growing new slots asynchronously within minutes. Exhausting every round is a hard refusal — never a silent launch into the project clone. Any isolation-check refusal now also kills the already-launched backend window via `fm_backend_kill` before exiting non-zero, so a refusal never leaves an agent-less window stranded in the tangled directory.

## Changes

- `bin/fm-spawn.sh` — the fix described above, plus review-round hardening (regex-validated `FM_SPAWN_WT_WAIT_INTERVAL` fallback, matching the codebase's other fractional-wait env vars)
- `tests/fm-backend.test.sh` — two new tests (stuck-pane refusal + window kill; late-settling-pane authoritative recording through a symlinked alias path) and an update to the existing symlink test to expect the new canonical recording
- `docs/configuration.md` — new env vars and behavior
- `docs/zellij-backend.md` — document-step fix dropping a now-stale "60s" figure
- `AGENTS.md` — section 7 update describing the new retry/kill/authoritative-recording behavior

Bash 3.2 (stock macOS) compatibility verified throughout (`/bin/bash -n`, and the full suite run under `/bin/bash` 3.2.57).

## Test plan

- [x] `bin/fm-lint.sh` (pinned ShellCheck) clean
- [x] `bash -n` clean on both GNU bash and stock macOS `/bin/bash` 3.2.57
- [x] Full `tests/*.test.sh` suite passes, including the two new tests, under both bash versions
- [x] Confirmed via `git stash` clean-checkout comparison that the one pre-existing local test failure (`fm-turnend-guard.test.sh`'s Node `ExperimentalWarning` quirk) is unrelated to this change and predates it
- [ ] CI green on this PR


kunchenguid:

PR #597, opened 2026-07-15
Windows (Git Bash/MSYS) support for spawn, supervision, adapters, and the test suite

Kallas95:
## Summary

This PR makes firstmate fully functional on Windows 11 under Git Bash (MSYS) with the herdr backend.
It is the stack we have been running in production locally since 2026-07-11; a crewmate was spawned, supervised, and shipped a merged PR end to end with these patches, and the targeted test suites now pass on Windows.

The commits fall into three groups.

### Runtime enablement (spawn and supervision path)

- **Session and wake-queue locks on Git Bash**: lock helpers gain an MSYS arm (PowerShell CIM pid walk, `MSYS=winsymlinks:sys`).
- **Worktree discovery for crewmate spawns**: herdr has no `foreground_cwd` on Windows, so pane cwd falls back to `.cwd` + `cygpath -u`; treehouse subshell detection points `COMSPEC` at PowerShell so the pane cwd updates via OSC 7.
- **Agent launch through Git Bash**: herdr panes run PowerShell on Windows, so the launch command is routed through `bash -l <launch.sh>`.
- **`fm_pid_identity` on MSYS**: BusyBox-style `ps` rejects `-o`, so the identity is parsed from the default `ps -p` columns (WINPID + STIME + COMMAND), preserving the pid-reuse guarantee the Unix `lstart+command` query gives. The Unix branch keeps the upstream `LC_ALL=C` pin.

### Node-side adapters

- **OpenCode plugin and Pi extension spawned `bin/*.sh` directly**: Windows node cannot exec a shebang script, and the spawn `error` event was swallowed as a pass, so the turn-end guard silently never ran. Scripts are now routed through bash on `win32`.
- **`fm-arm-command-policy.mjs` path matching**: `path.normalize`/`path.join` flip to backslashes on Windows node and never match the POSIX patterns, and MSYS hands `--root`/`--home` as `C:/...` while paths inside the command string stay `/c/...` - on Windows the deny policy was effectively inert. A `posixNormalize` helper folds every spelling into one forward-slash, uppercase-drive form.
- **TERM delivery to the Pi watch arm child**: native node's `child.kill()` is `TerminateProcess` on Windows, so the bash arm child died without running its TERM trap, and node's Windows pid cannot be mapped back after MSYS `exec` swaps the underlying Windows process. The launch command records the child's own MSYS pid (`$$` survives exec) in `state/.pi-watch-arm-msys-pid`, and `stopArm` delivers a real MSYS TERM via Git Bash `kill`, falling back to `child.kill` when unavailable.

### Test harness

- **TERM-resistant test peers must be MSYS processes**: MSYS `kill -TERM` to a non-MSYS binary (node) is emulated as `TerminateProcess`, which no in-process handler can ignore, so the watcher-lock healthy-peer scenario used a node peer that died and had its lock stolen. A `bash -c 'trap "" TERM; sleep 300'` peer ignores TERM identically on every platform.
- **`fm_fakebin_tool` replaces `ln -s` fakebins**: on MSYS `ln -s` copies the exe away from its DLLs (a copied `bash.exe` cannot load `msys-2.0.dll` from a bare-PATH fakebin), and bash resolves builtins like `printf` to a bare name no link can target. A `/bin/sh` wrapper execs the real path instead. Applied across the turnend-guard, cd-pretool, and arm-pretool suites.
- **Bootstrap suite**: the hermetic `BASE_PATH` (`/usr/bin:/bin`) has no git on Git-for-Windows (git lives in `/mingw64/bin`), which broke both the new `COMMON_TOOLS` git check and the fleet-sync fakes that delegate repo-state queries through `command git`; the ambient git's directory is appended on MSYS. The `add_real_jq` shim also strips the CRLF that native Windows jq builds emit, which poisoned exact-string assertions.

## Environment notes (not part of the diff)

Two host-level requirements surfaced while getting the suites green; they may be worth documenting in the Windows section of the docs:

- Native Windows jq builds (winget, even 1.8.x) open stdout in text mode and emit CRLF, which corrupts every bash `$(jq -r ...)` capture in the herdr adapter. We run an LF-stripping wrapper ahead of it on PATH.
- An empty locale (`LANG` unset) breaks multibyte glyph classification in `fm-composer-lib` (the composer `❯` row reads `pending` instead of `empty`); `LANG=C.UTF-8` fixes it.

## Test plan

On Windows 11 + Git Bash, with the environment notes above applied, all of the following pass (most failed or could not run before this stack):

`fm-bootstrap`, `fm-watcher-lock`, `fm-wake-queue`, `fm-turnend-guard`, `fm-spawn-batch`, `fm-spawn-dispatch-profile`, `fm-backend-herdr`, `fm-composer-lib`, `fm-arm-pretool-check`, `fm-cd-pretool-check`, `fm-pi-watch-extension`, `fm-pi-primary-types`, `fm-lint`.

No behavior change on macOS/Linux: every Windows arm is gated on `uname` `MINGW*|MSYS*` or `process.platform === "win32"`, and the shared-code changes (`posixNormalize`, wrapper fakebins, bash test peers) are platform-neutral.

### DIFF
--- a/.opencode/plugins/fm-primary-turnend-guard.js
+++ b/.opencode/plugins/fm-primary-turnend-guard.js
@@ -43,7 +43,14 @@ function resolvePath(anchor) {
 
 function runGuard(root) {
   if (!root) return Promise.resolve({ code: 0, stderr: "" });
-  return runProcess(`${root}/bin/fm-turnend-guard.sh`, [], '{"stop_hook_active":false}');
+  const script = `${root}/bin/fm-turnend-guard.sh`;
+  // Windows node cannot exec a shebang script directly; spawn would emit
+  // 'error' (swallowed as a pass) instead of running the guard. Route it
+  // through bash (Git Bash on PATH) there.
+  if (process.platform === "win32") {
+    return runProcess("bash", [script], '{"stop_hook_active":false}');
+  }
+  return runProcess(script, [], '{"stop_hook_active":false}');
 }
 
 async function letWatchArmRun(sessionID, client) {
--- a/.pi/extensions/fm-primary-pi-watch.ts
+++ b/.pi/extensions/fm-primary-pi-watch.ts
@@ -23,6 +23,7 @@ const state = process.env.FM_STATE_OVERRIDE || `${fmHome}/state`;
 const config = process.env.FM_CONFIG_OVERRIDE || `${fmHome}/config`;
 const armScript = `${fmRoot}/bin/fm-watch-arm.sh`;
 const marker = `${state}/.pi-watch-extension-loaded`;
+const armMsysPidFile = `${state}/.pi-watch-arm-msys-pid`;
 const extensionVersion = `sha256:${createHash("sha256").update(readFileSync(extensionFile)).digest("hex")}`;
 
 let child: any = null;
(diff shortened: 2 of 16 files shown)


Kallas95:
Closing - I'm maintaining these Windows/MSYS fixes as local patches for now. Happy to reopen with a refreshed stack if there's upstream interest.

PR #611, opened 2026-07-15
fix: deliver afk escalations when Herdr is idle

knowttl:
## Diagnosis

The primary read-only daemon log recorded repeated `inject deferred: supervisor pane busy (agent mid-turn)` events and a 31,726-second max-defer wedge.

`pane_is_busy` asked Herdr for native state but then re-scanned scrollback when it reported `idle`; historical quoted `esc to interrupt` output therefore made an idle supervisor appear busy indefinitely.

A related tmux submit-verification bug treated Claude’s `❯` plus U+00A0 empty composer as pending because POSIX `[:space:]` does not trim U+00A0.

## Fix

- Treat a legible Herdr native `busy` or `idle` status as conclusive for the supervisor injection guard, while retaining regex fallback for `unknown`.
- Keep the confirmed-empty composer guard and type-once, Enter-only submit model unchanged.
- Add shared U+00A0-aware edge trimming without discarding U+00A0 inside typed text.
- Record the 2026-07-15 incident and verification in `docs/herdr-backend.md`.

## Tests and verification

- `bash tests/fm-composer-lib.test.sh`
- `PATH="$PWD/.tmp-bin:$PATH" bin/fm-lint.sh` with the repository-pinned ShellCheck 0.11.0 installer.
- Regression coverage includes native-idle scrollback containing `esc to interrupt`, a max-defer delivery simulation, and the Claude U+00A0 composer state.

No Herdr session, workspace, or pane lifecycle command was used.

### DIFF
--- a/bin/fm-composer-lib.sh
+++ b/bin/fm-composer-lib.sh
@@ -159,6 +159,29 @@ fm_composer_strip_ghost() {
   '
 }
 
+# fm_composer_trim: trim ordinary whitespace plus the U+00A0 no-break-space
+# that Claude renders after an otherwise-empty prompt glyph. POSIX [:space:]
+# does not include U+00A0, so it must be handled explicitly without treating a
+# no-break space inside real typed text as disposable.
+fm_composer_trim() {  # <content>
+  local content=$1 nbsp=$'\302\240'
+  while :; do
+    case "$content" in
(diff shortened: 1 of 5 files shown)


knowttl:
opened against wrong base by mistake

PR #613, opened 2026-07-15
fix(watch): canonicalize watcher-lock paths against case-insensitive-fs aliases

saucetech:
## Summary

Fixes a path-casing false negative in the watcher arm's healthy-watcher verification, observed live on 2026-07-15.

`bin/fm-watch-arm.sh` (via the shared `fm_watcher_lock_matches_pid` in `bin/fm-wake-lib.sh`) verified a healthy watcher by comparing the lock's recorded `fm-home`/`watcher-path` fields against its own resolved `$FM_HOME`/`$WATCH` with exact string equality. `SCRIPT_DIR`/`FM_HOME` are derived via `cd + pwd`, which preserves the caller's casing rather than resolving to a canonical form. On macOS's case-insensitive filesystem, the same directory is reachable via differently-cased paths (e.g. a harness that `cd`'d via a lowercased alias), so:

- `healthy_watcher()` returned false for a genuinely live, fresh watcher — the arm reported `watcher: FAILED - no live watcher with a fresh beacon` and forked a duplicate child that always lost the singleton race.
- `--restart` could not match the lock's pid via `watch_lock_matches_pid`, so it silently no-op'd instead of restarting this home's own watcher.

## Changes

- `bin/fm-wake-lib.sh` — new `fm_canonical_path` helper (`cd` + `pwd -P`, bash 3.2 safe, no GNU `realpath` dependency); `fm_watcher_lock_matches_pid` now compares canonicalized forms of both `fm-home` and `watcher-path` on both sides, so a case-insensitive-filesystem alias (or a stray symlink) of either path matches.
- `bin/fm-watch-arm.sh` — `clear_stale_recorded_watcher_lock` uses the same canonical comparison for both fields, so a case-aliased recorded lock is still recognized as this home's own.
- `bin/fm-watch.sh` — the watcher's lock writer now records canonical `fm-home`/`watcher-path` at lock time too, so writer and reader converge on one form. The reader still re-canonicalizes independently, so a lock written by an older, un-fixed watcher still matches.
- `tests/fm-watcher-lock.test.sh` — three new tests: a direct unit test of the fixed comparator tolerating a path alias, an end-to-end test that arm mode attaches to (rather than falsely fails) a healthy watcher despite a path-aliased `FM_HOME`, and an end-to-end test that `--restart` actually restarts (rather than silently no-ops on) this home's own watcher under the same alias condition. These use a symlink-based alias rather than relying on the host's case-folding behavior, since that's deterministic on every OS/filesystem (a literal case-fold alias never occurs on a case-sensitive Linux CI runner, and proved unreliable even in this PR's own validation sandbox) — it exercises the identical `fm_canonical_path` codepath the real bug hits. Verified all three fail against the pre-fix code and pass against the fix.

Lock semantics, singleton behavior, and beacon freshness rules are unchanged — only path comparison/normalization.

### Audit of sibling guards

Checked the other places that compare recorded watcher/session state by path for the same exact-string-comparison class:

- `bin/fm-session-start.sh`'s lock step delegates to `bin/fm-lock.sh`, which is PID-only (no path comparison) — not affected.
- `bin/fm-turnend-guard.sh` already calls the shared `fm_watcher_healthy`/`fm_watcher_lock_matches_pid` — it inherits this fix for free with no direct edit needed.
- `bin/fm-guard.sh` delegates to `bin/fm-supervision-lib.sh`, which only checks beacon mtime freshness, no path comparison — not affected.

## Test plan

- [x] `bin/fm-lint.sh` (pinned ShellCheck) clean
- [x] `tests/fm-watcher-lock.test.sh` full suite passes, including the three new tests
- [x] Confirmed the three new tests fail against the pre-fix code (via `git stash`) and pass against the fix
- [x] `tests/fm-daemon.test.sh`, `tests/fm-secondmate-sync.test.sh`, `tests/fm-turnend-guard.test.sh` unaffected
- [x] no-mistakes pipeline: review, test, document, and lint steps all passed (one test-step fix round applied `NODE_NO_WARNINGS=1` to `tests/fm-turnend-guard.test.sh` to suppress a pre-existing, unrelated Node.js `ExperimentalWarning` quirk — the same known flake noted in #594's test plan)
- [ ] CI green on this PR

### DIFF
--- a/bin/fm-wake-lib.sh
+++ b/bin/fm-wake-lib.sh
@@ -50,14 +50,49 @@ fm_path_age() {
   echo $(( $(date +%s) - m ))
 }
 
+# fm_canonical_path <path>
+# Resolves <path> to its physical form (no symlinks, on-disk casing) via `cd` +
+# `pwd -P` - bash 3.2 safe, no GNU realpath dependency. On a case-insensitive
+# filesystem (macOS default), two differently-cased strings for the same
+# directory canonicalize to one identical string, so callers that compare
+# resolved paths instead of raw strings never false-negative on a casing or
+# symlink alias of the same location (e.g. a harness that cd'd via a
+# lower-cased path alias). Works for both directories and regular files: a
+# file's containing directory is resolved and its basename is reattached
+# unchanged, since a differently-cased leaf *filename* (as opposed to a
+# directory component in the chain) is not something this codebase's
+# fixed-literal script paths ever vary. Falls back to the original string
+# unresolved (and a non-zero return) when <path> is empty or its directory
+# cannot be entered (e.g. it does not exist), so a miss degrades to the prior
+# literal-string comparison rather than silently matching everything.
+fm_canonical_path() {
+  local path=$1 dir base resolved
+  [ -n "$path" ] || { printf '%s\n' "$path"; return 1; }
+  if [ -d "$path" ]; then
+    resolved=$(cd "$path" 2>/dev/null && pwd -P) || { printf '%s\n' "$path"; return 1; }
+    printf '%s\n' "$resolved"
+    return 0
+  fi
+  dir=$(dirname "$path")
+  base=$(basename "$path")
+  resolved=$(cd "$dir" 2>/dev/null && pwd -P) || { printf '%s\n' "$path"; return 1; }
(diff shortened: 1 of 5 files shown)


kunchenguid:

PR #621, opened 2026-07-16
fix(supervision): fingerprint lock identity from /proc start_ticks, not ps lstart

noahjobse:
## Problem

The supervision lock-identity fingerprint (`fm_pid_identity` in `bin/fm-wake-lib.sh`) is derived from `ps -p <pid> -o lstart=`. `lstart` is not a stable primitive: the kernel derives it from `btime` (boot time) plus the process's `starttime` ticks. `starttime` never moves, but `btime` is recomputed and jitters (observed ~1s, roughly every 30s, on a WSL2 host).

Every `btime` jump shifts the derived `lstart` of *every* live process. So a live lock holder's fingerprint drifts against its own process with no second writer. The health predicate `fm_watcher_lock_matches_pid` then misreads a healthy, beating, lock-holding watcher as "not ours" and raises false **"supervision off"** alarms. A forced re-arm off that misread could then act on the wrong process.

This is a self-inflicted flap: nothing external changes, yet supervision reports itself down.

## Fix

Fingerprint from `/proc/<pid>/stat` field 22 (start ticks) instead. That value is boot-relative, carries no `btime` term, and is constant for the process lifetime, so it cannot drift.

- New `fm_pid_start_ticks` reads field 22, counting fields *after the last `)`* so a `comm` containing spaces or `)` can't shift the count.
- `fm_pid_identity` now emits `<start_ticks> <command>`. The old `lstart` read is kept only as a fallback where `/proc` does not exist (macOS). `LC_ALL=C` is still pinned so the command half stays locale-invariant.

### Compatibility

The identity string is only *written* and *equality-compared* by every consumer (`fm-watch`, `fm-watch-arm`, `fm-afk-launch`, `fm-afk-start`, `fm-supervise-daemon`) — none parse its fields — so this is drop-in. The one transition effect: a watcher/daemon already running at upgrade whose recorded identity is an `lstart` string re-reads as not-matching once, is treated as stale, and re-records its ticks identity. A single, self-healing event.

## Verification

Two focused cases added to `tests/fm-watcher-lock.test.sh`:

1. **byte-stability** — the same live pid fingerprints byte-identically across 20 repeated reads.
2. **start-ticks, not lstart** — asserts the fingerprint's start half is `/proc/<pid>/stat` field 22, and that it stays fixed across a *simulated `btime` shift* under which the equivalent `lstart` derivation (`btime + starttime/HZ`) provably would move. This pins the cure so a future refactor can't quietly regress it back to `lstart`. Skips cleanly where `/proc` is absent.

The existing `test_pid_identity_is_locale_invariant` still passes. Full `tests/fm-watcher-lock.test.sh` suite is green (24 cases, exit 0) on Linux.

Verified on the original WSL2 host where the drift was observed: the new fingerprint held byte-identical across a real `btime` jitter that shifted `lstart`.

## Scope

Deliberately narrow: the root-cause primitive plus its regression tests. This is extracted from a larger torn-watcher-lock hardening set on the contributing fork; this PR is the focused, reviewable core of it.

### DIFF
--- a/bin/fm-wake-lib.sh
+++ b/bin/fm-wake-lib.sh
@@ -23,17 +23,45 @@ fm_pid_alive() {
   kill -0 "$pid" 2>/dev/null
 }
 
+# The kernel's own start time for a pid: boot-relative clock ticks, /proc/<pid>/stat
+# field 22. comm (field 2) is parenthesised and may itself contain spaces or ')',
+# so fields are counted from after the LAST ')'. Absent where /proc is (macOS).
+fm_pid_start_ticks() {
+  local pid=$1
+  [ -r "/proc/$pid/stat" ] || return 1
+  awk '{ s = $0; sub(/^.*\) /, "", s); n = split(s, f, " "); if (n < 20) exit 1; print f[20] }' \
+    "/proc/$pid/stat" 2>/dev/null
+}
+
 fm_pid_identity() {
-  local pid=$1 out
+  local pid=$1 start cmd
   case "$pid" in
     ''|*[!0-9]*) return 1 ;;
   esac
-  # Pin LC_ALL=C so lstart's date format is locale-invariant: the identity is
-  # written under one locale but re-read under the machine's ambient locale, which
-  # would otherwise mismatch on a non-C locale (e.g. ko_KR) and reject a live watcher.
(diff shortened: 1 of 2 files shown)


noahjobse:
Closing for now - shelving the no-mistakes-raised resubmission. The fix remains in noahjobse/firstmate and this can be reopened later if we return to it.

PR #634, opened 2026-07-16
feat(spawn): configurable CLAUDE_CONFIG_DIR for hands-free multi-account claude launch

### LINKED ISSUE #599: fm-spawn: claude crewmate launch hardcodes bare 'claude', failing on multi-account machines
## Summary

`bin/fm-spawn.sh`'s claude launch template runs **bare `claude`**, which resolves to the
default config dir (`~/.claude`). On a machine with **multiple Claude accounts** — where auth
lives in alternate config dirs and the default `~/.claude` is empty/unauthenticated — the
spawned crewmate lands on the "Select login method" wall and can never start.

## Details

Launch template (claude branch):

```
claude) printf '%s' 'CLAUDE_CODE_ENABLE_PROMPT_SUGGESTION=false claude --dangerously-skip-permissions __MODELFLAG____EFFORTFLAG__"$(cat __BRIEF__)"' ;;
```

There is no way to tell fm-spawn which account/profile a crewmate's claude should use. On a
multi-account setup the machine typically has a per-shell profile mechanism (e.g. an
`llm-profile`-style `lm <profile>` + `cl` flow that exports `CLAUDE_CODE_USE_WEB_LOGIN=1`, or a
`CLAUDE_CONFIG_DIR` pointing at an authenticated dir). fm-spawn bypasses all of that.

## Impact

FirstMate cannot spawn autonomous **claude** crewmates on a multi-account machine — a real
cross-machine portability blocker, since each machine has its own auth/profile flow.

## Suggested fix

Make the crewmate **launch verb** machine-configurable rather than hardcoding bare `claude`:

- Let a local, gitignored config (e.g. a launch-command/prefix knob, or honoring
  `CLAUDE_CONFIG_DIR` / the machine's profile flow) supply how claude is invoked, while
  fm-spawn keeps appending its own autonomy + brief args
  (`--dangerously-skip-permissions`, `CLAUDE_CODE_ENABLE_PROMPT_SUGGESTION=false`,
  `"$(cat <brief>)"`).
- Preserve today's behavior when the knob is absent (bare `claude`), so single-account
  machines are unaffected.

This keeps the mechanics in `fm-spawn.sh` while allowing each machine to point crewmates at an
authenticated account.

---
> 🤖 [Claude Code](https://claude.com/claude-code) · claude-opus-4-8 · 2026-07-15

tkt028:
Fixes #599

## Problem

`bin/fm-spawn.sh`'s claude launch template runs **bare `claude`**, which resolves to the default config dir (`~/.claude`). On a machine with multiple Claude accounts — where auth lives in alternate config dirs and the default `~/.claude` is empty/unauthenticated — a spawned crewmate lands on the "Select login method" wall and can never start. There was no way to tell fm-spawn which authenticated config dir a claude crewmate should use.

## Change

Add a local, gitignored **`config/crew-config-dir`** knob naming the Claude config dir (`CLAUDE_CONFIG_DIR`) a spawned `claude` crewmate authenticates with. When set, `fm-spawn` prefixes the claude launch command with `CLAUDE_CONFIG_DIR=<dir> `, while still appending its own autonomy and brief args unchanged.

- **Claude-harness only.** A new `__CLAUDECONFIGDIR__` placeholder lives solely in the claude launch template; every other harness ignores the knob.
- **Scoped, not global.** The value becomes a per-launch env prefix on the launch command, so it never mutates the captain's global config.
- **Backward compatible.** An absent or empty knob is **byte-identical** to prior behavior — bare `claude` on its default config dir — so single-account machines are unaffected.
- **Path-safe.** Only leading/trailing whitespace is trimmed (internal spaces in a config-dir path are preserved), and the value is shell-quoted into the launch string.

## Tests

`tests/fm-spawn-claude-config-dir.test.sh` drives a real claude ship spawn through `fm-spawn.sh` with a fake tmux that records the literal launch string, and asserts both directions:

1. With `config/crew-config-dir` set → the launch gains the `CLAUDE_CONFIG_DIR=<value>` prefix in the correct position.
2. With the knob absent → **no** `CLAUDE_CONFIG_DIR` prefix (the byte-identical prior launch).

The behavioral assertion drives the launch-string construction rather than grepping for a token, and was negative-tested (it fails when the substitution is removed).

## Validation

- `bin/fm-lint.sh` (full shellcheck set, pinned version) — clean.
- `bash -n` on the whole `bin/` toolbelt — clean.
- New test passes; the rest of the suite shows only pre-existing environment-dependent failures (missing/incompatible `tasks-axi`, optional backends absent) unchanged by this diff.

## Docs

`docs/configuration.md` documents the knob in the harness support section: what it names, claude-harness-only scope, and the absent-knob backward-compat guarantee.

---
> 🤖 [Claude Code](https://claude.com/claude-code) · claude-opus-4-8 · 2026-07-16

### DIFF
--- a/bin/fm-spawn.sh
+++ b/bin/fm-spawn.sh
@@ -67,6 +67,8 @@
 #   $vars and silently breaks ad-hoc `for ... in $pairs` loops).
 #   Launch templates live in launch_template() below; placeholders replaced before launch:
 #     __BRIEF__    absolute path to data/<task-id>/brief.md
+#     __CLAUDECONFIGDIR__  optional `CLAUDE_CONFIG_DIR=<dir> ` prefix from config/crew-config-dir
+#                  (claude only; absent knob = bare `claude`, default config dir; see #599)
 #     __TURNEND__  absolute path to state/<task-id>.turn-ended (for harnesses whose
 #                  turn-end signal rides the launch command, e.g. codex -c notify=[...])
 #     __PIEXT__    absolute path to state/<task-id>.pi-ext.ts (pi turn-end extension,
@@ -84,7 +86,7 @@ set -eu
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 
 usage() {
-  sed -n '2,78p' "$0" | sed 's/^# \{0,1\}//'
+  sed -n '2,80p' "$0" | sed 's/^# \{0,1\}//'
 }
 
 case "${1:-}" in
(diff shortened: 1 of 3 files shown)


tkt028:
Superseding this in favor of #1031.

#1031 ships the same captain-chosen `CLAUDE_CONFIG_DIR` feature, but improves on this PR in three ways:
1. **Rebased onto current `main`** — this branch was ~70 commits behind and conflicts on the `__OPINPUT__ encode launch-brief` line `main` has since changed (which is why its checks were stuck DIRTY).
2. **Includes the two fixes this PR was missing** — gitignoring `config/crew-config-dir` and expanding `~`/`$HOME` in its value.
3. **Re-raised through the no-mistakes pipeline** (review + tests green), so the required gate passes.

Closing here in favor of **#1031**. Thanks!

---
> 🤖 [Claude Code](https://claude.com/claude-code) · claude-opus-4-8[1m] · 2026-07-25

PR #648, opened 2026-07-16
fix(spawn): skip treehouse get's transient cwd in worktree-discovery poll

psacc:
## Intent

Contribute a single, self-contained bug fix to the upstream firstmate repo (kunchenguid/firstmate) THROUGH this pipeline, so the PR carries the no-mistakes signature that upstream's "PR must be raised via no-mistakes" CI gate requires. This supersedes the existing hand-made PR #480 (https://github.com/kunchenguid/firstmate/pull/480), which is permanently blocked by that gate; the PR body should say "Supersedes #480" and reuse its problem/root-cause writeup.

The change is a cherry-pick of commit 2a5fa59, "fix(spawn): skip treehouse get's transient cwd in worktree-discovery poll". It repairs a herdr-backend single-slot wedge. Root cause: `treehouse get` transits intermediate cwds while it sets a worktree up (momentarily `/` and then the treehouse root) before it creates and enters the pooled worktree, and herdr's foreground_cwd exposes those transients (tmux's pane_current_path reports the settled cwd and never surfaced them, which is why only herdr wedged). bin/fm-spawn.sh's worktree-discovery poll previously accepted the FIRST cwd that merely differed from the project dir, so it captured `/`, handed that transient to validate_spawn_worktree, failed the isolation gate, and aborted the spawn, while the still-initializing `treehouse get` completed and held the pool's only slot forever with no crewmate attached (so the next spawn also failed). The fix gates the poll on a new non-fatal is_isolated_worktree predicate (a real git-worktree root, physically distinct from the primary checkout, the same rule validate_spawn_worktree now delegates to), so it skips `/` and the treehouse root and keeps polling until the pane settles in the real worktree. It is a no-op for tmux, whose first observed non-project cwd already is that worktree.

Deliberate decisions a reviewer should not flag as mistakes:
- The branch was created OFF upstream/main (c115561) as a single commit carrying ONLY this fix - deliberately not off the fork's default branch, so the PR contains the fix alone and no fork merge history.
- The PR must target UPSTREAM (base kunchenguid/firstmate main) via fork routing: the branch is pushed to the psacc/firstmate fork and the PR is opened against kunchenguid. This is intentional, not a misconfiguration.
- Files touched are exactly: bin/fm-spawn.sh, docs/herdr-backend.md, tests/fm-backend.test.sh, tests/fm-tangle-guard.test.sh. Nothing else should ride along.
[...]

Issue #608, opened 2026-07-15
fm: crewmate --dangerously-skip-permissions spawn pollutes the user's shared/symlinked settings.json


tkt028:
## Summary

Launching a crewmate with `--dangerously-skip-permissions` and accepting its one-time **"Bypass
Permissions mode"** warning makes Claude Code persist `skipDangerousModePermissionPrompt: true`
into the **active `settings.json`**. When the user's `~/.claude*/settings.json` is a **symlink to
a shared, git-tracked source** (a common dotfiles/multi-account setup), this leaks a **global
safety-guardrail change** into the user's repo — affecting every account, and the whole fleet if
committed.

## Root cause

Crewmates run in the **user's own Claude config dir** (fm-spawn launches bare `claude`, so it
resolves the default/shared config). Their launch-mode acceptances — and their auth — therefore
write to the *user's* config, not a crewmate-scoped one.

## Reproduction

1. `~/.claude/settings.json` is a symlink to a git-tracked source file.
2. Spawn a crewmate (`--dangerously-skip-permissions`) and accept the Bypass Permissions warning.
3. `git -C <dotfiles> diff` shows an **uncommitted** `+"skipDangerousModePermissionPrompt": true`
   in the shared settings — a guardrail the user never intentionally disabled, now affecting all
   accounts.

## Impact

- Silent, **global** reduction of the dangerous-mode confirmation guardrail from a single spawn.
- Surfaces as an unexpected diff in the user's dotfiles repo; risks being committed/pushed
  fleet-wide.

## Suggested fix

Launch crewmates with a **dedicated `CLAUDE_CONFIG_DIR`** (crewmate-scoped), so their settings,
auth, and launch-mode acceptances never touch the user's shared config. **This is the same fix as
#599** (multi-account auth) — one dedicated-config-dir change resolves both: crewmates
authenticate in their own dir *and* stop polluting the user's shared settings.

## Related

- #599 (bare `claude` fails on multi-account machines) — same root, same fix.

---
> 🤖 [Claude Code](https://claude.com/claude-code) · claude-opus-4-8 · 2026-07-15

PR #653, opened 2026-07-16
fix(spawn): validate each sampled pane cwd in the worktree-discovery poll

karotkriss:
## Intent

**Rebased onto current `main`. This now completes #765 rather than competing with it - please read the relation section first, since the original framing of this PR is out of date.**

#765 (`fix(spawn): require two stable reads before accepting worktree path`) has landed and fixes a real half of this race: a brand-new window's `pane_current_path` can transiently report a stale, unrelated-but-real path before the shell catches up with `treehouse get`'s `cd`, so it now requires two consecutive reads to agree before accepting a sample. That guard is correct and this PR keeps it intact - the candidate-stability structure below is #765's, not a reimplementation.

What #765 does not close is *which* paths are eligible to be accepted at all. Its in-loop gate is "differs from the project checkout", and its post-loop `validate_spawn_worktree` gate is "is a real worktree top, distinct from the project". **The firstmate home/root passes both.** It is a genuine git worktree top, and it is not the project checkout. `docs/architecture.md` already notes the firstmate repo dispatches crewmates to work on itself, so `FM_ROOT` is a real repo root in production, not a hypothetical.

That matters because `treehouse` setup hops through intermediate directories (the user's home, the firstmate operational home) before settling in the pooled worktree. A hop that stalls in the firstmate home/root is stable across reads, so it satisfies #765's stability requirement and then passes both of its path gates.

### Residual gap, reproduced

Probe against unmodified `main`, fake pane reporting `FM_ROOT` persistently (what a stalled hop looks like):

```
--- exit status: 0
spawned probe-fmroot-a1 ... worktree=/tmp/probe653/main
--- recorded worktree= :
worktree=/tmp/probe653/main
```

Exit 0, and the firstmate root recorded as the task worktree. The consequence is the one the spawn assertion exists to prevent: the crew's turn-end (Stop) hook is installed into the primary checkout, so the primary session fires the crew's turn-ended marker on its own cadence, producing phantom watcher wakes that survive teardown (same failure family as #608).

Same probe with this change applied:

```
--- exit status: 1
error: treehouse get did not enter an isolated worktree within 3s ...
--- recorded worktree= : (no meta written)
```

### Both guards are required

Neither change subsumes the other, verified in both directions:

- **#765's regression test passes with this change.** `tests/fm-spawn-worktree-settle.test.sh` is green on this branch - the stale-read protection is preserved, not reverted.
- **This PR's earlier single-read version failed that test** (`not ok - meta did not record the settled worktree`). A transient stale read can itself be a valid, distinct worktree root, so per-sample validation alone is insufficient. That is exactly why the rebase keeps #765's two-consecutive-reads requirement instead of replacing it.

The merged accept condition is therefore: a sample must satisfy `fm_isolated_worktree_root` **and** two consecutive reads must agree on it.

## What Changed

- Add `fm_isolated_worktree_root` to `bin/fm-tangle-lib.sh`: a shared predicate accepting a sampled path only when its physical (`pwd -P`) form resolves back to itself via `git rev-parse --show-toplevel` and differs from every forbidden physical path passed to it (the project checkout, the firstmate home/root).
- Extend `bin/fm-spawn.sh`'s post-`treehouse get` wait loop to validate every sample with that predicate, layered on top of #765's existing candidate-stability loop. The predicate's echoed physical path feeds the stability comparison, so no extra `real_path_or_raw` call is added. A rejected sample clears the candidate and polling continues, rather than accepting a stable-but-wrong directory and then hard-failing after the loop.
- `validate_spawn_worktree` delegates to the same predicate so there is exactly one definition of a genuine isolated worktree; its orca-path call keeps the existing project-only exclusion (behavior-preserving, since orca worktrees come from orca's own API rather than a cwd poll). The now-redundant post-loop treehouse validation call is removed, since the loop proves the same property per sample.
- Make the wait bound `FM_SPAWN_WT_TIMEOUT`-overridable (production keeps 60s) and name the last sampled path in the timeout refusal for diagnosis.
- Add `tests/fm-spawn-worktree-wait.test.sh`: predicate acceptance/rejection cases (non-repo dir, nonexistent path, project clone, symlink to the project, firstmate home as a valid-but-wrong repo root, worktree subdir, genuine worktree top), a replay of the observed hop sequence (project -> user home -> firstmate home -> worktree) that must settle only on the last, and structural pins that the poll is wired to the predicate.
- Update `tests/fm-tangle-guard.test.sh` for the new per-sample refusal message and the test-scoped wait bound, and sync docs (`AGENTS.md` isolation wording, `docs/architecture.md`, the `FM_SPAWN_WT_TIMEOUT` entry in `docs/configuration.md`, the `fm-tangle-lib.sh` row in `docs/scripts.md`).

Supersedes the earlier relation note to #648: this branch is now built on `main` with #765 landed, and the residual gap described above is what remains.

## Risk Assessment

✅ Low: the accept condition is strictly narrower than `main`'s (a genuine isolated worktree top, agreed on by two consecutive reads, instead of any stable non-project cwd), #765's guard is preserved and pinned by its own test, the timeout and refusal paths are preserved, and the behavior is pinned by regression tests replaying the observed failure sequence.

## Testing

- `bash tests/fm-spawn-worktree-settle.test.sh` - green (#765's own regression test, unmodified, on this branch).
- `bash tests/fm-spawn-worktree-wait.test.sh` - green (predicate cases, hop-sequence replay, structural pins).
- `bash tests/fm-tangle-guard.test.sh` - green (real `bin/fm-spawn.sh` against a fake tmux/treehouse: aborts on non-worktree and primary-checkout panes, succeeds on a genuine worktree).
- `bash tests/fm-spawn-batch.test.sh`, `bash tests/fm-spawn-dispatch-profile.test.sh` - green.
- `bin/fm-lint.sh` - green under the pinned ShellCheck 0.11.0.

## Pipeline

This branch was pushed through `git push no-mistakes` and validated by the pipeline (run outcome: passed): intent, rebase, review, document, lint, and push completed; the review round surfaced no findings on the final head; the local test step was skipped in favor of the direct suite runs above because the only local failures are the environmental ones that reproduce identically on clean `main`.
The pipeline's PR-creation step was skipped by a transient local `gh` auth check failure, so this PR was opened directly from the pipeline-validated head (`90300f9`, the exact commit the pipeline pushed to the fork); CI on this PR runs the portable suite as usual.

The branch has since been rebased onto current `main` to resolve the conflict with #765; the conflict resolution is described in the Intent section above.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -342,7 +342,7 @@ A forced repair must use the home-scoped owner path emitted by supervision instr
 
 Guard warnings do not replace the contract.
 Queued wakes must be drained before other action, stale liveness must be repaired through the emitted protocol, and the worktree-tangle warning must be resolved without touching unlanded work.
-The spawn assertion and generated ship brief must both enforce that project work starts in an isolated disposable worktree, never the primary checkout.
+The spawn assertion and generated ship brief must both enforce that project work starts in an isolated disposable worktree, never the primary checkout or the firstmate home/root.
 Harness-aware turn-end guards are structural backstops, not permission to omit the live cycle.
 
 ### Away-mode stub
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -240,7 +240,7 @@ Write the task-specific brief under section 11 before spawning.
 ### Dispatch and supervision handoff
 
 Spawn only through `bin/fm-spawn.sh` after the profile and backend checks in section 4.
-The spawn must resolve a genuine isolated task worktree distinct from the primary checkout; a failed isolation assertion stops the task.
+The spawn must resolve a genuine isolated task worktree distinct from both the primary checkout and the firstmate home/root; a failed isolation assertion stops the task.
 After spawning, confirm the worker is processing the brief, handle any trust dialog through `harness-adapters`, and record ship or scout work as under way.
 A persistent secondmate is recorded in the secondmate registry and runtime state, never as a backlog work item.
 
--- a/bin/fm-spawn.sh
+++ b/bin/fm-spawn.sh
@@ -66,8 +66,14 @@
 #   provisioned firstmate home; the default is kind=ship.
 #   Before a secondmate launch, the home is locally fast-forwarded to the primary
 #   default-branch commit when safe; skipped syncs warn and launch unchanged.
-#   Ship/scout spawns refuse to launch unless the resolved task path is a real
-#   git worktree root distinct from the primary project checkout.
+#   Ship/scout treehouse-backed spawns accept a sampled task path only once it is a
+#   real git worktree root distinct from the primary project checkout and the
+#   firstmate home/root (treehouse setup hops through intermediate cwds; mid-hop
+#   samples are rejected and polling continues) and two consecutive reads agree on
+#   it (a transient stale pane read can itself be a valid but wrong worktree root),
+#   and refuse to launch if none
+#   appears within FM_SPAWN_WT_TIMEOUT (default 60s); orca spawns keep validating
+#   the worktree path orca returns against the project checkout.
 # Batch dispatch: pass one or more `id=repo` pairs instead of a single <id> <project>, e.g.
 #     fm-spawn.sh fix-a-k3=projects/foo add-b-q7=projects/bar [--scout]
 #   Each pair re-execs this script in single-task mode, so the single path stays the only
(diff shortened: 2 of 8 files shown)


kunchenguid:

<sub>Noted for firstmate#653 at `90300f97`.</sub>

karotkriss:
Rebased and resolved conflicts as asked

kunchenguid:

<sub>Noted for firstmate#653 at `e8ead7e8`.</sub>

karotkriss:
Closing this as the real issue behind the initial PR was fixed by another contributor, I dont see this bring much benefit as it is right now.

PR #681, opened 2026-07-17
fix(pi): distinguish stale locks when arming watcher

kunchenguid:
## Intent

Fix the Pi primary watcher arm refusal so an owned session lock still arms, a live other holder keeps the existing read-only refusal, and an absent or dead holder gets a distinct message saying no live session holds the lock and directing the operator to run bin/fm-session-start.sh to reclaim it and then call fm_watch_arm_pi to re-arm. Add focused coverage for owned, live-other, absent, and dead-holder states, and keep the relevant Pi, lock, type, turn-end, and lint checks green. Do not broaden this safety fix into a session-start work-under-way nudge because reliable work-state detection would require wider extension changes; record that optional idea as a PR follow-up instead.

## What Changed

- Updated Pi watcher arming so owned session locks still arm, live other holders stay read-only, and absent or dead locks report the `fm-session-start.sh` recovery path before re-arming.
- Added focused Pi extension coverage for owned, live-other, absent, and dead-holder lock states.
- Documented the stale-lock recovery step in the Pi supervision protocol and stabilized watcher extension async waits.

## Risk Assessment

✅ Low: The change is narrowly scoped to distinguishing Pi watcher arm lock states and adds focused coverage for the required owned, live-other, absent, and dead-holder cases without broadening session-start behavior.

## Testing

The prompt-provided baseline full-suite was already green; I then captured reviewer-visible Pi tool output for absent/dead/live-other/owned lock states and ran the focused Pi watcher, watcher-lock, turn-end guard, and repeated Pi watcher rearm-flake checks successfully, with lint/type static checks skipped only because this prompt forbids them.

<details>
<summary>Evidence: Pi lock-state fm_watch_arm_pi transcript</summary>

<code>absent-lock: ok=false message=watcher: not armed - no live session holds the lock; run bin/fm-session-start.sh to reclaim it, then call fm_watch_arm_pi to re-arm&#10;dead-holder-lock: ok=false message=watcher: not armed - no live session holds the lock; run bin/fm-session-start.sh to reclaim it, then call fm_watch_arm_pi to re-arm&#10;live-other-lock: ok=false message=watcher: read-only - session lock is held by another firstmate session&#10;owned-lock: ok=true message=watcher: started Pi extension arm child 1&#10;arm-log: arm args=--restart ...</code>

```text
absent-lock: ok=false message=watcher: not armed - no live session holds the lock; run bin/fm-session-start.sh to reclaim it, then call fm_watch_arm_pi to re-arm
dead-holder-lock: ok=false message=watcher: not armed - no live session holds the lock; run bin/fm-session-start.sh to reclaim it, then call fm_watch_arm_pi to re-arm
live-other-lock: ok=false message=watcher: read-only - session lock is held by another firstmate session
owned-lock: ok=true message=watcher: started Pi extension arm child 1
arm-log: arm args=--restart home=/var/folders/5x/4nqprlbx0518k3ybcb1sz6gr0000gn/T/no-mistakes-evidence/01KXRV1BXD7SNEK16C5999DBC4/pi-lock-arm-e2e/home root=/var/folders/5x/4nqprlbx0518k3ybcb1sz6gr0000gn/T/no-mistakes-evidence/01KXRV1BXD7SNEK16C5999DBC4/pi-lock-arm-e2e/repo
```
</details>
<details>
<summary>Evidence: Focused behavior test transcript</summary>

```text
tmux 3.6a
== tests/fm-pi-watch-extension.test.sh ==
ok - Pi primary watcher extension is tracked, self-hashing, and self-locating
ok - Pi secondmate launch wiring includes both tracked primary extensions
ok - Pi extension reports external healthy watcher output
ok - Pi custom tool returns text content and structured details
ok - Pi watcher arm distinguishes all session lock ownership states
ok - Pi process-exit cleanup listener has a bounded lifecycle
ok - Pi process-exit cleanup stops the attached arm child
ok - OpenCode primary watcher plugin has the verified TUI wake wiring
ok - OpenCode plugins have an explicit ESM boundary even under a typeless parent package
ok - OpenCode watcher plugin uses the effective FM_HOME state
ok - OpenCode watcher plugin sources the effective config
ok - OpenCode watcher plugin requires session lock ownership
ok - OpenCode watcher coordinator respects primary scope
ok - OpenCode watcher plugin rearms after a watcher wake
ok - OpenCode watcher plugin coordinates with the turn-end guard
ok - OpenCode healthy arm output does not suppress the turn-end guard
== tests/fm-watcher-lock.test.sh ==
ok - guard banner leads when down with pending wakes (re-arm-after-drain) and stays silent when fresh
ok - watch restart reports a healthy peer without attaching to it
ok - arm attaches to a live fresh watcher and exits only when that cycle ends
ok - arm attaches to a peer watcher after child stands down and exits when peer dies
== tests/fm-turnend-guard.test.sh ==
ok - fm_supervision_unhealthy: false with no state/*.meta at all
ok - fm_supervision_unhealthy: true with in-flight task and no beacon ever
ok - fm_supervision_unhealthy: true with in-flight task and a beacon far outside the grace window
ok - fm_supervision_unhealthy: false with in-flight task and a fresh beacon
ok - fm_supervision_status: FM_SUP_QUEUE_PENDING tracks state/.wake-queue
ok - fm-turnend-guard: silent no-op with nothing in flight
ok - fm-turnend-guard: blocks when a fresh beacon has no live watcher lock
ok - fm-turnend-guard: blocks on a dead watcher lock even when the beacon is fresh
ok - fm-turnend-guard: silent no-op with a live watcher lock and fresh beacon
ok - fm-turnend-guard: blocks on a live watcher lock with an ancient beacon
ok - fm-turnend-guard: blocks with the exact required reason in the primary when unhealthy
ok - fm-turnend-guard: blocks from active FM_HOME state, not only repo-root state
ok - fm-turnend-guard: X-mode repair reason sources the cadence config
ok - fm-turnend-guard: ignores stale repo-root state when FM_HOME is set
ok - fm-turnend-guard: uses FM_STATE_OVERRIDE ahead of FM_HOME/state
ok - fm-turnend-guard: stop_hook_active=true always allows the stop (never blocks twice in one turn)
ok - fm-turnend-guard: blocks a blind turn end in a secondmate's own home (.fm-secondmate-home no longer excludes it)
ok - fm-turnend-guard: idle-by-default - silent in a secondmate home with nothing in flight
ok - fm-turnend-guard: stop_hook_active=true allows the stop in a secondmate home (never blocks twice in one turn)
ok - fm-turnend-guard: secondmate deferred-death recovery - silent while watched, forces re-arm once the watcher exits
ok - fm-turnend-guard: inert in a secondmate's own child worktree (linked git worktree) even when unhealthy
ok - fm-turnend-guard: blocks a blind turn end in a treehouse-leased LINKED secondmate home (marker force-include)
ok - fm-turnend-guard: an invalid (empty) marker cannot spoof inclusion; linked worktree stays exempt
ok - fm-turnend-guard: a non-ASCII marker cannot spoof inclusion; linked worktree stays exempt
ok - fm-turnend-guard: inert in a crewmate/scout task worktree (linked git worktree) even when unhealthy
ok - fm-turnend-guard: fails open (never blocks) when jq is missing
ok - fm-turnend-guard: silent no-op on empty stdin
ok - fm-turnend-guard: runs well under the generous timing margin (0s)
ok - fm-turnend-guard-grok: forces one same-session resume when the shared predicate blocks
ok - fm-turnend-guard-grok: loop guard prevents a nested resume loop
ok - .claude/settings.json: Stop hook uses CLAUDE_PROJECT_DIR-anchored command
ok - .codex/hooks.json: Stop hook invokes the shared primary guard
ok - .codex/hooks.json: Stop hook uses hook process root when payload cwd is outside
ok - .codex/hooks.json: Stop hook ignores nested git root guard scripts
ok - .opencode primary plugin: session.idle forces one follow-up through the shared guard
ok - .opencode primary plugin: guard path is anchored to worktree, not directory
ok - .pi primary extension: agent_settled forces one follow-up through the shared guard
ok - .pi primary extension: no-tool and multi-tool runs each inject exactly one guard follow-up
ok - .pi primary extension: delivery failure resets the logical-run latch
ok - .grok primary hook: Stop hook invokes the grok adapter
== repeat tests/fm-pi-watch-extension.test.sh for prior rearm flake ==
-- repeat 1 --
ok - Pi primary watcher extension is tracked, self-hashing, and self-locating
ok - Pi secondmate launch wiring includes both tracked primary extensions
ok - Pi extension reports external healthy watcher output
ok - Pi custom tool returns text content and structured details
ok - Pi watcher arm distinguishes all session lock ownership states
ok - Pi process-exit cleanup listener has a bounded lifecycle
ok - Pi process-exit cleanup stops the attached arm child
ok - OpenCode primary watcher plugin has the verified TUI wake wiring
ok - OpenCode plugins have an explicit ESM boundary even under a typeless parent package
ok - OpenCode watcher plugin uses the effective FM_HOME state
ok - OpenCode watcher plugin sources the effective config
ok - OpenCode watcher plugin requires session lock ownership
ok - OpenCode watcher coordinator respects primary scope
ok - OpenCode watcher plugin rearms after a watcher wake
ok - OpenCode watcher plugin coordinates with the turn-end guard
ok - OpenCode healthy arm output does not suppress the turn-end guard
-- repeat 2 --
ok - Pi primary watcher extension is tracked, self-hashing, and self-locating
ok - Pi secondmate launch wiring includes both tracked primary extensions
ok - Pi extension reports external healthy watcher output
ok - Pi custom tool returns text content and structured details
ok - Pi watcher arm distinguishes all session lock ownership states
ok - Pi process-exit cleanup listener has a bounded lifecycle
ok - Pi process-exit cleanup stops the attached arm child
ok - OpenCode primary watcher plugin has the verified TUI wake wiring
ok - OpenCode plugins have an explicit ESM boundary even under a typeless parent package
ok - OpenCode watcher plugin uses the effective FM_HOME state
ok - OpenCode watcher plugin sources the effective config
ok - OpenCode watcher plugin requires session lock ownership
ok - OpenCode watcher coordinator respects primary scope
ok - OpenCode watcher plugin rearms after a watcher wake
ok - OpenCode watcher plugin coordinates with the turn-end guard
ok - OpenCode healthy arm output does not suppress the turn-end guard
-- repeat 3 --
ok - Pi primary watcher extension is tracked, self-hashing, and self-locating
ok - Pi secondmate launch wiring includes both tracked primary extensions
ok - Pi extension reports external healthy watcher output
ok - Pi custom tool returns text content and structured details
ok - Pi watcher arm distinguishes all session lock ownership states
ok - Pi process-exit cleanup listener has a bounded lifecycle
ok - Pi process-exit cleanup stops the attached arm child
ok - OpenCode primary watcher plugin has the verified TUI wake wiring
ok - OpenCode plugins have an explicit ESM boundary even under a typeless parent package
ok - OpenCode watcher plugin uses the effective FM_HOME state
ok - OpenCode watcher plugin sources the effective config
ok - OpenCode watcher plugin requires session lock ownership
ok - OpenCode watcher coordinator respects primary scope
ok - OpenCode watcher plugin rearms after a watcher wake
ok - OpenCode watcher plugin coordinates with the turn-end guard
ok - OpenCode healthy arm output does not suppress the turn-end guard
```
</details>
<details>
<summary>Evidence: Owned-lock watcher arm invocation</summary>

`arm args=--restart home=/var/folders/5x/4nqprlbx0518k3ybcb1sz6gr0000gn/T/no-mistakes-evidence/01KXRV1BXD7SNEK16C5999DBC4/pi-lock-arm-e2e/home root=/var/folders/5x/4nqprlbx0518k3ybcb1sz6gr0000gn/T/no-mistakes-evidence/01KXRV1BXD7SNEK16C5999DBC4/pi-lock-arm-e2e/repo`

```text
arm args=--restart home=/var/folders/5x/4nqprlbx0518k3ybcb1sz6gr0000gn/T/no-mistakes-evidence/01KXRV1BXD7SNEK16C5999DBC4/pi-lock-arm-e2e/home root=/var/folders/5x/4nqprlbx0518k3ybcb1sz6gr0000gn/T/no-mistakes-evidence/01KXRV1BXD7SNEK16C5999DBC4/pi-lock-arm-e2e/repo
```
</details>
- Outcome: ⚠️ 1 warning across 2 runs (56m41s)

## Pipeline

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>
<summary>⚠️ **Test** - 1 warning</summary>

🔧 Fix: Stabilize watcher extension async waits
1 warning still open:
- ⚠️ I did not run the intent&#39;s lint or TypeScript typecheck portions because this testing prompt explicitly forbids linters and static analysis tools. Behavioral Pi, lock, turn-end, and rearm-flake coverage passed, but green lint/type evidence for this target commit is intentionally missing under that constraint.
- <code>Prompt-provided baseline already completed before this round: `command -v tmux &gt;/dev/null || { echo &#34;tmux is required for e2e tests&#34; &gt;&amp;2; exit 1; }; tmux -V; rc=0; for t in tests/*.test.sh; do echo &#34;== $t ==&#34;; bash &#34;$t&#34; || rc=1; done; exit &#34;$rc&#34;`</code>
- <code>Manual evidence check: imported `.pi/extensions/fm-primary-pi-watch.ts` in a fixture and called `fm_watch_arm_pi` for absent lock, dead holder, live other holder, and owned lock, writing `/var/folders/5x/4nqprlbx0518k3ybcb1sz6gr0000gn/T/no-mistakes-evidence/01KXRV1BXD7SNEK16C5999DBC4/pi-lock-arm-transcript.txt`</code>
- `bash tests/fm-pi-watch-extension.test.sh`
- `bash tests/fm-watcher-lock.test.sh`
- `for i in 1 2 3; do bash tests/fm-pi-watch-extension.test.sh; done`
- `git status --short`

</details>

<details>
<summary>⚠️ **Document** - 1 warning</summary>

- ⚠️ Author intent requires recording the optional session-start work-under-way nudge as a PR follow-up, but no open PR exists for branch `fm/fm-piwatch-arm-stale-lock-msg`; I did not create or modify external GitHub state under this documentation-only scope.

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

### DIFF
--- a/.pi/extensions/fm-primary-pi-watch.ts
+++ b/.pi/extensions/fm-primary-pi-watch.ts
@@ -60,10 +60,6 @@ function lockOwnership(): LockOwnership {
   return pidAlive(lockPid) ? "other" : "missing";
 }
 
-function sessionOwnsLock(): boolean {
-  return lockOwnership() === "owned";
-}
-
 function markLoaded(): void {
   if (lockOwnership() === "other") return;
   mkdirSync(state, { recursive: true });
@@ -104,7 +100,14 @@ export default function (pi: ExtensionAPI) {
   }
 
   function startArm(): ArmResult {
-    if (!sessionOwnsLock()) return { ok: false, message: "watcher: read-only - session lock is held by another firstmate session" };
+    const ownership = lockOwnership();
+    if (ownership === "other") return { ok: false, message: "watcher: read-only - session lock is held by another firstmate session" };
+    if (ownership === "missing") {
+      return {
+        ok: false,
+        message: "watcher: not armed - no live session holds the lock; run bin/fm-session-start.sh to reclaim it, then call fm_watch_arm_pi to re-arm",
+      };
+    }
     markLoaded();
     if (child) return { ok: true, message: "watcher: healthy - Pi extension already has an arm child" };
     const id = ++seq;
--- a/docs/supervision-protocols/pi.md
+++ b/docs/supervision-protocols/pi.md
@@ -6,10 +6,11 @@ When this session owns supervision and away mode is not active:
 3. Arm supervision with the `fm_watch_arm_pi` tool.
    Use `/fm-watch-arm-pi` only as a human-entered fallback.
    Never run `bin/fm-watch-arm.sh` through Pi's bash tool because that foreground arm can wedge the agent and bypasses extension-owned cleanup.
-4. The extension starts `bin/fm-watch-arm.sh --restart`, keeps the child attached to the live Pi process, and sends a follow-up user message when the child exits with an actionable watcher reason.
-5. If the extension says the watcher is already healthy, do not start another cycle.
-6. If the extension reports a watcher failure, drain queued wakes, inspect the failure text, and restart Pi with both extensions loaded if needed.
-7. Never use shell `&` for watcher supervision.
+4. If the extension says no live session holds the lock, run `bin/fm-session-start.sh` to reclaim the session lock, then call `fm_watch_arm_pi` again.
+5. The extension starts `bin/fm-watch-arm.sh --restart`, keeps the child attached to the live Pi process, and sends a follow-up user message when the child exits with an actionable watcher reason.
+6. If the extension says the watcher is already healthy, do not start another cycle.
+7. If the extension reports a watcher failure, drain queued wakes, inspect the failure text, and restart Pi with both extensions loaded if needed.
+8. Never use shell `&` for watcher supervision.
    The arm mechanism above is extension-owned, not a model tool call, but a manual recovery probe that backgrounds, pipes, or bundles the arm is denied automatically by the PreToolUse seatbelt (`bin/fm-arm-pretool-check.sh`, wired into the turn-end guard extension at `__FM_PI_TURNEND_EXT__`).
 
 The turn-end guard extension lives at `__FM_PI_TURNEND_EXT__`.
--- a/tests/fm-pi-watch-extension.test.sh
+++ b/tests/fm-pi-watch-extension.test.sh
@@ -41,12 +41,16 @@ test_tracked_extension_present_and_self_hashing() {
   assert_contains "$text" ".pi-watch-extension-loaded" "tracked extension missing loaded marker"
   assert_contains "$text" 'createHash("sha256").update(readFileSync(extensionFile)).digest("hex")' "tracked extension does not self-hash its own content for extensionVersion"
   assert_contains "$text" 'fileURLToPath(import.meta.url)' "tracked extension does not self-locate via import.meta.url"
-  assert_contains "$text" "sessionOwnsLock" "tracked extension missing session lock ownership check"
   assert_contains "$text" 'type LockOwnership = "owned" | "missing" | "other"' "tracked extension does not distinguish missing lock from another owner"
   assert_contains "$text" "readFileSync(\`\${state}/.lock\`" "tracked extension does not read the effective session lock"
   assert_contains "$text" 'return pidAlive(lockPid) ? "other" : "missing"' "tracked extension does not allow a pre-lock load marker"
   assert_contains "$text" 'if (lockOwnership() === "other") return' "tracked extension overwrites another live session marker"
-  assert_contains "$text" "if (!sessionOwnsLock()) return { ok: false" "tracked extension arms without the session lock"
+  assert_contains "$text" 'const ownership = lockOwnership()' "tracked extension arm does not inspect the distinct lock ownership state"
+  assert_contains "$text" 'if (ownership === "other") return { ok: false' "tracked extension arm does not preserve the live-other read-only refusal"
+  assert_contains "$text" 'if (ownership === "missing")' "tracked extension arm collapses a stale or absent lock into the live-other refusal"
+  assert_contains "$text" "no live session holds the lock" "tracked extension arm missing stale-lock recovery guidance"
+  assert_contains "$text" "run bin/fm-session-start.sh to reclaim it" "tracked extension arm does not direct stale-lock reclamation"
+  assert_contains "$text" "call fm_watch_arm_pi to re-arm" "tracked extension arm does not direct supervision re-arm"
   assert_contains "$text" "writeFileSync(marker, \`\${extensionVersion}\\n\${process.pid}\\n\`)" "tracked extension does not write the content version and process marker"
   assert_contains "$text" "const config = process.env.FM_CONFIG_OVERRIDE" "tracked extension missing effective config resolution"
   assert_contains "$text" "FM_CONFIG_OVERRIDE: config" "tracked extension does not pass the effective config to the watcher arm"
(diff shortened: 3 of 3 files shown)


Kutsal04:
## Post-merge code review

Reviewed at head `351496b`. Line numbers below are against the files as of that commit; I also checked each item against current `main`, and noted where it has since moved.

**The through-line:** the PR splits the arm *message* into three states, but `lockOwnership()` — which it did not touch — only distinguishes two of them reliably. Empty locks and recycled PIDs still land in `"other"` (the state with no recovery path). `EPERM` and every non-`ENOENT` read error land in `"missing"` (the state whose new message tells the operator to take the lock). Because `bin/fm-lock.sh` used a *different* liveness rule than the extension, the new "run `fm-session-start.sh`, then re-arm" guidance also has no termination condition when the two disagree.

The six correctness items below are all still present on `main` — `pidAlive`, the `readFileSync` catch, and `markLoaded`'s recompute are unchanged in shape.

---

### Correctness

**1. An empty or truncated lock is classified `"other"` — the exact stale case this PR exists to fix.**
`.pi/extensions/fm-primary-pi-watch.ts:53`

```ts
if (!/^[0-9]+$/.test(lockPid) || lockPid === "1") return "other";
```

A session killed mid-write of `state/.lock` leaves `lockPid = ""`, the regex fails, and `startArm` returns `"watcher: read-only - session lock is held by another firstmate session"`. The operator is told a live session owns a lock nobody holds, never sees the new `fm-session-start.sh` guidance, and the watcher stays unarmed. Meanwhile `bin/fm-lock.sh` reclaims that same lock immediately (`holder_alive ""` → `kill -0 ""` fails). `tests/fm-watcher-lock.test.sh` already models the empty mid-acquire lock, so the case is known.

**2. The extension's liveness test is a bare `kill(pid, 0)`; `fm-lock.sh` additionally required the holder to look like a harness.**
`.pi/extensions/fm-primary-pi-watch.ts:60` vs `bin/fm-lock.sh:38-43` (at merge)

If a dead holder's PID gets recycled — or lingers as a zombie — `pidAlive` returns true, `lockOwnership()` returns `"other"`, and the arm is refused on every turn, permanently, with no recovery message. `holder_alive()` would have reclaimed it because `ps comm/args` don't match `HARNESS_RE`. This is precisely the class of stale lock the PR set out to route to the new message, and it still misses it.

**3. `pidAlive()` swallows `EPERM`, so a live lock held by another user reads as `"missing"` — and the new message tells the operator to take it.**
`.pi/extensions/fm-primary-pi-watch.ts:41`

On a shared host, `process.kill(otherUsersPid, 0)` throws `EPERM` because the process exists but isn't signalable. The catch returns `false` → `"missing"` → *"no live session holds the lock; run `bin/fm-session-start.sh` to reclaim it"*. Following that breaks the mutual exclusion the lock exists to enforce and lets two live sessions run bootstrap's mutating sweeps concurrently. Before this PR both branches returned the conservative read-only refusal, so the change converts a latent misclassification into a harmful instruction. Worth distinguishing `EPERM` (process exists) from `ESRCH` (it doesn't).

**4. The `readFileSync` catch is unconditional, conflating "cannot read the lock" with "no lock exists."**
`.pi/extensions/fm-primary-pi-watch.ts:50`

`EACCES` (wrong ownership after a sudo-run script), `EISDIR`, or a transient `EMFILE` under load all map to `"missing"`. Under `EACCES` the prescribed recovery also can't write the lock, so the operator runs session-start, re-arms, gets the identical message, and repeats. Only `ENOENT` should map to `"missing"`; every other errno is an unknown state that must not be reported as an absent holder.

**5. Three definitions of "this session owns the lock" now coexist, and the PR reconciled none of them.**
`.pi/extensions/fm-primary-pi-watch.ts:63`

`lockOwnership()` accepts any of 8 ancestors matching `lockPid`; `markLoaded()` writes `process.pid` as the marker; `bin/fm-session-start.sh`'s `pi_extension_loaded()` requires marker\_pid == lock\_pid exactly. If `fm-lock.sh`'s `harness_pid` resolved to an ancestor rather than the node process, the extension arms happily while session-start prints the "not loaded" reminder on every startup. (`bin/fm-session-lock-lib.sh` has since consolidated the shell side — the extension is still the odd copy out.)

**6. The new doc step has no termination condition.**
`docs/supervision-protocols/pi.md:9`

Re-running `bin/fm-session-start.sh` is deterministic: it recomputes the same `harness_pid` and rewrites the same value. If the extension still reports `"missing"`, it's because the two ownership algorithms disagree — divergent `FM_HOME`/`FM_STATE_OVERRIDE` resolution, ancestry deeper than 8 hops, or a harness `comm` that doesn't match. Session-start produces the same lock, the arm produces the same refusal, and the turn-end guard re-triggers every turn. Neither the message nor the doc bounds the loop or says what to do when it doesn't clear.

---

### Altitude / duplication

**7. The ownership decision procedure lives in three places; only the Pi copy was updated.**
`.opencode/plugins/fm-primary-watch-arm.js:113` (`sessionOwnsLock`)

Still a boolean that collapses missing and other into `{ status: "read-only" }` (line 383), and `tests/fm-pi-watch-extension.test.sh` still pins that shape. An OpenCode primary hitting a stale lock gets exactly the misleading refusal this PR just removed from Pi. Per `.agents/skills/firstmate-coding-guidelines/SKILL.md`: *"When you touch a contract, grep the repo for its other mentions and update the cross-references"* and *"review every affected supported primary harness … rather than checking only the adapters active in the current fleet."* Still true on `main`.

**8. `startArm` computes `ownership`, then `markLoaded()` recomputes it from scratch.**
`.pi/extensions/fm-primary-pi-watch.ts:111` (now `:146`/`:320` on `main`)

Each is an ancestry walk that forks `ps` synchronously up to 8 times, on the thread that renders Pi's TUI — up to 16 blocking `spawnSync` calls per arm, repeated on every settled turn. The diff already has the value in hand: pass it as `markLoaded(ownership)`.

**9. The refusal message tells the reader to "call `fm_watch_arm_pi`", but humans see this string too.**
`.pi/extensions/fm-primary-pi-watch.ts:108`, surfaced at `:168` via `ctx.ui.notify(result.message, …)`

`pi.md` step 3 documents `/fm-watch-arm-pi` as the human-entered fallback. A human who follows it against a stale lock is told to call a tool only the model can invoke. The message needs the slash-command spelling, or both.

---

### Tests

**10. The `FM_ARM_LOG` race the "stabilize async waits" commit was meant to fix is still open.**
`tests/fm-pi-watch-extension.test.sh:799`

`d9822ff` bumped the poll ceiling for `FM_GUARD_LOG` (50 → 250) but left the `existsSync(FM_ARM_LOG)` assertion three lines below with no wait of its own. The arm child and the guard child are independent async spawns off the same `session.idle`; when the guard's log lands first — likely, since the guard is a two-line `printf` while the arm goes through `bash -lc` plus config sourcing — the loop exits and the arm log hasn't been written. Flaky `"watch arm did not run"`.

**11. `999999` is not a guaranteed-dead PID.**
`tests/fm-pi-watch-extension.test.sh:252`

`pid_max` is commonly `4194304` on modern Linux and in containers, so `999999` is allocatable. If anything holds it, `pidAlive` returns true → `"other"` → the test fails for an environmental reason. Spawning a child, awaiting its exit, and reusing its now-dead PID makes the precondition guaranteed rather than assumed.

**12. Six new source-text assertions pin implementation shape that the new behavioral test already covers.**
`tests/fm-pi-watch-extension.test.sh:44-53`

`'const ownership = lockOwnership()'`, `'if (ownership === "other") return { ok: false'`, `'if (ownership === "missing")'`, plus three message fragments. Rewriting the branch as a `switch`, extracting the message to a constant, or reordering the guards — all behavior-preserving — breaks `test_tracked_extension_present_and_self_hashing` while `test_pi_arm_distinguishes_session_lock_ownership` still passes. These will block the consolidation items 5 and 7 call for.

**13. The five 50 → 250 bumps widen the window rather than close the race.**
Each loop goes from a 1s to a 5s ceiling; worst case adds ~25s per run. All five races have the same shape — a fixture script writes a log with no completion signal the test can await. Awaiting the spawned child's `close` event removes the race outright, and would have surfaced item 10.

---

### Conventions

**14. `bin/fm-lint.sh` covers `tests/*.sh`, and this PR added ~96 lines to one.**

At merge, `fm-lint.sh:76` was `exec shellcheck --norc bin/*.sh bin/backends/*.sh tests/*.sh`. The PR body records *"lint/type static checks skipped only because this prompt forbids them"* and *"green lint/type evidence for this target commit is intentionally missing"*, while a "✅ Lint — passed" block also appears further down — so the evidence is mixed. Either way the new heredoc-heavy test function merged without a shellcheck run against it, and the guidelines name `bin/fm-lint.sh` as the single owner of that gate.

---

Items 1–4 are the ones I'd fix first: they're small, they're in the function this PR is about, and two of them turn a conservative refusal into either a permanent wedge or an instruction to break the lock.

PR #689, opened 2026-07-18
feat(bin): add a concurrent crew admission cap to fm-spawn.sh

Bre77:
## Intent

Add a concurrent crew admission cap to `bin/fm-spawn.sh`: refuse a NEW ship/scout launch once too many are already live in this home.

This follows a 2026-07-17 investigation into several crews going silent within the same second. The investigation could not prove concurrency caused that incident - it ruled out a per-crew `MemoryMax` kill and a mid-session CLI update as the mechanism, but stopped short of proving the concurrency burst caused the synchronized failure. It did document a real, independent gap: the existing per-crew memory cap bounds one runaway crew's own cgroup, not how many crews can be live at once on a fixed-RAM host, and a burst of 9 concurrent launches (several `opus`/`high-effort`) in the 70 minutes before the event is exactly the kind of oversubscription that gap allows. The captain's ruling was to adopt a worker limit regardless of final root cause, explicitly as resource hygiene rather than a proven fix for that incident.

## What Changed

- `bin/fm-spawn.sh`: before a ship/scout AGENT launches, count live `state/*.meta` files recording `kind=ship` or `kind=scout` in this home and refuse the spawn once the count reaches the cap, naming the live count, the cap, and both override knobs. `--secondmate` launches are exempt (persistent supervisors, not burst load), and a task relaunching its own id excludes its own prior meta record so recovery relaunches are never self-blocked.
- Config: local, gitignored `config/max-crew` (one line, an integer), or the `FM_MAX_CREW` env var, which wins over the file. Default cap is 6, chosen as a middle ground on this host's 30GB RAM / 9GB swap - comfortably below the 9-launch burst the investigation observed, while `MemoryHigh`'s soft-throttle semantics mean most crews never actually consume their full per-crew share.
- `docs/crew-memory-cap.md` and `docs/configuration.md` document the new knob and the aggregate gap it closes, per the one-owner rule.
- `tests/fm-crew-admission-cap.test.sh` covers admission/refusal at the cap boundary, scout tasks counting toward the same cap, secondmate exclusion (both as an uncounted kind and as an exempt launch), self-exclusion on relaunch, and both override knobs.

## Testing

- `bin/fm-lint.sh` (shellcheck 0.11.0, pinned) - clean
- `bash tests/fm-crew-admission-cap.test.sh` - all 8 cases pass
- Existing `fm-crew-memory-cap`, `fm-spawn-dispatch-profile`, `fm-spawn-batch`, `fm-instruction-owners`, and `fm-lint` suites - all still pass

This is a direct-PR delivery (fork-only), so the no-mistakes pipeline did not run.


Bre77:
Opened by mistake (gh pr create defaulted to the fork's upstream base). Closing - the actual PR is against Bre77/firstmate.

PR #716, opened 2026-07-19
fix(bin): make fm-lock.sh harness detection and liveness work on Windows

kgrittner:
## Problem

On Windows (Git Bash / MSYS / Cygwin), `fm-lock.sh` could never find the harness process:

- The bundled MSYS `ps` only sees MSYS processes, while the harness (e.g. `claude.exe`) runs as a native Windows process, so the ancestry walk bottomed out at ppid 1 before reaching it.
- Liveness checks used `kill -0`, which operates on MSYS pids and misjudges native Windows pids.

## Fix

On MINGW/MSYS/CYGWIN:

- Climb the MSYS side first via `/proc/<pid>/ppid` (MSYS emulates fork with transient stubs, so native parent pids of nested shells may already be dead, while /proc tracks the logical parent correctly), then switch to the native side via `/proc/<pid>/winpid`.
- Continue the ancestry walk with a single `powershell.exe` CIM query loop following `ParentProcessId`. The pid and harness regex ride in as env vars so no bash-side escaping can corrupt the script, and `[int]` casts make a garbage pid a hard error rather than a WQL injection.
- Answer liveness through CIM as well, since the pid stored in the lock is now a native Windows pid.

Non-Windows platforms keep the existing `ps`-based walk unchanged.

## Testing

Verified end-to-end on Windows 11 (Git Bash under Claude Code):

- `fm-lock.sh` acquires the lock with the real native pid of `claude.exe` (confirmed via CIM that the detected pid is the harness, not an MSYS stub).
- `fm-lock.sh status` reports the holder as live; re-acquiring as the same holder succeeds.
- Stale-lock takeover: with a dead pid planted in an isolated `FM_STATE_OVERRIDE` dir, status reports it stale and acquire takes over.
- `bash -n` passes on all scripts.

### DIFF
--- a/bin/fm-lock.sh
+++ b/bin/fm-lock.sh
@@ -17,7 +17,70 @@ mkdir -p "$STATE"
 # Known harness command names; extend when a new adapter is verified.
 HARNESS_RE='claude|codex|opencode|grok|^pi$'
 
+# Windows (Git Bash / MSYS / Cygwin): the bundled MSYS ps has no -o support and
+# only sees MSYS processes, while the harness runs as a NATIVE Windows process
+# (e.g. claude.exe) that the MSYS ancestry never reaches - its view bottoms out
+# at ppid 1. So on these platforms the walk starts from this shell's native
+# Windows pid (/proc/$$/winpid) and follows ParentProcessId through one CIM
+# query loop. The pid written to the lock is then a native Windows pid, and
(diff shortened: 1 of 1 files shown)


kgrittner:
Withdrawing this - keeping the fix on my fork instead.

PR #725, opened 2026-07-19
fix(supervision): relocate long-lived processes out of disposable worktrees

thevwick:
## Intent

Add a guard that refuses to start a long-lived firstmate process from a disposable treehouse worktree, turning a silent failure into a loud one.

Background: treehouse's 'terminate lingering processes' step kills every process whose current working directory resolves under the worktree being returned. It matches on filesystem path, not process ancestry, so a process started by a completely different session dies when somebody returns an unrelated worktree. The signal is SIGURG (16), surfacing as exit 144. This repeatedly killed firstmate's supervision watcher, because the shell working directory persists between commands: once firstmate had cd'd into a worktree, everything launched afterwards inherited that doomed path. This was root-caused and reproduced three ways in a prior scout investigation (data/proc-reaper-scout-r7/report.md); the task is to implement its recommendation 2, not to re-investigate.

Deliberate decisions made while doing the work, which a reviewer reading only the diff would not know:

1. REFUSE rather than auto-relocate the cwd. This was an explicit judgment call left open by the task. Refusing is not merely more predictable: auto-correcting would be actively wrong, because the arm process is not the only process that must survive. The harness's own wrapper shell is the arm's parent and keeps the condemned cwd, so an internal 'cd $FM_HOME' would produce a watcher that still dies while looking safe. That is strictly worse than the current silent failure.

2. The check lives in the LAUNCHERS (bin/fm-watch-arm.sh), not in the PreToolUse guard (bin/fm-arm-command-policy.mjs / bin/fm-arm-pretool-check.sh). This was the other explicit judgment call. The pretool policy is a text classifier whose stated contract is that it never inspects the environment the command would run in, and it fails open on missing node/jq and on harnesses without PreToolUse hooks. Working directory is a runtime property invisible to command text. Duplicating the pool rule there would also violate the repo's one-owner rule and create a second copy that drifts. I added exactly one line to docs/arm-pretool-check.md marking that boundary so a future contributor does not add it there.

3. The rule is 'cwd is under the treehouse pool root AND is not inside FM_HOME', not simply 'cwd is under the pool'. This matters: a durably leased secondmate home IS itself a treehouse pool slot (bin/fm-home-seed.sh acquires homes via 'treehouse get --lease'), and is stable for that home's lifetime. A naive pool-only rule would have broken every secondmate's watcher. FM_HOME containment is the discriminator between a stable leased home and a disposable per-task slot.

4. The task asked me to consider whether the same check belongs on any other long-lived launcher and to say so either way. I found one genuine second instance and extended to it: bin/fm-afk-start.sh runs the away-mode supervision daemon with the identical lifecycle shape. Its consequence is worse, since away mode means nobody is watching, so it refuses BEFORE entering away mode rather than leaving state/.afk set with no daemon to own it. I deliberately did NOT guard bin/fm-watch.sh itself, because it is only ever forked by the already-guarded arm.

5. To honor the one-owner rule the check is factored into bin/fm-wake-lib.sh (fm_cwd_is_disposable_slot + fm_disposable_cwd_banner) rather than copied into both launchers. fm-afk-start.sh already sourced that lib, so this needed no new plumbing.

6. Test-harness change worth understanding: the new guard immediately fired on the PRE-EXISTING watcher tests, which arm from the repo checkout while pointing FM_HOME at a temp dir. That is genuinely the hazard shape, so those tests were latently wrong. I pinned TREEHOUSE_DIR to a non-existent path for that suite, which makes the guard inert for tests that are not about it and makes the suite behave identically in CI and inside a task worktree. The two tests that ARE about the guard set TREEHOUSE_DIR themselves.

Repo constraints that were enforced: load the firstmate-coding-guidelines skill first; apply its knowledge-placement decision tree so mechanics live in script headers rather than AGENTS.md (AGENTS.md was deliberately left untouched, as the task anticipated); one full sentence per line in tracked Markdown; plain dash and never an em dash, including in code comments and output strings; shellcheck must pass via bin/fm-lint.sh; tests colocated in tests/ extending existing runners rather than adding a new one; no agent name as a commit co-author.

Verification already done locally: bin/fm-lint.sh clean; tests/fm-watcher-lock.test.sh 28/28 across two full runs; tests/fm-afk-launch.test.sh passing including three new assertions; ten related suites passing. One flake was observed once in test_watcher_self_evicts_on_lock_takeover, which does not touch the arm path and passed on subsequent runs.

## What Changed

- Added `fm_cwd_is_disposable_slot` / `fm_relocate_from_disposable_cwd` to `bin/fm-wake-lib.sh`, which detect a working directory under the treehouse pool root but outside `FM_HOME` (so a durably leased secondmate home is exempt) and move the process into `FM_HOME` before it forks, with a loud stderr banner and a distinct typed error on chdir failure. Treehouse's path-matching process sweep otherwise killed watchers with SIGURG whenever an unrelated worktree was returned.
- Wired the relocation into every long-lived launcher: `bin/fm-watch.sh` (below its source guard, so sourcing for unit tests does not relocate the test shell), `bin/fm-watch-arm.sh`, `bin/fm-watch-checkpoint.sh`, and `bin/fm-afk-start.sh`. Path derivation was centralized (`fm_derive_paths`, `watch_derive_state_paths`) so STATE-derived paths are re-anchored by construction, and only genuinely relative `FM_HOME`/`STATE`/`FM_ROOT` values are rewritten, preserving `FM_HOME` as a byte-exact cross-process lock identity.
- Added `tests/fm-watcher-lock.test.sh` relocation cases, `tests/fm-afk-launch.test.sh`, `tests/fm-watch-checkpoint.test.sh` coverage, and `tests/fm-relocation-pool-pin.test.sh`, a static contract that every launcher-touching suite pins its pool root; `tests/lib.sh` now owns the shared pin so suites behave identically in CI and inside a task worktree. Docs updated in `docs/architecture.md`, `docs/configuration.md`, `docs/scripts.md`, `docs/watcher-continuity.md`, `docs/arm-pretool-check.md` (marking the boundary against duplicating this rule in the pretool policy), and `CONTRIBUTING.md`.

## Risk Assessment

✅ Low: The guard is well-bounded and self-correcting, the FM_HOME byte-identity invariant that other processes depend on is explicitly preserved and covered by a dedicated test, every path that executes the watcher now routes through a single choke point below the source guard, and the only remaining items are two narrow test-infrastructure gaps that cannot affect production behavior.

## Testing

On top of the already-green configured suite run, I ran the four suites that cover this change (watcher lock, away-mode launch, watcher checkpoint, and the new pool-pin contract test) and all passed. Because passing tests alone would not show the actual hazard, I also built a manual end-to-end simulation of treehouse's worktree-return sweep: with the guard unable to see the pool the watcher's working directory is the disposable slot, the sweep matches it and supervision dies; with the guard active the launcher prints its relocation banner, the watcher starts with its working directory in the home, and the sweep matches nothing so supervision survives. The away-mode launcher shows the same banner from a disposable slot and correctly stays silent when started from a durably leased secondmate home. One caveat recorded in the transcript: this bash build ignores SIGURG, so the demo terminates the same cwd-matched process set with a lethal signal, since the mechanism under test is the working-directory match rather than the signal number. No UI surface is involved, so the reviewer-visible evidence is CLI transcripts rather than screenshots.

<details>
<summary>Evidence: End-to-end treehouse sweep simulation: watcher dies without the guard, survives with it</summary>

<code>=== firstmate supervision watcher vs the treehouse worktree sweep ===&#10;&#10;--- without-guard&#10;captain runs the arm from a task worktree: cd .../pool/firstmate-abc123/3/firstmate&#10;| watcher: started pid=93862 (beacon fresh)&#10;supervision watcher pid 93862, working directory: .../pool/firstmate-abc123/3/firstmate&#10;an unrelated session now returns that worktree: treehouse return&#10;sweep: MATCHED pid 93836 (working directory is under the returned slot) -&gt; terminating&#10;sweep: MATCHED pid 93862 (working directory is under the returned slot) -&gt; terminating&#10;RESULT: supervision watcher DEAD - fleet silently unsupervised&#10;&#10;--- with-guard&#10;captain runs the arm from a task worktree: cd .../pool/firstmate-abc123/3/firstmate&#10;| ●━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━&#10;| ● RELOCATED THE WATCHER OUT OF A DISPOSABLE TASK WORKTREE&#10;| ● was: .../pool/firstmate-abc123/3/firstmate&#10;| ● now: .../home&#10;| ● A long-lived process started from a treehouse pool slot is killed by&#10;| ● signal 16 (exit 144) the moment any session returns that slot, because&#10;| ● the sweep matches on working directory, not on who started the process.&#10;| ● This process moved to the home before starting anything long-lived, so&#10;| ● neither it nor its children are in the sweep path. Nothing to do.&#10;| ●━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━&#10;| watcher: started pid=95468 (beacon fresh)&#10;supervision watcher pid 95468, working directory: .../home&#10;an unrelated session now returns that worktree: treehouse return&#10;sweep: MATCHED nothing - no firstmate process has its working directory under the returned slot&#10;RESULT: supervision watcher still running (pid 95468) - fleet stays supervised</code>

```text
=== firstmate supervision watcher vs the treehouse worktree sweep ===

--- without-guard
    captain runs the arm from a task worktree:  cd /var/folders/7t/wl9fn2g17xd8lsfw6rhmbb2w0000gn/T/tmp.QvCUkSiTtT/without-guard/pool/firstmate-abc123/3/firstmate
    | watcher: started pid=93862 (beacon fresh)
    supervision watcher pid 93862, working directory: /private/var/folders/7t/wl9fn2g17xd8lsfw6rhmbb2w0000gn/T/tmp.QvCUkSiTtT/without-guard/pool/firstmate-abc123/3/firstmate
    an unrelated session now returns that worktree:  treehouse return
    sweep: MATCHED pid 93836 (working directory is under the returned slot) -> terminating
    sweep: MATCHED pid 93862 (working directory is under the returned slot) -> terminating
    RESULT: supervision watcher DEAD (exit 127) - fleet silently unsupervised

--- with-guard
    captain runs the arm from a task worktree:  cd /var/folders/7t/wl9fn2g17xd8lsfw6rhmbb2w0000gn/T/tmp.QvCUkSiTtT/with-guard/pool/firstmate-abc123/3/firstmate
    | ●━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
    | ●  RELOCATED THE WATCHER OUT OF A DISPOSABLE TASK WORKTREE
    | ●  was:     /private/var/folders/7t/wl9fn2g17xd8lsfw6rhmbb2w0000gn/T/tmp.QvCUkSiTtT/with-guard/pool/firstmate-abc123/3/firstmate
    | ●  now:     /private/var/folders/7t/wl9fn2g17xd8lsfw6rhmbb2w0000gn/T/tmp.QvCUkSiTtT/with-guard/home
    | ●  A long-lived process started from a treehouse pool slot is killed by
    | ●  signal 16 (exit 144) the moment any session returns that slot, because
    | ●  the sweep matches on working directory, not on who started the process.
    | ●  This process moved to the home before starting anything long-lived, so
    | ●  neither it nor its children are in the sweep path. Nothing to do.
    | ●━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
    | watcher: started pid=95468 (beacon fresh)
    supervision watcher pid 95468, working directory: /private/var/folders/7t/wl9fn2g17xd8lsfw6rhmbb2w0000gn/T/tmp.QvCUkSiTtT/with-guard/home
    an unrelated session now returns that worktree:  treehouse return
    sweep: MATCHED nothing - no firstmate process has its working directory under the returned slot
    RESULT: supervision watcher still running (pid 95468) - fleet stays supervised
```
</details>
<details>
<summary>Evidence: Away mode relocates from a disposable slot but leaves a leased secondmate home alone</summary>

<code>=== away mode armed from a disposable task worktree ===&#10;●━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━&#10;● RELOCATED AWAY MODE OUT OF A DISPOSABLE TASK WORKTREE&#10;● was: .../pool/firstmate-abc/3/firstmate&#10;● now: .../home&#10;● ... This process moved to the home before starting anything long-lived, so&#10;● neither it nor its children are in the sweep path. Nothing to do.&#10;●━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━&#10;afk: starting supervise daemon in foreground; keep this command as a tracked background session&#10;&#10;=== away mode armed from a durably LEASED secondmate home (itself a pool slot) ===&#10;afk: starting supervise daemon in foreground; keep this command as a tracked background session&#10;(no relocation banner above = a leased home is correctly left alone)</code>

```text
=== away mode armed from a disposable task worktree ===
  ●  RELOCATED AWAY MODE OUT OF A DISPOSABLE TASK WORKTREE
  ●  was:     /private/var/folders/7t/wl9fn2g17xd8lsfw6rhmbb2w0000gn/T/tmp.dH66Q58CAI/pool/firstmate-abc/3/firstmate
  ●  now:     /private/var/folders/7t/wl9fn2g17xd8lsfw6rhmbb2w0000gn/T/tmp.dH66Q58CAI/home
  ●  A long-lived process started from a treehouse pool slot is killed by
  ●  signal 16 (exit 144) the moment any session returns that slot, because
  ●  the sweep matches on working directory, not on who started the process.
  ●  This process moved to the home before starting anything long-lived, so
  ●  neither it nor its children are in the sweep path. Nothing to do.
  afk: starting supervise daemon in foreground; keep this command as a tracked background session
  warn: could not auto-discover supervisor pane (no FM_SUPERVISOR_TARGET, TMUX_PANE, or HERDR_ENV/HERDR_PANE_ID); falling back to 'firstmate:0' — verify this is firstmate's pane
  error: supervisor target 'firstmate:0' does not resolve to a tmux pane; set FM_SUPERVISOR_TARGET

=== away mode armed from a durably LEASED secondmate home (itself a pool slot) ===
  afk: starting supervise daemon in foreground; keep this command as a tracked background session
  warn: could not auto-discover supervisor pane (no FM_SUPERVISOR_TARGET, TMUX_PANE, or HERDR_ENV/HERDR_PANE_ID); falling back to 'firstmate:0' — verify this is firstmate's pane
  error: supervisor target 'firstmate:0' does not resolve to a tmux pane; set FM_SUPERVISOR_TARGET
  (no relocation banner above = a leased home is correctly left alone)
```
</details>
<details>
<summary>Evidence: Reusable sweep simulation script</summary>

```text
#!/usr/bin/env bash
# End-to-end: does firstmate's supervision watcher survive treehouse's
# lingering-process sweep when armed from a disposable pool slot?
set -u
ROOT=$1
BASE=$(mktemp -d)
PIDS=()
trap 'for p in "${PIDS[@]:-}"; do kill "$p" 2>/dev/null; kill -9 "$p" 2>/dev/null; done; rm -rf "$BASE"' EXIT

proc_cwd() { lsof -a -p "$1" -d cwd -Fn 2>/dev/null | sed -n 's/^n//p' | head -1; }

# Stand in for treehouse's "terminate lingering processes" step: it matches every
# process whose working directory resolves under the slot being returned, then
# terminates it. The real sweep signals SIGURG (exit 144); this bash build ignores
# SIGURG, so the demo sends a lethal signal to the SAME matched set. The mechanism
# under test is the working-directory match, not the choice of signal.
treehouse_sweep() {
  local slot=$1 pid matched=0
  for pid in $(pgrep -f 'fm-watch|fm-watch-arm' 2>/dev/null); do
    case "$(proc_cwd "$pid")" in
      "$slot"|"$slot"/*)
        echo "    sweep: MATCHED pid $pid (working directory is under the returned slot) -> terminating"
        kill -URG "$pid" 2>/dev/null; kill -TERM "$pid" 2>/dev/null; matched=1 ;;
    esac
  done
  [ "$matched" = 1 ] || echo "    sweep: MATCHED nothing - no firstmate process has its working directory under the returned slot"
}

run_case() {
  local label=$1 mode=$2 pool_pin dir pool slot home state armpid wpid i
  dir="$BASE/$label"; pool="$dir/pool"; slot="$pool/firstmate-abc123/3/firstmate"
  home="$dir/home"; state="$home/state"
  mkdir -p "$slot" "$state"
  if [ "$mode" = active ]; then pool_pin="$pool"; else pool_pin="$dir/no-such-pool"; fi
  printf 'fm-pr-check-migration-scan-v1\n' > "$state/.pr-check-migration-scan-v1"
  printf 'fm-pr-check-migration-v1\n' > "$state/.pr-check-migration-v1"
  chmod 0600 "$state/.pr-check-migration-scan-v1" "$state/.pr-check-migration-v1"

  echo "--- $label"
  echo "    captain runs the arm from a task worktree:  cd $slot"
  ( cd "$slot" && FM_TREEHOUSE_POOL_ROOT="$pool_pin" FM_HOME="$home" FM_STATE_OVERRIDE="$state" \
      FM_POLL=2 FM_SIGNAL_GRACE=1 FM_CHECK_INTERVAL=999999 FM_HEARTBEAT=999999 \
      FM_ARM_CONFIRM_TIMEOUT=20 "$ROOT/bin/fm-watch-arm.sh" ) > "$dir/arm.out" 2>&1 &
  armpid=$!; PIDS+=("$armpid")

  for i in $(seq 1 60); do
    wpid=$(cat "$state/.watch.lock/pid" 2>/dev/null || true)
    [ -n "${wpid:-}" ] && kill -0 "$wpid" 2>/dev/null && break
    sleep 0.5
  done
  [ -n "${wpid:-}" ] || { echo "    watcher never started"; sed 's/^/    | /' "$dir/arm.out"; return 1; }
  PIDS+=("$wpid")

  sed 's/^/    | /' "$dir/arm.out"
  echo "    supervision watcher pid $wpid, working directory: $(proc_cwd "$wpid")"
  echo "    an unrelated session now returns that worktree:  treehouse return"
  treehouse_sweep "$(cd "$slot" && pwd -P)"
  sleep 2
  if kill -0 "$wpid" 2>/dev/null; then
    echo "    RESULT: supervision watcher still running (pid $wpid) - fleet stays supervised"
  else
    wait "$wpid" 2>/dev/null; echo "    RESULT: supervision watcher DEAD (exit $?) - fleet silently unsupervised"
  fi
  kill "$armpid" "$wpid" 2>/dev/null
  echo
}

echo "=== firstmate supervision watcher vs the treehouse worktree sweep ==="
echo
run_case without-guard inert    # guard cannot see the pool: pre-change behaviour
run_case with-guard    active   # guard active
```
</details>

## Pipeline

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>

- 🚨 `bin/fm-wake-lib.sh:115` - The refusal banner prints `cd &lt;home&gt; &amp;&amp; &lt;launcher&gt;` as the fix, but docs/cd-guard.md names `cd x &amp;&amp; cmd` as a blocked shape (`persistent-cd`), and the arm seatbelt denies the compliant subshell form `(cd home &amp;&amp; ...)` as `watcher-nested`. On a primary running both PreToolUse hooks neither form executes. Combined with bin/fm-continuity-pretool-check.sh, which permits only fm-wake-drain.sh and fm-watch-arm.sh while tasks are in flight and no live watcher holds the lock, a home that trips this guard has no allowed command that resolves it. Suggest phrasing the remedy the way bin/fm-guard.sh does (a location-independent invocation, or relaunching the harness background task with an explicit working directory), and updating docs/cd-guard.md line 53, whose &#39;steady state is always at the home&#39; premise this guard&#39;s existence falsifies.
- ⚠️ `bin/fm-afk-start.sh:122` - The comment claims the guard refuses &#39;before touching state, so away mode is never entered without a daemon&#39;, but neither documented entry path reaches this code before state is written. `fm_afk_launch_start_native` (bin/fm-afk-launch.sh:507) writes state/.afk and records the native terminal, then the agent separately invokes `FM_AFK_STATE_PREPARED=1 bin/fm-afk-start.sh`, which refuses with no automatic rollback - away mode is live with no daemon, recoverable only via the skill&#39;s manual `fm-afk-launch.sh stop`. On the terminal-backed `start` path the refusal is printed inside a non-visible detached terminal that is then closed, so the operator sees only &#39;daemon did not become ready&#39; and the loud failure becomes silent. Both launcher entrypoints run in the captain&#39;s shell with the same cwd, so the check belongs at the top of `fm_afk_launch_start` and `fm_afk_launch_start_native` as well.
- ℹ️ `bin/fm-wake-lib.sh:80` - `fm_resolve_dir` runs a bare `cd` in the caller&#39;s shell; it is only safe today because both call sites in `fm_cwd_is_disposable_slot` wrap it in command substitution. It is defined in a lib sourced by many scripts under a generic name, so a future direct call would silently relocate the calling process. Wrap the body in a subshell: `( CDPATH=&#39;&#39; cd -- &#34;$path&#34; 2&gt;/dev/null &amp;&amp; pwd -P )`.

🔧 Fix: relocate long-lived processes out of disposable worktrees
- ⚠️ `bin/fm-wake-lib.sh:120` - The relocation rewrites FM_HOME to the `pwd -P` resolved path (`FM_HOME=$FM_DISPOSABLE_HOME`). FM_HOME is not only a path here: fm-watch.sh:700 writes it verbatim into `.watch.lock/fm-home`, and fm_watcher_lock_matches_pid, fm-turnend-guard.sh:86, fm-continuity-pretool-check.sh:84, fm-pr-check-migrate.sh:269 and fm-watch-arm.sh:245 all compare it with a byte-exact `[ &#34;$lock_home&#34; = &#34;$home&#34; ]`. Every other process derives FM_HOME logically (`cd .. &amp;&amp; pwd` in the lib, or the exported env value), so if any component of the home path is a symlink the two strings diverge. Two concrete breakages: (a) FM_HOME unexported (primary case) - the forked watcher records the logical path while the relocated arm compares the physical one, so the arm&#39;s post-fork `fm_watcher_healthy` at fm-watch-arm.sh:245 never matches and the arm reports &#34;no live watcher with a fresh beacon&#34; and retries against a perfectly healthy watcher; (b) FM_HOME exported (secondmate case) - the assignment updates the exported value, so the child records the physical path while the guard, checkpoint and continuity check still hold the logical one and conclude the lock is not this home&#39;s. The re-anchoring is also unnecessary in practice: FM_HOME is already absolute in every path that sets it. Restrict the rewrite to the genuinely-relative case (leave the original string untouched when it already starts with `/`), and do the same for STATE/FM_STATE_OVERRIDE.
- ⚠️ `bin/fm-watch-checkpoint.sh:78` - bin/fm-watch-checkpoint.sh runs `$SCRIPT_DIR/fm-watch.sh` in the foreground for a bounded checkpoint and never calls fm_relocate_from_disposable_cwd, so it inherits whatever cwd the caller had. Per .agents/skills/harness-adapters/SKILL.md:184 this is Codex&#39;s PRIMARY watcher protocol, not fm-watch-arm.sh, so a Codex home whose shell has cd&#39;d into a task worktree gets no protection at all: every checkpoint runs a watcher in the condemned cwd and is killed by signal 16 mid-cycle whenever any session returns that slot, surfacing as a repeatedly-failing checkpoint. The change&#39;s reasoning excluded fm-watch.sh on the grounds that it is &#34;only ever forked by the already-guarded arm&#34;, which holds for the claude/grok path but not for this one. bin/fm-supervise-daemon.sh:1399 is genuinely covered, since it is reached only through the now-guarded fm-afk-start.sh.
- ⚠️ `tests/fm-watcher-lock.test.sh:104` - test_relocated_watcher_does_not_run_in_condemned_cwd forks a real arm (which forks a real fm-watch.sh) in the background and only kills them on the success path at the end. tests/lib.sh:44 defines `fail` as `exit 1`, so the `fail &#34;relocated arm never forked a watcher that took the lock&#34;` at line 106 and the `fail &#34;watcher inherited the condemned cwd&#34;` inside the case at line 110 both exit the suite with the arm and watcher still running. fm_test_cleanup only removes directories, so the leaked watcher keeps polling every FM_POLL=5 seconds forever against a state dir that has just been deleted. The two `sleep 300 &amp;` sleepers added at lines 58 and 136 leak the same way. Register the pids and `trap &#39;kill $PIDS 2&gt;/dev/null&#39; EXIT` at launch rather than killing at the end of the happy path.
- ℹ️ `tests/fm-watcher-lock.test.sh:114` - When lsof produces no output the test calls `pass` with &#34;cwd unreadable on this host, skipped&#34;, so the assertion the test exists to make silently degrades to a no-op that still reports ok. lsof also returns nothing when the watcher has already exited, which is a real failure mode rather than a host capability gap, so a genuine regression could be reported as a pass. A `skip` reporter, or at least asserting the watcher pid is still alive before treating empty output as unreadable, would keep the signal honest.

🔧 Fix: guard every watcher launcher and preserve home identity
- ⚠️ `bin/fm-wake-lib.sh:140` - `state_abs=$(fm_resolve_dir &#34;$STATE&#34;) || return 1` runs before the chdir and on every relocation, including the overwhelmingly common case where STATE is already absolute and the value is never used. Its failure is then reported by all four callers as &#34;cannot move out of the disposable task worktree ... into the home&#34;, which names the wrong cause: STATE failing to resolve (the lib&#39;s source-time `mkdir -p &#34;$STATE&#34;` silently failed on a read-only or full filesystem, or FM_STATE_OVERRIDE points at a path that was removed between source time and the call) aborts the launcher with a message about the home chdir. Move the resolution inside the non-absolute branch of the `case &#34;$STATE&#34;` at line 147, so the only failure mode reaching `return 1` is a genuine chdir failure and the caller&#39;s message is always accurate.
- ⚠️ `tests/fm-watch-checkpoint.test.sh:10` - tests/fm-watcher-lock.test.sh:24 exports TREEHOUSE_DIR to a non-existent path with the stated rationale that these suites run launchers from the repo checkout while pointing FM_HOME at a temp dir, which is exactly the condemned shape when the checkout is itself a pool slot. That rationale applies verbatim to tests/fm-watch-checkpoint.test.sh and tests/fm-afk-launch.test.sh, which now invoke the same relocating launchers and are not pinned. Their pre-existing cases (test_quiet_checkpoint_exits_124_cleanly, test_registered_check_uses_preserved_watcher_environment, test_existing_singleton_watcher_is_not_success, and every afk-launch case that runs $START) will relocate and emit the banner when the suite is run from inside a task worktree but not in CI. The assertions are all `assert_contains`, so nothing fails today, but the suites are environment-dependent for no reason and any future exact-output or empty-stderr assertion silently becomes CI-only. Pin TREEHOUSE_DIR to an absent path at the top of both suites, exactly as the watcher-lock suite does, since the cases that exercise the guard already set it themselves.
- ℹ️ `bin/fm-watch-checkpoint.sh:10` - bin/fm-wake-lib.sh is sourced at the top of the script, before argument parsing, and the lib performs `mkdir -p &#34;$STATE&#34;` at source time. `fm-watch-checkpoint.sh --help` and the two argument-error exits now create $FM_HOME/state as a side effect where they previously did not. Harmless in practice since any real checkpoint creates it anyway, but sourcing after the argument-parsing loop and before the relocation call would keep the informational paths side-effect free.

🔧 Fix: report relocation faults distinctly and pin test pool root
- ⚠️ `tests/fm-watch-triage.test.sh:39` - Round 3 pinned TREEHOUSE_DIR in the three suites that drive the launchers, but the guard has since moved into bin/fm-watch.sh itself, which pulls in every suite that executes the watcher directly. tests/fm-watch-triage.test.sh:39/387/428, tests/fm-wake-queue.test.sh:63/71/99/137/165 and tests/fm-pr-check-security.test.sh:587/2318/2339 all run &#34;$WATCH&#34; from the repo checkout while pointing FM_HOME at a temp dir outside the pool - exactly the shape fm_cwd_is_disposable_slot condemns when the checkout is a treehouse slot, which it is for any crewmate task worktree. Those runs will relocate to FM_HOME and emit the banner on stderr, so the watcher&#39;s cwd and the suites&#39; stderr differ between CI and a task worktree. Nothing fails today only because every assertion in these suites reads stdout; any future exact-output or empty-stderr assertion silently becomes CI-only. Apply the same TREEHOUSE_DIR pin and explanatory comment used at tests/fm-watcher-lock.test.sh:22 to these three suites.
- ℹ️ `bin/fm-wake-lib.sh:142` - fm_relocate_from_disposable_cwd re-anchors a relative FM_HOME and a relative STATE/FM_STATE_OVERRIDE before the chdir, but leaves FM_ROOT and FM_ROOT_OVERRIDE untouched. bin/fm-watch.sh:46 takes FM_ROOT=&#34;${FM_ROOT_OVERRIDE:-...}&#34; verbatim, so a relative override stays relative and silently resolves against the wrong directory after relocation, and the value is propagated to child processes. Only reachable when a caller sets a relative FM_ROOT_OVERRIDE, which today is a test-only shape, so this is a completeness gap in the re-anchoring rather than a live fault. Either re-anchor it alongside STATE using the same relative-only rule, or state in the header comment that FM_ROOT_OVERRIDE is required to be absolute.

🔧 Fix: re-anchor code root and centralize test pool pin
- ⚠️ `bin/fm-watch.sh:55` - The relocation call sits at top level, above the main-entry source guard at line 666. bin/fm-watch.sh is documented (lines 91-96) as sourceable for unit tests, with the contract that sourcing &#34;loads the functions ... and returns before acquiring the lock or starting the loop&#34; and that running it as a script behaves &#34;exactly as before, byte-for-byte&#34;. Both are now untrue when sourced: the sourcing shell&#39;s cwd is silently changed, and on a chdir failure the `exit 1` at line 58 terminates the sourcing test suite rather than returning, bypassing that suite&#39;s cleanup. tests/fm-supervision-events.test.sh:23 and tests/fm-backend-herdr-eventwait-smoke.test.sh:119 both source it. Moving the relocation inside the main-entry guard (or gating it on `[ &#34;${BASH_SOURCE[0]}&#34; = &#34;$0&#34; ]`) restores the documented contract while still relocating before the loop; note WATCH_LOCK at line 89 is derived from STATE, so if you move it, re-derive WATCH_LOCK after.
- ⚠️ `tests/fm-backend-herdr-eventwait-smoke.test.sh:119` - Round 4 made tests/lib.sh:48 the single owner of the TREEHOUSE_DIR pin, and tests/fm-afk-launch.test.sh:36 was given its own copy precisely because it is standalone and never sources that library. tests/fm-backend-herdr-eventwait-smoke.test.sh is the other standalone suite (it defines its own ROOT/fail/pass at lines 16-18 and never sources tests/lib.sh) and it sources bin/fm-watch.sh at line 119, which now relocates on source. It sets FM_ROOT_OVERRIDE but no FM_HOME, so FM_HOME defaults to $ROOT; any invocation whose cwd is under the treehouse pool but outside that root relocates the test shell mid-suite, and a chdir failure exits it before cleanup_all runs. Apply the same pin and rationale comment already carried by tests/fm-afk-launch.test.sh:24-36.
- ℹ️ `bin/fm-watch-arm.sh:76` - The relocation runs at line 76, but bin/fm-watch-arm.sh does not parse its arguments until line 342, so `fm-watch-arm.sh --bogus` now performs the chdir and prints the full relocation banner before emitting `usage:` and exiting 2. bin/fm-watch-checkpoint.sh:47-50 explicitly took the opposite approach in round 3, deferring the lib source until after argument handling so informational and error paths stay side-effect free; the two launchers are now inconsistent on the same point. Moving the arm&#39;s relocation below its argument `case` keeps it well before the fork and before any STATE-derived variable is consumed.
- ℹ️ `tests/fm-watcher-lock.test.sh:31` - fm_watcher_lock_test_cleanup iterates every pid ever passed to track_pid across the whole suite and issues `kill &#34;$p&#34;` with no liveness or identity check. By the time the EXIT trap fires, most tracked pids (the `sleep 300` sentinels and arms that already exited, plus the watcher pid read from state/.watch.lock/pid at line 105) are long dead, and on a busy host a wrapped pid can belong to an unrelated process. The individual tests already kill their own processes on the happy path, so the trap is the reuse-exposed one. Recording the pid alongside a cheap identity (fm_pid_identity is already available in the lib under test) and skipping any pid whose identity no longer matches would keep the safety net without the blind kill.

🔧 Fix: relocate below source guard and harden test cleanup
- ⚠️ `bin/fm-wake-lib.sh:9` - The relocation re-anchors a relative FM_HOME, STATE/FM_STATE_OVERRIDE and FM_ROOT/FM_ROOT_OVERRIDE, but not the paths already derived FROM STATE at lib source time. FM_WAKE_QUEUE and FM_WAKE_QUEUE_LOCK (bin/fm-wake-lib.sh:9-10) are computed as &#34;$STATE/.wake-queue&#34; before any relocation and are never re-derived, so with a relative STATE they stay relative: after the chdir into FM_HOME they resolve to $FM_HOME/&lt;relstate&gt;/.wake-queue while STATE itself has been re-anchored to &lt;slot&gt;/&lt;relstate&gt;. fm_wake_enqueue then appends to a directory that does not exist and the drain reads a different queue than the one the watcher writes. bin/fm-watch.sh:144 has the same shape: TRIAGE_LOG is set at top level above the source guard and, unlike WATCH_LOCK at line 674, is not re-derived after the relocation call. This is the identical completeness gap that was fixed for FM_ROOT in round 4, and the lib header comment currently asserts the re-anchoring is complete (&#34;every relative one is: FM_HOME, STATE ... and FM_ROOT&#34;), which is now inaccurate. Either re-derive the STATE-dependent globals inside fm_relocate_from_disposable_cwd (FM_WAKE_QUEUE, FM_WAKE_QUEUE_LOCK) and re-derive TRIAGE_LOG alongside WATCH_LOCK in bin/fm-watch.sh, or narrow the header claim and state that a relative STATE is unsupported.
- ℹ️ `tests/lib.sh:47` - tests/lib.sh now exports TREEHOUSE_DIR tree-wide to a path that does not exist. TREEHOUSE_DIR is treehouse&#39;s own pool-root variable, not a firstmate-private one, so the export reaches any suite that shells out to a real `treehouse` binary rather than the fakebin stub, and would point a genuine `treehouse get --lease` at a missing pool root instead of merely making the guard inert. Every current suite that touches treehouse installs a fake on PATH, so nothing is affected today, but the pin is broader than its stated purpose. Worth a line in the existing comment noting that the pin also overrides the real CLI, so a future suite that exercises real treehouse knows to unset it.

🔧 Fix: centralize path derivation and privatize pool-root override
- ⚠️ `tests/fm-secondmate-safety.test.sh:2009` - This suite is standalone (it never sources tests/lib.sh, so it does not inherit the tree-wide FM_TREEHOUSE_POOL_ROOT pin added at tests/lib.sh:47) and it EXECUTES the real &#34;$ROOT/bin/fm-watch.sh&#34; at line 2009 with FM_HOME pointed at a temp dir under TMP_ROOT. That is exactly the shape the guard now condemns whenever the repo checkout is itself a treehouse pool slot, which it is for any crewmate task worktree. The watcher will relocate and emit the banner on stderr inside a task worktree but not in CI, so the suite&#39;s watcher cwd and stderr are environment-dependent. Nothing fails today only because the assertions read $out (stdout), but this is the same class the change already fixed for tests/fm-afk-launch.test.sh, which was given its own standalone copy of the pin for precisely this reason. Round 5 audited the standalone suites for SOURCING the watcher (fm-backend-herdr-eventwait-smoke) but this one executes it. Add the same standalone FM_TREEHOUSE_POOL_ROOT pin and rationale comment carried by tests/fm-afk-launch.test.sh:26-37.
- ℹ️ `bin/fm-watch-arm.sh:347` - bin/fm-watch-arm.sh now assigns WATCH_LOCK, BEAT, CYCLE_LOG and CYCLE_LOG_LOCK twice verbatim: once at lines 75-84 and again at lines 347-350 after the relocation. Both bin/fm-wake-lib.sh (fm_derive_paths) and bin/fm-watch.sh (watch_derive_state_paths) were deliberately refactored in this same change so that a path derived from STATE lives in exactly one place and is re-anchored by construction. The arm is the one launcher left with a hand-maintained copy, so the next STATE-derived variable added near line 75 and not mirrored at line 347 silently keeps pointing at the pre-relocation state directory. Factor the four assignments into an arm_derive_state_paths function called at both points, matching the pattern the rest of the change establishes.
- ℹ️ `bin/fm-afk-start.sh:133` - Same duplication as the arm: FM_AFK_STATE and FM_AFK_LOCK are derived at lines 38-39 and re-derived byte-identically at lines 133-134 after the relocation. A future path derived from FM_AFK_STATE added at the top and not mirrored after the relocation would survive the chdir stale. Wrap the pair in a single derive function called from both sites.
- ℹ️ `bin/fm-wake-lib.sh:8` - FM_WAKE_QUEUE_PINNED is captured at source time from whether FM_WAKE_QUEUE is already set, and it is what makes fm_derive_paths skip re-anchoring that path. If bin/fm-wake-lib.sh is ever sourced twice within one process, the second source sees the value the FIRST source derived, records it as caller-pinned, and the relocation then leaves FM_WAKE_QUEUE and FM_WAKE_QUEUE_LOCK pointing at the pre-relocation state directory - reintroducing the exact enqueue/drain divergence round 6 fixed. No script double-sources it today (verified across bin/), so this is latent, but the mechanism gives no signal when it breaks. Guard the capture behind a source-once sentinel (e.g. only compute the _PINNED flags when FM_WAKE_LIB_SOURCED is unset) so a repeat source cannot mis-classify a lib-derived value as caller-supplied.

🔧 Fix: enforce pool-root pin by test and dedupe derived paths
2 infos still open:
- ℹ️ `tests/fm-relocation-pool-pin.test.sh:54` - The pin detector greps for any `FM_TREEHOUSE_POOL_ROOT=` assignment anywhere in the suite, including the per-invocation overrides that guard tests deliberately point at a REAL pool (e.g. tests/fm-afk-launch.test.sh:73, tests/fm-watch-checkpoint.test.sh:105). A future standalone suite that sets the variable only per-invocation to exercise relocation would therefore satisfy the check while every other case in that suite still runs unpinned - exactly the silent CI-versus-task-worktree divergence the meta-test exists to prevent. tests/lib.sh:47 and tests/fm-afk-launch.test.sh:39 both express the protective pin as a top-level `export FM_TREEHOUSE_POOL_ROOT=`, so anchoring the grep on that form (as test_shared_pin_is_owned_by_the_test_library already does for tests/lib.sh) would distinguish a real pin from a guard-exercising override.
- ℹ️ `tests/fm-watcher-lock.test.sh:145` - In test_relocated_watcher_does_not_run_in_condemned_cwd the forked watcher is only registered with track_pid after its pid is read from state/.watch.lock/pid. The `fail &#34;relocated arm never forked a watcher that took the lock&#34;` on the preceding line fires when the 20s discovery loop times out, which includes the case where the watcher process started but had not yet written the lock. Killing the tracked arm does not kill its already-forked watcher child, so that path exits the suite (tests/lib.sh `fail` is `exit 1`) leaving a real watcher polling against a state dir cleanup has removed - the precise leak the identity-checked trap was added for. Everywhere else in the suite the pid is tracked at launch; here the launch and the registration are separated by a race.

</details>

<details>

✅ No issues found.
- <code>`bash tests/fm-watcher-lock.test.sh` (35 assertions, includes the arm relocation, watcher-child cwd inheritance, symlinked/relative FM_HOME, relative STATE and FM_ROOT re-anchoring, and leased-home no-op cases)</code>
- <code>`bash tests/fm-afk-launch.test.sh` (52 assertions, includes the three new away-mode relocation assertions)</code>
- <code>`bash tests/fm-watch-checkpoint.test.sh` (checkpoint relocation path)</code>
- <code>`bash tests/fm-relocation-pool-pin.test.sh` (static contract that every launcher-touching suite pins its pool root)</code>
- <code>Manual end-to-end sweep simulation: armed `bin/fm-watch-arm.sh` from a simulated treehouse pool slot with the guard inert vs active, read each watcher&#39;s real working directory via `lsof -a -p &lt;pid&gt; -d cwd`, then ran a cwd-matching terminate sweep against both</code>
- <code>Manual `bin/fm-afk-start.sh` runs from a disposable pool slot and from a durably leased secondmate home, comparing banner presence</code>

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -76,8 +76,14 @@ for test_script in tests/*.test.sh; do bash "$test_script"; done   # behavior te
 tmp=$(mktemp -d) && printf 'done: smoke\n' > "$tmp/smoke.status" && FM_STATE_OVERRIDE="$tmp" FM_SIGNAL_GRACE=1 FM_POLL=1 FM_HEARTBEAT=999999 bin/fm-watch-arm.sh  # watcher re-arm smoke test (prints arm status, then an actionable signal)
 ```
 
+Run the watcher smoke test from a checkout that is itself a disposable treehouse worktree and it first prints a loud relocation banner on stderr before arming normally; that is the expected self-correction, not a failure.
+
 Discover tests by listing `tests/*.test.sh`: each is a self-contained bash script named `<subject>.test.sh`, and its header comment describes what it covers, so run one directly to focus on a subject.
 Tests that need a real optional backend or an explicit opt-in (real herdr/zellij/cmux smoke tests, the live Pi regression) skip themselves and print the tool or environment gate needed to enable them, so the run-all loop above is always safe.
+
+A suite that runs `bin/fm-watch.sh`, `bin/fm-watch-arm.sh`, `bin/fm-watch-checkpoint.sh`, or `bin/fm-afk-start.sh` needs a treehouse pool-root pin, because those launchers relocate themselves out of a disposable pool slot and a checkout that is itself a task worktree is exactly that shape.
+Sourcing `tests/lib.sh`, directly or through a helper that sources it, inherits the tree-wide `FM_TREEHOUSE_POOL_ROOT` pin; a standalone suite sets the variable itself, and a suite that only mentions a launcher without running it carries the `relocation-pool-pin: not-needed` marker instead.
+`tests/fm-relocation-pool-pin.test.sh` enforces that at authoring time.
 
 ## Questions
 
--- a/bin/fm-afk-start.sh
+++ b/bin/fm-afk-start.sh
@@ -128,7 +128,8 @@ fm_afk_start_main() {
     echo "afk: FAILED - cannot move out of the disposable task worktree ($FM_DISPOSABLE_CWD) into the home ($FM_DISPOSABLE_HOME)" >&2
     return 1
   fi
-  # The relocation re-anchors STATE, so re-derive the paths taken from it.
+  # The relocation may re-anchor a relative STATE, so re-derive the paths taken
+  # from it. An already-absolute STATE is left byte-identical.
   FM_AFK_STATE="${FM_STATE_OVERRIDE:-$FM_HOME/state}"
   FM_AFK_LOCK="$FM_AFK_STATE/.supervise-daemon.lock"
 
--- a/bin/fm-afk-start.sh
+++ b/bin/fm-afk-start.sh
@@ -119,15 +119,18 @@ fm_afk_start_main() {
 
   # The daemon is long-lived and carries supervision while nobody is watching, so
   # it has the same disposable-cwd hazard as the watcher (bin/fm-wake-lib.sh's
-  # fm_cwd_is_disposable_slot). Refuse before touching state, so away mode is
-  # never entered without a daemon that can survive to own it.
-  if fm_cwd_is_disposable_slot; then
-    fm_disposable_cwd_banner \
-      'REFUSING TO START AWAY MODE - CWD IS A DISPOSABLE TASK WORKTREE' \
-      'bin/fm-afk-start.sh'
-    echo "afk: FAILED - refusing to start from a disposable task worktree ($FM_DISPOSABLE_CWD)" >&2
+  # fm_relocate_from_disposable_cwd). Relocate before the exec, so the daemon
+  # inherits the safe working directory. Relocating rather than refusing also
+  # means the launcher paths that already wrote state/.afk can never leave away
+  # mode live with no daemon to own it.
+  if ! fm_relocate_from_disposable_cwd \
+    'RELOCATED AWAY MODE OUT OF A DISPOSABLE TASK WORKTREE'; then
+    echo "afk: FAILED - cannot move out of the disposable task worktree ($FM_DISPOSABLE_CWD) into the home ($FM_DISPOSABLE_HOME)" >&2
     return 1
   fi
+  # The relocation re-anchors STATE, so re-derive the paths taken from it.
+  FM_AFK_STATE="${FM_STATE_OVERRIDE:-$FM_HOME/state}"
+  FM_AFK_LOCK="$FM_AFK_STATE/.supervise-daemon.lock"
 
   mkdir -p "$FM_AFK_STATE"
   if [ "${FM_AFK_STATE_PREPARED:-0}" = 1 ]; then
--- a/bin/fm-afk-start.sh
+++ b/bin/fm-afk-start.sh
@@ -117,6 +117,18 @@ fm_afk_start_main() {
     * ) echo "usage: $(basename "${BASH_SOURCE[1]:-fm-afk-start.sh}")" >&2; return 2 ;;
   esac
 
+  # The daemon is long-lived and carries supervision while nobody is watching, so
+  # it has the same disposable-cwd hazard as the watcher (bin/fm-wake-lib.sh's
+  # fm_cwd_is_disposable_slot). Refuse before touching state, so away mode is
+  # never entered without a daemon that can survive to own it.
+  if fm_cwd_is_disposable_slot; then
+    fm_disposable_cwd_banner \
+      'REFUSING TO START AWAY MODE - CWD IS A DISPOSABLE TASK WORKTREE' \
+      'bin/fm-afk-start.sh'
+    echo "afk: FAILED - refusing to start from a disposable task worktree ($FM_DISPOSABLE_CWD)" >&2
+    return 1
+  fi
+
   mkdir -p "$FM_AFK_STATE"
   if [ "${FM_AFK_STATE_PREPARED:-0}" = 1 ]; then
     [ -f "$FM_AFK_STATE/.afk" ] || { echo "afk: launcher-prepared state is missing" >&2; return 1; }
--- a/bin/fm-afk-start.sh
+++ b/bin/fm-afk-start.sh
@@ -125,7 +125,7 @@ fm_afk_start_main() {
   # mode live with no daemon to own it.
   if ! fm_relocate_from_disposable_cwd \
     'RELOCATED AWAY MODE OUT OF A DISPOSABLE TASK WORKTREE'; then
-    echo "afk: FAILED - cannot move out of the disposable task worktree ($FM_DISPOSABLE_CWD) into the home ($FM_DISPOSABLE_HOME)" >&2
+    echo "afk: FAILED - $FM_DISPOSABLE_ERROR" >&2
     return 1
   fi
   # The relocation may re-anchor a relative STATE, so re-derive the paths taken
--- a/bin/fm-afk-start.sh
+++ b/bin/fm-afk-start.sh
@@ -35,8 +35,14 @@ set -eu
 FM_AFK_START_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 FM_ROOT="${FM_ROOT_OVERRIDE:-$(cd "$FM_AFK_START_DIR/.." && pwd)}"
 FM_HOME="${FM_HOME:-${FM_ROOT_OVERRIDE:-$FM_ROOT}}"
-FM_AFK_STATE="${FM_STATE_OVERRIDE:-$FM_HOME/state}"
-FM_AFK_LOCK="$FM_AFK_STATE/.supervise-daemon.lock"
+# The single owner of every away-mode path derived from FM_HOME or STATE. It runs
+# here and again after the relocation re-anchors those bases, so a path added to it
+# is re-anchored by construction rather than by a hand-maintained second copy.
+fm_afk_derive_state_paths() {
+  FM_AFK_STATE="${FM_STATE_OVERRIDE:-$FM_HOME/state}"
+  FM_AFK_LOCK="$FM_AFK_STATE/.supervise-daemon.lock"
+}
+fm_afk_derive_state_paths
 FM_AFK_DAEMON="$FM_AFK_START_DIR/fm-supervise-daemon.sh"
 
 # shellcheck source=bin/fm-wake-lib.sh
@@ -130,8 +136,7 @@ fm_afk_start_main() {
   fi
   # The relocation may re-anchor a relative STATE, so re-derive the paths taken
   # from it. An already-absolute STATE is left byte-identical.
-  FM_AFK_STATE="${FM_STATE_OVERRIDE:-$FM_HOME/state}"
-  FM_AFK_LOCK="$FM_AFK_STATE/.supervise-daemon.lock"
+  fm_afk_derive_state_paths
 
   mkdir -p "$FM_AFK_STATE"
   if [ "${FM_AFK_STATE_PREPARED:-0}" = 1 ]; then
--- a/bin/fm-wake-lib.sh
+++ b/bin/fm-wake-lib.sh
@@ -3,11 +3,24 @@
 
 FM_WAKE_LIB_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 FM_WAKE_DEFAULT_ROOT="$(cd "$FM_WAKE_LIB_DIR/.." && pwd)"
-FM_ROOT="${FM_ROOT_OVERRIDE:-${FM_ROOT:-$FM_WAKE_DEFAULT_ROOT}}"
-FM_HOME="${FM_HOME:-${FM_ROOT_OVERRIDE:-$FM_ROOT}}"
-STATE="${FM_STATE_OVERRIDE:-${STATE:-$FM_HOME/state}}"
-FM_WAKE_QUEUE="${FM_WAKE_QUEUE:-$STATE/.wake-queue}"
-FM_WAKE_QUEUE_LOCK="${FM_WAKE_QUEUE_LOCK:-$STATE/.wake-queue.lock}"
+# Whether the environment supplied these explicitly, captured once at source time
+# so a re-derivation never silently overwrites a caller's own value.
+FM_WAKE_QUEUE_PINNED="${FM_WAKE_QUEUE:+1}"
+FM_WAKE_QUEUE_LOCK_PINNED="${FM_WAKE_QUEUE_LOCK:+1}"
+
+# The single owner of every path this lib derives from FM_ROOT, FM_HOME or STATE.
+# It runs at source time and again after a relocation re-anchors the base paths,
+# so a derived path added here is re-anchored by construction and cannot drift out
+# of sync with the base it came from. Each assignment keeps an already-set value,
+# so re-running it recomputes only what the relocation invalidated.
+fm_derive_paths() {
+  FM_ROOT="${FM_ROOT_OVERRIDE:-${FM_ROOT:-$FM_WAKE_DEFAULT_ROOT}}"
+  FM_HOME="${FM_HOME:-${FM_ROOT_OVERRIDE:-$FM_ROOT}}"
+  STATE="${FM_STATE_OVERRIDE:-${STATE:-$FM_HOME/state}}"
+  [ -n "$FM_WAKE_QUEUE_PINNED" ] || FM_WAKE_QUEUE="$STATE/.wake-queue"
+  [ -n "$FM_WAKE_QUEUE_LOCK_PINNED" ] || FM_WAKE_QUEUE_LOCK="$STATE/.wake-queue.lock"
+}
+fm_derive_paths
 FM_LOCK_STALE_AFTER="${FM_LOCK_STALE_AFTER:-2}"
 mkdir -p "$STATE"
 
@@ -61,9 +74,12 @@ fm_path_age() {
 #
 # A durably leased secondmate home is itself a pool slot (bin/fm-home-seed.sh) but
 # is stable for that home's lifetime, so a cwd inside FM_HOME is always safe. Only
-# a pool slot that is NOT this home is refused. TREEHOUSE_DIR is treehouse's own
-# pool-root variable and defaults to ~/.treehouse; a pool root that does not
-# resolve leaves the guard inert.
+# a pool slot that is NOT this home is refused. The pool root resolves from
+# FM_TREEHOUSE_POOL_ROOT first, then TREEHOUSE_DIR, then ~/.treehouse; a pool root
+# that does not resolve leaves the guard inert. FM_TREEHOUSE_POOL_ROOT is
+# firstmate-private and exists so the test harness can point this guard at an
+# absent pool without redirecting the real treehouse CLI, which reads
+# TREEHOUSE_DIR as its own pool root and would act on a bogus value.
 #
 # The condition is self-correcting rather than operator-correctable. A launcher
 # calls fm_relocate_from_disposable_cwd, which moves the launcher process itself
@@ -100,7 +116,10 @@ fm_path_age() {
 # re-anchored environment is harder to reason about than a fully re-anchored one,
 # and FM_ROOT in particular is inherited by every child process the launcher
 # starts, so a stale relative value would resolve against the wrong directory
-# well after the relocation.
+# well after the relocation. Everything derived FROM those bases is then
+# recomputed through fm_derive_paths, the single owner of that derivation, rather
+# than re-anchored variable by variable, so a path added there is covered
+# automatically.
 fm_path_under() {
   local child=$1 parent=$2
   [ -n "$parent" ] || return 1
@@ -131,7 +150,7 @@ fm_cwd_is_disposable_slot() {
   FM_DISPOSABLE_CWD=
   FM_DISPOSABLE_HOME=
   cwd=$(pwd -P 2>/dev/null) || return 1
-  pool=$(fm_resolve_dir "${TREEHOUSE_DIR:-${HOME:-}/.treehouse}") || return 1
+  pool=$(fm_resolve_dir "${FM_TREEHOUSE_POOL_ROOT:-${TREEHOUSE_DIR:-${HOME:-}/.treehouse}}") || return 1
   fm_path_under "$cwd" "$pool" || return 1
   home=$(fm_resolve_dir "$FM_HOME") || home=$FM_HOME
   fm_path_under "$cwd" "$home" && return 1
@@ -193,6 +212,10 @@ fm_relocate_from_disposable_cwd() {
     FM_ROOT=$root_abs
     [ -z "${FM_ROOT_OVERRIDE:-}" ] || FM_ROOT_OVERRIDE=$root_abs
   fi
+  # Recompute every derived path from the re-anchored bases through their single
+  # owner, so nothing derived from STATE or FM_HOME survives the chdir as a stale
+  # relative value.
+  fm_derive_paths
   rule='━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━'
   {
     printf '●%s\n' "$rule"
--- a/bin/fm-wake-lib.sh
+++ b/bin/fm-wake-lib.sh
@@ -73,6 +73,25 @@ fm_path_age() {
 # do. It stays a warning rather than a silent repair so the condition remains
 # visible. A failed chdir into FM_HOME means a genuinely broken home, and the
 # helper returns non-zero so the caller can fail with its own typed line.
+#
+# bin/fm-watch.sh calls it itself, which is the choke point every watcher
+# launcher passes through, so no present or future launcher can miss the guard.
+# A launcher that also BLOCKS for the watcher's lifetime - bin/fm-watch-arm.sh
+# while it waits on its child, bin/fm-watch-checkpoint.sh for the whole bounded
+# checkpoint, bin/fm-afk-start.sh before the daemon exec - calls it on its own
+# account too, since a launcher killed by the sweep ends the work even when the
+# watcher it started was safe. The call is idempotent, so the second one detects
+# nothing and prints nothing.
+#
+# The relocation deliberately does NOT resolve an already-absolute FM_HOME or
+# STATE. FM_HOME doubles as a byte-exact cross-process identity string: the
+# watcher writes it into state/.watch.lock/fm-home and fm_watcher_lock_matches_pid,
+# bin/fm-turnend-guard.sh, bin/fm-continuity-pretool-check.sh,
+# bin/fm-pr-check-migrate.sh and bin/fm-watch-arm.sh all compare it verbatim
+# against their own logically derived value. Rewriting it to the resolved path
+# would desynchronize the lock from every one of them whenever a component of
+# the home path is a symlink. Only a genuinely relative value, which a chdir
+# really does invalidate, is re-anchored.
 fm_path_under() {
   local child=$1 parent=$2
   [ -n "$parent" ] || return 1
@@ -113,13 +132,24 @@ fm_cwd_is_disposable_slot() {
 fm_relocate_from_disposable_cwd() {
   local headline=$1 rule state_abs
   fm_cwd_is_disposable_slot || return 0
-  # Re-anchor the home and state paths first: a chdir invalidates any of them that
-  # was relative to the condemned cwd, and the resolved home is already absolute.
+  # Re-anchor only the paths a chdir would actually invalidate, which is the
+  # genuinely relative ones. FM_HOME doubles as a byte-exact cross-process
+  # identity string, written into the watcher lock and compared verbatim by the
+  # guard, the checkpoint and the continuity check, so an already-absolute value
+  # must survive relocation unchanged even when a path component is a symlink.
   state_abs=$(fm_resolve_dir "$STATE") || return 1
-  CDPATH='' cd -- "$FM_DISPOSABLE_HOME" 2>/dev/null || return 1
-  FM_HOME=$FM_DISPOSABLE_HOME
-  STATE=$state_abs
-  [ -z "${FM_STATE_OVERRIDE:-}" ] || FM_STATE_OVERRIDE=$state_abs
+  CDPATH='' cd -- "$FM_HOME" 2>/dev/null || return 1
+  case "$FM_HOME" in
+    /*) ;;
+    *) FM_HOME=$FM_DISPOSABLE_HOME ;;
+  esac
+  case "$STATE" in
+    /*) ;;
+    *)
+      STATE=$state_abs
+      [ -z "${FM_STATE_OVERRIDE:-}" ] || FM_STATE_OVERRIDE=$state_abs
+      ;;
+  esac
   rule='━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━'
   {
     printf '●%s\n' "$rule"
--- a/bin/fm-wake-lib.sh
+++ b/bin/fm-wake-lib.sh
@@ -50,6 +50,73 @@ fm_path_age() {
   echo $(( $(date +%s) - m ))
 }
 
+# Disposable-cwd guard for long-lived firstmate processes.
+#
+# treehouse's lingering-process sweep, which fires on every `treehouse return` and
+# on the exit of an interactive `treehouse get` shell, kills every process whose
+# working directory resolves under the worktree being returned. It matches on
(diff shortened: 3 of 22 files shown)


kunchenguid:

<sub>Noted for firstmate#725 at `98a1411e`.</sub>

kunchenguid:

kunchenguid:
I am closing this because it has been waiting on a rebase or merge-conflict fix since 2026-07-21, and I have not seen a comment or push since then.

PR #732, opened 2026-07-19
feat: add Pi supervised question bridge

stoneymarrow:
## Summary

- add the versioned Firstmate-owned Pi question request, response, owner, resolution, and locking protocol
- wire supervised Pi crew launches to the bridge and add guarded answer, startup recovery, timeout, cancellation, reissue, and teardown handling
- document the schema and operator workflow, including keyed status folding for the report-defined suffix format
- keep Phase 6 distribution, canary, and primary experiments out of scope

## Validation

- bash tests/fm-pi-supervised-question.test.sh
- bash tests/fm-pi-watch-extension.test.sh
- bash tests/fm-spawn-dispatch-profile.test.sh
- bash tests/fm-bootstrap.test.sh
- bash tests/fm-session-start.test.sh
- bash tests/fm-teardown.test.sh
- bash tests/fm-watch-triage.test.sh
- bash tests/fm-decision-hold-lifecycle.test.sh
- bin/fm-lint.sh
- git diff --check

### DIFF
--- a/.agents/skills/bootstrap-diagnostics/SKILL.md
+++ b/.agents/skills/bootstrap-diagnostics/SKILL.md
@@ -2,7 +2,7 @@
 name: bootstrap-diagnostics
 description: >-
   Agent-only handling playbook for session-start bootstrap diagnostics.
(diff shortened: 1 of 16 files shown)


stoneymarrow:
CI follow-up: pushed commit 89315d3 to fix the Linux behavior-test failure. GNU stat accepts -f as filesystem reporting, so the old BSD-or-GNU fallback never reached stat -c; the test now selects mode syntax by platform. The focused Pi bridge suite, simulated Linux mode path, ShellCheck, and git diff --check pass locally. The remaining no-mistakes check requires a PR-body signature from that pipeline; this repair task was explicitly instructed not to run or claim that pipeline.

stoneymarrow:
Withdrawing this contribution because existing Firstmate supervision, session inspection, and steering are sufficient for our current Pi question-handling needs. We will revisit a separate protocol only if a demonstrated gap remains.

PR #759, opened 2026-07-20
fix(teardown): refuse teardown when treehouse worktree slot was re-handed

LuckyB-UF:
## Intent

Guard fm-teardown so it refuses to tear down a task whose treehouse worktree slot was re-handed to another task after an outage; fail closed on unrecognized slot states; --force must not bypass it.

## What Changed

- Added a treehouse slot-ownership guard to `bin/fm-teardown.sh` that queries `treehouse status` for the task's worktree row and refuses to run `treehouse return --force` when the slot has been re-handed to another task after an outage.
- Split the guard's state check from a confident agent-liveness probe: `available`/free slots and slots still owned by a live agent proceed, while `leased`, occupied-by-a-dead-agent, and unrecognized states fail closed with no `--force` bypass — and the state-vs-liveness split lets normal zellij/cmux teardowns through instead of blocking them.
- Extended `tests/fm-teardown.test.sh` and `tests/fm-backend.test.sh` to cover the re-handed-slot refusal, unrecognized-state fail-closed behavior, and the expected status-log entries; documented the guard in `docs/architecture.md` and `docs/zellij-backend.md`.

## Risk Assessment

✅ Low: A well-bounded, defense-in-depth guard layered before destructive teardown actions with comprehensive tests and no dangerous fail-open paths; the sole residual edge is a deliberate, previously-decided fail-closed tradeoff with an operator escape.

## Testing

- ⏭️ **Test** - skipped

## Pipeline

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>

- ⚠️ `bin/fm-teardown.sh:576` - In the fall-through (in-use/busy/unrecognized) branch, the guard uses fm_backend_target_exists (pane/window PRESENCE only) to decide &#34;the live owner is us; proceed&#34;. fm-backend.sh (lines 697-717) explicitly documents that fm_backend_target_exists reports a lingering bare shell as &#34;alive&#34;, and provides fm_backend_agent_alive as the CONFIDENT agent-liveness probe precisely to close that gap. Failure scenario: a dead task&#39;s agent process exits but its tmux window lingers as a bare shell; treehouse frees the slot on owner-process death and re-hands it to a NEW task that treehouse now reports &#39;in-use&#39;. When tearing down the dead task, target_exists returns true for the lingering shell, the guard returns 0, and teardown proceeds to `treehouse return --force &lt;wt&gt;`, killing the new owner&#39;s live worktree — exactly the hazard this guard exists to prevent. Note the residual gap is no worse than the pre-change status quo (no guard at all); using fm_backend_agent_alive here would tighten it, at the cost of possibly refusing a genuine own-task teardown where the agent already exited — a deliberate tradeoff worth confirming.
- ℹ️ `bin/fm-teardown.sh:567` - The `*lease*` case matches any lowercased state containing the substring &#34;lease&#34;, which also matches a hypothetical free/transient state such as &#34;released&#34;. Such a state would be refused as &#34;leased to &#39;unknown&#39;&#34; (a misleading message) rather than being handled by the intended free-slot or unknown-state logic. Given treehouse&#39;s verified state set is only available/leased/in-use this is currently unreachable and fail-closed (safe direction), but the pattern is broader than the single &#34;leased&#34; word it targets.
- ℹ️ `bin/fm-teardown.sh:548` - The awk row parser takes path=$3 as a single field, so a treehouse status path containing whitespace would be truncated and fail to match wt_canon, silently skipping the row (guard returns 0 / proceeds). Treehouse pool worktree paths are managed and do not contain spaces today, so this is a latent robustness limit rather than an active bug.

🔧 Fix: harden teardown treehouse guard: confident liveness, exact lease, spaced paths
- ⚠️ `bin/fm-teardown.sh:585` - The in-use/unrecognized fall-through proves &#39;the slot is still ours&#39; via fm_backend_agent_alive, but that helper (bin/fm-backend.sh:718) only classifies tmux and herdr and returns &#39;unknown&#39; for every other backend. treehouse get runs for zellij and cmux tasks too (bin/fm-spawn.sh:834), and a live task&#39;s own slot reports &#39;in-use&#39; (not &#39;available&#39;/&#39;here&#39;) when treehouse status is queried from PROJ. Failure scenario: a healthy zellij- or cmux-backed task reaches teardown; the guard matches its own in-use slot, calls fm_backend_agent_alive(&#39;zellij&#39;|&#39;cmux&#39;, ...) which always returns &#39;unknown&#39;, and REFUSES with no --force bypass. Every normal teardown of a zellij/cmux task using a treehouse worktree is blocked, leaving the task un-tearable-down. Only tmux/herdr can ever satisfy the &#39;alive&#39; branch. Consider scoping the confident-liveness requirement to backends fm_backend_agent_alive can actually probe, and choosing a defined behavior for un-probable backends rather than the accidental always-refuse.
- ℹ️ `bin/fm-teardown.sh:581` - The &#39;available|*here*&#39; branch returns 0 (proceed) on any lowercased state containing &#39;here&#39;. treehouse status is queried from PROJ, not from within a slot, so no row should carry the self-occupancy &#39;you are here&#39; marker in its state column - making this arm likely dead code. More importantly it is the one match arm in the guard whose direction is fail-OPEN: were a foreign-owned row&#39;s state to ever contain &#39;here&#39;, the guard would allow the destructive return instead of refusing. Given treehouse&#39;s verified state set (available/leased/in-use) this is currently unreachable, but the substring match is broader than the intended self-occupancy signal and in the permissive direction.

🔧 Fix: split treehouse guard state vs liveness so zellij/cmux teardowns proceed
1 info still open:
- ℹ️ `bin/fm-teardown.sh:606` - Deliberate tradeoff (decided in prior review rounds): for tmux/herdr, an occupied (&#39;in-use&#39;) slot with fm_backend_agent_alive returning &#39;dead&#39; is refused even under --force. This correctly blocks re-handed-slot hazards, but also means a legitimately-owned task whose agent crashed to a bare shell while its treehouse slot is still &#39;in-use&#39; cannot be torn down by teardown OR --force; the operator must first kill the pane (making the slot &#39;available&#39;, as stuck-crewmate-recovery does) before teardown proceeds. Happy-path teardown is unaffected because the agent is alive at guard time (teardown itself does the killing). Noting for maintainer awareness; no change requested.

</details>

<details>

Step was skipped.

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>


LuckyB-UF:
Closing: this was auto-opened against the upstream repo (kunchenguid/firstmate) by mistake. Firstmate work lands on fork/main per fleet policy. Retargeted to the fork as LuckyB-UF/firstmate#1 (same branch, same commits).

PR #765, opened 2026-07-20
fix(spawn): require two stable reads before accepting worktree path

Freudator86:
## Intent

Fix a metadata-recording race in bin/fm-spawn.sh's treehouse-get worktree-detection loop. On some tmux/WSL setups, a brand-new tmux window's pane_current_path transiently reports a stale-but-real path (e.g. an old separate firstmate home checkout) before the pane actually settles into the real worktree treehouse get moved it to. The old loop accepted the first read that differed from the project path and broke immediately, so it could capture that stale value; because the stale path is itself a real, distinct git checkout, validate_spawn_worktree's isolation check also passed it, so the wrong path silently landed in state/<id>.meta's worktree= field (and, for the claude harness, the turn-end hook file path). This was reproduced live twice against a real crewmate that was actually running correctly the whole time - only the early metadata capture was wrong, not the crewmate's actual cwd. Deliberately chose the more robust fix over a simple pre-loop settle delay: the loop now requires two consecutive polls to report the same non-project path before accepting it as the worktree, using the loop's existing 1s inter-poll sleep as the confirmation gap so an already-settled pane isn't slowed by an extra full cycle. Explicitly out of scope per the task: validate_spawn_worktree() itself is left untouched, since it's a reasonable defense-in-depth layer that just isn't sufficient alone against a stale-but-real worktree path. Added tests/fm-spawn-worktree-settle.test.sh, a new regression test with a fake tmux stub simulating the transient-then-settled pane_current_path sequence; verified it fails against the pre-fix code and passes with the fix, plus a timing assertion that an already-settled pane doesn't pay the extra confirmation cycle.

## What Changed

- In `bin/fm-spawn.sh`'s treehouse-get worktree-detection loop, require two consecutive polls to agree on the same non-project path before accepting it as the settled worktree, instead of accepting the first differing read.
- Uses the loop's existing 1s inter-poll sleep as the confirmation gap, so an already-settled pane isn't penalized with an extra full polling cycle.
- Adds `tests/fm-spawn-worktree-settle.test.sh` with a fake tmux stub that simulates a transient stale `pane_current_path` followed by the real settled path, verifying the fix rejects the stale read and a timing assertion that an already-settled pane doesn't pay the extra confirmation cycle.

## Risk Assessment

✅ Low: Small, well-scoped fix confined to a single detection loop; logic correctly implements the two-consecutive-read confirmation described in the intent, leaves validate_spawn_worktree untouched as required, and is covered by a new regression test using existing test helpers.

## Testing

- Outcome: ⏭️ skipped across 1 run (17m53s)

## Pipeline

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>

</details>

<details>

✅ No issues found.

</details>

<details>

</details>

<details>

✅ No issues found.

</details>


kunchenguid:
Thanks @Freudator86 - merged! Really appreciate the contribution.

PR #772, opened 2026-07-20
fix(watch): report queued wakes at arm cycle end

raywest:
Fixes the arm-layer cycle-end misreporting introduced by #693: an owned or attached arm cycle that ends without visible reason text now checks whether the durable wake queue actually advanced during the cycle before synthesizing 'watcher: FAILED - ... without an actionable reason'. If wakes landed, it reports an accurate 'N wake(s) already queued' line with a distinct exit code (4), preserving the FAILED text and exit-1 semantics for the genuinely-empty case. Also hardens fm-watch.sh so an already-durable wake is never followed by a silent exit-1 (best-effort commit-transition, partial-append diagnostics), and fixes pre-existing SIGPIPE stderr noise in the herdr events-capability probe (herestring instead of printf-pipe). Counts queue records directly - a counter that advanced with no record behind it can never claim a wake; the seq snapshot serves only as a same-second boundary marker. Regression tests cover the queued-wake-with-no-visible-reason case for both owned and attached shapes. Diagnosed with deterministic reproductions on a live fleet; happy to share the repro scripts.


raywest:
Superseded by #807, re-raised through no-mistakes per this repo's contribution policy - thanks!

PR #775, opened 2026-07-20
fix(send): treat opencode busy-queued composer state as submitted

KostadinP:
## Root cause

`fm_tmux_submit_enter_core` (in `bin/fm-tmux-lib.sh`) types text, presses
Enter, then verifies delivery by reading the composer row.  When the target
opencode pane is mid-turn (busy, footer shows `esc interrupt`), opencode
accepts the Enter and queues the message for processing when the current
turn ends, but leaves the typed text visible in the composer row.  The
verification loop sees a `pending` composer, exhausts all retries, and
reports the submit as failed with `error: text not submitted ... (Enter
swallowed; text left in composer)` and exits 1 — even though the message
was delivered and will be consumed.

This is a false failure: callers that retry on non-zero exit send the
message twice.

## Fix

After the Enter retries are exhausted and the composer still shows
`pending`, check whether the pane is BUSY using the existing
`fm_pane_is_busy` (which scans the pane tail for the busy footer regex
`esc (to )?interrupt|Working...|Ctrl+c:cancel`).  If the pane is busy,
the harness has queued the message, so return the accepted verdict
(`empty`) instead of `pending`.  On an idle pane, keep returning
`pending` — a genuine swallow must still be detected.

This uses a robust, version-independent signal (the busy footer) instead
of a brittle label match.  Retries still run first (a transient render
clears on retry); only the post-retries-verdict is reclassified.

## Empirical evidence

opencode version: 1.18.4

### Before (busy pane)

Send to a pane showing `esc interrupt`:

```
$ FM_HOME=/fleet/firstmate fm-send oc-busy-test:opencode captain on deck
error: text not submitted to oc-busy-test:opencode (Enter swallowed;
       text left in composer; tried ...)
EXIT: 1
```

Message IS visible in the conversation with `QUEUED` label and is
consumed when the turn ends.

### After (busy pane)

The same scenario now exits 0.  Message is delivered exactly once.

### Idle genuine-swallow unchanged

An idle pane where Enter does not submit still reports `pending` and
exits 1.  The fix does not mask real swallows.

## Regression tests

`tests/fm-tmux-submit-busy.test.sh` covers four scenarios:

- busy pane + pending composer after Enter retries -> returns `empty`
- idle pane + pending composer after Enter retries -> returns `pending`
- busy pane + composer clears on first Enter -> returns `empty`
- idle pane + composer clears on first Enter -> returns `empty`

All mock the underlying tmux commands and override `fm_pane_is_busy`
with a test stub (controlled by `FM_FAKE_PANE_BUSY`).

### herdr backend

`fm_backend_herdr_send_text_submit` has a parallel gap: when the
baseline agent status is `busy` (pre-Enter agent was `working` or
`blocked`), it falls back to `fm_backend_herdr_composer_state` for
submit confirmation.  An opencode pane that is busy and queues the
message would also return `pending` after retries.  A busy-state check
analogous to the tmux fix would need to consult
`fm_backend_herdr_busy_state`, but the herdr case is more nuanced
because a `blocked` agent (stuck waiting on human input) should not
auto-accept.  Needs separate investigation.

### orca and cmux backends

Both use the same composer-state verification pattern (`empty`/`pending`
returned by `fm_backend_orca_composer_state` /
`fm_backend_cmux_composer_state`) but have no `is_busy` primitive.  Not
applicable without one.

### zellij backend

Uses content-diff (capture before/after Enter) instead of composer
state.  If opencode queues a message without changing the pane content,
the diff shows no change — same potential gap, but there is no
`is_busy` primitive.  Not applicable without one.

## Pipeline


kunchenguid:
Thanks @KostadinP - merged! Really appreciate the contribution.

PR #787, opened 2026-07-21
feat(harness): add verified Cursor worker adapter

Amplify-Logic:
## Summary

Adds `cursor` (Cursor CLI `agent`) as a verified **worker-only** adapter. Primary-orchestrator support was not verified and is not claimed.

Verification also found two shared composer-monitoring defects that would have made the adapter unsafe; both are fixed in the shared owners:

1. **Reverse-video idle cell** - cursor draws the terminal cursor as SGR 7 over the idle placeholder's first character. That survives ghost-stripping, so an idle composer reduced to a lone character and classified as `pending`, deferring away-mode escalations forever. The idle regex is now also matched against the plain (unstyled) row.
2. **False empty composer** - `#{cursor_y}` points at cursor's bottom status area, not the composer. Reading that row found an empty status line while real unsubmitted text was present - the dangerous false-empty direction for the away-mode injector. The tmux path now locates the composer structurally (scoped to panes positively identified as cursor).

Other integration points:

- Launch template: `--yolo`, `--workspace` pinned to the task worktree, deliberately no `-w` so cursor never allocates a second worktree
- Per-task gitignored `.cursor/hooks.json` `stop` hook
- Effort is a suffix on the model id (`cursor_model_with_effort`); `xhigh`/`max` cap at `high`
- `CURSOR_AGENT=1` detection is tested before `CLAUDECODE` because cursor does not clear an inherited Claude marker
- Busy signature `ctrl+c to stop` (not the spinner verb, which flips mid-turn)
- Bootstrap dispatch validation and secondmate liveness recognize cursor as a verified worker, including its effort range

Evidence and limitations: `docs/cursor-harness.md`. Backends: tmux verified; herdr composer-safe by construction with liveness unverified; zellij/orca/cmux not exercised with cursor.

## Test plan

- [x] `bin/fm-lint.sh`
- [x] `tests/fm-cursor-adapter.test.sh` (including under ambient `CURSOR_AGENT=1`)
- [x] `tests/fm-bootstrap.test.sh`
- [x] `tests/fm-secondmate-liveness.test.sh`


kunchenguid:

PR #808, opened 2026-07-21
code-shape: sharpen the tending brief and comment-hygiene bar

webjema:
## What

Sharpens the `/code-shape` skill so a tending pass does the right work by default, addressing a review of its guidance. The scoping/tracking engine and the inherited ship pipeline (`/code-review`, `/verify`, quality hooks) were already solid; the gap was in the crew brief that step 3 hands off — it named *what* to fix but not *how to do it safely*, and the comment rule covered only two of the common cases.

## Changes (all in step 3's `{TASK}` guidance)

- **Concrete smell checklist** for structure/testability instead of "maintainability and testability smells" — long functions, deep nesting, mixed I/O and logic, hidden deps/globals, primitive obsession, unclear names.
- **Deduplication guardrail** — rule-of-three, and keep extracted code in its natural home rather than reaching across the slice boundary (avoids premature abstraction and cross-slice coupling).
- **Broader comment-hygiene rule** — now also: delete redundant comments that restate the code and commented-out code; correct/delete a comment that has gone *wrong* (worse than none); and never strip docstrings, public-API docs, or license headers.
- **New safety bar** — the piece most missing before:
  - *Pure refactor*: behavior stays identical; no feature/perf/bugfix folded in; surface a found bug as a finding, don't bury it in a cleanup diff.
  - *Tests are the net*: a green suite only protects covered code, so confirm coverage before refactoring a unit — add characterization tests first, or skip it and say why.
  - *Reviewable diff*: small semantic commits; no mass reformat/rename that buries the change and destroys `git blame`.
- Direction-recording summary broadened to match the wider comment bar.

Style: one sentence per line, plain dashes, no agent co-author (per `firstmate-coding-guidelines`).


webjema:
Closing — opened against the wrong repo; re-opening on the fork (webjema/firstmate).

PR #862, opened 2026-07-22
fix(afk): refuse bare-shell supervisor targets and bound watcher shutdown

ManveerBhullar:
## What

Two away-mode supervision reliability repairs, both reproduced and root-caused in scout `afk-daemon-reliability-scout` (report in that task), implemented on the daemon only — no broad runtime-backend redesign, no manual target guessing, no broad tmux sweeps.

### 1. Fail fast at startup for a confidently dead/bare-shell supervisor target
`fm_backend_target_exists` only proved the pane was **present**, so an explicit `FM_SUPERVISOR_TARGET` pointing at a bare shell (e.g. the wrong tmux window such as `firstmate:0.0`, or an agent that had exited to its login shell) passed startup validation. The composer guard then read that shell as `pending`/`unknown` and deferred every escalation until the max-defer wedge (~300 s).

The daemon's startup now also probes `fm_backend_agent_alive` (`bin/fm-supervise-daemon.sh`):
- `dead` → refuse fast (log `startup failed: ... bare shell`, exit 1, release lock + pidfile), mirroring the secondmate-liveness policy that gates on `dead` only;
- `unknown` (pi's generic `node`, or an unreadable pane) → warn and continue, so a real primary is never falsely refused.

The composer guard is unchanged — it is what correctly prevented typing into a shell; this just surfaces the problem immediately instead of at the max-defer wedge.

**Test seam:** `FM_SUPERVISOR_ALLOW_DEAD_TARGET=1` downgrades the refusal to a warn. No external signal can distinguish an isolated E2E test's *simulated* agent pane (a shell-named process drawing a composer) from a real bare shell, so the opt-out is explicit. It is never set in production, and the composer guard still prevents any injection into a shell regardless.

### 2. Bound watcher shutdown so AFK stop completes inside its budget
The daemon's cleanup `wait()`ed its watcher child without a bound. `fm-watch.sh` blocks each cycle in an external `sleep $FM_POLL` (default 15) and bash defers a trapped SIGTERM until that sleep finishes, so on `fm-afk-launch.sh stop` the daemon could block ~`FM_POLL` and exceed stop's 10 s SIGTERM budget — surfacing as `away-mode daemon did not exit after SIGTERM; preserving lifecycle state` (the daemon did exit shortly after; a later `stop` found it absent).

Cleanup now SIGKILLs the watcher after a short grace (`FM_WATCHER_SHUTDOWN_GRACE_SECS`, default 5), well inside the 10 s budget. The watcher's durable state is the on-disk wake queue and suppression markers, so a forced kill never loses work.

## Why these are the right scope
- Defect 1 root cause is **caller target selection** (wrong explicit target) + **weak startup validation** (presence-only). The proven path is `$TMUX_PANE` auto-discovery; the fix makes a wrong-but-existing target fail fast.
- Defect 2 root cause is **daemon shutdown signaling** (the watcher's deferred SIGTERM during external sleep + the daemon's unbounded wait). The "terminal-close warning" on the later `stop` is a benign consequence (the detached session auto-collapses on pane-process exit), not a separate defect.

## Verification
- Reproduced both defects in isolated homes pre-fix (defect 1: a bash pane passes `target_exists`, reads composer `pending`; defect 2: `stop` = 10.4 s with the timeout message; minimal `trap … TERM; sleep 15` fires its trap only once the external sleep completes).
- Post-fix:
  - defect 1: `stop` completes in ~6.3–6.5 s with the normal "away mode stopped" line;
  - defect 2: a bare-shell target is refused at startup (exit 1, no pidfile).
- `bin/fm-lint.sh` (pinned shellcheck 0.11.0, CI's gate): clean.
- Regression tests added:
  - `tests/fm-daemon.test.sh`: `test_daemon_refuses_bare_shell_supervisor_target`, `test_daemon_does_not_refuse_unknown_supervisor_target`, `test_daemon_allow_dead_opt_out_permits_simulated_pane` (real tmux, real daemon executed directly);
  - `tests/fm-afk-launch.test.sh`: `e2e_tmux_daemon_stop_bounded_inside_budget` (real daemon, not the sleeper placeholder, so a watcher child actually exists).
- Existing daemon/inject E2E suites pass; the simulated-agent-pane suites set the new opt-out (`tests/fm-afk-inject-e2e.test.sh`, `tests/fm-afk-inject-herdr-e2e.test.sh`, `tests/fm-afk-pi-herdr-return-e2e.test.sh`).
- Evidence recorded in `docs/tmux-backend.md` ("Agent liveness probe") and `docs/herdr-backend.md` ("Daemon shutdown bounds the watcher wait").

## Notes
- Scoped to `bin/fm-supervise-daemon.sh` (both fixes) + tests + docs. No change to the shared watcher poll loop (a broader interruptible-sleep improvement for the always-on path is noted as out of scope).
- Based on current `main` (rebased; #841 landed in the meantime).

### DIFF
--- a/bin/fm-supervise-daemon.sh
+++ b/bin/fm-supervise-daemon.sh
@@ -79,6 +79,12 @@
 #                                   supported as supervisor backends; the daemon
 #                                   refuses loudly at startup rather than trying
 #                                   tmux primitives against a non-tmux pane.
+#          FM_SUPERVISOR_ALLOW_DEAD_TARGET
+#                                   test seam: when 1, a supervisor target that
+#                                   reads agent_alive=dead (a bare shell) downgrades
+#                                   from a startup refusal to a warn. For isolated
+#                                   E2E tests that simulate an agent pane with a
+#                                   shell-named process; never set in production.
 #          FM_INJECT_SKIP           |-prefixes force-self-handle bypassing
 #                                   classification (default "heartbeat"); empty
 #                                   disables. Use sparingly: it overrides the
@@ -94,6 +100,12 @@
 #                                   (default 300)
 #          FM_HOUSEKEEPING_TICK     seconds between housekeeping passes while
 #                                   the watcher is mid-cycle (default 15)
+#          FM_WATCHER_SHUTDOWN_GRACE_SECS
+#                                   seconds cleanup waits for the watcher child to
+#                                   exit on SIGTERM before SIGKILL (default 5).
+#                                   Bounds daemon shutdown inside fm-afk-launch
+#                                   stop's budget; the watcher cannot act on
+#                                   SIGTERM while in its external poll sleep.
 #          FM_BUSY_REGEX            OR-ed busy signatures (mirrors fm-watch.sh)
 #          FM_COMPOSER_IDLE_RE      empty-composer regex applied after dim-ghost
 #                                   and structural border stripping (default:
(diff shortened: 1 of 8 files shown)


kunchenguid:

PR #865, opened 2026-07-22
docs: describe fm-guard beacon-only predicate and its blind window

ManveerBhullar:
## What

Corrects documentation that overstated the pull-based supervision guard's predicate, and adds a regression test pinning the retained behavior. No runtime change.

## Why

`docs/architecture.md` claimed `fm-guard.sh` warns when "that watcher stops running". It does not: `fm-guard.sh` keys solely off the liveness beacon mtime (`state/.last-watcher-beat` within `FM_GUARD_GRACE`, 300s) and never inspects the watcher lock. So a watcher that has **died** while its beacon is still fresh stays silent for up to `FM_GUARD_GRACE` — a bounded blind window. The strict identity-matched live-lock backstop (`fm-turnend-guard.sh`) is what catches that case at the next turn end.

This split is **intentional and retained** (per the supervision-tooling-contract-audit): the cheap pull-based beacon guard and the strict push-based turn-end guard are complementary. This change keeps the predicate exactly as-is and only makes the prose precise.

## Changes

- `docs/architecture.md`: restate `fm-guard.sh`'s predicate as beacon-staleness, name the bounded blind window, and point to the push-based turn-end guard as the identity-matched live-lock backstop. The lib header (`fm-supervision-lib.sh`), script header, and `docs/turnend-guard.md` were already accurate; only this architecture prose was loose.
- `tests/fm-guard-stale-banner.test.sh`: add `test_fresh_beacon_silent_with_dead_lock`, pinning the beacon-only predicate — `fm-guard.sh` stays silent on a fresh beacon even with a dead watcher lock present. If the guard is ever strengthened to call `fm_watcher_healthy`, this test fails, surfacing the divergence as intentional.

## Scope

No behavior, watcher timing, locks, runtime code, or unrelated docs are touched. The strict complementary behavior is already covered by `test_hook_blocks_when_dead_lock_has_fresh_beacon` in `fm-turnend-guard.test.sh`.

## Verification

- `bin/fm-test-run.sh tests/fm-guard-stale-banner.test.sh` — all pass (incl. new test)
- `bin/fm-test-run.sh tests/fm-turnend-guard.test.sh` — all pass

Shipped via direct-PR (no-mistakes pipeline agent unavailable).


kunchenguid:

PR #752, opened 2026-07-20
fix(bin): make watcher process identity immune to Linux wall-clock changes

vvizlan:
## Intent

Make Firstmate's watcher process-identity check immune to Linux wall-clock changes and contribute the fix upstream for issue #433 through the authenticated account's fork. On Linux, fm_pid_identity must derive the stable lifetime component from /proc/<pid>/stat field 22 after parsing only after the final ')' in comm, and combine it with the full NUL-separated command line so PID reuse still mismatches. Preserve the existing LC_ALL=C ps lstart fallback unchanged on macOS and systems without usable proc data, keep fm_pid_identity as the one owner, and align any test fixture that derives identity itself. Add colocated regression coverage for a simulated btime change, tricky comm parsing, full cmdline inclusion, and genuine PID reuse; bin/fm-lint.sh and affected behavior tests must pass. Delivery must target kunchenguid/firstmate from vvizlan:fm/firstmate-clock-identity-fix without adding a fork remote or pushing elsewhere. The PR must reference #433 and explain the observed WSL2 evidence: btime moved 1784094040 -> 1784094031 -> 1784094016, dmesg reported 'systemd-journald: Time jumped backwards', and one live watcher lock recorded 'Sun Jul 19 23:52:46 2026 bash .../bin/fm-watch.sh' while a reread returned 'Sun Jul 19 23:52:32 2026 bash .../bin/fm-watch.sh'. Explain why starttime is stable, that macOS remains unchanged, that an old-format lock mismatches once after upgrade and resolves through one normal re-arm with no migration code, and summarize test coverage.

## What Changed
- `fm_pid_identity` in `bin/fm-wake-lib.sh` now prefers `/proc` on Linux: it parses starttime (stat field 22, read only after the final `)` of the comm field) and combines it with the hex-encoded full NUL-separated cmdline, so wall-clock steps (e.g. WSL2 btime drift) no longer change a live watcher's identity while PID reuse still produces a mismatch. The `LC_ALL=C ps lstart` fallback is preserved unchanged for macOS and hosts without readable proc data.
- Added `FM_PROC_ROOT_OVERRIDE` (documented in `docs/configuration.md`) to point the Linux identity read at an alternate proc root, and the `fm-pr-check-security` test fixture now derives its fake watcher-lock identity through `fm_pid_identity` instead of duplicating the `ps` invocation.
[...]

PR #900, opened 2026-07-23
fix(watcher): make Linux PID identity clock-drift stable

omnimaycha:
## Intent

Prevent a live, healthy Firstmate watcher from being rejected by process-identity checks when WSL2 or another Linux host adjusts its wall-clock or boot-time reference.

## Background

On WSL2, `ps lstart` changed for the same live PID after NTP synchronization or suspend.
Observed identities shifted from `07:09:22` when recorded to `07:09:27` on a live read, and a later watcher shifted from `07:11:22` to `07:11:33`.
The drift recurred after watcher restarts, causing the turn-end guard and continuity gate to treat a healthy watcher as mismatched.

## What Changed

- Use Linux `/proc/<pid>/stat` field 22 (`starttime`, in clock ticks since boot) together with the full `/proc/<pid>/cmdline` bytes.
- Split the `stat` record after its final `)` so process names containing spaces or parentheses parse correctly.
- Fail closed on Linux when the required `/proc` identity data is unavailable instead of falling back to drift-prone `ps lstart`.
- Keep the locale-pinned `ps lstart` plus command fallback for non-Linux hosts such as macOS.
- Return failure for omitted, invalid, zero, or missing PIDs.

## Compatibility and migration

Existing consumers compare the identity as an opaque string and require no format-specific changes.
A legacy `lstart` identity mismatches once without producing a false healthy result.
The supported watcher restart arm replaces that lock with a fresh watcher and writes the stable Linux identity.

## Regression coverage

- Repeated identity reads remain stable across simulated boot-time changes.
- PID reuse changes the identity.
- Invalid and missing PIDs fail.
- Linux never consults `ps` when `/proc` data is missing.
- A process name containing a closing parenthesis parses from the final delimiter.
- A live watcher lock remains matched while simulated `ps lstart` output shifts by 11 seconds.
- A real watcher with a legacy identity is rejected safely and repaired by `fm-watch-arm.sh --restart`.

## Validation

- `bin/fm-test-run.sh --changed --base origin/main`: 10 watcher/wake/lock scripts passed with 0 failures.
- `tests/fm-watcher-lock.test.sh`: passed on WSL2, including the live lock and legacy migration cases.
- `bin/fm-lint.sh`: passed with the repository-pinned ShellCheck 0.11.0 across the canonical script set.
- No-mistakes automated review, targeted test, documentation, and lint validation completed successfully before its upstream push failed because this contributor has read-only access.

## Risk assessment

Low.
The behavior change is limited to Linux process-identity collection and strengthens fail-closed handling.
The non-Linux fallback and opaque consumer contract remain intact.

### DIFF
--- a/bin/fm-wake-lib.sh
+++ b/bin/fm-wake-lib.sh
@@ -24,17 +24,19 @@ fm_pid_alive() {
 }
 
 fm_pid_identity() {
-  local pid=$1 out proc_root stat_line starttime cmdline_hex
+  local pid=${1:-} out proc_root stat_line starttime cmdline_hex
   local -a stat_fields
   case "$pid" in
-    ''|*[!0-9]*) return 1 ;;
+    ''|*[!0-9]*|0) return 1 ;;
   esac
   proc_root=${FM_PROC_ROOT_OVERRIDE:-/proc}
   # Prefer /proc on Linux: stat field 22 (starttime, clock ticks since boot) is
   # immune to the wall-clock steps that re-render the ps lstart fallback's date
   # (observed as WSL2 btime drift) and would evict a live watcher; combining the
   # full NUL-separated cmdline keeps PID reuse a mismatch even on a tick collision.
-  if [ "$(uname)" = Linux ] && [ -r "$proc_root/$pid/stat" ] && [ -r "$proc_root/$pid/cmdline" ]; then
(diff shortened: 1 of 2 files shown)


omnimaycha:
Superseded: the equivalent clock-drift-stable PID identity fix landed on main via #752 (a26b37c). Closing this fork PR as redundant.

Issue #883, opened 2026-07-22
herdr backend: prompt-glyph matching breaks under LC_ALL=C (bracket expression + ${content#??} operate on bytes)


marninms98-dotcom:
Under a C/POSIX locale, \`FM_BACKEND_HERDR_BARE_PROMPT_RE='^[❯›]'\` degrades from characters to bytes: a leading 0xE2 matches any box-drawing glyph (╭ ╰ │), so a border row is misread as a bare composer. Similarly \`\${content#??}\`/\`\${content#?}\` strips bytes, mangling the multibyte tail so an idle placeholder reads as pending.

Fix that worked for us: alternation instead of a bracket expression (\`'^(❯|›)'\`) and stripping each glyph by literal pattern rather than by \`?\` counts. Found via a locale-dependent test failure; reproducible with \`LC_ALL=C\` on macOS bash 3.2.

PR #929, opened 2026-07-23
fix(composer): byte-safe multibyte prompt-glyph matching under C locales

Doctor-Jayy:
Fixes both halves of #883: `LC_ALL=C` byte-safety in the composer classifier and in Herdr's bare-prompt-glyph regex.

## Problem

Under a C/POSIX locale on macOS Bash 3.2:

1. `fm_composer_classify_content` (`bin/fm-composer-lib.sh`) stripped a leading prompt glyph with `${content#??}`/`${content#?}` character-count patterns. Under `LC_ALL=C` those patterns strip **bytes**, not characters, so the 3-byte multibyte glyphs `❯` (claude) and `›` (codex) split mid-codepoint. An idle, empty composer (`❯ `) misclassified as `pending` instead of `empty`.
2. `FM_BACKEND_HERDR_BARE_PROMPT_RE` (`bin/backends/herdr.sh`) defaulted to the bracket expression `'^[❯›]'`. Under `LC_ALL=C`, a bracket expression decomposes into a byte-class - `[❯›]` becomes the raw UTF-8 byte set `{0xE2,0x9D,0xAF,0x80,0xBA}`. Since `0xE2` is the leading byte of *every* UTF-8 box-drawing character, not just `❯`/`›`, this wrongly matched the composer's own border rows (`╭`, `╰`, `│`, `╯`) as a bare prompt row, corrupting `fm_backend_herdr_composer_state`'s structural row scan.

Both are the same class of bug (character-count/class patterns implicitly assuming a multibyte-aware locale) against the same underlying data (the agent prompt glyphs), so both are fixed here together.

## Fix

1. Strip each prompt glyph by its literal byte sequence (`case ... in '❯'*) content=${content#'❯'} ;; ...`) instead of a character-count pattern. ASCII single-byte prompts (`>`, `$`, `%`, `#`) are handled the same way for consistency, though they were never affected by the original bug.
2. Change `FM_BACKEND_HERDR_BARE_PROMPT_RE` from the bracket expression to an alternation (`'^(❯|›)'`), which matches each glyph as a whole literal unit and is not vulnerable to byte-class decomposition.

Both fixes preserve existing behavior in a UTF-8 locale exactly (verified below) and do not touch the maintainer-selected zero-to-eight-row Pi composer classification logic.

## Testing

Reproduced and verified on macOS system Bash 3.2.57 under `LC_ALL=C`:

```
                 input=[❯ ]        input=[❯ hello]   input=[> ]
before fix        pending  (WRONG)  pending           empty
after fix         empty    (right)  pending           empty
```

- Added `test_agent_glyphs_strip_literally_under_c_locale` to `tests/fm-composer-lib.test.sh`, covering both `❯` and `›` under `LC_ALL=C` for both the empty and pending cases.
- Added `test_composer_state_bare_prompt_is_empty_under_c_locale` to `tests/fm-backend-herdr.test.sh`, pinning the exact border-row-misclassification regression under `LC_ALL=C`.
- Full existing suites (`tests/fm-composer-lib.test.sh`, `tests/fm-composer-ghost.test.sh`, `tests/fm-backend-herdr.test.sh`, `tests/fm-backend-cmux.test.sh`, `tests/fm-backend-orca.test.sh`, `tests/fm-daemon.test.sh`) pass, including under `LC_ALL=C LANG=C`.
- `shellcheck bin/*.sh bin/backends/*.sh tests/*.sh` passes clean.
- One-sentence doc updates to `docs/herdr-backend.md` and `docs/configuration.md` recording both fixes and their regression coverage.


kunchenguid:

Issue #479, opened 2026-07-11
Windows and Antigravity support


sendev1:
Hi,

I watched your video and am very interested in trying out this repository on Windows 11 with Antigravity. It appears this repo is tightly coupled with macOS/Linux. I have been trying to get it to work on my setup, but I am unable to successfully complete the setup on Windows.Could you please provide guidance or updates to support Windows/WSL environments?

julesLubrano:
Second independent WSL2 user request for Antigravity support.

My concrete need is narrower than making Antigravity a Firstmate primary: I want `agy` to be a verified selectable crewmate/scout harness that Firstmate can spawn, steer, supervise, interrupt, relaunch, and clean up like its existing worker harnesses.

The practical motivation is model access. Gemini 3.7 is already available through my authenticated Antigravity setup, but Firstmate cannot currently select that worker directly, so using the model requires awkward proxy/provider configuration instead of the Antigravity account and CLI I already have.

For an initial useful scope, crewmate/scout support on WSL2 is enough. Primary and persistent secondmate support can remain out of scope. The important behaviors are:

- verified WSL2 launch and process detection;
- explicit model selection, including the Gemini 3.7 models exposed by the installed `agy` catalog;
- permission/trust handling without silently mutating unrelated user configuration;
- reliable busy/idle and completion signaling;
- durable steering plus verified interrupt, exit, and relaunch behavior;
- safe cleanup that preserves unlanded work and removes only task-owned state;
- loud refusal rather than fallback when Antigravity or the requested model is unavailable.

PR #3008 appears closely aligned: it proposes a verified `agy` worker adapter and demonstrates `gemini-3.7-flash-high` against the authenticated CLI. I am adding this comment rather than opening a duplicate issue to show that the request now comes from multiple users and that WSL2 support is part of the demand.

The Pi custom-provider approach explored in #2998 may be a useful alternative for some setups, but it does not replace the zero-friction use case here: select the existing Antigravity CLI/account as the worker and let Firstmate manage it directly.

kunchenguid:
Speaking as Kun's firstmate:

**Triage:** Two asks in one issue. Antigravity-as-verified worker is VISION-aligned; open PR #3008 is the actual current help. Native Windows as a first-class host is a separate class already queued on #1796 and (this pass) #1667. Stamping `existing-pr`. Not adding `ready-for-pr`. No competing implementation PR opened.

[...]

PR #92, opened 2026-06-26
Add color to command center

Sahag117:
Summary:
- add ANSI color to the command center dashboard for headers, dividers, task ids, labels, and pane/session states
- keep redirected output plain by default, with FORCE_COLOR/CLICOLOR_FORCE opt-in and NO_COLOR opt-out
- add focused shell coverage for the color behavior

Verification:
- bash -n bin/fm-command-center.sh
- bash -n tests/fm-command-center-color.test.sh
- bash tests/fm-command-center-color.test.sh
- for test_script in tests/*.test.sh; do "$test_script"; done

Not run:
- shellcheck bin/fm-command-center.sh tests/fm-command-center-color.test.sh (shellcheck is not installed in this environment)

PR #934, opened 2026-07-23
fix: refine scout intake and parallel dispatch

kunchenguid:
## Intent

Clarify Firstmate's existing AGENTS.md intake contract so established reports and evidence answer informational questions without redundant design-only scouts; unclear implementation intent gets a concise implementation question rather than speculative design; authorized implementation keeps bounded research inside the ship unless uncertainty could materially change whether or what to build; genuine separately requested or decision-changing scouts remain valid and are promoted after separate implementation authorization. Also clarify the adjacent concurrency contract so file or subsystem overlap is a risk signal rather than an automatic wait: isolated, independently validatable work runs immediately with no concurrency cap when its selected delivery path can reconcile ordinary rebases or conflicts, while only concrete semantic dependencies, unsafe shared mutable external state, incompatible migrations, or equivalent reconciliation hazards serialize. Preserve information-versus-implementation authority, durable genuine blockers, all delivery paths, the existing section ownership, one sentence per Markdown line, and neutral AGENTS.md line count. Run this in parallel with the other AGENTS.md branch, let no-mistakes own rebasing and conflict handling, open a green PR, and leave it unmerged for captain review.

## What Changed

- Reuse established reports and evidence, keep bounded research within authorized ships, and reserve separate scouts for explicit knowledge requests or decision-changing uncertainty.
- Treat file and subsystem overlap as a risk signal, allowing independently validatable work to proceed concurrently while serializing concrete reconciliation hazards.
- Align delegation-guard guidance and documentation with intake classification, with regression coverage for the updated contracts and denial messages.

## Risk Assessment

✅ Low: The change is well-bounded, preserves the stated intake and concurrency contracts, and the follow-up consistently removes the previously conflicting scout-routing path across source, documentation, and focused regression coverage.

## Testing

Validated the five-file change with focused contract and guard tests, manual review of the intake and concurrency language, Markdown invariants, equal 484-line base and target counts, and an end-user CLI transcript for both scout configurations; all local checks passed, the worktree stayed clean, and PR creation plus remote CI remain for no-mistakes.

<details>
<summary>Evidence: Primary-session intake denial transcript</summary>

<code>The primary-session guard denied a delegation-shaped `Agent` call with exit 2. With `fm-scout.sh` present, it required intake classification first and reserved the scout path for already-classified scout work. Without it, it deferred to intake classification and used brief-then-spawn.</code>

```text
PRIMARY SESSION DENIAL - SCOUT ENTRY POINT PRESENT
{"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecision":"deny"},"systemMessage":"[subagent-dispatch] the firstmate primary dispatches through the fleet, not the harness's own delegation tools: work started that way has no durable fleet record, leaves every firstmate guard inert, and dies with this session. Instead, first classify the work under the AGENTS.md intake contract: work already classified as a scout goes to bin/fm-scout.sh \"<question>\" [project], while authorized ship work and its bounded research go to bin/fm-brief.sh then bin/fm-spawn.sh (blocked tool: Agent, delegation-shaped on \"agent\"). Launch the session with FM_ALLOW_SUBAGENT=1 for a deliberate exception."}
exit_status=2

PRIMARY SESSION DENIAL - SCOUT ENTRY POINT ABSENT
{"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecision":"deny"},"systemMessage":"[subagent-dispatch] the firstmate primary dispatches through the fleet, not the harness's own delegation tools: work started that way has no durable fleet record, leaves every firstmate guard inert, and dies with this session. Instead, first classify the work under the AGENTS.md intake contract, then use bin/fm-brief.sh followed by bin/fm-spawn.sh for dispatched work (blocked tool: Agent, delegation-shaped on \"agent\"). Launch the session with FM_ALLOW_SUBAGENT=1 for a deliberate exception."}
exit_status=2
```
</details>

## Pipeline

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>

- 🚨 `AGENTS.md:232` - The required behavior says “authorized implementation keeps bounded research inside the ship,” but the existing supported correction path in `bin/fm-subagent-pretool-check.sh:179-185` still tells a primary that all “investigation or diagnosis goes to bin/fm-scout.sh” whenever that entry point exists. Thus an authorized fix involving bounded diagnosis can still be redirected into the redundant separate scout this change forbids. Update that corrective route and its tests/docs to defer to the intake classification, reserving `fm-scout.sh` for work already classified as a scout.

🔧 Fix: Align scout guard with intake classification

</details>

<details>

✅ No issues found.
- `Inspected the complete five-file diff and manually checked changed Markdown for one sentence per physical line.`
- `bash tests/fm-instruction-owners.test.sh`
- `bash tests/fm-subagent-pretool-check.test.sh`
- `git diff --check ca4be1ca7fc4d2dca1614b9336f5080ee7ca5076..4f96a051083b233e4a9633d02b1bcfde650c2066 -- AGENTS.md docs/subagent-guard.md`
- <code>Compared base and target `AGENTS.md` line counts and confirmed 484 lines at both revisions.</code>
- <code>Invoked `./bin/fm-subagent-pretool-check.sh --claude --tool Agent` against isolated primary-home fixtures with and without `bin/fm-scout.sh`, capturing both denial surfaces and exit status 2.</code>
- <code>`gh-axi pr list --state all --head fm/fm-avoid-redundant-design-scouts-r1 --limit 10` confirmed no PR exists yet; PR creation and green remote CI remain owned by the later no-mistakes pipeline stage.</code>
- <code>`git status --short --branch` confirmed testing left the target worktree clean.</code>

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -226,17 +226,19 @@ Keep `local-only` work in the main home.
 Send in-scope work to the fitting secondmate unless it is blocked or the captain explicitly redirects it; do not read the secondmate's chat because marked routed replies return through its status or referenced document.
 If no secondmate scope fits, use the main home or discuss creating an appropriate persistent secondmate.
 
+Before commissioning an investigation, consult existing reports and established evidence.
 Classify the deliverable:
 
-- **Ship** is the default and produces a project change through the selected delivery mode.
-- **Scout** produces knowledge in `data/<id>/report.md`, never a PR, and is the default for investigation, diagnosis, planning, reproduction, or audit requests that do not clearly include implementation.
+- **Ship** is the default and produces a project change through the selected delivery mode; once implementation is authorized, dispatch a ship and keep any remaining bounded research inside it unless unresolved uncertainty could materially change whether or what to build.
+- **Scout** produces knowledge in `data/<id>/report.md`, never a PR, and is appropriate for investigation, diagnosis, planning, reproduction, or audit work when the captain explicitly requests a separate knowledge or design deliverable or unresolved uncertainty could materially change whether or what to build.
 
+If established evidence already answers an informational question, relay it without a design-only scout; when implementation intent is unclear, answer and ask one concise implementation question when useful rather than dispatching speculative design work.
+Never both present a likely-enough solution and launch a parallel design exercise that is not expected to change it.
 A diagnostic request, report, recommendation, or implementation-ready finding is evidence, not authorization to change code.
-Implementation requires a separate request or other clear implementation scope.
 Load `diagnostic-reasoning` before scoping a reported bug and before acting on a diagnostic report.
 
-Classify work as dispatchable when it does not overlap work under way, or queued and blocked when it touches the same project subsystem or depends on unlanded work.
-Dispatch independent work immediately with no concurrency cap, serialize coarse overlaps, and record blockers durably.
+Treat file or subsystem overlap as a risk signal rather than an automatic reason to wait, and dispatch isolated work immediately with no concurrency cap when each change can be independently implemented and validated and the selected delivery path can reconcile ordinary rebases or conflicts.
+Serialize only for a true semantic dependency, shared mutable external state, incompatible concurrent migration, or another concrete condition that makes independent progress or reconciliation unsafe; same-file editing alone is insufficient, and genuine blockers remain durable.
 Write the task-specific brief under section 11 before spawning.
 
 ### Dispatch and supervision handoff
@@ -305,13 +307,11 @@ Retire one only on an explicit captain or main-firstmate decision, after loading
 
 ### Scout outcome and promotion
 
-A completed scout must leave a self-contained report before its scratch worktree can be discarded.
-Read the report, relay its findings rather than merely saying it finished, record the report as the Done artifact, and re-evaluate the queue.
+A completed scout must leave a self-contained report before its scratch worktree can be discarded; read and relay its findings, record the report as the Done artifact, and re-evaluate the queue.
 A report may recommend implementation but does not authorize it.
 Before treating the investigation or any visual review as complete, load `decision-hold-lifecycle`; teardown enforces that shared completion gate.
 When implementation is separately authorized, promote the existing scout through `bin/fm-promote.sh` rather than creating a duplicate task.
-The promoted worker must inventory scratch state, return to a clean default-branch base, carry over only intended fix changes, create the ship branch, and follow the project's selected delivery path.
-Scratch commits and debug edits never ride along, and a reproduced bug becomes the regression test.
+The promoted worker must inventory scratch state, return to a clean default-branch base, carry over only intended fix changes, create the ship branch, and follow the project's selected delivery path while leaving scratch commits and debug edits behind and turning a reproduced bug into the regression test.
 
 ## 8. Supervision protocol
 
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ Launching a supported harness inside it instantiates your first mate - and makes
 - **One liaison** - you talk only to the first mate; it dispatches, supervises, escalates only real decisions, and reports plain outcomes.
 - **A visible crew** - every crewmate works in its own tmux window, experimental herdr/zellij tab, cmux workspace, or Orca terminal you can watch or type into; the first mate reconciles.
 - **Disposable worktrees** - each task runs in a clean [treehouse](https://github.com/kunchenguid/treehouse) git worktree, or an Orca-managed worktree when `backend=orca`, so parallel work on one repo never collides.
-- **Two task shapes** - ship tasks deliver a change; scout tasks investigate, plan, reproduce, or audit and leave a report.
+- **Two task shapes** - ship tasks deliver authorized changes; scout tasks leave standalone investigation reports when the intake contract warrants separate research.
 - **Explicit project modes** - each project ships via `no-mistakes`, `direct-PR`, or `local-only`, with an optional `+yolo` autonomy flag.
 - **Optional secondmates** - opt in to persistent second mates that run from isolated firstmate homes with their own `FM_HOME`, state, projects, and session lock, supervising project clones or a project-less firstmate-repo domain, kept on the primary firstmate version by guarded local fast-forwards and checked for live agent processes at session start.
 - **Event-driven, zero-token supervision** - a bash watcher sleeps on the fleet and wakes the first mate only when something needs you; verified primary harnesses also get a turn-end backstop that blocks or follows up on a blind stop when work is under way and supervision is not live.
--- a/bin/fm-subagent-pretool-check.sh
+++ b/bin/fm-subagent-pretool-check.sh
@@ -173,13 +173,13 @@ STATE=${FM_STATE_OVERRIDE:-$FM_HOME/state}
 . "$SCRIPT_DIR/fm-primary-scope-lib.sh"
 fm_primary_scope_matches "$FM_ROOT" "$STATE" || exit 0
 
-# Investigation has a dedicated entry point when this home carries it; degrade
+# Name the dedicated scout entry point only when this home carries it; degrade
 # to the two-step brief-then-spawn path when it does not, rather than naming a
 # script that is not there.
 if [ -f "$FM_ROOT/bin/fm-scout.sh" ]; then
-  ROUTE='investigation or diagnosis goes to bin/fm-scout.sh "<question>" [project], and ship work goes to bin/fm-brief.sh then bin/fm-spawn.sh'
+  ROUTE='first classify the work under the AGENTS.md intake contract: work already classified as a scout goes to bin/fm-scout.sh "<question>" [project], while authorized ship work and its bounded research go to bin/fm-brief.sh then bin/fm-spawn.sh'
 else
-  ROUTE='investigation and ship work both go to bin/fm-brief.sh then bin/fm-spawn.sh'
+  ROUTE='first classify the work under the AGENTS.md intake contract, then use bin/fm-brief.sh followed by bin/fm-spawn.sh for dispatched work'
 fi
 
 REASON="[subagent-dispatch] the firstmate primary dispatches through the fleet, not the harness's own delegation tools: work started that way has no durable fleet record, leaves every firstmate guard inert, and dies with this session. Instead, $ROUTE (blocked tool: $TOOL, delegation-shaped on \"$MATCHED\"). Launch the session with FM_ALLOW_SUBAGENT=1 for a deliberate exception."
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -128,7 +128,8 @@ The helper's header owns the exact signal detection, relocated-home limitation, 
 
 ## Two task shapes
 
-Ship tasks change projects and ship by project mode (`no-mistakes`, `direct-PR`, or `local-only`); scout tasks investigate, plan, reproduce bugs, or audit, then leave a report at `data/<id>/report.md` and never push.
+Ship tasks change projects and ship by project mode (`no-mistakes`, `direct-PR`, or `local-only`); scout tasks leave standalone investigation reports at `data/<id>/report.md` and never push.
+The intake and authority contract in `AGENTS.md` owns when separate scout research is warranted.
 
 ## Dispatch profiles
 
--- a/docs/subagent-guard.md
+++ b/docs/subagent-guard.md
@@ -161,8 +161,8 @@ A tool removed from the schema stays removed, so a genuinely intended use of a l
 - Malformed or empty stdin, invalid JSON, a payload with no tool name, and missing `jq` for stdin transport all fail open with exit 0 and no output.
 
 The deny message names the real dispatch path.
-When `bin/fm-scout.sh` exists in the home it routes investigation and diagnosis there and ship work to `bin/fm-brief.sh` then `bin/fm-spawn.sh`.
-When that script is absent the message degrades to naming `bin/fm-brief.sh` then `bin/fm-spawn.sh` for both, rather than pointing at a script that is not there.
+When `bin/fm-scout.sh` exists in the home the message first defers to the `AGENTS.md` intake classification, then routes work already classified as a scout there and authorized ship work with its bounded research to `bin/fm-brief.sh` then `bin/fm-spawn.sh`.
+When that script is absent the message still defers to intake classification and degrades to naming `bin/fm-brief.sh` then `bin/fm-spawn.sh` for dispatched work, rather than pointing at a script that is not there.
 
 ## Harness wiring
 
--- a/tests/fm-instruction-owners.test.sh
+++ b/tests/fm-instruction-owners.test.sh
@@ -192,6 +192,33 @@ test_compressed_agents_owner_map() {
   pass "compressed AGENTS.md records the approved one-owner map"
 }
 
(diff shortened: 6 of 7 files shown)


Boxyboxy:
## Reconciliation of every numbered finding you filed on this PR

**8 FIXED / 1 REFUTED / 3 DEFERRED.**

Your four review comments carry 17 numbered headings, which dedupe to **12 distinct findings**.
All 12 have exactly one outcome below, with the evidence measured at `cd0b2bf2b56c29f9b51c796a1d06ede0dbb94745` (= `origin/feat/campaigns`, including round 7).
Bodies verified complete by exact character count before reading: 63873 / 6709 / 17937 / 33092, all four MATCH.
Note for anyone re-checking: `wc -c` gives 64081 / 6730 / 17989 / 33181 - that is bytes, inflated by the severity emoji and the `·` separators.

**Merge map** (5 headings folded, all by byte-identical title plus matching tracking id):
the krewlib ordering blocker appears three times (comment 1 #1, comment 3 #1, comment 4 #2; `R1-02` = `F1`);
the emails ordering blocker three times (1 #2, 3 #2, 4 #3; `R1-03` = `F2`);
the deferred-until label twice (1 #4, 3 #3; `U61`).

**Your grades moved between rounds on the same items, always downward**, and we treated the grade as information rather than as a disposition: krewlib and emails ordering both went blocker to major to major; deferred-until went minor to nit "accepted tradeoff". Three more were self-downgraded inside their own Verification block before any fix existed (preview replay, the one-definition suite, dev_mode). An item graded nit and marked "accepted tradeoff" was still open until this pass wrote down what happened to it.

Six of the twelve were closed in our ledger only by your stated convention of dropping settled items rather than restating them. We do not accept that as evidence, so all six are re-proved below with a commit, a test, and in five of six a quoted red.

---

### FIXED (8)

**`R1-A3` contributor start-time join takes the route's default 100-row page** - minor.
The unpaged read was removed rather than paged: `DebtorContactBudgetWidget.tsx` no longer imports or calls `listCampaignInstances` at all, and `:318-392` is a bounded worker pool issuing one `getCampaignInstance(id)` per contributor id named by the budget payload, so there is no page and no unread tail.
Commit `a932d97a9`; guarded by three `expect(listCampaignInstances).not.toHaveBeenCalled()` assertions (`DebtorContactBudgetWidget.test.tsx:228`, `:302`, `:329`) plus `:262 "labels a contributor buried deep in the consumer's history"`, which is your own >100-instance scenario. Reverting to the single unpaged read reds 6 of 32: `labels a contributor buried deep...` gives `Expected length: 2 / Received length: 0`.

**`U61` deferred-until label resolves the zone from state alone** - minor, then nit.
Core fixed: `nextEligibleLabel.ts:199` now returns `from <clock> <abbrev>, based on the account state`, naming the basis instead of asserting the consumer's zone. Commits `5ee340295` + `bbd83173a`; guard `nextEligibleLabel.test.ts:136 "names the BASIS of the zone instead of asserting the consumer's"`. Deleting the qualifier reds 9 of 21, that guard reporting `Expected pattern: /account state/i / Received string: "from 12:30 PM PDT"`.
The residual (resolve ZIP-first on the sidebar, which does hold `postal_code`) is **refuted in crons code, not in a comment**: `compliance/zip_codes.py:1` maps a ZIP to a two-letter STATE, verified `ZIPS_DICT["32501"] == "FL"`, so ZIP-first yields "FL" on your own Pensacola example - identical to the account state, a literal no-op; and `compliance/phone_time_compliance.py:28` walks `STATE_TIMEZONES[state]`, the state's full zone list, returning `max(starts), min(ends)` at `:53`, so a per-consumer zone is not a value that gate computes.

**`R1-A5` responses landing after the dialog closes replay an abandoned run** - minor.
A session token now guards all three legs you named: `dialogSessionRef` at `DashboardEnrollmentModals.tsx:300`, bumped at `:385`, compared at `:655` (calls), `:832` (workflow) and `:970` (campaign confirm, deliberately narrower - the toast still fires because the bulk write already happened, only the results screen is withheld).
Commit `9b75b8e75`; guard `DashboardEnrollmentModals.test.tsx:1759 describe("a response that lands after its dialog closed")`. Deleting `:832` reds it with `Unable to find an accessible element with the role "dialog" and name "Workflow filters"` - the next launch opening on the abandoned run's confirm screen, exactly as you described - while the positive control `:1811 "still shows a preview that arrives while the dialog is OPEN"` stays green.

**`R1-A6` the whole Campaigns-tab suite ran on one definition group** - minor.
`makeInstance` gained an `overrides` parameter (`CampaignInstancesTab.test.tsx:243-253`, quoting your finding verbatim) and a second group exists at `:4440-4441` (`definition-2` / "Second notice", your own example name).
Commit `a37f5e2e8`; guard `:4435 describe("selection is scoped to one definition")`, four cases.
**Your own proposed mutation now reds**: deleting the reset ternary's else-branch at `CampaignInstancesTab.tsx:990-993` so `toggleRow` always spreads gives `halts only the definition the operator last selected, never both` - `Expected number of calls: 1 / Received number of calls: 2`, the terminal bulk halt firing into the unselected group - plus `clears the other definition's row tick when selection moves`.
One honest residual: we also deleted the `current.definitionId === definitionId` conjunct from `toggleGroup`'s `allSelected` and all four cases **stayed green**. The fixing round had already measured that and said so at `test:4545-4551` ("no input distinguishes the term being present from absent ... deleting that term leaves all 66 cases green"). We reproduced it. Two of the three pieces you named are now guarded with a red; the third is unreachable-by-construction defence in depth behind the guarded reset, and is recorded as such rather than claimed as covered.

**`R2-10` contact-budget contract omits `jurisdiction_unresolved` and inverts its JSDoc** - minor.
All four parts closed by `5a2f5a248`: the field is declared at `instances.ts:240` with the `blocked_states`-is-an-empty-intersection trap named at `:229-239`; the `limits_degraded` JSDoc is corrected at `:285-311`; the `partial_jurisdiction_count` JSDoc at `:450-458` now names no specific signal and says why; and the two stale reader sites you flagged are corrected (`DebtorContactBudgetWidget.tsx:523-535`, and the fixture at `DebtorContactBudgetWidget.test.tsx:1064-1065` now carries the live backend sentence instead of the pre-`f59d8a80` "No postal code on file...").
The guard is a consumed type pin - `type BlockedIsDeclared` at `instances.test.ts:47-70`, asserted through `const campaignContractPins: [BlockedIsDeclared, ...] = [true, ...]` at `:234-271`.
**One thing worth your knowing**: `jest.config.js:50-64` runs ts-jest with `isolatedModules: true`, so this pin cannot red under jest - the suite is green with or without the field. It reds only under a real typecheck. Proved under isolated `tsc`: deleting `jurisdiction_unresolved: boolean;` produces exactly one error, `TS2322: Type 'true' is not assignable to type 'never'`, at the `BlockedIsDeclared` slot.

**`R3-A8` a halted campaign renders its held step as an ACTIVE contact-window wait** - minor.
Fixed in the shape your finding argued for: terminality is folded in a separate function, not pushed into `classifyTimelineStatus`. `queuedWorkflowProgress.ts:409-415` `resolveRunEndedStatusClass` maps `waiting` to a new `waiting_ended` class only when the run ended; `IS_SETTLED` is a total `Record<TimelineStatusClass, boolean>` at `:474` with `waiting_ended: true` (`:498`) while `compliance` (`:507`) and `legal_halt` (`:509`) stay false; the timeline applies it at `QueuedWorkflowTimeline.tsx:520-536` with `runEnded` required on the campaign arm of the props union (`:257`).
Commit `b09bc58bc`; guard `queuedWorkflowProgress.test.ts:670 describe("resolveRunEndedStatusClass")`. Neutering the fold reds 2 of 61: `moves the contact-window hold, and only that` gives `Expected: "waiting_ended" / Received: "waiting"`, and `does settle the hold it moves, without calling it done` gives `Expected: true / Received: false`. The standing over-settle control (`IS_SETTLED.legal_halt` false to true) also reds 2.
Residual: no test renders the timeline with `runEnded={true}`, so the composed on-screen outcome rests on two composed unit guards plus the wire assertion at `WorkflowTimelineModal.test.tsx:937`, not on one render assertion.

**`R5-01` enroll gate blocks two debt-level flags the backend does not enforce** - major.
`DebtorCampaignsCard.tsx:170-175` is now exactly the backend's four `LEGAL_HALT_ENROLLMENT_FLAGS`; `disputed` and `refuse_to_pay` are gone. Commit `58fcc1987` (#1225); guard `DebtorCampaignsCardLegalState.test.tsx:419 "leaves enrollment available when only the two debt-level flags are set"`.
Re-adding `"disputed"` reds exactly that one test of 19: `Received element is not enabled: <button ... disabled:cursor-not-allowed ...>`. Adding `"dnc"` reds `"leaves enrollment available when only channel opt-outs are set"`, so the narrowing is pinned against re-widening too. The positive controls (`:384`, `:578`) are green, so the four enforced holds still refuse enrolment - the gate was narrowed, not neutered.

**`R5-05` "Stopped campaigns" caption asserts a compliance stop over a completed run** - minor.
`DebtorCampaignsCard.tsx:1505` gates the caption on `stoppedHasComplianceStop`, computed at `:643-651` through `campaignStopCause(...).kind === "compliance"` - the single owner, so the caption and the row pill cannot disagree; `operator` and `unstated` halts earn no caption either. Commit `574c95923`, which landed at 02:40:58Z, ahead of the 02:59:31Z comment reporting it.
Guard `DebtorCampaignsCardLegalState.test.tsx:601 describe("stopped-section caption")`, and it reds in **both** directions: changing the gate to `stoppedEntries.length > 0` reds the three negative cases, each quoting back `Stopped campaignsThese are not finished work. A compliance stop is a condition the platform is still enforcing.`; a constant `false` reds the positive case `"states the compliance stop when the group holds one, beside a completed row"` with `Expected element to have text content: / Received:` empty.

### REFUTED (1)

**`R5-04` the hot-table lock-window fix is not in the krewlib revision consumers pin** - major.
Refuted **for this repository only**; the upstream mechanism is not disputed. The file your finding names, `requirements.txt`, does not exist here, and this repository has no Python manifest of any kind. Its only dependency manifests are `package.json`, `package-lock.json`, `lambda/package.json` and `lambda/package-lock.json`, and `krewlib` occurs 0 times in all four; 0 times under `.github/` as well. Pin-shaped krewlib strings sit in exactly two non-manifest files: `__tests__/fixtures/invariants.yaml` (14 lines of kanon observation records about *other* repos' pins, sha256-byte-pinned at `__tests__/contracts/invariants-pin.test.ts:166-168`) and `docs/pci/vulnerability-remediation-example.md:120`, a context line inside a fenced diff block describing another repo's `requirements.txt` and naming `@main`, not a revision. Nothing here resolves, installs or applies a krewlib revision, so it cannot pin the wrong one.
Correcting our own prior summary rather than repeating it: 75 non-fixture files here *do* contain the string `krewlib`, as prose citations of krewlib symbols in docblocks (e.g. `app/api/campaigns/enrollmentRefusal.ts:18` cites krewlib `d9f5b2c83e01`). The zero is true of **pin sites** specifically.

### DEFERRED (3)

**`R1-02` krewlib #92 before crons #479** - blocker, then major.
Owner: the release sequencer, plus the krewlib and crons lanes. Reason, measured at this head: nothing in this repository participates in that ordering. Same census as `R5-04` - no Python manifest, zero `krewlib` in any of the four manifests, zero under `.github/`, pin-shaped strings only in the byte-pinned kanon fixture and one PCI doc example. This repository installs no krewlib, applies no Alembic chain and boots no strategy worker, so no frontend change could close it and we are not inventing a mitigation that would look like one. What is owed upstream, unchanged: a sentence in krewlib `AGENTS.md` naming `d9f5b2c83e01`, `claim_due_portfolios(0,0)`, `22023` and the crons-first direction, beside the existing consumer-contract-break bullet - better, a check in the promotion path, since `apply-main` fires on push to main with `protection_rules: []`.

**`R1-03` emails #454 before crons #479** - blocker, then major "accepted tradeoff".
Owner: the release sequencer, plus the emails and crons lanes. Same reason and same census; the named artifacts (`dispatch_contract_route.py`, `strategy_worker.py:444-447`) are not in this repository. Two control gaps stay open upstream: the obligation is stated only in crons while the releaser who can violate it works in emails, whose `AGENTS.md:121` still says the pin "is one-directional"; and the coupling is a hand-copy both ways, with crons `tests/test_terminal_causal_refusal_vocabulary.py:55-64` transcribing the eight flags as a literal so a ninth token reds nothing. We are also carrying forward a caution rather than resolving it: the emails-side "measured FALSE" refutation was taken on the incoming branch and concerns reason delivery, not vocabulary membership, so it does not close this.

**`R1-A9` dev_mode on campaign enrollment is gated only in the browser** - minor, downgraded from major in your own verification.
Owner: the krew-fastapi campaigns lane. Not fixed and not refuted: at this head `DashboardEnrollmentModals.tsx:908` and `:959` still send `dev_mode: isKrewAdmin && devModeEnabled ? true : undefined`, so your assertion still holds as written. Deferred because it cannot be closed here - the attack you describe is a hand-rolled POST carrying a session JWT, and no change to the client affects what the route accepts. The specific change owed is a role dependency on the campaigns router (`src/routes/secure/campaigns/routes.py`, currently `Depends(auth.validate_token)` only) or a `dev_mode` refusal on credentialed tokens where it is declared (`models.py`, `dev_mode: bool = False` on both enrollment requests). The downgrade rationale - that the untouched `/omnichannel-queue/confirm` is strictly larger and that `collector_users` has no role column - is a good reason not to call this a regression introduced here, and not a reason to call it closed.

---

### Method, since several of these were coverage findings

Everything above was measured from `git show cd0b2bf2b:<path>` blobs rather than a working tree, because sibling lenses were mutating tracked files throughout. Every mutation proof ran in a pristine `git archive cd0b2bf2b` export, one suite at a time; each mutated file was afterwards restored and diffed byte-identical against its blob. Nothing in the repository was edited to produce this reconciliation.

Nothing was closed by relaxing a gate, a threshold or an assertion. `R5-01` was the one to watch, since closing it meant narrowing a gate; that narrowing is bounded by measurement in both directions, and restoring the six-flag strictness is the over-block, not the fix.

Thank you for the ordering blockers in particular - they are the two we cannot close here, and they are the two most worth a human ratifying rather than a reviewer re-filing.

---

### Two notes the reconciliation raised, closed here

- **`R2-10`'s guard is a type pin, and it IS covered in CI.** The reconciliation correctly flagged that a
  green `jest` run on `instances.test.ts` proves nothing about that contract, because this repo runs ts-jest
  with `isolatedModules: true` and so does not typecheck. The repo gate's step 9 is `tsc --noEmit`
  (measured PASS on `cd0b2bf2b`), and `pr-lint-check.yml` runs the same step as a required check, so the
  pin is enforced - just by the typecheck job rather than the test job. Worth knowing before anyone reads a
  green suite as contract coverage.
- **Two residuals are stated rather than dressed up.** `R1-A6`'s `toggleGroup` `definitionId` conjunct
  cannot be mutated to red (the fixing round measured and disclosed that itself), and no test renders
  `QueuedWorkflowTimeline` with `runEnded={true}`. Both are carried as P3, not claimed closed.

Thanks for these - three of the twelve were things a source-only review would not have reached, and the
`R5-01` enrolment finding in particular was the one that made us re-derive the backend's own
`LEGAL_HALT_ENROLLMENT_FLAGS` set rather than trusting either side's prose.

PR #963, opened 2026-07-24
docs: explain Codex bwrap AppArmor requirement

Freudator86:
## Summary

- document the Ubuntu AppArmor user-namespace restriction that can make Codex `apply_patch` fail under `workspace-write`
- show a path-scoped profile for Codex's bundled `bwrap` and the in-worktree shell-write stopgap

## Validation

- `git diff --check`
- docs-only change

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -262,6 +262,21 @@ When X mode is opted in, bootstrap also requires `curl` and `jq` before arming t
 `tasks-axi` and `quota-axi` are required bootstrap tools in every profile, the same class as `lavish-axi`.
 An absent or incompatible `tasks-axi` reports `MISSING: tasks-axi (install: npm install -g tasks-axi)`; when `config/backlog-backend` is not `manual` and compatible `tasks-axi` is on `PATH`, bootstrap stays silent and firstmate uses its verbs for routine backlog mutations, otherwise it hand-edits `data/backlog.md` until installation is approved and completed.
 An absent `quota-axi` reports `MISSING: quota-axi (install: npm install -g quota-axi)`; firstmate cannot resolve a profile array until current quota output is available for every candidate.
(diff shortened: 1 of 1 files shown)


Freudator86:
Closing this ourselves. The problem it documents is real, but the content is harmful as written and
the PR cannot honestly be made mergeable, so it will be re-raised through the pipeline with rewritten
content rather than patched here.

**Why the content should not land as written**

1. **The exact attachment path stops matching after any upgrade, silently.** The profile names one
   version-stamped path. When Codex updates, the profile no longer attaches, and nothing reports
   that — the failure returns looking like a new problem.
2. **It covers one install family; three exist.** On the host this was written from, `bwrap` ships
   under a standalone release directory, an npm prefix, and an nvm node version. The single-path
   advice is wrong for two of the three.
3. **No behavioural verification step.** `apparmor_parser -r` exiting 0 proves the profile parsed,
   not that it attached. The absence of a "re-run the previously failing operation and confirm it
   now succeeds" step is exactly what allowed an earlier incident to read as fixed for four days.
4. **The shell-write stopgap weakens normal edit discipline** and should not be recommended without
   re-testing whether it is still needed at all.

**Why it cannot be fixed in place**

The required `PR must be raised via no-mistakes` check has failed since the `opened` event and again
on `synchronize`. The enforcement is a literal search for a pipeline marker in the PR body; a branch
re-push does not rewrite the body, and pasting the marker by hand would forge provenance for a
pipeline run that never happened. The same diagnosis closed an earlier PR by the same author.

**What the replacement will contain:** globbed stanzas for all three install families with every
varying segment globbed including the home directory, a discovery command for the real
`codex-resources/bwrap` paths, the sysctl left enabled, and a mandatory behavioural check in a fresh
workspace-write sandbox.

Thanks for the queue triage reminders — the fault here is ours, not the review backlog's.

PR #1002, opened 2026-07-24
feat: prefer native browser tools with isolated AXI fallback

RaFoyer:
## Summary

- Prefer browser capabilities native to the active harness while honoring an explicit user-selected browser.
- Keep chrome-devtools-axi as a capability-gated fallback with per-task isolated sessions.
- Remove legacy ambient browser hooks only through an explicit, guarded migration.
- Sanitize inherited browser connection and profile state, preserve failed sessions for recovery, and clean successful sessions during teardown.

## Current-state revalidation

Validated exact head `6b46b33` against current `main` at `61dc99d`.

- Clean synthetic merge into current `main`
- Bash syntax checks passed
- Focused AXI isolation, bootstrap, brief, secondmate safety, and teardown suites passed
- Diff checks passed
- Shellcheck found no browser-lane defect; only existing repository-wide informational include diagnostics and the existing unused-variable warning remain

## Safety boundary

Native browser tools remain preferred when available.
The AXI fallback never attaches to an existing browser or persistent profile unless the user explicitly authorizes that attachment.
Failed isolated sessions retain their recovery record; successful stop and teardown clear it.


RaFoyer:
Apologies — this was opened against the upstream by mistake (CLI defaulted to the parent repo from our fork). The change is being developed on the fork; nothing intended for upstream at this time.

PR #1065, opened 2026-07-26
fix(decision-hold): find durably resolved decisions in the archive

knowttl:
## The defect

`bin/fm-decision-hold.sh` resolved every task lookup through `task_show()`, which is `tasks-axi show <id> --full`.
That reads only the active backlog.

`.tasks.toml` sets `archive = "data/done-archive.md"` and `done_keep = 10`, so once ten newer Done items exist, tasks-axi prunes older Done items - including resolved captain-decision holds - out of `data/backlog.md` and into the archive.
`verify_hold_durable()` then failed with `captain decision <id> is absent from .../data/backlog.md`.
Scout teardown calls that gate, so the investigation could never be torn down again.

Observed on a live home: three resolved holds sat intact in `data/done-archive.md`, complete with their `Resolution recorded by fm-decision-hold` record, decision digest, and routed identities, while `tasks-axi show` returned `NOT_FOUND` for all three and the gate refused permanently.

The property that makes this worth fixing rather than working around is that **it is triggered by getting work done**.
The faster things land, the sooner resolved decisions age out, and the more scouts get permanently stuck.
A quiet week never exposes it.

## The fix

When a lookup misses in the active backlog, consult the configured archive before declaring the record absent.

- The archive path is read from `.tasks.toml` (`[markdown] archive`), never hardcoded.
  Absolute paths, relative paths, both quote styles, and odd spacing are handled; when the key is absent there is no archive and behavior is unchanged.
- The archive is **not** a parseable backlog document - it stores entries under `## Archived <date>` sections, which `tasks-axi` rejects outright.
  Rather than duplicating tasks-axi's entry parser, the archive is read by restoring the backlog headings in a private throwaway copy and letting tasks-axi parse that.
  Only column-0 headings are dropped, and entry bodies are always indented, so no archived content can be lost.

That last point is what makes the field parsing safe.
Because the record is parsed by tasks-axi itself, `state`, `held`, `kind`, `hold_kind`, and `body` come back byte-identical to `show --full`, including the escaped single-line body form that `verify_resolution_identity` pattern-matches against. No parallel parser to drift.

## Where the fallback applies, and why

The fallback lives in the single lookup owner, `task_show`. Per call site:

| Site | Effect | Why correct |
| --- | --- | --- |
| `verify_hold_durable` | **Fixed.** An archived resolved decision now verifies. | The must-fix; teardown depends on it. |
| `verify_hold_active` | **Still refuses** an archived record. | It asserts a hold is *currently* active; an archived record is Done, so it fails on `state != queued`. The reason is now accurate (`not queued (state=done)`) instead of a misleading "absent". |
| `command_hold` existence probe | **Closes a related hole.** | A resolved-then-archived identity used to miss, so `hold` silently created a fresh queued hold reusing it, bypassing the "already durably resolved; use a new decision key" refusal. It now refuses. |
| `verify_hold_resolved` | Idempotent `resolve` retry works after archival. | Previously fell through to an "absent" failure. |
| `origin_exists_here` | An origin whose backlog item was archived is still owned by this home. | Same defect class. |
| `resolve` routed-task lookups | Refuses with "already done" rather than "does not exist". | Both refuse; the new reason is the true one. |
| `add` / `update` / `unblock` / `done` mutations | Unchanged - active backlog only. | An attempt to mutate an archived record fails loudly rather than silently. |

## The gate is not weakened

- A genuinely absent decision still fails; a miss never becomes a pass.
- A record found in the archive must satisfy the **same** durable-resolution conditions - finding it is not enough, it must still carry `Resolution recorded by fm-decision-hold.` and `Routed work:`.
- An archived entry can satisfy only the durably-resolved branch, never the actively-held one. The archive holds pruned Done items, so an entry there is never actively held.
- If the archive cannot be parsed, the lookup misses and the gate refuses. There is no optimistic pass path.

## Regression coverage, both directions

In `tests/fm-decision-hold-lifecycle.test.sh`, using `tasks-axi prune --keep 0` to trigger exactly the archival that retention eventually performs:

1. An archived resolved decision verifies and lets teardown proceed - the fix.
2. A genuinely missing decision still refuses, by name.
3. An archived decision closed without a resolution record still refuses, with the right reason.

Tests 2 and 3 are the guard on the fix. Verified that test 1 fails without the code change, with the exact reported error:

```
fm-decision-hold: captain decision sample-archive-review-decision-route is absent from <home>/data/backlog.md
not ok - verification refused a resolved captain decision that retention had archived
```

## Recommendation, not implemented

The alternative shape - trusting the completion attestation already in `state/<id>.meta` instead of re-deriving from the backlog - was considered and deliberately **not** implemented.
It would weaken the gate: `decisions_reviewed=1` records that an inventory was *taken*, not that each decision is *still* durably held or resolved.
Re-checking the durable record at teardown time is precisely the gate's value.

## Scope

This defect only. No changes to retention, `.tasks.toml`, or `tasks-axi`. Raising `done_keep` would only reset the clock.

## Verification

GitHub Actions is disabled on this account, so this PR will show an empty check set. Local verification substitutes. What I actually ran:

```
bash tests/fm-decision-hold-lifecycle.test.sh     10/10 ok (run under both LC_ALL=C and en_US.UTF-8)
bash tests/fm-teardown.test.sh                    32/32 ok
bash tests/fm-bearings-snapshot.test.sh           42/42 ok
bin/fm-lint.sh                                    clean, ShellCheck 0.11.0 (pinned)
bin/fm-test-run.sh --lane portable-parallel-1|-2|portable-serial --exclude-family real-herdr-gated
                                                  87 scripts
```

ShellCheck was not installed locally; I fetched the pinned 0.11.0 build so `bin/fm-lint.sh` ran at CI parity rather than being skipped.

Two lane failures are **pre-existing and unrelated**, each reproduced identically on the baseline commit with this change reverted:

- `tests/fm-pi-watch-extension.test.sh` - the OpenCode watch plugin does not arm after the session lock PID matches (fails the "watch arm did not run after the session lock matched" branch).
- `tests/fm-session-start.test.sh` - "MISSING diagnostic did not appear at all".

Separately, `tests/fm-test-run.test.sh` fails only under a UTF-8 locale: the coverage guard builds its inputs with `LC_ALL=C sort` but invokes `comm` without it (`bin/fm-test-run.sh:434-435`), so the collations disagree. It passes under `LC_ALL=C`, which is why CI is green. Worth a separate fix; deliberately untouched here.

Four further lane failures (`fm-backend-herdr`, `fm-composer-lib`, `fm-backend-cmux`, `fm-afk-inject-e2e`) each pass when run individually. They were contention artifacts from concurrent suite runs on a shared machine that also hosts other active firstmate homes, not defects in this change.


kunchenguid:

<sub>Noted for firstmate#1065 at `9dbe690e`.</sub>

PR #1081, opened 2026-07-26
feat(homeguard): no-op Pi primary extensions outside FM home (ONM-1065)

ohzie:
## Summary

Incident 2026-07-26: a Pi session in critter_clash (non-firstmate project) loaded the user-scoped firstmate watcher extension, acquired `state/.lock`, and role-confused into supervisor mode (armed watcher, narrated supervision instead of building).

## Root cause

`fm-primary-pi-watch.ts` and `fm-primary-turnend-guard.ts` derive their `root` (firstmate home) from `import.meta.url` — two levels up from `.pi/extensions/`. When loaded user-scoped from firstmate's path while the session's cwd is a foreign project, `root` = firstmate but `cwd` ≠ firstmate, so the extension operates on firstmate's `state/` in the wrong session.

## Fix

Add `isAtFirstmateHome()` guard at the entry of both extensions' exported default functions:

```ts
function isAtFirstmateHome(): boolean {
  try {
    const cwd = realpathSync(process.cwd());
    const home = realpathSync(root);
    return cwd === home || cwd.startsWith(home + "/");
  } catch {
    return false;
  }
}
```

- Uses `root` (extension-file-location-based) not `FM_HOME` (env var) — prevents spoofing
- `realpathSync` on both sides resolves symlinks consistently
- `startsWith` covers subdirectory sessions

When `isAtFirstmateHome()` returns false: no lock acquisition, no watcher arming, no turn-end signals, no marker writes.

## What is NOT affected

- Primary session in firstmate: `cwd === root` ✓
- Pi secondmate in firstmate worktree: `PROJ_ABS = cwd = FM_HOME = root` ✓
- Crewmate turn-end signal extension (`state/$ID.pi-ext.ts`): different file, unaffected

## Regression test

`tests/fm-ext-homeguard.test.sh` (7 cases):
1. Static contract: tracked extensions contain the home guard
2. Watch extension no-ops in foreign cwd (no tools/events/markers/arm)
3. Watch extension works in home cwd
4. Watch extension no-op does not acquire or modify the session lock
5. Watch extension works when cwd is a subdirectory of FM home
6. Turn-end extension no-ops in foreign cwd
7. Turn-end extension registers events in home cwd

## Test results

```
tests/fm-ext-homeguard.test.sh        7/7 pass
tests/fm-pi-watch-extension.test.sh  32/32 pass
tests/fm-turnend-guard.test.sh       40/40 pass
```

Closes ONM-1065.

### DIFF
--- a/.pi/extensions/fm-primary-pi-watch.ts
+++ b/.pi/extensions/fm-primary-pi-watch.ts
@@ -1,7 +1,7 @@
 // Firstmate primary watcher bridge for Pi.
 import { spawn, spawnSync, type ChildProcess } from "node:child_process";
 import { createHash } from "node:crypto";
-import { mkdirSync, readFileSync, writeFileSync } from "node:fs";
+import { mkdirSync, readFileSync, realpathSync, writeFileSync } from "node:fs";
 import { dirname, resolve } from "node:path";
 import { fileURLToPath } from "node:url";
 import type { ExtensionAPI, Theme } from "@earendil-works/pi-coding-agent";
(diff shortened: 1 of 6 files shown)


ohzie:
Withdrawing: captain decided this lands in our local firstmate install rather than the public repo.

PR #1114, opened 2026-07-27
fix(bin): keep heredoc bodies out of command substitutions so fm-brief.sh parses under stock macOS Bash 3.2

guido-dev-talentix:
## Problem

`bin/fm-brief.sh` does not parse at all under the bash that stock macOS ships, so scaffolding any brief is impossible on a stock Mac:

```
$ /bin/bash --version | head -1
GNU bash, version 3.2.57(1)-release (arm64-apple-darwin25)
```

Bash 3.2 scans a `$( ... )` command substitution for its closing paren with a quote tracker that does not skip heredoc bodies. An **odd** number of quote characters inside a heredoc written *inside* a command substitution makes that scan swallow the rest of the file. The word `firstmate's` in the no-mistakes definition-of-done block (introduced in #945) was that odd apostrophe, and it killed the parse of the whole file - so every delivery mode was broken, not just no-mistakes. Bash 4+ parses the construct, which is why Linux CI never caught it.

This is the second occurrence: #166 fixed the same class, but its guard was a grep for one specific phrase, which a different word walked straight past.

## Fix (structural, not cosmetic)

Heredoc bodies now live in emitter functions (`cat <<DELIM` inside a function, captured by the caller as `VAR=$(fn)`). A heredoc inside a function body is parsed at top level, where quote characters are ordinary text, so no future wording can re-arm the trap. Generated text is byte-unchanged - the diff touches only wrapper lines.

Nine instances existed across the toolbelt and all are converted:

| File | Instances |
| --- | --- |
| `bin/fm-brief.sh` | three delivery-mode definition-of-done blocks + the Herdr declaration (the fatal ones) |
| `bin/fm-bootstrap.sh` | the generated `config/x-mode.env` cadence body |
| `bin/fm-fleet-snapshot.sh` | two embedded bounded-reader scripts + two jq filters |

The last five parsed today only because their quote characters happened to be balanced - the same latent trap, one edit from breaking.

`tests/*.sh` are deliberately untouched: six of them contain the construct (node/jq fixture bodies) but all parse cleanly under 3.2 today, and neither script the macOS CI lane executes is among them.

## Guard

`tests/fm-stock-bash-parse.test.sh` owns the class, not the phrase:

1. bans heredoc-inside-command-substitution across `bin/*.sh` and `bin/backends/*.sh` - portable, so it also fails on Linux where the parser itself cannot see the bug;
2. carries a positive-control fixture pair proving the banned shape really does break bash 3.2 while the function shape parses and emits identical text, so the guard cannot pass by being blind;
3. parses every bin script under a discovered bash 3.x (`FM_STOCK_BASH` overrides the search).

Against the pre-fix tree it fails and lists all nine sites; after the fix it passes.

The pre-existing `macos-stock-bash` CI job now runs this test in place of its single-file `bash -n bin/fm-fleet-snapshot.sh` check, which the toolbelt-wide sweep strictly supersedes. No new job, runner, or lint surface was added. The test is registered in `bin/fm-test-run.sh`'s `pure-contract-unit` family map and lands in the `portable-serial` lane by complement, so `--check-coverage` stays consistent.

## Verification

On the real `/bin/bash` 3.2.57 (the only bash on the affected machine):

- defect reproduced at base, clean parse across all 92 scripts in `bin/` and `bin/backends/` after;
- all six scaffold shapes (three delivery modes, scout, `--herdr-lab`, secondmate charter) generate complete briefs;
- a freshly inserted lone apostrophe in brief prose still parses, while the same sentence in the old `DOD=$(cat <<EOF` form breaks it again - the trap is demonstrably disarmed;
- `bin/fm-lint.sh` clean at the pinned ShellCheck 0.11.0; `bin/fm-doc-audience-check.sh` clean; `bin/fm-test-run.sh --check-coverage` ok;
- `tests/fm-brief.test.sh` 15/15, plus `fm-stock-bash-parse`, `fm-fleet-snapshot-view`, `fm-bearings-snapshot`, `fm-x-mode`, `fm-bootstrap`, `fm-lint`, `fm-documentation-audiences`, `fm-instruction-owners` green.

Two suites fail identically at base with these changes stashed and are therefore environmental, not caused by this change: `tests/fm-test-run.test.sh` (timing-based scheduler asserts) and `tests/fm-session-start.test.sh` (concurrent session-lock).

Validated through the no-mistakes pipeline (intent, rebase, review, test, document, lint all green). It is opened from a fork because the contributing account has pull-only rights on this repo.


guido-dev-talentix:
Closing and re-raising: this PR was opened by hand, so its body lacks the deterministic `## Pipeline` section that the required "Require no-mistakes" check looks for. The change was validated through the no-mistakes pipeline (review, test, document and lint all green), but the PR itself must be raised by the pipeline's own PR step to carry that signature - I will not hand-write the marker. Re-running the gate in its supported fork mode (`no-mistakes init --fork-url`) so it pushes to the fork and opens the PR against this repo itself.

PR #1137, opened 2026-07-27
fix(brief): generalize member-surface write-path doctrine

ICGNU3:
Bridges the no-mistakes heredoc parse failure that opened the brief-doctrine-generate-and-repair session, and generalizes fm-brief.sh so the same shape works for member-surface write-path tasks instead of just brief edits.

Sourced from session fm-brief-heredoc-fix-v0 + fm-dispatch-config-review-v0. 
Two commits: one for the generalized doctrine (bin/fm-brief.sh supports a member-surface write-path template), one for the heredoc wording fix (apostrophe escape).

### DIFF
--- a/bin/fm-brief.sh
+++ b/bin/fm-brief.sh
@@ -325,7 +325,7 @@ Two firstmate-specific rules layer on top of that guidance:
 - ask-user findings are never yours to answer: escalate to firstmate (rule 6) and stop.
   Firstmate applies the authority contract in its \`AGENTS.md\` and obtains any required captain decision.
   When the decision comes back, feed it to the gate with \`no-mistakes axi respond\` and let the pipeline apply it - do not route the question to "the user" or implement the fix yourself.
-- Avoid \`--yes\`: it would silently bypass firstmate's authority check and any required captain escalation.
+- Avoid \`--yes\`: it would silently bypass the firstmate authority check and any required captain escalation.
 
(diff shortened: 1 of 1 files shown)


ICGNU3:
Closing this in favor of #1163, which fixes the macOS parse bug without dropping the pinned contract text.

**This change breaks the ask-user authority contract test.** Verified on this branch:

```
fm-ask-user-authority.test.sh  8 ok, 1 fail
  not ok - generated implementation brief permits silent ask-user auto-resolution
fm-brief.test.sh              15 ok, 0 fail
fm-tangle-guard.test.sh         6 ok, 0 fail
```

`tests/fm-ask-user-authority.test.sh` asserts the generated brief contains the exact sentence:

```
silently bypass firstmate's authority check and any required captain escalation
```

This PR rewrites that section and removes the sentence, but touches only `bin/fm-brief.sh` and no test file, so the assertion fails.

That assertion was added in #945 alongside the text, which reads as deliberate: the brief wording is a contract, not incidental prose.

**Side effect worth naming.** Removing that line also removes the apostrophe in `firstmate's`, which happens to fix a separate bug: bash 3.2 mis-scans a bare apostrophe inside a heredoc nested in `$( )`, so `bin/fm-brief.sh` has not parsed under stock macOS bash since #945. CI is Linux bash 5, where the file is valid, so it is invisible from CI while brief scaffolding is completely dead on macOS.

#1163 fixes that directly by interpolating the apostrophe through `$APOS`, leaving the generated text byte-identical and `fm-ask-user-authority` at 9/9.

If the doctrine generalization here is still wanted, it should land on top of #1163 with the contract assertion updated in the same commit.

PR #1138, opened 2026-07-27
fix(teardown): uniformly apply worktree safety to scout and secondmate kinds

ICGNU3:
fm-teardown.sh refused the landed-work test for scout and secondmate kinds. This branch makes the safety contract uniform across all kinds: scout worktrees return scratch state, secondmate homes gate teardown by their own backlog/state, ship kinds keep the standard landed-work check.

Verified: bin/fm-teardown.sh -h unchanged; 40 ok / 0 not-ok on the test matrix, including 8 new unit tests covering scout/secondmate paths.

### DIFF
--- a/bin/fm-teardown.sh
+++ b/bin/fm-teardown.sh
@@ -668,9 +668,6 @@ validate_worktree_teardown_safety() {
   local dirty_raw dirty unpushed_raw unpushed DEFAULT unmerged_raw unmerged branch
   [ -d "$WT" ] || return 0
   [ "$FORCE" != "--force" ] || return 0
-  case "$KIND" in
-    secondmate|scout) return 0 ;;
-  esac
 
   if ! dirty_raw=$(git -C "$WT" status --porcelain 2>/dev/null); then
     if worktree_safety_blocked_by_lock "uncommitted changes"; then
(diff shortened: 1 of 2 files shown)


ICGNU3:
Closing this in favor of #1160, which fixes the same data-loss bug but keeps the `secondmate` half of the exemption.

**This change breaks every secondmate teardown.** Verified end-to-end on this branch:

```
fm-teardown.test.sh          40 ok, 0 fail   <- this PR's own suite, green
fm-secondmate-safety.test.sh 40 ok, 1 fail
  not ok - teardown failed for empty secondmate home
```

**Why the new tests pass anyway.** They call the safety function directly against a `make_case` fixture, which is a real git worktree. A secondmate home is not: it is a plain directory carrying a `.fm-secondmate-home` marker. Every check in `validate_worktree_teardown_safety` shells out to `git -C "$WT"`, which fails on a non-repo and correctly fail-safes to `REFUSED`. The unit fixture never reaches that condition, so the uniform-behavior assertions hold while the real path is broken.

Only `tests/fm-secondmate-safety.test.sh` exercises it, through `fm-teardown.sh` end to end.

Worth noting: that suite cannot run at all on macOS right now, because `bin/fm-brief.sh` has not parsed under bash 3.2 since #945. That likely masked this. #1163 fixes the parse; with it applied, the failure above reproduces reliably.

**The asymmetry is real, not an oversight.** `scout` should lose the exemption: a scout's worktree *is* a git worktree, so every check applies, and a scout lane authored an ADR and a guard script in our fleet, so "scouts are read-only" is false in practice. `secondmate` should keep it: there is no uncommitted git state there for these checks to protect, and removing it makes teardown impossible rather than safe.

#1160 also drops the same exemption from `post_lock_cleanup_check`, which this PR leaves in place. That recheck is what catches work written between the first safety check and the `treehouse return` that resets the worktree.

PR #1147, opened 2026-07-27
fix: avoid apostrophe in fm-brief.sh heredoc nested in command substitution

Seangee86:
## Summary
- `bin/fm-brief.sh` crashes with `unexpected EOF while looking for matching \`'\''` on macOS's default bash (3.2.57) when scaffolding a ship brief for a `no-mistakes`-mode project.
- Root cause: macOS ships bash 3.2, which mis-parses a literal apostrophe inside a heredoc when that heredoc is nested inside a `$(...)` command substitution (reproduced independently outside this repo; a standalone top-level heredoc with the same apostrophe is fine, only the nested-in-`$()` case breaks).
- The `no-mistakes`-mode DOD heredoc (`DOD=$(cat <<EOF ... EOF)`) in the ship-brief case block contains the word "firstmate's", which trips this.
- Fix: reworded that one line to drop the apostrophe ("firstmate's authority check" -> "the firstmate authority check"). No behavior change, no other heredocs in the file are affected (verified none of the other apostrophes in the file sit inside a heredoc nested in `$(...)`).

## Test plan
- [x] `bash -n bin/fm-brief.sh` passes (previously failed with the EOF/quote error)
- [x] `bin/fm-brief.sh <id> <repo>` (no-mistakes mode) scaffolds successfully end to end

### DIFF
--- a/bin/fm-brief.sh
+++ b/bin/fm-brief.sh
@@ -325,7 +325,7 @@ Two firstmate-specific rules layer on top of that guidance:
 - ask-user findings are never yours to answer: escalate to firstmate (rule 6) and stop.
   Firstmate applies the authority contract in its \`AGENTS.md\` and obtains any required captain decision.
   When the decision comes back, feed it to the gate with \`no-mistakes axi respond\` and let the pipeline apply it - do not route the question to "the user" or implement the fix yourself.
-- Avoid \`--yes\`: it would silently bypass firstmate's authority check and any required captain escalation.
+- Avoid \`--yes\`: it would silently bypass the firstmate authority check and any required captain escalation.
 
 After /no-mistakes reports CI green (the CI-ready return point - do not wait for it to keep monitoring in the background until merge), append \`done: PR {url} checks green\` and stop. You are finished.
 EOF


kunchenguid:
thanks for tracking this down. i merged https://github.com/kunchenguid/firstmate/pull/1093, which fixes the same macOS bash 3.2 parse failure - it removes the command substitution around those heredocs instead of rewording the apostrophe, so i'm closing this one as superseded. appreciate the help here.

timidri (changes requested):
❌ Needs changes

Routing ERROR/FATAL into AppSignal is the right call and the mechanism is sound — `BroadcastLogger` keeps `scalingo logs` byte-identical, the nil-logdev subclass makes non-overridden severities free, and excluding WARN with an explicit report at the one site that warranted it (`app/services/keycloak/setup_service.rb:353`) is exactly the right shape. Three things block it: one failure currently produces three incidents, the scrubber misses the message shape our Keycloak error paths actually produce, and a logging call can now raise into the caller.

**1. One failure becomes 3+ incidents.** Fifteen call sites log the backtrace as a *second* `Rails.logger.error`, including the shared API concern `app/controllers/concerns/api/v1/json_errors.rb:38` — so it is baked into our common error handling, not just ad-hoc sites. Concretely, a Frisbii API timeout during checkout now reports: `app/services/frisbii/client.rb:105` (one), `app/services/frisbii/client.rb:106` (two — whose incident *message* is the entire backtrace), then the re-raise at `:107` is rescued at `app/controllers/api/v1/checkout_controller.rb:89` and logged again (three). The controller renders a 422, so none of this was ever an exception incident before. Cheapest fix that needs no call-site churn: have `report` drop a message that is purely a backtrace (every line matching `\A\s*\S+:\d+:in `) — that removes the worst half, including the bare-backtrace-as-message incidents.

**2. The scrubber misses Ruby's Hash rendering — the shape three Keycloak error lines produce.** `SENSITIVE_VALUE_PATTERN`'s separator `\s*[:=]>?\s*` cannot cross the closing quote after the key, so the hash form falls straight through. Running the pattern from `app/services/appsignal/log_message_scrubber.rb:54`:

```
{"password"=>"hunter2"}                                 -> unchanged
{"password" => "hunter2"}                               -> unchanged   (Ruby 3.4 rendering)
{"credentials"=>[{"type"=>"password","value"=>"x"}]}    -> unchanged   (Keycloak's own shape)
Bearer eyJhbGciOiJSUzI1NiJ9.abc.def                     -> unchanged   (no separator at all)
{password: "hunter2"}                                   -> redacted    (the shape the spec tests)
```

`app/services/keycloak/user_service.rb:35`, `:297` and `:363` each interpolate a Faraday-parsed Hash directly into the error line, and the latter two are the set-password and reset-password paths. Suggest allowing an optional closing quote before the separator (`["']?\s*(?:=>|[:=])\s*`), adding a bearer/JWT pattern, and adding these hash-shaped strings to `spec/services/appsignal/log_message_scrubber_spec.rb` — the current examples only cover shapes that already pass.

**3. A logging call can now raise into the caller.** `ActiveSupport::BroadcastLogger#dispatch` has no rescue around `@broadcasts.map { |logger| logger.send(...) }`, so anything raised inside `LogSeverityReporter#report` (`app/services/appsignal/log_severity_reporter.rb:125-132`) propagates out of `Rails.logger.error` — and most of our error calls sit *inside* a `rescue` block, so a hiccup in the reporter turns a handled failure into a 500. Stdlib `Logger#error` never raises; this subclass silently drops that guarantee. Wrapping the body of `report` in `rescue StandardError` restores it.

Non-blocking, worth a line each:

- `error(message = nil)` (`log_severity_reporter.rb:113`) ignores the block form and anything logged via `add`. There are no block-form call sites in `app/` today, and Rails 8.1's `DebugExceptions#log_error` uses `logger.add` — which is lucky, since it means unhandled exceptions are *not* double-reported — but a gem or a future call site using either form is dropped without trace. Worth stating in the class comment.
- The email pattern stops at the first TLD label: `user@sub.example.co.uk` scrubs to `[FILTERED EMAIL].co.uk`. Not a leak, but the spec doesn't cover a multi-label TLD.
- Scope: the `CLAUDE.md` symlink→import-file swap is unrelated to the observability change. Tiny, but it's a second concern in the diff.
- All 139 error sites collapse into one `Appsignal::LoggedProblem` class, distinguishable only by message. Worth confirming how AppSignal will group and how an alert rule can target one problem rather than the class — otherwise the new signal is hard to route.

**Security (CIA).** 🔴 **Red — blocking.**
*Confidentiality* — 🔴 finding 2: the scrubber leaves Ruby-hash-rendered secrets and bearer tokens intact, and `keycloak/user_service.rb:297,363` interpolate exactly that shape on the password endpoints. Sending those messages to a third-party APM is a new egress path, so the gap matters more here than it did in `scalingo logs`. *Integrity* — 🟢 N/A: no user input, no queries, no migrations, no state-changing requests. *Availability* — 🔴 finding 3 (a logging call that can raise inside a rescue block) and finding 1 (incident volume on a path that fires on every Frisbii wobble).

One timing note: this wraps the global logger in production and is +257/−4. Worth landing it after the launch window rather than during it.

_Reviewed by office.eu pr-review_

PR #1152, opened 2026-07-27
fix(bin): re-own the Kimi turn-end hook region after a config rewrite

### LINKED ISSUE #1064: kimi turn-end hook markers are TOML comments and do not survive Kimi CLI config rewrites
Observed on the adapter's first day (4d6992a), Kimi CLI 0.29.x: the first kimi spawn installed the turn-end hook region in ~/.kimi-code/config.toml correctly. Between spawns, the Kimi CLI rewrote its own config.toml (TOML normalization - comments dropped). The [[hooks]] block survived byte-identical, but the '# BEGIN/END FIRSTMATE KIMI TURN-END HOOK' comment markers were stripped. The next spawn then refused: 'config.toml references fm-turn-end.sh outside the Firstmate-owned region.' - correct fail-closed behavior, but the recurrence makes every second spawn fail whenever Kimi rewrites its config.

Suggested fix directions: (a) treat an unmarked hook block whose bytes exactly match block() as owned and re-mark it in place; or (b) move ownership signaling out of comments into something Kimi's rewriter preserves. Workaround used in the field: remove the byte-identical orphaned block by hand and respawn (installer re-installs the marked region).

Evidence: bin/fm-kimi-turnend-hook.sh:230 refusal; config.toml post-rewrite had the hook block at top level with zero 'FIRSTMATE' matches.

august-agent:
Fixes #1064.

## Problem

The Kimi CLI rewrites its own `~/.kimi-code/config.toml` and drops every comment as part of TOML normalization. That erases the `# BEGIN FIRSTMATE KIMI TURN-END HOOK` / `# END ...` markers delimiting Firstmate's turn-end hook region, while leaving the `[[hooks]]` table itself byte-identical. The next crewmate spawn then refused:

```
fm-kimi-turnend-hook: refused: config.toml references fm-turn-end.sh outside the Firstmate-owned region.
```

So the second Kimi spawn onward could not launch, which made the Kimi crew adapter unusable for repeated work.

### Reproduction

Install into a fresh config, delete every comment line (the CLI's own normalization), then install again:

```
$ HOME=$repro bin/fm-kimi-turnend-hook.sh install          # ok, region written with markers
$ sed -i '' '/^[[:space:]]*#/d' $repro/.kimi-code/config.toml   # simulate the Kimi rewrite
$ HOME=$repro bin/fm-kimi-turnend-hook.sh install
fm-kimi-turnend-hook: refused: config.toml references fm-turn-end.sh outside the Firstmate-owned region.
rc=1
```

The `[[hooks]]` table survives the rewrite intact; only the two comment lines are gone.

## Fix

Region identification no longer depends on the comment markers surviving. Markers remain the fast path; when they are absent, the region is re-identified structurally:

1. Exactly one `[[hooks]]` entry may reference `fm-turn-end.sh`, and its parsed content must equal the hook entry the installer itself emits. That expected entry is derived by parsing the generated block rather than restated, so the block stays the single owner of the hook table's content.
2. The entry's text span is located by scanning `[[hooks]]` headers in document order, ending at the last non-blank, non-comment line before the next table header, so trailing blank lines and comments belonging to the following section are left alone.
3. The span is verified before use: the config with that span excised must parse, and must equal the original parsed document minus exactly that one entry. Any scanning mistake fails this check and refuses instead of rewriting the captain's config.
4. The verified span is then re-owned and re-marked in place. Repeated installs neither refuse nor accumulate duplicate hook tables, and removal still restores the surrounding bytes exactly.

Adoption runs before both the install and remove branches, so a config whose markers were stripped can also be cleanly uninstalled.

## Safety property preserved

The refusal exists to stop Firstmate clobbering a hook it does not own, and that is unchanged. A `[[hooks]]` table that references `fm-turn-end.sh` but is not the installer's own entry stays foreign and is still refused before any config write. All four shapes are covered by tests:

- a different script path
- Firstmate's entry plus an extra captain-added key
- two identical unmarked Firstmate blocks
- the inline-array `hooks = [{ ... }]` form

## Tests

`tests/fm-kimi-harness.test.sh` gains two cases:

- `test_kimi_hook_reowns_its_region_after_a_comment_stripping_rewrite` - fresh install, simulated comment-stripping rewrite, re-install succeeds and re-marks the region, exactly two hook tables remain (one foreign, one Firstmate's), a further install is byte-idempotent, and removal after another rewrite restores the original config byte-for-byte.
- `test_kimi_hook_still_refuses_unmarked_foreign_hook_tables` - the four foreign shapes above each refuse with the concrete reason, without changing config bytes and without writing the hook script.

All 21 cases in that file pass, as do `fm-turnend-guard`, `fm-documentation-audiences`, `fm-instruction-owners`, and `fm-lint`. `bin/fm-lint.sh`, `shellcheck`, and `bin/fm-doc-audience-check.sh` are clean.

## Docs

`docs/turnend-guard.md` records the rewrite behavior and the structural re-ownership as a current Kimi fact, and its regression-coverage line now names the new coverage.

## Note for maintainers

The installer requires a `python3` with `tomllib`, so Python 3.11 or newer. On a host whose first `python3` on PATH is older, it refuses before touching the config. That behavior is pre-existing and unchanged here, but it is worth confirming on Kimi hosts.

### DIFF
--- a/bin/fm-kimi-turnend-hook.sh
+++ b/bin/fm-kimi-turnend-hook.sh
@@ -7,6 +7,16 @@
 # and remove excises only that region. Missing, malformed, symlinked, partially
 # marked, or otherwise surprising config is refused without a config write.
 #
+# The Kimi CLI rewrites its own config.toml and drops comments, which erases the
+# region's comment markers while leaving the hook table itself intact. When the
+# markers are absent, the region is re-identified structurally: exactly one
+# [[hooks]] table whose parsed content equals Firstmate's own hook entry, whose
+# text span is confirmed by re-parsing the config with that span excised. That
+# span is then re-owned and re-marked in place, so repeated installs neither
+# refuse nor accumulate duplicate hook tables. A [[hooks]] table that references
+# fm-turn-end.sh but is not byte-for-byte Firstmate's own entry stays foreign and
+# is still refused.
+#
 # The installed Stop hook always exits 0 and stays silent. It reads cwd from the
 # hook payload, checks for a .fm-kimi-turnend pointer before registry work, and
 # touches a task turn-end marker only when the pointer names a Firstmate-created
@@ -69,6 +79,8 @@ BEGIN_OWNS_NEWLINE = BEGIN + b" (OWNS PRECEDING NEWLINE)"
 END = b"# END FIRSTMATE KIMI TURN-END HOOK"
 IDENTIFIER = b"FIRSTMATE KIMI TURN-END HOOK"
 HOOK_NAME = b"fm-turn-end.sh"
+HOOKS_HEADER = re.compile(rb"^[ \t]*\[\[[ \t]*hooks[ \t]*\]\][ \t]*(#.*)?$")
+TABLE_HEADER = re.compile(rb"^[ \t]*\[")
 TOKEN_NAME = re.compile(r"fm\.[A-Za-z0-9]{12}\Z")
 
 HOOK_BYTES = b'''#!/usr/bin/env bash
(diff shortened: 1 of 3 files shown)


august-agent:
Closing in favor of a single clean PR that delivers this fix correctly through the no-mistakes pipeline. Issue #1064 remains the tracking issue.

PR #1153, opened 2026-07-27
fix(watch): keep a declared pause on its bounded cadence

august-agent:
## The problem

A crew that declares `paused:` and then sits on long external monitors made the watcher churn. Every arm cycle started a watcher that immediately closed on `stale: <window>` for that same declared wait, so no watcher stayed up to hold the home lock, and the turn-end path then reported `watcher: FAILED - cycle ended without an actionable reason` and demanded a supervision repair. This repeated every few seconds and never settled. Declaring the pause helped the first surface only; the repetition continued.

## Root cause

Two independent defects, both reproduced before changing any code.

**Watcher, `bin/fm-watch.sh`.** `pause_state_class` demoted a still-declared pause to `none` whenever the crew's agent was still live, so a live declaration was designed to surface exactly once per stale hash. That holds only for a byte-static pane. A live crew's pane text keeps drifting, and on every drift the changed-hash branch reclassified the declaration as `none` and called `clear_pause_tracking`, which deletes `.paused-<key>` and the `.paused-resurfaced-<key>` cadence anchor. Two polls later the pane was stably stale again with no suppressor, so the same declared wait surfaced as a bare `stale:` wake and the cycle closed on it. Nothing bounded the repetition: the `FM_PAUSE_RESURFACE_SECS` cadence never applied to a live paused crew at all. A briefly busy pane wiped the same markers through the second clear in the busy branch.

Reproduction, five consecutive watcher cycles over a live paused crew whose pane gained one line between cycles:

```
round 1: EXITED -> stale: test:fm-op
round 2: EXITED -> stale: test:fm-op
round 3: EXITED -> stale: test:fm-op
round 4: EXITED -> stale: test:fm-op
round 5: EXITED -> stale: test:fm-op
stale wakes queued: 5
```

The same fixture with a byte-static pane surfaced once and absorbed the remaining four cycles, which isolates pane drift as the amplifier.

**Arm layer, `bin/fm-watch-arm.sh`.** Several arms can follow one watcher cycle. Only the arm that owns the watcher prints the wake reason; a follower attaches, then sees the cycle close with no successor and reported the typed failure for a close supervision had handled correctly:

```
round 1: armA rc=1 [watcher: already running pid N|watcher: attached pid=N (beacon 0s)|watcher: FAILED - cycle ended without an actionable reason|]
round 1: armB rc=0 [watcher: started pid=N (beacon fresh)|signal: .../op.status|]
```

## The fix

A declaration stays on its bounded `FM_PAUSE_RESURFACE_SECS` cadence once it has been surfaced inside the current window, whether or not its agent is still live, and its cadence markers survive both pane drift and a briefly busy pane. The withdrawal check at the top of the stale loop remains the one owner of clearing them.

Preserved: an active run still outranks the declaration and takes the ordinary wedge timer; a crew with no declaration still trips stale immediately; a forgotten pause still re-surfaces once per window as an awaiting-external recheck; a live declaration firstmate has not seen yet still surfaces once on first sight.

A follower arm now consults the durable wake queue, the shared record of what a cycle surfaced. A record appended at or after the cycle began, together with a fresh liveness beacon, explains the close, so the follower prints those queued reason lines and exits zero. A stale or absent beacon never qualifies, so genuinely down supervision still fails loudly and the guard keeps its real protection.

After the fix, the same five-cycle reproduction surfaces once and absorbs the rest, and both concurrent arms return zero with the queued reason.

## Tests

- `tests/fm-watch-triage.test.sh` - a live declared pause surfaces once, absorbs across pane drift with its cadence markers and wedge-timer exemption intact, and returns to ordinary stale surfacing once withdrawn. Fails on the pre-fix watcher at drift cycle 2.
- `tests/fm-watcher-lock.test.sh` - a follower arm whose cycle queued a wake reports that wake instead of a typed failure, and the same setup with an ancient beacon still fails loudly. The first fails on the pre-fix arm with exit 1.
- `tests/fm-turnend-guard.test.sh` - the guard stays silent for a live beating watcher with wakes queued for the next turn, and still blocks when no beacon exists.

Also fixes a pre-existing race in the watcher-lock suite: the TERM-resistant peer fixture is now waited on until it has installed its signal handler, since the restart's signal could otherwise land during interpreter startup, kill the peer, and fail the attach assertion for reasons unrelated to the code under test. That failure reproduces on `main`.

## Validation

`bin/fm-lint.sh` clean, `bin/fm-doc-audience-check.sh` clean, and `bin/fm-test-run.sh --family watcher-wake-lock` green (10 suites, 0 failed), plus the auto-arm, instruction-owner, and documentation-audience suites.

### DIFF
--- a/bin/fm-watch-arm.sh
+++ b/bin/fm-watch-arm.sh
@@ -30,15 +30,24 @@
 #                                                          this arm attaches and follows it
 #   watcher: FAILED - no live watcher with a fresh beacon  - could not confirm one
 #   watcher: FAILED - cycle ended without an actionable reason
-#                                                        - a clean cycle ended with no wake and no
-#                                                          verified healthy successor
+#                                                        - a clean cycle ended with no wake, no
+#                                                          verified healthy successor, and nothing
+#                                                          on the durable wake queue to explain it
+#   watcher: cycle closed on a wake already queued by a concurrent arm
+#                                                        - the followed cycle closed on a wake this
+#                                                          arm did not print itself; the queued
+#                                                          reason lines follow and this arm exits 0
 # It NEVER reports started/attached/healthy off a stale beacon or a dead/reused pid: a
 # stale-beacon or dead-pid holder either self-heals (the fresh child steals the
 # dead lock per the singleton self-eviction/steal path and is confirmed) or this
 # returns the FAILED line. On started it waits the child and propagates the wake
 # reason; on attached it stays live across identity-matched successors. An
-# attached cycle that ends without a healthy successor is a typed nonzero failure,
-# never a clean empty completion. On FAILED it exits non-zero so the failure is
+# attached cycle that ends without a healthy successor is a typed nonzero failure
+# whenever nothing explains the close, never a clean empty completion; when the
+# durable wake queue shows the cycle surfaced a wake at or after this cycle began
+# and the beacon is still fresh, that close is instead reported with its queued
+# reason lines, because supervision did its job and one of several concurrent
+# arms simply did not own the print. On FAILED it exits non-zero so the failure is
 # loud. A live cycle already present means re-arm attaches - do not start a second
 # watcher.
 #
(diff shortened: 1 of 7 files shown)


kunchenguid:

PR #1157, opened 2026-07-28
fix(session-start): print the secondmate charter in the context digest

brchue-ux:
## Problem

A seeded secondmate home carries its charter at `data/charter.md`, but `bin/fm-session-start.sh` never printed it.
The CONTEXT digest listed only `data/projects.md`, `data/secondmates.md`, `data/captain.md`, `data/captain-shared.md`, and `data/learnings.md`.

`AGENTS.md` line 3 ("You are the first mate.") is loaded every session in every home, while the charter that would correct it in a secondmate home was loaded in none.
The result is a reproduced identity bug: a second mate opens the session announcing itself as the first mate and describing a fleet-wide role it does not have.

## Changes

**`bin/fm-session-start.sh`** - print `data/charter.md` in the CONTEXT digest, first, before `data/projects.md`.
It answers "who am I", which governs how everything after it is read.
It uses the existing `print_file_or_absent` helper rather than a second emission path, so absence prints the same explicit `ABSENT` marker and stays distinguishable from an empty-but-present file.
Absence is meaningful: no charter means this is a primary home rather than a secondmate one.
`data/charter.md` is also added to the digest's closing "do NOT re-read" trailer and to the two header comments that enumerate the context files.

**`AGENTS.md`** - one line in the top address block stating the identity rule, plus `charter.md` in the section 2 `data/` layout listing, one line in the section 3 ABSENT-meaning sentence, and `data/charter.md` in the section 3 context-digest enumeration.
The last two are cross-reference updates to existing lists, not restatements of the rule.
The additions are deliberately tight because that file's token cost is paid by every session of every fleet member.

## Deliberate non-changes

The identity rule is **not** added to the charter template in `bin/fm-brief.sh`.
This is a one-owner-rule decision: `AGENTS.md` owns the rule and the charter supplies only the domain.
Duplicating it into the generated charter would create two copies to drift, and would leave already-seeded homes - whose charters were copied at seed time and are gitignored local state - still broken.
The two changes here fix every existing home with no per-home edits, which is a required property of the fix.

No home's `data/charter.md` was edited.

`AGENTS.md` section 1's "only point of contact for all software work across all of their projects" sentence was reviewed and deliberately left alone; that remit wording is tracked as its own separate item.

## Verification

Extended `tests/fm-session-start.test.sh` rather than adding a new runner, per the repo's colocated-test rule:

- `test_context_digest_charter_present` - content printed, ordered before `projects.md`, listed in the trailer.
- `test_context_digest_charter_absent` - `ABSENT` marker, not `(present, empty)`.
- The pre-existing digest test's expected `ABSENT` count moves from 4 to 5, since `charter.md` is now a sixth context file.

`bin/fm-lint.sh` and `bin/fm-doc-audience-check.sh` both run clean (`ok surfaces=55 local_links=151`).

One pre-existing unrelated failure in that suite, `not ok - MISSING diagnostic did not appear at all`, reproduces identically on the base commit b29621b and is not caused by this change.

### Digest output

Charter present:

```
data/charter.md (secondmate charter; ABSENT = this is a primary home)
--------------------------------------------------------------------------------
You are a persistent second mate managed by the main firstmate.
Your domain is the budget app (home_budget_app).

data/projects.md
--------------------------------------------------------------------------------
- home_budget_app [no-mistakes] - the budget app (added 2026-07-01)
```

Charter absent:

```
data/charter.md (secondmate charter; ABSENT = this is a primary home)
--------------------------------------------------------------------------------
ABSENT

data/projects.md
--------------------------------------------------------------------------------
- demo [no-mistakes] - a demo project (added 2026-07-01)
```

## Compatibility

Checked rather than assumed.
The CONTEXT section of `fm-session-start.sh` is a sequence of unconditional `print_file_or_absent` calls with no reference to `PRIMARY_HARNESS` or any backend variable.
Harness branching in that script is confined to the supervision-instructions step, and backend reads to the fleet-state endpoint liveness loop; neither gates CONTEXT.
This change is therefore independent of both the harness axis and the runtime-backend axis.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,6 +1,7 @@
 # Firstmate
 
 You are the first mate.
+If this home has a `data/charter.md`, you are instead a second mate: identify yourself as the second mate for the domain that charter names, never as the first mate.
 The user is the captain.
 This file is your entire job description.
 
@@ -77,6 +78,7 @@ data/                personal fleet records; LOCAL, gitignored as a whole
   backlog.md         task queue, dependencies, history
   captain.md         this home's domain-local captain preferences and working style; LOCAL, gitignored, canonical even if harness memory mirrors it, and updated with inspect-then-update
   captain-shared.md  main-authoritative shared captain preferences propagated read-only to secondmate homes; LOCAL, gitignored, owned by secondmate-provisioning
+  charter.md         secondmate charter, copied from the parent's charter brief at seed time; LOCAL, gitignored, present only in a secondmate home, and its presence is this home's identity signal
   learnings.md       fleet-local operational facts and gotchas; LOCAL, gitignored; dated, evidence-backed, curated, and updated with inspect-then-update - rewrite and prune rather than append forever, the same contract as captain.md; created lazily, absent until this home has a learning to store
   projects.md        thin fleet navigation registry; firstmate-private, parsed by fm-project-mode.sh (section 6)
   secondmates.md      secondmate routing table; firstmate-private, maintained by fm-home-seed.sh (section 6)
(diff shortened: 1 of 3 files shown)


brchue-ux:
Superseding this with a fresh PR rebased against our own fork (brchue-ux/firstmate), which is now the fleet's actual working source of truth. This branch is 24+ commits behind current main and was opened against the wrong repo for outward contribution purposes. Bug is real and still unfixed - a fresh PR is on the way.

PR #1160, opened 2026-07-28
fix(teardown): remove the scout worktree-safety exemption

ICGNU3:
## What

`validate_worktree_teardown_safety()` returns `0` before any check runs when `KIND` is `secondmate` or `scout`:

```sh
[ "$FORCE" != "--force" ] || return 0
case "$KIND" in
  secondmate|scout) return 0 ;;   # skips dirty / unpushed / unmerged
esac
```

This removes the `scout` half and keeps `secondmate`.

## Why

A teardown sweep across 104 worktrees hit 11 that held uncommitted work and destroyed **46 files**. They were untracked and unstaged, so they had never entered git's object store and were unrecoverable. Branches survived, so committed history was fine; the uncommitted deltas were not.

The safety function itself is well built (checks dirty, unpushed, unmerged; refuses on any). It simply never ran.

### Why `scout` goes

A scout's worktree is a real git worktree, so every check below the exemption applies to it. The stated rationale, that scouts leave scratch behind, is already covered by the dirty filter's `.claude/` and `.fm-*-turnend` exclusions.

The premise that scouts are read-only is false in practice: a scout lane in our fleet authored an ADR and a guard script, both lost.

### Why `secondmate` stays

This half is load-bearing rather than redundant, and worth stating explicitly so it is not "cleaned up" later.

A secondmate home is a plain directory carrying a `.fm-secondmate-home` marker, **not** a git worktree. Every check in the function shells out to `git -C "$WT"`, which fails on a non-repo and correctly fail-safes to `REFUSED`. Removing this exemption makes *every* secondmate teardown impossible. `tests/fm-secondmate-safety.test.sh` catches it immediately ("teardown failed for empty secondmate home").

There is no uncommitted git state there for these checks to protect.

## Also

The same exemption guarded `post_lock_cleanup_check`, the recheck that catches work written between the first safety check and the `treehouse return` that resets the worktree. Leaving it would have kept the hole open on the lock path. `secondmate` never reaches that branch, since the `elif` above it already excludes it, so no exemption is needed there.

`--force` remains the explicit escape hatch in both places.

## Tests

The exemption survived because it was never tested: `tests/fm-teardown.test.sh` was 1,405 lines with **zero** occurrences of `scout` or `secondmate`.

- `test_scout_dirty_worktree_refuses` reproduces the incident against the old code exactly: **exit 0, empty stderr**, worktree torn down with an uncommitted edit present.
- `test_scout_scratch_only_worktree_tears_down` pins the exemption's original rationale, so a scout holding only agent scratch still tears down clean.

Reaching the dirty check as a scout means satisfying two earlier scout gates first (`data/<id>/report.md`, and `decisions_reviewed=1`). Those are about the work product, not the worktree, which is why they are not a substitute for this check.

There is deliberately no `secondmate` twin: a secondmate refuses earlier on the seeded-home check, so such a case would pass with or without the exemption and would only look like coverage.

`tests/fm-teardown.test.sh`: **34/34**, up from 32.

Full suite on this branch shows no new failures. Note that `tests/fm-secondmate-safety.test.sh` currently fails on macOS at `origin/main` for an unrelated reason: `bin/fm-brief.sh` has not parsed under bash 3.2 since #945, which I am happy to send as a separate PR.

### DIFF
--- a/bin/fm-teardown.sh
+++ b/bin/fm-teardown.sh
@@ -668,8 +668,24 @@ validate_worktree_teardown_safety() {
   local dirty_raw dirty unpushed_raw unpushed DEFAULT unmerged_raw unmerged branch
   [ -d "$WT" ] || return 0
   [ "$FORCE" != "--force" ] || return 0
+  # This was `secondmate|scout`, which returned 0 before any check ran and
+  # destroyed 46 uncommitted files across 11 worktrees on 2026-07-26.
+  #
+  # scout is removed: a scout's worktree IS a git worktree, so every check below
+  # applies, and the "scouts leave scratch behind" rationale is already covered by
+  # the dirty filter, which ignores .claude/ and .fm-*-turnend. A scout lane wrote
+  # ADR 0058 and a guard script, so "scouts are read-only" was false in practice.
+  #
+  # secondmate stays, and is load-bearing rather than redundant: a secondmate home
+  # is a plain directory carrying a .fm-secondmate-home marker, not a git worktree
+  # (see the teardown case in tests/fm-secondmate-safety.test.sh). Every check
+  # below shells out to `git -C "$WT"`, which fails on a non-repo and correctly
+  # fail-safes to REFUSED, so dropping this would make every secondmate teardown
+  # impossible. There is no uncommitted git state here for the checks to protect.
+  #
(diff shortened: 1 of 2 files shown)


kunchenguid:

PR #1161, opened 2026-07-28
fix(teardown): recover transient and provably-stale git index.lock

RKucher1:
Hand-ports upstream firstmate's git `index.lock` recovery into `bin/fm-teardown.sh`, written to this repo's conventions rather than cherry-picked.

## The problem

A crew process killed mid-git-operation leaves a worktree `index.lock` behind, so `treehouse return --force` fails with `Unable to create '...index.lock': File exists`. Today that wedges teardown: the main worktree path warns and continues with the worktree un-returned, and the secondmate-home path fails outright.

## The safety invariant

The force-remove branch is the whole risk surface - deleting a lock believed stale corrupts the git index of a **live** worktree. So a lock is removed only when **both** proofs hold together, never either alone:

- `lsof` positively reports nothing holding the lock file **and** nothing holding the worktree directory, **and**
- the lock's mtime age is at least the staleness threshold.

Every state where `lsof` cannot answer - missing binary, run failure, probe timeout, or a warning naming a filesystem that *contains* the target - is UNKNOWN, and UNKNOWN never permits removal. There is deliberately no fallback that removes a lock without `lsof`.

The common case never reaches that branch at all: a transient lock is handled by patience alone (bounded retries). A lock that cannot be safely recovered is left in place and reported as a loud error, never a silent proceed.

## Deliberate divergences from upstream

Recorded in the `FM-LOCK-RECOVERY` provenance comment so a later diff against upstream (#296 `678f2b5`, #435 `38086ea`) reads as intent, not drift:

- **stdout classification.** Upstream captures `lsof <path> 2>&1` and treats any output as an error, so the routine `lsof: WARNING: can't stat() ... file system ...` emitted by any host with docker/NFS/autofs mounts turns a genuine no-match into UNKNOWN - making its own stale-removal path unreachable in practice. This uses `lsof -t`, judges on stdout PIDs, and keeps stderr as a printed diagnostic. The leniency is scoped: a warning whose mount *contains* the probe target still means UNKNOWN.
- **bounded wait, re-proved.** Upstream pairs a 30s threshold with a 3x1s retry window, so a lock born *during* teardown can never reach its own threshold. This waits out the difference, code-clamped, then re-reads age and re-runs the holder check - the timer is never itself the proof.
- **bounded probe.** `lsof` can block forever on a hung NFS/autofs mount, the very hosts this targets. The probe is wrapped in `timeout`/`gtimeout` at 5s where available; a timeout maps to UNKNOWN, so the bound can only ever refuse a removal. Stock macOS runs it unwrapped rather than degrading to a permanent UNKNOWN.

No `fm-backend.sh` or runtime-backend dependency is introduced (upstream routes through one; this teardown stays backend-free), and mtime reading reuses `fm-wake-lib.sh`'s `fm_path_mtime` rather than growing a second primitive.

## Review hardening

Several rounds of review findings are folded in:

- **ambient git env.** `git -C` does not override an inherited `GIT_DIR`/`GIT_WORK_TREE`/`GIT_INDEX_FILE`, and every treehouse worktree shares one repo - so an ambient value pointed the probe and the `rm` at the shared *parent* repo's lock. Resolution now clears those variables and validates the result: it must be named `index.lock` and lie under the worktree or its common git dir, or it is unresolvable.
- **threshold floor.** `FM_STALE_WORKTREE_LOCK_AGE_SECS=0` passed a digits-only check and switched the age proof off entirely. Now floored.
- **no futile stall.** When the clamped wait cannot arithmetically reach the threshold, teardown refuses immediately instead of blocking in the foreground for nothing.
- **branch drop moved.** The task-branch drop needed `git checkout --detach`, which writes the index, so the very lock this recovers from also blocked the drop and every rescued teardown leaked its `fm/<id>` ref. It now runs after a *successful* return, via the project repo; a refused or failed return leaves the branch alone, because its work is still checked out there.
- plus an explicit `return 0` on the no-timeout-binary path, one prefix-matching helper instead of two, and a guarded `rm` that logs a removal only when it actually happened.

## Behavior preserved

The main worktree caller keeps its F11 warn-and-continue husk cleanup, so no window or meta is stranded; the lock-blocked case is surfaced as a loud `ERROR` while non-lock failures keep their existing `warning: treehouse return` wording. The secondmate call site deliberately collapses the lock-refused code into a plain failure, so a still-held lease is never hidden.

## Tests

Developed TDD - tests written first, watched fail, then implemented. Each direction is proven by attempting what it forbids: a live-held lock is never removed; a provably stale one is removed and the return retried; an `lsof` that cannot answer blocks removal and fails loudly; a transient lock recovers on retry with the force-remove branch never entered (asserted by proving `lsof` was never invoked); a too-young lock is waited out and then re-proved; both call sites are exercised; and both branch-drop directions are pinned. Includes a regression test using this host's real `lsof` warning shape, since that is the bug that made the removal path unreachable in the first place.

`tests/fm-teardown-lock-recovery.test.sh` 27 cases, `tests/fm-teardown.test.sh` 21 cases.


RKucher1:
Opened in error - gh defaulted to the fork parent. This belongs on RKucher1/firstmate, not here. Closing; apologies for the noise.

PR #1169, opened 2026-07-28
feat(brief): add token-frugality rule to crewmate briefs

rishabhxverma:
## What changed

Lands the generalizable half of an unlanded local commit (`130d477`, "feat: token-frugality rules in crewmate contracts and supervision discipline") that never made it upstream before origin/main moved 96 commits ahead. That commit mixed two unrelated changes, so this PR splits them:

**Ported (generalizable, useful to every firstmate user):**
- `bin/fm-brief.sh`: added a numbered "Token frugality" rule to both the scout and ship brief templates (targeted reads over start-to-finish streaming, staged/script-parsed bulk data, capped and projected CLI/API output, scope kept to the task's files).
- `AGENTS.md`: added a cross-reference sentence in the supervision section stating firstmate's own token frugality mirrors the crewmate brief's rule. The original commit's anchor paragraph ("Token discipline: ...") no longer exists - it was removed by a later compression pass (#626) - so this re-applies the intent as a fresh one-line cross-reference rather than restating the crewmate-brief contract, per this repo's one-owner-rule documentation style.

**Deliberately NOT ported, and why:**
The original commit also rewrote the top-of-file address rules to hardcode one specific user's name and preferences ("The user is Rishabh...", "Do NOT address the user as captain"). This repo is a shared template used by many firstmate users, so a personal preference must never be hardcoded into it.

Instead, this PR adds the *generalizable* version of that idea: one line in AGENTS.md's existing address paragraph stating that the default form of address is itself a captain preference, and that `data/captain.md` - already the documented canonical home for captain preferences (AGENTS.md sections 2 and 6) - wins over the "captain" default when it specifies one. No personal name or preference is hardcoded anywhere in tracked files.

**Unrelated bug fix found during verification:**
While scaffolding throwaway scout and ship briefs to confirm the new rule renders correctly, the default (`no-mistakes` mode) ship brief failed to generate at all: `bash -n bin/fm-brief.sh` errored with `unexpected EOF while looking for matching ')'`. The cause was a stray apostrophe in "firstmate's authority check" inside a heredoc built via `$(cat <<EOF ... EOF)` - bash's quote-balance scan for the matching `)` breaks on an unescaped apostrophe inside such a heredoc. This is the same bug class as a previously-fixed issue (#166; see `test_no_mistakes_dod_wording` in `tests/fm-brief.test.sh`), reintroduced later via different wording that the existing string-matching regression test didn't cover. Fixed by rewording to "the firstmate authority check" (no apostrophe). Confirmed via `git stash` bisection that this reproduces identically on vanilla origin/main.

## Testing
- `bin/fm-lint.sh`: clean (ShellCheck 0.11.0, pinned version).
- `bash tests/fm-brief.test.sh`: full suite passes (16/16), including a new `test_token_frugality_rule_in_scout_and_ship` covering the new rule in both brief kinds.
- Manually scaffolded a throwaway scout brief and a throwaway ship brief via `bin/fm-brief.sh` and inspected `brief.md` to confirm rule 8's token-frugality text (and the ship-specific "stay within the files the task needs" clause) renders verbatim.
- Confirmed (via `git stash`) that both the apostrophe/heredoc bug and its generic regression coverage (`test_script_parses`) reproduce identically on vanilla origin/main without this branch's changes.

## Note
`tests/fm-test-run.test.sh` has two timing-sensitive scheduler assertions that flake in the local sandbox this branch was built in; confirmed via `git stash` that this reproduces identically on vanilla origin/main, so it is pre-existing environment-timing flakiness unrelated to this change and was left untouched.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,16 @@
+<!--
+STOP: do not open this pull request by hand.
+
+Pull requests targeting `main` must be raised through no-mistakes:
+https://github.com/kunchenguid/no-mistakes
+
+Run `git push no-mistakes` on your branch instead.
+That pipeline runs the required review/test/lint/CI steps, opens the PR for you,
+and writes the deterministic `## Pipeline` signature that the required
+`Require no-mistakes` check looks for in the PR body.
+
+A PR opened manually will fail that check and cannot be merged, even if every
+other check is green. Close it and re-push through the gate.
+
+See CONTRIBUTING.md for setup and the full workflow.
+-->
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -9,6 +9,7 @@ This is mandatory respectful address, not performance: it applies even when deli
 Do not force it into every sentence, but never send a response with zero direct address.
 Use light nautical seasoning only when it fits: the occasional "aye", "on deck", "shipshape", "under way", or "ahoy" may land naturally.
 Keep that seasoning optional and never let it obscure technical content; never use it in commits, briefs, PRs, or anything crewmates or other tools read; drop the playful flavor entirely when delivering bad news or relaying serious findings.
+This default form of address is itself a captain preference: when `data/captain.md` specifies a different one, it wins over "captain" here.
 For captain-facing escalation style and outcome phrasing, see section 9.
 
 ## 1. Identity and prime directives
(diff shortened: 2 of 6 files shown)


rishabhxverma:
Closing: opened against upstream by mistake. Re-raised on our own fork at https://github.com/rishabhxverma/firstmate/pull/1

timidri (commented):
⚠️ Mergeable with caveats

This correctly makes the plan's own interval authoritative for `billing_period`, filters every candidate (plan, webhook, checkout) through the same enum the `Subscription` model already enforces, and adds the missing check at the checkout boundary before any Frisbii call is made. I traced the resolution logic by hand and ran the new/changed specs locally (105+118+10 examples, 0 failures; rubocop clean; swagger regen shows no drift) — this closes the specific off-enum-`billing_period`-strands-a-paid-customer path described in the ticket.

Two things I'd want addressed or explicitly acknowledged before merge:

1. `subscription_manager.rb:146-148`'s `rescue ActiveRecord::RecordInvalid` still wraps the whole `Subscription.create!` call, and `status` (`:121`) reaches it straight from the webhook payload with no validation at all. An off-enum `status` from Frisbii would hit the exact same silent swallow this PR is meant to close, just via a different field. Worth a fast-follow ticket if it's out of scope here — but let's not close this out believing the rescue itself is now safe.
2. The spec doc says the reconciliation backfill query must run against production "before #1158 merges" — https://github.com/eufforic/office.eu/pull/1158 already merged before this PR was opened. Can someone confirm that query was run (or run it) before we merge this? If a subscription row is already mismatched in production, the add-on order-preview endpoint is likely already 422-ing for that customer as of #1158 landing, independent of this PR. I'm checking this now and will report the count here.

Security (CIA): 🟢 no new concerns, net improvement on integrity. No new authz surface and no new data returned to the client — the only new response is a fixed 422 message, not raw exception text. Input validation lands at the correct boundary (checkout controller, ahead of any external call) and the duplicated enum list is collapsed into one constant. Availability is unaffected: the enum check is a local comparison that can only reduce Frisbii round-trips. The residual item is finding 1 above — not a regression, but not newly closed either.

Nothing here blocks on code quality or test coverage — both are solid. This is a "confirm the two open items, then ship" situation.

_Reviewed by office.eu pr-review_

PR #1185, opened 2026-07-28
fix(spawn): keep the axi tools on PATH for spawned agents

robinbraemer:
## The defect

The axi tools (`tasks-axi`, `gh-axi`, `chrome-devtools-axi`, `lavish-axi`) are npm globals.
They install into whichever Node toolchain was active at install time, for example a mise-managed `~/.local/share/mise/installs/node/<version>/bin`.
That directory is on `PATH` only while that toolchain is the one being resolved.

`bin/fm-spawn.sh` launches an agent in a different project directory, and mise resolves toolchains per directory.
When the target project selects a different toolchain, the tools drop off `PATH` **all at once**.
A project pinning its own Node version does it, and so does a project whose config file mise declines to trust, since the failed config load takes the whole toolchain with it.

Nothing is uninstalled; the tools are simply unreachable from where the agent runs.

Two concrete consequences:

- `bin/fm-decision-hold.sh` refuses outright with `compatible tasks-axi is required`, so that path cannot be used at all from a spawned agent's worktree.
- Every generated brief instructs the agent to use `gh-axi` for GitHub operations. That instruction is silently false. An agent may notice and fall back to plain `gh`, or may not.

This is not specific to any one project layout. Any user whose axi tools live in a version-managed Node toolchain hits it as soon as an agent works in a directory that resolves a different toolchain.

## The fix

`bin/fm-spawn.sh` resolves the tool directory in the spawning process's own environment, where the tools do resolve, and exports it into the spawned agent's shell.
The resolution logic lives in a new `bin/fm-axi-path-lib.sh`.

Four properties worth checking in review:

1. **Resolved dynamically, with no hardcoded Node version.** The directory comes from `command -v` on the tools themselves, then `dirname`. A hardcoded `node/24` path would keep working until the day it becomes `node/25`, then silently recreate this exact bug. No Node version, mise layout, or install prefix is assumed anywhere in the change.

2. **The unresolvable case fails loudly at spawn.** If no axi tool resolves, the spawn is refused with an actionable message naming the missing tools and the likely cause, before any session endpoint or task metadata is created. The alternative — launching an agent that has been told to use tools it cannot see — moves the failure deep inside the task, where it is far more expensive. A single uninstalled tool while others resolve is an install gap rather than a `PATH` gap, so that warns loudly and proceeds rather than blocking every spawn.

3. **Proven by running the tools, not by printing `PATH`.** A directory being present on `PATH` is not evidence a binary executes. Verified in a shell in a directory that reproduces the bug (a project whose own mise config selects a different toolchain), before and after:

   ```
   BEFORE                          AFTER
   tasks-axi: not found            tasks-axi           0.2.2
   gh-axi: not found               gh-axi              0.1.25
   chrome-devtools-axi: not found  chrome-devtools-axi 0.1.26
   lavish-axi: not found           lavish-axi          0.1.37

   bin/fm-decision-hold.sh:
   "compatible tasks-axi is required"   ->   registers the hold successfully
   ```

   `fm-decision-hold.sh` is called out specifically because it is the concrete thing that was broken.

4. **The failure path is tested, not only the happy path.** `tests/fm-axi-path.test.sh` adds 8 cases. Two pin the refusal: the actionable message, and that `fm-spawn.sh` refuses before any endpoint or task metadata exists. The rest cover dynamic resolution against two unrelated install locations, deduplication, split installs, and the warn-but-proceed path.

The directory is **appended** to `PATH` rather than prepended, on purpose: it is also a Node toolchain `bin` directory, and prepending it would shadow a project's own pinned `node`/`npm`. Appending still resolves the axi tools, since no other directory on the agent's `PATH` provides those names.

## Regressions this change caused, and their correction

An earlier revision of this change made `tasks-axi` individually mandatory, so a spawn refused whenever that one tool was missing.
That was stricter than the defect requires and it broke four spawn-related suites that deliberately run with a stripped-down `PATH`: `fm-secondmate-sync`, `fm-shared-captain-inheritance`, `fm-session-start`, and `fm-kimi-harness`.

The refusal now triggers only when no axi tool resolves at all, which is the actual signature of a toolchain resolving away, since it takes all four together.
That fixed two of the four outright.
The other two had fixtures with no axi tools present at all and now supply stubs, since the spawn genuinely requires resolving the directory.
All four pass.

## Test results

Locally, 91 of 101 suites run; the remaining 10 are gated on real Herdr.

- Both portable parallel lanes: 0 failures.
- Portable serial lane: 3 failures, all pre-existing.
- `bin/fm-lint.sh`, the coverage guard, and `bin/fm-doc-audience-check.sh` are all clean.

The three serial failures were **proven** pre-existing rather than assumed. With this change fully removed and the working tree reverted to base `fa0d85d`, all three fail identically: a real-zellij pane check reporting an empty working directory, a bootstrap check that wants `cmux` installed on the host, and a signed-wrapper harness-ancestry check. None of them touch spawning or `PATH` resolution.

## Coverage limits, stated plainly

- The automated tests exercise resolution and the refusal through the tmux backend only. The export reaches the herdr, zellij, orca, and cmux backends via the same shared send path that already carries the existing `GOTMPDIR` export, so it is structurally identical, but it is not separately covered by a test.
- The end-to-end proof from a toolchain-pinned worktree, shown above, was performed manually. It is not automated, since it depends on a second project's toolchain configuration being present on the host.

## Notes

- CI's Herdr lane now installs `tasks-axi`, because that lane performs real spawns and would otherwise hit the new refusal.
- No change to any global mise configuration, shell profile, or installed package. The fix is entirely at spawn time, where the problem is created.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -170,6 +170,13 @@ jobs:
           command -v tar >/dev/null || { echo "::error::tar is required"; exit 1; }
           jq --version
           python3 --version
+      - name: Install tasks-axi
+        # fm-spawn.sh refuses to launch when it cannot resolve the axi agent tool
+        # directory, so the spawning Herdr lane needs it on PATH.
+        run: |
+          set -eu
+          npm install -g tasks-axi
+          tasks-axi --version
       - name: Install pinned Herdr
         run: |
           set -eu
--- a/bin/fm-axi-path-lib.sh
+++ b/bin/fm-axi-path-lib.sh
@@ -1,0 +1,87 @@
+#!/usr/bin/env bash
+# fm-axi-path-lib.sh - resolve the directory holding the axi agent tools so a
+# spawned direct report can actually run them.
+#
+# The hazard: the axi tools (tasks-axi, gh-axi, chrome-devtools-axi, lavish-axi)
+# are npm globals installed into whichever Node toolchain was active when they
+# were installed, e.g. a mise-managed ~/.local/share/mise/installs/node/<ver>/bin.
+# That directory is on PATH only while that toolchain resolves. A crewmate works
+# in a different project directory, and a project that pins its own toolchain
+# (or ships a config file mise refuses to trust) resolves away from it, so the
+# axi tools vanish from the crewmate's PATH ALL AT ONCE. Nothing is uninstalled;
+# they are simply unreachable from where crewmates run. The generated brief still
+# tells the crewmate to use gh-axi, and bin/fm-decision-hold.sh hard-requires
+# tasks-axi, so the crewmate fails deep inside its task instead of at launch.
+#
+# fm-spawn.sh therefore resolves the directory HERE, in firstmate's own
+# environment where the tools do resolve, and exports it into the spawned
+# agent's shell.
+#
+# Three properties this deliberately holds:
+#
+#   1. The directory is resolved dynamically from where the tools actually live
+#      (command -v, then dirname). A hardcoded node/<major> path would break
+#      silently on the next toolchain bump and recreate this bug with extra
+#      steps, so no Node version, mise layout, or install prefix is assumed.
+#   2. It fails closed when NO axi tool resolves at all. That is the signature of
+#      the actual bug - a whole toolchain resolving away - and it leaves no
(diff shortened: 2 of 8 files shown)


kunchenguid:

PR #1193, opened 2026-07-28
feat(bin): add read-only fleet claim reconciler [ci]

bramdokman:
## Defect class

Fleet records and delivery systems can claim success while observable reality disagrees. This change adds a read-only reconciler that prints only those disagreements and exits distinctly for clean (`0`), divergence (`1`), and could-not-verify (`2`).

The concrete incidents behind it were:

- a required CI gate reported success while every real job was skipped;
- deploys reported success while shipping nothing;
- healthy changes were silently ejected from the merge queue;
- worker cleanup failed quietly and retained a finished worker copy;
- queued worker records aged into falsehood after queue removal; and
- `main` sat red with four failing checks without being surfaced.

## What it checks

`bin/fm-fleet-reconcile.sh` compares durable firstmate claims with GitHub REST observations and reports only disagreements:

- a task recorded as landed whose PR is not merged;
- a merged PR whose recorded worker copy still exists;
- a ready-for-review open PR whose checks completed green but zero real test jobs ran;
- a successful deploy that ran only its selector job, within an age-bounded observation window; and
- failure-like checks on each repository's observed default branch.

It reads PR claims through the shared hardened `fm_pr_metadata_identity_parse` owner. Duplicate, aliased, or unparseable claims become could-not-verify. GitHub responses must use the documented `gh-axi` envelope; arbitrary JSON is rejected. Reads use REST only and never mutate GitHub or fleet state.

## Applicability, incompleteness, and exclusions

The vacuous-green detector distinguishes the PR #898 docs-only not-applicable case from a never-evaluated change:

- an explicit successful not-applicable attestation reconciles and is not counted as a real test;
- a successful applicability/detection/filter job whose decision is not exposed in check-run names becomes could-not-verify;
- completed successful checks with zero real tests and no evaluator are a divergence;
- still-running or queued checks are silent because incomplete is not untrue; and
- current draft PRs are excluded because drafts make no readiness claim. The exclusion uses the currently observed `draft` field, so marking a PR ready immediately returns it to scope.

It deliberately does not query merge-queue state on a schedule, infer worker status verbs, or apply an old-open-PR ejection heuristic. It is not wired into supervision, watcher, or teardown in this PR.

## Known gap discovered during this PR's own delivery

This reconciler **would not flag PR #1193's current approval-blocked CI state**. That is a real gap.

REST observation of head `9112724f555f76d4fe314146ecd436bb95baf0f1` showed:

```text
check-runs: total_count = 0
combined commit status: state = pending, total_count = 0
workflow runs:
  CI                    status = completed, conclusion = action_required
  Require no-mistakes   status = completed, conclusion = action_required
```

The current vacuous-green predicate starts only when at least one successful check-run exists. With zero check-runs it computes `green=0` and emits no vacuous verdict. It also does not query Actions workflow-run conclusions and has no independent input representing a supervisor's claim that CI is green. Therefore a never-started, approval-blocked workflow is missed rather than reported.

This surfaced an exact claim-versus-reality failure in the delivery of the reconciler itself: local tests and canonical lint passed, but GitHub CI had not run. The approval-blocked workflows must not be described as green. Extending the observer to reconcile never-started/action-required workflow runs against an explicit CI-green claim is follow-up scope, not silently implied coverage here.

## Evidence

The behavior suite contains 21 passing tests. Its REST double feeds real response bodies through the production jq filters rather than returning pre-filtered answers. Coverage includes:

- a deliberately constructed vacuous green:

  ```text
  divergence: acme/app PR #7 completed 1 successful checks, but zero real test jobs executed and no applicability job evaluated the change
  exit: 1
  ```

- draft exclusion and immediate ready-for-review re-entry;
- explicit not-applicable versus opaque applicability evaluation;
- real main-branch failure/action selection, including cancelled/stale as could-not-verify;
- per-repository default branches;
- selector-only deploy detection and age expiry;
- strict API-envelope rejection; and
- unavailable GitHub degradation:

  ```text
  could not verify GitHub observations: authentication, network access, or REST quota is unavailable
  exit: 2
  ```

A live read-only scratch-fleet run caught two real divergences:

```text
divergence: task ship-1172 has merged kunchenguid/firstmate PR #1172, but worker copy <scratch>/worktrees/ship-1172 still exists
divergence: task zellij-supervisor is recorded Done, but kunchenguid/firstmate PR #1189 is not merged
exit: 1
```

The live fleet-state hash was identical before and after observation. Canonical `bin/fm-lint.sh` passed on the exact pushed head with ShellCheck 0.11.0. That local validation is real evidence, but it is explicitly **not** evidence that GitHub CI passed.

## Follow-up recommendation

After the known never-started/action-required gap is decided, invoke this command from the supervision heartbeat in a separate wiring change. Keep the observer read-only, REST-only, and divergence-only so silence remains meaningful.


bramdokman:
Closing this at the captain's request, with an honest assessment rather than a quiet withdrawal.

The **thinking** in here holds up: distinguishing a reconciled state from an out-of-scope observation from a check that could not be carried out, and giving that third case its own exit code, is the property worth keeping. Three review rounds each found real defects — a detection path no test exercised, an alarm that could never clear, and a project-specific job name in a shared default.

The **vehicle** is the problem. 907 lines of bash for a correctness tool is a weak foundation, and classifying work by regex against job names means a rename silently miscategorises. Six configurable patterns are six places to be misconfigured, and the failure is quiet.

Reopening a better-shaped version later is preferable to landing this and maintaining it. No reflection on the review, which was consistently sharp.

PR #1198, opened 2026-07-28
fix(brief): parse fm-brief.sh on bash 3.2

jericho1050:
## Problem

`bin/fm-brief.sh` failed to parse on the bash that ships with macOS (3.2.57), which blocked every brief scaffold and therefore every crewmate dispatch on a stock Mac.

Bash 3.2 scans a command substitution for its closing parenthesis as a flat character stream and does not honour here-document boundaries, so the body of `VAR=$(cat <<EOF ... EOF)` is read as shell syntax. The apostrophe in `firstmate's` was taken as an opening quote that never closed:

```
```

Bash 4+ parses it correctly, so Linux CI never saw it.

## Approach: fix the construct, not the prose

Rewording dodges one instance but leaves the trap armed for the next contributor who writes a possessive, and only on macOS. A new `capture_block` helper reads here-documents with `read`, keeping the body out of any command substitution. All four capture sites use it.

The prose is deliberately unchanged, apostrophe included. `firstmate's` is still there. Preserving it is what proves the fix works.

## Two corrections to the original issue #166 diagnosis

Both verified empirically against bash 3.2.57 rather than assumed:

1. **Quoting the delimiter does not help.** `<<'EOF'` only affects expansion of the body, which is a later stage than the scan that already failed. The Herdr block previously believed immune was equally armed, and is converted too.
2. **Apostrophes are not the trigger.** Any unbalanced quote, backtick, or parenthesis breaks it. Even `it's a 'quote'` fails, on odd apostrophe parity.

So the defect is the construct, not the text inside it.

## Regression coverage

Extends `tests/fm-brief.test.sh` and the existing runner rather than adding a new one.

- **Structural guard** asserts no here-document in `fm-brief.sh` is captured by a command substitution. It checks source text, so it holds under every bash version. It also asserts `capture_block` is still the mechanism in use, so it cannot pass vacuously if the sections are deleted.
- **Repo-wide bash 3.2 parse check** across all 92 scripts in `bin/` and `bin/backends/`, wired into the **existing** `macos-stock-bash` CI job, which already pins `$BASH_VERSION` to `3.2.57*`. No new workflow, job, or matrix entry. This replaces an earlier design that skipped when no bash 3.x was present, and makes the guarantee unconditional on CI rather than weaker there.

`skip()` was removed rather than reformatted: it emitted `ok - ... # SKIP ...`, which matches the `grep -c '^ok - '` exact-count assertions in that same job (15 and 42). A skipped check would have counted as a passing one and silently lost coverage.

The bash 3.2 floor is now documented, including what to do when bash 4+ syntax is genuinely needed.

## Verification

| Check | Result |
|---|---|
| `/bin/bash -n` across `bin/` + `bin/backends/` on bash 3.2.57 | 92/92 (was 91/92) |
| Rendered brief output vs pre-change baseline | Byte-identical, 10 mode variants, under bash 3.2 and 5.3 |
| New test against unfixed script | Fails, as intended |
| New test against fixed script | Passes |
| `bin/fm-lint.sh` (pinned ShellCheck 0.11.0) | Pass |
| `ci.yml` structure | 9 jobs, no stray key, run block 39 lines |
| CI exact-count assertions | 15 and 42, unchanged |

Byte-identity was proven by capturing every mode variant before and after and diffing, not by inspection.

## Known follow-up, deliberately not in this PR

Workflow YAML has no structural validation. This branch demonstrated the gap on itself: an intermediate commit shipped a `run: |` block whose indentation silently closed the scalar and promoted a shell variable to a top-level job key. It parsed as valid YAML, so a linter would not have caught it, and every repo assertion about `ci.yml` is a string grep. Filed as separate work rather than widening this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -322,7 +322,23 @@ jobs:
           esac
           /bin/bash --version | head -1
           command -v jq >/dev/null || { echo "::error::jq is required"; exit 1; }
-          /bin/bash -n bin/fm-fleet-snapshot.sh
+
+          # Every shipped script must parse on the bash macOS ships, not just the
+          # snapshot ones: bash 3.2 reads a command substitution as a flat
+          # character stream, so a construct that is fine under bash 4+ can make a
+          # whole script unparseable here and nowhere else.
+          parse_failures=''
+          for script in bin/*.sh bin/backends/*.sh; do
+            [ -f "$script" ] || continue
+            if ! script_error=$(/bin/bash -n "$script" 2>&1); then
+              parse_failures="$parse_failures
+  $script: ${script_error:-no diagnostic emitted}"
+            fi
+          done
+          [ -z "$parse_failures" ] || {
+            echo "::error::these scripts do not parse under stock macOS Bash $BASH_VERSION:$parse_failures"
+            exit 1
+          }
 
           snapshot_output=$(/bin/bash tests/fm-fleet-snapshot-view.test.sh)
           printf '%s\n' "$snapshot_output"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -305,12 +305,12 @@ jobs:
           if-no-files-found: warn
 
   macos-stock-bash:
-    name: Stock macOS Bash snapshot compatibility
+    name: Stock macOS Bash compatibility
     runs-on: macos-latest
     timeout-minutes: 10
     steps:
       - uses: actions/checkout@v6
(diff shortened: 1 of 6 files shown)


kunchenguid:
thanks for digging into this. i merged https://github.com/kunchenguid/firstmate/pull/1093, which landed the same construct-level fix you took here - heredocs read outside command substitution with the apostrophe kept as the canary, a structural guard, and the full parse sweep on the stock macOS CI lane. closing this one as superseded. really appreciate the work here.

PR #1206, opened 2026-07-28
fix(session-lock): resolve Claude bg-spare ancestry to the outermost claude pid

trillium:
## Intent

Fix bin/fm-session-lock-lib.sh's fm_harness_ancestry_pid() so the Claude Stop auto-arm hook (bin/fm-claude-stop-autoarm.sh) correctly identifies the session's lock-owning process under Claude Code's bg-spare hook-worker architecture. The Stop hook fires several process levels below the actual session root (hook shell -> claude bg-spare -> claude bg-pty-host -> claude -> claude(lock)), and the prior code returned the FIRST ancestor matching a harness name (the bg-spare worker), which never matched the recorded lock pid, so fm_session_lock_owned_by_self() always returned false and the watcher silently never armed. Fix: keep walking past a claude-named match to find a still-more-ancestral claude-named match, stopping at the first non-match that follows an already-found match (bounds the walk to a contiguous run rather than the literal ancestry top, since an unrelated real claude-named process further up the actual process tree, e.g. this very live coding session that launches its own tests as subprocesses, must not be mistaken for part of the target session's own nested chain). Deliberately scoped this outermost-of-contiguous-run behavior to claude-named matches only: every other harness keeps the original first-match-wins (innermost) semantics, because Pi's shared signed-wrapper architecture is the opposite shape - the actual session lock is held by the INNER pi engine pid, with an outer pi-signed wrapper pid above it that is not the session owner (verified via the existing tests/fm-secondmate-harness.test.sh pi-signed identity test, which intentionally asserts the inner pid). Hop limit raised from 8 to 16 to cover the deeper bg-spare chain. Verified: bin/fm-lint.sh clean, all 17 tests in tests/fm-claude-stop-autoarm.test.sh pass, and tests/fm-secondmate-harness.test.sh's pi-signed ancestry/liveness assertions still pass (its one remaining failing test in that suite is a pre-existing environmental flake reproducible on unmodified main, unrelated to this change). Added a one-line dated verification note to docs/verification/supervision.md under the Turn-end guard section.

## What Changed

- `fm_harness_ancestry_pid()` in `bin/fm-session-lock-lib.sh` now walks past an initial `claude`-named ancestor match to find the outermost pid in a contiguous `claude`-named run, fixing session-lock ownership detection under Claude Code's bg-spare hook-worker chain (hook → bg-spare → bg-pty-host → claude → claude(lock)); it stops at the first non-match following an already-found match, so an unrelated `claude` process further up the real process tree isn't absorbed. All other harnesses keep the original first-match (innermost) semantics. The ancestry walk's hop limit was raised from 8 to 16 to cover the deeper chain.
- Added a regression test in `tests/fm-claude-stop-autoarm.test.sh` covering the nested bg-spare ancestry chain, and corrected `docs/sessionstart-nudge.md`'s stated hop-depth claim to match the updated limit.
- Added a dated verification note to `docs/verification/supervision.md`.

## Risk Assessment

✅ Low: Both round-1 findings (missing nested-chain regression test and stale doc-parity claim) were fixed correctly; the new test's use of a piped invocation reliably prevents bash's tail-exec pid collapse (verified empirically), so it genuinely exercises the two-hop claude ancestry the fix targets, and the doc edit accurately decouples the two independent ancestry walks without changing sessionstart-nudge behavior.

## Testing

Ran the full fm-claude-stop-autoarm test suite on the target commit (all 17 pass, including the new nested-bg-spare-chain regression test that directly exercises the actual Stop hook script end-to-end), then reproduced the bug by running that same new regression test against the pre-fix library code on the base commit, where it failed exactly as described (hook never armed, exit 0 instead of 2) — a clean before/after demonstration that the fix resolves the reported issue. Also confirmed the one failing test in tests/fm-secondmate-harness.test.sh is a pre-existing environmental flake by reproducing it identically on the unmodified base commit, unrelated to this change. No findings to report.

<details>
<summary>Evidence: Full fm-claude-stop-autoarm.test.sh run on target commit (all 17 tests pass, including the bg-spare nested-ancestry regression test)</summary>

```text
ok - settings.json registers the asyncRewake auto-arm with timeout >= 28800 and a foreground arm
ok - auto-arm: inert in a linked child worktree even when in-flight
ok - auto-arm: inert with no session lock
ok - auto-arm: a demonstrably dead recorded session owner is reclaimed through fm-lock.sh before arming
ok - auto-arm: inert without arm, rewake, or lock replacement when another live harness owns the home
ok - auto-arm: inert while AFK owns supervision
ok - auto-arm: stale-owner recovery leaves the AFK and supervision-need gates unchanged
ok - auto-arm: resolves the outermost pid of a nested contiguous claude ancestry (bg-spare chain)
ok - auto-arm: inert with nothing in flight and no X-mode need
ok - auto-arm: actionable close translates to exactly one exit-2 rewake with reason
ok - auto-arm: watcher: FAILED translates to an exit-2 alarm rewake
ok - auto-arm: clean close exits silently with a clean epoch
ok - auto-arm: X-mode poll need arms the cycle even with no tasks in flight
ok - auto-arm: concurrent firings admit one owner and one rewake translation
ok - auto-arm: need vanishing mid-cycle closes without a rewake
ok - auto-arm: mid-cycle AFK hands triage to the daemon with no rewake
ok - auto-arm: active in a marked secondmate home
ok - fm-lock: shared session-lock lib preserves the status path
```
</details>
<details>
<summary>Evidence: Before/after CLI transcript: same regression test fails on pre-fix code (exit 0 instead of 2, hook never arms) and passes on the fix</summary>

```text
AFTER (target commit caead41, fixed fm-session-lock-lib.sh):
ok - auto-arm: resolves the outermost pid of a nested contiguous claude ancestry (bg-spare chain)

BEFORE (base commit fa0d85d, same new test copied on top of the old library):
not ok - a nested contiguous claude ancestry must resolve to the outer lock-owning pid and arm: expected exit 2, got 0
```
</details>

## Pipeline

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>

- ⚠️ `tests/fm-claude-stop-autoarm.test.sh` - The fix&#39;s whole premise is a multi-level contiguous &#39;claude&#39;-named ancestry chain (bg-spare -&gt; bg-pty-host -&gt; claude -&gt; claude(lock)), and the new fm_harness_ancestry_pid() logic in bin/fm-session-lock-lib.sh specifically walks past a claude match looking for a still-more-ancestral one. No test in this diff, in tests/fm-claude-stop-autoarm.test.sh, or in tests/fm-secondmate-harness.test.sh&#39;s pi-signed case exercises more than a single fake &#39;claude&#39; process as the direct parent of the hook. tests/fm-claude-stop-autoarm.test.sh&#39;s run_autoarm() always launches the hook as a direct child of one fake claude PID (ln -s /bin/bash &#34;$FAKEBIN/claude&#34;), so every passing assertion only proves the unchanged single-hop-match path still works, not the new &#39;extending&#39; behavior that resolves the outermost pid of a contiguous run. A future edit that silently reverts to first-match-wins for claude (reintroducing the exact bug this PR fixes) would pass the full suite undetected.
- ⚠️ `docs/sessionstart-nudge.md:15` - docs/sessionstart-nudge.md:15 states bin/fm-sessionstart-nudge.sh walks &#39;at most eight parents ... matching bin/fm-lock.sh ... ancestry depth&#39;, but this PR raised bin/fm-session-lock-lib.sh&#39;s fm_harness_ancestry_pid() (used by bin/fm-lock.sh) from 8 to 16 hops while bin/fm-sessionstart-nudge.sh keeps its own independent, hard-coded 8-hop loop (bin/fm-sessionstart-nudge.sh:31) untouched. The two are now out of sync: the doc&#39;s parity claim is false, and if fm-lock.sh now records a more-ancestral outermost claude pid as the session lock owner (the same bg-spare-chain depth problem this PR targets could equally apply to a SessionStart hook invocation), fm-sessionstart-nudge.sh&#39;s shallower, differently-implemented self-ancestry search may never find that recorded pid within 8 hops and would keep re-printing the session-start nudge every hook firing instead of recognizing an already-started session.
- ℹ️ `bin/fm-session-lock-lib.sh:50` - While extending=1 (already inside a contiguous claude-named run), if the very next ancestor matches a different, non-claude harness name (hit=1, is_claude=0), the code sets best=pid to that different harness&#39;s pid and then breaks, discarding the previously-found claude match entirely rather than stopping and returning the last claude pid. This contradicts the documented &#39;stop at the first non-match that follows an already-found match&#39; contract, since a differently-named harness match is treated as a fresh first-match rather than a stopping boundary. The scenario (two distinct named harness processes directly parent-child) is unlikely in real process trees, so this is unlikely to trigger in practice.

🔧 Fix: Add nested-claude-ancestry regression test; fix nudge doc depth claim

</details>

<details>

✅ No issues found.
- <code>`bash tests/fm-claude-stop-autoarm.test.sh` on target commit caead41 — all 17 tests pass, including the new `test_resolves_outermost_claude_pid_in_nested_bgspare_chain` regression test</code>
- <code>`bash tests/fm-claude-stop-autoarm.test.sh` re-run with only the new regression test copied onto pre-fix base commit fa0d85d (old fm-session-lock-lib.sh) — fails with `expected exit 2, got 0`, proving the hook silently never armed before the fix</code>
- <code>`bash tests/fm-secondmate-harness.test.sh` on both target and base commits — same single pre-existing failure (&#39;unmarked shared signed-wrapper ancestry resolved claude, expected pi&#39;) reproduces identically on both, confirming it is an unrelated environmental flake and not a regression</code>
- <code>`git status --short` before and after testing confirms the worktree was left clean</code>

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>


kunchenguid:
Thanks @trillium - merged! Really appreciate the contribution.

PR #1237, opened 2026-07-29
fix(herdr): add blocked-transition confirm delay to suppress transient OpenCode false alarms

burakgizlice:
## Problem

Herdr reports agent-status changes via its transition event stream. When a pane
transitions to **blocked**, firstmate immediately escalates it as a stale wake,
with the message: `stale: <window> (herdr: agent blocked - waiting on human,
escalated immediately, not via wedge timer)`.

The OpenCode TUI briefly shows a "waiting for input" state between tool calls
as the agent cycles through its working loop. Herdr correctly reports these
transitions as `blocked` -> `working` -> `blocked` -> `working` on every tool
call. The existing per-pane dedupe marker prevents re-escalating the same
**blocked** edge, but each `working` transition clears the marker, so the next
`blocked` edge is treated as a fresh actionable event -- producing a firehose
of stale wake notifications every few seconds.

This is especially visible during:
- Normal agent working cycles (transient prompts between tool calls)
- Long-running commands like `sleep 600 && no-mistakes axi status` (not a
  wedge, just the agent waiting for itself)

## Solution

Add a **re-check step** in `fm_backend_herdr_apply_transition` (the herdr
backend function that routes transition events through the shared policy
table). When a `blocked` transition arrives:

1. Wait **3 seconds** (configurable via `FM_BACKEND_HERDR_BLOCKED_CONFIRM_DELAY`).
2. Re-query the live agent status via herdr's `agent.get` API.
3. If the agent has already resumed **working** (or gone idle/done),
   absorb the transition silently -- this was a transient false alarm.
4. If the agent is **still blocked**, escalate as before -- this is a real
   wedge (permission dialog, stuck prompt, interactive menu).

This handles all three cases correctly:
- **Transient prompts** (1-2s): after 3s the agent is `working` -> absorbed.
- **Long-running commands** (sleep 600): herdr correctly reports `working`
  during command execution, so no `blocked` transition is emitted at all.
- **Real wedges**: the agent stays `blocked` for 3+ seconds -> escalated.

The 3-second delay is a deliberate trade-off: a real wedge is delayed by
3 seconds, but the firehose of false alarms (dozens per minute) is eliminated.
The captain sees a notification only when something is genuinely wrong.

## Config

`FM_BACKEND_HERDR_BLOCKED_CONFIRM_DELAY` (default 3, in seconds). Set to 0
to skip the re-check entirely (used by tests that lack a live herdr server).

## Additional change

Also adds a **git-identity rule** (rule 8) to the crewmate brief scaffold
(`bin/fm-brief.sh`). Workers sometimes invent an agent identity
(`firstmate <firstmate@local>`) when committing, instead of using the
captain's global git config. The new rule explicitly forbids overriding any
git identity field.

## Testing

- `apply_transition` tests pass with `FM_BACKEND_HERDR_BLOCKED_CONFIRM_DELAY=0`
- Shellcheck clean on the modified files
- When the re-check API call fails (no herdr server, test environment), the
  function falls back safely to the old escalation behavior (fail-safe)

### DIFF
--- a/bin/backends/herdr.sh
+++ b/bin/backends/herdr.sh
@@ -2189,6 +2189,16 @@ fm_backend_herdr_busy_state() {  # <target>
 FM_BACKEND_HERDR_SUBMIT_POLLS=${FM_BACKEND_HERDR_SUBMIT_POLLS:-6}
 FM_BACKEND_HERDR_SUBMIT_MIN_SLEEP=${FM_BACKEND_HERDR_SUBMIT_MIN_SLEEP:-0.6}
 
+# FM_BACKEND_HERDR_BLOCKED_CONFIRM_DELAY (default 3): seconds to wait before
+# re-checking a `blocked` transition. herdr can emit transient `blocked` events
+# when the OpenCode TUI shows a brief input prompt between tool calls; the
+# re-check queries the live agent status and absorbs the transition if the
+# agent has already resumed `working`. This also covers long-running commands
+# (e.g. `sleep 600`) where herdr correctly reports `working` during execution.
+# A real wedge (permission dialog, stuck prompt) persists far longer than this
+# delay, so genuine escalations are only delayed by this many seconds.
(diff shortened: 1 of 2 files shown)


kunchenguid:

PR #1300, opened 2026-07-30
fix(bin): close three merge and decision-hold papercuts

jamesstomberg:
Three small papercuts in firstmate's own merge and decision-hold tooling, found by hitting each of
them in real operation. Offered upstream for review — I have no push access, so this comes from a fork.

## 1. `fm-pr-merge` refused a PR whose task was already torn down

A merge that arrives after teardown had no guarded path at all, so the operator's only option was to
reach past the guards to a raw merge command. Adds `--torn-down`.

**It requires positive evidence the task genuinely existed** — `data/<id>/brief.md`, or a Done entry
for that id in the backlog. Absent metadata alone is not enough, because a task id that never existed
is equally absent: without the evidence check, a typo would merge whatever PR URL was passed. The
whole point of the flag is to be the guarded path, and a guarded path that merges an arbitrary PR on a
typo has defeated itself.

The name `--torn-down` is also now mentioned in the ordinary refusal message, so the path is
discoverable at the moment you need it.

## 2. The merge ledger could assert a merge that never happened

The ledger line was written *before* the merge was attempted, so a failed merge left a permanent
record claiming success. Now the line is written only after the merge succeeds, and records the actual
outcome. The ledger's only job is to be a true record, and firstmate itself reads it.

Ledger files are also created with `umask 077`.

## 3. `fm-decision-hold resolve` refused a routed task that was already Done

A decision answered after its dependent work had finished could not be closed, leaving a captain hold
open forever with nothing left to route it to.

## Verification

Full local pipeline: review, tests, docs and lint all passed. The two behavioural changes above (1 and
2) came out of that review as findings and were both fixed rather than accepted.


kunchenguid:

PR #1315, opened 2026-07-30
fix(herdr): stop Broken-pipe stderr noise on every capability probe

wameson:
## Problem
`bin/backends/herdr.sh`'s capability probe (`fm_backend_herdr_events_capable`) writes `printf: write error: Broken pipe` to stderr on essentially every watcher cycle.

## Root cause
The probe piped the (large) API schema into `grep -Fq`:
```sh
printf '%s' "$schema" | grep -Fq 'events.subscribe' || return 1
printf '%s' "$schema" | grep -Fq 'pane.agent_status_changed' || return 1
```
`grep -q` exits on the first match and closes the read end of the pipe while `printf` is still writing the schema, so `printf` gets EPIPE and prints the broken-pipe message to stderr.

## Fix
Replace each pipe-into-`grep` with an in-memory `case` glob against the variable:
```sh
case "$schema" in *events.subscribe*) ;; *) return 1 ;; esac
case "$schema" in *pane.agent_status_changed*) ;; *) return 1 ;; esac
```
No subshell, no pipe, no reader that can close early, so no EPIPE (and it is faster).

## Behavior is unchanged, not weakened
`*literal*` is a literal substring match equivalent to `grep -F`; neither token contains a glob metacharacter. The function still returns success only when BOTH `events.subscribe` and `pane.agent_status_changed` are present and `return 1` otherwise. The only other pipe-into-grep in the file is a `grep -qE` regex on a single stripped terminal line — a tiny buffer that never overflows the pipe — so it produces no such noise and is deliberately left unchanged.

## Tests
Adds three colocated regression tests in `tests/fm-backend-herdr.test.sh`: both tokens present ⇒ capable, either token missing ⇒ not capable, and the probe leaves stderr clean. The clean-stderr test runs under `trap '' PIPE` (the disposition under which the noise surfaces) against a >64KiB schema; verified to fail against the old grep-piped code and pass against the fix. Full herdr suite green.

### DIFF
--- a/bin/backends/herdr.sh
+++ b/bin/backends/herdr.sh
@@ -2341,8 +2341,13 @@ fm_backend_herdr_events_capable() {  # <session>
   case "$protocol" in ''|*[!0-9]*) return 1 ;; esac
   [ "$protocol" -ge "$FM_BACKEND_HERDR_MIN_EVENTS_PROTOCOL" ] || return 1
   schema=$(herdr api schema --json 2>/dev/null) || return 1
-  printf '%s' "$schema" | grep -Fq 'events.subscribe' || return 1
-  printf '%s' "$schema" | grep -Fq 'pane.agent_status_changed' || return 1
+  # In-memory substring tests, not `printf ... | grep -Fq`: grep -q exits on
+  # first match and closes the pipe while printf is still writing the large
+  # schema, which prints `printf: write error: Broken pipe` to stderr on every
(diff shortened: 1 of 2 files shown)


wameson:
Superseded: re-raising through no-mistakes so the PR body carries the required deterministic signature. The original push 403'd before the pr step could run, so this hand-opened PR lacks the signature line the upstream 'no-mistakes-required' check enforces. A replacement PR from the same fork branch will follow.

PR #1285, opened 2026-07-30
fix(composer): read an NBSP-padded empty composer as empty and fail unshown herdr wedge alarms

sbracewell64:
## Intent

Fix the root cause of firstmate's away-mode escalation delivery wedge: buffered captain-relevant escalations sat undelivered for ~9.5 hours in each of three separate stretches, and steers into an idle pane reported false "delivery unconfirmed" errors.

This implements recommendations R1 and R3 of a completed, already-reviewed diagnosis (private report at data/away-injection-wedge-diagnosis/report.md). The design was reviewed and motivated before this work started; the task was explicitly to implement it, not to redesign it. Do not propose an alternative architecture for the fix.

R1 (the fix). Real Claude Code 2.1.220 pads its EMPTY composer row with U+00A0 NBSP, so the captured row is exactly `❯` + \xc2\xa0. bash's [[:space:]] does not match U+00A0, so no trim could remove it, the leading-glyph strip left a lone NBSP behind, and the shared classifier concluded "real, unsubmitted content remains" -> `pending` on a genuinely idle pane, stably, on every poll. Away-mode injection and verified submit both proceed only on an affirmative `empty`, which is why delivery wedged. fm_composer_classify_content now normalizes the non-ASCII blanks a TUI can use as padding before its trims (U+00A0, U+2007, U+202F -> ASCII space; U+200B, U+FEFF dropped).

Deliberate decisions in R1, so they are not mistaken for oversights:
- The normalization lives in bin/fm-composer-lib.sh, the ONE fleet-wide owner of the empty|pending|unknown verdict, specifically so both ANSI readers (tmux capture-pane -e and herdr pane read) and all four adapters (tmux, herdr, Orca, cmux) are covered at once and cannot drift back into per-adapter copies. Per-adapter normalization was rejected for exactly that reason.
- Callers already trim ASCII whitespace before calling, and that is deliberately NOT sufficient - the whole defect is that their trims cannot see U+00A0. The classifier re-trims both content and plain_content after normalizing so the rest of the function keeps the trimmed inputs its contract comment documents.
- The safety property is load-bearing and asserted in tests: every character normalized here renders as blank and nothing else is touched, so the change can only ever make an OTHERWISE-BLANK row read as blank. It is impossible for real typed text to become `empty` - a row holding any visible glyph keeps that glyph byte for byte, NBSP-joined text stays `pending`, and an NBSP-padded dead-shell prompt stays non-empty (it becomes the documented `unknown` rather than `pending`).
[...]

PR #1322, opened 2026-07-30
fix(afk): end the away-mode wedge from an invisible composer padding

KD5694:
## Why

Away mode stopped delivering escalations overnight on 2026-07-29/30: 1562 deferrals over 6h14m against a pane that was ready for input the whole time.

Root cause, proven byte by byte in an isolated lab: claude 2.1.220 draws its idle composer as the bare agent glyph plus U+00A0 NO-BREAK SPACE (`e2 9d af c2 a0`). U+00A0 is not in `[[:space:]]` in any glibc locale, so the shared classifier's trims removed nothing, its exact-glyph case missed, and an idle composer classified as `pending` - the verdict meaning "a human typed something unsent". The injector defers on anything but `empty`, so it deferred every escalation.

Shutdown latency was a second, independent defect on the same path: the daemon's shutdown trap did an unbounded `wait` on its watcher child, and the watcher spends each cycle inside a blocking foreground wait where bash defers its TERM trap. Shutdown took about 14s and blew the 10s stop window, so a correct shutdown was reported as "daemon did not exit after SIGTERM".

## What changed

1. **`bin/fm-composer-lib.sh`** - invisible padding is normalized in the one owner of the verdict, before any decision, then re-trimmed. The fix lives here rather than in the herdr reader because the tmux path delegates to the same function and had the identical bug. Octal byte escapes, not `\u`, because parts of the fleet run under `LC_ALL=C`; the verdict is asserted identical in both locales. Every pre-existing branch keeps its semantics, including the load-bearing dead-shell refusal where a bare `>`/`$`/`%`/`#` row must stay `unknown` and never become a safe injection target.

   One behavior widening is intentional: a row containing only invisible padding was `pending` and is now `empty`. Only characters with zero typed meaning are normalized (U+00A0, U+2007, U+202F, U+200B, U+2060, U+FEFF), so this cannot swallow a half-typed line.

2. **`bin/backends/herdr.sh`** - composer scan window 20 to 40 rows. A tall custom statusline can occupy 11-12 rows below the composer; at 20 the composer row itself falls out of the window, which reads `unknown` and defers injection exactly as a false `pending` does. Wire cost is zero: the capture already fetches at least 200 rows.

3. **Bounded watcher reap plus a cross-process budget contract.** The shutdown trap now reaps the watcher within a budget and hard-kills only as a backstop. That budget is `FM_POLL + 5` with a floor of 20 so the graceful path is actually reachable - a 3s bound would have made SIGKILL the routine path, skipping the watcher's own cleanup trap and leaving a stale `state/.watch.lock` after every clean stop.

   The away-mode stop window must outlast a full graceful reap at every `FM_POLL`, not just the stock one. Two hand-tuned constants could not hold that, and re-deriving the number in the launcher could not either: the daemon is spawned with a three-variable environment allowlist into a terminal owned by the multiplexer server, so the two sides can legitimately read a different `FM_POLL`. The daemon therefore records the budget it actually resolved in `state/.supervise-daemon.reap-secs` beside its pidfile and removes it on every exit path; the stop path reads that record and never re-derives it. One producer of the number, no environment agreement required. An absent, empty, non-numeric or zero record falls back to the plain 160-poll floor - the 40s window this path used before - never to a guess.

4. **`wedge_alarm_via_herdr`** reads delivery from the response `shown` field instead of the exit code. herdr ACKs with exit 0 on a host whose toast delivery is off (`{"result":{"shown":false,"reason":"disabled"}}`), which made the alarm a silent stub exactly where the captain would see nothing. Only an explicit `shown:false` counts as failure, so an older herdr that prints nothing still counts as delivered, and the match is textual so the alarm path takes on no jq dependency.

`docs/wedge-alarm.md` states plainly that channels fire independently with no failover between them, and that an explicit `shown:false` is recorded as a failed channel rather than counted as delivered.

## Tests

`tests/fm-composer-nbsp.test.sh` guards the outage at two levels: the shared classifier directly, and the recorded night screen replayed end to end through the real herdr adapter, including a tall-statusline case that fails at a 20-row window and passes at 40. Its jq gate deliberately sits below the level-1 classifier assertions so the regression guard still runs on a host without jq.

The shared fake herdr CLI moved verbatim into `tests/herdr-helpers.sh` so a second suite can source it instead of copying it; `tests/fm-backend-herdr.test.sh` is unchanged at 126 assertions. `bin/fm-test-run.sh` gains one family-map line without which changed-file selection would never run the new test.

Added guards in `tests/fm-daemon.test.sh` and `tests/fm-afk-launch.test.sh` cover the bounded reap and hard-kill backstop, the alarm channel's `shown:false` / `shown:true` / silent responses, the recorded reap budget matching what the daemon resolved, the stop path following that record, and every malformed-record path falling back to the floor. Each was verified to fail against the pre-change behavior.

Local verification of this head: `fm-afk-launch` 55, `fm-daemon` 104, `fm-composer-nbsp` 8, `fm-backend-herdr` 126 assertions, zero failures; `bin/fm-lint.sh` clean at the pinned ShellCheck 0.11.0; `bin/fm-test-run.sh --check-coverage` ok; `bin/fm-doc-audience-check.sh` ok.

Two suite scripts fail on the development host under `LC_NUMERIC=pl_PL` (awk emits `0,1000` for `%.4f`). Both fail identically on the base commit and both pass under `LC_ALL=C` on base and head alike, so they are pre-existing and unrelated to this change.

## Known residue, deliberately not in this change

- `docs/wedge-alarm.md` still carries older "continues to the next channel" wording a few lines below the new no-failover line. It is literally true - the emit loop does not abort - but can read as failover. Rewording that paragraph goes past what this change made stale.
- The herdr adapter's structural gate that decides whether a row is a composer at all still trims with `[[:space:]]` only, so a harness padding the leading edge of the composer row would fail its anchor. The observed harness pads only after the glyph.
- Real all-channels-failed escalation for the wedge alarm was explicitly ruled out of scope; this change makes the failure visible, not escalated.


eduardstan:
Your byte-level root cause reproduced exactly on a herdr home — `e2 9d af c2 a0`, U+00A0 outside
`[[:space:]]` in every glibc locale — and the normalisation here is the fix for it. Thank you for
proving it to the byte; it saved us a lot of guessing.

**One thing worth knowing before this lands: on herdr there is a second defect in front of it, and
this fix alone will not reach it.**

Our incident was 2524 deferrals over 5h54m against a live idle claude primary, with correct target
discovery, and the verdict was `unknown` rather than `pending`. The reason it was not `pending` is
that the classifier never saw the content at all:

`fm_backend_herdr_composer_state` treats a lone trailing horizontal rule with no matching opening
rule as proof that a composer row found above it is stale. That heuristic was written for Pi, whose
composer genuinely sits between two bare separators. **claude frames its composer the same way except
its upper rule carries a label** (`--- name --`), so it is not a bare separator, the capture ends in
exactly the incomplete-pair shape, and the live composer row is discarded before normalisation runs.

So on herdr the order is: structural rejection first, NBSP second. With only this pull request
applied, a herdr home with a claude primary still reads `unknown` and still wedges.

We have both fixed on our fork and opened it as #1654, with your normalisation
credited and applied rather than reimplemented — only one of #1322 / #1285 is taken, since they are
equivalent for that defect. Happy to fold it into this pull request instead if you would rather keep
it in one place; it is your work that made the second half tractable.

KD5694:
Closing this in favour of #1668, which carries the same normalisation plus the structural defect in front of it.

Your analysis settled it: on a herdr home with a claude primary, the composer row is discarded before normalisation ever runs, so this pull request alone would not have ended the wedge on exactly the setup that reported it. Better one landed change that covers the whole path than two partial ones, and #1668 already carries the provenance the required gate accepts.

The `FM_BACKEND_HERDR_COMPOSER_LINES` bump from 20 to 40 in #1668 is worth calling out separately, because it fixed a third way into the same symptom that we had not isolated: a multi-line custom statusline pushes the composer row out of the scanned tail, which reads as `unknown` for a reason that has nothing to do with either the NBSP padding or the labelled rule. Our own home runs such a statusline, so that one would have bitten us next.

No need to fold anything back in here. Thanks for crediting the normalisation rather than reimplementing it, and for taking it through the pipeline so it can actually land.

PR #1333, opened 2026-07-30
feat(herdr): opt-in per-primary-home workspace label

msaniawa-coder:
## The problem

`fm_backend_herdr_workspace_label()` returned the hardcoded constant `firstmate` for any home without a `.fm-secondmate-home` marker. Only secondmate homes got a distinct label (`2ndmate-<id>`), and there was no way to change it.

That is fine for one primary home. It breaks as soon as two primary homes share a single Herdr session: both resolve to the same `firstmate` label, so `fm_backend_herdr_workspace_find` matches the first home's workspace and the second home's crew panes are created inside it, mixed in with the first home's panes. A workspace created by hand for the second home is never a candidate, because the lookup only ever matches that one constant.

This was hit in practice, with the second home's panes landing alongside seven panes belonging to the first home.

## The change

An opt-in, local, gitignored `config/herdr-workspace-label`. A primary home that shares a Herdr session with another primary home writes its own workspace label there.

**Absent means unchanged.** An absent, empty, or whitespace-only file resolves to `firstmate`, byte for byte what it does today. This is the point of the design, not a side effect: every existing home keeps working untouched, every already-recorded task label stays valid, and there is no migration step. A second primary home opts in explicitly rather than any home being moved for it.

**A secondmate marker still wins.** `.fm-secondmate-home` is checked first and keeps precedence over the config file, so a secondmate home still resolves to `2ndmate-<id>` regardless of what the file says. The existing per-secondmate guarantee cannot be configured away. This also keeps `fm-spawn.sh` correct: it narrowly overrides the home when a primary launches a secondmate, and the marker is what that path resolves on.

The label is read fresh on every call rather than cached at load time, so it stays stable across every respawn and recovery for the life of that home.

### Accepted value shape

Deliberately conservative, since the value is passed to the Herdr CLI:

- one line, 1 to 64 characters
- drawn from `A-Z`, `a-z`, `0-9`, `.`, `_`, `-`
- first character a letter or digit
- not the reserved `2ndmate-` prefix

Surrounding whitespace is trimmed; inner whitespace is rejected by the character set. That set also excludes control characters, plus `/` (the separator in the legacy `<owner>/<task> · p:<token>` presentation-child labels) and a leading `-` (which the CLI would read as a flag). `2ndmate-` is refused because that namespace belongs to secondmate homes, and a primary home claiming it would recreate the same cross-home collision this change removes. Explicitly configuring `firstmate` is accepted and idempotent.

**An unusable value refuses; it never falls back.** Falling back to `firstmate` would put the home straight back into the collision the file was written to remove, so resolution returns an error naming the file and the accepted shape, and emits no label at all. That refusal is propagated through every caller: `workspace_find`, `workspace_ensure`, `container_ensure`, `list_live`, and the spawn path. `container_ensure` resolves the label before starting a Herdr server, so a bad value refuses without side effects, and `list_live` refuses rather than reporting an empty workspace, since "no live tasks" from a home whose own label cannot be resolved would read as a fleet with nothing running.

### Not inherited

The item is deliberately absent from the inheritance allowlist in `bin/fm-config-inherit-lib.sh`. Propagating it would put two homes back into one workspace, which is exactly what it exists to prevent. Both the config-push path and the bootstrap inheritance pass go through that explicit allowlist, so no change was needed to either; a regression assertion in the existing propagation test now pins the exclusion.

## Adoption safety, re-verified empirically

With a custom label, workspace lookup will adopt a workspace a human created by hand. That is the same shape as the 2026-07-02 incident recorded in the comment above `fm_backend_herdr_workspace_prune_seeded_default_tab`, where an adopted workspace had its live pane closed.

The existing structural gate should already make this safe: only a workspace created by that same `fm_backend_herdr_workspace_ensure` call carries a non-empty `seeded_tab_id`, and only that exact id is ever a prune candidate. Rather than assume it still holds, this was re-verified against a real Herdr 0.7.5 (protocol 17) session:

- a pre-existing workspace carrying a configured label is adopted and reports an empty seeded-tab id
- after the task tab is created, the workspace still holds both its original tab and the new one

The gate is structural rather than label-derived, so it holds for any label, not only the derived ones. That evidence is recorded in `docs/verification/runtime-backends.md` with date, version, commands, and exact output, and is captured as a permanent regression in the prune-safety end-to-end test.

## Known limitation, stated plainly

The optional, default-off presentation-space ordering step recognizes only `firstmate` and `2ndmate-*` as top-level homes. A workspace under a configured label is therefore not recognized as a top-level home by the *other* homes on that session, so when one sits between another home's block and its new child, that placement step skips with a warning and leaves the layout in Herdr's current order.

This is presentation-only and never affects a worker: ordering is best-effort, always returns success, and never fails a spawn. A complete fix would require every home's configured label, which no single home can know, so this is documented in `docs/herdr-backend.md` rather than worked around.

## Tests

- 13 new cases in `tests/fm-backend-herdr.test.sh`, all driving the public function surface rather than asserting source text: absent/empty/whitespace-only stays `firstmate`, a valid value wins, trimming, two primary homes not colliding, the secondmate marker beating a configured label, every rejected shape refusing loudly without falling back, and the container path creating, adopting, and refusing under a configured label
- a configured-label adoption section in `tests/fm-backend-herdr-prune-safety-e2e.test.sh`
- a second-primary-home section in `tests/fm-backend-herdr-workspace-per-home-e2e.test.sh` that drives the real spawn path end to end and asserts the crew lands in its own workspace, that recovery discovery stays scoped to it, and that cleanup closes only its own task

Three real-Herdr suites now point at a scratch config directory, because running them from a home that adopts this feature would otherwise change the very label they assert.

## Verification

- `bin/fm-lint.sh` clean (pinned ShellCheck 0.11.0)
- `bin/fm-doc-audience-check.sh` ok
- `bin/fm-test-run.sh --check-coverage` ok
- portable parallel shards 1 and 2: no failures
- real-Herdr gated family: 10 scripts, no failures, including the full presentation end-to-end suite

Two failures in the serial lane, `tests/fm-calm-pi-extension.test.sh` and `tests/fm-secondmate-harness.test.sh`, were confirmed pre-existing by stashing this branch and re-running on a clean tree. Neither touches the Herdr adapter.

## Documentation

Following the one-owner rule: `docs/configuration.md` owns the schema and per-field semantics, `docs/herdr-backend.md` owns the behavior description and cross-references it, `AGENTS.md` gets one line in its layout listing, and `docs/verification/runtime-backends.md` holds the empirical record.


msaniawa-coder:
Withdrawing this contribution at the repository owner's request on our side. Apologies for the noise.

PR #1330, opened 2026-07-30
feat(bin): route crew dispatch by subscription capacity

adibirzu:
## What Changed
- Added `bin/fm-dispatch-select.mjs` to resolve already-matched dispatch profile arrays using fresh quota telemetry, reserve thresholds, provider cooldowns, and persisted least-recent rotation.
- Threaded subscription provider identity through bootstrap validation and `fm-spawn.sh`, including native harness/provider consistency checks, provider metadata, and explicit rejection of unsupported Kimi subscription dispatch.
- Updated crew-dispatch docs, examples, agent skills, and shell coverage for subscription routing settings, non-native provider requirements, selector behavior, and spawn metadata.

## Risk Assessment

✅ Low: The reviewed changes are bounded to subscription-aware dispatch routing, and the follow-up bootstrap/schema tightening now aligns with the selector contract without exposing a material source-verifiable defect.

## Testing

Baseline inspection found a clean worktree and required tools; targeted selector, spawn-profile, and bootstrap dispatch-validation tests passed; manual CLI evidence shows healthy provider rotation, reserve and cooldown failover, and `fm-spawn` persisting `provider=codex` while launching Codex with model/effort flags only.

<details>
<summary>Evidence: Subscription routing CLI transcript</summary>

```text
Subscription-aware dispatch routing CLI evidence
Generated in /tmp/no-mistakes-evidence/01KYT7407HMRKKPKWT0N0JF689/subscription-routing-cli

$ FM_HOME=/tmp/no-mistakes-evidence/01KYT7407HMRKKPKWT0N0JF689/subscription-routing-cli/home FM_STATE_OVERRIDE=/tmp/no-mistakes-evidence/01KYT7407HMRKKPKWT0N0JF689/subscription-routing-cli/home/state FM_CONFIG_OVERRIDE=/tmp/no-mistakes-evidence/01KYT7407HMRKKPKWT0N0JF689/subscription-routing-cli/home/config bin/fm-dispatch-select.mjs select --quota-json /tmp/no-mistakes-evidence/01KYT7407HMRKKPKWT0N0JF689/subscription-routing-cli/quota-healthy.json --now 1000 [{"harness":"claude","provider":"claude","model":"sonnet","effort":"high"},{"harness":"codex","provider":"codex","model":"gpt-5","effort":"high"}]
exit: 0
stdout:
{"harness":"claude","provider":"claude","model":"sonnet","effort":"high"}
stderr:
[...]

PR #1217, opened 2026-07-28
feat(bin): federated fleet add-on with multi-account spawn and quota-aware routing

adibirzu:
## Intent

Two follow-ups a closing review flagged on the fleet quota work (PR #1187). First: tests/federation/test_fleet_ops.sh consulted the live quota-axi in its register/route lifecycle cases, so it went red whenever the operator's real subscription window was exhausted (observed 4/16 failing, all green after the window reset) — now hermetic via a stubbed quota-axi, with lifecycle semantics still genuinely exercised and one bounded case deliberately retained on the real quota_now path so live coverage is not silently deleted. Second: bin/fm-fleet-lib.sh had grown to 816 lines, past the repo's 800-line convention — split at a verified one-directional seam (surface->KB crossings: zero) into a new leaf bin/fm-fleet-quota-lib.sh, leaving 393 and 460 lines. Behaviour-preserving: proven by the committed golden fixtures being compared, never regenerated.

## What Changed

- **Federated multi-operator fleet** — new `bin/fm-fleet.sh` CLI over `bin/fm-fleet-lib.sh`, coordinating several OS operators through a shared, cross-uid-safe, git-backed KB: `init`/`register`/`heartbeat`/`leave`, `queue`/`claim`/`handoff`/`reap` under a lock, scope routing with heartbeat-freshness staleness (with a non-GNU `date` fallback), and `status`/`view`, plus `fm-fleet-wait.sh` and `fm-fleet-join.sh`. Documented in `docs/fleet-addon.md`, `docs/fleet-quickstart.md`, `docs/fleet-token-economy.md`, and the `federation`/`multi-account` agent skills.
- **Per-spawn `--account` axis** — account registry, resolution and validation across three isolation methods (`bin/fm-accounts-lib.sh`, `fm-account-env.sh`, `fm-account-exec.sh`, `fm-spawn-acct.sh`), riding `fm-spawn`'s existing raw-launch escape hatch so no core FirstMate script is modified and no secret lands on argv. Adds a user-scoped on-demand prereq installer (`fm-accounts-prereq.sh`) and one reviewable idempotent root step (`scripts/fleet-root-prereq.sh`).
[...]

PR #1142, opened 2026-07-27
feat(bin): give each project its own leased worktree pool on the repo's filesystem

adibirzu:
## Intent

PR #1142 (per-project worktree pools): fix the two Behavior-tests-Herdr CI failures — post-create abort fixture updated for leased acquisition, and the projected-teardown focus regression root-caused (treehouse return --force killed the pane's shell before the presentation restore block) and fixed by closing the presentation pane before returning the worktree; 22/22 presentation e2e x3 + 132/132 backend suite green locally. Goal: green Herdr CI lane + the Require-no-mistakes signature on PR #1142.

## What Changed

- Spawn now acquires crew worktrees through `treehouse get --lease` from a per-project pool placed on the repo's own filesystem. The new `bin/fm-treehouse-lib.sh` selects the pool by running treehouse under a per-command `HOME` and verifies the acquired worktree is colocated with the object store, preventing the cross-filesystem split that made `no-mistakes init`/`axi run` hang inside git's gitfile open.
- Teardown and the spawn-abort cleanup path now close the Herdr presentation pane before returning the leased worktree (previously `treehouse return --force` killed the pane's shell first, letting Herdr's last-pane cleanup destroy the workspace and steal focus), and lease failures now surface treehouse's stderr instead of a generic error. Watcher lock checks also resolve identity paths before comparing.
- Tests and docs follow the new flow: leased-acquisition and post-create abort fixtures, treehouse stubs in every fakebin, new watcher-lock and pool-settle coverage (11/11 settle, 22/22 Herdr presentation e2e across three runs), plus pool/spawn/teardown doc alignment and recorded verification evidence.

## Risk Assessment

✅ Low: All three accepted round-1 findings are fixed correctly in 94148ff (abort-path pane-close-before-return ordering, teardown abort instead of focus-unsafe return under lock contention, lease stderr surfaced with proper temp-file hygiene), no test pins the replaced behaviors, and the only remaining item is an informational version-skew note that CI will confirm.

## Testing

[...]

PR #1104, opened 2026-07-27
feat(harness): add verified cline and cursor-agent crewmate adapters

adibirzu:
## Intent

Crewmate harness adapters for cline and cursor-agent (PR #1104), rebased onto upstream's pi-signed adapter: empirically verified launch templates, busy/idle signatures, interrupt/exit semantics, and backend idle-regex extensions, with 12-check suites per adapter. Goal: ship through the sanctioned pipeline so PR #1104 carries the no-mistakes signature and passes the Require-no-mistakes gate.

## What Changed
- Added crewmate adapters for Cline CLI 3.0.46 and Cursor CLI 2026.07 with empirically verified launch templates, busy/idle signatures, and interrupt/exit semantics: harness detection in `bin/fm-harness.sh`, launch templates and adapter allowlists in `bin/fm-spawn.sh` (including Cursor's workspace-trust gate predicates), and idle-regex extensions across the herdr, cmux, and orca backends.
- Consolidated composer idle-placeholder and agent-glyph defaults into shared definitions in `bin/fm-composer-lib.sh`/`bin/fm-tmux-lib.sh` so all backends recognize the new adapters' idle composers (e.g. cline's "Ask anything..." and cursor's "→ Plan, search, build anything") instead of each backend carrying its own copies.
- Added per-adapter behavior test suites (`tests/fm-cline-harness.test.sh` 12 checks, `tests/fm-cursor-agent-harness.test.sh` 14 checks) plus verification evidence docs under `docs/verification/`, and updated the harness-adapters SKILL.md and backend/configuration docs to register the new adapters.

## Risk Assessment

✅ Low: The fix commit resolves all three round-1 warnings at the recommended shared boundary (fleet-wide idle/glyph defaults consumed by all four adapters, shared busy-default extension, and a kimi-pattern trust-gate readiness step that fails loudly), preserves the dead-shell safety rule with new test coverage, and leaves only informational nits and the honestly documented live end-to-end acceptance step.

## Testing

[...]

PR #1338, opened 2026-07-30
feat: consolidate federated fleet and crew routing work

adibirzu:
## Intent

Use #1338 as the single consolidation vehicle for the captain-approved upstream PR cleanup.
This replaces the overlapping federated fleet, quota routing, adapter, pool relocation, capacity guard, and runtime handoff PRs with one branch based on current upstream `main`.

## Consolidated Scope

- Federated fleet add-on with multi-operator routing, shared fleet state, cross-operator handoff, and quota-aware account dispatch.
- Multi-account spawn support with account registry resolution, quota surface selection, and the prior security fixes for split fork-sync permissions, fleet-directory ownership, and stale handoff references.
- Subscription-capacity crew routing deduped into one implementation.
- Verified harness coverage for `agy`, `cline`, `cursor`, `copilot`, and related local adapter paths.
- Machine-capacity guard that refuses spawn when the machine has no memory headroom.
- Per-project leased worktree pool relocation and freshen behavior.
- Runtime handoff on quota, including adapter coverage for `agy`, `cline`, `cursor`, `copilot`, and `muse`.
- No-mistakes review fixes for account handoff quoting, home-boundary guards, muse handoff support, and lint/test routing.

## Supersedes

- #1217: older federated fleet lineage is carried forward here through the #1338 reraised branch.
- #1559: machine-capacity guard plus agy and sync-CI work are folded into this branch.
- #1473: agy adapter verification is folded into this branch.
- #1330: original subscription routing branch is deduped into this branch.
- #1461: subscription-aware routing successor is folded into this branch.
- #1104: local `cline` and `cursor-agent` adapters are folded into this branch.
- #1142: per-project leased worktree pool relocation is folded into this branch.
- #1468: runtime handoff on quota is folded into this branch with adapter coverage completed.
- #1815: prior consolidation attempt is superseded by this clean #1338 vehicle.

## Branch State

- Base: `origin/main` at `76355e20b4f44d968ca43c14e1bb21c100ac90d7`.
- Fork main synced by non-force merge to `5446bf33cf134223d02285931f2682f74d9ab3e2`.
- Head: `b0fa72ced5b57c073dfcf9859f1643a0dea30c8e`.
- Push target: `adibirzu/firstmate:fed-lib-split-reraised`.

## Validation

- `bin/fm-lint.sh` passed on the consolidated branch.
- `bin/fm-test-run.sh --changed --base origin/main` passed as the accepted repo merge-truth gate: 161 scripts, 0 failed.
- `no-mistakes` review fixes were applied and recovered into the branch.
- `no-mistakes` final green attestation is blocked by infrastructure, not by code: the shared `/tmp/no-mistakes-evidence` directory is owned by another uid (`barf-ai`), so evidence creation fails with permission denied.
- The no-mistakes blocker needs an upstream per-user evidence-root fix.


kunchenguid:

<sub>Noted for firstmate#1338 at `aca5fc6c`.</sub>

kunchenguid:

<sub>Noted for firstmate#1338 at `a2b1ce0c`.</sub>

adibirzu:
Pushed two test-only commits (`746c703`, `d52c8ec`) after running the full 120-suite on **macOS / BSD userland**. No production code touched; Linux behaviour is unchanged in every case.

The run surfaced two distinct classes of problem, and in both the tests — not the code under test — were at fault.

### 1. The fleet suites assumed GNU coreutils (`746c703`)

The fleet *production* code is already BSD-first — `bin/fm-fleet-lib.sh:46` tries `date -u -j -f` before `date -u -d`, and both `bin/fm-fleet-preflight.sh` and `scripts/fleet-root-prereq.sh` fall back from `getent` to `/etc/group`. The tests never got the same treatment.

The serious part is that two of the three were not merely failing, they were **silently vacuous**. Nine in-place edits used the GNU form `sed -i EXPR FILE`; BSD sed requires a backup-suffix argument, so it read `EXPR` as the suffix and `FILE` as the script, errored to stderr, and **left the file untouched**. Each of those edits is what *sets up* its scenario — ageing a claim stamp, flipping a status to `in-flight`, marking an operator offline, forcing a stale heartbeat — so the assertions that followed graded the unmutated state. The stale-heartbeat, quota-ceiling, reap and routing cases were not testing what they claim on macOS.

`tests/fm-admiral-optin.test.sh` had a different defect: `set +e` / `set -e` around a probe, in a file that runs under `set -u` **only**. The trailing `set -e` did not *restore* errexit, it *enabled* it for the first time; the next line then legitimately returned non-zero on any host without the privileged Tier C setup, so the shell exited 1 silently and the sixth test never ran. Every assertion printed `ok` while the file exited 1, which is why it read as a harness quirk.

### 2. Fixture helpers let hook output into their return value (`d52c8ec`)

Five suites failed for one shared reason. Their setup helpers **return a directory on stdout** — `w=$(new_world ...)`, `case_dir=$(make_case ...)`, `GATE_WT=$(make_gate_worktree ...)` — while also running `git push` to seed a fixture remote. `git push` runs the caller's pre-push hook, and a hook that logs informationally to **stdout** rather than stderr has its banner captured by the command substitution. The captured value becomes `"<banner>\n/real/path"`, so every derived path is corrupt. `git push -q` silences git itself and has no effect on hook output.

None of the symptoms pointed at the cause:

| suite | reported failure |
|---|---|
| `fm-update` | `firstmate fast-forwarded (missing: 'firstmate: updated ')` |
| `fm-teardown` | `fork-allow: ... expected exit 0, got 1` |
| `fm-gate-refuse` | `must exit 3 even with the marker unset: ... got 111` — 111 is `run_guard_lib`'s own `cd "$GATE_WT" \|\| exit 111` |
| `fm-review-diff` | `diff should show the PR head content (missing: '+pr-fixed')` |

Fix is hook-agnostic: group each helper's setup and send its stdout to `/dev/null`, so the only thing on stdout is the value the helper contracts to return. That is a property these helpers arguably should have had regardless — a function that returns a value on stdout must own stdout.

### Result

| suite | before | after |
|---|---|---|
| `federation/test_fleet.sh` | 15 pass / 4 fail | **19 / 0** |
| `federation/test_fleet_ops.sh` | 23 pass / 5 fail | **28 / 0** |
| `fm-admiral-optin` | 5 of 6 tests run, exit 1 | **6 of 6, exit 0** |
| `fm-gate-refuse` | 2 pass / 1 fail | **7 / 0** |
| `fm-review-diff` | 0 pass / 1 fail | **5 / 0** |
| `fm-update` | 0 pass / 1 fail | **9 / 0** |
| `fm-teardown` | 0 pass / 1 fail | 22 / 1 *(remaining `squash-replayed-patch` case is unrelated)* |

Full suite on macOS: **104 → 110 of 120**. `bin/fm-test-run.sh --check-coverage` and `bin/fm-lint.sh` both pass.

### Notes for maintainers

- The federation suites also need **`flock`** (util-linux, absent from macOS). Not worked around — `bin/fm-fleet-preflight.sh` already reports it correctly (*"Tier A is blocked. Install: flock"*) and `brew install flock` provides the portable discoteq build. That is a declared prerequisite, not a portability defect.
- The remaining 10 macOS failures are untouched here and look like genuinely separate issues (herdr/Pi e2e environment, `fm-berth` live-session locking, `fm-fleet-sync` stale-lock removal, `fm-session-start` recovery digest, `fm-gotmp`, `fm-backend`). Happy to take those in a follow-up if useful.

adibirzu:
Follow-up to the previous comment. Full suite on **macOS / BSD userland** is now **119 / 120**, up from 104. Four more test-only commits plus one small product change; every fix below is verified by running the affected suite directly.

### Fixture rot: derived closures instead of hand-maintained lists (`6333e76`)

`fm-gotmp` and `fm-backend` each build a fake `bin/` for `fm-teardown.sh` from a hand-listed set of sibling libs. Both had fallen behind `bin/fm-treehouse-lib.sh`, which teardown sources at line 112 — so teardown died on a missing `source` immediately. Neither failure said so: `fm-gotmp` runs teardown under `>/dev/null 2>&1` and reports only *"teardown exited non-zero with a valid tasktmp"*. Both fixtures now derive the closure from the actual `. "$SCRIPT_DIR/<lib>.sh"` lines, transitively; the explicit lists stay as documentation.

⚠️ **This also closed a way for a test to corrupt the repo.** `fm-gotmp` symlinks real libs into its fake bin and then stubs some of them with `cat > "$fake/bin/<lib>"` — and `cat >` onto a symlink writes *through* to the real file. While adding the closure pass it truncated the tracked `bin/fm-tasks-axi-lib.sh` from 76 lines to a 1-line stub, which then broke unrelated suites (`fm-decision-hold.sh`, `fm-backlog-handoff.sh` and `fm-bootstrap.sh` all call `fm_tasks_axi_compatible` from it). Two guards now: the closure pass **copies** rather than symlinks, and every stub write is preceded by `rm -f`. The hazard pre-dates this change — any lib added to the symlink list and later stubbed would have done the same.

### Assertions pinned to things that are explicitly not stable (`e34d1df`)

- `fm-session-start` asserted the exact header `--- compact-startup ---`, but `fm-session-start.sh` prints `--- <id> (<crew-name>) ---` whenever `fm-name.sh` resolves a name, falling back to the bare form only when it *fails*. The assertion passed only on hosts where naming does not work. Now asserts the id.
- `fm-calm-pi-extension` probed the version with `pi --version 2>/dev/null`. **`pi` prints its version to stderr** and exits 0, so the redirect discarded the only output and the suite failed with *"could not determine the installed Pi version"* against a working Pi.

### macOS code signing killed a test's fake process (`3a00c7e`)

`fm-berth`'s `start_fake_harness` copied the system `sleep` to a file named `claude` so the holder check (`ps -o comm=` matched against `FM_HARNESS_RE`) would recognise it. Copying an Apple-signed binary invalidates its signature and macOS **SIGKILLs** the copy on start. The fake holder was dead before the berth looked, the berth correctly reported a free lock, and the suite blamed the berth for *"a berth held by a live session must refuse a second"*. `codesign -s -` does not rescue it and a shell wrapper would report the interpreter as `comm`; a **symlink** execs the real binary while `comm` still resolves to the link name, on BSD and Linux alike. 5 ok → 11 ok (six assertions had never run).

### The one product change (`56fdced`)

The herdr presentation lock **serializes** concurrent teardowns — it retries 50 × 0.1s, i.e. waits **5s** before refusing. `fm-backend-herdr-presentation-e2e` races two teardowns against one session and asserts *both* complete; its pass line is literally "concurrent projected cleanup is serialized". On macOS one of the pair reliably lost, and **which one alternated between fixture A and fixture B run to run** — that alternation is what identified a timing bound rather than an ordering defect.

The default is **unchanged at 50**; it is now overridable via `FM_HERDR_PRESENTATION_LOCK_RETRIES`, the same idiom `fm-teardown.sh` already uses for `FM_TREEHOUSE_RETURN_LOCK_RETRIES` and `FM_STALE_WORKTREE_LOCK_AGE_SECS` a few hundred lines above. Raising a ceiling only makes a waiter wait longer before giving up; it cannot turn a refusal into a wrong mutation. The suite sets 600 so it measures serialization rather than machine speed. 9 ok → **22 ok**.

### Prerequisites, not defects

Three failures were purely environment and are **not** patched here:

| tool | had | needed | evidence |
|---|---|---|---|
| `flock` | absent | any | util-linux; `fm-fleet-preflight.sh` already says *"Tier A is blocked. Install: flock"* |
| `herdr` | 0.7.1 | **0.7.4** | `bin/fm-install-herdr.sh` pins it (protocol 16) |
| `pi` | 0.78.0 | ≥ 0.81 | the calm launch line passes `--approve`, which 0.78 does not have |

With the CI-pinned herdr, `fm-backend-herdr-focus-flash-e2e` and `fm-herdr-session-cleanup-e2e` pass **unmodified** — they were never broken. Worth noting that the Pi `--approve` floor is not documented anywhere; `docs/calm-mode-feasibility.md` records a verified range but not the flag requirement.

### Still failing: 1

`fm-calm-pi-extension` — *"/export did not complete while calm mode was on"* (4 ok / 1 fail; was 2 ok / 1 fail). The live-Pi pane shows the composer still holding the typed text and Pi's own warning *"tmux extended-keys is off. Modified Enter keys may not work."* The suite submits with **`M-s`**, a modified key, and never enables `extended-keys` on its tmux socket — a plausible cause. I tried enabling it three ways (`set-option` after `new-session`, `start-server` first, and a `-f` config read at server start) and **could not demonstrate the warning clearing**, so I reverted rather than ship an unproven fix. Flagging it for whoever owns this suite: the wait bound there is also only 120 × 0.05s = 6s for a live TUI round-trip.

`bin/fm-lint.sh` and `bin/fm-test-run.sh --check-coverage` pass throughout.

adibirzu:
## Sync + validation update

**Merge scope.** This branch now carries the federated trunk reconciled against upstream `main`, and it is current with upstream at the time of writing (`0` commits behind). Conflicts were resolved by keeping upstream's structure and re-applying the federated delta onto it, so no remote-only commits were dropped.

**Review fixes.** The no-mistakes review raised 13 findings; 11 were fixed on this branch (2 remaining are informational no-ops). The substantive ones:

- The machine-capacity guard matched process trees against only 7 of 12 verified adapters, so a machine full of the other 5 reported a near-zero fleet footprint and never tripped its memory limit.
- An account helper placed an API key on `env`'s argv, readable from `/proc/<pid>/cmdline`; it now exports into the child instead.
- A shared composer pattern regressed from an alternation to a bracket expression, which matches *bytes* under `LC_ALL=C`, so unrelated box-drawing characters read as an empty composer.
- The shared fleet knowledge base no longer holds a git repository at all, and its directory permissions no longer allow one operator to plant hooks that execute as another.
- `fork-sync.yml` was split so freshly fetched upstream code never executes while a write-scoped credential is reachable: a read-only `validate` job merges and runs the validators, and a gated `publish` job only moves refs.
- Instructions referencing a recovery script that exists in no commit on any ref were removed rather than reworded.

**Gates run locally.** `bin/fm-lint.sh` green across the full canonical set; `bin/fm-test-run.sh --check-coverage` green; both portable parallel lanes and all four portable serial shards green; `bin/fm-doc-audience-check.sh` green. One test failed only under heavy parallel load and passed in isolation and on a clean re-run of its shard. The Herdr lane was deliberately not run locally and is left to CI.

**New CI.** `merge-check.yml` adds only an upstream-merge probe, since `ci.yml` already owns lint and the complete behavior-test partition on every pull request. `fork-sync.yml` keeps the fork current with upstream daily, pushing only a validated clean merge and parking a conflicting one on a `sync/upstream` pull request instead.

**no-mistakes attestation: blocked on infrastructure, not on this code.** The pipeline completes intent, rebase and review, then fails in its test step at `mkdir /tmp/no-mistakes-evidence/<run-id>: permission denied`. That path is compiled into no-mistakes v1.41.2 and is a fixed shared `/tmp` location; on this machine it is owned by a different operator, so it collides across users. There is no config or environment override for the evidence root in that version, the shared daemon is serving other lanes and must not be restarted, and storing evidence in-repo is disallowed by this repository's `CONTRIBUTING.md`. The durable fix belongs upstream in no-mistakes as a per-user evidence root.

**Related.** See the supersession note in the description regarding #1559 and #1473.

PR #1343, opened 2026-07-30
fix(watcher): deliver actionable exits to attached arms

### LINKED ISSUE #863: fm-watch-arm: attached cycles report 'FAILED - cycle ended without an actionable reason' when the starter arm consumed the actionable reason
## Summary

When `bin/fm-watch-arm.sh` **attaches** to an already-running watcher cycle (started by another arm — in our case the turn-end guard's backstop arm), and that cycle later exits with a legitimate actionable reason, the attached arm cannot see the reason. It prints:

```
watcher: attached pid=<pid> (beacon Ns)
watcher: FAILED - cycle ended without an actionable reason
```

and exits 1. The harness surfaces this as a *failed* background task, so every ordinary wake delivered through an attached arm masquerades as a watcher failure. On an active fleet (5+ crews appending status) this produced a supervisor-facing "failure" every few minutes for hours, while the underlying supervision was actually working.

## Environment

- firstmate @ 673b6ad (after 59ece45..673b6ad fast-forward, 2026-07-22)
- backend: herdr 0.7.4 (`config/backend=herdr`), macOS (Darwin 25.5.0)
- primary harness: claude (background-notify supervision per docs/supervision-protocols/)

## Evidence

`state/.watch-cycle-exits.log` shows the pattern repeatedly (fields trimmed):

```
arm_pid=51641  watcher_pid=51653  origin=started   exit_code=0       reason=actionable-stale
arm_pid=70227  watcher_pid=51653  origin=attached  exit_code=unknown reason=attached-cycle-ended
...
origin=started   exit_code=0       reason=actionable-signal
origin=attached  exit_code=unknown reason=attached-cycle-ended   <- reported as FAILED to the harness
```

The starter arm records the real reason (`actionable-signal` / `actionable-stale`) in the ledger; the attached arm records `attached-cycle-ended` with `exit_code=unknown` and then reports the no-actionable-reason FAILURE (`bin/fm-watch-arm.sh:251`).

Sequence that triggers it constantly:

1. The primary's turn ends with tasks in flight; the turn-end guard's backstop arms a watcher (untracked by the harness session).
2. The interactive session re-arms per protocol; its arm **attaches** to that healthy cycle.
3. A crew appends a status line -> the cycle exits 0 with `actionable-signal`.
4. The starter arm (backstop, detached) consumes the reason; nobody is listening to it.
5. The attached (tracked) arm sees only "cycle ended", finds no reason, and reports FAILED — the session gets a failure notification instead of a wake reason, and `bin/fm-wake-drain.sh` is frequently empty by the time it runs.

## Expected

An attached arm whose cycle ends should recover the ended cycle's recorded exit reason and re-emit it as an ordinary wake (exit 0 with the reason line), reserving FAILED for cycles that genuinely died without a reason.

## Suggested fix

`state/.watch-cycle-exits.log` already contains what's needed: on `attached-cycle-ended`, look up the ledger row for the same `watcher_pid` with `origin=started` and a recorded `reason=actionable-*`, and exit with that reason (`stale:`/`signal:`/`check:`/`heartbeat` line) instead of the FAILED line. Alternatively (or additionally), have the attach path follow the identity-matched successor chain through an actionable exit the same way the started path does, so tracked and untracked arms never race for the reason.

## Minor related note

Under the herdr backend, `fm_backend_herdr_events_capable` sprays `printf: write error: Broken pipe` to stderr (bin/backends/herdr.sh:1802-1803, and 1960-1961 in the current revision) because `printf '%s' "$schema" | grep -Fq ...` lets grep exit on first match while printf is still writing ~220KB. Harmless but noisy inside arm output; `grep -Fq ... <<<"$schema"` or ignoring SIGPIPE there would silence it. Happy to split this into its own issue if preferred.

palinopr:
## Defect

Closes #863.

When one `fm-watch-arm.sh` process owned a watcher and another arm attached to that same healthy cycle, only the owner could read the watcher's actionable output.
The attached arm saw the lock disappear, reported `watcher: FAILED - cycle ended without an actionable reason`, and exited 1 even though the watcher had delivered a legitimate wake.
That turns an ordinary wake into a failed background task and can lead the supervisor to dismiss the notification.

## Reproduction

The regression test starts two real arm processes against one real watcher.
Appending a captain-relevant status makes the watcher enqueue and print a `signal:` reason.
Before this change, the owning arm returned that reason with status 0 while the attached arm printed the typed failure and returned status 1.

## Fix

`state/.watch-cycle-exits.log` is not a sound correctness handoff for the attached arm.
Its writes are intentionally best-effort, and the owning arm cannot append its row until after the watcher has exited and released the singleton lock.
It also records a reason classification rather than the exact reason text.

The watcher now publishes an atomic `state/.watch-cycle-result` record before its EXIT trap releases `state/.watch.lock`.
The record binds the exact reason to the watcher PID and full process identity.
An attached arm reads that exact-instance result before following a replacement or declaring the close unexplained, surfaces the same reason as the owner, and exits 0.
A missing, malformed, or identity-mismatched result remains a nonzero failure.
The existing cycle-exit ledger remains bounded diagnostic evidence rather than a delivery dependency.

## Verification

The positive regression deliberately holds the lifecycle-ledger lock so neither arm can depend on a post-exit ledger row.
The attached arm still surfaces the owner's exact `signal:` reason and exits 0.
The negative control kills the watcher without an actionable result and confirms the attached arm still prints `watcher: FAILED - cycle ended without an actionable reason` and exits nonzero.
A reason matrix covers `signal:`, `stale:`, `check:`, and `heartbeat`.

```text
bin/fm-test-run.sh --family watcher-wake-lock
FM_TEST_SUMMARY total=10 failed=0 skipped_gate=0

bin/fm-lint.sh
fm-lint.sh: ShellCheck 0.11.0 (pinned 0.11.0)

/bin/bash -n bin/fm-watch-cycle-lib.sh bin/fm-push-transition-lib.sh bin/fm-watch.sh bin/fm-watch-arm.sh tests/fm-watcher-lock.test.sh

bin/fm-doc-audience-check.sh
fm-doc-audience-check: ok surfaces=57 local_links=163
```


kunchenguid:

PR #1346, opened 2026-07-31
fix(scripts): refuse CLAUDE.md promotion when a repo guards it by literal path

wongsuwarn:
## Summary
- `fm-ensure-agents-md.sh` promotes a bare `CLAUDE.md` into `AGENTS.md` (with a symlink left behind) whenever a project has only `CLAUDE.md` and no `AGENTS.md` yet.
- If that project's own tracked automation hardcodes a `--disallowedTools` entry naming `CLAUDE.md` by literal path (to stop unattended agent turns from editing their own "constitution"), promotion silently defeats that protection: the real content moves to `AGENTS.md`, but the guard keeps matching only the now-empty `CLAUDE.md` symlink.
- Found the hard way: caught and reverted by hand in a project after this exact regression happened.
- This adds a narrow, scoped detector (`has_disallowed_tools_guard_referencing_claude`) that refuses promotion when a git-tracked file mentions both `disallowedTools` and `CLAUDE.md` together, with an actionable error explaining the fix (point the guard at `AGENTS.md` instead, or promote manually).
- Deliberately narrow: this is a textual heuristic for the one confirmed pattern, not general guard-mechanism detection - a guard built from a variable, a separate config file, or a non-CLI enforcement path isn't caught. The header comment says so explicitly.

## Test plan
- [x] `bin/fm-lint.sh bin/fm-ensure-agents-md.sh tests/fm-ensure-agents-md.test.sh` (ShellCheck 0.11.0, pinned)
- [x] `bash tests/fm-ensure-agents-md.test.sh` - 11/11 passing, including two new tests: one confirming refusal when a `--disallowedTools` guard names `CLAUDE.md`, one confirming an ordinary README/CONTRIBUTING mention of `CLAUDE.md` alone does not false-trip the heuristic

### DIFF
--- a/bin/fm-ensure-agents-md.sh
+++ b/bin/fm-ensure-agents-md.sh
@@ -10,6 +10,13 @@
 # Refuses a case-variant real memory file such as a lowercase agents.md, whose
 # CLAUDE.md symlink would carry an uppercase literal target that dangles on a
 # case-sensitive filesystem (issue #389).
+# Refuses to promote CLAUDE.md when the repo's own tracked automation guards
+# it by literal path (a --disallowedTools entry naming CLAUDE.md) - promoting
+# would move the real content to AGENTS.md while that guard keeps matching
+# only the now-empty CLAUDE.md symlink, silently defeating the protection.
+# This is a narrow textual heuristic for that one known shape, not general
+# detection of every possible guard mechanism; a guard built from a variable,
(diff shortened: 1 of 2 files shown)


wongsuwarn:
Opened in error against kunchenguid/firstmate. This was meant for our own firstmate (wongsuwarn/firstmate), where the change already landed. Not an upstream contribution. Closing; can be reopened if needed.

PR #1256, opened 2026-07-29
fix(bin): normalize relative durable paths

sparkus:
## Intent

Fix the relative-FM_HOME fault that can bake an unresolvable status path into a durable secondmate charter and silently hide replies. Reproduce relative FM_HOME, FM_STATE_OVERRIDE, and FM_DATA_OVERRIDE inputs; normalize relative directory inputs to absolute paths before deriving durable or cross-process paths; fail loudly with the offending variable named when an input cannot be resolved; preserve existing absolute-path output byte-for-byte and preserve charter wording, contracts, and semantics. Survey bin/ and fix only equivalent durable-path or cross-process exposures, leaving transient relative-home consumers unchanged. Preserve safety boundaries, do not rewrite live charters or live data/state, keep the diff proportionate, add regression coverage across affected scripts, and ship a PR without merging.

## What Changed

- Normalize relative `FM_HOME`, `FM_STATE_OVERRIDE`, and `FM_DATA_OVERRIDE` values before generating charters or constructing durable and cross-process paths, while preserving absolute path spellings and naming invalid inputs in errors.
- Embed an absolute home in generated X-mode poll shims without changing transient relative-home consumers.
- Document the resolution behavior and add regression coverage across brief generation, spawning, AFK launch, and X-mode bootstrap.

## Risk Assessment

✅ Low: The change is well-bounded and now closes the reviewed relative FM_HOME, FM_STATE_OVERRIDE, and FM_DATA_OVERRIDE durable/cross-process paths while preserving already-absolute path spelling and existing charter semantics.

## Testing

Inspected the exact base-to-target change, ran the four focused regression suites, and manually exercised real secondmate charter generation. Relative home/state/data inputs produced an absolute durable reply path byte-identical to equivalent physical absolute inputs; absolute spelling was preserved; every unresolvable input failed loudly naming its variable. Reviewer-visible CLI and charter evidence was captured. No screenshot applies because this is a shell-only change.

<details>
<summary>Evidence: Relative status-path CLI transcript</summary>

```text
$ fm-brief.sh with physical absolute FM_HOME/FM_STATE_OVERRIDE/FM_DATA_OVERRIDE
scaffolded: /private/var/folders/bd/ky17fvwx45353ky4s2bg8ys80000gn/T/no-mistakes-evidence/01KYQDJKBR0VDEEQR0Z626A7E3/demo2-data/route-proof-2/brief.md (secondmate charter)
$ same command from evidence root with relative directory inputs
[...]

PR #1355, opened 2026-07-31
fix(watch): judge watcher liveness in awake time across macOS system sleep

### LINKED ISSUE #1280: bin: basename "illegal option -- z" noise on macOS - login-shell ancestor "-zsh" parsed as a flag in the harness ancestry walk
## Symptom

On macOS (BSD basename), every session start and every Claude Stop-hook cycle prints:

```
basename: illegal option -- z
usage: basename string [suffix]
       basename [-a] [-s suffix] string [...]
```

It appears in the LOCK section of `bin/fm-session-start.sh` output and in Stop-hook feedback, so it fires on effectively every supervision turn.

## Root cause

`fm_harness_ancestry_pid` in `bin/fm-session-lock-lib.sh` walks the process ancestry and runs `basename "$comm"` on each ancestor's `ps -o comm=` value (line 34). A macOS login shell reports its command name with a leading dash, literally `-zsh`, which BSD basename parses as an option and rejects.

Reproduced ancestry (Claude Code launched from a herdr-managed login shell):

```
41619: /bin/zsh
4684:  claude
59422: -zsh          <- basename "-zsh" errors here
59421: /opt/homebrew/bin/herdr
```

```
$ basename "-zsh"
basename: illegal option -- z   (exit 1)
$ basename -- "-zsh"
-zsh                            (exit 0)
```

Impact is noise only in the observed case: `bc` ends up empty for that ancestor, the walk had already matched the outermost `claude` pid, and the lock resolves correctly. But any layout where the ancestor ABOVE a login shell also matters would mis-evaluate that hop, and the stderr noise lands in every digest and hook message.

## Fix

Use the end-of-options guard everywhere a process comm is passed to basename:

- `bin/fm-session-lock-lib.sh:34` -> `bc=$(basename -- "$comm")`
- `bin/fm-session-lock-lib.sh:72` -> `basename -- "$comm"`
- `bin/fm-harness.sh:51` has the same pattern (`basename "$comm"` inside a case) and should get `--` as well.

`basename --` is POSIX-standard end-of-options and works on both BSD/macOS and GNU coreutils.

adonis-garcia-git:
## Summary

Fixes the overnight "silent watcher deaths" on macOS and locks in the BSD basename portability fix on the Stop-hook path.

**Defect 1 - watcher cycles reported dead while healthy.** Root cause, established from the cycle-exit ledger, a live process monitor, `pmset -g log`, and `kern.waketime`: the watcher never crashed. A clamshell Mac cycling ~900s maintenance sleeps suspends the watcher with everything else, so its liveness beacon aged 900s+ in wall-clock terms the instant each DarkWake began, and every wall-clock judge (attach-mode arm, turn-end guard, Stop auto-arm, singleton re-arm refusal) declared a live watcher dead. Every "died" cycle in the ledger actually exited 0 with an actionable reason; a 15-second-budget herdr event reader showed 13+ minutes of suspended wall time.

Fix: beacon staleness is now judged in awake time. `fm_path_age_since_system_wake` in `bin/fm-wake-lib.sh` (single owner) clamps the age to the system's last wake, read from `kern.waketime` on Darwin - empirically verified to update on DarkWake too. Non-Darwin platforms keep plain wall-clock behavior byte-for-byte. The attach failure path now distinguishes a live identity-matched holder whose beacon froze while awake, printing an actionable possible-hang reason instead of claiming the cycle ended. The absorb/surface contract is untouched, and the lifecycle ledger's `beacon_age` stays wall-clock on purpose as diagnostic evidence.

**Defect 2 - `basename: illegal option -- z` on every Claude Stop hook (#1280).** Root cause: the pre-#1256 harness ancestry walk called bare `basename` on a login shell's `-zsh` comm; BSD basename parses it as flags. HEAD already carries the portable `basename --` fix from #1256 (the noise was pure stderr; classification was unaffected), so this PR adds the missing regression test locking silent classification of dash-leading comms, skipped where `ps -o comm=` does not carry argv[0] (Linux procps). A sweep found no other GNU-only utility flags on macOS-reachable paths.

**Test-harness fixes found along the way.** `tests/lib.sh` pins `FM_SYSTEM_WAKE_EPOCH_OVERRIDE=0` suite-wide so backdated-beacon tests stay deterministic on hosts that recently slept, and exports `--experimental-strip-types` via `NODE_OPTIONS` when node accepts it - node 22.6-23.5 could not import the `.ts` Pi/OpenCode extension fixtures at all, which had been masking that coverage. The pipeline also made the pi-watch suite hermetic against slow login profiles.

## Validation

- `bin/fm-lint.sh` (ShellCheck 0.11.0 pinned) and `bin/fm-doc-audience-check.sh` pass.
- New regression tests: sleep-window survival, awake-hang verdict, typed live-holder hang failure (`tests/fm-watcher-lock.test.sh`), sleep-aware supervision status (`tests/fm-turnend-guard.test.sh`), dash-leading comm silence (`tests/fm-claude-stop-autoarm.test.sh`).
- Full changed-file suite sweep green; the no-mistakes pipeline (review, test, document, lint) passed with only info-level findings.
- Three pre-existing failures reproduce identically at the base commit on the diagnosis machine and are out of scope: the Pi hung-successor recovery test on node 22, the secondmate-harness pi-signed ancestry test under a Claude session, and the Kimi hook install test.
- Maintainer-verification record for the `kern.waketime` DarkWake fact added under `docs/verification/supervision.md`.

Fixes #1280.


kunchenguid:

PR #1365, opened 2026-07-31
feat(spawn): refuse dispatch onto a branch another live session is using

DigiThomas:
## Why

A clean working tree at a known tip is not evidence that a branch is idle. The pre-dispatch check was point-in-time; the hazard has duration.

The incident: firstmate inspected a branch, saw a clean tree at a known tip, and dispatched a worker onto it. An interactive session in a **different directory for the same repository** was still working that branch and pushed twenty minutes later. The two sessions overlapped 31 minutes. The worker's push was rejected, it misattributed the cause to one of its own review helpers, and then "fixed" that non-problem in a way that broke something real.

The fault is the dispatcher's, not the worker's.

## What this adds

`bin/fm-branch-liveness.sh` - refuse-grade evidence that a session is currently alive on a branch. Three detectors, all must clear:

| detector | signal |
| --- | --- |
| `session` | live agent session transcripts, matched on `gitBranch` plus a `cwd` that resolves into the same **repository** |
| `checkout` | local checkouts of the repository on that branch **showing current activity** - a dirty tree, or an index/HEAD/reflog touched inside the activity window |
| `remote` | a branch head on the remote this checkout has never seen |

`bin/fm-spawn.sh` runs the check before any worktree or session provider is created, and refuses rather than warning.

### Design points worth knowing

- **Repository identity is git-common-dir or normalized remote URL, never path equality.** This is load-bearing: the collision was two different directories for one repository, so a path comparison would have missed it entirely.
- **Freshness comes from the transcript record's own timestamp, not file mtime.** Verified: the incident transcript's mtime was about three hours later than its last real turn, so mtime alone over-reports live sessions. mtime is still used as a cheap pre-filter, since it can only ever be newer than the last appended record.
- **The checkout detector requires activity, not merely having the branch checked out.** Branch state was already known to be insufficient; a clean, idle, abandoned checkout on the branch is not a live session and does not refuse.
- **The remote detector only runs when the branch already exists locally**, so ordinary per-task-branch dispatch never touches the network and an offline machine can still dispatch normal work.
- **Ownership resolution lives in `fm-spawn.sh`, not in the detector**, so the detector stays a pure "is anyone live on this branch" question.

### Fail-closed direction

Detection that cannot complete exits non-zero and refuses. `exit 0` means "checked, and nothing is there", so degrading an unreadable signal to `0` would restore exactly the silent all-clear that caused the incident.

An **absent** signal is deliberately different and clears: no transcript root means no agent sessions were ever recorded, and no remote means no remote actor. Both are real answers, not gaps.

### Override

`--allow-branch-collision` dispatches anyway. It is explicit, prints every finding as a warning, and records `branch_collision_override=1` in the task's durable metadata, so an overridden dispatch stays visible after the fact. It is not settable by config or environment.

### No false positives on the dispatcher's own workers

Those workers live in isolated worktrees of the same repository by design, so treating them as collisions would refuse every ordinary dispatch. The home, the project clone, the task's own worktree, and the task's own recorded metadata are excluded. The regression test for this uses a **control**: the same fixture refuses without the ownership evidence and clears with it, so the assertion cannot pass vacuously.

Ownership exclusion is scoped to the task's own record, so a *sibling* worker already live on a declared existing branch is refused rather than cleared. A firstmate worker is another live session too.

## Disclosed limitation - please read

**The brief scan matches command form only.** It looks for `git checkout` / `git switch` in the task brief and requires the branch to be declared with `--branch` when it names a branch other than the per-task one. A brief that names a branch **in prose** - "continue the work on story/1-4" with no git command - is **not** caught, so the guard silently checks the per-task branch instead of the branch the worker will actually use.

That exact case has already occurred in practice on another task, and **this guard would not have caught it**. The check itself is sound once pointed at the right branch; the gap is in inferring which branch that is. Closing it properly means declaring the branch at intake rather than recovering it from prose, which is outside this change.

Two further limits, stated rather than hidden, and both documented in the script header:

- The session detector reads Claude Code transcripts. Other harnesses are covered only through the harness-agnostic `checkout` and `remote` detectors.
- The checkout detector's candidate inventory is bounded to the clone's own worktrees, directories named by live sessions, and the home's project clones. An independent clone that no live session has touched is outside it; finding those would mean scanning the filesystem, which this guard deliberately does not do.

## Tests

12 cases in `tests/fm-branch-liveness.test.sh`, all passing, covering: a genuine cross-clone collision refused with named evidence; the same branch name in an unrelated repository cleared; a session quiet beyond the window not treated as live; the dispatcher's own recorded worktree cleared with a control run that refuses without the ownership evidence; sibling workers on their own branches never blocking an ordinary dispatch; the checkout detector refusing on activity rather than branch state; a remote head never seen locally refusing and clearing once fetched; a fresh per-task branch never reaching the network; unreadable signals failing closed while an absent signal stays a real answer; the dispatch path refusing and dispatching only under the explicit override; a brief pointing at another branch refusing until declared; and a task's own isolated copy on its own branch never blocking that task.

Lint (`bin/fm-lint.sh`), the documentation-audience check, and the test-coverage guard all pass.


kunchenguid:

PR #1377, opened 2026-07-31
fix(afk): deliver away-mode escalations when the primary pane is unusable

jamesstomberg:
## The incident

Away mode stalled overnight twice with the identical log line — 2026-07-21 and 2026-07-30:

```
inject deferred: supervisor composer not confirmed-empty (state=pending)
```

The second occurrence buffered **every escalation for 5.6 hours** while four finished PRs waited to be merged, and the daemon still reported itself armed the whole time. Nothing was lost — the durable queue replayed everything on return — but the entire point of away mode (landing work unattended) did not happen.

**It reproduced a third time on 2026-07-31, within ten minutes of arming**, with a worker genuinely blocked and waiting:

```
fm away-mode inject WEDGED: 307s undelivered as of 2026-07-31T13:45:06+0200
```

So this is not intermittent bad luck. On a home where the primary sits in a state the shared classifier reads as `pending` or `unknown`, it reproduces on essentially every session.

## Why the guards are kept

The delivery guards are each correct and stay. There is no safe way to type into a composer holding unsubmitted text, and no backend offers a write path that bypasses the composer, so a refusal cannot be relaxed into an injection without risking a corrupted prompt.

**The defect was that a refusal carried no duration and no cause.** A primary parked in one blocking state blocked every injection indefinitely, while the only channel a captain could actually reach said nothing useful. The max-defer escape could not help either, because its one retry needs the same idle pane that is unavailable.

## The change

- **Record which guard is refusing, and when it first refused**, in `state/.subsuper-delivery-blocked`. A changed blocking state restarts the clock, so only an unchanging state accrues a reportable stall and a genuinely in-use pane is never misreported as wedged.
- **Make the pane-independent active alert an actual delivery path.** It now carries the buffered escalations themselves plus the recorded cause and duration — *"the primary composer has held unsent text for 5h39m"* — instead of pointing at a marker file an away captain cannot open. The durable marker keeps the full buffer and records the real cause rather than asserting "busy or wedged" for every failure.
- **Add `bin/fm-afk-artifacts-lib.sh`** as the single owner of the delivery-artifact set, which was previously spelled out literally in five places across three scripts. A failed source in the return path is now fatal rather than silently leaving delivery state uncleared.

## Testing

Regression coverage in `tests/fm-daemon.test.sh` reproduces the incident shape and **fails against the unfixed daemon**: the guard still refuses, the buffer survives, and the alert carries the work plus the cause and duration.

Full suite green (101 files, 0 failures).

## Notes for review

- No behaviour change when delivery is healthy — the new state file is only written on a refusal.
- The guards' refusal semantics are unchanged; this only adds observability and an alternate delivery path for the already-buffered work.
- Docs updated: `docs/architecture.md`, `docs/configuration.md`, `docs/wedge-alarm.md`, and the `afk` skill.

### DIFF
--- a/.agents/skills/afk/SKILL.md
+++ b/.agents/skills/afk/SKILL.md
@@ -111,6 +111,8 @@ If that submit cannot be confirmed, it raises a loud, rate-limited wedge alarm:
 an ERROR in the daemon log, a durable
 `state/.subsuper-inject-wedged` marker (surface it on the "while you were out"
 catch-up if present), a tmux status-line flash when applicable, and a configurable backend-independent active alert.
+Because no backend can write to an agent except through the composer the guard may be refusing, that active alert is the only channel left while the pane is unusable, so it carries the buffered escalations themselves plus the recorded cause and duration of the block.
+`state/.subsuper-delivery-blocked` records which guard is refusing and when it first refused; the clock restarts whenever the blocking state changes, so only an unchanging state accrues a reportable stall.
 `docs/wedge-alarm.md` owns the alert channel setup, and `docs/verification/supervision.md` "Wedge-alarm channels" owns active evidence.
 So a guard false-positive becomes a visible stall, never an unbounded silent no-op.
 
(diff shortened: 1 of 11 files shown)


kunchenguid:

PR #1392, opened 2026-07-31
feat(cursor): Cursor Cloud agent view, steering, and creation

SH4DY:
## Why

Firstmate had no way to see or drive the Cursor Cloud agents that the team's work actually runs on. Cloud agents run on Cursor's infrastructure, so they fit none of firstmate's existing seams: there is no pane to capture, no composer to submit into, and no local worktree to isolate.

## How

`bin/fm-cursor.sh`, a guarded helper over the Cloud Agents API using `curl` and `jq` - the same pair X mode already requires, so no new runtime dependency.

Deliberately **not** a runtime backend and **not** a harness. Most of `bin/fm-backend.sh`'s adapter contract could only be stubbed with lies, and a cloud spawn could never satisfy the worktree-isolation assertion in `bin/fm-spawn.sh`, which protects a real safety invariant. It follows the companion-surface precedent already set for Codex Desktop threads. `bin/fm-backend.sh`, `bin/fm-spawn.sh` and `FM_BACKEND_KNOWN` are untouched, and it creates no task records and registers no watcher check - which is what stops a windowless, worktreeless agent from being read as a stalled crewmate by session-start recovery.

Two API facts drive the shape, both verified against the live API:

- An agent's `status` is **lifecycle only** (`ACTIVE|ARCHIVED`); execution status lives on runs. `ACTIVE` means "not archived", never "currently running" - 12 of 12 sampled ACTIVE agents had FINISHED latest runs, some six weeks old. `list` therefore resolves each agent's latest run and reports the run status as the primary column.
- A Cursor **agent** maps onto a firstmate task and a Cursor **environment** maps onto a project. `POST /v1/agents` takes either a named environment or a bare repo list, documented as mutually exclusive, and the environment is what carries the predefined secrets and MCP configuration. An agent belongs to its environment, never to one repository inside it, so `list` and `show` lead with the environment.

`config/cursor-environment` optionally names a default environment. It is a **target** for operations that need one, not a view preference, so it marks the default with `*` and never filters `list` implicitly.

## Credential handling

Presence-gated on `CURSOR_API_KEY` in the gitignored `.env`, read from that file only - never an ambient variable, which would silently change which account firstmate speaks for, and never the macOS keychain or `cursor-agent`'s stored credentials. The key reaches `curl` through a mode-0600 config file, never `argv`, since a process argument is readable via `ps`. Both the key charset and the timeout are validated before touching that file, because it carries the bearer header and its syntax is injectable.

Work in progress: read-only `list`/`show`/`runs`/`usage` is complete and validated; steering and creation are being added to this branch.

### DIFF
--- a/.agents/skills/firstmate-cursor-cloud/SKILL.md
+++ b/.agents/skills/firstmate-cursor-cloud/SKILL.md
@@ -42,6 +42,15 @@ A finished agent stays `ACTIVE` indefinitely until somebody archives it, so a fl
 Never report agent lifecycle as though it were activity.
 Take the run status from the latest run, which `bin/fm-cursor.sh list` resolves and reports as its primary column: `CREATING` and `RUNNING` mean work is in flight, and `FINISHED`, `ERROR`, `CANCELLED`, and `EXPIRED` are terminal.
 
+That enum is not the whole column, and the two values outside it are the ones most likely to be misread.
+`unknown` is a third category that is neither in flight nor terminal: it means resolving that agent's latest run failed, so its real state was never observed.
+`unresolved` is the same absence by choice rather than by failure, and appears only under `--no-runs`, which skips resolution entirely.
+Never read either value as idle, finished, or terminal, and never fold an `unknown` row into a "nothing running" count.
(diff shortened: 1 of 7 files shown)


SH4DY:
Closing — this work is moving to the SH4DY fork, where development will continue. Reopening against upstream later if it's wanted there.

PR #1404, opened 2026-07-31
fix(snapshot): route registry parsing through the #1452 owner

JaredHuynhning:
## Rescope

This pull request now builds on #1452, which established `bin/fm-secondmate-registry-lib.sh` as the strict owner of `data/secondmates.md` parsing and rejects ambiguous duplicate bindings.
The former competing `fm-registry-lib.sh` implementation and its last-match-wins contract have been removed from this branch.

## What changed

- `bin/fm-fleet-snapshot.sh` now parses its bounded registry window through the #1452 owner instead of carrying an independent jq field extractor.
- The owner exposes a strict ID-only helper so malformed rows with valid IDs can retain their existing snapshot diagnostic while invalid IDs remain rejected.
- Duplicate IDs remain errors; no row is silently selected.
- The stock macOS Bash snapshot check now expects the added 16th behavior test.

## Reference audit

- `bin/fm-bootstrap.sh` already routes its parser call and registry helper handoffs through `fm-ff-lib.sh` to the strict owner, so it is unchanged.
- `bin/fm-spawn.sh` already uses the strict field and binding validators; its remaining registry reference is only an existence check, so it is unchanged.
- `bin/fm-update.sh` already calls the strict line parser through `fm-ff-lib.sh`; its other registry references are the path variable and explanatory comment, so it is unchanged.
- `bin/fm-fleet-snapshot.sh` contained the only remaining independent parser among the four audited scripts, and that parser is replaced here.
- `bin/fm-test-run.sh` has no registry references on current `main` and is unchanged.

## Validation

- The new snapshot behavior test was observed failing against the independent jq parser and passing after the owner integration.
- Stock `/bin/bash` snapshot suite: 16 tests passed.
- Adjacent update, bootstrap, and secondmate safety suites passed.
- Documentation audience check and pinned ShellCheck gate passed.
- A 256-row full snapshot completed in 1.12 seconds with the registry timeout set to 2 seconds.

## Pipeline

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -350,8 +350,8 @@ jobs:
           snapshot_output=$(/bin/bash tests/fm-fleet-snapshot-view.test.sh)
           printf '%s\n' "$snapshot_output"
           snapshot_count=$(printf '%s\n' "$snapshot_output" | grep -c '^ok - ')
-          [ "$snapshot_count" -eq 15 ] || {
-            echo "::error::expected 15 snapshot/fleet-view tests, got $snapshot_count"
+          [ "$snapshot_count" -eq 16 ] || {
+            echo "::error::expected 16 snapshot/fleet-view tests, got $snapshot_count"
             exit 1
           }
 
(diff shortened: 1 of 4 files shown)


kunchenguid:

Issue #1336, opened 2026-07-30
spawn: a failed bootstrap leaves a labeled pane with no state - invisible to all supervision


jokim1:
## Summary

A crewmate spawn whose bootstrap fails (e.g. `treehouse get` refusing) leaves a live, labeled backend pane with **no `state/<id>.meta` and no `state/<id>.status`**. Because the watcher and every supervision mechanism key exclusively on state files, these panes are invisible orphans: no stale detection, no heartbeat surfacing, no teardown path ever reaches them.

## Reproduction (observed 2026-07-30, herdr backend, two independent instances)

1. Exhaust the treehouse pool for a project (`max_trees = 16`, all worktrees in use or dirty).
2. Spawn a crewmate for that project through `bin/fm-spawn.sh`.
3. The backend pane is created and labeled first; bootstrap then runs `treehouse get`, which fails:
   ```
   josephkim@Mac clawtalk % treehouse get
   🌳 Setting up worktree...
   all 16 worktrees are in use or dirty (max_trees = 16). ...
   ```
4. No agent starts; no `state/<id>.meta` or `state/<id>.status` is ever written; zero watcher `.seen-*`/`.hb-*` markers.

Observed at scale: 6 panes from one failed per-PR triage fan-out (spawned within 300ms of each other) sat as byte-identical dead login shells for 11h43m+; 3 more from a second fan-out (failed on a stale remote-ref lock `refs/remotes/origin/<branch>` in `treehouse get`) the same day. A same-day fallback respawn on tmux for two of those tasks then produced duplicate dead herdr panes that survived even a manual state cleanup, because pane closure and state deletion are not coupled for stateless panes.

## Root cause

The spawn path creates the pane before it records state, and records state only on success. Supervision watches only state. A failed spawn therefore produces a pane no mechanism can see.

## Proposed fix (smallest structural)

In `bin/fm-spawn.sh`, write `state/<id>.meta` (or at minimum a `failed:` line to `state/<id>.status`) when bootstrap fails - or write meta before pane creation and update it - so a failed spawn presents as a terminal state that the ordinary done/failed handling path can reap with the existing guarded teardown. No new reconciler or watcher needed: just make every pane born with state.

## Secondary observation

Worktree-pool exhaustion is commonly caused by long-parked landing-gated tasks holding slots, so this failure mode clusters exactly when the fleet is busiest.

kunchenguid:
Speaking as Kun's firstmate: this is still real on latest main.

[...]

PR #1409, opened 2026-07-31
fix(watch): attach --restart to a healthy watcher instead of force-replacing

calebkeller2:
## Intent

Repair the watcher restart contract so `bin/fm-watch-arm.sh --restart` attaches to an already healthy, identity-verified watcher without signaling it. Preserve home scoping, singleton ownership, stale-lock recovery, PID-reuse resistance, bounded replacement behavior, Bash 3.2 compatibility, and the live primary watcher. Keep this change separate from the parked #1336 branch.

## What changed

- Moved the shared healthy-watcher gate ahead of arm/restart mode dispatch.
- Both ordinary arm and `--restart` now attach to a healthy identity-matched watcher and follow that cycle.
- Replacement remains limited to an unhealthy holder recorded in the same `FM_HOME` lock.
- Strengthened `tests/fm-watcher-lock.test.sh` with a ready-synchronized TERM-resistant peer, a no-TERM assertion, and failure-path fixture cleanup.
- Updated the maintained architecture contract for the new restart behavior.

## Diagnosis

- Initiating trigger: `--restart` entered its replacement path before checking shared watcher health.
- Masking condition: a peer that had already installed its TERM handler could survive the signal and make the later attach path appear correct.
- Visible symptom: the focused test observed `watcher: started` instead of attaching to the verified healthy peer, and the failed fixture could retain the test pipe.
- Counterfactual: ordinary arm attached successfully to the same healthy, identity-verified state, isolating the divergence to restart mode.

## Validation

No-mistakes run `01KYWKT69M4Y1NT901Z8JTEXW1` completed rebase, independent review, full tests, documentation review, lint, fork push, PR creation, and its checks-passed decision point on rebased head `eaf5c96`.

The branch was pushed to the approved public fork `calebkeller2/firstmate` and this PR targets `kunchenguid/firstmate`. The pull request is intentionally unmerged.

### DIFF
--- a/bin/fm-watch-arm.sh
+++ b/bin/fm-watch-arm.sh
@@ -49,11 +49,10 @@
 # state/.watch-triage.log remains exclusively the watcher's absorbed-wake debug
 # log and is never written here.
 #
-# --restart: stop ONLY this FM_HOME's watcher (the pid recorded in THIS home's
-# state/.watch.lock) and own a fresh cycle, or attach if a verified live peer
-# wins the singleton while the duplicate child stands down. It
-# resolves and signals exactly that pid, so it can never touch another home's
-# watcher. NEVER `pkill -f
+# --restart: attach when THIS home's recorded watcher is identity-verified and
+# healthy. Only when replacement is required does it stop the recorded watcher
+# and own a fresh cycle. It resolves and signals exactly that pid, so it can
+# never touch another home's watcher. NEVER `pkill -f
 # bin/fm-watch.sh`: that pattern matches every firstmate home's watcher
 # (secondmate homes run the same script) and would kill siblings.
 set -u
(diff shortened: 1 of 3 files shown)


josh-padnick:
Reinforcing this from a separate home: we hit the arm/restart path failing in a way this PR looks relevant to.

Observed today on a home with ~12 lanes in flight (claude primary, herdr backend):

- `bin/fm-watch-arm.sh` exited non-zero with `watcher: FAILED - no live watcher with a fresh beacon`, twice in a row, while the Stop-hook alarm reported supervision down. A plain re-arm did not recover it; `bin/fm-watch-arm.sh --restart` did, immediately reporting `watcher: started pid=... (beacon fresh)`. So the ordinary arm path could not re-establish a cycle that `--restart` could.
- Separately and earlier, we saw repeated `watcher: FAILED - cycle ended without an actionable reason` Stop alarms while a successor cycle was demonstrably healthy (its triage log was absorbing wakes normally throughout).

Two further symptoms we believe are distinct from this PR, noted here only so they are not mistaken for it:

- Long tool executions (multi-minute builds and pipeline calls) appear to defeat the busy-pane signature at poll instants, producing `possible wedge` escalations up to `demand-deep-inspection` for panes that deep inspection proved were mid-productive-turn.
- Panes whose reconciled state is terminal or parked awaiting a human keep re-alarming as wedge candidates; with many parked lanes this produces sustained no-op wake churn.

We have a local fix for the latter two (suppressing wedge candidacy for panes that are parked, finished, or unprovable, without weakening genuine wedge detection) that we have deliberately not pushed while this PR is open, to avoid duplicating or conflicting with work already in flight. Happy to open it separately, fold it in, or drop it if it is already covered here - whichever the maintainers prefer.

PR #1410, opened 2026-07-31
feat(bin): refuse a byte-identical or unpairable before/after evidence image pair

wongsuwarn:
## Summary
- Add `bin/fm-evidence-check.sh`, a deterministic check that refuses a merge when a PR's before/after evidence images are byte-identical, or when a directory's before-*/after-* images can't be paired at all - both are the fabricated- or unverified-visual-check shape this tool exists to catch.
- Pairing supports the naming conventions seen across the fleet's history: exact `before-<suffix>`/`after-<suffix>` (or reversed `<prefix>-before`/`<prefix>-after`), and a shared leading `<N>-` ordinal when the suffix instead describes state (e.g. `before-1-typed-not-saved.png` / `after-1-typed-with-save-button.png`). An explicit, greppable `.allow-identical` marker opts a genuinely-intentional identical pair back in.
- Wire it into `bin/fm-pr-merge.sh`: before merging, the check runs against the actual GitHub PR head (freshly fetched `refs/pull/<n>/head`, falling back to the recorded `pr_head=`, then the local worktree HEAD with a warning) rather than a possibly-stale local worktree, following the same head-resolution pattern `bin/fm-review-diff.sh` already uses.
- One-line pointer added to `AGENTS.md` so a ship brief capturing before/after evidence points a worker at `bin/fm-evidence-check.sh --help`, which remains the sole owner of the naming convention.

## Test plan
- [x] `bin/fm-lint.sh` (shellcheck clean)
- [x] `tests/fm-evidence-check.test.sh` (12 cases: identical pair refused, genuine diff passes, after-only/no-images pass silently, reversed convention, ordinal pairing on real fleet-history filenames, unpairable set refuses instead of a silent ok, a leftover mismatched pair beside a matched pair still refuses, a genuinely one-sided extra beside a matched pair still passes, `.allow-identical` opt-out, `--local` mode, usage errors)
- [x] `tests/fm-pr-merge.test.sh` (identical-evidence pair refuses the merge, a genuinely different pair still merges, evidence is checked against the fetched remote PR head rather than a stale local worktree HEAD)
- [x] Hand-verified against four real commits from `we-pack-together`'s PR-evidence history (the fabricated pair this tool exists to catch, two genuine before/after sets, one legitimate after-only set) - all behave as expected.


wongsuwarn:
Superseded - this fix landed directly on the local firstmate installation instead, since upstream contribution isn't required for this project and push access to this repo isn't available. Closing rather than leaving open indefinitely.

PR #1429, opened 2026-07-31
feat(dashboard): mission-control board, typed queue, and fm-mission-control.v1 schema

josh-padnick:
Phase 1 of the captain's mission-control dashboard: a single-viewport, self-updating view of everything waiting on him, generated from live fleet state. It replaces a hand-maintained HTML page that firstmate rewrote by hand on every change.

## What ships

- **A structured state document** (`data/mission-control.json`, schema `fm-mission-control.v1`) as the primary output. The HTML is a rendering *of* that document and nothing else.
- **A board** led by one dominant "Needs your attention" queue at ~60% of the first viewport, over compact active-work and landed summaries.
- **An experimental inbox view** in the same artifact, so the two can be compared before one is chosen.
- **A generator**, `bin/fm-dashboard.sh`, that projects `bin/fm-fleet-snapshot.sh` rather than re-reading fleet state.
- **19 behaviour tests** plus a bounded per-task activity feed added to the canonical snapshot.

## The JSON schema, `fm-mission-control.v1`

Stated explicitly because **Phase 2's lane is building against it now** — it is a contract, not an implementation detail. The script header is its authoritative owner.

Top level:

- `schema`, `generated` (UTC ISO-8601), `generated_display` (local, human), `fm_home`, `refresh_seconds`
- `source: { fleet_snapshot: {schema, generated}, pr_data: {present, fetched, age_seconds, stale, ttl_seconds, records} }`
- `counts: { inbox, mine, worker_turn, external_turn, merge_queue, awaiting_captain, ui, plan, decision, decisions, in_flight, in_flight_live, completed, notices }`
- `merge_queue[]`, `awaiting_captain[]`, `in_flight[]`, `completed[]`, `notices[]`

**TURN — carried by every `merge_queue`, `awaiting_captain` and `in_flight` item:**

- `turn`: `captain` | `worker` | `external` — whose move it is, derived from what the fleet shows, never from which list the item sits in. An open PR whose lane is still running its pipeline is the *worker's* turn; checks that have not reported are *external*.
- `headline` — one line saying which situation this is: "Yours to review and merge" or "Blocked on <the concrete thing>".
- `action`: `merge` | `look` | `read` | `decide` | `null` — null unless it is his turn.
- `waiting_for`, `turn_reason`, `since`, `since_source` (`backlog` | `status-log` | null), `since_epoch`, `age_seconds`
- `activity: { text, source: run-step | status-log | state, state }` — what is happening now, in the captain's nouns. Phase 2 pushes transitions of this field as events.
- `attended`: `true` | `false` | `null`, with `attended_by` (`pipeline` | `worker` | null) and `attended_evidence` (`run-step` | `live-endpoint` | `no-endpoint` | `idle-endpoint` | `unverified`)
- `stuck` — true when a worker reported `blocked:` (needs help), false for `needs-decision:` (waiting on the captain by design)

**`awaiting_captain[]`** additionally carries `type` (`ui` | `plan` | `decision`), `kind`, `title`, `note`, `link`, `link_kind`, `project`, `source`, `decision_count`, `recommended_count`, and `decisions[]` of `{id, key, question, recommendation, recommendation_source, reasoning}`.

**`merge_queue[]`** additionally carries `url`, `number`, `repo`, `status` (`ready` | `held`), `checks: {state, summary, source}` and `blockers[]` of `{kind, text}`.

**`in_flight[]`** additionally carries `state`, `state_source`, `live`, `needs_captain`, `pr`, and `feed[]` — the history, distinct from `activity`, which is the single thing happening now.

**The inbox is derived, not stored.** It is every item across those three arrays whose `turn` is `captain`, deduped by `id`; `counts.inbox` is its size. Both views read the same arrays, so they cannot disagree about what is waiting.

## Why this shape

- **JSON first.** Phase 2 serves the same object, so nothing here is thrown away. A test proves the page is byte-reproducible from the state document alone.
- **Turn is structural, not cosmetic.** The board previously showed a mid-pipeline PR as "ready" — something in the captain's queue he could not act on. That failure is what `turn` exists to prevent.
- **Blocked is three states.** Being-fixed, nobody-on-it, and ownership-unverified are different situations. `attended: true` requires positive evidence; a session that merely exists is not someone working. `null` is rendered as its own state, because guessing "stalled" is a false alarm and guessing "being worked" is false comfort.
- **A plan with decisions is one item.** `awaiting_captain[]` already modelled one row per origin with nested decisions; the separate decisions column was duplicating them.
- **Typing uses dispatch kind, not file extension.** Plans render to HTML here too, so the extension cannot separate "read this" from "look at this". Scout work produces knowledge to read; ship work produces a page to look at.
- **Honesty over convenience.** PR checks are fetched only by `--refresh-prs` and cached, with the cache's age published and marked stale past a TTL. Links are published only when they resolve to a file this home really has. Trimmed history, incomplete inventory and unreadable homes are disclosed rather than left to be discovered by clicking.

## Which view I would keep

**The inbox** — but only because the turn model exists; without it an inbox would list things that are not actually his.

Its honest caveat is that the board still does two things the inbox does not: decisions grouped inside their plan so several related questions get answered in one pass, and the per-workstream narrative that answers "is anything stuck?". Before the inbox could stand alone it needs expand-in-place on a row's decisions.

## Verification

- 19 behaviour tests in `tests/fm-dashboard.test.sh`, plus a bounded activity-feed test in `tests/fm-fleet-snapshot-view.test.sh`. `bin/fm-lint.sh` clean.
- Verified in Chrome at 1512x945, light and dark, both views: no page scroll, panels scrolling internally, expansion, filtering, and `j`/`k` navigation all exercised on live fleet state.
- Auto-update proven end to end: a status event appended to a live lane appeared without a manual refresh.
- `fetch` and `XMLHttpRequest` are both blocked from a `file://` origin in Chrome, which is why the page reloads itself rather than polling.

## Deliberate exceptions

- **Page scroll below 760px wide or 560px tall.** The no-scroll invariant holds at every desktop size; a height-locked five-panel console at phone size would be five unusable slivers.
- **Persistent secondmate lanes are excluded** from `in_flight`. An idle secondmate is healthy, and drawing it as a stalled lane would misreport it.
- **Epochs are resolved in shell, never in jq.** jq 1.6's `fromdateiso8601` *and* `strptime`/`mktime` both apply local daylight rules to UTC input, which put the captain-facing stamp an hour out. BSD `date -j -f` has the same flaw unless `TZ=UTC` is set for the parse.

## Known follow-up

An **identity-model rework** is queued separately. Today a task id doubles as the identity of a reviewable item, so one task can produce a PR row, an ask row and a lane row that all share a key, and a type-preference constant decides which is shown. Three defects in this PR's review traced to that. It also blocks Phase 2, which cannot address a row that has no unique id. The proposal is `task_id` as a field, a unique `item_id` per row kind, and an explicit `supersedes` relationship replacing the constant.


josh-padnick:
Opened against the wrong base - mission control is the captain's own feature and lands on the josh-padnick fork, not upstream. Reopening there.

PR #1434, opened 2026-07-31
fix: launch codex crews at max effort

kwangmoonbae:
## What

Permit `max` as a reasoning effort for the `codex` harness, so a configured or captain-chosen codex `max` profile is actually launched at max instead of being accepted and then silently dropped.

Before this change the two halves of the contract disagreed:

- `bin/fm-bootstrap.sh` (`effort_ok` in `crew_dispatch_validate`) rejected `codex:max` outright, so a dispatch profile asking for it was reported as `CREW_DISPATCH: invalid config/crew-dispatch.json - invalid effort: codex:max`.
- `bin/fm-spawn.sh` (`effort_flag_for_harness`) accepted `max` but emitted no flag for it, so the worker launched at codex's own default effort while `effort=max` was still recorded in task meta.

The result was a silent downgrade: recorded intent and actual launch disagreed, with no diagnostic.

## Backend verification

Verified against the installed CLI rather than trusting the older documented ceiling.

```
$ codex --version
codex-cli 0.145.0

$ codex debug models --bundled | jq -r '.models[] | "\(.slug): \([.supported_reasoning_levels[].effort] | join(","))"'
gpt-5.6-sol: low,medium,high,xhigh,max,ultra
gpt-5.6-terra: low,medium,high,xhigh,max,ultra
gpt-5.6-luna: low,medium,high,xhigh,max
gpt-5.5: low,medium,high,xhigh
gpt-5.4: low,medium,high,xhigh
gpt-5.4-mini: low,medium,high,xhigh
gpt-5.2: low,medium,high,xhigh
codex-auto-review: low,medium,high,xhigh
```

An end-to-end check confirmed the flag is honored rather than downgraded: `codex exec --model gpt-5.6-sol -c 'model_reasoning_effort="max"'` reports `reasoning effort: max` in its session header.

Note that `codex doctor -c model_reasoning_effort=<value>` does NOT discriminate - it accepts a deliberately bogus value too, so it is not evidence either way.

## Model scoping, deliberately harness-wide

`max` is model-scoped in codex-cli 0.145.0: only the `gpt-5.6` family carries it, and every other model caps at `xhigh`. Firstmate's acceptance is harness-wide by decision, not because every codex model supports it.

Two catalogs exist and they differ in which older models they carry - `codex debug models --bundled` is the static compiled catalog and ships `gpt-5.2`, while `codex debug models` is the account's effective list and ships `gpt-5.3-codex-spark` instead. Both agree exactly on which models carry `max`. The contract therefore states the max-bearing family positively and names both recheck commands rather than enumerating a model list that is version- and account-dependent and would rot.

This matters because `model_reasoning_effort` is a free-form string in codex 0.145.0 config loading, unlike `approval_policy` and `sandbox_mode` which reject unknown values. Pairing `max` with a model that caps at `xhigh` is accepted silently and its effective effort is undefined. The contract records that caveat instead of adding a per-model capability check, which would have to be maintained against a shifting catalog.

## Scope

- `bin/fm-bootstrap.sh` - accept `codex:max` in dispatch-profile validation.
- `bin/fm-spawn.sh` - emit `-c 'model_reasoning_effort="max"'` for codex.
- `.agents/skills/harness-adapters/SKILL.md` - the authoritative contract row, plus the model-scoping and free-form-string caveats.
- `tests/` - colocated regression coverage on both sides.

Unchanged: every other harness ceiling (`grok` stays capped at `high`, `opencode` remains effort-less) and the generic effort-selection policy. Invalid efforts stay loud - `codex:bogus` is still flagged by bootstrap and still refused by `fm-spawn` before any meta write or launch.

## Known follow-up, not in this change

The per-harness effort ceilings live in two executable copies (`bin/fm-bootstrap.sh` and `bin/fm-spawn.sh`) plus the prose owner, so moving one ceiling touches three places. This change adds pointer comments only; consolidating the shell copies behind one shared table changes executable code and belongs in its own change.

### DIFF
--- a/.agents/skills/harness-adapters/SKILL.md
+++ b/.agents/skills/harness-adapters/SKILL.md
@@ -127,6 +127,12 @@ The supported launch-profile flags below are verified locally; each row records 
 | opencode | `--model <provider/model>` | none for firstmate's interactive launch | Verified on opencode 1.17.6. `opencode run` has `--variant`, but firstmate launches the interactive `opencode --prompt` path, which has no verified effort flag. |
 | kimi | `--model <model>` | none | Verified 2026-07-25 on Kimi Code CLI 0.29.1. |
 
+Codex's `max` is model-scoped in codex-cli 0.145.0: the bundled catalog advertises it for `gpt-5.6-sol`, `gpt-5.6-terra`, and `gpt-5.6-luna`, while `gpt-5.5`, `gpt-5.4`, `gpt-5.4-mini`, `gpt-5.2`, and `codex-auto-review` cap at `xhigh`.
+Firstmate accepting `max` harness-wide for codex is therefore a deliberate decision, not a claim that every codex model supports it.
+Codex 0.145.0 also parses `model_reasoning_effort` as a free-form string at config load, unlike `approval_policy` and `sandbox_mode`, which reject an unknown value with `unknown variant`.
+So an unsupported model paired with `max` is accepted silently and its effective effort is undefined rather than loud.
+Pinning `max` on a model outside the `gpt-5.6` family is unvalidated.
+
 The concrete `harness` field owns adapter identity independently of the model provider: `harness=pi` with `model=xai/grok-*` is Pi using xAI, not `harness=grok`, and does not require Grok CLI login; `harness=grok` remains the standalone Grok Build CLI adapter.
 
 ### Model support discovery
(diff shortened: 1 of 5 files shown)


kwangmoonbae:
Closing this at the author's request. No action needed from maintainers.

PR #134, opened 2026-06-29
fix(spawn): make tmux window handling robust under non-default config

Deeds67:
## Intent

The developer was working on the firstmate codebase as an ordinary software project, tasking a coding agent with applying a small robustness fix to bin/fm-spawn.sh's tmux window handling to address two empirically-reproduced bugs that surface under a non-default tmux config (base-index 1 and automatic-rename on): a window-creation collision (fixed by targeting the session with a trailing colon so tmux appends at the next free index) and a dangerous lost-window-name issue where a renamed window caused the wait-loop to misread firstmate's own primary checkout as the worktree (fixed by capturing the stable window id, disabling automatic-rename/allow-rename, and targeting by id). They required the change be confined to that one script plus an optional deterministic test, that bash -n and shellcheck pass, and that the behavior test suite (especially the spawn/tangle tests) pass, with the work shipping through the no-mistakes contribution pipeline. After the code was committed and verified locally, the pipeline was blocked by a no-mistakes gate transport error (invalid gate path: '.'); the developer fixed gate infrastructure on their end (restarting the daemon and clearing the stale ref), and when that did not resolve it, told the agent to stand by while they worked the deeper no-mistakes transport/path issue from the firstmate side, instructing it not to retry the gate or change anything.

## What Changed

- `bin/fm-spawn.sh`: create the task window by targeting the session with a trailing colon so tmux appends at the next free index instead of colliding when `base-index` is non-zero, and capture the resulting stable window id from `new-window -P`.
- `bin/fm-spawn.sh`: disable `automatic-rename`/`allow-rename` on the new window and target the worktree wait loop and `treehouse get` send-keys by the captured window id, so a window rename can no longer make the wait loop misread firstmate's own primary checkout as the worktree.
- Added `test_spawn_tmux_window_construction` to `tests/fm-tangle-guard.test.sh` covering the window-construction behavior, and listed it in the `CONTRIBUTING.md` coverage list.

## Risk Assessment

✅ Low: A small, well-bounded robustness fix confined to one script's tmux handling plus a hermetic deterministic test; the only finding is a low-practical-risk consistency gap where the launch and persisted meta target still use name-based addressing the fix otherwise hardened against.

## Testing

[...]

PR #1435, opened 2026-07-31
fix(tmux): force session resolution in list-windows and agent-state probes

zachmarmolejo:
## Summary

- Apply trailing-colon form (`"$ses:"`) to `tmux list-windows` in `fm_backend_tmux_create_task` so a session and a window sharing the same name (e.g. session "DarkOrbit" with a window also named "DarkOrbit") cannot collide during the duplicate-window check. The `new-window` call two lines below already uses this form.
- Apply the same fix to `fm_backend_tmux_agent_state`'s `tmux list-windows` call, which feeds the recovery-grade agent liveness probe - a false `missing` verdict there can authorize relaunching a duplicate agent.

## Test plan

- [ ] Existing `tests/fm-backend-herdr.test.sh` passes (verified locally)
- [ ] Existing `tests/fm-backend-herdr-smoke.test.sh` passes (verified locally)
- [ ] CI green

### DIFF
--- a/bin/backends/tmux.sh
+++ b/bin/backends/tmux.sh
@@ -168,6 +168,8 @@ fm_backend_tmux_agent_state() {  # <target>
   esac
   session=${target%%:*}
   window=${target#*:}
+  # Trailing colon forces tmux to resolve $session as a session, not a window
+  # whose name happens to match (e.g. session "DarkOrbit" + window "DarkOrbit").
   if windows=$(LC_ALL=C tmux list-windows -t "$session:" -F '#{window_name}' 2>&1); then
     inventory_status=0
   else
--- a/bin/backends/tmux.sh
+++ b/bin/backends/tmux.sh
@@ -168,7 +168,7 @@ fm_backend_tmux_agent_state() {  # <target>
   esac
   session=${target%%:*}
   window=${target#*:}
-  if windows=$(LC_ALL=C tmux list-windows -t "$session" -F '#{window_name}' 2>&1); then
+  if windows=$(LC_ALL=C tmux list-windows -t "$session:" -F '#{window_name}' 2>&1); then
     inventory_status=0
   else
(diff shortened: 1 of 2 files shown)


zachmarmolejo:
Closing this: the defect it fixes does not exist.

I tested the claim against real tmux (3.4) rather than reasoning from the code. With a session and a window deliberately sharing the same name, both target forms return byte-identical output:

```
$ tmux list-windows -t "tmuxprobe-collision" -F '#{window_name}'
window0
tmuxprobe-collision
unique-window

$ tmux list-windows -t "tmuxprobe-collision:" -F '#{window_name}'
window0
tmuxprobe-collision
unique-window
```

The reason is in tmux's own synopsis:

```
list-windows [-a] [-F format] [-f filter] [-t target-session]
```

`-t` here is a **target-session**, not a target-window. Target-session resolution never consults window names, so a bare session name is already unambiguous. Passing a window name fails outright:

```
$ tmux list-windows -t "A-window2" -F '#{window_name}'
can't find session: A-window2
```

The asymmetry with the neighboring `new-window -t "$ses:"` line is deliberate, not an oversight. That trailing colon was added in 6d90240 (#134) for an unrelated reason: `new-window` takes a target-window, and an empty window part means "append at the next free index", which is what makes it safe under a non-default `base-index`. Different command, different argument type, different reason.

Since both forms behave identically, the change is harmless in itself. The problem is the comment it adds, which explains tmux's target resolution incorrectly ("Trailing colon forces tmux to resolve $ses as a session, not a window whose name happens to match"). That would leave a wrong explanation in the tree for the next reader, which is worse than no change at all.

Note for anyone hitting this later: the Codex review on this PR endorsed the premise without testing it, and its additional finding pointed at a call site that had already been fixed in dc38ef1 - it was reviewing a superseded commit.

PR #1439, opened 2026-08-01
fix(spawn,teardown): compare worktree paths by identity, not string equality

austinm911:
## Problem

Spawn compared directory paths as strings. On a case-insensitive, case-preserving filesystem (macOS default), `/x/Coding/y` and `/x/coding/y` name the same directory but compare as different strings.

The worktree-settle loop in `fm-spawn.sh` waits for the pane's cwd to differ from the project path before accepting it as the isolated task worktree. When `FM_ROOT` and the pane's reported cwd disagree only in case, that check passes immediately and spawn records **the project clone itself** as `worktree=` in `state/<id>.meta`.

That defeats the isolation assertion: it fails open, accepting the primary project directory as an "isolated worktree".

`fm-teardown.sh` then made it destructive. It ran, in this order:

```sh
branch=$(git -C "$WT" rev-parse --abbrev-ref HEAD)   # = main, because WT is the clone
git -C "$WT" checkout --detach -q
git -C "$WT" branch -D "$branch"                     # deletes the clone's default branch
...
teardown_treehouse_return ... || { echo "error: ..."; exit 1; }   # fails, aborts AFTER the delete
```

The project clone is left on a detached HEAD with its default branch deleted, and teardown exits non-zero with no rollback.

## Reproduction

Observed live on macOS, twice, with `FM_ROOT` resolving as `/Users/<u>/coding/firstmate` while the pane reported `/Users/<u>/Coding/firstmate`:

```
project=  /Users/<u>/coding/firstmate/projects/loadout
worktree= /Users/<u>/Coding/firstmate/projects/loadout
```

```sh
[ "$a" = "$b" ]   # false  <- what the code used
[ "$a" -ef "$b" ] # true   <- the actual truth
```

Teardown deleted `main` in the clone and aborted. Correcting `worktree=` by hand made the same teardown succeed, isolating the cause.

## Fix

**`bin/fm-spawn.sh`** — added `same_directory()` using the POSIX `-ef` same-file test, with a string fallback when a path does not exist yet. Applied to the worktree-settle comparison, `validate_spawn_worktree`, and the secondmate home/operational-dir guards. `path_is_ancestor_of` now walks parents with the same identity test instead of a string prefix match, so it is also correct across case variants and symlinks.

**`bin/fm-teardown.sh`** — two changes:

1. Refuses up front, before any mutation, when the recorded worktree and project are the same directory, naming both paths and the suspect metadata file.
2. Defers hook cleanup and branch deletion until **after** `teardown_treehouse_return` succeeds, and deletes the branch via `$PROJ` rather than detaching `$WT`. An aborted teardown can no longer leave the repository worse than it found it.

## Tests

- `tests/fm-spawn-worktree-settle.test.sh` — case-variant spellings of the project are treated as the same directory
- `tests/fm-teardown.test.sh` — teardown refuses same-directory metadata before changing the project clone; branch is preserved when the return step fails

Full suites pass: spawn-worktree-settle, teardown, teardown-endpoint-safety, spawn-dispatch-profile, secondmate-safety (134 assertions). ShellCheck clean on both scripts (the one remaining SC2034 is pre-existing).

### DIFF
--- a/bin/fm-spawn.sh
+++ b/bin/fm-spawn.sh
@@ -684,6 +684,17 @@ resolved_existing_dir() {
   cd "$path" && pwd -P
 }
 
+same_directory() {
+  local left=$1 right=$2
+  [ -n "$left" ] || return 1
+  [ -n "$right" ] || return 1
+  if [ -e "$left" ] && [ -e "$right" ]; then
+    [ "$left" -ef "$right" ]
+  else
+    [ "$left" = "$right" ]
+  fi
+}
+
 resolve_project_dir_arg() {
   local path=$1
   case "$path" in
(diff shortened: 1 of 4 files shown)


kunchenguid:

PR #1453, opened 2026-08-01
fix(turnend): detect Grok 4.5 Esc:cancel busy; optional mute hatch

aiandwebservices-cyber:
## Summary
- Grok Build 4.5 shows `Esc:cancel` mid-turn instead of `Ctrl+c:cancel`. Busy detection only matched the old token, so long turns looked idle and stop/stale recovery thrashed.
- Match both cancel footers in `fm-busy-lib` and `fm-tmux-lib`, with a unit case and harness-adapters note.
- Optional operator hatch: `state/.mute-turnend-guard` silences turnend stop recovery when present (env-only resolution on the Grok adapter so stripped-PATH fail-open still works). Default is unmuted.

## Test plan
- [x] `bash tests/fm-busy-state.test.sh` (all passed, including Esc:cancel case)
- [x] `bash tests/fm-turnend-guard.test.sh` (grok/jq cases ok; pre-existing Pi undici proxy noise on this host)
- [ ] Smoke: Grok primary mid-turn shows busy via Esc:cancel; stop no longer thrash-loops when a healthy watcher is armed


kunchenguid:

Issue #1383, opened 2026-07-31
watcher: cycle exits 1 with no actionable reason, and arm attaches to the dying cycle and reports it healthy


gcdcintra:
## Summary

Two related failures observed together:

1. A watcher cycle exited non-zero with no actionable reason: `watcher: FAILED - watcher cycle exited 1 without an actionable reason`.
2. A concurrent `bin/fm-watch-arm.sh` **attached to that same cycle while it was already failing**, reported it as healthy on a fresh beacon, and then exited 1 itself.

The net effect is that an arm can report success against a process that is in the act of dying, leaving supervision down while the operator believes it was restored.

## Evidence

Two arms were running against the same home moments apart. The first:

```
watcher: started pid=1178161 (beacon fresh)
watcher: FAILED - watcher cycle exited 1 without an actionable reason
```

The second, launched seconds later, attached to that same PID and reported it healthy:

```
watcher: already running pid 1178161
watcher: attached pid=1178161 (beacon 11s)
```

Both background jobs then terminated with exit code 1. Immediately afterwards:

```
$ pgrep -c -f 'bin/fm-watch.sh'
0
$ echo "beacon: $(( $(date +%s) - $(stat -c %Y state/.last-watcher-beat) ))s"
beacon: 27s
```

So the beacon was fresh - within the 300s grace - while **no watcher process existed at all**. A liveness check based on beacon freshness alone would have concluded supervision was healthy.

`bin/fm-watch-arm.sh --restart` recovered it cleanly (`watcher: started pid=1223496 (beacon fresh)`, 1 process, 15s beacon).

Environment: firstmate `f7d0d0a`, Claude Code 2.1.220, tmux 3.4, tmux backend, Linux, 5 tasks in flight.

## Two distinct problems

**The unexplained exit.** `exited 1 without an actionable reason` is by definition undiagnosable from the message. Whatever the cycle hit, it is not surfaced. Some detail - the failing step, the last poll, a captured stderr tail - would make this reportable rather than mysterious. I cannot supply a root cause because the tooling does not emit one.

**The attach race.** `attached pid=<n> (beacon <t>)` is documented as proof that one live cycle exists, and the supervision protocol says to treat it that way. Here it was proof of nothing: the PID existed and the beacon was fresh, yet the cycle was already failing. A fresh beacon is evidence the process was alive *recently*, not that it is alive *now* or that it will still be alive after the arm returns.

## Suggested fix

[...]

Issue #1381, opened 2026-07-31
supervision: stale auto-arm claim with a dead owner_pid silently disables watcher re-arm


gcdcintra:
## Summary

`bin/fm-claude-stop-autoarm.sh` can leave a claim in `state/.claude-autoarm.lock` whose `owner_pid` is dead. Every subsequent Stop sees an existing claim, concludes recovery is already under way, and stands down. The watcher it originally launched has long since exited (normal after it delivers a wake), so nothing restarts it and **supervision goes dark silently**.

The only thing that surfaces it is `bin/fm-turnend-guard.sh`, which blocks at the next turn end with `TURN WOULD END BLIND`. So the automatic re-arm stops working and the last-resort backstop quietly becomes the primary mechanism.

## Evidence

Observed repeatedly across one long session with 4-5 tasks in flight:

```
$ cat state/.claude-autoarm.lock
epoch=8 owner_pid=315315 outcome=rewake updated_at=1785491744

$ kill -0 315315 ; echo $?
1                      # dead

$ echo $(( $(date +%s) - 1785491744 ))
558                    # claim 9+ minutes stale
```

Meanwhile the guard fired on essentially every turn end:

```
●  TURN WOULD END BLIND - SUPERVISION IS OFF
●  5 task(s) in flight, but no live watcher holds this home lock (last beat: 13s ago).
●  The Stop-owned auto-arm did not claim this home either, so recovery is NOT already under way.
```

Note the second line: the beacon was *fresh* (13s) yet no watcher held the lock, because the cycle had just exited after delivering a wake and nothing re-armed it.

Environment: firstmate `f7d0d0a`, Claude Code 2.1.220, tmux 3.4, tmux backend, Linux.

## Impact

Supervision lapsed on roughly every turn for several hours. Each lapse required a manual `bin/fm-watch-arm.sh --restart` as its own background task. Because the operator cannot repair it any other way, the documented "routine arm and re-arm are owned by the Stop hook, never by you" contract is effectively inverted: manual arming becomes the norm and the recovery-only path becomes routine.

## Suggested fix

Treat a claim whose `owner_pid` is not alive as reclaimable. A liveness check on `owner_pid` before honouring an existing claim would let the next Stop take ownership instead of standing down behind a dead one. A staleness bound on `updated_at` would also work, but a PID liveness check is exact rather than heuristic.

## Note on what I deliberately did not do

[...]

PR #1477, opened 2026-08-01
fix(bin): keep watcher supervision continuous across cycle ends

gcdcintra:
Fixes two defects that took watcher supervision down repeatedly on a live fleet, reported as #1383 and #1381.

Both were reproduced end to end against the real failure before any code changed, and the before/after below comes from running tracked scripts only. That evidence is the substance of this PR, so it leads.

## Before and after, measured

Two arms against one home, then the incumbent cycle is signalled the same way `bin/fm-watch-arm.sh --restart` signals it:

```sh
export FM_STATE_OVERRIDE="$scratch/state" FM_POLL=2 FM_SIGNAL_GRACE=1
bin/fm-watch-arm.sh > armA.out 2>&1 &
bin/fm-watch-arm.sh > armB.out 2>&1 &
kill -TERM "$(cat "$FM_STATE_OVERRIDE/.watch.lock/pid")"
```

Before:

```text
armA.out: watcher: started pid=1578935 (beacon fresh)
armA.out: watcher: FAILED - watcher cycle exited 1 without an actionable reason
armB.out: watcher: attached pid=1578935 (beacon 1s)
live watchers: 0
beacon age: 13s
```

Zero live watchers, behind a beacon still well inside its own 300 second grace. Anything judging supervision healthy from beacon freshness alone would have been wrong. The attached arm's own `watcher: FAILED - cycle ended without an actionable reason` arrived about ten seconds later, by which point the home had already been unsupervised for that entire window.

After, from the same commands:

```text
armA.out: watcher: started pid=1809720 (beacon fresh)
armA.out: watcher: cycle stopped by SIGTERM
armA.out: watcher: FAILED - watcher cycle was stopped by SIGTERM (exit 143)
armB.out: watcher: attached pid=1809720 (beacon 2s)
armB.out: watcher: started pid=1812513 (beacon fresh)
live watchers: 1
beacon age: 1s
```

And for the auto-arm claim, three consecutive Stop firings against a home with in-flight work and a claim whose recorded owner was dead:

```text
before:  Stop 1: exit=0 arms_run=    Stop 2: exit=0 arms_run=    Stop 3: exit=0 arms_run=    epoch: NONE
after:   Stop 1: exit=2 arms_run=1   Stop 2: exit=2 arms_run=2   Stop 3: exit=2 arms_run=3   epoch: outcome=rewake
```

Measured 2026-08-01 with bash 5.2.21, ShellCheck 0.11.0, Linux 6.8.0-136-generic, against an isolated scratch state directory.

## Defect 1: an arm reported a dying cycle as healthy, and could not say why it died (#1383)

Two separate problems arrived together.

**The unexplained exit.** `bin/fm-watch.sh` carried `trap 'exit 1' HUP INT TERM`. A signalled cycle exited a bare `1` and discarded which signal ended it, so the arm layer had nothing to report but an exit code. That is the entire content of `exited 1 without an actionable reason`: the dominant real cause was a signal, and nothing recorded it. A dozen further `|| exit 1` sites, covering a wake that could not be appended to the durable queue and a state check that could not be run, were silent in the same way.

**The attach race.** `watcher: attached pid=<n> (beacon <t>)` is documented as proof that one live cycle exists, and the supervision protocol tells the operator to treat it that way. It was decided from a single read. A pid that exists with a beacon inside the grace window proves the cycle was alive recently, not that it survives the arm. When the attached chain then ended with nothing holding the singleton, the arm simply returned, leaving the home with zero cycles.

Fixed by:

- The watcher names every non-zero exit. A signalled cycle prints `watcher: cycle stopped by SIG<name>` and exits with the conventional 128 plus signal number, so the arm can classify it. A wake that cannot be recorded, or a check that cannot be run, names that step.
- The arm captures the watcher's stderr separately and quotes a bounded tail alongside every `watcher: FAILED` line, so a failure is reportable rather than mysterious.
- An attach is reported only after the target keeps passing the health gate for `FM_ARM_ATTACH_VERIFY` seconds. A target that dies inside the arm's own lifetime is a failed attach, and the arm owns a cycle instead of reporting one.
- Every path that would otherwise return while nothing holds the singleton starts a cycle instead. Repeated arming therefore converges on exactly one live cycle: the watcher singleton makes a redundant start stand down, and the arm attaches to the winner. `FM_ARM_MAX_TRANSITIONS` keeps a flapping home loud rather than endless.

The one deliberate exception: when the ended cycle's actionable reason is already sitting in the durable wake queue, that needs a handling turn now, so it stays a typed non-zero failure naming the queued record count rather than being silently absorbed into a new cycle.

## Defect 2: a claim held by a dead process suppressed every re-arm (#1381)

`bin/fm-claude-stop-autoarm.sh` stood down on any refused single-flight claim. A refusal is not proof that an owner exists. `fm_lock_try_acquire` can refuse with no live owner at all, returning an empty `FM_LOCK_HELD_PID`, so nothing distinguished that state from a genuine concurrent owner.

The concrete wedge: a claim interrupted between `ln -s` and its stray-link cleanup leaves an owner link inside a directory-shaped lock. Because the lock path is dot-prefixed, that entry is hidden, a plain glob never matches it, and `rmdir` fails on every subsequent acquire, permanently. Every later Stop then read an existing claim, concluded recovery was under way, and stood down while no watcher ran at all.

The practical consequence inverted the documented contract: routine re-arm is meant to be owned by the Stop hook with manual arming reserved for recovery, and instead manual arming became routine while the turn-end guard became the primary mechanism.

Fixed by checking owner liveness before honouring a refusal. Owner-pid liveness is exact where a staleness bound on a timestamp is heuristic, so a live owner's claim is still honoured and only a claim nothing owns is reclaimed. `fm_lock_reclaim_unowned` handles the shapes an interrupted claim leaves, including the hidden stray owner link, and never touches a lock whose owner is alive.

## Scope and what is not covered

- `bin/fm-turnend-guard.sh` is deliberately unmodified. It was the backstop that surfaced all of this, and making the symptom quieter there would have hidden the defect rather than fixed it.
- An owned cycle that is signalled, or that self-evicts because the lock no longer names it, still fails loudly rather than restarting. A signal is an explicit stop, and a foreign lock holder means something else is managing the singleton. Restarting in either case would fight the caller.
- Compatibility was reviewed across every supported primary harness and runtime backend, and each axis marked not-applicable only after inspecting its integration surface:
  - Claude, OpenCode, Pi and pi-signed all match `^watcher: FAILED` and classify by exit code and signal. The matched prefix and the exit contract are unchanged.
  - Codex runs the watcher directly under `timeout`, which reports 124 on timeout regardless of the child status, and prints its quiet-checkpoint line without the captured stderr. No classification or output change.
  - Grok reads the arm's status lines through a tracked background task. The reported-reason wording is carried in its supervision protocol doc.
  - Kimi has no tracked arm or watcher integration and follows the generic emitted protocol, so it is not applicable.
  - tmux, herdr, zellij, orca, cmux and codex-app are reached only for crew state. No backend adapter reads the watcher lock, the arm's status lines, or the auto-arm claim, so they are not applicable.
- The reclaim path is intentionally narrow. It removes only symlink entries inside a lock directory, never arbitrary files, and it deliberately does not re-apply the acquisition path's mid-claim grace: `fm_lock_try_acquire` already evaluated that grace against the untouched lock and returns early while a live pid holds it, but its own failed steal clears the known files and refreshes the lock mtime, which would make every dead claim look mid-acquire to a second evaluation.
- Defect 2 is specific to the Claude Stop hook. Defect 1 is in the harness-agnostic arm and watcher path.

## Tests

Colocated with the existing suites, exercised through the real executables rather than by asserting implementation source.

New coverage in `tests/fm-watcher-lock.test.sh`:

- an attach target that dies during verification is never reported as a live cycle
- a signalled watcher exit is reported with a concrete, non-empty reason
- an attached chain ending with no holder results in a takeover, not a blind return
- repeated arming converges on exactly one live cycle, and reconverges after the holder is stopped

New coverage in `tests/fm-claude-stop-autoarm.test.sh`:

- a claim whose owner pid is dead is reclaimed rather than honoured
- a claim with a live owner is still honoured, so reclaiming never becomes stealing

Two existing tests asserted the previous contract, that an attached arm fails loudly and leaves the home unsupervised, and now assert the takeover instead.

`bin/fm-lint.sh` and `bin/fm-doc-audience-check.sh` are clean, and the affected suites pass with zero failures:

```text
FM_TEST_SUMMARY total=5 failed=0 skipped_gate=0
```

Documentation is updated in `docs/watcher-continuity.md`, which owns the arm-layer cycle contract, with the measured evidence recorded in `docs/verification/supervision.md` and cross-references updated in `docs/architecture.md`, `docs/scripts.md`, and the Claude and Grok supervision protocols.


kunchenguid:

PR #1480, opened 2026-08-01
feat(bin): resolve delivery per task and add the no-mistakes-prod-only project policy

kunchenguid:
## Problem

Delivery mode and yolo were project-wide. `bin/fm-brief.sh` and `bin/fm-spawn.sh` each resolved them from `data/projects.md` independently, so a task that had to ship differently could only be steered by hand-editing its brief while its recorded state kept the project default, and the later lifecycle followed a contract the worker was never given.

Reproduced first through the public interfaces in an isolated home: with a project registered at the legacy default, an adjusted brief said direct-PR while the dispatched task recorded `mode=no-mistakes yolo=off`. Three separate layers: the intake selection was never recorded, the project-default lookup was split across two callers, and the durable task record is what teardown, local merges, and the fleet view consume.

## What this adds

`bin/fm-task-delivery.sh` is the single owner of one task's concrete delivery resolution, recorded once at intake in `state/<id>.delivery` and read unchanged by brief generation and dispatch:

- a task-specific captain instruction naming a concrete mode and yolo value, stronger than any project default;
- an explicit surface classification that resolves the new conditional project policy;
- exact-value validation, atomic publication through an exclusive link, and refusal of malformed, partial, conflicting, duplicate, corrupted, symlinked, wrong-identity, scout-inapplicable, and secondmate-inapplicable input before any brief, task copy, home, endpoint, or metadata exists.

`no-mistakes-prod-only` is a fourth registry policy for a remote-backed project: internal-only tooling, automation, contributor or operator process, and release or submission work ships direct-PR, while product-facing, mixed, and uncertain work runs the pipeline. It resolves per task from firstmate's explicit classification, never from file location or project name, and a ship task on such a project refuses until that classification exists. It is also the stated default for a newly added or created remote-backed project, while existing entries keep the meaning they already have and are never migrated, and yolo stays off unless the captain says otherwise.

Generic consumers follow: secondmate seeding preserves the policy verbatim and prepares the clone for the pipeline path it may take, batch dispatch resolves each task's own record, and teardown clears the record with the rest of the task's state.

## Backward compatibility

An omitted task record resolves exactly as before, including `bin/fm-project-mode.sh`'s own warning and safe default for an unregistered project. A registry line with no bracket keeps its legacy `no-mistakes off` meaning; the parser never reinterprets an existing record.

## Verification

- `tests/fm-task-delivery.test.sh` (new): every mode and both yolo values, all three surface classifications, project-default fallback, brief-and-metadata agreement, refusals before any mutation, policy drift, scout and secondmate refusal, batch isolation, local-only containment, new-project default registration, and secondmate seeding.
- Mutation-checked: reverting either consumer to the project lookup, or dropping the conditional-policy refusal, fails the suite.
- `bin/fm-test-run.sh --changed --base origin/main`: 58 suites, 1180 assertions, 0 failures.
- `bin/fm-lint.sh` and `bin/fm-doc-audience-check.sh` clean.

One behavior change worth noting: a brief's definition of done now says "This task ships ...", because the mode is the task's resolution rather than the project's.


kunchenguid:
Superseded by captain decision 2026-08-02 after design scout `fm-delivery-explicit-mode-design-r1`.

We are **not** continuing this PR. A fresh implementation will land the revised contract:
- ship-scoped required `--mode`/`--yolo` at brief+spawn with brief/spawn agreement check
- keep `fm-project-mode.sh` only for mechanical fleet-sync/home-seed consumers
- `no-mistakes-prod-only` via thin parser mapping + tracked AGENTS/skill text
- no `.delivery` control plane

Salvage for the replacement: project-management skill prose and test scenarios only.
Report: will ship as `fm-delivery-explicit-mode-impl-r1`.

PR #1481, opened 2026-08-01
fix(bin): tolerate an operator's local treehouse.toml during fleet sync

sparkus:
## Problem

An untracked `treehouse.toml` at a clone's root permanently blocks fleet sync. `dirty_status` in `bin/fm-ff-lib.sh` treats any porcelain entry as dirt, so every sync reports the clone `STUCK: ... N commits behind` and it silently falls further behind for as long as the file exists.

Anyone using [Treehouse](https://github.com/kunchenguid/treehouse) to manage worktrees hits this on every clone that has a local pool config. Observed in practice across four project clones, all steadily drifting behind with no way to recover short of deleting a file the operator owns or adding a global ignore.

## Change

- Add `TOLERATED_UNTRACKED_LOCAL_CONFIG` — a narrow, exact-match allowlist for untracked root paths that are known operator-owned machine configuration — alongside the existing secondmate seed-marker exception.
- Route `bin/fm-fleet-sync.sh` through the shared `fm-ff-lib.sh` dirty-status gate rather than its own duplicated `default_branch`/dirty logic, so the fleet sync and the secondmate fast-forward sweep agree on what counts as dirt.

The match is on the **complete porcelain record**, so `treehouse.toml.bak`, `sub/treehouse.toml`, or a tracked-and-modified `treehouse.toml` all still read dirty and still block the fast-forward. Nothing is forced, stashed, or discarded, and a genuinely dirty tree is still refused — prime directive #3 is untouched.

## Tests

Added to the existing suites, all passing on a clean `upstream/main` base:

- `tests/fm-fleet-sync.test.sh` — 27 checks
- `tests/fm-secondmate-sync.test.sh` — 23 checks

New coverage: the tolerance applies (T15), it does not mask real dirt, and the pre-existing seed-marker tolerance is unchanged (T16/T17).

Branched from current `main`; contains only this fix.

### DIFF
--- a/bin/fm-ff-lib.sh
+++ b/bin/fm-ff-lib.sh
@@ -17,14 +17,20 @@
 # A tracked-files fast-forward never touches the gitignored operational dirs
 # (data/, state/, config/, projects/, .no-mistakes/), so it cannot disturb a
 # secondmate's backlog, projects, or in-flight work.
-# The seeded .fm-secondmate-home identity marker is gitignored too; the local
-# sync tolerates only that marker during the one-time upgrade of pre-ignore
-# linked-worktree homes.
+# The shared dirt gate preserves an exact untracked root treehouse.toml operator
+# machine configuration while every other porcelain record blocks fast-forward.
+# The seeded .fm-secondmate-home identity marker is gitignored too; local sync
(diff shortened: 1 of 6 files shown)


kunchenguid:

PR #1485, opened 2026-08-01
fix(codex): anchor trusted hook payload root

coreldh:
## Intent

Prevent a trusted Codex hook declaration from being retargeted to scripts in a writable worker checkout.

## What changed

- Bind Firstmate-launched Codex hooks to the verified parent or secondmate tracked root through `FM_CODEX_HOOK_ROOT`.
- Pass each agent's canonical worktree or home only to the root-scope predicate, preserving the linked-worktree exemption.
- Resolve the cd policy executable from the trusted wrapper directory.
- Add hermetic behavior coverage and update only the documentation that describes this root-anchoring contract.

## Scope

The manifest generator, stored manifest, spawn/update enforcement, generator-candidate ignores, and manifest-only tests and documentation are deliberately absent.

This PR is one commit across 11 files: 258 insertions and 14 deletions.

## Validation

- RED before GREEN: the pre-fix fixture executed a substituted worktree payload; the fixed fixture refused it and executed the trusted payload.
- Falsification: neutralizing the Stop anchor made the dedicated regression fail; restoring it returned the test to green.
- Focused affected suites: 5 passed, 0 failed.
- Full suite: 105 scripts, 102 passed and 3 failed, with 24 expected gate skips.
- All three full-suite failures reproduce identically on detached `origin/main` `8c21b10`: `fm-backend-orca`, `fm-session-start`, and `fm-teardown`; the candidate adds no new failures.
- Coverage inventory: 105 total, 24 parallel, 70 serial, 11 Herdr.
- Shell lint, documentation audience check, JSON validation, and `git diff --check` passed.
- Live Codex trust-dialog behavior remains `NOT_VERIFIABLE`; the regression is hermetic and does not modify the operator trust store.

No merge is requested or performed.


karotkriss:
@coreldh no-mistakes is required for submission. you can go ahead an tell firstrmate to run no-mistakes on the PR in place

coreldh:
Superseded by https://github.com/kunchenguid/firstmate/pull/1537.

coreldh:
Thanks Kriss, and sorry the close landed before this reply. Bad timing on my side, not a brush off.

We did try in place first. That branch's no-mistakes state was stuck at blocked_remote_rewritten, because the earlier force push (the reduction you asked for) left the recorded head out of sync with the fork. From there the tool offered no recovery path, so a run in place could not start at all.

Rather than hand edit the state pointer to get around that refusal, I raised it fresh through no-mistakes on a clean branch: #1537. Same fix, genuine pipeline signature. It is 14 files rather than 11, and the three extra are documentation lines this change makes false, one of which claimed the Codex hooks file is unchanged. Zero code movement from what you reviewed, verified by blob hash.

One thing you may want either way: getting stuck like that looks reachable for any contributor who force pushes after a review, which is a normal thing to be asked to do. Happy to write it up as an issue with a reproduction if that is useful.

PR #1521, opened 2026-08-02
fix(bin): open exactly one decision per status line

Ndidi:
## The defect

`status_open_decisions` in `bin/fm-classify-lib.sh` folds a task's append-only status log into the set of decisions still open. A `needs-decision:` or `blocked:` line opens a keyed decision, and only a `resolved:` or `captain-held:` line carrying that same key closes it. `bin/fm-decision-hold.sh` then refuses to complete or verify while any open key sits outside the reviewed inventory.

Two line shapes broke that fold, both silently.

**1. An unresolvable key dropped the entire line.** `_fm_decision_key` returned non-zero when a `[key=...]` token could not be parsed as a slug, and both folds did `continue`. So this:

```
needs-decision [key=bad key]: which storage backend?
```

opened no decision at all. The escalation disappeared from the durable set and the completion gate passed as though nothing were pending. A malformed `resolved:` line likewise closed nothing, leaving no trace of the attempt. Losing a decision outright is the worse of the two failures here.

**2. Two key tokens bound the decision to whichever came first.**

```
needs-decision [key=one] [key=two]: pick a rollout order
```

folded to `one`, with nothing to tell the author that the second key had been ignored.

### The rule that is working as designed

Only the prefix before the first colon is parsed for a key. A `[key=...]` token after the colon is summary prose, so this opens a decision under `default`:

```
needs-decision: half or full framing [key=phone-framing]
```

That is deliberate and stays unchanged. A summary may legitimately read `should docs mention [key=prose]?`, and nothing in the line distinguishes prose from intent once past the colon. There is an existing test asserting exactly this.

It was, however, completely silent, and it is the shape that motivated this change. An author who writes one correctly keyed line and later writes a second line with the token after the colon ends up with a `default` record sitting beside their real keys. The gate then rejects `default` as outside the reviewed inventory, and nothing explains why. The only remedy was to know, by folklore, to append a bare unkeyed `resolved:` after every escalation. This PR keeps the parsing rule and removes the silence.

## Reproduction

```console
$ . bin/fm-classify-lib.sh
$ printf 'needs-decision [key=bad key]: which storage backend?\n' > /tmp/t.status
$ status_open_decisions /tmp/t.status
$ # before: no output at all, the decision is gone and the gate passes
```

```console
$ printf 'needs-decision [key=one] [key=two]: pick a rollout order\n' > /tmp/t.status
$ status_open_decisions /tmp/t.status
one	needs-decision	pick a rollout order
$ # before: silently bound to the first token
```

## The fix

The invariant is now explicit: **every non-blank `needs-decision:` or `blocked:` line opens exactly one record.**

- A resolvable key opens one decision under that key. Unchanged.
- No key token opens one decision under `default`. Unchanged.
- A line the parser cannot resolve to exactly one slug (two `[key=...]` tokens, an unterminated token, or a value that is not a slug) opens one decision under a reserved `!unresolved-key-<line>` key, and reports the file and line on stderr.

Refusing beats guessing for the ambiguous cases. Taking the first token binds the decision to whichever one happened to be typed first. Falling back to `default` reopens whatever an earlier bare `resolved:` had closed. Both are silent, and neither is recoverable by an author who is never told.

`!` can never appear in a valid slug, so `validate_slug` in `bin/fm-decision-hold.sh` rejects it. The reserved record therefore can never acquire an inventory entry, and `complete` and `verify` reject it by name until the offending status line is fixed. The line number is part of the key so that two malformed lines in one file stay distinct instead of overwriting each other; the status log is append-only, so the number is stable across folds.

The activity fold (`_fm_status_open_activities_stream`) deliberately keeps dropping such lines. Those records are supersession evidence rather than authoritative state, so both directions of a drop are conservative, and a reserved per-line key would strand a phase that no later event could ever close. The reasoning is recorded inline at both call sites.

`bin/fm-brief.sh` now states the token position rule in the generated status protocol, and `.agents/skills/decision-hold-lifecycle/SKILL.md` points at it from the gate failure, so the rule is discoverable at both the moment of writing and the moment of failing.

## Tests

`test_decision_key_grammar` in `tests/fm-watch-triage.test.sh`, alongside the existing key-grammar assertions:

- a keyed line opens exactly one record, under its own key
- a keyed line opens no `default` record beside it
- an unkeyed line opens exactly one record, under `default`
- a two-key line opens exactly one record, under neither key nor `default`, and emits a diagnostic naming the file and line
- a malformed key opens exactly one record instead of vanishing
- two malformed lines stay distinct, and a malformed `resolved:` closes neither them nor an unrelated real decision
- the end-to-end shape: a token after the colon stays prose under `default`, and the earlier keyed decision is untouched by it

Every case asserts the exact record count, not just the presence of a key, so a regression that adds or loses a record fails rather than passing on a substring match.

## Compatibility note

A status log that already contains a line with an unparsable key will now fail the completion gate where it previously passed. That is the intended correction, since such a log does contain an unaccounted decision, but it is a behaviour change on existing data rather than a pure bug fix.

## Verification

`bin/fm-lint.sh` and `bin/fm-doc-audience-check.sh` are clean. The full suite passes apart from three suites that fail identically with this branch's tree removed (`fm-bootstrap`, `fm-pi-watch-extension`, `fm-test-run`), so none of them is caused by this change; the first looks environment dependent, and the second reproduces on elapsed wall time alone rather than test ordering.

### DIFF
--- a/.agents/skills/decision-hold-lifecycle/SKILL.md
+++ b/.agents/skills/decision-hold-lifecycle/SKILL.md
@@ -31,6 +31,7 @@ Bearings reads the resulting structured state and must never compensate by scrap
 2. Inventory only genuine unresolved choices that require the captain.
 3. For each choice, choose a stable key and use the script's `hold` command with a concise title, reason, and repository.
 4. Run the script's `complete` command with the full unresolved-key inventory for that review pass.
+   When the gate reports an open decision outside the reviewed inventory, read the status line that opened it before adding the key: a `[key=<slug>]` token written after the colon is summary prose, so that decision is filed under `default`, and a `!unresolved-key-<line>` entry means the named line resolves to no single key at all.
 5. Relay the choices to the captain as decisions from Bearings' Captain's Call section under `AGENTS.md` section 9; do not use the word hold in captain chat.
 6. After the captain decides, record dependent work with normal tasks-axi commands and block it by the hold identity.
 7. Put the captain's exact durable decision in a file and use the script's `resolve` command with every routed task.
--- a/bin/fm-brief.sh
+++ b/bin/fm-brief.sh
@@ -278,8 +278,10 @@ The report is the only thing that survives, so anything worth keeping must be in
    treating it as a possible wedge. Use \`blocked:\` when you are stuck and need help.
 5. If you hit the same obstacle twice, append \`blocked: {why}\` and stop; firstmate will help.
 6. If a decision belongs to a human (product choices, destructive actions),
-   append \`needs-decision: {summary of options}\` and stop. Firstmate will reply with the decision.
-   When firstmate replies or a blocker clears and you resume, append \`resolved: {how it was decided or unblocked}\` (add the same \`[key=<slug>]\` if you opened it with one) so the decision or blocker is durably closed and does not keep resurfacing.
+   append \`needs-decision [key=<slug>]: {summary of options}\` and stop. Firstmate will reply with the decision.
+   The \`[key=<slug>]\` token must sit before the colon: anywhere after it is read as summary prose, the
+   decision is filed under \`default\` instead, and your later keyed \`resolved:\` line will not close it.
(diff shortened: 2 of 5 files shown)


Ndidi:

PR #1553, opened 2026-08-02
fix(bin): let an attached watcher arm take over instead of alarming

Coutinho072:
Only the arm that forks a watcher holds its stdout, so an attached arm can never read the wake reason of the cycle it follows: a healthy wake close and a death look identical to it. Under the Claude Stop-owned protocol the successor arm starts only at the next Stop, far beyond the attached arm's confirmation window, so every attached arm ended in a false "FAILED - cycle ended without an actionable reason". Worse, the alarm's repair text asked for another manual arm, which created another attached arm and closed a feedback loop.

Fix: an attached arm that outlives its followed cycle takes over as the next owner instead of alarming. The false-alarm reproduction from the investigation is included as a regression test.

Observed live on 2026-08-02 (4+ consecutive false FAILED alarms with a fresh liveness beacon each time); root-cause investigation and scratch-home validation preceded the fix.


kunchenguid:

<sub>Noted for firstmate#1553 at `a7c9880b`.</sub>

Coutinho072:
Closing this without merging.

Upstream commit 88b2a94 ("correct session lock and attached watcher supervision", #1545) already fixes the same false attached-arm alarm, through an identity-bound delivery-log mechanism. It landed on main and touches the same files this branch changes.

Merging this as written would conflict with, and partly regress, that tested mechanism. The upstream fix is the one to keep.

Thanks to whoever picked it up.

PR #1555, opened 2026-08-03
feat(bin): publish the Captain's Log machine projection for Engineering

0ldtimer:
## Intent

Publish the implementation-ready FirstMate Engineering Machine Projection: a versioned backend-neutral Captain's Log interface with opaque Cycle/Build/Scope and FirstMate correlation; continuous Scope, qualitative Hill, blocker, evidence, verification, and outcome reports; supervisor-owned guarded ShapeUp submission with protected credentials and durable replay; existing decision holds as the sole Captain-attention lifecycle; revision-bound Build Review across every active mission and latest required evidence; and optional exact bounded read-only session inspection. Preserve rejected records, distinguish typed stale/unavailable/conflict outcomes, prevent credential leakage, keep session output non-authoritative, and ship all four implementation units with fixtures, documentation, and tests. Focused and affected regression suites pass and lint is clean. The full 111-script local suite passed 94 scripts; 11 real-Herdr gates lacked their required default session and six unrelated environment-sensitive baseline suites also failed, while every new feature suite passed.

## What Changed

- Added the versioned Captain's Log interface: `bin/fm-captains-log-projection.sh` (`capabilities`, `snapshot`, `intent`, optional `inspectSession` over a bounded JSON stdin contract), `bin/fm-mission.sh` for Build-revision-bound mission accept/supersede/defer, `bin/fm-report.sh` for immutable Scope, qualitative Hill, blocker, evidence, verification, and outcome records, and the shared `bin/fm-engineering-lib.sh` (privacy-safe identities, credential vocabulary, checked atomic writes, liveness-safe `mkdir` mutex). Rejected and schema-invalid records are retained and non-executable, partially readable stores and failed writes return typed `projection_unavailable` / `record_not_durable` / `outcome_not_durable` results, and acceptance order comes from a FirstMate-assigned `acceptanceSequence` allocated under the lock rather than from crewmate-supplied timestamps.
- Wired the surrounding surfaces: `bin/fm-shapeup-client.sh` submits Scope and Hill reports through one operator-configured closed transport under a bounded timeout, reading the workspace credential fresh from an owner-only file and passing it only via the child environment, with typed `shapeup_unavailable` outcomes and durable replay; `bin/fm-decision-hold.sh` gains `id`, `project`, and `status --json` so existing decision holds remain the sole Captain-attention lifecycle and a hold mid-resolution reports `retained` instead of overwriting the retry identity; Build Review acceptance is revision-bound across every active mission and fails closed while any bearing record is isolated.
- Added a bounded read-only `--json` capture mode to `bin/fm-peek.sh` that redacts credentials over reconstructed logical lines (wrap stitching, `=` vs `:` separator rules, row boundaries preserved unless a redaction straddled a wrap), embedded the projection snapshot in `bin/fm-fleet-snapshot.sh` with graceful `freshness: "unavailable"` degradation, and shipped the correlation-envelope and compatibility-matrix fixtures, `docs/captains-log-machine-projection.md`, `docs/shapeup-client.md`, updated architecture/configuration/scripts/decision-hold docs, and eight portable test suites (all passing, lint clean).

## Risk Assessment

✅ Low: This is the first round with no warning-or-above findings: every correctness, durability, and credential-leak defect from the five prior rounds is fixed and re-verified here by direct reproduction, the redaction candidate selection is now safe by construction rather than by comparison, and what remains is duplicated rule text and one vacuous test assertion.

## Testing

Ran the eight focused feature suites named in docs/captains-log-machine-projection.md plus the seven regression suites this change touches (fleet-snapshot embedding, decision-hold lifecycle, documentation inventory, and the fm-peek consumers) — all pass with no skips, since jq and tasks-axi are both available. Because passing suites alone do not show the interface an integrator uses, I also wrote and ran an end-to-end walkthrough that drives the real CLIs against a throwaway FM_HOME and captured the full request/response transcript: capability negotiation; a revision-bound mission with its evidence contract and reporting instructions; all seven report kinds including a backward Hill judgment; numeric progress refused as invalid_hill and a cross-Build report refused as cross_build, both retained on disk with status rejected; a Captain Call raised on a durable hold, acknowledged without resolving it, then resolved and closed through fm-decision-hold.sh, with identity_conflict on intentId reuse and condition_not_found for an unknown condition; a Build Review packet that becomes ready only once terminal and evidence-verified, accepts with delivery/buildCloseout/cycleClose all false, replays exactly, then returns review_revision_conflict and stale_packet once a second mission binds a different Build revision; guarded ShapeUp submission where the transport receives argc=0 and the token only via the environment, with credential_permissions on a non-owner-only file and shapeup_unavailable on a missing transport while the prior durable outcome still replays; bounded read-only session inspection returning authoritative false with a pane-wrapped Bearer value and an inline SERVICE_TOKEN both redacted (verified by a sliding-window fragment scan); an isolated hand-edited record failing Build Review closed with review_records_isolated while every other operation stays up; and the projection embedded in the fleet snapshot at freshness fresh. This change ships no UI, HTML, or rendered surface — its end-user surface is a stdin/stdout JSON CLI contract — so the reviewer-visible evidence is the CLI transcript and the persisted store state rather than screenshots. I did not run lint, since the task forbids linters, so the intent's "lint is clean" claim is unverified here; I also did not run the full 111-script suite, because .no-mistakes.yaml explicitly reserves broad regression for CI and scopes local Test to intent-targeted validation. Transient demo homes under /tmp were removed and the worktree is clean.

<details>
<summary>Evidence: End-to-end CLI transcript: all ten acts of the Captain&#39;s Log machine projection driven against a real FM_HOME</summary>

```text


═══ 1. Versioned backend-neutral interface: what does this build advertise? ═══

$ echo '{...capabilities}' | fm-captains-log-projection.sh
{
  "accepted": true,
  "acceptedIntents": [
    "acknowledgeCondition",
    "acceptBuildReview",
    "resolveCondition"
  ],
  "capabilities": {
    "core": "available",
    "sessionInspection": {
      "mode": "bounded-read-only",
      "preferredBackend": "herdr",
      "required": false,
      "status": "negotiated"
    },
    "shapeUp": {
      "required": false,
      "status": "negotiated"
    }
  },
  "capturedAt": "2026-08-01T12:00:00Z",
  "operations": [
    "capabilities",
    "snapshot",
    "intent",
    "inspectSession"
  ],
  "schemaVersion": "fm-captains-log-projection.v1",
  "sourceRevision": "capabilities-v1"
}


═══ 2. Revision-bound Engineering mission with a required evidence contract ═══

$ fm-mission.sh accept - <<< <mission-42 bound to build-8:r7>
{
  "accepted": true,
  "mission": {
    "acceptedBuildRevision": "build-8:r7",
    "evidenceRequirements": [
      "automated-tests"
    ],
    "missionId": "mission-42",
    "reporting": {
      "acceptedKinds": [
        "scope_discovery",
        "scope_revision",
        "hill_judgment",
        "blocker",
        "evidence",
        "verification_instructions",
        "outcome"
      ],
      "continueUntilTerminal": true,
      "establishInitialScopes": true
    },
    "sourceRevision": "fbe37ad57bc42906b4d737165da22a73b509eb8a76798987c7c54aaad03c4474",
    "state": "accepted"
  }
}


═══ 3. Continuous reporting: Scope, qualitative Hill, blocker, evidence, verification, outcome ═══

$ fm-report.sh append -  # scope_discovery
{
  "acceptanceSequence": 1,
  "accepted": true,
  "kind": "scope_discovery",
  "scope": {
    "judgment": "A distinct integration risk emerged.",
    "title": "Context fencing"
  }
}

$ fm-report.sh append -  # scope_revision (Scope is continuous)
{
  "acceptanceSequence": 2,
  "accepted": true,
  "kind": "scope_revision",
  "scope": {
    "change": "split",
    "reason": "Fencing separated from the transport rewrite."
  }
}

$ fm-report.sh append -  # hill_judgment, uphill/forward
{
  "accepted": true,
  "hill": {
    "judgment": "The unknowns are naming themselves.",
    "movement": "forward",
    "phase": "uphill"
  }
}

$ fm-report.sh append -  # hill_judgment may move BACKWARD as learning changes the picture
{
  "accepted": true,
  "hill": {
    "judgment": "A hidden coupling put the Scope back uphill.",
    "movement": "backward",
    "phase": "uphill"
  }
}

$ fm-report.sh append -  # numeric progress is REFUSED: no count can determine Hill position
{
  "accepted": false,
  "error": "invalid_hill",
  "message": "Hill reports must be qualitative judgments and may move backward",
  "storedStatus": "rejected"
}

$ fm-report.sh append -  # blocker
{
  "accepted": true,
  "kind": "blocker"
}

$ fm-report.sh append -  # evidence against the mission contract, verified
{
  "accepted": true,
  "evidence": {
    "contract": "automated-tests",
    "executionRevision": "exec:r1",
    "producer": "keiko",
    "reference": {
      "kind": "file",
      "mediaType": "text/plain",
      "value": "/tmp/fm-e2e.XHkqYw/projects/task-42/evidence/tests.txt"
    },
    "verification": {
      "instructions": "Run bash tests/fm-report.test.sh and read the summary.",
      "status": "verified"
    },
    "verifier": "firstmate"
  }
}

$ fm-report.sh append -  # verification_instructions for a human verifier
{
  "accepted": true,
  "kind": "verification_instructions"
}

$ fm-report.sh append -  # outcome
{
  "accepted": true,
  "kind": "outcome",
  "outcome": {
    "state": "completed",
    "summary": "Fencing landed with a verified suite."
  }
}


═══ 4. Rejected records are PRESERVED and typed, never silently dropped ═══

$ fm-report.sh append -  # a report correlated to a different Build
{
  "accepted": false,
  "error": "cross_build",
  "storedStatus": "rejected"
}

$ snapshot | .reports[] | {reportId, status, error}   # both rejects survive as history
[
  {
    "error": null,
    "reportId": "r-blocker-1",
    "status": "accepted"
  },
  {
    "error": "cross_build",
    "reportId": "r-cross",
    "status": "rejected"
  },
  {
    "error": null,
    "reportId": "r-evidence-1",
    "status": "accepted"
  },
  {
    "error": null,
    "reportId": "r-hill-1",
    "status": "accepted"
  },
  {
    "error": null,
    "reportId": "r-hill-2",
    "status": "accepted"
  },
  {
    "error": "invalid_hill",
    "reportId": "r-hill-bad",
    "status": "rejected"
  },
  {
    "error": null,
    "reportId": "r-outcome-1",
    "status": "accepted"
  },
  {
    "error": null,
    "reportId": "r-scope-1",
    "status": "accepted"
  },
  {
    "error": null,
    "reportId": "r-scope-2",
    "status": "accepted"
  },
  {
    "error": null,
    "reportId": "r-verify-1",
    "status": "accepted"
  }
]


═══ 5. Decision holds are the SOLE Captain-attention lifecycle ═══

$ fm-report.sh append -  # a consequential report raises one durable Captain Call
{
  "accepted": true,
  "condition": {
    "affectedObjects": [
      "build-8",
      "scope-context"
    ],
    "boundRevisions": {
      "build": "build-8:r7"
    },
    "choices": [
      "defer",
      "retain"
    ],
    "conditionId": "scope-drop",
    "consequenceOfDelay": "Execution stays blocked on an undecided Scope.",
    "creationSequence": 1,
    "evidence": [],
    "explanation": "Dropping the Scope changes the accepted Build outcome.",
    "holdId": "mission-42-decision-scope-drop",
    "lifecycle": "raised",
    "packetRevision": "627ced6aa9fc0283edf5932c9eb53be0674c2d758e157af6970d1f62ddbc17bb",
    "projection": {
      "projectedAt": "2026-08-01T12:00:00Z",
      "status": "projected"
    },
    "recommendation": "Defer the Scope explicitly.",
    "title": "Drop or retain the fencing Scope"
  }
}

$ fm-decision-hold.sh status mission-42 scope-drop --json
{
  "schemaVersion": "fm-captain-call-status.v1",
  "conditionId": "scope-drop",
  "holdId": "mission-42-decision-scope-drop",
  "lifecycle": "raised",
  "packetRevision": "627ced6aa9fc0283edf5932c9eb53be0674c2d758e157af6970d1f62ddbc17bb",
  "durableState": {
    "state": "queued",
    "held": true
  }
}

$ projection intent acknowledgeCondition  # attention only - never resolves the hold
{
  "accepted": true,
  "error": null,
  "outcome": {
    "action": "acknowledgeCondition",
    "implications": {
      "buildCloseout": false,
      "cycleClose": false,
      "delivery": false
    },
    "intentId": "ack-1",
    "recordedAt": "2026-08-01T12:00:00Z",
    "replayed": false,
    "status": "acknowledged"
  }
}

$ fm-decision-hold.sh status mission-42 scope-drop --json  # still held
{
  "schemaVersion": "fm-captain-call-status.v1",
  "conditionId": "scope-drop",
  "holdId": "mission-42-decision-scope-drop",
  "lifecycle": "raised",
  "packetRevision": "627ced6aa9fc0283edf5932c9eb53be0674c2d758e157af6970d1f62ddbc17bb",
  "durableState": {
    "state": "queued",
    "held": true
  }
}

$ tasks-axi add routed-scope-work ... && tasks-axi block routed-scope-work --by <holdId>
routed work is queued and blocked by the Captain Call

$ projection intent resolveCondition  # routes the decision and closes that same hold
{
  "accepted": true,
  "error": null,
  "outcome": {
    "action": "resolveCondition",
    "implications": {
      "buildCloseout": false,
      "cycleClose": false,
      "delivery": false
    },
    "intentId": "resolve-1",
    "recordedAt": "2026-08-01T12:00:00Z",
    "replayed": false,
    "status": "resolved"
  }
}

$ fm-decision-hold.sh status mission-42 scope-drop --json  # durably resolved
{
  "schemaVersion": "fm-captain-call-status.v1",
  "conditionId": "scope-drop",
  "holdId": "mission-42-decision-scope-drop",
  "lifecycle": "resolved",
  "packetRevision": null,
  "durableState": {
    "state": "done",
    "held": false
  }
}

$ snapshot | .conditions[0]
{
  "conditionId": "scope-drop",
  "holdId": "mission-42-decision-scope-drop",
  "lifecycle": "resolved",
  "packetRevision": "627ced6aa9fc0283edf5932c9eb53be0674c2d758e157af6970d1f62ddbc17bb",

... [1938 bytes truncated] ...

r": null,
  "outcome": {
    "action": "acceptBuildReview",
    "implications": {
      "buildCloseout": false,
      "cycleClose": false,
      "delivery": false
    },
    "intentId": "review-1",
    "recordedAt": "2026-08-01T12:00:00Z",
    "replayed": false,
    "status": "captainAccepted"
  }
}

$ projection intent acceptBuildReview  # exact replay returns the prior outcome
{
  "accepted": true,
  "replayed": true,
  "status": "captainAccepted"
}

$ fm-mission.sh accept -  # a second active mission bound to a DIFFERENT Build revision
{
  "accepted": true,
  "acceptedBuildRevision": "build-8:r8",
  "missionId": "mission-43"
}

$ snapshot | .buildReviews[0]  # one packet covers BOTH active missions
{
  "buildRevision": null,
  "buildRevisions": [
    "build-8:r7",
    "build-8:r8"
  ],
  "missionSet": [
    {
      "buildRevision": "build-8:r7",
      "missionId": "mission-42",
      "terminal": true
    },
    {
      "buildRevision": "build-8:r8",
      "missionId": "mission-43",
      "terminal": false
    }
  ],
  "ready": false,
  "revisionConflict": true
}

$ projection intent acceptBuildReview  # conflicting bound revisions are a typed refusal
{
  "accepted": false,
  "error": "review_revision_conflict",
  "message": "Active missions in this Build are bound to conflicting Build revisions"
}

$ projection intent acceptBuildReview  # the superseded packet is typed stale, not silently retried
{
  "accepted": false,
  "error": "stale_packet",
  "message": "Build Review packet changed before Captain acceptance"
}


═══ 7. Supervisor-owned guarded ShapeUp submission with protected credentials ═══

$ fm-shapeup-client.sh capabilities
{
  "accepted": true,
  "capabilities": {
    "accepted": true,
    "capabilities": [
      "scope.write",
      "hill.write"
    ],
    "contractVersion": "shapeup-engineering.v1"
  },
  "schemaVersion": "fm-shapeup-client.v1"
}

$ fm-shapeup-client.sh submit r-scope-1
{
  "accepted": true,
  "error": null,
  "outcome": {
    "authoritativeRevision": "shapeup:r9",
    "idempotencyKey": "fm-report:r-scope-1:612b3e0a6e83888a02a8dae686b9e798682eb5369f32cdd1fcba45a0b501c5cc",
    "recordedAt": "2026-08-01T12:00:00Z",
    "replayed": false,
    "reportId": "r-scope-1",
    "reportRevision": "612b3e0a6e83888a02a8dae686b9e798682eb5369f32cdd1fcba45a0b501c5cc",
    "status": "applied"
  }
}

$ fm-shapeup-client.sh submit r-scope-1   # durable replay, transport not re-invoked
{
  "accepted": true,
  "replayed": true,
  "status": "applied"
}

$ grep -c s3cr3t-workspace-token <(all client output)   # credential leakage check
clean: the workspace credential never appears in client output

$ cat transport-argv.log   # the credential reaches the child only through the environment
argc=0 argv=[] token_seen_in_env=yes
argc=0 argv=[] token_seen_in_env=yes
argc=0 argv=[] token_seen_in_env=yes

$ chmod 644 config/shapeup-token && fm-shapeup-client.sh submit r-hill-1
{
  "accepted": false,
  "error": {
    "code": "credential_permissions",
    "message": "ShapeUp credential file must be owner-only"
  }
}

$ <transport removed> fm-shapeup-client.sh submit r-hill-1   # typed unavailable, not a false success
{
  "accepted": false,
  "error": {
    "code": "shapeup_unavailable",
    "message": "ShapeUp capability negotiation is unavailable"
  }
}

$ <transport removed> fm-shapeup-client.sh submit r-scope-1   # prior durable outcome still replays
{
  "accepted": true,
  "replayed": true,
  "status": "applied"
}

$ snapshot | .shapeUpOutcomes   # ShapeUp stays authoritative for its own facts
[
  {
    "outcome": {
      "authoritativeRevision": "shapeup:r9",
      "idempotencyKey": "fm-report:r-scope-1:612b3e0a6e83888a02a8dae686b9e798682eb5369f32cdd1fcba45a0b501c5cc",
      "recordedAt": "2026-08-01T12:00:00Z",
      "replayed": false,
      "reportId": "r-scope-1",
      "reportRevision": "612b3e0a6e83888a02a8dae686b9e798682eb5369f32cdd1fcba45a0b501c5cc",
      "status": "applied"
    },
    "requestDigest": "612b3e0a6e83888a02a8dae686b9e798682eb5369f32cdd1fcba45a0b501c5cc"
  }
]


═══ 8. Optional, exact, bounded, non-authoritative session inspection ═══

$ cat pane.txt   # what the live 40-column pane actually holds
----------------------------------------| <- 40-column pane edge
running tests...                        |
Authorization: Bearer fm_fake_51QsecretP|
AYLOADabcdefghijklmnop                  |
SERVICE_TOKEN=another-live-secret       |
suite: 8 files, 0 failures              |

$ projection inspectSession (exact mission revision)
{
  "accepted": true,
  "error": null,
  "inspection": {
    "authoritative": false,
    "available": true,
    "capture": {
      "bytes": 101,
      "lines": 20,
      "text": "running tests...\nAuthorization: Bearer [REDACTED]\nSERVICE_TOKEN=[REDACTED]\nsuite: 8 files, 0 failures",
      "truncated": false
    },
    "descriptor": {
      "backend": "tmux",
      "lifecycle": "running",
      "targetId": "firstmate:fm-task-42",
      "taskId": "task-42"
    },
    "mode": "bounded-read-only"
  }
}

$ fragment check: does any 8-char window of the wrapped secret survive?
clean: no 8-character window of the pane-wrapped credential survives redaction

$ projection inspectSession with a STALE expected mission revision
{
  "accepted": false,
  "error": {
    "code": "stale_revision",
    "message": "Mission revision changed before session inspection"
  }
}


═══ 9. A hand-edited record is isolated: Build Review fails CLOSED, everything else stays up ═══

$ printf "{ not json" > data/engineering/reports/r-evidence-1.json

$ snapshot | .invalidRecords   # the broken record stays visible and non-executable
[
  {
    "buildRef": null,
    "cycleRef": null,
    "error": {
      "code": "malformed_record",
      "message": "Record is not a valid JSON object"
    },
    "kind": "report",
    "recordId": "r-evidence-1"
  }
]

$ snapshot | .buildReviews[0]  # never ready while an isolated record bears on it
{
  "isolatedRecords": [
    {
      "error": "malformed_record",
      "kind": "report",
      "recordId": "r-evidence-1"
    }
  ],
  "ready": false,
  "revisionConflict": true
}

$ projection intent acceptBuildReview <current packet>  # refuses with its own typed error,
                                              # ahead of the revision conflict verdict
{
  "accepted": false,
  "error": "review_records_isolated",
  "message": "A stored mission or report record for this Build is invalid, so the packet cannot prove it covers every active mission"
}

$ echo <capabilities> | fm-captains-log-projection.sh   # every other operation is unaffected
{
  "accepted": true,
  "core": "available"
}


═══ 10. The projection is embedded in the read-only fleet snapshot ═══

$ fm-fleet-snapshot.sh | .captains_log_projection
{
  "buildReviews": [
    {
      "buildRef": "build-8",
      "ready": false
    }
  ],
  "capturedAt": "2026-08-01T12:00:00Z",
  "conditions": [
    {
      "conditionId": "scope-drop",
      "lifecycle": "resolved",
      "projection": {
        "projectedAt": "2026-08-01T12:00:00Z",
        "status": "projected"
      }
    }
  ],
  "freshness": "fresh",
  "invalidRecords": [],
  "missions": [
    {
      "acceptedBuildRevision": "build-8:r7",
      "missionId": "mission-42",
      "state": "accepted"
    },
    {
      "acceptedBuildRevision": "build-8:r8",
      "missionId": "mission-43",
      "state": "accepted"
    }
  ],
  "outcomes": [
    {
      "action": "acknowledgeCondition",
      "implications": {
        "buildCloseout": false,
        "cycleClose": false,
        "delivery": false
      },
      "status": "acknowledged"
    },
    {
      "action": "resolveCondition",
      "implications": {
        "buildCloseout": false,
        "cycleClose": false,
        "delivery": false
      },
      "status": "resolved"
    },
    {
      "action": "acceptBuildReview",
      "implications": {
        "buildCloseout": false,
        "cycleClose": false,
        "delivery": false
      },
      "status": "captainAccepted"
    }
  ],
  "reportCount": 12,
  "schemaVersion": "fm-captains-log-snapshot.v1"
}


Demo FM_HOME: /tmp/fm-e2e.XHkqYw
```
</details>
- Evidence: Walkthrough script that produced the transcript (re-runnable: bash e2e-captains-log-walkthrough.sh &lt;repo-root&gt;) (local file: <code>/var/folders/2g/z6s_jsfx13b37_l46phgrnn00000gn/T/no-mistakes-evidence/01KZ296JNYPNXW5DBYWCEZG4PM/e2e-captains-log-walkthrough.sh</code>)
<details>
<summary>Evidence: Persisted store state after the walkthrough: retained rejected records, durable outcomes, closed Captain hold, no credential anywhere on disk</summary>

```text
$ find data/engineering -type f | sort   # durable record store the projection reads
data/engineering/decisions/resolve-1.txt
data/engineering/missions/mission-42.json
data/engineering/missions/mission-43.json
data/engineering/outcomes/ack-1.json
data/engineering/outcomes/resolve-1.json
data/engineering/outcomes/review-1.json
data/engineering/reports.sequence
data/engineering/reports/r-blocker-1.json
data/engineering/reports/r-condition-1.json
data/engineering/reports/r-cross.json
data/engineering/reports/r-evidence-1.json
data/engineering/reports/r-hill-1.json
data/engineering/reports/r-hill-2.json
data/engineering/reports/r-hill-bad.json
data/engineering/reports/r-outcome-1.json
data/engineering/reports/r-outcome-2.json
data/engineering/reports/r-scope-1.json
data/engineering/reports/r-scope-2.json
data/engineering/reports/r-verify-1.json
data/engineering/shapeup-outcomes/r-scope-1.json

$ jq '{reportId,kind,status,rejectedAt,error,hill}' data/engineering/reports/r-hill-bad.json
# a numeric-progress Hill report: refused, but retained verbatim and typed
{
  "error": {
    "code": "invalid_hill",
    "message": "Hill reports must be qualitative judgments and may move backward"
  },
  "hill": {
    "judgment": "Nearly there.",
    "phase": "downhill",
    "progress": 80
  },
  "kind": "hill_judgment",
  "rejectedAt": "2026-08-01T12:00:00Z",
  "reportId": "r-hill-bad",
  "status": "rejected"
}

$ jq '{reportId,kind,status,error,shapeUp:.correlation.shapeUp}' data/engineering/reports/r-cross.json
{
  "error": {
    "code": "cross_build",
    "message": "Report correlation does not match the accepted Build"
  },
  "kind": "scope_discovery",
  "reportId": "r-cross",
  "shapeUp": {
    "buildRef": "build-9",
    "buildRevision": "build-9:r1",
    "cycleRef": "cycle-13",
    "scopeRef": "scope-context"
  },
  "status": "rejected"
}

$ tasks-axi show mission-42-decision-scope-drop --full
# the sole Captain-attention lifecycle, durably closed by the resolveCondition intent
task:
  id: mission-42-decision-scope-drop
  title: Drop or retain the fencing Scope
  state: done
  blocked: no
  blocked_by: none
  held: no
  hold_reason: captain decision pending for scope-drop
  hold_kind: captain
  hold_until: "-"
  kind: captain
  repo: firstmate
  priority: "-"
  created: "-"
  closed: 2026-08-02
  deps: none
  links: none
  body: "Resolution recorded by fm-decision-hold.\nDecision digest: f78efc4e5028a8d184e1570c067e809e57c58365d8d79243fcd92adade3acf0a\nRouted identities: routed-scope-work\n\nCaptain decision:\nDefer the fencing Scope to the next Cycle.\n\nRouted work:- routed-scope-work"

$ grep -ri 's3cr3t-workspace-token' data/ state/ config/shapeup-client.json  # credential never persisted outside the owner-only file
(no matches)
```
</details>
<details>
<summary>Evidence: Captain-attention lifecycle and Build Review refusals (excerpt from the transcript)</summary>

```text
$ projection intent acknowledgeCondition # attention only - never resolves the hold
{"accepted": true, "outcome": {"action": "acknowledgeCondition", "status": "acknowledged",
"implications": {"buildCloseout": false, "cycleClose": false, "delivery": false}}}

$ fm-decision-hold.sh status mission-42 scope-drop --json # still held
{"lifecycle": "raised", "durableState": {"state": "queued", "held": true}}

$ projection intent resolveCondition # routes the decision and closes that same hold
{"accepted": true, "outcome": {"action": "resolveCondition", "status": "resolved"}}

$ fm-decision-hold.sh status mission-42 scope-drop --json # durably resolved
{"lifecycle": "resolved", "durableState": {"state": "done", "held": false}}

$ projection intent resolveCondition # reusing an intentId with different content
{"accepted": false, "error": "identity_conflict"}

$ projection intent acceptBuildReview # conflicting bound revisions
{"accepted": false, "error": "review_revision_conflict"}

$ projection intent acceptBuildReview # superseded packet
{"accepted": false, "error": "stale_packet"}

$ projection intent acceptBuildReview # a hand-edited record makes Review fail CLOSED
{"accepted": false, "error": "review_records_isolated"}
```
</details>
<details>
<summary>Evidence: Credential protection: ShapeUp argv boundary and pane-wrap redaction (excerpt from the transcript)</summary>

```text
$ cat transport-argv.log # the credential reaches the child only through the environment
argc=0 argv=[] token_seen_in_env=yes

$ grep s3cr3t-workspace-token <(all client output)
clean: the workspace credential never appears in client output

$ cat pane.txt # what the live 40-column pane actually holds
----------------------------------------| <- 40-column pane edge
running tests... |
Authorization: Bearer fm_fake_51QsecretP|
AYLOADabcdefghijklmnop |
SERVICE_TOKEN=another-live-secret |
suite: 8 files, 0 failures |

$ projection inspectSession (exact mission revision)
{"accepted": true, "inspection": {"available": true, "authoritative": false,
"mode": "bounded-read-only",
"capture": {"text": "running tests...\nAuthorization: Bearer [REDACTED]\nSERVICE_TOKEN=[REDACTED]\nsuite: 8 files, 0 failures"},
"descriptor": {"backend": "tmux", "lifecycle": "running", "targetId": "firstmate:fm-task-42", "taskId": "task-42"}}}

$ fragment check: does any 8-char window of the wrapped secret survive?
clean: no 8-character window of the pane-wrapped credential survives redaction
```
</details>

## Pipeline

<details>

🔧 Fix: fix(engineering): close redaction, replay, ordering, and durability gaps
6 issues (4 warnings, 2 infos) still open:
- ⚠️ `bin/fm-peek.sh:104` - The new colon rule `s/($names)[[:space:]]*:[[:space:]]*[^[:space:]]+/\1: [REDACTED]/g`, combined with the `contiguous` detection at line 80 (`row ~ (&#34;(&#34; names &#34;)[[:space:]]*[=:]$&#34;)`), fires on ordinary inert pane output because the vocabulary matches bare words like `tokens`, `password`, and `access-key`. Reproduced by running the current `redact_capture` over a synthetic pane: rows `ok 12 usage summary for this run    tokens:` (pane-wide, inert) and `ok 13 unrelated assertion on next line` come back as the single line `ok 12 usage summary for this run    tokens: [REDACTED] 13 unrelated assertion on next line` — the next row&#39;s `ok` is consumed as the credential value, the redaction suppresses the span re-split, and two unrelated pane events are merged into one. A standalone `Total tokens: 4821 in this run` becomes `Total tokens: [REDACTED] in this run`, and `Retrieved access-key: none configured` / `Password: not required for this step` merge into one mangled line. Two things this contradicts: docs/captains-log-machine-projection.md still states &#34;A row that merely fills the pane without wrapping keeps its own line, so unrelated events are never merged into one&#34;, and tests/fm-engineering-lib.test.sh:284 pins `&#34;Total tokens: 4821 in this run, so we split the Scope.&#34;` as benign for record scanning while the pane path now redacts exactly that string. This fails closed (over-redaction, not a leak), so it is not a security regression — but it degrades the readability that is the entire point of session inspection. Narrowing the colon form is a product tradeoff: require the matched name to be a whole token bounded by start/whitespace/`-`/`_` rather than allowing the `[A-Za-z0-9_]*` wrapper, restrict the colon form to compound names (`api-key:`, `x-auth-token:`) and drop bare `token:`/`key:`, or additionally require a secret-shaped value.
- ⚠️ `bin/fm-report.sh:99` - `next_acceptance_sequence` reads `$SEQUENCE_FILE`, computes `next=$((current + 1))`, and writes it back with no mutual exclusion, while `fm-report.sh append` has no lock at all (unlike the projection&#39;s intent path, which uses a `mkdir` mutex precisely because concurrency is expected). Two crewmates appending into the same FM_HOME concurrently both read N and both stamp `acceptanceSequence: N+1`. In `build_reviews`, `def authoritative: sort_by(.acceptedAt // &#34;&#34;,.acceptanceSequence // 0,.reportId) | last` then falls through to `.reportId`, which is crewmate-supplied — reopening the exact ordering hole this fix closes, just behind a concurrency window instead of a same-second window. The code comment at line 71 and docs/captains-log-machine-projection.md both assert the stronger property (&#34;two reports accepted inside the same wall-clock second are still ordered by FirstMate rather than by anything the crewmate supplied&#34;), which does not hold under a concurrent append. Secondary: the same comment claims &#34;an interrupted append can only skip a number, never reuse one&#34;, but if `$SEQUENCE_FILE` is missing or corrupt, `acceptance_sequence_floor` recomputes the max from `xargs -0 cat` over the report store, and an unreadable report file is silently skipped by `cat 2&gt;/dev/null`, yielding a floor below the true max and reusing a sequence. Guard the read-increment-write with the same `mkdir`-based mutex the projection uses.
- ⚠️ `bin/fm-captains-log-projection.sh:321` - `partition_records` now drops a schema-invalid mission from `records`, so `build_reviews`&#39; `select((.state // &#34;accepted&#34;) == &#34;accepted&#34;)` never sees it and the packet&#39;s `missionSet` silently shrinks. tests/fm-build-review.test.sh:147 pins this: with `mission-44` isolated, `missionSet` is asserted to be exactly `[&#34;mission-42&#34;,&#34;mission-43&#34;]`. Before this commit the projection refused outright (`projection_unavailable`); now it proceeds fail-open. `intent_response` gates `acceptBuildReview` only on `.revisionConflict`, `.packetRevision`, and `.ready` (line 554) and never consults `snapshot.invalidRecords`, so if the isolated record were the only mission still non-terminal, the packet reports `ready: true` and the Captain can accept a Build Review that does not cover every active mission — contradicting &#34;One Build Review packet covers every active accepted mission of a Cycle and Build&#34; and &#34;The packet becomes ready only when each mission is terminal&#34;. The same applies to an isolated accepted `evidence` or `outcome` report, whose exclusion lets an older record decide readiness. The isolation itself is what was asked for; what needs a decision is whether readiness/acceptance should be suppressed (or the packet should carry an `isolatedRecords` flag) while any mission- or report-kind record is in `invalidRecords`.
- ⚠️ `bin/fm-report.sh:214` - `command_project` now short-circuits with `printf &#39;retained: ...&#39;` and `return 0` when the hold already carries a resolution record (bin/fm-decision-hold.sh:407) — correct for the replay fix, but it is indistinguishable from success to its callers. `project_condition_packet` redirects stdout to /dev/null and branches only on exit status, so it stamps `.condition.projection={status:&#34;projected&#34;,projectedAt:...}` even though the hold body was never updated with the new packet revision. Failure scenario: a `resolveCondition` partially fails after `resolve` wrote the retry-identity body (hold still queued); a crewmate then appends an updated report for the same condition; `hold` succeeds, `project` returns `retained`, and the durable record claims the packet landed. docs/captains-log-machine-projection.md guarantees the opposite: &#34;Every projected condition carries its `projection` state, so a held condition that never received its packet is visible to the Captain.&#34; Either have `command_project` return a distinct status for the retained case so `project_condition_packet` can record `degraded` (which `heal_condition_projection` already retries), or add an explicit `retained` projection state — the choice is user-visible, so it needs your call.
- ℹ️ `bin/fm-shapeup-client.sh:98` - Two rough edges in the perl timeout fallback. (1) `if (!$pid) { setpgrp(0, 0); exec @ARGV }` has no `or exit` guard: a transport with a broken shebang makes `execve` fail with ENOENT, `exec` returns false, and the child falls through into the parent&#39;s code — installing an ALRM handler and calling `waitpid $pid, 0` with `$pid == 0`, which returns immediately and exits with an arbitrary `$? &gt;&gt; 8`. The `-f`/`-x` precheck at line 76 does not catch a bad interpreter line. Add `exec @ARGV or exit 127`. (2) `exit($? &gt;&gt; 8)` reports 0 when the transport is killed by a signal, so a transport that emitted a complete JSON object and was then SIGKILLed is treated as a successful submission and journalled. Map a signal death to a non-zero status (e.g. `exit($? &amp; 127 ? 126 : $? &gt;&gt; 8)`).
- ℹ️ `bin/fm-shapeup-client.sh:105` - With no `timeout`, `gtimeout`, or `perl` on PATH, `transport_call` now returns 8 and every submission fails with `shapeup_unavailable`/`transport_unbounded`, where it previously ran unbounded and worked. docs/shapeup-client.md documents this deliberately (&#34;a host with no way to bound a child process refuses the call instead of running it unbounded&#34;) and the fallback chain covers macOS and mainstream Linux, so this is a noted tradeoff rather than a defect — recording it so the new runtime dependency is visible if the client is ever run in a minimal container image.

🔧 Fix: fix(engineering): narrow colon redaction, lock sequence, fail-closed review
5 issues (2 warnings, 3 infos) still open:
- ⚠️ `bin/fm-peek.sh:111` - The narrowed colon rule `s/(^|[[:space:]])($labels)[[:space:]]*:[[:space:]]*[^[:space:]]{6,}/` correctly stops redacting inert telemetry, but it gives up two shapes the previous revision covered. Reproduced against the current `redact_capture`: `apitoken: fm_fake_gluedvalue123` and `dbpassword: fm_fake_gluedvalue456` come back unchanged, because `fm_eng_credential_label_ere` (bin/fm-engineering-lib.sh:82) requires the vocabulary word to be a whole separator-delimited component (`([A-Za-z0-9]+[-_])*(ALT)([-_][A-Za-z0-9]+)*`) and a glued lowercase name has no separator; the previous `($names)` form, which kept the `[A-Za-z0-9_]*` wrapper, matched them. `Password: hunt3` is also unchanged now, because the operand is below the `{6,}` bound. The `=` form still covers both (`apitoken=[REDACTED]` verified), so the gap is colon-specific. Lowercase glued keys are ordinary in YAML, `.env`, and config dumps a pane echoes. Separately worth noting while you are in this rule: a quoted label such as `{&#34;api_key&#34;: &#34;sk_fm_fake_live...&#34;}` or `  &#34;password&#34;: &#34;...&#34;` is matched by neither form, because `(^|[[:space:]])` requires the label to start immediately after whitespace and a leading quote blocks it — that one is pre-existing rather than new. Options if you want the coverage back without reintroducing the `tokens:` false positive: allow a leading quote/brace/dash in the anchor class, permit a glued prefix only when the credential word ends the label (`(^|[[:space:]&#34;&#39;{\[])[A-Za-z0-9]*(ALT)([-_][A-Za-z0-9]+)*`), or lower the operand bound and instead require the operand to look secret-shaped.
- ⚠️ `bin/fm-engineering-lib.sh:189` - `fm_eng_lock_release_all` iterates `for dir in $FM_ENG_LOCK_HELD`, an unquoted space-joined string built at line 214 (`FM_ENG_LOCK_HELD=&#34;${FM_ENG_LOCK_HELD}${FM_ENG_LOCK_HELD:+ }$dir&#34;`). A lock directory whose path contains a space is split into fragments, so `fm_eng_lock_remove` is called on paths that do not exist and the real directory is never removed. Reproduced with an FM_HOME of `.../My Home`: `fm_eng_lock_acquire` then `fm_eng_lock_release_all` leaves the lock directory behind (`fm_eng_lock_release` survives only because it calls `fm_eng_lock_remove &#34;$1&#34;` directly, ignoring its own broken bookkeeping). This matters most in `bin/fm-captains-log-projection.sh`, where `trap fm_eng_lock_release_all RETURN` (line 456) is the *only* release path for the intent mutex — every intent on such a home leaks a directory under `state/fm-projection-intent-locks/`, which then accumulates one per intentId. It self-heals (the next acquirer sees a dead owner and reclaims at once) rather than wedging, and it also defeats the INT/TERM/HUP traps in `bin/fm-report.sh`. `fm_eng_lock_release`&#39;s `kept` rebuild is corrupted the same way once more than one lock is held. Use a newline-delimited list read under `IFS=$&#39;\n&#39;`, or track held locks in a bash array.
- ℹ️ `bin/fm-engineering-lib.sh:232` - `fm_eng_lock_acquire` returns 1 both for contention (`mkdir &#34;$dir&#34;` failed after reclaim) and for a permanent condition (`mkdir -p &#34;$(dirname &#34;$dir&#34;)&#34;` failed — unwritable or missing `$STATE`). `fm_eng_lock_acquire_wait` cannot tell them apart and retries 100 times regardless, so `fm-report.sh append` on a home with a non-writable `state/` stalls the reporting crewmate for the full wait before returning `record_not_durable` — and it now takes that path on *every* append, since `next_acceptance_sequence` (bin/fm-report.sh:110) is unconditional. Return a distinct status for the un-creatable-parent case and fail fast on it. While there: `sleep 0.05 2&gt;/dev/null || sleep 1` means the wait is ~5s on hosts whose `sleep` accepts fractions and ~100s on one that does not, so the bound the retry count implies is not the bound a caller actually gets.
- ℹ️ `bin/fm-engineering-lib.sh:82` - `fm_eng_credential_label_ere` recovers the alternation by stripping the literal wrapper text off `fm_eng_credential_name_ere`&#39;s output (`word=${names#&#39;[A-Za-z0-9_]*(&#39;}` / `word=${word%&#39;)[A-Za-z0-9_]*&#39;}`). That couples the label form to the exact `printf &#39;[A-Za-z0-9_]*(%s)[A-Za-z0-9_]*&#39;` format string two functions away: change the wrapper and the strip silently no-ops, leaving `word` as the whole name ERE and producing a label regex that matches far more than intended, with no error. Extract the alternation into a `fm_eng_credential_alternation` helper that both `fm_eng_credential_name_ere` and `fm_eng_credential_label_ere` wrap, so neither has to parse the other&#39;s output.
- ℹ️ `bin/fm-captains-log-projection.sh:259` - The isolation filter `select((.cycleRef == null or .buildRef == null) or (.cycleRef == $shape.cycleRef and .buildRef == $shape.buildRef))` means any isolated mission or report whose own correlation is unreadable withholds readiness from *every* Build Review, not just the bound Build — and an unparseable JSON file always lands in that bucket, since `collect_records` stamps `cycleRef: null, buildRef: null` for `malformed_record`. That is the documented and requested fail-closed behavior (docs/captains-log-machine-projection.md: &#34;could belong to any Build, so it withholds readiness across the Review set&#34;), and the record is named in `invalidRecords` so the operator can find it. Recording the blast radius rather than proposing a change: one corrupt file in `data/engineering/reports/` blocks Captain acceptance of every Build until it is repaired or removed.

🔧 Fix: fix(engineering): restore colon secrets, path-safe locks, fast lock refusal
- ⚠️ `bin/fm-engineering-lib.sh:296` - `fm_eng_lock_acquire_wait` derives `attempts` from the requested seconds (`attempts=$((seconds * 20))` at line 296) and counts only the `sleep &#34;$unit&#34;` between attempts, but each attempt calls `fm_eng_lock_acquire` -&gt; `fm_eng_lock_reclaim` -&gt; `fm_eng_lock_owner_state`, which forks roughly ten processes (`dirname`, `mkdir`, three `sed`, three `head`, `ps`, `tr`). That overhead is unbudgeted, so the delivered bound is about 2.5x the requested one. Measured on an idle machine against a live-owner lock: requested 1s took 2551ms, 2s took 5390ms, and the default 5s took 12470ms. This contradicts the comment at line 276 (&#34;the bound a caller asks for is the bound it gets whether or not this host&#39;s sleep accepts a fraction&#34;) and the new docs/captains-log-machine-projection.md:113 claim that &#34;contention is waited out for a bound expressed in seconds&#34;. The practical cost is a ~12.5s stall on `fm-report.sh append` — the crewmate&#39;s reporting path — whenever two appends contend, plus ~1000 process spawns for that one wait. Two consequences to fix together: compute the deadline from a wall clock and stop when it passes rather than counting attempts, and stop re-running the full owner probe on every retry (probe once, then just retry `mkdir`, re-probing only occasionally). Note also that the new tests/fm-engineering-lib.test.sh:220 ceiling `[ &#34;$elapsed&#34; -le 6 ]` for a 2s bound leaves only ~0.6s of headroom over the 5.4s measured here, so it is likely to flake on a slower or loaded CI runner.
- ℹ️ `bin/fm-engineering-lib.sh:267` - `FM_ENG_LOCK_PAUSE_UNIT` is intended to memoize the fractional-sleep probe, but `fm_eng_lock_pause_unit` is invoked as `unit=$(fm_eng_lock_pause_unit)` (line 289), so the assignment happens in a command-substitution subshell and never reaches the parent. Verified: after three `fm_eng_lock_acquire_wait` calls the parent&#39;s `FM_ENG_LOCK_PAUSE_UNIT` is still empty. Because the probe body runs a real `sleep 0.05`, every `fm_eng_lock_acquire_wait` pays it — including the uncontended common path, since the probe is called before the first acquire attempt. Measured: an uncontended `fm_eng_lock_acquire_wait` costs 154ms against 26ms for a bare `fm_eng_lock_acquire`, so roughly 128ms is added to every `fm-report.sh append` for a capability check whose answer never changes. Have the function assign the global directly and return, rather than being consumed through `$(...)`; better still, defer the probe until the first retry is actually needed.
- ℹ️ `bin/fm-peek.sh:117` - The new line-final colon rule accepts an operand of any length, which makes the pre-existing &#34;redaction suppresses the span re-split&#34; mechanism fire on short, inert-looking values and merge unrelated pane rows again — a narrower instance of the round-2 class. Reproduced: rows `checking configuration values and other things` (pane-wide, inert) and `password: unset` come back as the single line `checking configuration values and other thingspassword: [REDACTED]`, while the same shape with a non-line-final operand (`password: unset for now`) correctly keeps both rows. docs/captains-log-machine-projection.md:134 still states &#34;A row that merely fills the pane without wrapping keeps its own line, so unrelated events are never merged into one.&#34; The redaction itself is defensible fail-closed behavior; the lost event boundary is not. Root cause is in the final awk stage, which prints the merged text whenever `index(text, &#34;[REDACTED]&#34;) &gt; 0` regardless of where the marker sits: re-split on the recorded spans when every `[REDACTED]` marker lies wholly inside one span, and only keep the merged line when a redaction actually straddles a span boundary.

🔧 Fix: fix(engineering): real lock deadline, deferred probe, row boundaries
- ⚠️ `bin/fm-peek.sh:138` - `settle()` prints `logical` (the joined form) whenever `joined_rows != logical`, on the assumption that any disagreement means a redaction straddled a wrap. It can also mean the per-row pass redacted something the joined pass missed, and in that case the more-redacted result is discarded. Reproduced: a pane-wide row `...password: hunt3` followed by `  more context here` returns `checking configuration values and other  step 4 of 9 note now  password: hunt3  more context here` with the value in the clear, while the identical rows with the credential row last return `password: [REDACTED]`. Mechanism: the row pass matches rule 2 (line-final operand, any length) because `hunt3` ends row B, but in the joined line `hunt3` is followed by row C&#39;s leading space so it is neither line-final nor long enough for rule 1&#39;s `{6,}` bound; the forms disagree and the un-redacted joined form wins. Confirmed bounded to operands under six characters — `hunt36` and `fm_fake_longsecret1` are redacted correctly in the same geometry. This is not a regression (the previous revision leaked the same input through the single-form path), but this commit already computes the correct answer in the R pass. Fix: on disagreement prefer whichever form carries more `[REDACTED]` markers, falling back to the joined form only on a tie — that keeps the straddling-wrap case (joined has one marker, rows have none) and stops discarding a row-only redaction. The docs claim at docs/captains-log-machine-projection.md:137 (&#34;only a redaction that straddles a wrap keeps the joined line&#34;) describes the intended rule, not the implemented one.
- ℹ️ `bin/fm-engineering-lib.sh:313` - `deadline=$(( $(date +%s) + seconds ))` reads a whole-second clock, so the delivered wait is anywhere from `seconds - 1` to `seconds`. Measured: a 1s request completed in 403ms — under 60% of the requested grace period — while 2s/5s/6s came in at 2030ms/4849ms/6132ms. Only very short overrides of `FM_REPORT_SEQUENCE_LOCK_WAIT_SECONDS` or `fm_eng_lock_acquire_wait`&#39;s fourth argument are affected, and the default 5s is accurate enough, so this is polish rather than a defect. Add one second to the deadline (or derive it from the same sub-second source the pause unit already proves is available) so the requested bound is a floor rather than a ceiling.

🔧 Fix: fix(engineering): safe redaction candidate, lock deadline floor
2 infos still open:
- ℹ️ `bin/fm-peek.sh:160` - The six-expression sed rule block is now written out twice, verbatim, at lines 121-127 and 160-166, and the `claim` awk function is duplicated at lines 89-94 and 133-138. The safety invariant survives drift — both candidates pass through the second block, and the joined candidate is built from rows that already passed the first, so an expression present in only one block cannot release a value the other masked — but it does silently change behavior: a rule added only to the first block stops catching straddling wraps, and one added only to the second makes the two candidates disagree on inputs where nothing straddled, costing row boundaries for no reason. Neither failure produces an error. Collect the `-e` expressions into a shell array (`local -a rules=(-e &#39;...&#39; -e &#39;...&#39;)`) built once from `$names`/`$labels`/`$anchor`/`$quotes` and expand it at both call sites, and pass the shared `claim` body to both awk stages through a single here-doc or `-v` prologue.
- ℹ️ `tests/fm-peek.test.sh:350` - `refute_fragments &#34;session inspection&#34; &#34;$candidate_inspection&#34; &#34;$CANDIDATE_SECRET&#34;` never executes its body. `refute_fragments` (line 105) loops `for ((i = 0; i + window &lt;= ${#secret}; i++))` with `window=8`, and `CANDIDATE_SECRET=&#39;hunt3&#39;` is five characters, so `0 + 8 &lt;= 5` is false and zero comparisons run. The peek output is genuinely covered by the explicit `case` at line 347, but the projection&#39;s `inspectSession` path — which the new case&#39;s own comment says must never publish a value the other candidate masked — has no effective assertion at all. The same is true of the `refute_fragments` call for `SHORT_VALUE=&#39;hunt3&#39;` earlier in the file. Add a direct `case &#34;$(printf &#39;%s&#39; &#34;$candidate_inspection&#34; | jq -r &#39;.inspection.capture.text&#39;)&#34; in *&#34;$CANDIDATE_SECRET&#34;*) fail ...` check alongside the existing one, or have `refute_fragments` fall back to a whole-string containment check when the secret is shorter than its window.

</details>

<details>

✅ No issues found.
- `bash tests/fm-captains-log-projection.test.sh`
- `bash tests/fm-mission-shapeup.test.sh`
- `bash tests/fm-report.test.sh`
- `bash tests/fm-shapeup-client.test.sh`
- `bash tests/fm-decision-hold.test.sh`
- `bash tests/fm-build-review.test.sh`
- `bash tests/fm-peek.test.sh`
- `bash tests/fm-engineering-lib.test.sh`
- <code>`bash tests/fm-bearings-snapshot.test.sh` (suite modified by this change)</code>
- `bash tests/fm-fleet-snapshot-view.test.sh`
- `bash tests/fm-decision-hold-lifecycle.test.sh`
- `bash tests/fm-documentation-audiences.test.sh`
- <code>`bash tests/fm-backend.test.sh`, `bash tests/fm-backend-tmux-smoke.test.sh`, `bash tests/fm-composer-ghost.test.sh` (fm-peek.sh consumers)</code>
- <code>Manual end-to-end walkthrough driving bin/fm-mission.sh, bin/fm-report.sh, bin/fm-captains-log-projection.sh, bin/fm-shapeup-client.sh, bin/fm-decision-hold.sh, bin/fm-peek.sh and bin/fm-fleet-snapshot.sh against a throwaway FM_HOME: `bash e2e-captains-log-walkthrough.sh &lt;repo-root&gt;`</code>
- <code>Manual credential-leakage checks: grep of client output and the whole persisted store for the workspace token, transport argv log showing `argc=0`, and an 8-character sliding-window fragment scan over the pane-wrapped secret in the inspectSession capture</code>
- <code>Manual durability inspection of the resulting record store (`find data/engineering -type f`, `jq` over the retained rejected records, `tasks-axi show mission-42-decision-scope-drop --full`)</code>

</details>

<details>

- ℹ️ `AGENTS.md:78` - AGENTS.md section 2&#39;s private-home tree enumerates every config/ file and the durable data/ records, but this change&#39;s config/shapeup-client.json, its credential file, and data/engineering/ are absent. I deliberately did not add them: docs/configuration.md is the stated owner of the operational-home layout and was updated with data/engineering/ and a ShapeUp client section, and AGENTS.md is reserved for knowledge almost every session needs (it is also under a 7,500-token startup budget). Flagging it as a judgment call in case the maintainer wants the always-loaded inventory to stay exhaustive for config/.
- ℹ️ `docs/documentation-audiences.json:35` - Out-of-scope hardening for a follow-up: docs/documentation-audiences.json&#39;s requiredOwnerPointers does not register the change&#39;s two new pointer edges (docs/configuration.md -&gt; docs/shapeup-client.md and docs/architecture.md -&gt; docs/captains-log-machine-projection.md). Those pointers are the only path from README-linked docs to the new surfaces, so registering them would let tests/fm-documentation-audiences.test.sh catch their silent removal the way it already guards the calm, supervision, and backend pointers. I left the inventory alone because nothing is currently stale and the check passes (surfaces=63, local_links=176).

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

### DIFF
--- a/.agents/skills/decision-hold-lifecycle/SKILL.md
+++ b/.agents/skills/decision-hold-lifecycle/SKILL.md
@@ -22,6 +22,8 @@ A completed investigation and an ended visual review use this same owner and com
 Run the command in the originating work's authoritative `FM_HOME`; main-home work creates main-home holds, and secondmate-owned work creates holds in that secondmate home's backlog rather than copying them into the main backlog.
 Do not close a hold merely because the originating investigation completed, its report was archived, its visual review ended, or its task was torn down.
 The hold remains the authoritative Captain's Call item until the captain's answer is durably recorded, dependent work is created in the same backlog and blocked by that hold, and `bin/fm-decision-hold.sh resolve` routes the answer by clearing those dependency edges before closing the hold.
+A Captain Call that `bin/fm-report.sh` raises from an accepted Engineering mission is the one exception: it lives in this same backlog and closes through this same `resolve` owner, but only the `resolveCondition` intent of `bin/fm-captains-log-projection.sh` may drive it.
+A bare `resolve` would close that hold without the durable outcome the projection derives resolution from, leaving the captain's answer unprojected.
 Resolved findings, recommendations that need no captain choice, and prose that merely sounds decision-like do not create holds.
 Bearings reads the resulting structured state and must never compensate by scraping historical reports, visual-review artifacts, terminal output, chat, or other prose.
 
--- a/bin/fm-captains-log-projection.sh
+++ b/bin/fm-captains-log-projection.sh
@@ -449,10 +449,12 @@ intent_response() {  # <intent-json>
     2) return 2 ;;
   esac
   lock_dir="$STATE/fm-projection-intent-locks/$intent_id"
-  fm_eng_lock_acquire "$lock_dir" "${FM_PROJECTION_INTENT_LOCK_STALE_MINUTES:-}" "$NOW" || {
-    fail_json intent_busy "The intent is already being processed"
-    return 2
-  }
+  fm_eng_lock_acquire "$lock_dir" "${FM_PROJECTION_INTENT_LOCK_STALE_MINUTES:-}" "$NOW"
+  case $? in
+    0) ;;
+    1) fail_json intent_busy "The intent is already being processed"; return 2 ;;
+    *) fail_json intent_lock_unavailable "The intent mutex could not be created, so the intent was not applied"; return 2 ;;
+  esac
   trap fm_eng_lock_release_all RETURN
   replay_prior_outcome "$record" "$digest_value"
   case $? in
--- a/bin/fm-captains-log-projection.sh
+++ b/bin/fm-captains-log-projection.sh
@@ -1,0 +1,679 @@
+#!/usr/bin/env bash
+# fm-captains-log-projection.sh - bounded machine projection for Captain's Log.
+#
+# Reads one JSON request from stdin and prints one JSON response.
+# The stable request schema is fm-captains-log-projection.v1.
+#
+# Supported operations:
+#   {"schemaVersion":"fm-captains-log-projection.v1","operation":"capabilities"}
+#   {"schemaVersion":"fm-captains-log-projection.v1","operation":"snapshot"}
+#   {"schemaVersion":"fm-captains-log-projection.v1","operation":"intent","intent":{...}}
+#
+# Core projection data is stored below data/engineering as closed JSON records.
+# Invalid records remain visible in snapshot.invalidRecords and are never used
+# to derive executable conditions, review packets, or command outcomes.
+set -u
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+FM_ROOT="${FM_ROOT_OVERRIDE:-$(cd "$SCRIPT_DIR/.." && pwd)}"
+FM_HOME="${FM_HOME:-${FM_ROOT_OVERRIDE:-$FM_ROOT}}"
+DATA="${FM_DATA_OVERRIDE:-$FM_HOME/data}"
+STATE="${FM_STATE_OVERRIDE:-$FM_HOME/state}"
+ENGINEERING="$DATA/engineering"
+MISSIONS="$ENGINEERING/missions"
+REPORTS="$ENGINEERING/reports"
+OUTCOMES="$ENGINEERING/outcomes"
+SHAPEUP_OUTCOMES="$ENGINEERING/shapeup-outcomes"
+NOW="${FM_PROJECTION_NOW:-$(date -u +%Y-%m-%dT%H:%M:%SZ)}"
+SCHEMA=fm-captains-log-projection.v1
+
+# shellcheck source=bin/fm-engineering-lib.sh
+# shellcheck disable=SC1091
+. "$SCRIPT_DIR/fm-engineering-lib.sh"
+
+command -v jq >/dev/null 2>&1 || {
+  echo "fm-captains-log-projection: jq not found" >&2
+  exit 1
+}
+
+digest() {
+  if command -v shasum >/dev/null 2>&1; then
+    shasum -a 256 | awk '{print $1}'
+  else
+    sha256sum | awk '{print $1}'
+  fi
+}
+
+fail_json() {  # <code> <message>
+  jq -n --arg schemaVersion "$SCHEMA" --arg code "$1" --arg message "$2" \
+    '{accepted:false,schemaVersion:$schemaVersion,error:{code:$code,message:$message}}'
+}
+
+valid_identity() {  # <value>
+  case "$1" in
+    ''|*[!A-Za-z0-9._-]*) return 1 ;;
+  esac
+  [ "${#1}" -le 128 ]
+}
+
+INTENT_LOCK_DIR=''
+INTENT_LOCK_HOST=${HOSTNAME:-$(hostname 2>/dev/null || printf 'unknown-host')}
+INTENT_LOCK_STALE_MINUTES=5
+case "${FM_PROJECTION_INTENT_LOCK_STALE_MINUTES:-}" in
+  ''|*[!0-9]*) ;;
+  *)
+    if [ "${#FM_PROJECTION_INTENT_LOCK_STALE_MINUTES}" -le 9 ] \
+      && [ "$FM_PROJECTION_INTENT_LOCK_STALE_MINUTES" -ge 1 ]; then
+      INTENT_LOCK_STALE_MINUTES=$FM_PROJECTION_INTENT_LOCK_STALE_MINUTES
+    fi
+    ;;
+esac
+
+remove_intent_lock() {  # <lock-dir>
+  rm -f "$1/owner" 2>/dev/null || true
+  rmdir "$1" 2>/dev/null || true
+}
+
+release_intent_lock() {
+  [ -n "$INTENT_LOCK_DIR" ] || return 0
+  remove_intent_lock "$INTENT_LOCK_DIR"
+  INTENT_LOCK_DIR=''
+}
+
+PROCESS_START_METHOD=lstart
+
+# The signature is rendered in a fixed locale and timezone so the same process
+# always yields the same identity whatever environment observes it. It carries
+# its method so a signature produced a different way is never mistaken for a
+# different process.
+process_start_signature() {  # <pid>
+  local value
+  value=$(LC_ALL=C TZ=UTC ps -o lstart= -p "$1" 2>/dev/null | tr -d '[:space:]')
+  [ -n "$value" ] || return 1
+  printf '%s:%s' "$PROCESS_START_METHOD" "$value"
+}
+
+# A lock whose recorded owner is provably gone - exited, or its pid recycled by a
+# process that started later - is reclaimed at once. An owner that cannot be
+# compared, including one recorded by another identity method, falls back to the
+# bounded age window, so no lock can wedge forever and none is reclaimed merely
+# because its identity is presented differently. A confirmed live owner is never
+# reclaimed, however long it runs and whoever owns it.
+intent_lock_owner_state() {  # <lock-dir>
+  local dir=$1 owner_host owner_pid owner_started current_started
+  if [ ! -f "$dir/owner" ]; then
+    printf 'unknown'
+    return 0
+  fi
+  owner_host=$(sed -n 's/^host=//p' "$dir/owner" 2>/dev/null | head -1)
+  owner_pid=$(sed -n 's/^pid=//p' "$dir/owner" 2>/dev/null | head -1)
+  owner_started=$(sed -n 's/^started=//p' "$dir/owner" 2>/dev/null | head -1)
+  if [ "$owner_host" != "$INTENT_LOCK_HOST" ]; then
+    printf 'unknown'
+    return 0
+  fi
+  case "$owner_pid" in
+    ''|*[!0-9]*) printf 'unknown'; return 0 ;;
+  esac
+  current_started=$(process_start_signature "$owner_pid") || current_started=''
+  if [ -z "$current_started" ]; then
+    if kill -0 "$owner_pid" 2>/dev/null; then
+      printf 'unknown'
+    else
+      printf 'dead'
+    fi
+    return 0
+  fi
+  case "$owner_started" in
+    "$PROCESS_START_METHOD":?*) ;;
+    *) printf 'unknown'; return 0 ;;
+  esac
+  if [ "$owner_started" = "$current_started" ]; then
+    printf 'alive'
+  else
+    printf 'dead'
+  fi
+}
+
+reclaim_stale_intent_lock() {  # <lock-dir>
+  local dir=$1
+  [ -d "$dir" ] || return 0
+  case "$(intent_lock_owner_state "$dir")" in
+    alive) return 0 ;;
+    dead) ;;
+    *)
+      [ -n "$(find "$dir" -maxdepth 0 -mmin "+$INTENT_LOCK_STALE_MINUTES" -print 2>/dev/null)" ] || return 0
+      ;;
+  esac
+  remove_intent_lock "$dir"
+}
+
+trap release_intent_lock EXIT
+trap 'release_intent_lock; exit 130' INT
+trap 'release_intent_lock; exit 143' TERM
+trap 'release_intent_lock; exit 129' HUP
+
+collect_records() {  # <dir> <kind>
+  local dir=$1 kind=$2 file records invalid
+  local -a files=() valid=() malformed=()
+  if [ -d "$dir" ]; then
+    while IFS= read -r file; do
+      [ -n "$file" ] || continue
+      files[${#files[@]}]=$file
+    done < <(find "$dir" -type f -name '*.json' -print 2>/dev/null | LC_ALL=C sort)
+  fi
+  if [ "${#files[@]}" -eq 0 ]; then
+    printf '{"records":[],"invalid":[]}'
+    return 0
+  fi
+  if records=$(printf '%s\0' "${files[@]}" | xargs -0 cat 2>/dev/null | jq -cn '[inputs]' 2>/dev/null) \
+    && [ "$(printf '%s' "$records" | jq -r 'length')" = "${#files[@]}" ] \
+    && [ "$(printf '%s' "$records" | jq -r 'all(type == "object")')" = true ]; then
+    jq -cn --argjson records "$records" '{records:$records,invalid:[]}'
+    return 0
+  fi
+  for file in "${files[@]}"; do
+    if jq -e 'type == "object"' "$file" >/dev/null 2>&1; then
+      valid[${#valid[@]}]=$file
+    else
+      malformed[${#malformed[@]}]=$(basename "$file" .json)
+    fi
+  done
+  records='[]'
+  if [ "${#valid[@]}" -gt 0 ]; then
+    records=$(printf '%s\0' "${valid[@]}" | xargs -0 cat 2>/dev/null \
+      | jq -cn '[inputs | select(type == "object")]') || return 1
+    [ -n "$records" ] || return 1
+    [ "$(printf '%s' "$records" | jq -r 'length')" = "${#valid[@]}" ] || return 1
+  fi
+  invalid='[]'
+  if [ "${#malformed[@]}" -gt 0 ]; then
+    invalid=$(jq -cn --arg kind "$kind" '
+      [ $ARGS.positional[]
+        | {kind:$kind,recordId:.,error:{code:"malformed_record",message:"Record is not a valid JSON object"}} ]
+    ' --args "${malformed[@]}") || return 1
+  fi
+  jq -cn --argjson records "$records" --argjson invalid "$invalid" '{records:$records,invalid:$invalid}'
+}
+
+condition_projection() {  # <reports-json> <outcomes-json>
+  jq -cn --argjson reports "$1" --argjson outcomes "$2" '
+    [ $reports[]
+      | select(.condition? and (.condition | type == "object"))
+      | select(.status? == "accepted")
+      | {
+          reportId:.reportId,
+          missionId:.missionId,
+          conditionId:.condition.conditionId,
+          holdId:(.condition.holdId // null),
+          packetRevision:(.condition.packetRevision // .sourceRevision),
+          lifecycle:(.condition.lifecycle // "raised"),
+          affectedObjects:(.condition.affectedObjects // []),
+          explanation:(.condition.explanation // ""),
+          recommendation:(.condition.recommendation // ""),
+          choices:(.condition.choices // []),
+          consequenceOfDelay:(.condition.consequenceOfDelay // ""),
+          evidence:(.condition.evidence // []),
+          boundRevisions:(.condition.boundRevisions // {}),
+          creationSequence:(.condition.creationSequence // 0),
+          projection:(.condition.projection // null),
+          cycleRef:(.correlation.shapeUp.cycleRef // ""),
+          buildRef:(.correlation.shapeUp.buildRef // ""),
+          scopeRef:(.correlation.shapeUp.scopeRef // null),
+          capturedAt:(.capturedAt // ""),
+          acceptedAt:(.acceptedAt // "")
+        }
+    ]
+    | sort_by(.missionId,.conditionId,.acceptedAt,.capturedAt,.reportId)
+    | group_by([.missionId,.conditionId])
+    | map(last)
+    | map(. as $condition
+      | .lifecycle = (
+          [ $outcomes[]
+            | select(.intent.action == "resolveCondition")
+            | select(.intent.missionId == $condition.missionId)
+            | select(.intent.conditionId == $condition.conditionId)
+            | select(.outcome.status == "resolved")
+          ]
+          | if length > 0 then "resolved" else $condition.lifecycle end))
+    | sort_by(.cycleRef,.buildRef,(.scopeRef // ""),.creationSequence,.missionId,.conditionId)
+  '
+}
+
+build_reviews() {  # <missions-json> <reports-json> <outcomes-json>
+  local missions=$1 reports=$2 outcomes=$3
+  jq -cn --argjson missions "$missions" --argjson reports "$reports" --argjson outcomes "$outcomes" '
+    def authoritative: sort_by(.acceptedAt // "",.capturedAt // "",.reportId) | last // null;
+    def terminal($missionId):
+      ([ $reports[]
+        | select(.status == "accepted" and .missionId == $missionId)
+        | select(.kind == "scope_discovery" or .kind == "scope_revision" or .kind == "hill_judgment" or .kind == "blocker" or .kind == "outcome")
+      ] | authoritative) as $latest
+      | ($latest != null
+        and $latest.kind == "outcome"
+        and ($latest.outcome.state == "completed" or $latest.outcome.state == "failed" or $latest.outcome.state == "cancelled"));
+    def required_evidence_verified($mission):
+      (($mission.evidenceRequirements // []) | all(. as $requirement |
+        ([ $reports[]
+          | select(.status == "accepted")
+          | select(.missionId == $mission.missionId and .kind == "evidence")
+          | select(.evidence.contract == $requirement)
+        ] | authoritative) as $latest
+        | ($latest != null and $latest.evidence.verification.status == "verified")));
+    [ $missions[]
+      | select((.state // "accepted") == "accepted")
+      | select(.correlation.shapeUp.buildRef? != null)
+    ]
+    | sort_by(.correlation.shapeUp.cycleRef,.correlation.shapeUp.buildRef,.missionId)
+    | group_by([.correlation.shapeUp.cycleRef,.correlation.shapeUp.buildRef])
+    | map(
+        . as $group
+        | ($group[0].correlation.shapeUp) as $shape
+        | ($group | map(.correlation.shapeUp.buildRevision // "") | unique) as $buildRevisions
+        | (($buildRevisions | length) > 1) as $revisionConflict
+        | ($group | map(. as $mission | {
+            missionId:$mission.missionId,
+            missionRevision:($mission.sourceRevision // ""),
+            buildRevision:($mission.correlation.shapeUp.buildRevision // ""),
+            terminal:terminal($mission.missionId),
+            evidenceVerified:required_evidence_verified($mission),
+            reportRevisions:([ $reports[]
+              | select(.status == "accepted" and .missionId == $mission.missionId)
+              | .sourceRevision ] | sort),
+            evidenceRevisions:([ $reports[]
+              | select(.status == "accepted" and .missionId == $mission.missionId and .kind == "evidence")
+              | .sourceRevision ] | sort)
+          })) as $missionSet
+        | {
+            cycleRef:$shape.cycleRef,
+            buildRef:$shape.buildRef,
+            buildRevision:(if $revisionConflict then null else ($buildRevisions[0] // "") end),
+            buildRevisions:$buildRevisions,
+            revisionConflict:$revisionConflict,
+            missionSet:$missionSet,
+            ready:(($revisionConflict | not) and ($missionSet | all(.terminal and .evidenceVerified))),
+            packetMaterial:({
+              cycleRef:$shape.cycleRef,
+              buildRef:$shape.buildRef,
+              buildRevisions:$buildRevisions,
+              revisionConflict:$revisionConflict,
+              missionSet:$missionSet
+            } | @json)
+          }
+      )
+  '
+}
+
+review_revisions() {  # <reviews-json>
+  local reviews=$1 count index material revision result='[]'
+  count=$(printf '%s' "$reviews" | jq -r 'if type == "array" then length else "invalid" end') || return 1
+  case "$count" in
+    ''|*[!0-9]*) return 1 ;;
+  esac
+  index=0
+  while [ "$index" -lt "$count" ]; do
+    material=$(printf '%s' "$reviews" | jq -r --argjson index "$index" '.[$index].packetMaterial') || return 1
+    revision=$(printf '%s' "$material" | digest) || return 1
+    [ -n "$revision" ] || return 1
+    result=$(jq -cn --argjson result "$result" --argjson reviews "$reviews" \
+      --argjson index "$index" --arg revision "$revision" '
+      $result + [$reviews[$index] | del(.packetMaterial) | . + {packetRevision:$revision}]') || return 1
+    index=$((index + 1))
+  done
+  printf '%s' "$result"
+}
+
+snapshot_json() {
+  local mission_bundle report_bundle outcome_bundle shapeup_bundle missions reports outcomes shapeup_outcomes invalid conditions material source_revision reviews
+  mission_bundle=$(collect_records "$MISSIONS" mission) || return 1
+  report_bundle=$(collect_records "$REPORTS" report) || return 1
+  outcome_bundle=$(collect_records "$OUTCOMES" outcome) || return 1
+  shapeup_bundle=$(collect_records "$SHAPEUP_OUTCOMES" shapeup-outcome) || return 1
+  missions=$(printf '%s' "$mission_bundle" | jq -c '.records | sort_by(.missionId // "")') || return 1
+  reports=$(printf '%s' "$report_bundle" | jq -c '.records | sort_by(.capturedAt // "",.reportId // "")') || return 1
+  outcomes=$(printf '%s' "$outcome_bundle" | jq -c '.records | sort_by(.recordedAt // "",.intentId // "")') || return 1
+  shapeup_outcomes=$(printf '%s' "$shapeup_bundle" | jq -c '.records | sort_by(.outcome.recordedAt // "",.outcome.reportId // "")') || return 1
+  invalid=$(jq -cn \
+    --argjson a "$(printf '%s' "$mission_bundle" | jq -c '.invalid')" \
+    --argjson b "$(printf '%s' "$report_bundle" | jq -c '.invalid')" \
+    --argjson c "$(printf '%s' "$outcome_bundle" | jq -c '.invalid')" \
+    --argjson d "$(printf '%s' "$shapeup_bundle" | jq -c '.invalid')" \
+    '$a + $b + $c + $d') || return 1
+  conditions=$(condition_projection "$reports" "$outcomes") || return 1
+  [ -n "$conditions" ] || return 1
+  material=$(jq -cS -n --argjson missions "$missions" --argjson reports "$reports" \
+    --argjson outcomes "$outcomes" --argjson shapeUpOutcomes "$shapeup_outcomes" --argjson invalid "$invalid" \
+    '{missions:$missions,reports:$reports,outcomes:$outcomes,shapeUpOutcomes:$shapeUpOutcomes,invalidRecords:$invalid}') || return 1
+  source_revision=$(printf '%s' "$material" | digest) || return 1
+  [ -n "$source_revision" ] || return 1
+  reviews=$(build_reviews "$missions" "$reports" "$outcomes") || return 1
+  [ -n "$reviews" ] || return 1
+  reviews=$(review_revisions "$reviews") || return 1
+  reviews=$(jq -cn --argjson reviews "$reviews" --argjson outcomes "$outcomes" '
+    $reviews
+    | map(. as $review
+      | .acceptance = (
+          [ $outcomes[]
+            | select(.intent.action == "acceptBuildReview")
+            | select(.intent.cycleRef == $review.cycleRef)
+            | select(.intent.buildRef == $review.buildRef)
+            | select(.intent.packetRevision == $review.packetRevision)
+            | .outcome
+          ] | last // null))
+  ') || return 1
+  jq -cn --arg capturedAt "$NOW" --arg sourceRevision "$source_revision" \
+    --argjson missions "$missions" --argjson reports "$reports" \
+    --argjson conditions "$conditions" --argjson outcomes "$outcomes" \
+    --argjson shapeUpOutcomes "$shapeup_outcomes" \
+    --argjson buildReviews "$reviews" --argjson invalidRecords "$invalid" '
+    {
+      schemaVersion:"fm-captains-log-snapshot.v1",
+      sourceRevision:$sourceRevision,
+      capturedAt:$capturedAt,
+      freshness:"fresh",
+      missions:$missions,
+      reports:$reports,
+      evidence:[$reports[]
+        | select(.status == "accepted" and .kind == "evidence")
+        | {reportId,sourceRevision,capturedAt,missionId,taskId,crewmateId,correlation,evidence}],
+      conditions:$conditions,
+      buildReviews:$buildReviews,
+      outcomes:$outcomes,
+      shapeUpOutcomes:$shapeUpOutcomes,
+      invalidRecords:$invalidRecords
+    }
+  '
+}
+
+capabilities_response() {
+  jq -n --arg schemaVersion "$SCHEMA" --arg capturedAt "$NOW" '
+    {
+      accepted:true,
+      schemaVersion:$schemaVersion,
+      sourceRevision:"capabilities-v1",
+      capturedAt:$capturedAt,
+      operations:["capabilities","snapshot","intent","inspectSession"],
+      capabilities:{
+        core:"available",
+        shapeUp:{status:"negotiated",required:false},
+        sessionInspection:{status:"negotiated",required:false,mode:"bounded-read-only",preferredBackend:"herdr"}
+      },
+      acceptedIntents:["acknowledgeCondition","acceptBuildReview","resolveCondition"]
+    }
+  '
+}
+
+replay_prior_outcome() {  # <record-path> <request-digest>
+  local record=$1 digest_value=$2 prior prior_digest
+  [ -f "$record" ] || return 1
+  prior=$(jq -c . "$record" 2>/dev/null) || {
+    fail_json malformed_record "The prior intent outcome is unreadable"
+    return 2
+  }
+  prior_digest=$(printf '%s' "$prior" | jq -r '.requestDigest // empty')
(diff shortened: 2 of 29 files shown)


kunchenguid:

<sub>Noted for firstmate#1555 at `01fb94e6`.</sub>

kunchenguid:

PR #1579, opened 2026-08-03
fix(bin): match every Claude busy footer without the idle timer trap

saucetech:
## Problem

Claude Code renders **three** busy footers and the shipped signature keyed on a shape only two of them have. A pane waiting on a tool renders `<Gerund> for 53s · 1 shell still running` — no token counter, no parenthesized duration — so the rendered delivery guards read a genuinely working pane as idle.

Consequences: `fm-send` reported "Enter swallowed; text left in composer" on healthy panes, and the away-mode and pending-reply guards misjudged working crews.

## The trap this avoids

The obvious fix is to anchor on the elapsed timer, since all three busy shapes carry one. Live capture shows that is wrong: a **finished** turn renders the same timer (`✻ Baked for 8m 42s`). A timer-only alternative reports an idle pane as busy — the dangerous direction, because a wedged crew then looks productive forever and never surfaces as stale, and the submit guard treats a swallowed Enter as delivered.

The discriminator is instead a **complete parenthesized duration-plus-detail footer** (streaming and thinking) or a **complete count-and-nouns wait hint** (tool wait), both anchored as full rows. A finished turn renders neither. A regression fixture pins the idle timer line so the timer anchor cannot be reintroduced.

## Changes

- Rewrite `FM_TMUX_CLAUDE_BUSY_REGEX_DEFAULT` to match all three footers by structure, anchored as full rows so ordinary transcript fragments cannot hold an idle pane busy.
- Thread the recorded harness through the submit path so `fm_pane_is_busy` selects the per-harness signature instead of the shared default.
- Add `fm_tmux_harness_supports_busy_queued_enter`, an explicit capability gate for treating a rendered busy footer as submit proof. Only `claude` and `opencode` opt in; every other harness and the unknown-harness case fail closed to `pending`. This prevents Kimi's locale-sensitive moon spinner from becoming submit proof, which the semantic contract in `bin/fm-busy-lib.sh` explicitly forbids.
- Consolidate the duplicated busy-queued-Enter contract into the `bin/fm-tmux-lib.sh` header, leaving one-line pointers in the afk and harness-adapters skills.

## ASCII-safe by construction

Every alternative is ASCII. The spinner glyph, the middot separator and the ellipsis are all locale-fragile, and the gerund is randomised and not always ASCII (`Sautéed`), so none of them are matched.

## Testing

Fixtures are live pane captures taken 2026-08-03 from both working and finished panes: all three busy shapes, the hour duration form, an accented gerund, ASCII-degraded rendering, the finished-turn lines that must read idle, and non-footer transcript text (`Retry completed (5s)`, `2 jobs still running after cleanup`) that must not match.

Each new busy assertion was proven to fail against the previous signature before the fix, and a mutation matrix confirmed every arm of the pattern is load-bearing.

## Not included

- `FM_TMUX_BUSY_REGEX_DEFAULT` (the shared harness-agnostic default) is deliberately unchanged: widening it would break the harness-scoping invariant that stops one harness's output classifying another.
- `bin/fm-spawn.sh` untouched.
- The signature encodes specific footer vocabulary, so a future Claude release that adds a new suffix will need a re-capture. Worth a periodic live-capture check as follow-up.


saucetech:
Closing: this PR was opened by hand and therefore cannot carry the no-mistakes signature that .github/workflows/no-mistakes-required.yml requires. Superseded by the PR the no-mistakes pipeline raises for the same branch, which carries the deterministic Pipeline section. Same commits, no code change.

PR #1586, opened 2026-08-03
fix(bin): stop the captain decision queue compounding

timidri:
Four defects in firstmate's captain-decision handling let the decision queue grow without draining. Together they mean a decision can be registered twice, can wait indefinitely without anyone noticing, is expensive to close, and can strand the investigation that raised it.

## A. A second review pass can duplicate an open decision

`bin/fm-decision-hold.sh hold` builds its identity as `<origin-id>-decision-<decision-key>`, so a later pass over the same subject registers a second captain decision whenever it picks a different key, even though `decision-hold-lifecycle` asks the agent to match semantically. Nothing put the already-open set in front of the caller before the write.

`hold` now refuses to create a *new* identity while the home holds other open captain decisions: it prints the open set with each decision's age and exits 3. A refusal rather than a warning, because output after a successful write is not reliably read, and semantic matching needs the open set in context before the write rather than after it. It is not a hard block - `--distinct` clears it and records the attestation in the new decision's body. `fold` is the first-class alternative for a finding that belongs to a question already open: it appends the finding to the existing decision and records the fold in the origin's metadata, so the origin still satisfies the completion gate via `complete --folded`. Exact-key repetition never reaches the gate, so idempotent retry and recovery paths are unchanged. `open` prints the same listing on demand, with `--aging-only` for the threshold subset.

## B. An open decision's age is invisible

`bin/fm-fleet-snapshot.sh` marked a row `captain_actionable` but carried nothing about how long it had waited, so `bin/fm-bearings-snapshot.sh`'s Captain's Call could not distinguish a decision raised an hour ago from one waiting a week, and the bounded `decisions_open` cap could truncate the oldest away entirely.

Every captain-actionable row now carries `waiting_days` and `decision_aging`, computed from the `since` metadata `tasks-axi add` already writes. There is no schema change and nothing to backfill, so decisions registered before this change report their age immediately. Bearings carries `since`, `waiting_days`, and `aging` into `decisions_open` and orders it ageing-first then oldest-first, so an ageing decision leads the section and cannot be truncated away by the cap. One `FM_DECISION_AGING_DAYS` (default 3) drives both the fleet view and the CLI listing, so they cannot disagree about which decisions are ageing. Three days is the shortest span that always covers a full working day plus a weekend edge, so it cannot fire on a decision the captain simply has not reached yet, while still surfacing a stalled one long before the week-long silences it exists to prevent.

## C. Recording an answer cost more than repeating the question

`resolve` required a decision *file* plus at least one dependent task already blocked by the decision. Where the answer spawned no dependent work, satisfying that meant inventing a task purely to block it - so an answer that arrives in conversation is cheaper to leave unrecorded, which is how a decision stays open and gets asked again.

`--decision <text>` and `--no-routed-work` remove that cost, and `--decided-by captain|firstmate` records who actually settled it. The routing requirement survives where it matters: `--no-routed-work` is refused while any task is blocked by the decision, checked against `tasks-axi list --blocked` rather than against the caller's claim. Closing now carries the prior registration record forward under `Origin record:` instead of overwriting it, so the distinctness attestation and any folded findings survive closure.

Separately, `verify_hold_durable` - which backs both the completion gate and scout teardown - accepted only this script's own resolution formatting. A decision closed with a real written record by any other route was rejected as "neither actively held nor durably resolved", and one that ordinary Done retention had rotated into `data/done-archive.md` was rejected as absent, stranding the finished investigation that raised it. The gate was rejecting on formatting, not on substance. One shared `closed_body_has_record()` predicate now applies to a closed decision's body, and `archived_done_body()` reads the archive's preserved indented body lines so the archived shape is held to that same test. Retention therefore only changes where a decision's record lives, never whether the gate accepts it.

This does not weaken the gate. A closed decision whose body is empty, or one still carrying the untouched registration record saying `State: awaiting captain decision.`, is refused wherever it lives - it says in its own words that no answer was ever written down, which is exactly the loss the gate exists to prevent. The regression covers that refusal both in the live backlog and after `prune` rotates it into the archive, so retention cannot launder a closure the live branch refused.

## D. The bar for asking was never written down

`ask-user-authority` owned the ask-user authority boundary but stated no threshold for whether a question belongs to the captain at all, so an investigation could register anything it found unclear. It now carries a concrete four-step escalate-or-decide test with worked examples on both sides, and `decision-hold-lifecycle` routes every candidate through that bar before registering anything, so the queue stops growing at the source. `AGENTS.md` section 13's trigger is widened to match, and `AGENTS.md` section 8 gains one line requiring an ageing decision to be resurfaced as still waiting rather than re-asked.

## Verification

```
bash tests/fm-decision-hold-lifecycle.test.sh   # 14 cases
bash tests/fm-bearings-snapshot.test.sh         # 42 cases
bin/fm-lint.sh
bin/fm-doc-audience-check.sh
```

The tests are behavioural, through the CLI and the snapshot contracts, on synthetic `sample-*` identities, with no assertions on implementation source. `FM_DECISION_NOW` and `FM_SNAPSHOT_NOW` pin the clock so the ageing arithmetic is deterministic.

New cases cover duplicate refusal and its exit code, the open set reaching the caller, folding (idempotence, no second decision, and refusal when the origin has no metadata to record the fold in), `--none` refused against recorded folds, `--distinct` passing a genuinely distinct decision, exact-key retry staying ungated, one-call answer recording, `--no-routed-work` refused while dependent work is blocked, externally closed and archived recognition, recordless closure refused both in the backlog and after it rotates into the archive, firstmate-decided attribution, and an ageing decision carrying its age, leading the section, and surviving the bounded cap.

## Scope

No migration and no schema change. Ageing is computed from the `since` metadata every row already carries, and the duplicate gate applies only to new registrations, so an existing queue stays readable and closable.

## Pipeline

<details>

✅ No issues found.

</details>

<details>
<summary>🔧 **Rebase** - 1 issue found → auto-fixed ✅</summary>

- ⚠️ `docs/scripts.md` - merge conflict rebasing onto origin/main

🔧 Fix applied.

</details>

<details>

- ⚠️ `bin/fm-decision-hold.sh:673` - `complete --folded` cannot satisfy an origin that has an open keyed *status* decision, leaving the new fold path a dead end. The status-open loop at 670-675 requires each open status key to be present in `$keys`, but on the `--folded` branch (633-636) `supplied` is empty, so `keys` is empty (unless an earlier `complete` already recorded keys) and `list_has_key &#34;&#34; &#34;default&#34;` fails. Failing sequence: a crewmate writes `needs-decision: ...` into `state/&lt;origin&gt;.status` (opening key `default` per fm-classify-lib.sh:148-159); a later review pass judges that question already owned by another home decision and runs `fold &lt;origin&gt; --into &lt;other-hold&gt;`; `complete &lt;origin&gt; --folded` then aborts with &#34;open structured decision &lt;origin&gt;/default has no captain-held inventory entry&#34;. Only `hold` appends the `captain-held [key=...]` transfer that closes a status fold (line 690), and `fold` writes no such transfer. The sole way forward is to register a second captain decision under that key — the exact compounding this change exists to prevent — or to attest `--distinct` falsely. Recommend making `fold` record the transfer for the covered status key(s), or teaching the `complete` status loop to accept a recorded `decision_folds` owner, at the shared status-fold boundary rather than in each caller.
- ⚠️ `bin/fm-decision-hold.sh:163` - `waiting_days` can under-report by a full day because `date_epoch` (157-161) parses a bare `%Y-%m-%d` and BSD `date -j -f` inherits the *current* H:M:S — confirmed on this host: `date -u -j -f &#39;%Y-%m-%d&#39; 2026-07-26 +%T` returns the live wall-clock time, not `00:00:00`. `waiting_days` makes two separate `date` invocations (now first, then since), so if a second ticks between them the difference is `N*86400 - 1`, and `/ 86400` floors to `N-1`. Concretely: a decision registered exactly `FM_DECISION_AGING_DAYS` days ago is rendered as `waiting 2 days` instead of 3 and is silently dropped by `open --aging-only`, roughly whenever the two calls straddle a second boundary. Fix by pinning the time explicitly, e.g. `date -u -j -f &#39;%Y-%m-%d %H:%M:%S&#39; &#34;$1 00:00:00&#34; +%s` with the GNU fallback already in place, so both endpoints are midnight-anchored and the subtraction is exact. The jq path in bin/fm-fleet-snapshot.sh:415 is unaffected (it uses `strptime`/`mktime` against a single `SNAPSHOT_EPOCH`).
- ℹ️ `bin/fm-decision-hold.sh:441` - `verify_resolution_identity` compares only the decision digest and routed identities, so `--decided-by` is outside the retry identity. Re-running `resolve` on an already-closed decision with the same text but `--decided-by captain` after it was first closed `--decided-by firstmate` (or vice versa) short-circuits at line 786-791 and prints `resolved: &lt;id&gt;` while the durable `Decided by:` line still records the original attributor. Given the change&#39;s own stated point that attribution must be honest, the recorded `Decided by:` value should be compared alongside the digest and routes and mismatches rejected like the others.
- ℹ️ `bin/fm-decision-hold.sh:202` - `list_open_captain_decisions` filters only on `--state held --kind captain` plus `hold_kind == captain`, while the fleet view&#39;s `captain_actionable` (bin/fm-fleet-snapshot.sh:409-411) additionally requires `state == &#34;queued&#34;` and `(.unresolved_blocker_ids | length) == 0`. A captain decision carrying an unresolved `blocked-by` edge therefore appears in `open`, in `open --aging-only`, and in the `hold` refusal set, but never in Bearings&#39; Captain&#39;s Call — so the refusal can cite a decision the captain cannot see. docs/decision-hold-lifecycle.md:65 states the fleet view and `open --aging-only` &#34;cannot disagree about which decisions are ageing&#34;, which holds for the threshold but not for the decision set. Either add the blocked/queued filter to the listing or narrow the doc claim to the threshold.
- ℹ️ `bin/fm-decision-hold.sh:649` - The fold-verification loop is byte-for-byte duplicated between `command_complete` (646-652) and `command_verify` (708-715), and `command_verify` re-reads `decision_folds` via `meta_value` (708) instead of the `origin_fold_ids` helper added at 259-263. Extracting a single `verify_folds &lt;folds&gt;` helper and using `origin_fold_ids` in both places removes the drift risk between the completion gate and the teardown gate, which the docs assert apply the same check.

🔧 Fix: fix folded status-decision deadlock and ageing boundary
- ⚠️ `bin/fm-decision-hold.sh:406` - An unqualified `fold` silently transfers the origin&#39;s *sole* open status decision even when the fold is about a different question, which can lose a captain question. `fold_covered_keys` only demands `--key` when more than one status decision is open (406-407); at exactly one it returns that key unconditionally (408) with nothing tying it to the note being folded. Concrete path: origin X&#39;s crewmate writes `needs-decision: &lt;question A&gt;` into `state/X.status`; a review pass of X finds a different question B, judges B already owned by open hold H, and runs `fold X --into H --note &#34;B&#34;`. Coverage resolves to `default` (question A), line 711 appends `captain-held [key=default]: tracked by H`, and H&#39;s body records only `Also raised by X: B`. Question A is now marked as owned by a decision that never mentions it, `complete X --folded` succeeds, and A is gone — the exact loss class this gate exists to prevent. This is likely rather than theoretical because .agents/skills/decision-hold-lifecycle/SKILL.md:23 and :44 tell the agent to run `fold` with no mention of `--key`, and one open status decision is the common shape. Suggest requiring `--key` whenever any status decision is open (`-ge 1` rather than `-gt 1`) — the caller always knows which question it folded — and naming `--key` in the skill&#39;s fold step. This tightens rather than contradicts the instruction not to let one fold satisfy several.
- ℹ️ `bin/fm-decision-hold.sh:392` - Repeating the same `--key` in one `fold` invocation appends the transfer line twice. `fold X --into H --key a --key a` builds `requested=&#34;a a&#34;`; both loop iterations at 392-400 find `a` in `open_keys` (the `status_transfer_recorded` skip at 395 cannot fire because nothing has been written yet), so `covered` becomes `a\na` and the loop at 710-712 appends `captain-held [key=a]: tracked by H` twice to `state/X.status`. Harmless to the gate but it corrupts the status log the crew-state reader consumes, and the existing idempotency test only covers repeated *invocations*. Dedupe `requested` (or `covered`) with the `sorted_key_union` helper already in the file.
- ℹ️ `bin/fm-decision-hold.sh:213` - Noting the accepted consequence of the requested fleet-view parity: because `list_open_captain_decisions` now excludes a captain decision carrying an unresolved blocker (new `f9` check at 213), that decision no longer appears in the `hold` refusal set, so a later review pass can register a second key for a question that is open-but-blocked without any warning. This is the direct trade the instruction asked for — the listing can no longer cite a decision Bearings does not show — and blocked decisions are not captain-actionable anyway, so no action is implied. Flagged only so the narrowing of the duplicate gate is on the record.

🔧 Fix: require fold --key while a status decision is open
2 warnings still open:
- ⚠️ `bin/fm-decision-hold.sh:410` - There is now no way to fold a finding that covers none of the origin&#39;s open status decisions, so the mixed case dead-ends. `fold` accepts only `--into`, `--note` and `--key` (654-663), and `fold_covered_keys` fails at 409-410 whenever `origin_open_decisions` is non-empty. Concrete path: origin X&#39;s live crewmate status ends `working: ...` / `needs-decision: &lt;question A&gt;` (exactly the round-3 fixture shape, so `origin_open_decisions` does not gate it to empty); A is a genuine captain question needing its own `hold`, and the same review pass separately finds B, already owned by open hold H. `fold X --into H --note B` is refused with &#34;origin X has 1 open structured decision&#34;, and `--key &lt;A&gt;` would append `captain-held [key=A]: tracked by H`, marking A as owned by a decision that never mentions it - the loss the refusal exists to prevent. `hold` does not write the transfer (only `complete` at 779 and `fold` at 711 do), so registering A&#39;s hold first does not clear it either. The only escape is to run `complete X &lt;A&gt;` before the fold and again after, inverting the documented order (skill Operating sequence puts fold at step 3 and complete at step 4). Suggest an explicit no-coverage attestation on `fold` - e.g. `--covers-no-status-decision` - so the strict `--key` requirement is kept while a fold that genuinely covers nothing can still say so.
- ⚠️ `.agents/skills/decision-hold-lifecycle/SKILL.md:23` - The skill now tells the agent to pass `--key` on every fold, but `--key` is rejected whenever the origin has no open status decision, so following the documented flow verbatim fails in the common shapes. `fold_covered_keys` (391-406) fails with &#34;origin X has no open structured decision under key K&#34; when the key is not in `origin_open_decisions`&#39; output, and that output is empty both when the origin has no status file at all (the shape in tests/fm-decision-hold-lifecycle.test.sh:607, which folds unqualified) and - more importantly - when the status ends in a terminal verb: `origin_open_decisions` (bin/fm-decision-hold.sh:357-369) returns empty for a non-secondmate origin whose last status line is `done:` or `failed:`, which is precisely the completed-investigation shape this skill is invoked on (see the `needs-decision` ... `done:` fixture at tests/fm-decision-hold-lifecycle.test.sh:147-151). SKILL.md:23 (&#34;naming the origin&#39;s open decision key with `--key`&#34;) and :44 (&#34;naming the covered decision key with `--key`&#34;) should condition the instruction on the origin actually having an open status decision, and tell the agent how to see that set, so the documented flow matches the script&#39;s contract.

🔧 Fix: decouple folding from status accounting at sign-off
2 infos still open:
- ℹ️ `bin/fm-decision-hold.sh:384` - `status_transfer_recorded` now matches `captain-held [key=&lt;k&gt;]: ` as a substring anywhere in a line rather than at line start, so an ordinary status note that happens to contain that text would let `--folded-key &lt;k&gt;` be accepted for a key that was never open. Path: a crewmate writes `working: captain-held [key=ghost]: copied from the other task` into `state/X.status`; `_fm_decision_key` reads the prefix before the first colon so the verb stays `working` and `ghost` never enters the open set, but the liveness check at 758-766 falls through to `status_transfer_recorded`, which matches, and the third tolerance accepts `--folded-key ghost` — recording a folded-key attestation for a question that never existed. Narrow because it needs that literal in crew prose, but the anchor is free: match `^captain-held \[key=&lt;k&gt;\]: ` (a `grep -q` with the brackets escaped, or an awk index test) so only a real transfer event satisfies it.
- ℹ️ `bin/fm-decision-hold.sh:795` - For a key accounted for as folded, the transfer event is written as `tracked by $folds`, and `folds` is the comma-separated union of every fold the origin recorded — so with two or more folds the durable status line reads `captain-held [key=K]: tracked by H1,H2` and cannot say which decision actually carries K. Exact in the single-fold case (what the new test asserts), and harmless mechanically: nothing parses the text after `tracked by` — `status_open_decisions` and `status_is_paused_or_captain_held` key off the verb only — and docs/decision-hold-lifecycle.md:45 documents it as &#34;the recorded folds&#34;. Noted only because the status log is a durable audit record; if per-key precision is ever wanted, `--folded-key` would need to carry the owning hold id.

</details>

<details>

✅ No issues found.
- <code>Replicated the `macos-stock-bash` CI lane verbatim under `/bin/bash` 3.2.57 (parse sweep over `bin/fm-lint.sh --list-files`, then both pinned suites) — lane exit 0</code>
- <code>`/bin/bash tests/fm-fleet-snapshot-view.test.sh` → 15 `ok -` lines, matches the lane&#39;s pin of 15</code>
- <code>`/bin/bash tests/fm-bearings-snapshot.test.sh` → 42 `ok -` lines, matches the new pin of 42; verified the base pin of 41 would emit `::error::expected 41 Bearings tests, got 42`</code>
- <code>`bash tests/fm-decision-hold-lifecycle.test.sh` → 17 cases pass, exit 0 (targeted product suite for the four decision-queue defects)</code>
- <code>Manual CLI render of the ageing decision: `FM_HOME=&lt;fixture&gt; FM_BEARINGS_NOW=2026-08-02T12:00:00Z /bin/bash bin/fm-bearings-snapshot.sh` showing `waiting_days,aging` and the ageing decision leading `decisions_open`</code>
- <code>Manual end-to-end `bin/fm-decision-hold.sh` transcript: `hold` (register), duplicate-key `hold` refused with exit 3, `fold --into` happy path plus `decision_folds=` recorded in origin metadata, `open` / `open --aging-only` at and below the ageing threshold, single-call `resolve`, then `open` showing a cleared queue</code>
- <code>Privacy scan of `git diff cf95112..f705263` and of the commit messages for concrete deployment identifiers — no hits</code>
- <code>`git status --porcelain` — worktree clean, no transient artifacts left behind</code>

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

### DIFF
--- a/.agents/skills/ask-user-authority/SKILL.md
+++ b/.agents/skills/ask-user-authority/SKILL.md
@@ -1,8 +1,8 @@
 ---
 name: ask-user-authority
 description: >-
-  Agent-only decision procedure for ask-user findings.
-  Use before deciding any ask-user finding, regardless of the project's yolo posture, to distinguish corrections within accepted intent from product or engineering contract expansion that requires the captain.
+  Agent-only decision procedure for ask-user findings and for any other question about to be escalated to the captain.
+  Use before deciding any ask-user finding, regardless of the project's yolo posture, and before registering a captain decision from an investigation or review, to distinguish corrections within accepted intent from product or engineering contract expansion that requires the captain.
 user-invocable: false
 metadata:
   internal: true
@@ -10,8 +10,41 @@ metadata:
 
 # ask-user-authority
 
-This skill is the single owner of the decision procedure for ask-user findings.
+This skill is the single owner of the escalate-or-decide boundary.
+It governs every question about to reach the captain, whether it arrived as an ask-user finding from a validation gate or as an unresolved choice found by an investigation or review, which `decision-hold-lifecycle` routes here before registering anything.
 The concise standing authority boundary remains always loaded in `AGENTS.md` section 7.
+
+## The bar for asking
+
+Escalation is expensive and the queue compounds, so the default is to decide.
+Ask only when the answer is genuinely not derivable from the accepted contract, the code, and the evidence in front of you.
+
+Apply this test before escalating anything:
+
+1. Is there a defensible answer a careful engineer would reach from the accepted contract and the evidence?
+   If yes, take it and record the reasoning; a difficult or debatable call is still yours.
+2. Would the two candidate answers commit the project to materially different products, costs, or accepted risks?
+   If yes, it is the captain's.
+3. Does answering it require preference the evidence cannot supply, such as which of two defensible products to build, how much risk or spend to accept, or what an external commitment should say?
+   If yes, it is the captain's.
+4. Is it destructive, irreversible, or genuinely security-sensitive?
+   If yes, it is the captain's under the stronger existing boundary.
+
+An engineering trade-off with a defensible answer is Firstmate's call even when the trade-off is real, the alternatives are close, or getting it wrong would cost rework.
+Difficulty, reviewer emphasis, and the presence of two options are not evidence that a question belongs to the captain.
+Uncertainty about which answer is best is the ordinary condition of engineering work, not grounds to escalate.
+
+Worked examples on both sides:
+
+- The order in which two related pull requests should land, where one order is clearly safer, is Firstmate's call: the evidence settles it and neither order changes what gets built.
+- Where a piece of infrastructure should live within an accepted architecture is Firstmate's call, even when two placements are defensible, because the choice does not change the product's behavior or its accepted risk.
+- Whether to fix a defect in place or split it across two pull requests is Firstmate's call when both routes deliver the same accepted behavior; it becomes the captain's only if splitting changes what ships or when.
+- Which of two documents is authoritative for a team's rules is the captain's: it is a standing policy choice about how the team works, and the evidence cannot rank the options.
+- Accepting a known risk in order to ship sooner is the captain's, because it trades product risk for time.
+- Anything that changes an outward-facing commitment, a cost, or a security posture is the captain's.
+
+When a question fails this bar, do not register it as a captain decision and do not park the work waiting for one.
+Decide it, record the reasoning where the work lives, and continue.
 
 ## Decide who has authority
 
--- a/.agents/skills/bearings/SKILL.md
+++ b/.agents/skills/bearings/SKILL.md
@@ -71,6 +71,7 @@ Every `/bearings` chat response renders EXACTLY these four sections, in THIS ord
 
 1. **Captain's Call** - ONLY items that need the captain's own action now: a decision to make, a PR to approve or merge, a credential or login to provide, or a blocker only the captain can clear.
    Empty-state: "Nothing needs your action right now."
+   A decision the snapshot marks `aging` leads this section and renders as still waiting since its date, for example "still waiting since 3 March"; it is never re-asked as though it were new.
 2. **Recently Landed** - the bounded current recent-completions baseline: merged PRs, completed scouts, and finished local-only merges across the main fleet and every registered secondmate home.
    Empty-state: "No recent completions are in the current baseline."
 3. **Underway** - live work progressing on its own, one line of current state per direct report.
--- a/.agents/skills/decision-hold-lifecycle/SKILL.md
+++ b/.agents/skills/decision-hold-lifecycle/SKILL.md
@@ -16,12 +16,24 @@ This skill is the single policy owner for unresolved captain decisions discovere
 
 Every unresolved decision that belongs to the captain and is discovered while producing, reading, presenting, or ending an investigation or visual review must become a structured captain-held work item in the authoritative backlog of the home that owns the originating work before that work or review may be treated as complete.
 The agent performs the semantic inventory because scripts must not infer decisions from report prose, visual-review artifacts, terminal output, or chat.
-Give each distinct unresolved decision a stable privacy-safe key, register it through `bin/fm-decision-hold.sh hold`, and use the same key on retry so registration is idempotent while different decisions retain different durable identities.
+Apply `ask-user-authority`'s bar for asking to every candidate first; a question that fails that bar is Firstmate's to decide and must not be registered here at all.
+Give each remaining distinct unresolved decision a stable privacy-safe key, register it through `bin/fm-decision-hold.sh hold`, and use the same key on retry so registration is idempotent while different decisions retain different durable identities.
+Before a new identity is created the script puts the home's open captain decisions in front of you and refuses, because a second key covering an already-open question is how this queue compounds.
+Read that listing and judge it semantically: an earlier review of the same pull request, subsystem, or policy question usually already owns the decision under a different key.
+When it does, use `bin/fm-decision-hold.sh fold` to add this pass's finding to the existing decision instead of opening a second one, then complete against that fold.
+Re-run with `--distinct` only when the question is genuinely different from every decision in that listing; the attestation is recorded in the new decision.
 After inventorying the whole report and review surface, run `bin/fm-decision-hold.sh complete` with every unresolved key, or with `--none` only when the reviewed surface contains no unresolved captain decision.
 A completed investigation and an ended visual review use this same owner and completion command; a visual tool, including Lavish, never owns a parallel completion policy.
 Run the command in the originating work's authoritative `FM_HOME`; main-home work creates main-home holds, and secondmate-owned work creates holds in that secondmate home's backlog rather than copying them into the main backlog.
 Do not close a hold merely because the originating investigation completed, its report was archived, its visual review ended, or its task was torn down.
-When the captain's answer authorizes follow-up work, the hold remains the authoritative Captain's Call item until that answer is durably recorded, dependent work is created in the same backlog and blocked by the hold, and `bin/fm-decision-hold.sh resolve` routes the answer by clearing those dependency edges before closing the hold.
+The hold remains the authoritative Captain's Call item until the captain's answer is durably recorded by `bin/fm-decision-hold.sh resolve`, which closes it.
+Record the answer in the same turn it arrives.
+An answer that lives only in a chat message is worse than no answer, because it acquires authority it cannot evidence: later work cites it as settled with nothing to point at, and the decision stays open and gets asked again.
+Recording is deliberately cheap so nothing stands between hearing the answer and writing it down - pass the captain's words to `resolve` with `--decision`, and use `--no-routed-work` when the answer spawns no dependent task.
+Two closures other than a fresh captain answer are equally first class: a question that turns out to be firstmate's own call under `ask-user-authority`'s bar closes with `--decided-by firstmate`, and one an earlier captain answer already settled closes with that answer and its source as the decision text.
+Neither is a shortcut around the captain; each is the honest record of who actually decided.
+A decision closed with a real record but not through this script still counts as durably resolved, so an earlier bulk close does not strand the investigation that found it.
+Where dependent work genuinely exists, create it in the same backlog, block it by the hold identity, and route it with `--routed-to`; the script refuses `--no-routed-work` while any task is blocked by the hold, so that requirement cannot be skipped.
 When the captain's answer routes no follow-up work at all, such as a declined proposal, `bin/fm-decision-hold.sh decline` records that answer and closes the hold; it never substitutes for routing work the captain did authorize.
 When the captain simply answers a hold that has no follow-up work routed behind it yet, `bin/fm-decision-hold.sh answer` records that answer and closes the hold, so answering is closing rather than a separate later act that can be forgotten.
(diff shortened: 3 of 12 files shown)


timidri:
Maintainer check-in: this one is finished and mergeable, but every workflow run on it sits at `action_required` and none has ever started — our access here is read-only, so a maintainer has to release them before any check can report.

Could someone release the runs when convenient? That's all it needs; no write access required on our side.

Happy to rebase or split it if that helps review.

timidri:
@kunchenguid — this one has been open since 3 August; ready for review whenever it fits.

PR #1590, opened 2026-08-03
fix(podman): verify credential seeding and support live re-seed

anbclausen:
Podman workers frequently booted with an expired login and froze at `Login expired · Run /login`, doing no work, while firstmate saw a healthy container and pane.

**Root cause**: `bin/backends/podman.sh` seeded credentials with `podman cp "$HOME/.claude/." <container>:/home/agent/.claude` followed by `|| true`. A failed copy - or one capturing a stale/truncated file mid credential refresh - was completely silent. Observed 2026-08-03: a container held a stale 276-byte `.credentials.json` against the primary's valid 504-byte file; the agent cached the bad token at startup and never authenticated.

## Changes

- **Verified seeding** (`fm_backend_podman_seed_credentials`): copies the primary's current `~/.claude`, then requires the container's `.credentials.json` to match the primary's byte count, retrying (`FM_BACKEND_PODMAN_SEED_RETRIES`, default 3) to ride out a concurrent refresh. A copy failure, a size mismatch, or a truncated file on the primary removes the container and fails the spawn with a diagnostic. Absent host credentials stay a warning, not an error (API-key auth, non-claude harnesses).
- **Live re-seed** (`bin/fm-podman-reseed.sh <id> [--restart]`): re-credentials a running container from the primary's current file and, with `--restart`, relaunches its agent from the launch command now recorded in podman task metadata (`launch=`, written only for podman spawns) - no teardown, worktree preserved.
- **Docs** (`docs/podman-backend.md`): new "Credential seeding" section with a "Known issue: worker boots with expired login" + recovery subsection covering the symptom, root cause, the re-seed path, the `bin/fm-teardown.sh <id> --force` + respawn fallback, and why not to hand-copy the credential file.
- **Tests**: `tests/fm-backend-podman.test.sh` gains four seeding cases (match accepted; mismatch, copy failure, and truncated primary file all refuse loudly).

## Verification

`tests/fm-backend-podman.test.sh` passes (19 checks). `bash -n` clean on all three changed scripts. `bin/fm-lint.sh` could not run in this environment (ShellCheck 0.11.0 not installed; it refuses any other version) - CI's lint job covers it. `bin/fm-doc-audience-check.sh` reports only a pre-existing, unrelated `docs/wedge-alarm.md` README-link failure, present on `main` too.

Validated by review and tests rather than the no-mistakes pipeline, at the captain's instruction.


anbclausen:
Sorry, my crewmate wanted to contribute :))

PR #1595, opened 2026-08-03
docs: require deliberate context discipline

pa-arth:
## What

Adds one context-discipline rule to firstmate's two instruction surfaces, so agents keep their own context small and compact at a boundary they choose.

Auto-compact fires on a token threshold, not a semantic boundary. It routinely lands mid-gate, mid-investigation, or mid-recovery, and the summary keeps the prose while dropping the operational state the agent needed next. With auto-compact off at the harness level, the failure mode flips: an agent that does not manage its own context runs out of it rather than being rescued.

## Where, and why only there

Two audiences, one place each - duplicated instructions drift.

- **`AGENTS.md` section 1** for firstmate. Placed at the tail of "Identity and prime directives", which already carries standing self-conduct that is not a numbered hard rule. Section 3 was rejected because its heading is scoped "run once at every session start"; section 8 was rejected because it is fleet supervision, where the rule would read as being about supervising crew rather than about the agent's own context.
- **`bin/fm-brief.sh`'s `# Rules`, as rule 8** for crewmates, who never read `AGENTS.md`.

Ship and scout share a single `CONTEXT_RULE` shell variable, so the two rule lists cannot diverge. The secondmate charter is deliberately excluded: a secondmate is a firstmate in its own home and reads its own `AGENTS.md`. The new test asserts that exclusion.

Also corrects a line in `.agents/skills/stuck-crewmate-recovery/SKILL.md` that had become false. It told firstmate "A low context reading is not wedging; modern harnesses auto-compact and keep going." With auto-compact off, that is the exact rule firstmate applies when deciding whether to help a crewmate approaching exhaustion, so leaving it would have made this change actively counterproductive. The correction is that firstmate should *notice*, not that firstmate should intervene: the existing guarantee that a low reading is never grounds to interrupt, exit, or relaunch a working crewmate is preserved explicitly.

## The rule states a principle and names no mechanism

Two review rounds established this the hard way. Earlier drafts named a "how" - a scratch notes file, the `stow` skill, a recovery step - and every named mechanism collided with a contract that already owned it: the scratch file tripped `fm-teardown.sh`'s dirty check, the `stow` pointer summoned its mandatory complete pass at every seam, and the recovery-step pointer routed a nudge through an *interrupt*, landing mid-edit - precisely the "partway through" moment this change exists to prevent.

Three for three is not bad luck, so the mechanisms are gone. The rule says what must be true and stops: keep context small; compact at a seam you choose; write down what must survive before you compact; no threshold and no cadence. It names no destination file, no skill, and no recovery step. Notably, no teardown exemption was added to accommodate a scratch file - that would have traded a rule collision for a permanent hole in the unlanded-work check.

No machinery: no context-measuring script, no hook, no threshold checker. No safety boundary was reworded or reordered.

## Rendered `# Rules` section from a scaffolded ship brief

```
7. Never stop, restart, or update the shared `no-mistakes` daemon - it is one instance serving
   every lane/home, so restarting it kills other lanes' in-flight pipeline runs. On ANY no-mistakes
   daemon error, append `blocked: {the daemon error}` and stop; only firstmate manages the daemon.
8. Keep your context small and compact it deliberately. Read what the task needs, not what sits
   near it, and point at an authoritative file instead of copying it. Compact at a seam you
   choose - a finished investigation, a landed step, a closed decision - never partway through
   one. Anything that must survive (findings, a decision, a blocker) is written down BEFORE you
   compact, never only in this chat. No threshold, no cadence: compact when you judge it right.
```

The scout contract renders rule 8 identically; the secondmate charter omits it.

## Verification

- New behavior test `test_context_discipline_rule_reaches_every_crewmate_scaffold` asserts the rule renders in all three ship delivery modes and in the scout contract, and is absent from the secondmate charter. **Proven to fail against the pre-change script** (restored `bin/fm-brief.sh` from `HEAD`, reran, got `not ok - brief-ctx-nomistakes: ship brief lost the context-discipline rule`) - a green run on unchanged code would have proved nothing.
- `tests/fm-brief.test.sh` green.
- `bin/fm-lint.sh` **could not run locally**: ShellCheck is not installed on this machine, and installing it would have reached outside the worktree. CI owns that gate.

### DIFF
--- a/.agents/skills/stuck-crewmate-recovery/SKILL.md
+++ b/.agents/skills/stuck-crewmate-recovery/SKILL.md
@@ -44,7 +44,6 @@ Escalate in order:
    For example, for a single-Escape adapter: `FM_HOME=<this-firstmate-home> bin/fm-send.sh <window> --key Escape`.
 4. If the crewmate is genuinely wedged after redirection, exit the agent with the adapter's exit command and relaunch with the same brief plus a `progress so far` note appended to it.
    Genuine wedging means looping, unresponsive, repeating the same obstacle, or truly dead.
-   A low context reading is not wedging by itself and never justifies exiting a working crewmate.
-   The fleet runs with harness auto-compact off, so treat it as a cue to redirect through step 3: tell the crewmate to write its live state to its durable file and compact deliberately.
+   A low context reading is not wedging and is never grounds to interrupt, exit, or relaunch a working crewmate, but with harness auto-compact off it is no longer harmless, so note it.
    The worktree and commits persist, so relaunch is cheap.
 5. If a second relaunch fails too, write `failed` to the backlog and tell the captain the plain failure, preserved work, and consequence using `AGENTS.md` section 9; do not mention metadata, harness, window, or worktree unless the path itself is needed for action.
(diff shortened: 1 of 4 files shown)


kunchenguid:

PR #1592, opened 2026-08-03
fix(bin): stop the Claude turn-end guard blocking stops it cannot repair

pa-arth:
## Intent

Raise the turn-end guard fix compliantly through the gate per CONTRIBUTING.md, so the repo's 'PR must be raised via no-mistakes' check passes. An earlier PR (#1592) for this same change was opened by hand with gh and fails that check.

The branch fixes three defects in bin/fm-turnend-guard.sh's --claude mode, in two commits.

Commit 1 (6938bc4), the original work:
- A session that does not hold this home's session lock re-blocked on every turn for ~15 consecutive turns with three pieces of work in flight, while the lock-owning session sat idle. bin/fm-claude-stop-autoarm.sh refuses to arm from a non-lock-owning session, so blocking it demanded a repair it could never perform. The guard now allows that stop with an advisory.
- The advertised ceiling of FM_CLAUDE_TURNEND_BLOCK_BUDGET (default 3) consecutive blocks never arrived, because a stalled auto-arm stops rewriting state/.claude-autoarm-epoch and the epoch-gated counter froze in exactly the stalled case the bound exists to cover. A genuine block now consumes one count per blocked stop; a BUDGET_ADVANCED flag keeps a single stop accounted at most once in total, and the allow paths keep their per-epoch idempotence.

Commit 2 (5b4b8ac), fixing what a previous run of this pipeline correctly caught in commit 1:
- The foreign-lock allow used '! fm_session_lock_owned_by_self', which also reports failure when this process's harness ancestry cannot be resolved, so it conflated 'someone else owns the lock' with 'I cannot tell who owns it'. With an unresolvable ancestry and state/.lock holding this session's OWN live harness pid, the guard allowed a blind stop for the one session permitted to repair supervision. bin/fm-session-lock-lib.sh now owns the decision through one lock-file reader plus fm_session_lock_live_foreign_owner, which requires complete positive evidence including a RESOLVED ancestry; a missing lock, malformed lock, dead owner, and unresolvable ancestry all keep the ordinary blocking path.
- The advisory asserted 'The lock-owning session restores supervision when it ends its next turn'. That is the assumption the motivating incident disproved, since an idle lock owner has no next turn, and it is wrong under away mode where the away daemon owns the watcher. It now states only what was verified and takes its recovery step from bin/fm-supervision-instructions.sh with the same --afk and --x-mode inputs block_stop uses.
- docs/turnend-guard.md now describes the corrected predicate rather than the intended one.

[...]

PR #1597, opened 2026-08-03
fix(pi): stop the turn-end guard follow-up wedging a Pi primary

Mani212005:
A firstmate PRIMARY session running on the Pi harness hung forever on "Working..." for any input, including `hi` — a collapsed tool call, no error, no stderr, no timeout.

## Root cause

`pi.sendUserMessage(content, { deliverAs: "followUp" })` called from inside an awaited extension handler does not queue a follow-up. It starts a fully re-entrant nested agent run inside the parent run's still-open stack frame.

Verified against the installed Pi SDK, version 0.83.0:

- `core/agent-session.js` — `_emitAgentSettled()` sets `this._isAgentRunActive = false` and *then* awaits `this._extensionRunner.emit({ type: "agent_settled" })`.
- `core/agent-session.js` — `get isStreaming() { return this._isAgentRunActive; }`, so inside an `agent_settled` handler the session reports NOT streaming.
- `core/agent-session.js` — `prompt()` honors `options.streamingBehavior` (what `deliverAs` maps to) only inside `if (this.isStreaming) { ... return; }`. Not streaming means the follow-up queueing branch is skipped entirely.
- `core/agent-session.js` — execution falls through to `_runAgentPrompt()`, which sets `_isAgentRunActive = true`, runs a whole new agent turn, and in its `finally` calls `_emitAgentSettled()` again.
- `core/extensions/runner.js` — `emit()` awaits each extension's handler serially, so one slow settled handler holds the run frame open for *every* extension. This was established during this work and matters for the watcher race below.

The trigger: `bin/fm-turnend-guard.sh` returns exit 2 deterministically whenever tasks are in flight and no healthy watcher holds the home lock. That condition fired the re-entrant send on every single turn.

### 1. New owner for follow-up delivery

`.pi/extensions/lib/fm-followup-delivery.ts` is the single owner of when an operational follow-up may be sent:

- `withSettledFrame(body)` marks an `agent_settled` handler as holding Pi's run frame open.
- `deliverOperationalFollowUp(pi, content)` holds a delivery while any Firstmate settled frame is open, then sends from a fresh macrotask once every frame has closed.

Both Pi extensions import the same module, so the frame count is shared in the extension host process.

**A plain macrotask hop is not sufficient, and this was proven rather than assumed.** The first implementation deferred by a macrotask only; the watcher test written for this change failed against it. The turn-end guard's helper spawn genuinely holds Pi's parent `_runAgentPrompt` frame open for hundreds of milliseconds, so a watcher arm-child close landing in that window still nested. The frame gate is what makes delivery safe.

`deliverAs: "followUp"` is still passed deliberately: a send can legitimately land while some *other* run is genuinely streaming, and without it Pi throws `Agent is already processing`. That rationale now lives in the delivery module header.

### 2. `fm-primary-turnend-guard.ts`

- The `agent_settled` handler body runs inside `withSettledFrame`, and delivery is no longer awaited inside it. The existing one-shot `guardFollowupActive` semantics are preserved, so the follow-up turn does not re-trigger the guard.
- **Consecutive-block ceiling.** `bin/fm-turnend-guard.sh` applies `FM_CLAUDE_TURNEND_BLOCK_BUDGET` only in `--claude` mode; the Pi path reaches `block_stop()` and returns 2 before any budget logic runs, so on Pi there was no ceiling at all. After `FM_PI_TURNEND_BLOCK_BUDGET` (default 3) consecutive follow-ups the extension stops re-prompting and posts one visible, non-blocking transcript warning instead. Any healthy turn (guard exit 0) resets both the count and the warning.
- **Spawn timeouts.** `runGuard()` and `runChecker()` (used by both `runPretoolCheck` and `runCdCheck`) now share one `collectSpawn` helper with a hard budget — `FM_PI_TURNEND_GUARD_TIMEOUT_MS` (15000) and `FM_PI_PRETOOL_CHECK_TIMEOUT_MS` (10000). On expiry the child is killed and the result is treated as *allow*, matching the fail-open posture `bin/fm-turnend-guard.sh` documents in its own header. An unbounded spawn awaited inside a hook is the same bug class as the re-entrant send: `runChecker` is awaited inside the `tool_call` hook, so a wedged checker would hang a tool call the same way.

### 3. `fm-primary-pi-watch.ts`

`sendWake()` routes through the shared delivery module. Watcher arm, retry, continuity, and the session-generation ownership rules in its header are unchanged.

### 4. Supporting changes

- `positiveInteger` extracted from `fm-primary-pi-watch.ts` into `.pi/extensions/lib/fm-extension-env.ts` so both extensions share one owner.
- `.agents/skills/harness-adapters/SKILL.md` still documented the raw `pi.sendUserMessage(..., { deliverAs: "followUp" })` call as the correct Pi primary pattern — exactly the call this change removes. A future agent reading it would have reintroduced the bug. It now points at `lib/fm-followup-delivery.ts` as the owner and states the failure mode.

### Scope

**No file under `bin/` is modified.** This is deliberate: PR #1592 is concurrently rewriting the block-budget logic in `bin/fm-turnend-guard.sh`, so the ceiling is implemented in the extension rather than in the guard script where `FM_CLAUDE_TURNEND_BLOCK_BUDGET` already lives.

## How this was verified

Tests are colocated per `firstmate-coding-guidelines` and extend the existing suites rather than adding a runner. The shared Pi session fake lives in the new `tests/pi-extension-helpers.sh`, using the established `tests/*-helpers.sh` convention that `bin/fm-test-run.sh` already maps.

The fake reproduces the real SDK ordering: the run flag is cleared before the settled emit, several handlers share one open frame, and a send from a settled frame starts a nested run. It counts `reentrantRuns` — agent runs started while another run frame is still open — which is precisely the defect.

### Fail-before / pass-after evidence

Each fix was individually reverted to confirm its test actually catches the defect:

| Change reverted | Result |
| --- | --- |
| Deferred delivery in the guard | `not ok — Pi guard follow-up must be delivered outside the agent_settled frame` |
| Deferred delivery in the watcher | `Error: the wake started 1 nested agent runs` |
| Consecutive-block ceiling | `not ok — Pi guard must bound consecutive follow-ups and reset on a healthy turn` |
| Spawn timeouts | the suite **hangs** past a 120s timeout — the bug shape itself |

With the fixes in place all of these pass.

An earlier iteration is worth recording: the watcher test failed against a macrotask-only fix with `1 run frames stayed open`, which is what drove the design to the frame gate.

### Suites run

- `tests/fm-turnend-guard.test.sh` — 66 ok, 0 failures
- `tests/fm-pi-watch-extension.test.sh` — 21 ok vs 20 on the untouched base commit (the +1 is the new test), 0 new failures
- `tests/fm-session-start.test.sh`, `tests/fm-supervision-instructions.test.sh`, `tests/fm-test-run.test.sh`, `tests/fm-documentation-audiences.test.sh`, `tests/fm-lint.test.sh`, `tests/fm-startup-memory-budget.test.sh` — all clean
- `bin/fm-lint.sh` — clean against the pinned ShellCheck 0.11.0
- `bin/fm-doc-audience-check.sh` — ok, 61 surfaces, 175 local links
- `bin/fm-test-run.sh --changed` — resolves the new files correctly, no unmapped-path refusal

### Documentation

Routed to existing owners rather than restated inline: `docs/configuration.md` for the three new environment variables, `docs/turnend-guard.md` for the Pi adapter behavior and regression pointers, `docs/verification/supervision.md` for the empirical Pi 0.83.0 SDK facts.

## What could NOT be verified

Stated plainly:

1. **No real Pi primary was driven.** All verification is against the SDK source plus the session fake. The opt-in `FM_PI_LIVE_E2E=1 tests/fm-pi-primary-live-e2e.test.sh` path has its fixtures updated for the new modules but was not run.
2. **The frame gate covers Firstmate's own settled handlers only.** A third-party extension registering a long-awaiting `agent_settled` handler would re-open the same window, and no extension API can observe that. This limit is documented in the module header.
3. **`no-mistakes` did not run.** The pipeline refused to launch because the globally configured gate agent is `pi`, which it will not run against a repo with `AGENTS.md`/`CLAUDE.md`. The run failed at launch with zero steps executed. This PR was opened on an explicitly approved direct-PR path instead.
4. **Pre-existing failures in the development environment, not caused by this change** — each reproduces identically on the untouched base commit:
   - `tests/fm-pi-watch-extension.test.sh` — roughly five Pi and five OpenCode arm-recovery tests (e.g. `expected one successor plus two retries, got 1`). Consistent, not flaky. They were excluded from the pass/fail comparison above on both sides equally.
   - `tests/fm-calm-pi-extension.test.sh` — `/export did not complete while calm mode was on`.
   - `tests/fm-pi-primary-types.test.sh` — two `fm-calm.ts` type errors against Pi 0.83.0. The files changed here typecheck clean.
5. **A verification claim was withdrawn during this work.** An early draft of the `docs/verification/supervision.md` entry recorded an exact `bin/fm-turnend-guard.sh` exit-2 reproduction command. Attempting it in a synthetic primary fixture returned exit 0 rather than 2, so rather than record output that was not actually observed, that block was replaced with a pointer to the regression test that owns the predicate. The SDK facts in that record were each read directly from the installed `dist/` and stand.


Mani212005:
Superseded: re-raising this change through the no-mistakes pipeline so it carries the required pipeline signature. Same commits, same branch. Closing to free the head branch for the pipeline-raised PR.

PR #1628, opened 2026-08-04
fix(bin): unblock session lock and CLAIM_LOCK on Windows Git Bash

campbell-reid-displayr:
## Summary

Firstmate is permanently stuck read-only on Windows Git Bash (MSYS).

- **`bin/fm-session-lock-lib.sh`**: the harness-ancestry walk depends entirely on `ps -o comm=/-o ppid=/-o args=`. Git Bash's bundled Cygwin `ps` (3.6.9) doesn't support `-o` at all (`ps: unknown option -- o`), so the walk dies on hop one and `fm-lock.sh` always reports `cannot locate harness process in ancestry`. `fm-harness.sh` already trusts verified env markers (`CLAUDECODE=1`, `PI_CODING_AGENT`, `GROK_AGENT`) ahead of the `ps` walk; the session lock had no equivalent fallback. It now probes `ps -o` support once and falls back to the same verified-marker trust when unavailable.

This fallback is additive: `ps -o`-capable platforms (Linux, macOS - CI's only runners) take the exact same code path as before.

Separately, the session lock's `CLAIM_LOCK` mutex depends on `ln -s` producing a real symlink, which requires Windows Developer Mode (or the "Create symbolic links" privilege) to be enabled - without it `ln -s` on a directory silently creates a plain empty directory instead, and the lock can never be verified as owned. Rather than work around this in code, it's now documented as a hard requirement in the README.

## Verified

- `fm-lock.sh` now acquires instantly instead of hanging; `fm-session-start.sh` no longer falls into the read-only branch.
- `tests/fm-session-lock-ancestry.test.sh`: all pass, including a new case for the broken-`ps` fallback.
- CI only runs `ubuntu-latest`/`macos-latest`, so none of this was previously exercised on Windows.

## Test plan

- [ ] CI (Linux/macOS) stays green - the new code path is gated behind a capability probe that should evaluate to "supported" there, so behavior is unchanged.
- [ ] On Windows Git Bash (Developer Mode enabled): `bash bin/fm-session-start.sh` acquires the lock without a READ-ONLY banner.

### DIFF
--- a/bin/fm-session-lock-lib.sh
+++ b/bin/fm-session-lock-lib.sh
@@ -74,6 +74,38 @@ fm_harness_process_matches() {  # <comm> <args>
   return 1
 }
 
+# Cached result of whether this platform's `ps` understands the POSIX -o
+# custom-format option (`-o comm=`, `-o ppid=`, `-o args=`), which every
+# function below depends on to read process identity. Verified broken: the
+# Cygwin ps 3.6.9 shipped with Git for Windows supports only
+# [-aefls] [-u UID] [-p PID] and rejects -o outright, so every comm/args/ppid
+# read above silently comes back empty (stderr is normally discarded) and the
+# ancestry walk fails on its very first hop. Probed once and cached because ps
+# is invoked many times per session.
+FM_PS_SUPPORTS_O=
+fm_ps_supports_o() {
+  if [ -z "$FM_PS_SUPPORTS_O" ]; then
+    if ps -o comm= -p $$ >/dev/null 2>&1; then
+      FM_PS_SUPPORTS_O=1
+    else
+      FM_PS_SUPPORTS_O=0
+    fi
+  fi
+  [ "$FM_PS_SUPPORTS_O" = 1 ]
+}
+
+# True when this process's environment carries one of the verified harness
+# markers fm-harness.sh's detect_own() trusts as its Layer 1: only claude, pi,
+# and grok set one of their own (codex, opencode, and kimi are markerless, so
+# they have no fallback here and still depend entirely on the ps ancestry walk
+# below). Keep in sync with detect_own() by hand - see comment there.
+fm_harness_verified_env_marker() {
+  [ "${CLAUDECODE:-}" = "1" ] && return 0
+  [ "${PI_CODING_AGENT:-}" = "true" ] && return 0
+  [ "${GROK_AGENT:-}" = "1" ] && return 0
+  return 1
+}
+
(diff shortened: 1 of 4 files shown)


bin/fm-session-lock-lib.sh
```diff
@@ -74,6 +74,38 @@ fm_harness_process_matches() {  # <comm> <args>
...
+# function below depends on to read process identity. Verified broken: the
+# Cygwin ps 3.6.9 shipped with Git for Windows supports only
+# [-aefls] [-u UID] [-p PID] and rejects -o outright, so every comm/args/ppid
+# read above silently comes back empty (stderr is normally discarded) and the
+# ancestry walk fails on its very first hop. Probed once and cached because ps
+# is invoked many times per session.
+FM_PS_SUPPORTS_O=
+fm_ps_supports_o() {
+  if [ -z "$FM_PS_SUPPORTS_O" ]; then
+    if ps -o comm= -p $$ >/dev/null 2>&1; then
```
campbell-reid-displayr, line 88: Probes with `-p $$` (this process's own pid, always guaranteed to exist) rather than an arbitrary pid, so a `-p` lookup failure can only mean "no `-o` support" - never "pid not found" muddying the signal.

bin/fm-session-lock-lib.sh
```diff
@@ -94,6 +126,14 @@ fm_harness_process_matches() {  # <comm> <args>
 # reported and the callers below decide what they need from it.
 fm_harness_ancestry_pids() {
   local pid=$$ comm args extending=0 printed=0
+  if ! fm_ps_supports_o && fm_harness_verified_env_marker; then
```
campbell-reid-displayr, line 129: Worth flagging for future readers: on this fallback path the reported pid is just `$$` - whatever process happens to be executing `fm-lock.sh` right now - not a walk up to a longer-lived ancestor, since there's no `ps -o ppid=` available to walk with. Verified empirically on a real Cygwin/Git-Bash box: that pid is already dead a moment after `fm-lock.sh` exits, so `fm_harness_pid_alive` on a previously-recorded lock owner always reports "dead" here, and the "another live session holds the lock" check in `fm-lock.sh` never fires on this platform - every acquisition attempt just claims a fresh pid instead of ever refusing. Doesn't affect this PR's goal (unblocking the permanent hang), but the concurrent-session protection is effectively a no-op wherever this fallback is active, which seems worth a follow-up issue rather than silently living with it.

campbell-reid-displayr (commented):
A few inline notes on the trickier parts of the two fallbacks - no blocking concerns, just context for future readers.

Issue #1506, opened 2026-08-02
Away-mode escalation undelivered for 3.16 days: discovery falls back to a crew shell pane, and the dead-man's alarm
  writes only to the failing channel


tkt028:
## Why this matters

**Away-mode escalation to the operator's terminal was undelivered for 3.16 days, and the alarm about it was written only to a log file.**

Measured 2026-08-02 from `state/.supervise-daemon.log` on a live home:

```
[2026-08-02T18:45:43+0700] ERROR: away-mode escalation undelivered 273104s;
  inject could not confirm a submit (supervisor pane busy or wedged).
  Buffer + wake-queue preserved; alarm marker written.
```

`273104s` = **3 days 3 hours**. In that single daemon run: **85 deferrals, 0 deliveries.**

The dead-man's switch worked exactly as designed — it detected the outage, preserved the buffer, wrote the alarm. **It wrote the alarm to the same channel that was already failing.** An undeliverable-escalation alarm that is itself undeliverable cannot do its job, and that is why this went unnoticed for three days rather than three minutes.

### A. Target discovery fails, and the daemon falls back to a pane belonging to the crew

```
[2026-08-02T18:25:20+0700] daemon starting (pid 54106); target=firstmate:0;
  target_source=FALLBACK(firstmate:0); backend=tmux; backend_source=FALLBACK(tmux)
```

A healthy start, from the same home five days earlier:

```
daemon starting (pid 34974); target=%1; target_source=TMUX_PANE
```

**Trigger:** discovery reads `$TMUX_PANE`. When away-mode is armed from inside an agent session rather than a plain tmux shell, that variable is unset — verified directly (`TMUX_PANE=UNSET` in the arming context). Discovery then takes `FM_SUPERVISOR_TARGET_DEFAULT` (`bin/fm-supervisor-target-lib.sh:21`), consumed at `bin/fm-supervise-daemon.sh:1089`.

**`firstmate:0` is not the operator's terminal.** On the measured machine it resolves to:

```
1: zsh#* (1 panes) [80x24] ... @0 (active)
```

— a bare login shell inside the **crew** tmux session. Escalations were being aimed at a shell prompt with no reader.

**This is `absent` rendered as a value.** *"I could not discover the supervisor pane"* is `unknown`; the daemon converted it into a confident target and armed anyway. The `FALLBACK(...)` marker proves it knew it was guessing — it logged the uncertainty and then behaved as though certain.

[...]

PR #1654, opened 2026-08-04
fix(herdr): a labelled composer rule makes an idle claude pane read unknown, in front of the NBSP defect

eduardstan:
## The incident

Away-mode escalation undelivered for **21249s (5h54m)** on a herdr home, and ~8h in an earlier
stretch, against a **live, idle claude primary that the daemon had resolved correctly**:

```
daemon starting; target=default:w3:pJ; target_source=HERDR_ENV(HERDR_PANE_ID);
  backend=herdr; backend_source=HERDR_ENV
```

Deferral reasons that night:

```
2524  deferred: supervisor composer not confirmed-empty (state=unknown)
  57  deferred: supervisor pane busy (agent mid-turn)
```

The injector was right to refuse — an unreadable pane is genuinely unsafe to type into. The bug is
that the pane was readable and the classifier said otherwise.

## Two defects, stacked on one row

**1. Structural (this pull request's own finding).** `fm_backend_herdr_composer_state` treats a lone
trailing horizontal rule with no matching opening rule as proof that a composer row found above it is
stale. That heuristic is correct for Pi, whose composer sits between two *bare* separators. claude
frames its live composer the same way **except its upper rule carries a label** (`--- name --`), so it
is not a bare separator, the capture ends in exactly the incomplete-pair shape, and the live composer
is discarded before its content is ever examined.

The complete-pair branch immediately above already corroborates the Pi shape against herdr's native
agent identity; this branch did not. It now does: only a Pi pane's separator invalidates a match, and
an unreadable identity still invalidates — so **doubt keeps blocking injection exactly as before**.

**2. Content.** With the row admitted, claude's idle composer reads as the bare agent glyph plus
U+00A0 NO-BREAK SPACE (`e2 9d af c2 a0`), which is not in `[[:space:]]` in any glibc locale, so no
trim removes it and the exact-glyph cases miss.

**Defect 1 sits in front of defect 2.** On herdr, the composer is rejected structurally before
normalisation runs — so #1322 or #1285 landing alone still leaves a herdr home with a claude primary
reading `unknown` and still wedging.

## Attribution

The normalisation for defect 2 is **taken from #1322 by @KD5694**, which root-caused it byte by byte
against a 1562-deferral incident, together with that pull request's composer scan-window increase.
**#1285 carries an equivalent independent fix for the same defect; only one is applied.** Neither is
reimplemented here. If you would rather keep this in one place, I am happy to close this and fold the
structural half into #1322 instead — that pull request did the harder half.

Related: #1506 records the same symptom from a third cause (target discovery falling back to a crew
shell). Our discovery worked and we wedged anyway, so that root cause is sufficient but not necessary;
I have added our measurements there.

## Evidence

Verified on the exact pane that wedged, same pane and same moment:

```
before:  fm_backend_composer_state herdr <primary>  ->  unknown
after:   fm_backend_composer_state herdr <primary>  ->  empty
```

A real buffered escalation then delivers end to end. All three affected suites pass **on this
repository's base**, not only on our fork: `fm-composer-lib`, `fm-backend-herdr`, and
`fm-afk-inject-herdr-e2e` (which drives an actual delivery rather than mocking one — a unit test
cannot prove a defect whose whole nature is that the unit looked fine).

### DIFF
--- a/bin/backends/herdr.sh
+++ b/bin/backends/herdr.sh
@@ -2440,7 +2440,13 @@ fm_backend_herdr_strip_ansi() {  # <text>
 # afk-herdr-false-pending); it superseded a herdr-only faint byte-pattern check
 # that recognized only codex's bold-wrapped bare prompt and missed claude's own
 # dim ghost - the overnight away-mode injection wedge on the primary claude pane.
-FM_BACKEND_HERDR_COMPOSER_LINES=${FM_BACKEND_HERDR_COMPOSER_LINES:-20}
+# Rows of the pane tail the structural composer scan looks at. WHY 40: a harness
+# footer plus a multi-line custom statusline can occupy 11-12 rows under the
+# composer, and at 20 the composer row itself falls out of the window, which
+# reads as `unknown` and defers injection just as a false `pending` does. The
+# read is free: the capture already pulls 200 rows and this only bounds the tail
+# it scans.
+FM_BACKEND_HERDR_COMPOSER_LINES=${FM_BACKEND_HERDR_COMPOSER_LINES:-40}
 # Known ghost/placeholder composer text. Extend this if another
 # herdr-verified harness needs its own idle placeholder recognized.
 FM_BACKEND_HERDR_IDLE_RE=${FM_BACKEND_HERDR_IDLE_RE:-'^Type a message\.\.\.$'}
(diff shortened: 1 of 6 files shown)


eduardstan:
Superseded by https://github.com/kunchenguid/firstmate/pull/1668, which carries the same change raised through no-mistakes.

PR #285, opened 2026-07-06
fix(watcher): make PID identity locale-invariant

pchuri:
## Intent

Captain, the developer wanted a minimal fix for firstmate's watcher process-identity check so `fm_pid_identity()` pins the `ps` call to `LC_ALL=C`, preventing locale-dependent `lstart` date strings from breaking watcher health checks on non-C locales. They required a colocated regression test in `tests/fm-watcher-lock.test.sh` proving the identity output is invariant under non-C `LC_ALL` or `LC_TIME`, with no unrelated changes. They explicitly constrained the work to an isolated disposable worktree, a new `fm/fm-locale-idfix-q2` branch, no default-branch pushes or PR merges, use of `gh-axi`, and loading the `firstmate-coding-guidelines` skill before editing tracked firstmate material. They required local validation with shellcheck and the full bash test suite, then no-mistakes validation via the existing fork-backed gate, ultimately rerunning validation once codex was fixed and reporting `done: PR <url> checks green`.

## What Changed
- Captain, pinned `fm_pid_identity` process inspection to `LC_ALL=C` so watcher PID identity strings stay locale-invariant.
- Extended watcher lock tests to cover `fm_pid_identity` output and lock matching under non-C `LC_ALL` and `LC_TIME` settings.
- Updated contributing guidance to document the watcher PID identity locale coverage.

## Risk Assessment

✅ Low: Captain, the branch is narrowly scoped to making watcher PID identity generation locale-stable and adds a colocated regression check, with no obvious behavioral regression in the reviewed paths.

## Testing

The previously completed baseline full bash suite was accounted for; on the target change I ran the focused watcher-lock suite and captured end-to-end locale health evidence from a live isolated watcher, with all checks passing and no working-tree artifacts left behind.

<details>
<summary>Evidence: Real watcher locale health transcript</summary>

<code>identity_byte_equal_across_locales=yes&#10;fm_watcher_healthy_under_LC_ALL_ko_KR_UTF_8=ok&#10;fm_watcher_healthy_under_LC_TIME_ko_KR_UTF_8=ok</code>

```text
real fm-watch.sh locale health verification
locale_available=ko_KR.UTF-8
watch_process_pid=7835
lock_pid=7835
recorded_lock_identity=Mon Jul  6 15:09:50 2026     bash /Users/al02628760/.no-mistakes/worktrees/1c407ac2b9d3/01KWTZS2E1T9VC76MNXDQ2MR74/bin/fm-watch.sh
[...]

PR #1658, opened 2026-08-04
fix(wake): stop bash locale restores segfaulting forked supervision children

### LINKED ISSUE #1656: fm-wake-drain.sh annotation phase segfaults intermittently on macOS, and the same fault makes watcher liveness checks report a live watcher as gone
#### Environment

- macOS 26.4.1 (build 25E253), Darwin 25.4.0, arm64.
- GNU bash 5.3.9(1)-release (aarch64-apple-darwin25.1.0), installed from Homebrew at `/opt/homebrew/bin/bash`.
- Locale: `LANG` unset, `LC_ALL` unset, `LC_CTYPE=UTF-8`.
  This is the default a macOS terminal produces when it exports `LC_CTYPE` without `LANG`.

The ambient locale matters, and so does which bash is running.
Both are part of the reproduction rather than incidental detail.

#### What happens

`bin/fm-wake-drain.sh` intermittently prints a segfault line for its annotation phase while the drained wake records themselves print correctly:

```
bin/fm-wake-drain.sh: line 77: 86871 Segmentation fault: 11     ( fm_wake_print_annotations "$RAW_ROWS" )
```

The same fault has a second, quieter effect that is easy to misread.
`fm_pid_identity` in `bin/fm-wake-lib.sh` reads a process identity through a command substitution.
That substitution prints a correct identity and then dies, so the substitution returns a signal status and `|| return 1` fires.
`fm_watcher_lock_matches_pid` and `fm_watcher_healthy` therefore report a live, healthy watcher as absent.

Downstream that surfaces as watcher cycles closing with `watcher: FAILED - cycle ended without an actionable reason`, and as `bin/fm-watch-arm.sh` exiting non-zero with `watcher: FAILED - no live watcher with a fresh beacon` seconds after the watcher touched its own liveness beacon.

Measured on the environment above, against unmodified code, using synthetic fixtures in a throwaway state directory so anyone can reproduce the same numbers:

- Drain annotation phase: 3 crashes in 60 drains.
- `fm_watcher_healthy` against a genuinely live, correctly locked watcher: 20 false negatives in 250 checks.

#### Minimal reproduction

This needs nothing from this repository:

```sh
c=0
for i in $(seq 1 300); do
  /opt/homebrew/bin/bash -c 'f() { local LC_ALL=C; }; (f)' >/dev/null 2>&1 || c=$((c+1))
done
echo "crashes: $c/300"
```

Observed: `crashes: 18/300`.

Two controls isolate the trigger:

```sh
# Stock Apple bash, same machine, same locale.
for i in $(seq 1 300); do /bin/bash -c 'f() { local LC_ALL=C; }; (f)'; done   # 0/300

# Homebrew bash, but with a concrete ambient locale.
for i in $(seq 1 300); do env LANG=en_US.UTF-8 /opt/homebrew/bin/bash -c 'f() { local LC_ALL=C; }; (f)'; done   # 0/300
```

The crash needs all three conditions together: a bash that resolves the empty locale through gettext, an ambient locale with no concrete value to resolve to, and a forked child performing the restore.

#### Expected behavior

Setting `LC_ALL=C` for a scoped command or a function should not be able to terminate the shell that does it.
A wake drain should annotate or skip, and a liveness check should answer, rather than dying after producing a correct answer.

#### Root cause

Bash restores `LC_ALL` when an `LC_ALL=C <cmd>` assignment prefix or a `local LC_ALL=C` goes out of scope.
That restore runs bash's internal `reset_locale_vars`, which calls `setlocale(LC_ALL, "")`.

Homebrew's bash is linked against gettext, so that call lands in `libintl_setlocale` rather than the libc entry point:

```
$ otool -L /opt/homebrew/bin/bash | grep -iE 'intl|CoreFoundation'
	/opt/homebrew/opt/gettext/lib/libintl.8.dylib (compatibility version 13.0.0, current version 13.5.0)
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 4109.1.255)

$ otool -L /bin/bash | grep -iE 'intl|CoreFoundation'
(no matches)
```

With no concrete ambient locale, resolving `""` makes gettext ask CoreFoundation for the user's preferred language.
CoreFoundation is not fork-safe: it may not be called in a child after `fork()` without an intervening `exec()`.
Every forked bash child that performs this restore is therefore doing something undefined, and it faults depending on process state.

The crash report's faulting thread shows exactly that path:

```
libsystem_trace.dylib  _os_log_preferences_refresh + 56
libsystem_trace.dylib  os_log_type_enabled + 772
CoreFoundation         -[CFPrefsSearchListSource alreadylocked_copyValueForKey:] + 204
CoreFoundation         -[CFPrefsSource copyValueForKey:] + 44
...
Foundation             LocaleCache.preferredLanguages(forCurrentUser:) + 164
CoreFoundation         CFLocaleCopyPreferredLanguages + 28
libintl.8.dylib        libintl_setlocale + 468
bash                   reset_locale_vars + 68
```

Exception type `EXC_BAD_ACCESS (SIGSEGV)`, `KERN_INVALID_ADDRESS`.

This is a bash-on-macOS platform hazard rather than a defect in this repository's logic.
It reaches this repository because the wake library forks on every wake and pins `LC_ALL=C` in those forked children.
Linux and WSL are unaffected: their bash resolves `""` through libc with no CoreFoundation involved.

dennismolund:
Fixes #1656.

#### What breaks, and for whom

On macOS with a gettext-linked bash (the Homebrew build) and no concrete ambient locale, restoring `LC_ALL` inside a forked child segfaults that child intermittently, at roughly one call in twelve.
The issue has the environment, the measurements, and a self-contained reproduction.

Two supervision paths are affected:

- `bin/fm-wake-drain.sh` ran its annotation phase in a subshell, so the crash killed that subshell and printed a bare `Segmentation fault` line. Drained wake records were never lost, because they are printed and committed before the annotation runs, but the annotation was.
- `fm_pid_identity` returned a signal status after printing a correct identity, so `fm_watcher_healthy` reported a live watcher as absent. That is what produced `cycle ended without an actionable reason` and `no live watcher with a fresh beacon` moments after a beacon touch.

Linux and WSL are unaffected; their bash never reaches CoreFoundation for this.

#### The fix, and why the locale guarantee is kept

The C locale was pinned in these places for real reasons, and every one of them still holds.
The change is only in how the C locale is reached, never in whether it applies.

- `fm_pid_identity` uses `env LC_ALL=C ps ...`. `ps` still renders `lstart` under C, so the identity stays locale-invariant as #285 intended, but this shell's own locale never moves and there is nothing to restore.
- `fm_wake_clean_field` uses `env LC_ALL=C tr ...`, keeping control-character collapsing byte-wise so a stray control character still cannot desync the fixed five-field record.
- `fm_wake_print_annotations` pins `LC_ALL=C` exported inside a subshell that exits without restoring it. The renderer still measures `${#line}` and slices `${line:0:n}` under C, so the 2048-byte per-item and 8192-byte global caps stay byte bounds. A subshell cannot modify its parent's environment, so the pin is confined by a process boundary instead of by a scope-exit restore.

The general rule is recorded in the coding guidelines: reach the C locale with `env` when only the external command needs it, and with a non-restoring exported pin when bash itself needs C semantics.
One documentation example in `docs/fm-test-portable-shards.md` is aligned to the same spelling, so the docs do not demonstrate the pattern the guidelines now ban.

The drain's annotation boundary is also made structural rather than incidental.
The annotation is consumed through a command substitution, so however that process dies, the death is a status the drain inspects rather than an event that reaches the drain's output.
It degrades to a skipped annotation and one stderr note, and the already-committed raw wake records are untouched.

#### Platform impact

The crash being fixed is macOS-specific.
The hardening is platform-neutral: `env`-scoped locales, a non-restoring pin, and an annotation phase that cannot take down its caller are correct everywhere and cost nothing on Linux.
No behavior changes on any platform where the crash never occurred - the annotation output is byte-identical before and after on the same input.

#### Risk assessment

Low.

- The change alters only how the C locale is reached (an `env`-scoped assignment for external commands, a non-restoring exported pin where bash itself needs C semantics), never whether it applies; annotation output is byte-identical before and after on identical input.
- Platforms without the fault see no behavior change; the added isolation is strictly protective (a dying annotation phase degrades to a skipped annotation and one stderr note, and raw wake records are committed before annotation runs).
- The riskiest surface, watcher liveness verification, was measured before and after on the exact gate `fm-watch-arm.sh` uses: 20 false negatives in 250 checks against a live watcher, then 0 in 250, with identity reads stable across repeated forks.
- Four regression tests are each proven to fail against the unfixed code, and a full-suite walk classified every observed failure as pre-existing at equal-or-lower rates against pristine `main`.
- Residual risk is reintroduction by future code: the crash class is wider than the two call sites fixed here, so the guard rule lives in the coding guidelines with the rationale comment in `bin/fm-wake-lib.sh` as its owner.

#### Regression tests

Four cases in `tests/fm-wake-queue.test.sh`, each verified to fail against the unfixed code:

- Repeated drains under the crash-triggering ambient locale must all complete normally. The repeat count turns an ~8% per-drain crash into a near-certain catch. The run also asserts that every drain actually produced an annotation, so survival cannot pass vacuously through a renderer that quietly emits nothing.
- A dying annotation phase must not reach the drain. A `perl` shim walks its own ancestry and sends `SIGSEGV` to the drain processes between itself and the outermost drain - the same signal, on the same processes, the real crash hit. The drain must still exit zero, keep its raw records, report the skip, and not surface the abnormal termination. A platform whose `ps` cannot report that ancestry skips explicitly rather than passing without killing anything.
- Identity reads of a live process must stay successful and stable across repeated forks under the same ambient locale.
- The annotation caps must remain byte bounds when status content is multibyte. Every other fixture in this suite is ASCII, where bytes and characters are the same number, so a lost locale pin would go unnoticed. With multibyte content, an unpinned renderer counts characters and the 2048-byte cap runs to roughly 4.6 KB.

All fixture content is synthetic.

What each case is worth on a platform without the fault is worth stating plainly.
The annotation-death case is fully real everywhere: it kills real processes with a real signal, so it tests the drain's boundary on Linux exactly as it does on macOS.
The byte-bound case is a genuine bound everywhere.
The repeated-drain and identity-stability cases cannot catch a crash that platform does not have, but they still assert that every drain completes and annotates and that identity reads stay stable, so neither is inert.

#### Verification

- `bin/fm-lint.sh` clean (ShellCheck 0.11.0, the pinned version).
- `bin/fm-doc-audience-check.sh` clean.
- Full `bin/fm-test-run.sh --all` regression run.
- Before and after on the exact liveness gate `fm-watch-arm.sh` uses: 20 false negatives in 250 checks against a live watcher, then 0 in 250.

#### Follow-ups found during validation

This PR fixes the two sites behind the reported symptoms and deliberately stops there, so the change stays small enough to review against the reproduction.
Validation surfaced more of the same pattern, recorded here rather than swept in:

- `bin/fm-pending-reply-lib.sh:730` carries the identical shape: an `LC_ALL=C` prefix inside a checked command substitution, in `fm_pending_reply_pid_identity`. The consequence matches the one fixed here - a live sender can be misread as dead, which would mis-trigger pending-reply recovery and escalation. This is a genuine second instance, not a stylistic one.
- Roughly eighty further `LC_ALL=<x> <cmd>` command-substitution sites exist across `bin/`, along with `local LC_ALL=C` in `bin/fm-pr-lib.sh` (lines 95, 121, 143, 170, 212) and `bin/fm-primary-scope-lib.sh:8`. Notable others include `bin/backends/tmux.sh:220`, `bin/fm-remote-file.sh:121`, and `bin/fm-backlog-handoff.sh:284`. Sites whose substitution status is checked with `|| return 1` or `|| die` share the same misread-as-failure hazard on macOS; sites whose status is unchecked do not, because the captured output survives a child death at locale-restore time.

Both are suitable for a follow-up PR applying the convention this change documents.
They were left out here on purpose: a first fix is easier to judge when its diff matches the failure it reproduces.

## Pipeline

<details>

✅ No issues found.

</details>

<details>

- ⚠️ `.agents/skills/firstmate-coding-guidelines/SKILL.md` - branch carries 3 commit(s) that exist on your local main branch but were never pushed to origin/main; rebasing would bundle this unrelated work (30 file(s)) into the PR:
- 8571451 docs(guidelines): rule out reintroducing the forked-child locale restore
- 9324d54 test(wake): cover the locale-restore crash and the annotation boundary
- a793674 fix(wake): stop bash restoring LC_ALL in forked supervision children

</details>

<details>

- ⚠️ `bin/fm-pending-reply-lib.sh:730` - bin/fm-pending-reply-lib.sh:730 (fm_pending_reply_pid_identity) retains the exact crash shape this branch fixes in fm_pid_identity: `identity=$(COLUMNS=10000 LC_ALL=C ps -p &#34;$pid&#34; ...) || return 1` runs the LC_ALL=C prefix inside a checked command substitution, so bash&#39;s locale restore in the forked child can segfault after correct output on macOS with the bare-UTF-8 ambient locale, making fm_pending_reply_sender_alive misread a live sender as dead and mis-trigger pending-reply recovery/escalation. Mechanical fix per the new convention: `env LC_ALL=C ps ...`.
- ⚠️ `.agents/skills/firstmate-coding-guidelines/SKILL.md:121` - The new guideline bans bash LC_ALL restores in forked children, but the pattern remains widespread outside fm-wake-lib.sh: `local LC_ALL=C` in bin/fm-pr-lib.sh (lines 95, 121, 143, 170, 212) and bin/fm-primary-scope-lib.sh:8, plus dozens of `$(LC_ALL=C cmd ...)` command-substitution sites across bin/ (e.g. fm-remote-file.sh:121, fm-backlog-handoff.sh:284, backends/tmux.sh:220, fm-fleet-snapshot.sh). Sites whose substitution status is checked (`|| return 1` / `|| die`) share the misread-as-failure hazard on macOS. Decide whether to sweep these in this branch or file a follow-up task; the fix is mechanical (env LC_ALL=C, or exported pin in a non-restoring subshell) but expands the branch&#39;s deliberate scope.
- ℹ️ `tests/fm-wake-queue.test.sh:526` - The new byte-bounds test itself uses the banned pattern: `widest=$(LC_ALL=C awk ...)` and `total=$(LC_ALL=C awk ...)` at tests/fm-wake-queue.test.sh:526 and 528 run LC_ALL=C prefixes inside command substitutions under the harness&#39;s ambient (potentially crash-triggering) locale. Harmless in outcome today - the captured value survives a child death at locale-restore time and the assignment status is unchecked - but `env LC_ALL=C awk` would keep the regression suite consistent with the convention it enforces (the pre-existing lines 343/345 have the same shape).

🔧 Fix: use env LC_ALL=C prefixes in wake-queue tests

</details>

<details>

✅ No issues found.
- <code>`bash tests/fm-wake-queue.test.sh` - all 16 cases pass (exit 0), including the 4 new regression tests: annotation-process-death boundary, 60-drain crash-locale repeat, multibyte byte-cap bound, and identity-read stability</code>
- <code>`bash tests/fm-watcher-lock.test.sh` - all watcher lock/liveness cases pass (exercises fm_pid_identity callers)</code>
- <code>`bash tests/fm-documentation-audiences.test.sh` - passes (exit 0) after the coding-guidelines SKILL.md addition</code>
- <code>Manual before/after repro: 120 runs of `bin/fm-wake-drain.sh` under `env -u LANG -u LC_ALL LC_CTYPE=UTF-8` against base cf95112 (5 drains printed `Segmentation fault: 11` from the annotation subshell) vs target 2e218b2 (0 signal deaths, 120/120 annotations produced, all exit 0)</code>
- <code>Manual before/after repro: 120 `fm_pid_identity` reads of a live process under the same locale against base (5 failed reads of a live process) vs target (0 failures, identity stable across all reads)</code>
- `Captured one fixed drain's user-visible transcript under the crash-triggering locale: raw wake row + annotation, exit 0, silent stderr`

</details>

<details>

✅ No issues found.

</details>

<details>

🔧 Fix: silence SC2016 on literal awk programs in wake-queue tests

</details>

<details>

✅ No issues found.

</details>

### DIFF
--- a/.agents/skills/firstmate-coding-guidelines/SKILL.md
+++ b/.agents/skills/firstmate-coding-guidelines/SKILL.md
@@ -118,6 +118,8 @@ Run `bin/fm-doc-audience-check.sh`; it enforces classification, README setup rou
 - Plain dash `-`, never an em dash.
 - Never add an agent name as a commit co-author.
 - `bin/*.sh` and `bin/backends/*.sh` must pass `shellcheck`.
+- Never let bash restore `LC_ALL` inside a forked child: an `LC_ALL=C <cmd>` prefix or a `local LC_ALL=C` in a command substitution, pipeline stage, or subshell can segfault that child on macOS, silently turning a correct result into a signal status.
+  Use `env LC_ALL=C <cmd>` when only the external command needs the locale, and an exported pin in a subshell that never restores it when bash itself needs C semantics; the locale convention comment in `bin/fm-wake-lib.sh` owns the full rationale.
 - Run `bin/fm-lint.sh` before treating a script change as done; it is the single owner of the lint definition (file set, config, and pinned shellcheck version) that CI and the no-mistakes pre-push gate both invoke, and it refuses to run under any other shellcheck version.
 - Colocate tests with the existing pattern in `tests/`, name them `<subject>.test.sh`, and extend an existing script rather than inventing a new runner.
 - Tests must exercise behavior through an executable or public interface and must never assert implementation-source bytes, including through parsers, regexes, snapshots, or indirect wrappers.
--- a/bin/fm-wake-drain.sh
+++ b/bin/fm-wake-drain.sh
@@ -12,6 +12,7 @@ SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 DRAIN_TMP=
 DRAIN_LOCK_HELD=false
 RAW_ROWS=
+ANNOTATIONS=
 
 # Defense in depth for the supervision chain: this script runs at the top of
 # every wake-handling and recovery turn, so assert supervision health here too. A
@@ -126,7 +127,19 @@ DRAIN_LOCK_HELD=false
 
 # Raw output and queue deletion are authoritative. Everything below is
 # best-effort and cannot restore, duplicate, hide, or fail the consumed rows.
-(fm_wake_print_annotations "$RAW_ROWS") || true
+#
+# The annotation is consumed through a command substitution rather than run
+# inline, so its process death is a status this script inspects instead of an
+# event that reaches the drain's own output. Whatever kills that process -
+# an unexpected row shape, an unreadable status file, or an abnormal
+# termination - degrades here to a skipped annotation and one stderr note.
+# The raw records above are already printed and the queue is already committed,
+# so a skipped annotation costs context, never a wake.
+if ANNOTATIONS=$(fm_wake_print_annotations "$RAW_ROWS" 2>/dev/null); then
+  [ -z "$ANNOTATIONS" ] || printf '%s\n' "$ANNOTATIONS"
+else
+  printf 'fm-wake-drain: wake annotation skipped; raw wake records above are complete\n' >&2
+fi
 (print_open_decisions_section) || true
 assert_watcher_liveness
 exit 0
--- a/bin/fm-wake-lib.sh
+++ b/bin/fm-wake-lib.sh
@@ -15,6 +15,29 @@ FM_LOCK_STALE_AFTER="${FM_LOCK_STALE_AFTER:-2}"
 _FM_UNAME=$(uname 2>/dev/null || echo unknown)
 mkdir -p "$STATE"
 
+# Locale convention for this library.
+#
+# Bash restores LC_ALL when an `LC_ALL=C <cmd>` prefix or a `local LC_ALL=C`
+# goes out of scope, and that restore calls setlocale(LC_ALL, ""). With no
+# concrete ambient locale - the macOS default, where LC_CTYPE is the bare
+# "UTF-8" and LANG is unset - resolving "" asks CoreFoundation for the user's
+# preferred language. CoreFoundation is not fork-safe, so that restore
+# intermittently segfaults whatever forked child it runs in: a command
+# substitution, a pipeline stage, or an explicit subshell. This library forks on
+# every wake, so the consequence is not cosmetic. It killed the drain's
+# annotation phase outright, and it made the identity read below return a
+# correct answer and then die, which its callers could only read as a live
+# watcher having vanished.
+#
+# A bash-level locale prefix is therefore safe here only where the ambient
+# locale is already C, so that the restore is C to C and never resolves "".
+# Everything else reaches the C locale one of two ways:
+#   - `env LC_ALL=C <cmd>` when only the external command needs it. This never
+#     moves this shell's own locale, so there is nothing to restore.
+#   - an exported pin inside a subshell that exits without restoring it, when
+#     bash itself needs C semantics; see fm_wake_print_annotations, which is
+#     what makes the prefixes beneath it safe.
+
 fm_current_pid() {
   printf '%s\n' "${BASHPID:-$$}"
 }
@@ -59,7 +82,7 @@ fm_pid_identity() {
   # Pin LC_ALL=C so lstart's date format is locale-invariant: the identity is
   # written under one locale but re-read under the machine's ambient locale, which
   # would otherwise mismatch on a non-C locale (e.g. ko_KR) and reject a live watcher.
-  out=$(LC_ALL=C ps -p "$pid" -o lstart= -o command= 2>/dev/null) || return 1
+  out=$(env LC_ALL=C ps -p "$pid" -o lstart= -o command= 2>/dev/null) || return 1
   [ -n "$out" ] || return 1
   printf '%s\n' "$out" | sed 's/^[[:space:]]*//'
 }
@@ -516,7 +539,7 @@ fm_failure_episode_reset() {
 }
 
 fm_wake_clean_field() {
-  LC_ALL=C tr '\t\r\n' '   '
+  env LC_ALL=C tr '\t\r\n' '   '
 }
 
 fm_wake_append() {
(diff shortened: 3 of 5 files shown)


kunchenguid:

<sub>Noted for firstmate#1658 at `0f4ac48e`.</sub>

kunchenguid:

<sub>Noted for firstmate#1658 at `153b7a82`.</sub>

kunchenguid:

PR #1671, opened 2026-08-04
fix(bin): deliver away-mode escalations when the runtime owns the watcher

jericho1050:
## Away mode could not deliver a single escalation on a runtime-owned watcher

On Pi (and any runtime whose extension owns the watcher), `/afk` starts the
sub-supervisor daemon, the daemon reports healthy, and **zero escalations ever reach
the captain**. Measured over a full night: ~2809 injection attempts, 0 delivered.

The daemon assumes it owns the watcher and tries to start its own. When the runtime
extension already owns one, the daemon's child logs `watcher: already running pid N`
every ~16s forever and the daemon is starved of wakes - it never classifies anything.
Only the 300s catch-all fleet scan ever produced an escalation.

That is the worst possible failure shape: away mode looks armed, every health check
passes, and the captain silently receives nothing.

### What this changes

The daemon detects a runtime-owned watcher and consumes its durable wake queue instead
of competing for ownership, so escalations flow on the path that is actually running.

### Verification

Diagnosed with two live probes before the fix, and the failure is reproducible:
probe 2 had the extension watcher wake firstmate directly with the probe line while
`state/.subsuper-escalations` stayed EMPTY - proving the daemon never saw it.

### DIFF
--- a/.agents/skills/afk/SKILL.md
+++ b/.agents/skills/afk/SKILL.md
@@ -49,8 +49,10 @@ batched digest rather than per-wake injections.
    The daemon is **presence-gated**: it injects escalations only while
    `state/.afk` exists, and stays quiet otherwise.
 
-3. **Do not separately arm `fm-watch.sh`.** The daemon manages the watcher as
-   its child; the singleton lock no-ops a stray arm harmlessly.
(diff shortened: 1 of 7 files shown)


kunchenguid:

PR #1674, opened 2026-08-04
fix(guard): never report a fresh-beacon live watcher as down

jericho1050:
## Problem

Firstmate's supervision guard falsely declared `WATCHER DOWN - SUPERVISION IS OFF` while a watcher was demonstrably alive and beating.
Measured live: the beacon was 34 seconds old against a 300-second grace, yet the guard reported "no watcher has a fresh beacon."
Because the banner reads as a hard alarm, it pushes the operator to tear down and re-arm supervision that was never actually down.

The trigger is that the alarm gate used the strict `fm_watcher_healthy` check, which also requires the recorded watcher identity and path to byte-match the live process.
That match transiently fails on a benign identity read, a re-render, or a mid-handoff window, so a live, beating watcher was intermittently reported as dead.

## Fix

The advisory guard now gates its alarm on a weaker, beacon-anchored predicate, `fm_watcher_beating`: true when a live process holds this home's watcher lock and the beacon is fresh within grace, independent of the recorded identity bytes.
`fm_watcher_healthy` stays the strict gate for the turn-end block and the arm/attach paths, so the relaxation is confined to the pull-based advisory guard and does not weaken any decision that starts or attaches a watcher.
Home scoping is preserved, so a foreign or stale lock record is never counted as this home's live watcher.

When the guard genuinely does need to alarm, it now words the verdict from the actual failure instead of always blaming a stale beacon:

- A stale or absent beacon is reported as no fresh beacon (unchanged wording).
- A fresh beacon that still fails the beating check - a watcher that died and left a not-yet-aged beacon - is reported as no live watcher holding this home's lock, which is what actually happened.

## Verification

- `bash tests/fm-watcher-lock.test.sh` - all tests pass, including the new coverage below.
- `bash bin/fm-lint.sh` (pinned ShellCheck 0.11.0) - clean.

New regression tests:

- A fresh beacon with a live lock holder and an identity mismatch is silent (never `WATCHER DOWN`).
- Same, with a watcher-path mismatch, is silent.
- A dead recorded pid with a not-yet-aged leftover beacon still alarms, and uses the lock-truthful wording rather than falsely claiming a stale beacon.
- `fm_watcher_beating` directly requires a live, home-matched lock holder and a fresh beacon (rejects a stale beacon, a foreign-home record, and a dead pid).

## Scope note

This change is the guard-verdict half of the Pi-primary supervision work.
The ordinary-wake delivery path to a Pi primary is owned by the harness-identity fix and the sibling supervision-delivery PRs and is intentionally not duplicated here.

## Known unrelated failures

`tests/fm-pi-primary-types.test.sh` and `tests/fm-calm-pi-extension.test.sh` fail on a clean tree as well (Pi SDK type drift); they are unrelated to this branch and untouched by it.

### DIFF
--- a/bin/fm-guard.sh
+++ b/bin/fm-guard.sh
@@ -154,6 +154,17 @@ watcher_healthy=false
 if fm_watcher_healthy "$STATE" "$WATCH" "$GRACE" "$FM_HOME"; then
   watcher_healthy=true
 fi
+# The alarm gate is the weaker, beacon-anchored liveness, not the strict identity
+# match: a fresh beacon with a live lock holder means a watcher IS beating, so
+# supervision is happening even if the recorded identity does not byte-match (a
+# transient identity read, a benign re-render, or a mid-handoff). This is what
+# keeps a demonstrably fresh beacon from ever being reported as "watcher down".
+# fm_watcher_healthy stays authoritative for the turn-end block and arm/attach
+# paths; only this advisory relaxes to the beacon.
+watcher_beating=false
+if [ "$watcher_healthy" = true ] || fm_watcher_beating "$STATE" "$GRACE" "$FM_HOME"; then
+  watcher_beating=true
+fi
 if [ "$needed" = false ]; then
   # Leave the unhealthy state (nothing riding on the watcher): clear so a later
   # work or X-mode need + stale combination is a fresh episode even if the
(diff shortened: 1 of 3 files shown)


kunchenguid:

PR #1675, opened 2026-08-04
fix(bin): resolve firstmate's harness from live ancestry before env markers

jericho1050:
## What went wrong

A captain exited a Claude Code session with `/exit` and started `pi` in the **same shell**.
Claude Code's exported environment (`CLAUDECODE=1` and the other `CLAUDE_CODE_*` variables) survived that transition, because an exported variable outlives the process that set it.
Firstmate then ran on Pi for the whole session while reporting itself as Claude Code.

Harness identity selects the supervision protocol firstmate emits for its own turns.
Believing it was Claude Code, this Pi session emitted the Claude Stop-hook auto-arm protocol instead of Pi's extension-owned one. The observed fallout:

- repeated `TURN WOULD END BLIND - supervision is off` guard fires with no working repair path;
- `bin/fm-watch-arm.sh` refusing (surfaced as "this session is shutting down"), which was misread as a separate defect and drove a full session restart that did not fix it;
- the captain reporting that away-mode messages never arrived ("maybe it's not working in the Pi harness") - which was correct and was misdiagnosed at the time.

This is upstream of the `afk-pi-native` fix: that work assumes firstmate already knows it is Pi. Here it did not.

## Root cause

`bin/fm-harness.sh`'s `detect_own()` checked verified environment markers **first** and only fell back to live process ancestry.
It short-circuited on `CLAUDECODE=1` and returned `claude` before ever reaching the `PI_CODING_AGENT=true` check. A stale exported marker from a dead ancestor therefore outranked the currently-running agent's own live marker.

## The fix

Reorder detection so **live process ancestry is authoritative** and verified env markers are only a fallback used when ancestry cannot name a harness.

The precedence is not an arbitrary line swap - it follows which signals are launch-scoped and which can survive an `exec`:

- **Process ancestry** reflects the process tree as it exists *right now*. The walk starts at this process and stops at pid 1, so it can never name a dead ancestor: a process whose parent exits is reparented to pid 1, where the walk stops. The harness whose process actually launched firstmate always appears in this chain.
- **An exported marker** (`CLAUDECODE=1`, `PI_CODING_AGENT=true`, `GROK_AGENT=1`) outlives the process that set it - it persists in an interactive shell across `/exit` and into the next program launched there. So it must never outrank live ancestry.

Ancestry-first also correctly resolves genuine nesting (e.g. a `pi` launched from a Claude tool shell) to the **closest** harness in the chain.

`resolve_pi_identity` is extracted so the `pi-signed` launch boundary is applied consistently in both the ancestry and fallback paths: the signed identity is selected only when both live Pi markers are present (`PI_CODING_AGENT=true` **and** `FM_PI_HARNESS=pi-signed`). A plain-Pi launch cannot be relabelled by an inherited `FM_PI_HARNESS=pi-signed`, and unmarked shared Pi ancestry stays plain `pi`.

The env-marker fallback still matters: `codex`, `opencode`, and `kimi` are markerless and are detectable **only** through ancestry, while `claude`, the Pi family, and `grok` expose a marker used when ancestry is blind (no `ps`, or a bare interpreter whose script path did not match).

## Where harness identity is consumed (what was wrong this session)

Every consumer of firstmate's own harness identity was fed `claude` instead of `pi`:

1. **Supervision protocol emission** - `bin/fm-supervision-instructions.sh` selects the emitted protocol snippet purely from harness identity, so the Pi session got the Claude protocol. (`bin/fm-session-start.sh` and the away-mode `bin/fm-supervise-daemon.sh` read the same identity.)
2. **Turn-end guard mode** - the same snippet selection wires the turn-end guard: `pi.md` wires Pi's extension-owned guard, `claude.md` wires the Claude Stop-hook auto-arm whose `fm-watch-arm` refused. This is the direct source of the "turn would end blind" fires.
3. **Busy-state / turn-end signal** - the emitted protocol determines which harness's turn-end signal firstmate waits on, so a Pi primary told it was Claude waited on the wrong signal and never woke correctly.
4. **`fm-harness.sh crew` fallback** - `resolve_crew()` mirrors `detect_own` when `config/crew-harness` is absent or `default`, so crewmates would have been launched on the wrong harness too.

## Does spawn propagate the pollution? (reported, not fixed here)

Yes. `bin/fm-spawn.sh` does **not** scrub `CLAUDECODE` / `PI_CODING_AGENT` / `GROK_AGENT` from the child environment before launching a crewmate (it only sets `FM_PI_HARNESS` for Pi launches and unsets `TRACEPARENT`), so stale markers do propagate into every child.

This fix makes that largely harmless for the normal case: a child's own live parent is its launching harness, so ancestry-first detection now correctly ignores any inherited marker. The only residual gap is a markerless child whose ancestry walk cannot match (no `ps`, or a wrapper hiding the harness name in its path), which would then fall back to a stale inherited marker. Scrubbing markers at spawn would be defense-in-depth; I did **not** expand scope to do it here, per the task constraint - flagging it as a separate follow-up.

## Verification

- **Reproduced first.** With a fake `ps` supplying a live `pi` parent chain plus a stale `CLAUDECODE=1`, the pre-fix `bin/fm-harness.sh` (from `origin/main`) returns `claude`; the fixed script returns `pi`.
- **Regression coverage** added in `tests/fm-secondmate-harness.test.sh` (a table covering both leak directions, the `pi-signed` launch boundary, markerless-harness leaks, and every blind-ancestry fallback row) and `tests/fm-kimi-harness.test.sh` (markerless kimi: live ancestry beats a stale inherited `CLAUDECODE`). The two config-resolution suites were pinned with a neutral fake `ps` so their env-marker assertions no longer depend on whichever harness ran the suite.
- `bash tests/fm-secondmate-harness.test.sh` - all pass.
- `bash tests/fm-kimi-harness.test.sh` - all pass.
- `bin/fm-lint.sh` (the repo's pinned shellcheck owner, 0.11.0) - clean.

Note: `tests/fm-pi-primary-types.test.sh` and `tests/fm-calm-pi-extension.test.sh` fail on a clean tree too (Pi SDK type drift, unrelated to this branch); they are not touched here.

### DIFF
--- a/.agents/skills/harness-adapters/SKILL.md
+++ b/.agents/skills/harness-adapters/SKILL.md
@@ -39,7 +39,7 @@ If the captain asks for a new harness, propose verifying it first: spawn a trivi
 
 ## Detection
 
-`bin/fm-harness.sh` prints firstmate's own harness, using verified env markers first and then process ancestry.
+`bin/fm-harness.sh` prints firstmate's own harness, resolving live process ancestry first and falling back to verified env markers only when ancestry cannot name a harness, so a stale marker inherited from a dead ancestor (such as a `CLAUDECODE=1` that survived a prior session's `/exit`) never outranks the harness whose process is actually running firstmate.
 Within the Pi family, only the exact launch-boundary marker `FM_PI_HARNESS=pi-signed` alongside `PI_CODING_AGENT=true` selects the signed identity; unmarked shared launcher ancestry remains `pi`.
 `bin/fm-harness.sh crew` resolves the effective crewmate harness from `config/crew-harness` (absent or `default` -> own).
 `bin/fm-harness.sh secondmate` resolves the secondmate-launch harness through the chain `config/secondmate-harness` -> `config/crew-harness` -> own, so an unset `config/secondmate-harness` matches the crew harness.
--- a/bin/fm-harness.sh
+++ b/bin/fm-harness.sh
@@ -18,8 +18,15 @@
 # harness only, no model/effort. Only the first non-empty, non-comment line is parsed.
 # Model/effort come ONLY from this file - config/crew-harness stays a bare adapter
 # name and is never parsed for a model.
-# Detection layers: verified environment markers first, then process ancestry.
-# Record each newly verified env marker here.
+# Detection layers: live process ancestry FIRST, then verified environment
+# markers as a fallback when ancestry cannot see the harness. Ancestry is the
+# authoritative signal because the parent chain reflects the process tree as it
+# exists right now and can never name a dead ancestor (a process whose parent
+# exits is reparented to pid 1, where the walk stops). An exported marker such as
+# CLAUDECODE=1 outlives the process that set it - it persists in an interactive
+# shell across `/exit` and into the next program launched there - so a stale
+# inherited marker must never outrank the harness whose process actually launched
+# this one. Record each newly verified env marker in the fallback layer below.
 set -u
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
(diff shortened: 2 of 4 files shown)


kunchenguid:

PR #1681, opened 2026-08-04
fix(watch): absorb a live declared-paused crewmate on the bounded pause cadence

MattH-ca:
## Problem

A healthy crewmate whose last status line is `paused:` emitted repeated bare `stale: <window>` wakes, each costing firstmate a supervision turn. Observed live on 2026-08-03 on two separate tasks.

`bin/fm-crew-state.sh` was already correct, reporting `state: paused · source: status-log`. Only the watcher's absorption path was wrong.

`pause_state_class` short-circuited to `none` whenever `fm_backend_agent_alive` reported anything other than `dead`, in both its fast path and its fallback path. A crewmate that declared `paused:` and then sat alive and idle therefore classified as `none` and fell through to ordinary stale handling, never reaching `handle_paused_stale`. Pause absorption was reachable only for a confidently dead agent, which inverts the documented contract: `AGENTS.md` section 8 and the `bin/fm-brief.sh` status protocol both state that a declared `paused:` means firstmate leaves the idle pane alone and rechecks on a long cadence rather than treating it as a possible wedge.

The `.paused-*` / `.paused-rechecked-*` / `.paused-resurfaced-*` markers were still being written as a side effect of `surface_nonterminal_stale`, which is why they looked freshly stamped even though pause handling never ran.

## Fix

Split the liveness gate by declaration type rather than dropping it.

A genuine `paused:` is a live external wait, so the crew is expected to sit alive and idle. It now classifies `paused` and is absorbed on the bounded cadence whenever crew state still agrees, independent of agent liveness.

The dead-agent gate now applies only to a `captain-held` transfer. There the agent was meant to have exited, so an endpoint still running its harness contradicts the declaration and must surface. `map_log_state` maps `captain-held` to `unknown` and never to `paused`, so the new early return cannot absorb a live captain-held endpoint.

This aligns the normal-mode watcher with `bin/fm-supervise-daemon.sh`, which already classified declared pauses without a liveness gate, rather than creating a second owner for the behavior.

### How the safety property is preserved

The liveness gate existed so a live-but-wedged worker that wrongly declared `paused:` could not hide forever. That property is preserved, not removed.

It now routes through `handle_paused_stale`'s existing `PAUSE_RESURFACE_SECS` recheck and the `.paused-resurfaced-*` throttle. Because that cadence is anchored on the status file's own mtime, a churny idle pane cannot reset it, so a forgotten pause still resurfaces on a bounded cadence as an awaiting-external recheck rather than a wedge.

## Coverage

Added to `tests/fm-watch-triage.test.sh`, in the file's colocated style:

- `test_live_declared_pause_churny_pane_not_flooded` - the 2026-08-03 flood reproduction. A live declared pause on a pane that keeps ticking must be absorbed on every distinct hash, never surfaced per hash.
- `test_live_declared_pause_resurfaces_on_bounded_cadence` - a live declared pause that stops matching still resurfaces within the bounded cadence.
- `test_live_captain_held_still_surfaces` - the other half of the split. A still-live captain-held endpoint keeps producing a bare stale wake, so the retained gate stays under test.
- `test_exited_and_live_declared_pause_and_captain_held_are_bounded` - the dead-agent pause case.
- `test_nonterminal_stale_not_working_surfaced` now pins `FM_FAKE_TMUX_CURRENT_COMMAND`, so the live and not-paused case is asserted with a genuinely live agent and wedge detection is protected against regression.

## Documentation

`docs/architecture.md` owned the inaccurate liveness precondition and is corrected there, along with the `status_is_paused_or_captain_held` header and the watcher comments. `AGENTS.md` section 8 and `bin/fm-brief.sh` never asserted a liveness gate, so nothing there went stale and the contract is not restated in a second place.

## Known limitation, filed as separate follow-up

This does not cure harnesses with no verified busy source. `fm_busy_classify` hardcodes `unknown` for `codex*` and `kimi*` (`bin/fm-busy-lib.sh`), so `fm-crew-state.sh` emits `unknown pane` and exits before its status-log fallback. `pause_state_class` then still receives `class=none` and a live declared-paused Codex or Kimi crewmate continues to emit a bare stale wake roughly every `STALE_ESCALATE_SECS`.

That is better than the previous per-hash flood but is the same underlying failure. Closing it means changing the crew-state and busy-classification contract, which is deliberately out of scope here and tracked as separate follow-up work.

### DIFF
--- a/bin/fm-classify-lib.sh
+++ b/bin/fm-classify-lib.sh
@@ -129,9 +129,11 @@ status_is_paused() {  # <status-line>
 
 # 0 if a status line declares either an external-wait pause or a verified
 # captain-held transfer.
-# Both declarations can intentionally leave an exited crew's endpoint idle, so
-# the watcher applies its bounded pause cadence when agent death confirms that
-# no live decision gate is being silenced.
+# A declared paused: is a live external wait - the crew stays alive and idle - so
+# the watcher applies its bounded pause cadence regardless of agent liveness. A
+# captain-held transfer instead expects the endpoint to have exited, so the
+# watcher applies the same cadence only when agent death confirms it and surfaces
+# a still-live captain-held endpoint. pause_state_class owns that split.
 status_is_paused_or_captain_held() {  # <status-line>
   local line=$1 verb
   status_is_paused "$line" && return 0
--- a/bin/fm-watch.sh
+++ b/bin/fm-watch.sh
@@ -152,9 +152,10 @@ STALE_ESCALATE_SECS=${FM_STALE_ESCALATE_SECS:-240}  # idle secs before a provabl
 # between completed turns, including long tool calls, builds, or test runs.
 BUSY_TURN_MAX_SECS=${FM_BUSY_TURN_MAX_SECS:-3600}
 # A crew that declared a pause is idling on a known external wait, so its stale
-# pane is absorbed rather than wedge-escalated.
-# A captain-held or paused crew whose agent has confidently exited uses the same
-# bounded cadence, while a live or ambiguously read agent still surfaces once.
+# pane is absorbed rather than wedge-escalated - whether the agent is still live
+# (the normal external-wait case) or has confidently exited. A captain-held
+# transfer whose agent has confidently exited uses the same bounded cadence,
+# while a still-live captain-held endpoint is suspect and still surfaces once.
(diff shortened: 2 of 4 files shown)


MattH-ca:
Closing: this PR was opened manually and therefore lacks the no-mistakes pipeline signature the required check looks for. Reopening it properly through `git push no-mistakes` with the fork configured as the push target.

PR #1703, opened 2026-08-04
fix(watcher): report why a cycle exited and verify an attach before claiming it

DigiThomas:
Fixes the two watcher-supervision defects reported in #1383.

## Defect 1 - a watcher cycle exits without saying why

`bin/fm-watch.sh` trapped `HUP`/`INT`/`TERM` to a bare `exit 1` and printed nothing, so a watcher torn down with its owning arm at a turn boundary was byte-identical to a crash. The reported line, `watcher: FAILED - watcher cycle exited 1 without an actionable reason`, named no step, no signal, and no stderr.

The watcher now tracks the phase it is in and reports it on any non-zero exit:

```
watcher: FAILED - watcher cycle exited <rc> during <step>[ after SIG<name>]
```

This covers a pre-lock refusal, a trapped signal, and a `set -u` abort. The exit code deliberately stays `1` rather than `128+signal`, because other layers key off it; the signal travels in the reason line.

The arm captures the watcher's stderr, replays it, and quotes a bounded tail in its own failure line. When the watcher could not print its own reason, the arm synthesizes one carrying the exit code, signal, lock and beacon state, queued-wake count, and that stderr tail. An arm interrupted at a turn boundary replays the watcher's own failure line from the captured stdout before deleting it, filtered to `^watcher: FAILED` only - an unfiltered replay would feed `bin/fm-claude-stop-autoarm.sh`, which runs the arm as `>"$OUT" 2>&1` and greps that output for actionable wakes, and would manufacture a wake for a cycle being torn down on purpose.

## Defect 2 - attach reports healthy against a dying cycle

`watcher: attached pid=<n> (beacon <t>)` is documented as proof one live cycle exists, but it was decided by a single read. The reporter observed a fresh beacon inside the grace with no watcher process left at all.

Each attach candidate must now keep verifying as this home's live watcher - process alive, lock identity unchanged, beacon fresh - for the whole `FM_ARM_ATTACH_VERIFY` window before the attached line is printed. The line reports the window it survived, plus `beacon advanced` when the beacon moved inside it, which is the only positive proof of progress rather than of recent existence.

A candidate that stops verifying is a failed attach, never a healthy one: the arm re-executes itself as `--restart`, bounded by `FM_ARM_RESTART_MAX` so a watcher that dies on every launch reports the failure instead of looping. A mid-window lock move to a pid that just passed the same liveness, identity and beacon gate is a handover, not a failure - the arm retargets onto that successor and verifies it from scratch, bounded by `FM_ARM_ATTACH_RETARGET_MAX`, rather than TERMing a watcher that is fine. Both knobs reject `0`, since a zero window or zero retarget budget reinstates exactly the behaviour this change removes.

## Root cause

Found, not merely instrumented. The lifecycle ledger recorded this shape three times on 2026-08-02: `exit_code=143 signal=TERM reason=arm-interrupted` on the owning arm, then a duplicate attached arm reporting a failure with `exit_code=unknown signal=unknown`. Sending `TERM` to a running watcher reproduces the upstream message exactly on pre-change code.

## Interaction with #1545

This branch is rebased onto #1545, which fixes the complementary defect of an attached arm reporting a *delivered* cycle as failed. Both survive: the identity-bound delivery ledger stays authoritative and is consulted first, so a cycle proven to have delivered a wake still exits 0 and prints no failure line. The typed failures here are reachable only for a cycle that delivered nothing. The verified successor identity is threaded through the retarget path so the new identity-mismatch condition stays truthful.

## Tests

Four regression cases in `tests/fm-watcher-lock.test.sh`, each proved to fail before the change:

- `test_nonzero_watcher_exit_reports_an_actionable_reason` - terminates a running watcher and refuses a pre-lock start, asserting the exit names its step, signal and stderr.
- `test_arm_refuses_to_attach_to_a_dying_watcher` - holds a lock with a fresh beacon through a holder that dies only after an ordered handshake proves the arm read it. No sleep budget.
- `test_arm_retargets_to_a_healthy_lock_successor` - hands the lock to a second genuinely healthy holder mid-window and asserts both the attach and that the successor is still alive. On pre-change code the arm execs `--restart` and TERMs it.
- `test_arm_signal_replays_watcher_failure_line` - TERMs an arm owning a live watcher, asserts the failure line is replayed, and asserts a seeded non-`watcher: FAILED` line is replayed onto neither stream.

Local results on macOS 26.6 (Darwin 25.6.0), bash 3.2.57, ShellCheck 0.11.0:

| suite | result |
|---|---|
| `fm-watcher-lock` | 34/34 |
| `fm-watch-arm` | 3/3 |
| `fm-claude-stop-autoarm` | 22/22 |
| `fm-turnend-guard` | 63/63 |
| `fm-arm-pretool-check` | 146/146 |
| `fm-watch-triage` | 46/46 |
| `fm-session-lock-ancestry` | 7/7 |
| `fm-subagent-pretool-check` | 13/13 |

`bin/fm-lint.sh` clean at pinned ShellCheck 0.11.0. `bin/fm-doc-audience-check.sh` clean.

`docs/watcher-continuity.md`, `docs/configuration.md` and `docs/verification/supervision.md` are updated to match, with dated before/after evidence for each case.

## Known limitation

The negative half of `test_arm_signal_replays_watcher_failure_line` depends on a seeded line surviving at a byte offset the watcher can overwrite, with a measured margin of 5 bytes. The positive assertion is unaffected and the test cannot fail falsely, but a longer step name or shorter temp root would silently retire that half of the coverage. Worth replacing with a positive assertion that the replayed stream contains only the failure line.


kunchenguid:

Issue #1694, opened 2026-08-04
wake-drain guard reports WATCHER DOWN and blames a stale beacon while printing a beacon age inside the grace window


akshatpatel64:
## Summary

The wake-drain supervision guard prints `WATCHER DOWN - SUPERVISION IS OFF` and blames a stale beacon, while printing a beacon age that is well inside the grace window. The check itself is correct; the *diagnostic* misattributes the cause, and the alarm fires during the normal gap between an actionable watcher exit and the next Stop-owned arm.

Observed on an up-to-date home (`fc3684a`), primary harness claude, tmux backend, macOS.

```
●  WATCHER DOWN - SUPERVISION IS OFF
●  2 task(s) in flight, but no watcher has a fresh beacon (last beat: 37s ago, grace 300s).
```

`37s ago` with `grace 300s` says the beacon *is* fresh, so the message contradicts its own evidence.

## Why it happens

`fm_watcher_healthy` (`bin/fm-wake-lib.sh:99`) requires three things together: a live pid from `state/.watch.lock/pid`, a lock identity matching that pid, and a beacon age under grace. Between cycles the watcher has exited and released the lock, so the pid check fails first and the function returns 1 while the beacon is still fresh.

The caller then renders that single false return as "no watcher has a fresh beacon" and appends the age. So a missing *process* is reported as a stale *beacon*, with a number that disproves staleness.

## Why it matters more than wording

That gap is the designed steady state, not a fault. The watcher queues an actionable wake and exits; the Stop hook arms the next cycle when the handling turn ends. During that window the guard declares supervision off, so a healthy fleet emits an alarm on essentially every wake.

Concretely, in one session this sent the supervisor chasing supervision twice. Each time the conclusion was the same: registration intact, `state/.claude-autoarm-epoch` freshly written with `outcome=rewake`, no failure markers, real wakes arriving with real reasons. Nothing was wrong. That is the cost, an alarm that reliably fires when nothing is wrong trains the reader to discount it, which is exactly what you do not want from a supervision guard.

Note this is adjacent to but distinct from #863 and #1383, both of which are about `fm-watch-arm.sh` exit classification. This one is the wake-drain guard's own liveness diagnostic, and it persists after #1545 landed.

## Suggested fix

Two small things, either independently useful.

[...]

Issue #1692, opened 2026-08-04
Firstmate burns through my credits too quickly because of excessive watcher events 


gamedevcb:
# Summary
When running Firstmate through my chatGPT subscription  (Sol= firstmate, Luna= crewmates), it rapidly burned through my weekly allowance. When I investigated where my credits went, it turned out that over 80% of them were used up by Sol, repeatedly resending the cache, thousands of times via watcher events triggering checking-in on crewmates. Not sure if this is a bug, but it is definitely undesirable.

# Overview

With Pi as the primary harness, one actionable Firstmate watcher wake can expand into many model provider requests because the injected operational message contains only the event reason and an instruction to run `bin/fm-wake-drain.sh`.

The model then reconstructs current task state through multiple separate tool/result cycles. Each cycle is another provider request carrying the primary session context.

This is not a request to suppress genuine actionable wakes or weaken lifecycle safety. It is a request to provide one bounded, structured wake-context packet so the primary agent does not have to reconstruct routine operational state incrementally.

## Environment

- Firstmate commit: `99533c5d7d3702050e6084429dddff6ea4fe1aa0`
- Primary harness: Pi
- Runtime backend: Herdr
- Primary model: GPT-5.6 Sol
- Platform: Linux
- Firstmate home: dedicated profile

## Measured evidence

After removing exact copied usage entries introduced by a parent-session clone/fork:

- 1,605 unique Sol provider-activity records
- 1,200 watcher-originated provider calls
- 178,589,184 of 220,382,208 unique cached-input tokens were watcher-originated
- watcher-originated activity therefore represented about 74.8% of calls and 81.0% of cached input

One inspected `watcher:signal` run expanded into 18 provider requests.

The sequence included:

1. `fm-wake-drain.sh`
2. read full report
3. `fm-crew-state.sh`
4. read status log
5. Git/evidence/decision checks
6. teardown
7. drain another wake
8. process and PID investigation
9. creation and dispatch of a follow-up review task
10. worker-state confirmation
11. final user-facing response

Some of these actions were substantively necessary. The performance problem is that routine state acquisition and lifecycle reconciliation required many separate model continuations.

## Current behavior

The Pi watcher injects a message equivalent to:

> Run bin/fm-wake-drain.sh first and handle the queued wake.

The model must then decide which state files, reports, endpoint checks, decision holds and repository checks to query.

## Expected behavior

[...]

Issue #1685, opened 2026-08-04
Stop-owned watcher auto-arm never claims home despite valid live session lock


schenchu:
## Symptom

`bin/fm-turnend-guard.sh --claude` fires "TURN WOULD END BLIND - SUPERVISION IS OFF" on every turn end for an entire session, with the Stop-owned auto-arm (`bin/fm-claude-stop-autoarm.sh`) never claiming the home, despite:

- The session lock (`state/.lock`) holding a live harness pid the whole time (`bin/fm-lock.sh status` -> `lock: held by live harness pid <N>`).
- `.claude/settings.json` correctly registering both the `Stop` guard hook and the `asyncRewake` auto-arm hook.
- Manually running `bin/fm-claude-stop-autoarm.sh` directly always exits 0 with no output/error, but never advances recovery.
- One manual `bin/fm-watch-arm.sh` run did successfully start a watcher with a fresh beacon (`watcher: started pid=<N> (beacon fresh)`), but it ran a single bounded cycle and exited rather than persisting, so the beacon went stale again shortly after (matching the documented design where continuity is meant to come from the Stop hook re-arming every turn, not from one long-lived process).

## Evidence

- `state/.claude-autoarm-epoch` stayed at `epoch=28 owner_pid=<pid-A> outcome=rewake updated_at=<ts>` unchanged across many turns and several manual auto-arm invocations, even though the session's live lock-holder pid was a *different* pid (`<pid-B>`) for the whole session - the epoch record never advanced to reflect the current session identity.
- No failure-notice or failure-alarm files were ever created (`state/.claude-autoarm-failure-notified`, `state/.claude-autoarm-failure-alarmed` both absent throughout), so the auto-arm's own typed-failure/notice path never triggered even though it also never succeeded - it appears to be silently no-op'ing rather than either succeeding or recording a diagnosable failure.
- `state/.turnend-claude-blocks` stayed at `count=1` for `epoch=28` across multiple separate guard firings in the same session, which never advanced enough to trip the documented bounded attended fail-open.
- The guard kept repeating the same "same stale episode" warning each poll for related commands (`fm-send.sh`, `fm-wake-drain.sh`), consistent with the beacon staying stale for the whole session.

## Impact

[...]

PR #1734, opened 2026-08-05
fix: accommodate mode-inert filesystems in PR-check migration

AhmadHammad21:
## Summary

On a filesystem that cannot enforce POSIX permission bits - a WSL2 `v9fs`/DrvFs mount of a Windows drive reads every file back as `0777` no matter what `chmod` requested - the PR-check migration's exact `0600`/`0700` assertions can never be satisfied. `fm-pr-check-migrate.sh --checks-safe` then exits 1 ("PR check migration blocked; refusing to execute state checks") and the watcher never arms, so supervision stays down for the whole session on a home whose `state/` lives on a Windows drive.

## Change

- Add a cached `fm_pr_fs_enforces_mode` probe: create a private temp file, `chmod 0600`, read the mode back, and cache the verdict per device. It defaults to "enforced" whenever it cannot run, so a real POSIX filesystem is never weakened by a failed probe.
- Add `fm_pr_mode_matches`, which skips **only** the exact-mode equality on a mode-inert filesystem and keeps every other protection at the call site (regular file, single hard link, expected device, content) fully in force.
- Route the private-artifact and quarantine mode assertions through the new helper.

## Testing

- New coverage in `tests/fm-pr-check-security.test.sh` exercises both an enforcing filesystem (strict assertions still apply) and a mode-inert one (equality skipped, other guards intact).
- Verified `fm-pr-check-migrate.sh --checks-safe` returns 0 on an affected `v9fs` `state/` directory.

Related watcher-down reports on Windows-drive homes: #1692, #1685, #1694.

### DIFF
--- a/bin/fm-pr-check-migrate.sh
+++ b/bin/fm-pr-check-migrate.sh
@@ -6,6 +6,12 @@
 # registered custom checks remain armed, and every other task poll is
 # quarantined for private review. A current X-mode shim is preserved by exact
 # content, while the recognized older byte-static shim is refreshed in place.
+#
+# Private artifacts are expected at mode 0600 and the quarantine directory at
+# 0700, but some filesystems cannot enforce POSIX permission bits (a WSL2 v9fs
+# or DrvFs mount of a Windows drive reads every file back as 0777). There the
+# exact-mode assertions are skipped through fm_pr_mode_matches while every other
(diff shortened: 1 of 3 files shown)


AhmadHammad21:
Closing: this PR was raised by hand and cannot satisfy the `Require no-mistakes` check. Re-raising the same fix through the no-mistakes pipeline, which stamps the required Pipeline signature into the PR body.

PR #1742, opened 2026-08-05
fix(bin): count only live endpoints as live lanes

willchen95:
## Intent

Close a false-liveness defect in the watcher: `count_live_lanes` counted recorded `state/*.meta` files rather than live endpoints, so a home whose endpoints had all died still reported work in progress.

Reproduced live before the change: five recorded tasks, every endpoint resolving `missing` through `fm_backend_agent_state`, and the captain pulse still reporting live lanes. The same count gates the refill-aware heartbeat, so a fully drained fleet stayed above its refill target and never refilled — the failure mode that rule was added to prevent.

## What Changed

- `count_live_lanes` now excludes a recorded non-secondmate task whose endpoint is authoritatively gone. Only `dead` (the backend proving the endpoint missing or the agent gone) is excluded; `unknown` still counts, so an unreadable backend can never inflate dispatch by under-reporting. Cost is one bounded backend probe per recorded non-secondmate task.
- New `count_recorded_lanes`; the captain pulse prints `live-lanes=N (of M recorded)` when the two differ, so a fleet of dead records reads as dead instead of as work.
- Two regression tests in `tests/fm-watch-triage.test.sh`: the counter contract across dead/unknown/alive endpoints, and the pulse line itself under an all-dead fleet.
- Separately, the two competing captain-facing rule blocks at the top of `AGENTS.md` are merged into one ordered MAXIMUM RULE (DECIDE, WRITE, TELL THE TRUTH), with the honesty section outranking the brevity section.

## Risk Assessment

Low. The change only narrows a count, and narrows it in the safe direction: an unreadable or ambiguous backend keeps counting as live, so the refill gate cannot over-dispatch on a read failure. The pulse string gains a parenthetical only when live and recorded disagree.

## Testing

`bin/fm-test-run.sh tests/fm-watch-triage.test.sh` — 0 failures, exit 0, including both new assertions:

```
ok - a live lane is a recorded task with an endpoint that is not authoritatively gone
ok - the captain pulse reports a fleet of dead endpoints as zero live lanes
FM_TEST_END tests/fm-watch-triage.test.sh exit=0 duration_ms=157472
```

`bin/fm-lint.sh` exits 0. The full suite was not run for this change.

### DIFF
--- a/.agents/skills/afk/SKILL.md
+++ b/.agents/skills/afk/SKILL.md
@@ -100,12 +100,13 @@ backend (tmux or herdr; see "Auto-discovered supervisor pane" below):
   It preserves idle bordered composers such as claude's `│ > … │` and bare agent glyphs as empty, but a bare shell glyph is unknown unless inside a genuine bordered composer box; see `docs/herdr-backend.md` "Composer and injection safety" for the complete contract.
   `pane_input_pending` remains the tested predicate for callers that only need to know whether real unsubmitted text is present, but it is insufficient for an injection-safety decision because it cannot distinguish `empty` from `unknown`.
 
-A busy primary pane, or any composer verdict other than `empty`, defers the injection; the buffered escalation survives in `state/.subsuper-escalations` and is retried on the next housekeeping tick.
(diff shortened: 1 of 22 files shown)


willchen95:
Closing: opened without authorization from the repository owner's side. Withdrawing.

PR #1743, opened 2026-08-05
fix(bin): stop reading an unpushed pipeline head as run divergence

mremond:
## The defect

`bin/fm-crew-state.sh` reported `state: failed · source: run-step · run failed` for a no-mistakes run that was alive and validating normally.

`AGENTS.md` section 7 makes this script the authority for judging a validation: *"Judge validation by the current-code-matched run step through `bin/fm-crew-state.sh`, not by shell liveness or the last status event."* A `failed` verdict on a live run invites tearing down or restarting a worker mid-pipeline that still holds unlanded work, and it reaches a supervisor as a `stale:` wake — already disposed to act.

Reproduced live on task `delegated-run-durability`, and every link confirmed by running it, not by reading it:

| step | evidence |
|---|---|
| verdict | `state: failed · source: run-step · run failed` |
| `axi status` | run `01KZ8JGTAW3DN81Y658CVWYKJ8`, branch matched exactly, `status: running`, step `document`, `agent_pid` alive |
| `git rev-parse --verify 16f026f7^{commit}` | `fatal` — the object is not in the worktree |
| `no-mistakes runs` | `running … 16f026f7` (skipped as unresolvable) then `failed … 2e96f463` (matched local HEAD) |
| result | `RUN_SOURCE=coarse`, `COARSE_STATUS=failed` |

## Root cause: two facts conflated

`fm_nm_head_matches_worktree` resolved the run head with a single `git rev-parse --verify <run_head>^{commit}` and treated any failure as "not our run". That call fails for two unrelated reasons:

1. **The run head is on a divergent line of history** — real evidence of non-ownership.
2. **The run head is simply not an object in this worktree** — no evidence either way.

The gate commits into its own copy, so cause 2 is the *normal* mid-run condition. Attribution was lost, the coarse runs-list fallback took over, skipped the same unresolvable row, and answered from an **older, superseded `failed` run of the same branch**. That is how a missed attribution became a wrong *terminal* verdict.

### The window is wider than it first appears

The original report scoped this to the `document` and `lint` steps. Checking the same live run again after it pushed and opened a PR, its head was **still** unresolvable (`rc=2`) at step `ci` — because the gate's push goes to the **remote**, not into the crew's worktree. The window therefore spans from the first pipeline commit through push, pr and ci: most of every no-mistakes delivery.

That result is why the fix is deliberately **not** gated on `phase: pre_push`. Doing so would have left the same false verdict live across the second half of every run.

## What the head check was protecting, stated before relaxing it

Branch name alone is not identity. The same branch name is reused across stages and across crews, so a **historical run** — one whose head was rewritten, or one the local tree has since committed past — must never be reported as this crew's current state.

That protection is a claim about **divergent history**, and it needs both commits in hand to make it. It says nothing about a head that is merely absent. Relaxing the absent case therefore gives up none of it.

## The fix

`fm_nm_head_matches_worktree` becomes **tri-state**:

| verdict | meaning |
|---|---|
| `0` match | equal commits, or worktree HEAD is an ancestor of the run head |
| `1` mismatch | both resolve here and the run head is a strict ancestor or diverged; also unbindable inputs (missing head, unreadable worktree) |
| `2` unresolvable | the run head is not an object here at all |

Divergence rejects exactly as before. An unresolvable head is settled from `branch_sync` rather than from the sha, and binds only when **all four** line up:

1. `branch_sync.pipeline.run` is the run being attributed — the report is about *this run*.
2. `branch_sync.pipeline.current_head` prefix-matches the head that failed to resolve — its absence is *explained*, not assumed.
3. `branch_sync.local.head` is this worktree's HEAD — the report is about *this tree*. `axi status` answers repo-wide, and two worktrees can carry the same branch name.
4. `branch_sync.pipeline.submitted_head` passes the **unchanged** code-identity rule. It is resolvable precisely because it predates the pipeline's own unpushed commits, and this is where a rewritten or locally-advanced branch is still caught.

Any one absent — **including an older gate that emits no `branch_sync`** — keeps the previous conservative rejection. The relaxation is evidence-driven, never assumed.

`branch_sync` repeats key names the run block already uses (`head`, `status`, `run`, `branch`), so the existing unscoped `fm_nm_field` would answer from the wrong block. A scoped `fm_nm_branch_sync_field` was added rather than loosening `fm_nm_field`, and verified against real v1.41.2 output before being trusted.

### The coarse fallback stays a fallback

It carries no `branch_sync`, so an unresolvable sha cannot be corroborated there. It now distinguishes the same three outcomes: `match` answers, `mismatch` keeps scanning (unchanged — an older row on the same branch can still be ours), and `unresolvable` **declines** rather than falling through to the older rows that the live run supersedes.

## Tests

Six added to `tests/fm-crew-state.test.sh`, driven through the real executable with throwaway git repos and fake `no-mistakes`/`tmux`.

Two reproduce the defect — they fail against pre-fix code with the exact reported output and pass after:

- a live run whose head is absent locally reports `working`, not `failed`
- the coarse fallback declines instead of answering from a superseded row

Four pin the protection that had to survive:

- no `branch_sync` → conservative rejection preserved
- diverged `submitted_head` on a rewritten branch → rejected
- `branch_sync` naming a different run → rejected
- `branch_sync` reporting a different local head → rejected

Each of the four corroboration conditions was **mutation-tested individually** and confirmed to be caught by its own test, so none is vacuous. A `UNRESOLVABLE_SHA` fixture plus an `assert_absent_object` guard keep the "object absent" precondition from going quietly vacuous either.

## Related finding, not fixed here

`bin/fm-teardown.sh:1155` shares this one-owner function and has the same conflation. Its `|| return 1` call means this change leaves its behavior **byte-identical** — it still refuses to bind on an unresolvable head. It is fail-open by design ("never a guess"), but it can fail to abort a genuinely-parked run it owns during the same window. Reported rather than changed, to keep this PR scoped to the reported defect.

## Verification

- `tests/fm-crew-state.test.sh` green; `fm-teardown`, `fm-gotmp`, `fm-wake-queue`, `fm-watch-triage` green.
- `bin/fm-lint.sh` clean under the pinned ShellCheck 0.11.0.
- Live re-check after the fix: `state: working · source: run-step · validating (running)`.
- `tests/fm-backend.test.sh` and `tests/fm-backend-orca.test.sh` fail on an old-teardown conformance case with `fm-decision-hold: compatible tasks-axi is required`. Verified to fail **identically on the pristine base** — pre-existing and environmental, unrelated to this change.

### DIFF
--- a/bin/fm-crew-state.sh
+++ b/bin/fm-crew-state.sh
@@ -28,9 +28,10 @@
 #      is an ancestor of the run head (pipeline fix commits advanced the run on
 #      the same line of history). Local work that advanced past the run head, or
 #      diverged from it, invalidates attribution. A run head that does not
-#      resolve here at all is a THIRD case, not divergence: the pipeline commits
-#      before it pushes, so it is routine and expected mid-run. It is settled
-#      from branch_sync (see nm_run_pipeline_owns_branch), never from the sha.
+#      resolve here at all is a THIRD case, not divergence: the pipeline writes
+#      commits in its own copy, so its advanced head can remain absent from this
+#      worktree throughout the run, including after it pushes. It is settled from
+#      branch_sync (see nm_run_pipeline_owns_branch), never from the sha.
 #      The run-step is AUTHORITATIVE: running/fixing -> working, ci -> working,
 #      awaiting_approval/fix_review -> parked (with gate findings), terminal
 #      passed/checks-passed -> done, failed/cancelled -> failed. EXCEPT: while
@@ -423,7 +424,7 @@ nm_run_pipeline_owns_branch() {  # <unresolvable-run-head>
   # 4. Same code: the commit this run took custody of binds to our code identity
   #    under the UNCHANGED rule. This is where a rewritten or locally-advanced
   #    branch is still caught - the submitted head is resolvable precisely
-  #    because it predates the pipeline's own unpushed commits.
+  #    because it predates the pipeline's own commits.
   submitted=$(strip_quotes "$(fm_nm_branch_sync_field "$RUN_OUT" pipeline submitted_head)")
   fm_nm_head_matches_worktree "$WT" "$submitted"
 }
--- a/bin/fm-crew-state.sh
+++ b/bin/fm-crew-state.sh
@@ -27,7 +27,10 @@
 #      A run matches when its head equals the worktree HEAD, or the worktree HEAD
 #      is an ancestor of the run head (pipeline fix commits advanced the run on
 #      the same line of history). Local work that advanced past the run head, or
-#      diverged from it, invalidates attribution.
+#      diverged from it, invalidates attribution. A run head that does not
+#      resolve here at all is a THIRD case, not divergence: the pipeline commits
+#      before it pushes, so it is routine and expected mid-run. It is settled
+#      from branch_sync (see nm_run_pipeline_owns_branch), never from the sha.
 #      The run-step is AUTHORITATIVE: running/fixing -> working, ci -> working,
 #      awaiting_approval/fix_review -> parked (with gate findings), terminal
 #      passed/checks-passed -> done, failed/cancelled -> failed. EXCEPT: while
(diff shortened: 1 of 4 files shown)


mremond:
Closing in favour of a pipeline-raised PR for the same branch.

This repository requires PRs to be raised through `no-mistakes` (`.github/workflows/no-mistakes-required.yml` checks the PR body for the pipeline signature), so a hand-opened PR cannot pass its checks by design. Re-opening this through the gate's fork route instead, which writes that signature itself. Same branch, same validated head `acfab5b8`.

PR #1744, opened 2026-08-05
fix(bin): preserve live run attribution for pipeline-owned heads

mremond:
## The problem in one sentence

`bin/fm-crew-state.sh` reports `state: failed` for a validation run that is alive, healthy, and running normally — and it does so during the *majority* of every pipeline run.

## What you would see

```
$ bin/fm-crew-state.sh delegated-run-durability
state: failed · source: run-step · run failed

$ (cd <that worktree> && no-mistakes axi status)
status:  running
step:    document (in progress)
agent_pid: alive
```

Two sources of truth, one of them confidently wrong. Reproduced and traced on 2026-08-05 against task `delegated-run-durability`, gate v1.41.2.

## Why it happens

Firstmate decides "is this run mine?" by comparing the run's head commit against the crew worktree. That comparison lives in `fm_nm_head_matches_worktree` (`bin/fm-nm-run-lib.sh`):

```sh
run_full=$(git -C "$wt" rev-parse --verify "${run_head}^{commit}" 2>/dev/null) || return 1
```

Here is the trap. The pipeline commits **into the worktree's branch, then pushes later**. Between those two moments the run's head is a commit that the worktree cannot resolve — not because it belongs to someone else, but because it does not exist locally *yet*.

`rev-parse` fails. The function returns 1. And 1 already means something else: *"this run is not yours."*

So the chain runs:

1. `axi status` answers correctly — right run id, branch identical to the crew's branch.
2. The head check cannot resolve the object and returns "not a match".
3. Attribution is declared **missed**, despite an exact branch match.
4. Control falls through to the coarse runs-list fallback.
5. That fallback returns a *stale, older row* — which says `failed`.

The verdict is not a wrong reading of this run. It is the right reading of a **different, dead run**.

## Why this is not cosmetic

`AGENTS.md` section 7 makes this script **the authority** for judging a validation. A `failed` verdict on a live run invites tearing down or restarting a worker that is mid-pipeline and holding unlanded work — hard rule 3 territory. The verdict also surfaces as a `stale:` wake, so it reaches a supervisor already primed to intervene.

## The window is wider than it first looked

The original report guessed the bad window was `document`/`lint` — i.e. `phase: pre_push`. Re-checking the same live run *after* it had pushed and opened a PR showed its head **still unresolvable**, at step `ci`. The reason: the gate pushes to the **remote**, never back into the crew's worktree.

This is why the fix is deliberately **not** gated on `pre_push`. Doing so would have left the same false verdict live across `push`, `pr` and `ci`.

## What this PR changes

**1. The head check becomes tri-state.** `fm_nm_head_matches_worktree` now separates two situations it used to conflate:

| Return | Meaning |
| --- | --- |
| `0` | Heads match — the run is ours. |
| `1` | **Proven divergence**, or unbindable inputs (missing head, unreadable worktree). Rejected exactly as before. |
| `2` | **Unresolvable** — the run head is simply not an object in this worktree. Not proof of anything. |

An absent object and a diverged history are different facts. Only the second is evidence of non-ownership.

**2. An unresolvable head is settled by evidence, never by the sha.** A new `nm_run_pipeline_owns_branch` in `fm-crew-state.sh` corroborates from the gate's own `branch_sync` block, and requires **all four** to line up:

1. `branch_sync.pipeline.run` equals the run id being attributed;
2. `branch_sync.pipeline.current_head` prefix-matches the head that failed to resolve — so its absence is *explained* rather than assumed;
3. `branch_sync.local.head` equals this worktree's `HEAD` — because `axi status` answers repo-wide, and two worktrees can carry the same branch name;
4. `branch_sync.pipeline.submitted_head` passes the unchanged code-identity rule — resolvable precisely because it predates the pipeline's own unpushed commits, and this is where a rewritten or locally-advanced branch is still caught.

Any one of the four missing — **including an older gate that emits no `branch_sync` at all** — keeps the previous conservative rejection. The relaxation is evidence-driven; it is never assumed.

**3. The coarse fallback stops reporting superseded failures.** This is the same conflation in the same one-owner function, and it is what turned a *missed attribution* into a *wrong terminal verdict*. It now distinguishes three outcomes: match answers; mismatch keeps scanning (unchanged — an older row on the same branch can still be ours); unresolvable **declines** rather than falling through to older rows that the live run supersedes.

## What is deliberately preserved

The head check exists to stop firstmate attributing to one crew a run that does not concern it. **A wrong attachment in that direction is at least as dangerous as the defect being fixed.** The goal here was to *distinguish two conflated situations*, never to silence a control.

Four tests pin exactly that, and each was mutation-tested individually to prove it is not vacuous:

- no `branch_sync` → conservative rejection;
- diverged `submitted_head` on a rewritten branch → rejected;
- `branch_sync` naming a different run → rejected;
- `branch_sync` reporting a different local head → rejected.

`bin/fm-teardown.sh` shares this one-owner function through a plain `|| return 1` caller (`fm-teardown.sh:1155`). Its behaviour is **byte-identical**: it still refuses to bind on an unresolvable head. Teardown was not changed — the reported defect was scoped to `fm-crew-state.sh`, and the same conflation there is reported without expanding scope.

## Tests

Six tests added to `tests/fm-crew-state.test.sh`, colocated per repo convention, driven through the real executable with fake `no-mistakes`/`tmux` and throwaway git repositories.

- **Two fail against pre-fix code** with the exact reported output, and pass after.
- **Four pin the protections above.**
- An `UNRESOLVABLE_SHA` fixture plus an `assert_absent_object` guard keep the "object absent" precondition from going quietly vacuous.

The second trap called out in the brief was *verifying by reading the code rather than by running it* — the defect itself was born that way. Every claim here was reproduced and watched.

## Verification

- `tests/fm-crew-state.test.sh` green; `fm-teardown`, `fm-gotmp`, `fm-wake-queue`, `fm-watch-triage` green.
- `bin/fm-lint.sh` clean under the pinned ShellCheck 0.11.0.
- `tests/fm-backend.test.sh` and `tests/fm-backend-orca.test.sh` fail on an old-teardown conformance case (`fm-decision-hold: compatible tasks-axi is required`) — verified to fail **identically on the pristine base**, so pre-existing and environmental, not caused by this change.

A scoped `fm_nm_branch_sync_field` was added rather than loosening `fm_nm_field`, because `branch_sync` repeats key names the run block already uses (`head`, `status`, `run`, `branch`) and the unscoped reader would answer from the wrong block. Verified against real v1.41.2 output before being trusted.

## Pipeline

| Step | Outcome |
| --- | --- |
| intent | ✅ passed |
| rebase | ✅ passed |
| review | ⏭️ skipped — see the note below |
| test | ⏭️ skipped — see the note below |
| document | ⏭️ skipped — see the note below |
| lint | ⏭️ skipped — see the note below |
| push | ✅ passed |

## Note on the gate's step summary

The gate's own summary for the run that published this PR marks `review`, `test`, `document` and `lint` as **skipped**. They were not skipped for this work.

They passed in run `01KZ8PGA3SYQ2SW0E1Y5E2Q72Y` — review 0 findings, test 0 findings — and only the `push` step failed there, on a 403, because the authoring account has no write access to this repository. Publication was then re-routed through a fork. The head being published is byte-identical to the head that passed, on an unchanged base, so the steps were resumed rather than restarted and the same code was not re-validated from scratch.


kunchenguid:

PR #1752, opened 2026-08-05
fix(bin): bound the bootstrap gh auth status probe

cipherholdingsllc:
## Problem

`bin/fm-bootstrap.sh` probes GitHub auth with a bare `gh auth status`, which can hang indefinitely (interactive prompt, network stall). A hung probe blocks bootstrap and anything waiting on it, and a timeout is indistinguishable from being unauthenticated.

## Change

- Adds `gh_auth_probe`: the same single `gh auth status` call, but bounded (default 10s, override via `FM_GH_AUTH_TIMEOUT_SECS`, non-numeric/zero values reset to 10), with prompting and update notifications disabled, using `timeout`, `gtimeout`, or a Perl process-group alarm as fallback.
- Adds `gh_auth_report`: exit 0 stays silent by default, ordinary nonzero keeps the legacy `NEEDS_GH_AUTH` line, and 124 reports `GH_AUTH: indeterminate (probe timed out after Ns)` so a timeout is never misreported as unauthenticated. `FM_BOOTSTRAP_GH_AUTH_DIAGNOSTICS=1` opts into explicit typed lines for the authenticated and unauthenticated states too.
- Existing default output is unchanged for the authenticated and unauthenticated cases.

## Tests

Two new hermetic cases in `tests/fm-bootstrap.test.sh` (fake `gh`; no network, no credentials): typed-state reporting for authenticated / unauthenticated / hung probes with a bound-time assertion, and a mirror of the pilot preflight capture proving `PREFLIGHT_RESULT` is reached in all three states. Full suite: 23 ok / 0 not ok, exit 0.

Note: ShellCheck is not installed on the authoring machine, so `bin/fm-lint.sh` was not run locally; relying on CI for lint.

### DIFF
--- a/bin/fm-bootstrap.sh
+++ b/bin/fm-bootstrap.sh
@@ -7,6 +7,7 @@
 #          Silent = all good.
 #          Lines: "MISSING: <tool> (install: <command>)",
 #                 "MISSING_MANUAL: <tool> (instructions: <url>)", "NEEDS_GH_AUTH",
+#                 "GH_AUTH: authenticated|not authenticated|indeterminate (probe timed out after <seconds>s)"
 #                 "BACKEND_INVALID: <name> (known: <names>)",
 #                 "STARTUP_MEMORY_BUDGET: invalid config/startup-memory-budget - <reason>",
 #                 "CREW_DISPATCH: invalid config/crew-dispatch.json - <reason>",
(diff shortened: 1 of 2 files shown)


cipherholdingsllc:
Closing: raised directly with `gh pr create`, so it missed the `no-mistakes` gate required by CONTRIBUTING. Re-raising the same change through the gate.

PR #1766, opened 2026-08-06
fix(bin): accept opencode/kimi effort as documentation in dispatch validator

x45dev:
## Problem

The crew-dispatch.json validator in `fm-bootstrap.sh` rejected any `effort` field on an opencode or kimi profile, even though `fm-spawn.sh` already accepts it: `effort_flag_for_harness` records the requested `effort=` in task meta and emits no flag for harnesses without a verified effort knob (`harness-adapters` confirms opencode's interactive launch has none).

The validator was the outlier, forcing operators to drop effort fields from their routing rules to unblock dispatch, losing the recorded intent.

## Fix

Align the validator with `fm-spawn`'s behavior: any string effort value on an opencode or kimi profile is accepted as documentation. The two test rows that expected rejection now expect acceptance.

## Blast radius

Low. The change only relaxes validation to match existing spawn behavior; no new code paths, no behavior change for other harnesses.

### DIFF
--- a/bin/fm-bootstrap.sh
+++ b/bin/fm-bootstrap.sh
@@ -742,7 +742,7 @@ crew_dispatch_validate() {
       elif $h == "codex" then (["low","medium","high","xhigh"] | index($e))
       elif $h == "grok" then (["low","medium","high"] | index($e))
       elif $h == "pi" or $h == "pi-signed" then (["low","medium","high","xhigh","max"] | index($e))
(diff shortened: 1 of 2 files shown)


x45dev:
Closing to re-run through no-mistakes per the repo's PR-source policy. All 12 functional CI checks pass; only the policy check failed. Re-raising on a fresh branch from current main.

PR #1774, opened 2026-08-06
fix(orca): validate composite worktree identities

### LINKED ISSUE #1227: Orca teardown refuses every task after #1171: composite orca_worktree_id fails fm_backend_endpoint_atom_valid
## What happens

After updating to `c0c0881`, **every Orca-backed task refuses teardown**:

```
REFUSED: Orca endpoint metadata for task <id> is malformed or inconsistent; preserving task state.
```

The metadata is not malformed. It is exactly what `bin/fm-spawn.sh` wrote.

## Cause

`fbece9c` (#1171, "harden endpoint cleanup") routes the Orca branch's `orca_worktree_id` through `fm_backend_endpoint_atom_valid` (`bin/fm-backend.sh:380`), which rejects any value containing a character outside `A-Za-z0-9._@%+-`:

```sh
fm_backend_endpoint_atom_valid() {  # <value>
  case "$1" in
    ''|*[!A-Za-z0-9._@%+-]*) return 1 ;;
  esac
}
```

But Orca's worktree id is a **composite** `<uuid>::<absolute-path>`, written by `fm-spawn.sh:297` / `:1464`:

```
orca_worktree_id=e72d88c7-db30-4ae0-89e1-e826fe4ce584::/Users/rodrigocampos/orca/workspaces/estimate-os/fm-audit-est-product-security
```

`:` and `/` are both outside the allowed set, so the check can never pass for a well-formed Orca task. The validator is correct for single-atom endpoints such as a tmux pane or a terminal handle (`terminal=` passes fine); it is simply the wrong predicate for this composite field.

## Impact

Teardown is the only guarded path that closes a task, and it now refuses for every Orca task on a home that has updated. Landed work cannot be cleaned up, task metadata stays live, and the watcher keeps raising stale wakes for a finished task indefinitely — one supervision turn burned per wake.

The refusal is safe (it preserves state rather than destroying anything), but it is unconditional, so `--force` is the only way past, which is exactly the wrong habit to build around an unlanded-work guard.

## Reproduce

1. Spawn any scout or ship task with `--backend orca`.
2. Let it finish and land.
3. Run `bin/fm-teardown.sh <id>`.

Observed on firstmate `c0c0881`, macOS 25.5.0, bash 3.2. Three Orca tasks torn down cleanly on `a5fe1bc` earlier the same day; the first teardown attempted after the update refused.

## Suggested fix

Validate the two halves separately rather than treating the composite as one atom — the uuid half against `fm_backend_endpoint_atom_valid`, and the path half against whatever the tmux/herdr worktree paths already use. That keeps the hardening intent (#1171 is guarding against injected or ambiguous endpoint values) without rejecting the format firstmate itself produces.

pawridisepetmanagement:
## Summary

- validate Orca worktree identities as the generated `repo-id::absolute-worktree-path` composite
- preserve exact terminal/task binding and fail-closed resolved-path ownership checks
- add regression coverage for valid, malformed, ambiguous, and teardown identities
- document the composite identity contract

Fixes #1227.

## Validation

- `bash tests/fm-backend-orca.test.sh`
- `bash tests/fm-teardown-endpoint-safety.test.sh`

Both focused suites pass.


kunchenguid:

PR #1775, opened 2026-08-06
fix(bin): bound the bootstrap gh auth probe and report unfinished probes as indeterminate

cipherholdingsllc:
## Problem

`bin/fm-bootstrap.sh` probed GitHub auth with a bare `gh auth status`. That call is unbounded, so a stalled network or a credential helper waiting on approval hangs session start indefinitely. And because any non-zero result printed `NEEDS_GH_AUTH`, a probe that never answered was indistinguishable from a confirmed logout.

## Change

`gh_auth_report` makes the same single `gh auth status` call, bounded at 10s (`FM_GH_AUTH_TIMEOUT_SECS`; non-numeric, zero, and zero-padded values reset to 10) via `timeout`, `gtimeout`, or a Perl process-group alarm. Stock macOS ships none of the GNU tools, so that fallback chain is load-bearing rather than defensive.

An unfinished probe now reports `GH_AUTH: indeterminate`, with distinct wording for its three causes: the bound expired, the probe was killed before answering, or the home has no bounding tool at all. `AGENTS.md` and `.agents/skills/bootstrap-diagnostics/SKILL.md` register the `GH_AUTH:` prefix, so a session that sees only that line has a defined response.

Default output for the two determinate states is unchanged: authenticated stays silent, unauthenticated still prints exactly `NEEDS_GH_AUTH`. The typed lines are opt-in behind `FM_BOOTSTRAP_GH_AUTH_DIAGNOSTICS=1`.

## Two deliberate behavior changes

Flagging these rather than leaving them in the diff.

Signal-terminated probes now report indeterminate where they previously fell through to `NEEDS_GH_AUTH`. A killed probe proves nothing about the credential. Every status at or above 128 routes through one classifier, so the answer no longer depends on which bounding tool the machine happens to have.

Unifying that message required unifying the kill grace, so both branches now escalate after a shared `GH_AUTH_KILL_GRACE_SECS=2`. The Perl fallback's worst case widens from bound plus 0.2s to bound plus 2s on stock macOS. A per-branch grace would have let the installed bounder decide a slow-but-cooperative probe's wording.

## Tests

Hermetic cases (fake `gh`, no network, no credentials) that shim `PATH` to force each branch of the chain rather than exercising only whichever tool the host ships. Suite: 27 passing. `bin/fm-lint.sh` clean on pinned ShellCheck 0.11.0.

One gap I would rather name than hide: the Perl branch's SIGKILL escalation is verified by hand, not pinned by a test. Reverting that handler to a bare `exit 124` would leave the suite green. A `wedged` fixture closes it, and I am happy to add it here or as a follow-up.

## Follow-up, deliberately out of scope

`bin/fm-bearings-snapshot.sh`, `bin/fm-fleet-snapshot.sh`, and `bin/fm-nm-run-lib.sh` carry the same timeout/gtimeout/Perl chain. All three still end in `exit($? >> 8)`, the signal-status bug fixed here, and none passes `-k`, so the four implementations now disagree about both hardening measures. A shared `fm_bounded_run` helper would give one implementation. Kept out of this PR to keep the diff reviewable; happy to send it separately.

## Pipeline

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>

- ⚠️ `bin/fm-bootstrap.sh:108` - Zero-padded zero values bypass the sanitizer and disable the bound entirely. The guard is `&#39;&#39;|*[!0-9]*|0)`, so `FM_GH_AUTH_TIMEOUT_SECS=00` (or `000`) is non-empty, all-digits, and not literally `0`, and passes through unchanged. Verified: `timeout 0 sleep 4` exits 0 after the full 4s (GNU timeout treats 0 as no limit), and perl&#39;s `alarm(&#34;00&#34;)` numifies to 0, which cancels any pending alarm — the probe snippet with t=00 against `sleep 20` was still running after 7s. Both the primary and fallback paths therefore hang unboundedly, reintroducing exactly the failure this change exists to prevent. This contradicts the intent criterion &#34;non-numeric and zero values sanitized back to 10&#34;: `00` is a zero value that is not sanitized. Fix by rejecting all-zero strings (e.g. add `0*` to the case pattern, or compare arithmetically after the digit check).
- ⚠️ `bin/fm-bootstrap.sh:238` - The Perl fallback&#39;s `exit($? &gt;&gt; 8)` discards signal-termination status, so a `gh auth status` killed by a signal is reported as authenticated and bootstrap stays silent. Verified by running the exact snippet against a child that does `kill -SEGV $$`: the probe exits 0, whereas invoking the child directly gives 139. The `timeout`/`gtimeout` branches return 128+N here (-&gt; NEEDS_GH_AUTH), and so did the pre-change direct call, so the fallback silently swallows a real auth failure signal on precisely the stock-macOS path this chain exists to serve. Map signal death to a non-zero, non-124 status, e.g. `my $st = $?; exit($st &amp; 127 ? 128 + ($st &amp; 127) : $st &gt;&gt; 8);`.
- ⚠️ `bin/fm-bootstrap.sh:10` - `GH_AUTH: indeterminate (probe timed out after &lt;n&gt;s)` prints unconditionally but is not registered as an actionable diagnostic anywhere the agent reads. AGENTS.md:501 and .agents/skills/bootstrap-diagnostics/SKILL.md both enumerate the trigger prefixes (MISSING, NEEDS_GH_AUTH, TANGLE, FLEET_SYNC, ...) and neither lists GH_AUTH, and the skill has no per-line handling entry for it. A session whose only bootstrap output is the indeterminate line therefore does not load the playbook and has no defined response, which defeats the point of distinguishing indeterminate from unauthenticated. The header line added here also presents all three GH_AUTH states as if they always print, without noting that `authenticated` and `not authenticated` require FM_BOOTSTRAP_GH_AUTH_DIAGNOSTICS=1.
- ℹ️ `bin/fm-bootstrap.sh:106` - `FM_GH_AUTH_TIMEOUT_SECS` and `FM_BOOTSTRAP_GH_AUTH_DIAGNOSTICS` are absent from the canonical env-var block in docs/configuration.md, which documents sibling bootstrap knobs `FM_FLEET_SYNC_BOOTSTRAP_TIMEOUT` and `FM_FLEET_PRUNE` (lines 550-551), and from the fm-bootstrap.sh header&#39;s own env documentation, which covers FM_BOOTSTRAP_DETECT_ONLY, FM_BOOTSTRAP_VERBOSE_FACTS, and FM_FLEET_PRUNE. Operators have no documented way to discover the 10s default, the override, or the diagnostics opt-in.
- ℹ️ `bin/fm-bootstrap.sh:220` - The `timeout`/`gtimeout` branches send only SIGTERM and then wait for the child indefinitely, while the Perl fallback deliberately escalates TERM -&gt; 0.2s -&gt; KILL. A `gh` (or credential helper) that blocks or ignores SIGTERM keeps the primary path hanging, so the strongest guarantee lives only in the least-used branch. Adding `-k &lt;n&gt;` (e.g. `timeout -k 2 &#34;$FM_GH_AUTH_TIMEOUT_SECS&#34; ...`) makes the primary path match the fallback&#39;s escalation.
- ℹ️ `bin/fm-bootstrap.sh:255` - When none of timeout/gtimeout/perl exist, gh_auth_probe returns 124 without ever invoking gh, yet gh_auth_report prints &#34;probe timed out after 10s&#34;. Nothing ran and nothing timed out; the operator is told a bounded probe expired when in fact the auth check was skipped for lack of a bounding mechanism. Returning 124 there is required by the intent, but the two causes warrant distinguishable wording so the remediation (install coreutils/perl) is discoverable.
- ℹ️ `bin/fm-bootstrap.sh:218` - The timeout and gtimeout branches are byte-identical apart from the binary name, and gh_auth_report repeats the same `[ &#34;${FM_BOOTSTRAP_GH_AUTH_DIAGNOSTICS:-0}&#34; != 1 ] || echo ...` gate twice. Resolving the bounding binary into a variable (`for b in timeout gtimeout; do command -v &#34;$b&#34; ... done`) and factoring the diagnostics gate into a one-line helper removes the duplication without changing behavior; the Perl branch stays separate since its shape genuinely differs.
- ℹ️ `tests/fm-bootstrap.test.sh:104` - The new tests pin PATH to `$fakebin:$BASE_PATH` (/usr/bin:/bin:/usr/sbin:/sbin) but never control which bounding tool is visible, so each host exercises exactly one branch: Linux CI hits `timeout` (/usr/bin/timeout), stock macOS hits the Perl fallback (/usr/bin/perl). The Perl process-group path and the no-tool `return 124` path — the two portability guarantees the change is built around, and where the signal-status bug above lives — are never both covered on any single run. A fakebin shim that shadows/hides `timeout`, `gtimeout`, and `perl` per case would force all three branches deterministically.

🔧 Fix: harden bounded gh auth probe and register its diagnostics
- ⚠️ `bin/fm-bootstrap.sh:256` - The new `-k &#34;$GH_AUTH_KILL_GRACE_SECS&#34;` makes the escalated-kill case report NEEDS_GH_AUTH instead of indeterminate. GNU timeout documents and returns 137 (128+9), not 124, when it has to send SIGKILL after the grace period; verified on coreutils 9.11 that `timeout -k 1 1 bash -c &#39;trap &#34;&#34; TERM; sleep 10&#39;` exits 137 while `timeout -k 1 1 sleep 10` exits 124. 137 misses the `124)` arm at line 299 and falls through `*)` at line 309 to `echo &#34;NEEDS_GH_AUTH&#34;`, so a `gh` or credential helper that ignores SIGTERM - precisely the case `-k` was added to handle - now produces a confident &#34;not authenticated&#34; report from a probe that never finished. That contradicts the intent criterion &#34;stop it reporting an indeterminate probe as unauthenticated&#34;. Fix inside the bounder branch so the Perl and no-tool paths keep shell semantics, e.g. run the bounder, capture its status, and `return 124` for 137 (and 143, which non-GNU `timeout` implementations return on expiry) while leaving other non-zero statuses alone.
- ⚠️ `tests/fm-bootstrap.test.sh:112` - `set_fake_bounding_timeout` records argv and then `exec &#34;$@&#34;` with no bound at all, so the primary branch - the one every coreutils-equipped host and Linux CI actually takes - is only asserted on the argv string (`-k 2 10 gh auth status`). Nothing verifies what `gh_auth_report` does with the bounder&#39;s exit status, which is why the 137-on-SIGKILL misreport above is invisible to the suite even though the same test file now covers the Perl branch&#39;s timeout and signal cases behaviorally. Extend the shim to exit a caller-selected status (124 for a plain expiry, 137 for an escalated kill) and assert both map to `GH_AUTH: indeterminate (probe timed out after 10s)`.
- ℹ️ `bin/fm-bootstrap.sh:258` - `gh_auth_probe` hand-rolls a fourth copy of a timeout/gtimeout/Perl-process-group-alarm chain that already exists in the repo. `bin/fm-bearings-snapshot.sh:194-199` is a near-twin (it bounds a `gh` call with the identical `GH_PROMPT_DISABLED=1 GH_NO_UPDATE_NOTIFIER=1` prefix and the byte-identical Perl one-liner), and `bin/fm-fleet-snapshot.sh:791-796` and `bin/fm-nm-run-lib.sh:17-24` carry the same chain. All three still end in `exit($? &gt;&gt; 8)`, the exact signal-status bug this change just fixed in its own copy, and none passes `-k`, so the four implementations now disagree about both hardening measures. Extracting the hardened version into a shared `bin/fm-*-lib.sh` helper (`fm_bounded_run &lt;secs&gt; &lt;cmd&gt;...`) would give one implementation and propagate the fix; touching the three other scripts expands this change&#39;s scope, so it is the author&#39;s call.
- ℹ️ `bin/fm-bootstrap.sh:237` - With `GH_AUTH_KILL_GRACE_SECS=2` the worst-case wall clock of the bounder branch is `FM_GH_AUTH_TIMEOUT_SECS + 2` (12s by default), but the header at line 88, docs/configuration.md:505, and .agents/skills/bootstrap-diagnostics/SKILL.md:31 all describe a flat 10 second bound, and the printed line still says &#34;timed out after 10s&#34;. The extra 2s only elapses when the probe ignores SIGTERM, so this is an accuracy nuance rather than a behavior problem, and the existing `elapsed -lt 5` assertions against a 1s bound still hold (worst case 3s).

🔧 Fix: report escalated-kill bounder status as indeterminate
1 warning still open:
- ⚠️ `bin/fm-bootstrap.sh:266` - The new `[ &#34;$status&#34; -lt 128 ] || status=124` applies only to the bounder branch, so the two bounding implementations now report a signal-killed probe differently. A `gh auth status` killed by something other than the bound (SIGSEGV, the OOM killer, an external kill) returns 139/137 from `timeout`, is collapsed to 124 at line 266, and prints `GH_AUTH: indeterminate`; the same condition on the Perl fallback is mapped to 128+N at line 287, misses the `124)` arm at line 309, and prints `NEEDS_GH_AUTH`. The suite pins both semantics - tests/fm-bootstrap.test.sh:182 asserts indeterminate for forced 137/143, and tests/fm-bootstrap.test.sh:210 asserts NEEDS_GH_AUTH for the Perl branch&#39;s SEGV case - so a home on stock macOS and a home on Linux give opposite answers for an identical unanswered probe, under an intent that names portability a hard requirement. The bounder branch&#39;s own comment states the principle (&#34;Every 128+N outcome means the probe was killed rather than answered, so it is indeterminate and never a confident unauthenticated&#34;), which the Perl branch contradicts; that branch also still reports an unanswered probe as unauthenticated, the behavior the intent&#39;s headline goal forbids. Note the reverse tradeoff too: on the bounder path a probe that segfaults in 0.1s now prints &#34;probe timed out after 10s&#34;, which is inaccurate wording for a non-timeout kill. Both mappings were separately requested in earlier rounds, so which semantic should win is the author&#39;s call - either extend the collapse to the Perl branch (map 128+N to 124 there too, ideally with cause wording that does not claim a timeout), or narrow the bounder collapse to 124/137/143 only.

🔧 Fix: unify killed-probe reporting across both bounding branches
1 info still open:
- ℹ️ `bin/fm-bootstrap.sh:301` - One residual sub-wording difference survives the unification, by design. A probe that ignores SIGTERM and must be SIGKILLed by its own bound reports `GH_AUTH: indeterminate (probe terminated abnormally with status 137)` on the timeout/gtimeout path, because GNU timeout&#39;s 137 cannot be attributed to the kill grace versus an external kill, but reports `GH_AUTH: indeterminate (probe timed out after 10s)` on the Perl fallback, because that branch&#39;s alarm handler exits 124 at line 279 before waitpid observes the signal and so knows it caused the kill. The classification is identical on both paths, .agents/skills/bootstrap-diagnostics/SKILL.md:34 states the ambiguity explicitly, and the skill&#39;s handling for the two lines is the same, so this is a documented information asymmetry rather than a defect; noting it only so the residual gap against &#34;report identically on every machine&#34; is on record.

</details>

<details>

✅ No issues found.
- `a`

</details>

<details>
<summary>⚠️ **Document** - 2 warnings</summary>

- ⚠️ `bin/fm-bootstrap.sh:310` - The final commit&#39;s stated goal was to &#34;unify killed-probe reporting across both bounding branches&#34;, but the bound&#39;s own escalation is still worded differently depending on which bounding tool the home ships. I measured this: against a SIGTERM-ignoring probe, `timeout -k 2 1` exits 137, which the shared mapping classifies as `abnormal` and reports as &#34;probe terminated abnormally with status 137&#34;; the Perl fallback&#39;s ALRM handler calls `exit 124` itself (line 283-287), so the identical hang reports &#34;probe timed out after 10s&#34;. Only externally-killed probes (the SIGSEGV/139 case the test pins) actually read identically. I documented the divergence rather than hid it, because the safety invariant still holds everywhere -- both wordings are indeterminate and neither becomes NEEDS_GH_AUTH -- and the SKILL.md tells the agent to handle the two lines the same way. But if the author intended literal unification, the fix belongs in the Perl handler (exit a distinguishable status when it had to escalate, or have the handler report the same abnormal cause), not in the prose I just aligned to current behavior. That is an executable-behavior change and therefore outside this task&#39;s scope, so it needs the author&#39;s call.
- ℹ️ `docs/configuration.md:303` - The gh-auth probe mechanism (10s bound, sanitization rules, SIGKILL grace, three indeterminate causes, opt-in diagnostics) is now stated in four places: the `bin/fm-bootstrap.sh` header, the `docs/configuration.md` Toolchain prose, the `docs/configuration.md` env-var block, and `.agents/skills/bootstrap-diagnostics/SKILL.md`. This is exactly the synchronization cost the one-owner rule warns about -- the single wrong &#34;2 second grace&#34; figure had to be chased through all four. I reduced the SKILL.md copy to a pointer as part of this fix, but collapsing the remaining overlap (notably the env-block comment restating the prose two paragraphs above it) is a consolidation beyond what this change made stale. Proposing it as a follow-up rather than multiplying edits here.

🔧 Fix: unify killed-probe wording across both gh auth bounding branches
- ⚠️ `tests/fm-bootstrap.test.sh:208` - The behavior I just changed has no test pinning it. The Perl fallback now exits 128+SIGKILL when it has to escalate, instead of always exiting 124, and that is what makes a wedged probe read identically on both branches. But the existing Perl-branch hang case uses a fake gh of `sleep 300`, which honours SIGTERM, so it only ever exercises the 124 path; the 137 escalation path is exercised on the `timeout` branch solely through `set_fake_bounding_timeout`&#39;s forced status, which fakes the bounder rather than running one. I verified the parity by hand against the real script (a SIGTERM-ignoring gh reports `probe was killed before it answered, status 137` on both branches, and a cooperative hang reports `probe timed out after 1s` on both), but a manual check does not survive the next refactor: someone could revert the Perl handler to a bare `exit 124` and the whole suite would still pass. Closing this is cheap given the existing fixtures - add a `wedged` mode to `set_fake_gh_auth_status` (`trap &#39;&#39; TERM; sleep 300`) and assert the Perl branch emits the same line the bounder branch does, at a cost of about bound+2s of runtime. I did not write it because this step&#39;s rules forbid changing tests, and the author&#39;s instruction authorized the message unification, not new test authoring. It needs their call on whether it lands in this PR or as a follow-up.
- ℹ️ `bin/fm-bootstrap.sh:243` - Unifying the message required unifying the grace, which widens the Perl fallback&#39;s worst case. It previously waited 0.2s between SIGTERM and SIGKILL; it now takes the shared GH_AUTH_KILL_GRACE_SECS of 2s, so worst-case wall clock on stock macOS goes from bound+0.2s to bound+2s (measured 3.2s with a 1s bound). This was not optional: leaving the graces different would have meant a probe that takes, say, 0.5s to honour SIGTERM reported as killed on the Perl branch and as a timeout on the timeout branch, which is exactly the tool-dependent message the author asked to remove. The bound still cannot hang, the 2s figure now matches what the docs claim on every path, and the suite&#39;s `elapsed &lt; 5` assertion still holds. Flagging it only because it is a timing change on the stock-macOS path that the wording-focused instruction did not explicitly ask for, and the contributor&#39;s intent statement called portability a hard requirement - so if bound+2s is considered too slow for session start, the lever is GH_AUTH_KILL_GRACE_SECS and it should be lowered for both branches together, never for one.

🔧 Fix: drop stale SIGKILL clause from gh auth timeout bullet
2 warnings still open:
- ⚠️ `tests/fm-bootstrap.test.sh:208` - The regression test you approved last round is still not in the tree, and this step is forbidden from editing tests, so I could not add it. The Perl-branch hang case still uses a `sleep 300` fake gh, which honours SIGTERM and therefore only ever exercises the 124 path; the ALRM handler&#39;s escalation to 128+SIGKILL is exercised nowhere. The `signal` fixture at line 222 pins a self-crashing probe (139) rather than a bound that had to escalate, and the 137 case on the `timeout` branch comes from `set_fake_bounding_timeout`&#39;s forced status, which fakes the bounder instead of running one. I re-confirmed the parity by hand against the real bounders this round (bound 1s, grace 2s: a SIGTERM-ignoring probe exits 137 on both branches, a slow-but-cooperative one exits 124 on both), but that is exactly the kind of check a refactor silently voids: reverting the Perl handler to a bare `exit 124` would still leave the whole suite green. The fix is unchanged from what you approved - a `wedged` mode in `set_fake_gh_auth_status` (`trap &#39;&#39; TERM; sleep 300`) plus an assertion that the Perl branch emits the same killed line the bounder branch does, costing about bound+2s of runtime. It needs a step that is allowed to write tests.
- ⚠️ The two behavior-change statements you asked to be stated plainly have no live PR body to land in, and I could not resolve that from the document step. PR kunchenguid/firstmate#1752 is CLOSED and its head (7a4fdc3) predates every commit on this branch, so its body still describes only the original bounded-probe commit - it does not mention the killed or no-bounding-tool causes, and it claims 23 passing tests where the suite is now 27. Editing a closed PR that points at different code would not carry the statements to whatever PR this branch opens, and the body is not a repository file I can fix in place, so the PR step has to author them. The two lines it must include, per your instruction: signal-terminated probes now report `GH_AUTH: indeterminate` where they previously fell through to `NEEDS_GH_AUTH`, because a killed probe proves nothing about the credential and reporting it as unauthenticated is the exact failure this change exists to remove; and unifying the message required unifying the kill grace, so the Perl fallback&#39;s worst case widens from bound plus 0.2 seconds to bound plus 2 seconds on stock macOS, because a per-branch grace would decide a slow-but-cooperative probe&#39;s wording purely from which bounder the home ships. Current-state versions of both facts already live in their owner docs (`docs/configuration.md` &#34;Toolchain&#34; and the `bin/fm-bootstrap.sh` header); only the before-and-after framing belongs in the PR narrative.

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

### DIFF
--- a/.agents/skills/bootstrap-diagnostics/SKILL.md
+++ b/.agents/skills/bootstrap-diagnostics/SKILL.md
@@ -29,7 +29,7 @@ When any diagnostic needs captain attention, report the plain consequence and re
   Treat GitHub auth as unproven rather than broken: do not dispatch work that needs it, and do not ask the captain to re-authenticate on this evidence alone.
   Ask the captain to run `! gh auth status` themselves to settle the state, since a wedged credential helper, an unreachable network, or a blocking keychain prompt are the usual causes.
   `FM_GH_AUTH_TIMEOUT_SECS` raises the 10 second bound when the home is merely slow rather than stuck.
-  The reported seconds are that bound, not the elapsed wall clock: a probe that ignores SIGTERM is escalated to SIGKILL after a further 2 second grace, and that escalated kill still reports indeterminate rather than unauthenticated.
+  The reported seconds are that bound, not the elapsed wall clock, so a probe held past it by a SIGKILL escalation still reports indeterminate rather than unauthenticated; `docs/configuration.md` "Toolchain" owns the escalation grace and worst-case timing.
 - `GH_AUTH: indeterminate (probe terminated abnormally with status <n>)` - the preflight was killed before it answered, so the credential state is unknown for the same reason the timeout case is.
   The status is reported raw because it cannot say whether the kill grace escalated to SIGKILL, the OOM killer fired, or something external killed the probe.
   Handle it exactly like the timeout case above, and treat a repeat across sessions as a sign the home itself is unhealthy rather than a GitHub auth problem.
--- a/.agents/skills/bootstrap-diagnostics/SKILL.md
+++ b/.agents/skills/bootstrap-diagnostics/SKILL.md
@@ -29,6 +29,7 @@ When any diagnostic needs captain attention, report the plain consequence and re
   Treat GitHub auth as unproven rather than broken: do not dispatch work that needs it, and do not ask the captain to re-authenticate on this evidence alone.
   Ask the captain to run `! gh auth status` themselves to settle the state, since a wedged credential helper, an unreachable network, or a blocking keychain prompt are the usual causes.
   `FM_GH_AUTH_TIMEOUT_SECS` raises the 10 second bound when the home is merely slow rather than stuck.
+  The reported seconds are that bound, not the elapsed wall clock: a probe that ignores SIGTERM is escalated to SIGKILL after a further 2 second grace, and that escalated kill still reports indeterminate rather than unauthenticated.
 - `GH_AUTH: indeterminate (no bounded-probe tool: install coreutils or perl)` - the home has none of `timeout`, `gtimeout`, or `perl`, so bootstrap skipped the auth check entirely rather than risk hanging session start on it.
   Report the missing bounding tool and the printed coreutils-or-perl remediation, then rerun session start once one is installed.
   The credential state stays unknown until then, so treat it exactly like the timeout case above and do not dispatch work that needs GitHub auth.
--- a/.agents/skills/bootstrap-diagnostics/SKILL.md
+++ b/.agents/skills/bootstrap-diagnostics/SKILL.md
@@ -29,7 +29,8 @@ When any diagnostic needs captain attention, report the plain consequence and re
   Treat GitHub auth as unproven rather than broken: do not dispatch work that needs it, and do not ask the captain to re-authenticate on this evidence alone.
   Ask the captain to run `! gh auth status` themselves to settle the state, since a wedged credential helper, an unreachable network, or a blocking keychain prompt are the usual causes.
   `FM_GH_AUTH_TIMEOUT_SECS` raises the 10 second bound when the home is merely slow rather than stuck.
-  The reported seconds are that bound, not the elapsed wall clock, so a probe held past it by a SIGKILL escalation still reports indeterminate rather than unauthenticated; `docs/configuration.md` "Toolchain" owns the escalation grace and worst-case timing.
+  The reported seconds are that bound, not the elapsed wall clock: a probe that honours the bound's SIGTERM slowly can die up to the kill grace later and still report this line, while one that never honours it reports the killed line below instead.
+  `docs/configuration.md` "Toolchain" owns that grace and the worst-case timing.
 - `GH_AUTH: indeterminate (probe was killed before it answered, status <n>)` - the preflight was killed before it answered, so the credential state is unknown for the same reason the timeout case is.
   The line names no culprit on purpose: the status alone cannot say whether the kill grace escalated to SIGKILL, the OOM killer fired, or something external killed the probe, so do not report a cause the captain would have to unlearn.
   Handle it exactly like the timeout case above, and treat a repeat across sessions as a sign the home itself is unhealthy rather than a GitHub auth problem.
--- a/.agents/skills/bootstrap-diagnostics/SKILL.md
+++ b/.agents/skills/bootstrap-diagnostics/SKILL.md
@@ -30,8 +30,8 @@ When any diagnostic needs captain attention, report the plain consequence and re
   Ask the captain to run `! gh auth status` themselves to settle the state, since a wedged credential helper, an unreachable network, or a blocking keychain prompt are the usual causes.
   `FM_GH_AUTH_TIMEOUT_SECS` raises the 10 second bound when the home is merely slow rather than stuck.
   The reported seconds are that bound, not the elapsed wall clock, so a probe held past it by a SIGKILL escalation still reports indeterminate rather than unauthenticated; `docs/configuration.md` "Toolchain" owns the escalation grace and worst-case timing.
-- `GH_AUTH: indeterminate (probe terminated abnormally with status <n>)` - the preflight was killed before it answered, so the credential state is unknown for the same reason the timeout case is.
-  The status is reported raw because it cannot say whether the kill grace escalated to SIGKILL, the OOM killer fired, or something external killed the probe.
+- `GH_AUTH: indeterminate (probe was killed before it answered, status <n>)` - the preflight was killed before it answered, so the credential state is unknown for the same reason the timeout case is.
+  The line names no culprit on purpose: the status alone cannot say whether the kill grace escalated to SIGKILL, the OOM killer fired, or something external killed the probe, so do not report a cause the captain would have to unlearn.
   Handle it exactly like the timeout case above, and treat a repeat across sessions as a sign the home itself is unhealthy rather than a GitHub auth problem.
 - `GH_AUTH: indeterminate (no bounded-probe tool: install coreutils or perl)` - the home has none of `timeout`, `gtimeout`, or `perl`, so bootstrap skipped the auth check entirely rather than risk hanging session start on it.
   Report the missing bounding tool and the printed coreutils-or-perl remediation, then rerun session start once one is installed.
--- a/.agents/skills/bootstrap-diagnostics/SKILL.md
+++ b/.agents/skills/bootstrap-diagnostics/SKILL.md
@@ -2,7 +2,7 @@
 name: bootstrap-diagnostics
 description: >-
   Agent-only handling playbook for session-start bootstrap diagnostics.
-  Use whenever the session-start digest's bootstrap section prints an actionable diagnostic line - MISSING, MISSING_MANUAL, BACKEND_INVALID, NEEDS_GH_AUTH, TANGLE, STARTUP_MEMORY_BUDGET, CREW_DISPATCH invalid, FLEET_SYNC, PR_CHECK_MIGRATION, SECONDMATE_SYNC, SECONDMATE_LIVENESS, SECONDMATE_HANDOFF, NUDGE_SECONDMATES, or FMX - or when a standalone bin/fm-bootstrap.sh run prints one of those lines.
+  Use whenever the session-start digest's bootstrap section prints an actionable diagnostic line - MISSING, MISSING_MANUAL, BACKEND_INVALID, NEEDS_GH_AUTH, GH_AUTH, TANGLE, STARTUP_MEMORY_BUDGET, CREW_DISPATCH invalid, FLEET_SYNC, PR_CHECK_MIGRATION, SECONDMATE_SYNC, SECONDMATE_LIVENESS, SECONDMATE_HANDOFF, NUDGE_SECONDMATES, or FMX - or when a standalone bin/fm-bootstrap.sh run prints one of those lines.
   A silent bootstrap section, or a BOOTSTRAP_INFO fact, means no skill load.
 user-invocable: false
 metadata:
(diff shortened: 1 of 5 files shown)


kunchenguid:

<sub>Noted for firstmate#1775 at `f30fda9a`.</sub>

kunchenguid:

PR #1791, opened 2026-08-06
fix(teardown): gate the herdr presentation lock on the projection being on

bingb0t5:
`fm-spawn.sh` already gates its projection work on `fm_backend_herdr_presentation_enabled`, but teardown reached for the presentation session lock **unconditionally**, at two sites: the endpoint preflight (`teardown_herdr_preflight_target`) and the plain pane close.

So a home that wrote `off` to `config/herdr-presentation-spaces` still had its cleanup depend on a feature it had switched off. That is an inconsistency on any platform - spawn honours the opt-out, teardown ignored it.

Where the lock namespace cannot satisfy its own validity check the consequence is worse than cosmetic. Teardown refuses, the task's state files are retained by design, supervision stays permanently "needed", and there is no manual recovery. `--force` does not help: it skips only the dirty and landed-work checks, and the same preflight runs in both branches.

Both sites now consult the same predicate. **Default behaviour is unchanged** - the projection is on unless a home opts out, so a contended lock still refuses exactly as before.

The focus-preserving close itself is kept in the opted-out path; it simply no longer demands a lock whose only purpose is ordering a projection this home does not have.

## Relationship to #1790

#1790 fixes the underlying reason the namespace can be unobtainable on Windows, and is the more important of the two. This one is independent: it is a real inconsistency in its own right, and it is what lets an opted-out home tear down cleanly regardless of why the lock is unavailable. Neither depends on the other, and they touch disjoint files.

## Testing

New `test_herdr_flat_teardown_opted_out_of_presentation_ignores_the_session_lock` makes the lock unavailable in both ways it can be unavailable - held by a contending process, or unresolvable because the namespace cannot be validated - and asserts teardown still completes, closes the pane, returns the isolated copy, and clears its records. The existing `test_herdr_flat_teardown_refuses_orphaning_records_then_retry_completes` is the negative control: with the projection at its default, a contended lock must still refuse.

Verified end to end against a real stuck task on the affected host: a finished task that had been uncollectable for a day tore down cleanly, state files cleared, worktree returned to the pool, and the delivered work intact on its branch.

`tests/fm-teardown.test.sh` reaches the same point on this branch as on unmodified `main` on that host (9 ok, then a pre-existing failure caused by the namespace fault that #1790 addresses), so this introduces no regression there.

ShellCheck is not installable on that host, so `bin/fm-lint.sh` could not be run for CI parity.

https://claude.ai/code/session_01F7L1m56txHs7hrNcTeq7AP

### DIFF
--- a/bin/fm-teardown.sh
+++ b/bin/fm-teardown.sh
@@ -1893,6 +1893,7 @@ teardown_herdr_require_prerequisites() {  # <task-id>
     fm_backend_herdr_workspace_presence_state \
     fm_backend_herdr_endpoint_confirmed_gone \
     fm_backend_herdr_explicit_close_pane_confirmed \
+    fm_backend_herdr_presentation_enabled \
     fm_backend_herdr_presentation_session_lock_path; do
     if ! declare -F "$prerequisite" >/dev/null 2>&1; then
       echo "error: herdr teardown prerequisites are unavailable for $task_id; nothing was changed - restore the adapter and rerun teardown" >&2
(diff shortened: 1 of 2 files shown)


kunchenguid:

PR #1816, opened 2026-08-06
fix(bin): prevent failed misread when newer same-branch run owns the pipeline head

spirispark:
## Intent

Root-cause and fix a dangerous state misread in bin/fm-crew-state.sh: when bare `no-mistakes axi status` returned an older same-branch failed run while a newer pipeline run on the same branch had advanced its head and was parked at `fix_review`, the helper reported `state: failed - source: run-step - run failed` for a healthy parked worker. A failed misread can drive the watcher to relaunch a healthy parked worker, so this must be fixed root-cause not patched at the symptom. Root cause: bare `axi status` branch+head attribution alone cannot distinguish an older same-code terminal run from a newer active run on the same branch. Fix: when the bare answer is a terminal run on this branch, consult the `axi` home surface (active_run block first, runs[N]{id,branch,status,head,pr} table second) to recover the exact current-branch run ID, then re-query `axi status --run <id>`, verify the ID/branch/code head, and only treat the bare answer as authoritative if no exact-ID run is found. The coarse `no-mistakes runs` fallback also now prefers an active row over a same-code terminal row, so a branch without the home surface still gets the right answer. Coverage: test_newer_parked_run_beats_older_failed_run (fails on pre-fix code, passes post-fix) and test_single_parked_run_counterfactual (smallest counterfactual: single current run still reports parked). Diff scope: bin/fm-crew-state.sh helper set + same-branch same-code precedence tweak; tests/fm-crew-state.test.sh fake axi home surface and the two new tests. No public contracts change; fm-lint clean; existing 48 tests unchanged.

## What Changed
- When bare `axi status` returns a terminal same-branch run, re-resolve through the `axi` home surface (active_run block first, then the `runs[N]{id,branch,status,head,pr}` table) to recover the exact current-branch run ID, then re-query `axi status --run <id>` and verify ID/branch/code head match before treating the bare answer as authoritative.
- The coarse `no-mistakes runs` fallback now prefers an active row over a same-code terminal row, so a branch without the home surface still gets the right answer.
- Added regression test `test_newer_parked_run_beats_older_failed_run` (fails pre-fix), smallest counterfactual `test_single_parked_run_counterfactual` (single current run still reports parked), a fake `axi` home surface fixture, and a documented `FM_CREW_STATE_DEBUG` toggle for run-attribution diagnosis.

## Pipeline

<details>
</details>
<details>
</details>
<details>
<summary>✅ **Review** - completed</summary>
</details>
<details>
</details>
<details>
</details>
<details>
</details>
<details>
</details>

### DIFF
--- a/bin/fm-crew-state.sh
+++ b/bin/fm-crew-state.sh
@@ -21,8 +21,10 @@
 # Logic, in order:
 #   1. Resolve worktree + backend target + kind from state/<id>.meta.
 #   2. Matching no-mistakes run for this crew's branch AND current code identity,
-#      active or terminal (from `axi status`, or the coarse `no-mistakes runs`
-#      fallback)? Branch name alone is not enough: a historical run on a reused
+#      active or terminal (from `axi status`, re-resolved through the `axi` home
+#      surface when the bare answer is a stale same-branch terminal run, or
+#      falling back to the coarse `no-mistakes runs` list when no exact ID is
+#      exposed)? Branch name alone is not enough: a historical run on a reused
 #      branch whose head was rewritten or diverged must not be attributed.
 #      A run matches when its head equals the worktree HEAD, or the worktree HEAD
 #      is an ancestor of the run head (pipeline fix commits advanced the run on
--- a/bin/fm-crew-state.sh
+++ b/bin/fm-crew-state.sh
@@ -50,6 +50,7 @@
 # Read-only and side-effect free. Always exits 0 on a successful read regardless
 # of state; exit 2 only on a usage error (no id).
 set -u
+[ "${FM_CREW_STATE_DEBUG:-0}" = 1 ] && set -x
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 FM_ROOT="${FM_ROOT_OVERRIDE:-$(cd "$SCRIPT_DIR/.." && pwd)}"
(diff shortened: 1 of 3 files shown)


spirispark (commented):
## Request changes — failed-misread fix can still miss the original bug

When a same-branch `no-mistakes axi status` hit is terminal, the helper now consults the axi home surface (`active_run`, then the runs table) and re-queries `axi status --run <id>` so an older failed run cannot be treated as current. The coarse `no-mistakes runs` fallback prefers `pending|running` over a terminal row.

The incident model is right: local HEAD equals the old run head and is an ancestor of a newer pipeline head. Failed refresh restores `RUN_OUT`. Exact-ID status is rejected unless id/branch/head match. CI is green and the PR is mergeable.

### Must-fix

1. **`nm_axi_run_id_for_branch` returns as soon as `active_run` matches branch+head, with no status check.** If home `active_run` is the same stale failed run that bare `axi status` returned, the table's live row is never consulted. If `active_run` is terminal, fall through to the table.

2. **Home-table only treats `pending|running` as active.** A parked `fix_review` / `awaiting_approval` / `parked` row is skipped and the failed `terminal_id` wins — same misread. Treat any non-terminal as active, or at least the gate statuses this helper already maps to parked.

3. **Coarse path: `pending` now beats `failed` but then maps to `unknown` instead of working.** Map `pending` like `running`.

### Test gaps

The regression table has the running row first (first-match would pass). There is no `active_run` fixture, and the fake `axi status --run` ignores the ID. Put the failed row first, add an `active_run`-stale case, and dispatch `--run` by id.

PR #1817, opened 2026-08-06
session-start: queue a due sprint-check wake before the first watcher poll

vladimyr1991:
## Summary
- Session start now asks fm-sprint-poll.sh whether a configured sprint-board scan is due, and when locked, queues the same check wake the watcher would have queued, so a home whose poll interval already elapsed doesn't sit idle until the next watcher cycle.
- fm-supervision-lib.sh now recognizes state/sprint-watch.check.sh as an in-flight supervision need, alongside the existing X-mode relay-poll and process-event-source checks.
- Adds two tests: a configured session start queues one due sprint-check wake before draining (and does not fire again inside the interval), and a read-only (lock-refused) session start leaves sprint-poll state untouched.

## Test plan
- [x] `bash tests/fm-session-start.test.sh` - all pass, including the two new tests
- [x] `bash tests/fm-turnend-guard.test.sh` - pass
- [x] `bash tests/fm-backend.test.sh` - pass
- [x] `bash tests/fm-procevent.test.sh` - pass
- [x] `bash tests/fm-session-lock-ancestry.test.sh` - pass
- [x] `bash tests/fm-claude-stop-autoarm.test.sh` - pass


vladimyr1991:
Opened in error against the wrong base repository - closing. The change will be resubmitted as a PR against vladimyr1991/firstmate instead.

PR #1828, opened 2026-08-06
fix(bin): refresh away-mode wedge markers for provably-working crews

sbracewell64:
## Intent

docs(daemon): record the shipped wedge-aging design and its accepted cost

This commit carries no code change. The tree already describes the shipped
design correctly; what was stale was the recorded narrative, which fed the
commit body and the pull request description. It is corrected here as a new
commit rather than by amending, because amending replaced a head the validation
pipeline owned and cost a run.

What actually ships

Away-mode wedge aging decided purely from pane text: a pane whose rendered
content stopped changing for STALE_ESCALATE_SECS escalated as a possible wedge,
and did so again every threshold for as long as it stayed static. It never asked
whether the crew was provably working, though crew_absorb_class in
bin/fm-classify-lib.sh already answers that from the authoritative current state.
A worker waiting on a completion artifact rather than printing holds a static
pane while demonstrably alive, so it alarmed indefinitely.

The shipped fix is a single gate in housekeeping's stale persistence recheck, at
the exact point that emits the alarm. When an aged marker is about to escalate,
crew_absorb_class is read once. A positive working verdict REFRESHES the marker
timestamp in place; anything else escalates and removes the marker as before.
classify_stale and handle_wake are unchanged from base apart from comments: the
gate is deliberately not on the per-wake path, because crew_absorb_class reuses
bin/fm-crew-state.sh and may make a bounded no-mistakes call, and a static pane
can re-enter the wake path arbitrarily often.

Refreshing rather than dropping is the load-bearing choice. It makes the
suppression fail toward noise: silence must be re-earned every threshold by a
fresh positive verdict, so if the crew freezes, or the state reader stops
answering, the marker ages out and escalates with nothing needing to notice and
re-arm.

The accepted cost

Detection after a freeze now takes at most two thresholds rather than one. The
daemon's guarantee is preserved - a wedge is found within a bounded time and
never lost - and the constant doubles. This was weighed and accepted against
alternatives that were either permanently silent or a hot loop.

Relatedly, the age in the escalation line now measures time since positive
evidence was last seen rather than time since the pane first went idle. That is
the more useful number for triage under these semantics, and preserving the
original epoch would need a second marker, which the design forbids.

Designs withdrawn along the way, so they are not re-proposed

Round one dropped the task from wedge aging outright on a working verdict, via a
new working action in classify_stale and handle_wake. Withdrawn: the drop never
re-armed. In away mode the watcher enqueues at most one stale wake per distinct
pane hash, so once the marker was removed, a crew that was working when checked
and froze afterwards on an unchanged pane went from one escalation to zero,
permanently. That converts a noisy but safe detector into a silent-failure one.

Round two re-armed by clearing the watcher's per-hash stale marker through
clear_pause_tracking. Withdrawn: it did not fire once, it oscillated. Clearing
that marker makes the watcher re-enqueue the same unchanged hash immediately on
restart, costing a watcher respawn, a crew-state read and a durable wake-queue
record roughly every second for the worker's entire run - a hot-path regression
against a contract this change itself introduced.

Verification carried forward unchanged

Five negative controls, each witnessed red rather than asserted:

1. A pane transitioning from working to frozen escalates within two thresholds.
2. A genuinely hung pane that was never working still escalates on the unchanged
schedule; that control leg runs first in its test, so the escalation path is
witnessed firing before the suppression is trusted.
3. A continuously-working pane does not escalate, including across two
consecutive thresholds, so the later escalation cannot be an artefact of the
marker merely ageing twice. This is what distinguishes a fix from a revert.
4. The fail-safe: once the working verdict stops being re-earned, the pane
escalates rather than staying silent, covered both for a crew that freezes and
for a state reader that stops answering.
5. The hot-path check, measured rather than reasoned about: the gate costs one
crew-state read per threshold rather than one per tick, counted across several
ticks, and the working path leaves every watcher per-hash marker alone.

bin/fm-lint.sh, tests/fm-daemon.test.sh and tests/fm-watch-triage.test.sh pass.

Contribution base

The original report located the defect at two call sites in the running fleet
code that test crew_absorb_class for a settled verdict. That settled mechanism
does not exist at this branch's contribution base, where crew_absorb_class
returns only working, paused or none. The defect itself is fully present here, so
the fix is a single gate. There is consequently no parallel gate to create, and
when the settled work reaches this base the two conditions fold into the same
branch rather than becoming a second mechanism.

Repo conventions that apply: this is the firstmate repo's shared tracked material, so firstmate-coding-guidelines applies - one sentence per line in prose, plain dash rather than em dash, shellcheck-clean bin scripts, colocated tests, and no agent name as a commit co-author.

Deliberately out of scope, so it should not be reported as an omission: the watcher's enriched demand-deep-inspection wedge still overrides to escalate in handle_wake, which is a separate owner named in the original exclusion list. STALE_ESCALATE_SECS, the escalation cadence, the deep-inspection demand, the max-defer wedge alarm and presence gating are all untouched. crew_absorb_class returns the correct class for this worker shape, so there is no second finding to report.

The final commit on this branch is intentionally empty of code: it carries the corrected narrative as a new commit rather than an amend, because amending replaced a head the pipeline owned and failed a previous run. Do not squash, amend, or drop the three commits beneath it - they are the tracked pipeline history.

MERGE STATUS - this PR is deliberately BLOCKED and must not be merged yet

Review found that crew_absorb_class returns working FOREVER for a worker killed mid-turn: bin/fm-busy-lib.sh parses ts= but uses it only for format validation, never expiry, so a crash with no stop event leaves state=busy permanently, and the watcher BUSY_TURN_MAX_SECS backstop cannot help because it is gated on pane-busy while a dead pane is footer-idle. The stale-busy shape predates this branch, but the consequence does not: before this change such a crew still escalated, and after it the refresh gate suppresses it forever, converting a detectable crash into an undetectable one.

That is an upstream defect in the busy record, filed separately as cfvc-05-crew-state-json. This task's commission explicitly forbids compensating for a wrong crew_absorb_class class inside the daemon, so there is deliberately no cap on consecutive refreshes, no advancing-evidence check at the gate, and no expiry logic at the call site. This PR unblocks when the busy-record work lands. Do not treat the absence of that compensation as an oversight.

FURTHER WORK IN THIS BRANCH BEYOND THE ORIGINAL FIX

Per-pass cost bound: the gate is one crew-state read per crew per threshold, but several markers due in the same housekeeping pass would serialize bounded no-mistakes reads inline in the daemon's main loop and delay wake delivery. Reads are now budgeted per pass. A marker that does not get its read this pass is never suppressed - it stays aged and is reconsidered next pass - so deferring a read can never become a missed escalation. The deferral is ceil(D/B) - 1 passes for D due markers at budget B, stated as a formula rather than a single constant so it cannot go stale.

Escalation wording: the operator-facing line and the tree now record that the age measures time since positive evidence was last seen rather than time since the pane first went idle, and the superseded one-threshold detection bound is qualified wherever it was still stated as current.

Test-harness integrity: control 2, the safety control asserting a genuinely hung pane still escalates, was passing for the WRONG REASON twice - first because a test unset the library-owned FM_CREW_STATE_BIN, then because a replacement stub pointed at a fake reader the test never created. Both made the reader unresolvable, and crew_absorb_class absorbs that error and returns none, which is not working, so the escalation assertion passed through an error rather than through the verdict it named. Fixed by creating the fake in make_supercase, asserting via a call log that the reader actually answered, and adding a whole-file guard that aborts the run if crew_absorb_class is ever reached with an unusable FM_CREW_STATE_BIN, with an explicit opt-out for the test that exercises the unreadable-reader leg on purpose.

Control 2 has since been re-witnessed honestly: green when the reader answers stopped, red when it answers working (proving the escalation follows the verdict rather than an error), and red when the reader path is absent. tests/fm-daemon.test.sh 104 assertions and tests/fm-watch-triage.test.sh 46 assertions pass, bin/fm-lint.sh is clean, and the daemon suite emits no unbound-variable or command-not-found noise.

## What Changed

- `bin/fm-supervise-daemon.sh` housekeeping now consults `crew_absorb_class` at the one point an aged `.subsuper-stale` marker is about to escalate: a positive `working` verdict rewrites the marker timestamp in place instead of alarming, while stopped, parked, failed, and unreadable crews escalate on the unchanged schedule. Refreshing rather than dropping keeps silence conditional - it has to be re-earned every threshold - so detection after a freeze costs at most two thresholds instead of one, and the operator line now reads `no progress evidence for <n>s` because a refresh restarts that clock. `classify_stale` and `handle_wake` are unchanged apart from comments, so the per-wake path stays a pure status-file read.
- Added `FM_STALE_WORKING_GATE_READS` (default 3) to cap those crew-state reads per housekeeping pass, since housekeeping runs inline in the daemon's main loop. A marker past the budget is left aged and untouched for the next pass, so the budget delays an escalation and can never suppress one. `stale_gate_read_budget` assigns rather than prints (a `$( )` call site would discard its throttle state), restores the default on an invalid value, and logs that on first sight, on change, and otherwise hourly. Documented in `docs/configuration.md`, `docs/architecture.md`, and `.agents/skills/afk/SKILL.md`, with the read-budget capacity arithmetic owned by `docs/architecture.md`.
- Test harness hardening: `make_supercase` now creates the crew-state fake itself, the fake can log its invocations so read counts are counted rather than assumed, and a new `fm_guard_crew_state_reader` aborts the run if `crew_absorb_class` is ever reached with a nonexistent or the real reader - the two ways control 2 had been passing for the wrong reason. New `tests/fm-daemon.test.sh` cases cover the gate, the fail-safe once refreshes stop, the one-read-per-threshold bound, budget deferral without suppression, the sanitizer, and guard idempotence; `tests/fm-wake-daemon-lifecycle-e2e.test.sh` stubs the reader at its housekeeping call sites.

## Notes

- Merge is deliberately blocked pending the busy-record work filed as `cfvc-05-crew-state-json`: `crew_absorb_class` reports `working` indefinitely for a worker killed mid-turn (a `state=busy` record with no expiry), which this gate would convert from a detectable wedge into permanent silence. The commission forbids compensating inside the daemon, so there is intentionally no refresh cap, advancing-evidence check, or call-site expiry here.
- Known follow-up, deliberately not fixed in this branch: the inert-assertion class this branch hit twice still has an unfixed instance at `tests/fm-wake-daemon-lifecycle-e2e.test.sh` phase 3, whose assertions are satisfied by a capture failure returning 2 as well as by the busy verdict they name. Generalising the new crew-state-reader harness guard to cover that shape is the real fix.

## Risk Assessment

🚨 High: The round-5 error is properly fixed - the throttle state now travels through the daemon's real call path and a four-pass regression test would have caught the inert version - leaving only an info-level grep-count idiom in the new test, but the author-declared merge block on the upstream stale-busy-record fix is still standing, so merge needs an explicit human gate.

## Testing

Ran the two suites the change touches plus the lifecycle e2e (`fm-daemon.test.sh` 108 assertions, `fm-watch-triage.test.sh` 46, `fm-wake-daemon-lifecycle-e2e.test.sh`), all green and free of unbound-variable/command-not-found noise, then went past unit coverage and drove the real daemon housekeeping loop end-to-end against a real state dir whose working verdict comes from the real `bin/fm-crew-state.sh` reading real `bin/fm-busy-event.sh` records and a real git worktree HEAD. On one identical fixture the base daemon alarms a healthy validating crew at the first threshold while the target stays silent for three thresholds and escalates one threshold after evidence stops, a never-working crew still escalates on the unchanged schedule, the per-pass read budget defers without ever suppressing, and a misconfigured budget restores the default and is logged once; I also re-witnessed the safety control red under two independent mutations rather than trusting its green. No screenshot or rendered-UI artifact applies: this is a background daemon with no rendered surface, so its end-user surface is the escalation digest it types into the supervisor pane and the daemon log, both captured verbatim as CLI transcripts. No actionable issues found.

<details>
<summary>Evidence: Base vs target end-to-end wedge-aging transcript (real crew-state reader, real busy records)</summary>

<code>=== BASE 2cf0283 ===&#10;crew-state reads: state: working · source: run-step · validating (running)&#10;threshold 1 validation run still running ESCALATED -&gt; stale persisted 300s (possible wedge): fmdemo:fm-corpus-w1&#10;threshold 2 validation run still running silent (no marker left to age)&#10;threshold 3 validation run still running silent (no marker left to age)&#10;-- crew freezes; pane text still unchanged --&#10;threshold 4 no progress evidence silent (no marker left to age)&#10;supervisor pane: FIRSTMATE_OP: v1 away-supervisor: Supervisor escalate (1 event(s)): stale persisted 300s (possible wedge): fmdemo:fm-corpus-w1&#10;&#10;=== TARGET 6296cde (identical fixture) ===&#10;crew-state reads: state: working · source: run-step · validating (running)&#10;threshold 1 validation run still running silent (marker refreshed, still aging)&#10;threshold 2 validation run still running silent (marker refreshed, still aging)&#10;threshold 3 validation run still running silent (marker refreshed, still aging)&#10;-- crew freezes; pane text still unchanged --&#10;threshold 4 no progress evidence ESCALATED -&gt; no progress evidence for 300s (possible wedge): fmdemo:fm-corpus-w1&#10;supervisor pane: FIRSTMATE_OP: v1 away-supervisor: Supervisor escalate (1 event(s)): no progress evidence for 300s (possible wedge): fmdemo:fm-corpus-w1 (pre-read; re-arm not needed — watcher daemon-managed)&#10;daemon log:&#10;stale marker refreshed (provably working, wedge aging restarts): fmdemo:fm-corpus-w1 x3&#10;&#10;=== TARGET - CONTROL: never provably working ===&#10;crew-state reads: state: working · source: status-log · ...&#10;threshold 1 hung, never provably working ESCALATED -&gt; no progress evidence for 300s (possible wedge): fmdemo:fm-corpus-w1</code>

```text
==================================================================================
 BASE  2cf0283  - away-mode wedge aging decided purely from pane text
==================================================================================
daemon under test : /tmp/no-mistakes-evidence/01KZBX804SBBHVDV7C7EMY9WXF/base-bin/fm-supervise-daemon.sh
crew              : corpus-w1 on feat/corpus-rebuild, kind=ship, harness=claude
pane              : static since the handoff (never changes again)
busy record       : v1 gen=g1786038585.4122233.8426 seq=2 state=idle source=claude-hook event=stop ts=1786038585

away-mode wedge aging, one housekeeping pass per STALE_ESCALATE_SECS (240s):
  crew-state reads: state: working · source: run-step · validating (running)
  threshold 1  validation run still running           ESCALATED -> stale persisted 300s (possible wedge): fmdemo:fm-corpus-w1
  threshold 2  validation run still running           silent (no marker left to age)
  threshold 3  validation run still running           silent (no marker left to age)

  -- the crew now freezes: its validation run is gone (aborted / detached),
  -- and the pane text has still not changed by a single byte.
  crew-state reads: state: working · source: status-log · corpus rebuilt, handed to validation, waiting on the run
  threshold 4  no progress evidence                   silent (no marker left to age)

what the operator actually receives in the supervisor pane:
  ⁣FIRSTMATE_OP: v1 away-supervisor: Supervisor escalate (1 event(s)): stale persisted 300s 
  (possible wedge): fmdemo:fm-corpus-w1 (pre-read; re-arm not needed — watcher daemon-managed)

daemon log (state/.supervise-daemon.log), timestamps elided:
  self-handle: stale: fmdemo:fm-corpus-w1 -> transient stale (fmdemo:fm-corpus-w1): working: corpus rebuilt, handed to validation, waiting on the run

==================================================================================
 TARGET 6296cde - identical fixture, records, and stubbed CLI
==================================================================================
daemon under test : /home/shane/.no-mistakes/worktrees/5f306883d81c/01KZBX804SBBHVDV7C7EMY9WXF/bin/fm-supervise-daemon.sh
crew              : corpus-w1 on feat/corpus-rebuild, kind=ship, harness=claude
pane              : static since the handoff (never changes again)
busy record       : v1 gen=g1786038587.4128224.23981 seq=2 state=idle source=claude-hook event=stop ts=1786038587

away-mode wedge aging, one housekeeping pass per STALE_ESCALATE_SECS (240s):
  crew-state reads: state: working · source: run-step · validating (running)
  threshold 1  validation run still running           silent (marker refreshed, still aging)
  threshold 2  validation run still running           silent (marker refreshed, still aging)
  threshold 3  validation run still running           silent (marker refreshed, still aging)

  -- the crew now freezes: its validation run is gone (aborted / detached),
  -- and the pane text has still not changed by a single byte.
  crew-state reads: state: working · source: status-log · corpus rebuilt, handed to validation, waiting on the run
  threshold 4  no progress evidence                   ESCALATED -> no progress evidence for 300s (possible wedge): fmdemo:fm-corpus-w1

what the operator actually receives in the supervisor pane:
  ⁣FIRSTMATE_OP: v1 away-supervisor: Supervisor escalate (1 event(s)): no progress evidence for 
  300s (possible wedge): fmdemo:fm-corpus-w1 (pre-read; re-arm not needed — watcher 
  daemon-managed)

daemon log (state/.supervise-daemon.log), timestamps elided:
  self-handle: stale: fmdemo:fm-corpus-w1 -> transient stale (fmdemo:fm-corpus-w1): working: corpus rebuilt, handed to validation, waiting on the run
  stale marker refreshed (provably working, wedge aging restarts): fmdemo:fm-corpus-w1
  stale marker refreshed (provably working, wedge aging restarts): fmdemo:fm-corpus-w1
  stale marker refreshed (provably working, wedge aging restarts): fmdemo:fm-corpus-w1

==================================================================================
 TARGET - CONTROL: a genuinely hung crew that was never provably working
==================================================================================
daemon under test : /home/shane/.no-mistakes/worktrees/5f306883d81c/01KZBX804SBBHVDV7C7EMY9WXF/bin/fm-supervise-daemon.sh
crew              : corpus-w1 on feat/corpus-rebuild, kind=ship, harness=claude
pane              : static since the handoff (never changes again)
busy record       : v1 gen=g1786038590.4137363.32652 seq=2 state=idle source=claude-hook event=stop ts=1786038590

away-mode wedge aging, one housekeeping pass per STALE_ESCALATE_SECS (240s):
  crew-state reads: state: working · source: status-log · corpus rebuilt, handed to validation, waiting on the run
  threshold 1  hung, never provably working           ESCALATED -> no progress evidence for 300s (possible wedge): fmdemo:fm-corpus-w1

what the operator actually receives in the supervisor pane:
  ⁣FIRSTMATE_OP: v1 away-supervisor: Supervisor escalate (1 event(s)): no progress evidence for 
  300s (possible wedge): fmdemo:fm-corpus-w1 (pre-read; re-arm not needed — watcher 
  daemon-managed)

daemon log (state/.supervise-daemon.log), timestamps elided:
  self-handle: stale: fmdemo:fm-corpus-w1 -> transient stale (fmdemo:fm-corpus-w1): working: corpus rebuilt, handed to validation, waiting on the run
```
</details>
<details>
<summary>Evidence: Per-pass gate read budget and invalid-budget diagnostic (daemon log)</summary>

<code>five crews go due in the same housekeeping pass; four provably working, &#39;hung&#39; is not&#10;&#10;housekeeping pass 1&#10;crew-state reads this pass : 3 (alpha,bravo,charlie)&#10;markers still aged : alpha,bravo,charlie,delta,hung&#10;escalations buffered : none&#10;&#10;housekeeping pass 2&#10;crew-state reads this pass : 2 (delta,hung)&#10;markers still aged : alpha,bravo,charlie,delta&#10;escalations buffered : no progress evidence for 501s (possible wedge): fmdemo:fm-hung&#10;&#10;daemon log:&#10;stale marker refreshed (provably working, wedge aging restarts): fmdemo:fm-alpha&#10;stale marker refreshed (provably working, wedge aging restarts): fmdemo:fm-bravo&#10;stale marker refreshed (provably working, wedge aging restarts): fmdemo:fm-charlie&#10;stale gate deferred to next pass (read budget 3 spent, marker kept aged): fmdemo:fm-delta&#10;stale gate deferred to next pass (read budget 3 spent, marker kept aged): fmdemo:fm-hung&#10;stale marker refreshed (provably working, wedge aging restarts): fmdemo:fm-delta&#10;&#10;=== misconfigured budget: FM_STALE_WORKING_GATE_READS=&#39;4 &#39; ===&#10;reads over 4 passes with a standing typo : 12 (the sanitized default of 3 per pass)&#10;daemon log:&#10;invalid FM_STALE_WORKING_GATE_READS &#39;4 &#39;; using 3 &lt;- exactly once across all four passes</code>

```text
five crews go due in the same housekeeping pass; four provably working, 'hung' is not
FM_STALE_WORKING_GATE_READS is left at its default

housekeeping pass 1
  crew-state reads this pass : 3  (alpha,bravo,charlie)
  markers still aged         : alpha,bravo,charlie,delta,hung
  escalations buffered       : none

housekeeping pass 2
  crew-state reads this pass : 2  (delta,hung)
  markers still aged         : alpha,bravo,charlie,delta
  escalations buffered       : no progress evidence for 501s (possible wedge): fmdemo:fm-hung

housekeeping pass 3
  crew-state reads this pass : 0  ()
  markers still aged         : alpha,bravo,charlie,delta
  escalations buffered       : no progress evidence for 501s (possible wedge): fmdemo:fm-hung

daemon log so far, timestamps elided:
  stale marker refreshed (provably working, wedge aging restarts): fmdemo:fm-alpha
  stale marker refreshed (provably working, wedge aging restarts): fmdemo:fm-bravo
  stale marker refreshed (provably working, wedge aging restarts): fmdemo:fm-charlie
  stale gate deferred to next pass (read budget 3 spent, marker kept aged): fmdemo:fm-delta
  stale gate deferred to next pass (read budget 3 spent, marker kept aged): fmdemo:fm-hung
  stale marker refreshed (provably working, wedge aging restarts): fmdemo:fm-delta

==========================================================================
misconfigured budget: FM_STALE_WORKING_GATE_READS='4 ' (a trailing space)
==========================================================================
reads over 4 passes with a standing typo : 12 (the sanitized default of 3 per pass)
daemon log, timestamps elided:
  invalid FM_STALE_WORKING_GATE_READS '4 '; using 3
  stale marker refreshed (provably working, wedge aging restarts): fmdemo:fm-alpha
  stale marker refreshed (provably working, wedge aging restarts): fmdemo:fm-bravo
  stale marker refreshed (provably working, wedge aging restarts): fmdemo:fm-charlie
  stale gate deferred to next pass (read budget 3 spent, marker kept aged): fmdemo:fm-delta
  stale gate deferred to next pass (read budget 3 spent, marker kept aged): fmdemo:fm-hung
  stale marker refreshed (provably working, wedge aging restarts): fmdemo:fm-alpha
  stale marker refreshed (provably working, wedge aging restarts): fmdemo:fm-bravo
  stale marker refreshed (provably working, wedge aging restarts): fmdemo:fm-charlie
  stale gate deferred to next pass (read budget 3 spent, marker kept aged): fmdemo:fm-delta
  stale gate deferred to next pass (read budget 3 spent, marker kept aged): fmdemo:fm-hung
  stale marker refreshed (provably working, wedge aging restarts): fmdemo:fm-alpha
  stale marker refreshed (provably working, wedge aging restarts): fmdemo:fm-bravo
  stale marker refreshed (provably working, wedge aging restarts): fmdemo:fm-charlie
  stale gate deferred to next pass (read budget 3 spent, marker kept aged): fmdemo:fm-delta
  stale gate deferred to next pass (read budget 3 spent, marker kept aged): fmdemo:fm-hung
  stale marker refreshed (provably working, wedge aging restarts): fmdemo:fm-alpha
  stale marker refreshed (provably working, wedge aging restarts): fmdemo:fm-bravo
  stale marker refreshed (provably working, wedge aging restarts): fmdemo:fm-charlie
  stale gate deferred to next pass (read budget 3 spent, marker kept aged): fmdemo:fm-delta
  stale gate deferred to next pass (read budget 3 spent, marker kept aged): fmdemo:fm-hung
```
</details>
<details>
<summary>Evidence: Control-2 mutation witness (escalation follows the verdict, not a swallowed reader error)</summary>

<code>--- A-reader-answers-stopped ---&#10;ok - persistent stale escalates after threshold and clears its marker&#10;exit=0&#10;&#10;--- B-reader-answers-working ---&#10;not ok - persistent stale was not escalated&#10;exit=1&#10;&#10;--- C-reader-path-absent ---&#10;not ok - crew_absorb_class reached with an unusable FM_CREW_STATE_BIN (.../fakebin/no-such-reader.sh): its verdict would be an error, not a verdict. Point it at a reader that exists (make_fake_crew_state), or set FM_TEST_CREW_STATE_UNREADABLE=1 to exercise the unreadable-reader leg on purpose.&#10;exit=143</code>

```text
Negative-control witness for control 2 -- 'a genuinely hung pane still escalates'
(tests/fm-daemon.test.sh::test_housekeeping_persistent_stale_escalates)

--- A-reader-answers-stopped ---
exit=0

--- B-reader-answers-working ---
not ok - persistent stale was not escalated
exit=1

--- C-reader-path-absent ---
not ok - crew_absorb_class reached with an unusable FM_CREW_STATE_BIN (/tmp/fm-daemon-mutant.KJc4F2/stale-persistent/fakebin/no-such-reader.sh): its verdict would be an error, not a verdict. Point it at a reader that exists (make_fake_crew_state), or set FM_TEST_CREW_STATE_UNREADABLE=1 to exercise the unreadable-reader leg on purpose.
exit=143
```
</details>
<details>
<summary>Evidence: Scenario driver: end-to-end wedge aging (reusable)</summary>

```text
#!/usr/bin/env bash
# End-to-end operator-level demonstration of away-mode wedge aging.
#
# Nothing about the change under test is faked: the real housekeeping loop of the
# daemon passed as $1 runs against a real state dir, reading a real
# bin/fm-crew-state.sh verdict driven by a real bin/fm-busy-event.sh lifecycle
# record and a real git worktree HEAD. Only the terminal and the no-mistakes CLI
# are stubbed, because this box has no live fleet and no validation service.
#
# The crew shape is the reported one: a worker that handed its branch to the
# validation pipeline and is waiting on the result. Its harness turn has ENDED,
# so the pane is idle and byte-identical on every capture, while the pipeline run
# attributed to its exact HEAD is still running - demonstrably alive.
#
# usage: FM_DEMO_REPO=<checkout> wedge-scenario.sh <daemon> <outdir> <scenario>
set -u

DAEMON=$1
OUT=$(mkdir -p "$2" && cd "$2" && pwd)
SCENARIO=$3
REPO=${FM_DEMO_REPO:?set FM_DEMO_REPO to the checkout}

rm -rf "$OUT"; mkdir -p "$OUT"
STATE="$OUT/state"; FAKEBIN="$OUT/fakebin"; WT="$OUT/worktree"
mkdir -p "$STATE" "$FAKEBIN" "$WT"

ID=corpus-w1
WIN="fmdemo:fm-$ID"
KEY=$(printf '%s' "$ID" | tr ':/.' '___')
BRANCH=feat/corpus-rebuild
PANE="$OUT/crew-pane.txt"
SUPERVISOR_PANE="$OUT/supervisor-pane.txt"
SENT="$OUT/supervisor-typed.txt"
LOGFILE="$OUT/supervise-daemon.log"
: > "$SUPERVISOR_PANE"; : > "$SENT"; : > "$LOGFILE"

# --- the crew's worktree: a real git repo on the crew's branch --------------
git -C "$WT" init -q -b "$BRANCH"
git -C "$WT" -c user.email=demo@example.com -c user.name=demo commit -q --allow-empty -m "rebuild corpus"
HEAD_SHA=$(git -C "$WT" rev-parse HEAD)

# --- the crew's pane: byte-identical for the whole run ----------------------
cat > "$PANE" <<'PANETXT'
> handed the corpus rebuild to validation; waiting on the run before reporting
  no-mistakes run started for feat/corpus-rebuild
>
PANETXT

# --- stub terminal (the shared test fake, verbatim) -------------------------
cat > "$FAKEBIN/tmux" <<'SH'
#!/usr/bin/env bash
set -u
case "${1:-}" in
  display-message)
    [ "${FM_FAKE_TMUX_PANE_ALIVE:-1}" = "1" ] || exit 1
    _print=0
    for _a in "$@"; do
      case "$_a" in *cursor_y*) printf '%s\n' "${FM_FAKE_TMUX_CURSOR_Y:-0}"; exit 0 ;; esac
      [ "$_a" = "-p" ] && _print=1
    done
    [ "$_print" = 1 ] && printf 'fakepane\n'
    exit 0 ;;
  list-windows)
    [ -n "${FM_FAKE_TMUX_WINDOW:-}" ] && printf '%s\n' "$FM_FAKE_TMUX_WINDOW"
    exit 0 ;;
  capture-pane)
    _S=""; _E=""; shift
    while [ "$#" -gt 0 ]; do
      case "$1" in
        -S) _S="${2:-}"; shift 2; continue ;;
        -E) _E="${2:-}"; shift 2; continue ;;
        *) shift ;;
      esac
    done
    [ -n "${FM_FAKE_TMUX_CAPTURE:-}" ] || exit 0
    if [ -n "$_S" ] && [ -n "$_E" ]; then
      case "$_S$_E" in
        *[!0-9]*) cat "$FM_FAKE_TMUX_CAPTURE" 2>/dev/null ;;
        *) sed -n "$((_S + 1)),$((_E + 1))p" "$FM_FAKE_TMUX_CAPTURE" 2>/dev/null ;;
      esac
    else
      cat "$FM_FAKE_TMUX_CAPTURE" 2>/dev/null
    fi
    exit 0 ;;
  send-keys)
    while [ "$#" -gt 0 ]; do
      case "$1" in
        -l) shift; [ "$#" -gt 0 ] && {
          printf '%s\n' "$1" >> "${FM_FAKE_TMUX_SENT:-/dev/null}"
          [ -n "${FM_FAKE_TMUX_CAPTURE:-}" ] && printf '%s\n' "$1" >> "$FM_FAKE_TMUX_CAPTURE"
        } ;;
        Enter)
          printf '[ENTER]\n' >> "${FM_FAKE_TMUX_SENT:-/dev/null}"
          if [ -n "${FM_FAKE_TMUX_CAPTURE:-}" ] && [ -s "$FM_FAKE_TMUX_CAPTURE" ]; then
            _tmp="${FM_FAKE_TMUX_CAPTURE}.tmp"
            sed '$d' "$FM_FAKE_TMUX_CAPTURE" > "$_tmp" 2>/dev/null && mv -f "$_tmp" "$FM_FAKE_TMUX_CAPTURE"
          fi ;;
      esac
      shift
    done
    exit 0 ;;
esac
exit 1
SH
chmod +x "$FAKEBIN/tmux"

# --- stub no-mistakes CLI: serves whatever run object the scenario sets -----
cat > "$FAKEBIN/no-mistakes" <<'SH'
#!/usr/bin/env bash
set -u
case "${1:-}" in
  axi) shift
    case "${1:-}" in status) printf '%s\n' "${FM_FAKE_AXI_STATUS:-}" ;; logs) printf '\n' ;; esac ;;
  runs) printf '%s\n' "${FM_FAKE_RUNS_LIST:-}" ;;
esac
exit 0
SH
chmod +x "$FAKEBIN/no-mistakes"
export PATH="$FAKEBIN:$PATH"

# The validation run attributed to this crew's exact HEAD, still running.
RUN_RUNNING="run:
  id: \"01RUNCORPUS\"
  branch: $BRANCH
  status: running
  head: \"$HEAD_SHA\"
  pr: \"\"
  findings: none
  steps[2]{step,status,findings,duration_ms}:
    intent,completed,0,0
    review,running,0,0"
export FM_FAKE_AXI_STATUS="" FM_FAKE_RUNS_LIST=""

# --- the crew's recorded metadata and status log ----------------------------
cat > "$STATE/$ID.meta" <<META
worktree=$WT
window=$WIN
backend=tmux
harness=claude
kind=ship
META
printf 'working: corpus rebuilt, handed to validation, waiting on the run\n' \
  > "$STATE/$ID.status"

# --- a REAL busy lifecycle record, written by the real writer ---------------
# The harness turn is OVER (the crew submitted and stopped talking), so the
# semantic busy contract correctly reports idle. That is exactly why the pane
# goes static, and why a pane-text-only wedge timer cannot tell this apart
# from a hang.
GEN=$("$REPO/bin/fm-busy-event.sh" arm "$STATE" "$ID" --state busy --source fm-spawn --event launch-brief)
"$REPO/bin/fm-busy-event.sh" apply "$STATE" "$ID" idle --gen "$GEN" \
  --source claude-hook --event stop >/dev/null

crew_state_line() {
  FM_STATE_OVERRIDE="$STATE" FM_CREW_STATE_NM_TIMEOUT=5 "$REPO/bin/fm-crew-state.sh" "$ID"
}

printf 'daemon under test : %s\n' "$DAEMON"
printf 'crew              : %s on %s, kind=ship, harness=claude\n' "$ID" "$BRANCH"
printf 'pane              : static since the handoff (never changes again)\n'
printf 'busy record       : %s\n' "$(cat "$STATE/$ID.busy-state")"
echo

# --- source the daemon under test ------------------------------------------
export FM_TEST_DAEMON_SOURCED=1 FM_WEDGE_ALARM_CHANNEL=off FM_ROOT_OVERRIDE="$REPO"
# shellcheck disable=SC1090
. "$DAEMON"

export FM_STATE_OVERRIDE="$STATE"
export FM_STALE_ESCALATE_SECS=240
export FM_CREW_STATE_BIN="$REPO/bin/fm-crew-state.sh"
export FM_CREW_STATE_NM_TIMEOUT=5
export FM_FAKE_TMUX_WINDOW="$WIN"
export FM_FAKE_TMUX_CAPTURE="$PANE"
export LOG="$LOGFILE"
afk_enter "$STATE"          # away mode: the daemon owns triage and may inject

MARKER="$STATE/.subsuper-stale-$KEY"

# ONE away-mode stale wake for the unchanged pane, through the daemon's own
# path. In away mode the watcher enqueues at most one stale wake per distinct
# pane hash, so this happens exactly once for the whole run; everything after it
# is the daemon's own aging.
handle_wake "stale: $WIN" "$STATE" >/dev/null 2>&1
[ -e "$MARKER" ] || { echo "FATAL: the stale wake recorded no wedge marker"; exit 1; }

# Advance the clock one STALE_ESCALATE_SECS the only way a script can: back-date
# the marker the daemon itself owns. This script never re-creates it, only ages
# it - so a design that DROPPED the marker instead of refreshing it would leave
# nothing to age, and would show up below as permanent silence.
age_one_threshold() {
  [ -e "$MARKER" ] || return 0
  echo $(( $(cat "$MARKER") - 300 )) > "$MARKER"
}

SEEN_ESCALATIONS=0
report_pass() {  # <n> <label>  — reports what THIS pass added, not the buffer total
  local esc="$STATE/.subsuper-escalations" now_n added
  if [ -e "$esc" ]; then now_n=$(wc -l < "$esc" | tr -d ' '); else now_n=0; fi
  added=$(( now_n - SEEN_ESCALATIONS )); SEEN_ESCALATIONS=$now_n
  printf '  threshold %s  %-38s ' "$1" "$2"
  if [ "$added" -gt 0 ]; then
    printf 'ESCALATED -> %s\n' "$(tail -1 "$esc")"
  elif [ -e "$MARKER" ]; then
    printf 'silent (marker refreshed, still aging)\n'
  else
    printf 'silent (no marker left to age)\n'
  fi
}

echo "away-mode wedge aging, one housekeeping pass per STALE_ESCALATE_SECS (240s):"
case "$SCENARIO" in
  working-then-frozen)
    FM_FAKE_AXI_STATUS="$RUN_RUNNING"
    echo "  crew-state reads: $(crew_state_line)"
    for n in 1 2 3; do
      age_one_threshold; housekeeping "$STATE"; report_pass "$n" "validation run still running"
    done
    echo
    echo "  -- the crew now freezes: its validation run is gone (aborted / detached),"
    echo "  -- and the pane text has still not changed by a single byte."
    FM_FAKE_AXI_STATUS=""
    echo "  crew-state reads: $(crew_state_line)"
    for n in 4 5; do
      age_one_threshold; housekeeping "$STATE"; report_pass "$n" "no progress evidence"
      [ -s "$STATE/.subsuper-escalations" ] && break
    done
    ;;
  never-working)
    # Control: a genuinely hung crew that was NEVER provably working. Same static
    # pane, same marker, same threshold; no run was ever attributed to it.
    FM_FAKE_AXI_STATUS=""
    echo "  crew-state reads: $(crew_state_line)"
    for n in 1 2; do
      age_one_threshold; housekeeping "$STATE"; report_pass "$n" "hung, never provably working"
      [ -s "$STATE/.subsuper-escalations" ] && break
    done
    ;;
esac

echo
echo "what the operator actually receives in the supervisor pane:"
FM_FAKE_TMUX_CAPTURE="$SUPERVISOR_PANE" FM_FAKE_TMUX_SENT="$SENT" FM_ESCALATE_BATCH_SECS=0 \
  escalate_flush "$STATE" >/dev/null 2>&1 || true
if grep -q . "$SENT" 2>/dev/null; then
  grep -v '^\[ENTER\]$' "$SENT" | fold -s -w 96 | sed 's/^/  /'
else
  echo "  (nothing injected - the daemon stayed silent)"
fi

echo
echo "daemon log (state/.supervise-daemon.log), timestamps elided:"
if [ -s "$LOGFILE" ]; then sed 's/^\[[^]]*\] /  /' "$LOGFILE"; else echo "  (empty)"; fi
```
</details>
<details>
<summary>Evidence: Scenario driver: per-pass gate read budget (reusable)</summary>

```text
#!/usr/bin/env bash
# Operator-level demonstration of the per-pass gate read budget and its
# misconfiguration diagnostic - the second half of this branch's work.
#
# Five crews go due in the SAME housekeeping pass. Four are provably working
# (a validation run attributed to their exact HEAD); one is genuinely hung.
# The gate is allowed FM_STALE_WORKING_GATE_READS crew-state reads per pass, so
# the pass cannot serialize five bounded reads inline in the daemon's main loop.
# What must be visible: every read the budget denies is DEFERRED (marker left
# aged, untouched), never absorbed - so the hung crew still escalates, just a
# pass later.
#
# usage: FM_DEMO_REPO=<checkout> gate-budget-scenario.sh <daemon> <outdir>
set -u

DAEMON=$1
OUT=$(mkdir -p "$2" && cd "$2" && pwd)
REPO=${FM_DEMO_REPO:?set FM_DEMO_REPO to the checkout}

rm -rf "$OUT"; mkdir -p "$OUT"
STATE="$OUT/state"; FAKEBIN="$OUT/fakebin"
mkdir -p "$STATE" "$FAKEBIN"
PANE="$OUT/crew-pane.txt"; LOGFILE="$OUT/supervise-daemon.log"
READS="$OUT/crew-state-reads.txt"
printf '>\n' > "$PANE"; : > "$LOGFILE"; : > "$READS"

cat > "$FAKEBIN/tmux" <<'SH'
#!/usr/bin/env bash
set -u
case "${1:-}" in
  display-message) printf 'fakepane\n'; exit 0 ;;
  list-windows) exit 0 ;;
  capture-pane) [ -n "${FM_FAKE_TMUX_CAPTURE:-}" ] && cat "$FM_FAKE_TMUX_CAPTURE"; exit 0 ;;
esac
exit 1
SH
chmod +x "$FAKEBIN/tmux"

# A crew-state reader that logs every call, so the budget is COUNTED rather than
# reasoned about, and answers per-crew from the fixture the daemon can see.
cat > "$FAKEBIN/fm-crew-state.sh" <<'SH'
#!/usr/bin/env bash
set -u
id=${1:-}
printf '%s\n' "$id" >> "${FM_DEMO_READ_LOG:-/dev/null}"
if [ -e "${FM_DEMO_STATE:-}/$id.validating" ]; then
  printf 'state: working · source: run-step · validating (running)\n'
else
  printf 'state: unknown · source: none · no current-state source available\n'
fi
exit 0
SH
chmod +x "$FAKEBIN/fm-crew-state.sh"
export PATH="$FAKEBIN:$PATH"

CREWS="alpha bravo charlie delta hung"
for c in $CREWS; do
  cat > "$STATE/$c.meta" <<META
window=fmdemo:fm-$c
backend=tmux
harness=claude
kind=ship
META
  printf 'working: handed to validation, waiting on the run\n' > "$STATE/$c.status"
  [ "$c" = hung ] || : > "$STATE/$c.validating"
  echo $(( $(date +%s) - 500 )) > "$STATE/.subsuper-stale-$c"
done

export FM_TEST_DAEMON_SOURCED=1 FM_WEDGE_ALARM_CHANNEL=off FM_ROOT_OVERRIDE="$REPO"
# shellcheck disable=SC1090
. "$DAEMON"

export FM_STATE_OVERRIDE="$STATE" FM_STALE_ESCALATE_SECS=240
export FM_CREW_STATE_BIN="$FAKEBIN/fm-crew-state.sh"
export FM_DEMO_READ_LOG="$READS" FM_DEMO_STATE="$STATE"
export FM_FAKE_TMUX_CAPTURE="$PANE"
export LOG="$LOGFILE"

show_pass() {  # <n>
  local before_reads=$1 esc="$STATE/.subsuper-escalations" now
  now=$(wc -l < "$READS" | tr -d ' ')
  printf '  crew-state reads this pass : %s  (%s)\n' "$(( now - before_reads ))" \
    "$(tail -n +$(( before_reads + 1 )) "$READS" | paste -sd, -)"
  printf '  markers still aged         : %s\n' \
    "$(cd "$STATE" && ls .subsuper-stale-* 2>/dev/null | sed 's/.subsuper-stale-//' | paste -sd, - || echo none)"
  printf '  escalations buffered       : %s\n' \
    "$( [ -s "$esc" ] && cat "$esc" | paste -sd' | ' - || echo none )"
}

echo "five crews go due in the same housekeeping pass; four provably working, 'hung' is not"
echo "FM_STALE_WORKING_GATE_READS is left at its default"
echo
for pass in 1 2 3; do
  n=$(wc -l < "$READS" | tr -d ' ')
  echo "housekeeping pass $pass"
  housekeeping "$STATE"
  show_pass "$n"
  echo
done

echo "daemon log so far, timestamps elided:"
sed 's/^\[[^]]*\] /  /' "$LOGFILE"

echo
echo "=========================================================================="
echo "misconfigured budget: FM_STALE_WORKING_GATE_READS='4 ' (a trailing space)"
echo "=========================================================================="
: > "$LOGFILE"; : > "$READS"; : > "$STATE/.subsuper-escalations"
for c in $CREWS; do echo $(( $(date +%s) - 500 )) > "$STATE/.subsuper-stale-$c"; done
for pass in 1 2 3 4; do
  FM_STALE_WORKING_GATE_READS='4 ' housekeeping "$STATE"
  for c in $CREWS; do
    [ -e "$STATE/.subsuper-stale-$c" ] && echo $(( $(date +%s) - 500 )) > "$STATE/.subsuper-stale-$c"
  done
done
printf 'reads over 4 passes with a standing typo : %s (the sanitized default of 3 per pass)\n' \
  "$(wc -l < "$READS" | tr -d ' ')"
echo "daemon log, timestamps elided:"
sed 's/^\[[^]]*\] /  /' "$LOGFILE"
```
</details>
<details>
<summary>Evidence: tests/fm-daemon.test.sh full output (108 assertions)</summary>

```text
ok - a non-terminal stale wake records its marker without reading crew state
ok - housekeeping refreshes a persistent stale that is provably working, while an unreadable crew still escalates
ok - a refreshed wedge marker escalates within one more threshold once the working verdict stops
ok - the provably-working gate costs one crew-state read per threshold and leaves watcher markers alone
ok - a gate read deferred by the per-pass budget delays an escalation by one pass and never suppresses it
ok - installing the crew-state reader guard twice leaves the alias unwrapped, answering, and still guarded
ok - an invalid gate budget restores the default and is logged, while 0 floors to 1 quietly
ok - housekeeping spends the sanitized budget and reports the invalid value it replaced
ok - through housekeeping itself, the invalid-budget report fires once per standing typo, again on change, and only past its reminder interval
```
</details>
- Evidence: tests/fm-watch-triage.test.sh full output (46 assertions) (local file: <code>/tmp/no-mistakes-evidence/01KZBX804SBBHVDV7C7EMY9WXF/fm-watch-triage.test.out</code>)

## Pipeline

_... (5 earlier update rounds omitted to keep the PR body within GitHub's 65536-char limit; full history is in the run log.)_

<details>

🔧 Fix: restore default on invalid gate budget, report it loudly
- ⚠️ `bin/fm-supervise-daemon.sh:1013` - The comment at line 998 states the warning &#34;re-reports every pass it is read on, because the log is trimmed to its newest lines and a single startup warning can age out&#34;. But stale_gate_read_budget is called once per housekeeping() invocation, which the main loop runs every FM_HOUSEKEEPING_TICK (default 15s), so one config typo emits ~4 lines/min, ~5760/day, at roughly 75 bytes each. trim_log (line 1648) keeps only FM_LOG_KEEP_LINES (2000, line 258), so the retained daemon log shrinks to under ~9 hours and is dominated by this single repeated line - evicting exactly the watcher-crash, escalation and wedge-alarm history an operator needs when investigating. The cure crowds out the same trimmed window the reasoning was protecting. A throttle keeps both properties: log on the first read and then once per N passes (or whenever the raw value changes), so the warning stays present in any recent window without becoming the window.
- ℹ️ `bin/fm-supervise-daemon.sh:1140` - Acknowledgement retained at the author&#39;s request, no action wanted: the intent&#39;s MERGE STATUS section still declares this PR blocked because crew_absorb_class reports working indefinitely for a worker killed mid-turn (stale state=busy record with no expiry), which this gate converts from a detectable wedge into permanent silence. The commission forbids compensating inside the daemon, and this commit correctly adds no refresh cap, no advancing-evidence check and no call-site expiry. Recorded so a clean review is not mistaken for the merge gate being satisfied.

🔧 Fix: throttle invalid gate-budget diagnostic to bounded cadence
4 issues (1 error, 3 infos) still open:
- 🚨 `bin/fm-supervise-daemon.sh:1058` - The report throttle is inert at its only call site. housekeeping does `working_reads_max=$(stale_gate_read_budget)`, and command substitution runs in a subshell, so the `_fm_gate_budget_last_invalid` and `_fm_gate_budget_reported_at` assignments at lines 1029-1030 are discarded the moment the substitution returns. Every pass re-enters with `${_fm_gate_budget_last_invalid:-}` empty, which never equals a non-empty $raw, so the first arm of the || at line 1027 is always true and the warning fires every FM_HOUSEKEEPING_TICK - the same ~4 lines/min flood this round was meant to bound. I confirmed the scoping with a direct bash probe: a global set inside $(f) reads back empty, set by a direct call it persists. The comment at line 1012 (&#34;The two facts it remembers live in this process for the daemon&#39;s lifetime&#34;) is false at the only place the function is used. The new test at tests/fm-daemon.test.sh:687-691 cannot catch it because it calls stale_gate_read_budget directly with &gt;/dev/null instead of through a substitution, and test_housekeeping_spends_the_sanitized_budget runs a single pass asserting the message is present, which passes either way. Fix the invocation, not the helper&#39;s logic: have housekeeping call it as a plain command that assigns a named global (or hoist the report out of the substituted path), and change the throttle test to drive it through the production form so a repeat pass can be observed suppressed.
- ℹ️ `tests/fm-daemon.test.sh:691` - The periodic-reminder leg sets STALE_GATE_INVALID_REMIND_SECS=0, which makes `[ $((now - reported_at)) -ge 0 ]` unconditionally true for any value of reported_at, including one that was never set. It proves the elapsed arm can emit, but not that the interval is honoured or that the timestamp is recorded correctly - a broken elapsed computation would still pass. Use a small non-zero interval and back-date _fm_gate_budget_reported_at so the comparison itself is what decides the outcome.
- ℹ️ `bin/fm-supervise-daemon.sh:1` - The round-4 instruction asked that the inert-assertion class be recorded as follow-up work in the commit body, naming the tests/fm-wake-daemon-lifecycle-e2e.test.sh phase-3 case whose assertions are satisfied by a capture failure returning 2 as well as by the busy verdict they name. Commit fa17ea1 has no body, but that is the pipeline&#39;s convention rather than an omission: every no-mistakes(review) commit on this branch is subject-only, while the two hand-authored commits (ae6d0ce, 6187dd7) carry full bodies. The record therefore needs to land in the pull-request description that a later step of this run produces. Noted so it is not lost, not as a defect in the commit.
- ℹ️ `bin/fm-supervise-daemon.sh:1162` - Acknowledgement retained at the author&#39;s request, no action wanted: the intent&#39;s MERGE STATUS section still declares this PR blocked because crew_absorb_class reports working indefinitely for a worker killed mid-turn (stale state=busy record with no expiry), which this gate converts from a detectable wedge into permanent silence. The commission forbids compensating inside the daemon, and this commit correctly adds no refresh cap, no advancing-evidence check and no call-site expiry. Recorded so a clean review is not mistaken for the merge gate being satisfied.

🔧 Fix: carry gate-budget throttle state through housekeeping&#39;s real call
3 infos still open:
- ℹ️ `tests/fm-daemon.test.sh:725` - `n=$(grep -Fc &#34;$msg&#34; file 2&gt;/dev/null || echo 0)` is unsafe when the file exists with zero matches: grep -c prints &#34;0&#34; AND exits 1, so the || fires and n becomes the two-line string &#34;0\n0&#34;. I probed this to confirm - existing file with no match yields &#34;0\n0&#34;, missing file yields &#34;0&#34;. The zero-assertion at line 726 therefore passes only because log() never creates within.log when nothing is emitted; the moment housekeeping emits any unrelated line during that leg (a stale-gate deferral, a marker refresh, an inject alarm), the file exists, n is &#34;0\n0&#34;, and `[ &#34;$n&#34; = 0 ]` reports a false failure. The -ge 1 legs at 721 and 728 would hit `integer expression expected` in the same situation. Use `n=$(grep -Fc &#34;$msg&#34; file 2&gt;/dev/null) || n=0` so the count is a single token either way. Related, same test: the probe&#39;s `|| fail` at line 716 depends on housekeeping exiting 0, but its final construct is the heartbeat-scan `while read` loop, which returns 1 when the scan is due and scan_captain_relevant_statuses is empty - it is 0 here only because .subsuper-last-scan is written on the first of the five passes, so the scan is not due again. Reordering the probe so a scan-due pass runs last would produce a spurious failure.
- ℹ️ `bin/fm-supervise-daemon.sh:1` - Tracking note, no source defect: the round-5 instruction redirected the inert-assertion class record from the commit body to the pull-request description, and widened it to cover this round&#39;s case - a fix, not just an assertion, that was inert on the real path while its tests passed. That record is owned by the PR-authoring step later in this run, so there is nothing to verify in the tree now. It must name tests/fm-wake-daemon-lifecycle-e2e.test.sh phase 3 as the concrete unfixed example (assertions satisfied by a capture failure returning 2 as well as by the busy verdict they name), state that generalising the existing crew-state-reader harness guard is the real fix, and say explicitly that it is deliberately not fixed in this branch.
- ℹ️ `bin/fm-supervise-daemon.sh:1167` - Acknowledgement retained at the author&#39;s request, no action wanted: the intent&#39;s MERGE STATUS section still declares this PR blocked because crew_absorb_class reports working indefinitely for a worker killed mid-turn (stale state=busy record with no expiry), which this gate converts from a detectable wedge into permanent silence. The commission forbids compensating inside the daemon, and this commit correctly adds no refresh cap, no advancing-evidence check and no call-site expiry. Recorded so a clean review is not mistaken for the merge gate being satisfied.

</details>

<details>

✅ No issues found.
- <code>`bash tests/fm-daemon.test.sh` - 108 assertions, exit 0, no unbound-variable or command-not-found noise</code>
- <code>`bash tests/fm-watch-triage.test.sh` - 46 assertions, exit 0</code>
- <code>`bash tests/fm-wake-daemon-lifecycle-e2e.test.sh` - exit 0 (its housekeeping call sites were changed by this branch)</code>
- <code>Manual end-to-end: `wedge-scenario.sh &lt;daemon&gt; &lt;out&gt; working-then-frozen` run against BOTH the base daemon (`2cf0283:bin/fm-supervise-daemon.sh`, staged with symlinked sibling libs) and the target daemon, on one fixture using the real `bin/fm-crew-state.sh`, real `bin/fm-busy-event.sh arm/apply` records, and a real git worktree HEAD</code>
- <code>Manual end-to-end control: `wedge-scenario.sh &lt;target daemon&gt; &lt;out&gt; never-working` - a genuinely hung crew, never provably working, still escalates at threshold 1</code>
- <code>Manual end-to-end: `gate-budget-scenario.sh` - five crews due in one housekeeping pass; counted crew-state reads per pass, inspected which markers stayed aged, and confirmed the hung crew escalates one pass later</code>
- <code>Manual end-to-end: same script with `FM_STALE_WORKING_GATE_READS=&#39;4 &#39;` over four passes - 12 reads total (sanitized default of 3/pass) and exactly one `invalid FM_STALE_WORKING_GATE_READS &#39;4 &#39;; using 3` log line</code>
- <code>Mutation witness of control 2 (`test_housekeeping_persistent_stale_escalates`) in three variants: reader answers `stopped` (green), reader answers `working` (red - `not ok - persistent stale was not escalated`), reader path absent (aborts via `fm_guard_crew_state_reader` with exit 143 and its diagnostic)</code>
- <code>`git status --porcelain` and `git clean -nd` - worktree left clean; removed the `/tmp/fm-daemon-mutant.*` temp roots my mutation runs created</code>

</details>

<details>

- ℹ️ `.agents/skills/afk/SKILL.md:175` - The change states the gate&#39;s per-pass read-budget capacity arithmetic twice in full: docs/architecture.md:26-30 and .agents/skills/afk/SKILL.md:174-177 both carry the ceil(D/B) - 1 burst formula, the ~48-simultaneously-stale-crew starvation threshold, the stable-glob-order unfairness note, and the advice to raise FM_STALE_WORKING_GATE_READS. Both copies are currently accurate, so nothing is stale today, but they are synchronized restatements of one contract and will drift the moment only one is edited (for example if the default budget or FM_HOUSEKEEPING_TICK changes, since the &#39;45s&#39; and &#39;48 crews&#39; numbers are derived from both). The repo&#39;s one-owner rule in .agents/skills/firstmate-coding-guidelines/SKILL.md forbids exactly this. Deciding which surface owns it is a judgment call I left to the author: the agent-runtime /afk skill plausibly owns the away-mode classification table, while capacity tuning reads as maintainer-architecture plus the configuration.md tunable. Proposed follow-up: keep the agent-operative facts in the skill (the gate exists, a deferred read delays but never suppresses an escalation) and reduce the burst formula, starvation threshold, and glob-order fairness note to a one-line cross-reference to docs/architecture.md. Deliberately not done here to avoid restructuring documentation this change did not make stale.

🔧 Fix: consolidate gate read-budget arithmetic under docs/architecture.md

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>


kunchenguid:

PR #1835, opened 2026-08-06
watcher: use herdr native idle state for declared-pause cadence

mish18:
Fix the watcher's declared-pause cadence for live-but-idle herdr agents.

Problem: A crew with a declared paused:/captain-held status and a herdr agent whose agent_status is idle would previously get pause_state_class = none on every new pane hash, because fm_backend_agent_alive returns alive for any registered herdr agent. The watcher then surfaced a bare stale: wake roughly every time the idle pi TUI footer re-rendered.

Fix:
- Add fm_backend_agent_is_idle in bin/fm-backend.sh: for herdr it returns true only when the native agent_status is confidently idle; other backends return false.
- In bin/fm-watch.sh pause_state_class, check native idle before the existing live-agent fallback so that declared-pause + herdr idle keeps the bounded PAUSE_RESURFACE_SECS cadence instead of firing per-hash bare stales.
- Unknown/unreadable liveness still classifies as none; tmux and backends without a native idle semantic keep current behavior; working, terminal, wedge-escalation, secondmate, and afk paths are untouched.

Tests:
- Add install_fake_herdr to tests/wake-helpers.sh.
- Extend tests/fm-watch-triage.test.sh with four herdr cases: declared-pause + live-idle agent uses pause cadence and survives new pane hashes; declared-pause + dead agent unchanged; declared-pause + unreadable liveness surfaces; non-paused + idle agent unchanged.

Motivated by the 2026-08-06 p5 noise pattern observed for task-mctaskface-mvp-p1 on default:wA:p5.

### DIFF
--- a/bin/fm-backend.sh
+++ b/bin/fm-backend.sh
@@ -905,6 +905,25 @@ fm_backend_agent_alive() {  # <backend> <target>
   esac
 }
 
+# fm_backend_agent_is_idle: 0 only when a backend exposes a NATIVE semantic
+# idle state for the agent at <target> and that state is confidently idle.
+# Herdr is the first supported backend: its `agent get` surface reports an
+# explicit "idle" agent_status. Backends without such a primitive (tmux and
+# all others) always return 1, preserving the existing "live means not dead"
+# triage path for them.
+fm_backend_agent_is_idle() {  # <backend> <target>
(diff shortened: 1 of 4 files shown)


kunchenguid:

PR #1862, opened 2026-08-07
Add latent PR worker hibernation lifecycle

ivannovak:
## Summary

- add crash-safe latent worker hibernation for exact authenticated GitHub PR heads
- preserve recoverable commits under protected local refs while releasing worker processes and temporary Treehouse slots
- add resume, attention, startup recovery, merge finalization, visibility, and project-removal safeguards
- use durable `pr_ready_head`/`pr_head` identity rather than status-log pause tails for eligibility
- keep GitLab and tasks with dirty/divergent work, open decisions, validation, or pending obligations active
- reconcile the branch with main's deferred session-start network stage (#1860, #1918)

## Reconciliation with main

This branch and `main` both restructured how session start sequences its work, so the merge was a composition decision rather than a hunk-by-hunk pick.

`main` split `bin/fm-bootstrap.sh` into a local pass and a network-only pass (`FM_BOOTSTRAP_NETWORK=skip|only`), moved four network sweeps into the detached bounded worker owned by `bin/fm-startup-network.sh`, and reports them in the digest's own `NETWORK CHECKS` section. This branch added an additional mutating sweep - latent transaction recovery - written against the pre-deferral ordering, where that local/network split did not exist.

Picking either side would have destroyed the other: taking the branch's shape puts the network back on the blocking path, which is the exact regression #1860 fixed; taking main's shape leaves the hibernation sweep attached to a stage that no longer runs, so the lifecycle would look installed while being inert.

### Where the hibernation lifecycle belongs, and why

The lifecycle splits cleanly along the same seam main introduced, because its two session-start-relevant halves have different network characters.

**`recover-all` (crash recovery) runs on the LOCAL pass, inside main's `local_phase` guard.** `latent_recover_one` reconciles an interrupted hibernation transaction against `git rev-parse` on the local clone, the local metadata hash, and the local manifest. It makes no external-network call, so deferring it would buy no startup latency. It would also cost correctness: the deferred stage is bounded at 120s and routinely still `IN PROGRESS` when the digest is composed, so a half-applied transaction would remain unreconciled underneath the very `FLEET STATE` section that reports that task's endpoint. It is placed alongside the PR-check migration and the startup-memory budget, and it is gated by `FM_BOOTSTRAP_DETECT_ONLY` exactly as they are, so a lock-refused read-only session never touches it.

**`enter` (hibernation itself) stays off the session-start path entirely.** This is the network half - `latent_fetch_exact_head` fetches the exact PR object and the authenticated canonical poll must be valid before a worker is released. It is event-driven and always has been: `bin/fm-pr-check.sh` drives it when a PR is registered, and `bin/fm-watch.sh` drives it from the supervision poll. Neither is on the blocking startup path, so no part of hibernation needed to move into the deferred stage. Session start therefore still makes no blocking external-network call.

The loop in bootstrap only probes for a transaction journal, so a home with no interrupted hibernation pays a single failed glob.

### Documentation

The sweep count was six on main and seven on the branch, and the two lists disagreed. It is now seven in all three places that state it - `AGENTS.md` section 3, `bin/fm-bootstrap.sh`'s header, and `bin/fm-session-start.sh`'s ordering comment - with main's "the four network ones among them run in the deferred stage" clause preserved, which stays accurate because latent recovery is local. The `bootstrap-diagnostics` trigger list is the union of both sides (`NETWORK_CHECKS:` from main, `LATENT_RECOVERY:` from this branch) in both `AGENTS.md` and the skill's own front matter.

## Verification

Run against the merged tree in an isolated scratch `FM_HOME`, not read off the diff.

### 1. Composed ordering, with both stages present

`bin/fm-session-start.sh` emits main's nine-stage ordering intact:

```
LOCK → BOOTSTRAP → WAKE QUEUE → (supervision block) → FLEET STATE → NETWORK CHECKS → CONTEXT → NEXT STEP
```

With a latent transaction journal planted at `data/fm-hib-probe/latent/transaction`, the hibernation sweep reports inside the blocking `BOOTSTRAP` section:

```
LATENT_RECOVERY: quarantined fm-hib-probe
```

and the network stage reports separately, off the blocking path:

```
NETWORK CHECKS
completed off the startup path in 1s: GitHub authentication, dead-secondmate relaunch,
secondmate convergence, pending handoff delivery, and project clone refresh with its drift reporting.
```

### 2. Watching it fail

The sweep's wiring was broken deliberately and session start re-run each time.

| Wiring | `LATENT_RECOVERY` in digest |
| --- | --- |
| intact (as merged) | present |
| sweep removed | **absent** |
| sweep moved to the deferred network phase | **absent from the entire digest** |
| restored | present |

The second break is the more instructive one. Moved to the deferred phase the sweep is not merely delayed - the digest completes with `NETWORK CHECKS` still `IN PROGRESS - ... Started 2s ago, bounded at 120s`, so the recovery result is unavailable at the moment the digest reports that task's endpoint. That is the "installed but inert" failure this placement avoids.

Phase gating was then checked directly against `bin/fm-bootstrap.sh`:

| Pass | Sweep runs? | Required |
| --- | --- | --- |
| `FM_BOOTSTRAP_DETECT_ONLY=1` (lock-refused read-only) | no | no |
| `FM_BOOTSTRAP_NETWORK=skip` (locked local, blocking path) | yes | yes |
| `FM_BOOTSTRAP_NETWORK=only` (deferred worker) | no | no - must not repeat |

An earlier version of this check used `FM_BOOTSTRAP_NETWORK_PHASE`, which is the internal normalized variable rather than the input `FM_BOOTSTRAP_NETWORK`. It silently fell back to the complete run and appeared to show the sweep repeating on the deferred pass. The instrument was wrong, not the code; the table above is from the corrected run.

### 3. The network deferral survives

A project clone was pointed at an unroutable origin (`https://10.255.255.1/`) so the fleet-sync fetch would hang, then session start was timed:

- session start returned in **2s**, exit 0
- the deferred worker took **21s**, of which `phase fleet-sync ... elapsed=20000` ms was the hung fetch
- the digest printed `IN PROGRESS`, named each unconfirmed check, and never reported one as passed

The 20s of blocked network never touched the blocking path, and the hibernation sweep still reported on it in the same run.

### 4. Hibernated-task visibility in the composed digest

A `tier=latent` task with no window or worktree is reported as released rather than misread as broken:

```
endpoint: released (state: quarantined · source: latent-manifest · latent recovery integrity failed)
```

Without this branch's `fm-session-start.sh` case that line reads `endpoint: unknown (no window recorded)`.

### 5. Regression test

`tests/fm-bootstrap.test.sh` gains `test_latent_recovery_sweep_runs_on_the_local_half`, which pins the sweep to the local half, asserts the network half does not repeat it, and asserts a detect-only session does not run it. Each of its three assertions was watched failing before it landed, with unpiped exit 1:

- sweep removed → `not ok - the local half did not run latent transaction recovery`
- sweep moved outside both guards → `not ok - the deferred network half repeated latent transaction recovery`
- only the detect-only guard dropped → `not ok - a detect-only session ran latent transaction recovery`

Restored, it returns `ok - bootstrap: latent transaction recovery runs on the local half only` with exit 0.

## Safety boundaries preserved

- **Lock-refused read-only sessions** run no mutating sweep, hibernation recovery included (verified above).
- **Sweeps only when this session holds the lock** - the latent sweep sits inside the same `FM_BOOTSTRAP_DETECT_ONLY` guard as the other local sweeps.
- **Hibernation is not teardown** - `bin/fm-latent.sh`'s header keeps that statement, `finish` `exec`s into `bin/fm-teardown.sh`, and teardown still requires the forge's exact merged head to contain the saved generation before deleting the ref.
- **Exact-PR-head verification before releasing a worker** - `latent_enter` still proves PR identity, ancestry, cleanliness, obligations, validation, endpoint, and recovery-ref state twice, before and after backend termination.
- **No force or discard bypass** anywhere in the lifecycle.

## Local evidence

- `bin/fm-lint.sh` passed, ShellCheck 0.11.0 (pinned 0.11.0), exit 0, including a `CI=true` full-parity run over the complete file set
- `bin/fm-doc-audience-check.sh` passed: `ok surfaces=65 local_links=216`
- `bin/fm-test-run.sh --check-coverage` passed: `FM_TEST_COVERAGE ok total=133 parallel=24 serial=98 serial_shards=4 herdr=11`
- `bin/fm-test-run.sh --all`, verbatim: `FM_TEST_SUMMARY total=133 failed=10 skipped_gate=15 duration_ms=3861068`, exit 1

Every test that covers this merge passed on the merged tree:

| Test | Result |
| --- | --- |
| `tests/fm-latent.test.sh` (whole hibernation lifecycle) | `exit=0` |
| `tests/fm-bootstrap.test.sh` (incl. the new placement test) | `exit=0` |
| `tests/fm-session-start.test.sh` | `exit=0` |
| `tests/fm-startup-network.test.sh` | `exit=0` |
| `tests/fm-crew-state.test.sh` (latent-tier reconciliation) | `exit=0` |

Nine of the ten failures were reproduced on clean `main` (833a9a2) and are pre-existing in this
environment - missing or wrong-version external dependencies (Orca, Pi, Kimi, Muse, Herdr 0.7.5
against a documented 0.8.0 floor, and remote SSH):

`fm-backend-orca`, `fm-calm-pi-extension`, `fm-gotmp`, `fm-kimi-harness`, `fm-muse-harness`,
`fm-remote-secondmate-lifecycle-e2e`, `fm-remote-secondmate-trace-context`,
`fm-backend-herdr-focus-flash-e2e`, `fm-backend-herdr-presentation-e2e`.

The tenth was a real defect in the hibernation commit; it is described and FIXED below, so the teardown suite is now green.

## Exit-status defect found in the hibernation commit, and fixed here

`tests/fm-teardown.test.sh` case `herdr-preflight-missing-adapter` fails differently before and
after the hibernation commit, and the difference is a genuine behavioural regression that this
merge only carries forward. It is reported here rather than silently repaired, because fixing it
changes the feature rather than the merge.

Bisected in one clone under identical untangled git conditions, so the only variable is the code:

| Commit | `bin/fm-latent.sh` | teardown exit | assertion reached |
| --- | --- | --- | --- |
| `8398d31` merge base | absent | **1** | fails later, on the stderr-wording assertion |
| `833a9a2` current main | absent | **1** | fails later, on the stderr-wording assertion |
| `c150076` hibernation commit | present | **0** | fails earlier: "teardown continued without its required preflight" |
| `2583e05` this merge | present | **0** | same as `c150076` |

The later stderr-wording failure is pre-existing in this environment on every commit above.
The `1 -> 0` exit-status change is introduced exactly by `c150076`.

Mechanism, isolated to one line and then reduced to a minimal reproduction. Disabling only
`trap teardown_lifecycle_lock_release EXIT` at `c150076` restores `exit 1`, with the
`unresolvable-lock` mode unchanged at `exit 1` in both:

| c150076 | `missing-adapter` | `unresolvable-lock` |
| --- | --- | --- |
| with the trap | **0** | 1 |
| trap disabled | 1 | 1 |

The cause is not the handler's body. Under `set -e`, a `source` of a MISSING file is a fatal shell
abort, and that abort reaches an EXIT trap with `$?` **already 0**:

```
set -eu; cleanup() { echo "STATUS-INSIDE-TRAP=$?"; }; trap cleanup EXIT; . /nonexistent/adapter.sh
  -> STATUS-INSIDE-TRAP=0, script exits 0
same script with no trap installed
  -> script exits 1
```

So merely installing an EXIT trap converts this fatal abort from `exit 1` into `exit 0`, whatever the
handler does. Ordinary `exit 1` refusals are unaffected: they arrive at the trap with `$?`=1 and exit 1.

Severity: teardown still **refuses correctly and destroys nothing**. Probed directly at the point of
refusal on the merged tree: `worktree_exists=YES branch=fm/task-x1 meta=YES status=YES`. The defect
is that the refusal is invisible to the caller - a refusal that reports success. That matters because
`AGENTS.md` treats a teardown refusal as a stop-and-investigate result, and `fm-latent.sh finish`
`exec`s into `fm-teardown.sh` and would read the masked status.

### Reachable beyond teardown refusal

The hibernation commit adds **two** whole-script EXIT traps: `fm-teardown.sh:190` and
`fm-latent.sh:78`. Both scripts run under `set -eu`, and both reach `fm-backend.sh`'s LAZY adapter
source (`fm-backend.sh:609`) after their trap is installed, so both can hit the fatal-abort class:

- `bin/fm-teardown.sh` - proven end to end by the repository's own test (`missing-adapter`, `rc=0`).
- `bin/fm-latent.sh` - `fm-latent.sh:503` (`fm_backend_source herdr`) sits on the `enter` path, after
  the trap. Reproduced against the REAL `bin/fm-backend.sh` with `bin/backends/herdr.sh` removed:
  the script's own construct exits **0** with the trap and **1** without it.

That second one is the more consequential consumer, because `bin/fm-watch.sh:723` BRANCHES on the
status (`if fm-latent.sh enter "$id" ...; then`), so a fatal missing-adapter failure would be read as
a successful hibernation. `bin/fm-pr-check.sh:134` calls `enter` with `|| true` and is unaffected
either way.

Scope is bounded, and deliberately verified rather than assumed: this affects ONLY the
fatal-source-abort class. Ordinary safety refusals - dirty worktree, unlanded work, latent integrity,
lock contention - use explicit `exit 1` (46 sites in teardown) and still exit non-zero with the trap
installed, as the `unresolvable-lock` column above shows.

No handler-side rescue of `$?` works, because the status is gone before the handler runs: `return $rc`,
`exit $rc`, and a bare `:` handler all still produce `exit 0`.

### The fix

The only evidence that survives a fatal abort is that the script never reached a completion point, so
each script now marks its legitimate terminations and its handler treats a `$?=0` arrival WITHOUT
that mark as an abort, exiting 1. The lock-release work the traps exist for is unchanged, and
ordinary refusals are untouched - they carry their own non-zero `$?` straight through.

Teardown's abort message is conditional on whether teardown had begun changing the task, so it can
never claim "nothing was changed" once something might have been. That flag is set EARLY, at the
point teardown commits to its destructive phase: an over-eager flag only makes the message more
conservative, while a late one would let it overclaim.

### Watched failing first, on the real scripts

Both captured unpiped, before and after, driving the real code rather than a synthetic shape:

| Script | Trigger | Before | After |
| --- | --- | --- | --- |
| `bin/fm-latent.sh` | real `enter` path, lazily sourced adapter removed | **rc=0** - `not ok - fm-latent.sh enter reported success (rc=0) after a fatal abort` (suite exit 1) | **rc!=0** - `ok - fm-latent.sh preserves a non-zero exit status through its EXIT trap on a fatal abort` (suite exit 0) |
| `bin/fm-teardown.sh` | repo's own `herdr-preflight-missing-adapter` case | **rc=0** - `not ok - ... teardown continued without its required preflight` (suite exit 1) | **rc!=0** - suite exit 0, 57 assertions passing, 0 failures |

The new `tests/fm-latent.test.sh` case asserts the abort ACTUALLY fired (`grep -q "No such file or
directory"`) before asserting the status, so it cannot pass vacuously if the fixture stops triggering
the abort.

### Suite result after the fix, by CI lane

The monolithic `--all` run was repeatedly killed by the environment partway through, so the suite was
re-run as the six lanes CI actually composes. Every summary line is verbatim:

| Lane | Result |
| --- | --- |
| `portable-parallel-1` | `FM_TEST_SUMMARY total=11 failed=0 skipped_gate=0` |
| `portable-parallel-2` | `FM_TEST_SUMMARY total=13 failed=0 skipped_gate=0` |
| `portable-serial-1of4` | `FM_TEST_SUMMARY total=21 failed=0 skipped_gate=2` |
| `portable-serial-2of4` | `FM_TEST_SUMMARY total=26 failed=4 skipped_gate=3` |
| `portable-serial-3of4` | `FM_TEST_SUMMARY total=25 failed=2 skipped_gate=5` |
| `portable-serial-4of4` | `FM_TEST_SUMMARY total=26 failed=2 skipped_gate=5` |

122 tests, 8 failures, and all 8 are the SAME pre-existing environment failures baselined against
clean `main` earlier: `fm-calm-pi-extension`, `fm-remote-secondmate-trace-context`,
`fm-remote-secondmate-lifecycle-e2e`, `fm-muse-harness`, `fm-kimi-harness`, `fm-backend-orca`,
`fm-gotmp`, and `fm-backend-herdr-focus-flash-e2e`.

Against the pre-fix `--all` run (`total=133 failed=10`) the delta is exactly one failure REMOVED -
the teardown preflight case this commit fixes - and nothing added. `grep` for
"continued without its required preflight" across every lane log returns 0 occurrences. The
remaining `real-herdr-gated` lane is opt-in and needs Herdr >= 0.8.0; this host has 0.7.5.

Regression check that the fix does not invent failures: `fm-gotmp` still fails exactly as it does on
clean `main`, and its output contains none of the new abort messages, so that pre-existing failure is
untouched. `fm-bootstrap` + `fm-crew-state` (78 assertions) and `fm-session-start` +
`fm-startup-network` (59 assertions) are green, `bin/fm-lint.sh` passes at CI parity, and the
coverage guard reports `total=133`.

## CI

CI did not run on the earlier revision. This is a fork pull request, so workflow runs require maintainer approval. An empty check set is not being reported as passing checks.


kunchenguid:

<sub>Noted for firstmate#1862 at `c1500766`.</sub>

PR #1870, opened 2026-08-07
Fix agy effort double-passing in fm-spawn.sh

prajwal-395:
Fixes a bug where agy was being passed an --effort flag when it was already baked into the display-name model.

### DIFF
--- a/bin/fm-spawn.sh
+++ b/bin/fm-spawn.sh
@@ -1056,9 +1056,21 @@ model_flag_for_harness() {
 }
 
 effort_flag_for_harness() {
-  local harness=$1 effort=$2
+  local harness=$1 effort=$2 model=${3:-}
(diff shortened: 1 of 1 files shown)


prajwal-395:
Closing as superseded.

This fix has been folded into the consolidated agy harness adapter branch, where it is implemented more completely: rather than inspecting the effort value alone, the resolved model is threaded into `effort_flag_for_harness` so the `--effort` flag is skipped whenever the model name already carries its own effort suffix (agy's display names such as `Gemini 3.1 Pro (High)` bake it in, and passing both makes agy silently fall back to its default model).

Kebab-case model ids still receive `--effort`, and `agy:max` is still rejected as unsupported.

Superseded by the equivalent commit on `fm/agy-harness-adapter`; no longer needed as a standalone change.

PR #1930, opened 2026-08-08
fix(crosscheck): let the merge gate reach a clear verdict

ruby-dlee:
## Intent

Fix the crosscheck merge gate so it can produce a CLEAR verdict. It could RECORD findings but never CLEAR one, so no PR in the fleet could pass its merge gate.

**This change fixes the gate that reviews it.** Crosscheck cannot meaningfully validate this PR, because the machinery under review is the machinery that would do the reviewing. That circularity is stated here deliberately rather than left to be discovered — the bootstrap problem is the captain's to resolve at merge time.

## Validation, and how it was reached

**The local no-mistakes pipeline was bypassed for the final push.** CI plus independent review are the validation for this PR. That is a deliberate, captain-authorised decision, not an omission — three other lanes hit the same gate-custody divergence tonight and took the same route. Nothing was reset, forced, or rewritten; the branch is a clean fast-forward of ten commits.

What the pipeline *did* complete before it was bypassed:

- **Review: completed clean.** Five earlier review rounds found six real defects — five of them in code written during this task — and all were fixed and re-reviewed. The final run's review produced three auto-fix findings and nothing requiring a decision.
- **`tests/fm-crosscheck.test.sh`: 40 → 60 cases, green on Python 3.14.6 and 3.9.6.** Every new test was confirmed to fail against pre-fix code first.
- **Test step: red on contention, proven not to be this change.** The failing files are byte-identical to `origin/main`, unreachable from this diff, and clean `origin/main` at `4b6beeaa` runs them standalone at 46 ok / 0 not ok. All eight failing suites pass standalone with zero failures. Details below.

## The gate could pass a fix it never tested

The most important defect here was not in the brief and is a **safety** problem rather than an availability one.

A mutation proof clears a finding by showing the named test passes before a mutation and fails after. The gate read "fails after" as simply `exit != 0` — but a run that never executed the test also exits nonzero. So a mutation that broke the runner rather than the behaviour was recorded as "the test caught the regression," and the ledger stored a durable `verified-fixed` proof for a regression nothing had detected. **The gate could certify a fix on the strength of a test that never ran.**

This is not hard to hit: mutating implementation is exactly what breaks collection, and a mutated Python module that no longer imports takes its whole test file down with it.

Closing it makes the gate **stricter**. Some proofs that would have cleared before are now refused, which is correct — they were never evidence.

## One bypass, four doors

Closing that hole revealed that the same forged-verdict could arrive through several channels, each found only because the previous one was closed:

1. **A positional second test target.** `test_invocation.arguments` were spliced into argv, so a vacuous named test plus `arguments: ["tests/other_test.py"]` cleared a finding on a file the gate never validated — routing around the tracked check, the symlink check, and the patch guard at once.
2. **A flag altering exit semantics.** `--continue-on-collection-errors` converts a collection failure into a plain test-failure status.
3. **An ambient environment variable.** `PYTEST_ADDOPTS` reproduced the same flip with no reviewer involvement at all — any developer with it exported silently turned the gate from refusing to approving.
4. **An ancestor config file.** `pytest` walks parents to the filesystem root, so an operator `pytest.ini` above the proof checkout did the same.

Each was measured, not reasoned about. Resolutions are absolute rather than enumerated: proofs take **no runner arguments**, run under an **explicit minimal environment allow-list**, and execute beneath a **gate-owned neutral `pytest.ini`**. No allow-list of flags and no guessed exit codes, because an enumeration is only as good as the cases someone thought of.

**`verified-fixed` is restricted to runners whose non-execution signal is classified — `pytest` only.** Shell runners were measured and *proven* indistinguishable: a mutation deleting a sourced implementation, and an implementation exiting during load, both collide with a genuine catch at exit 1, and a vacuous test clears under both. The structural argument that shell was safe read persuasively and was false; only measurement caught it. Accepted cost: this repo cannot clear a *finding* by mutation proof, though a review that finds nothing still clears normally.

## Also fixed

- **The gate punished reviewers for producing evidence.** The post-review integrity check listed every untracked file through a 200,000-byte limit, so its cost scaled with how much evidence a reviewer wrote. A review that found nothing completed; one that *substantiated* a finding was refused. Now `--untracked-files=normal`, with its own budget so heavy unauthorised scratch is refused by name rather than as a bare output-limit error. Detection is unchanged and separately guarded.
- **Failures to execute were reported as verdicts about the code** — an absent runner, an unmatched selector, an unfound command, and a helper needing reviewer-only environment all exit nonzero and were each reported as "the test does not pass." Refusals now name the real cause and carry the command's own output.
- **pytest node ids** (`file.py::test`) were treated as filesystem paths and could never resolve.
- **A lexical symlink check** made any home under a symlinked ancestor permanently unable to clear.
- **`Path.stat(follow_symlinks=…)` is Python 3.10+** while `fm-crosscheck.sh` execs ambient `python3`, so evidence capture crashed on an older interpreter.

## Secondary, separable: `bin/fm-lint.sh` batching

`bin/fm-lint.sh` passed **203 tracked scripts — not the 186 the original task records; the defect grew** — to a single ShellCheck invocation, and printed nothing between its version banner and its exit.

**That is why a healthy lint step is indistinguishable from a dead one, and why two lanes died tonight looking alive.** A step doing exactly what it should produces no output for its entire runtime, so silence gets read as a wedge.

Now lints in batches of 25 with a progress line before each, so the quiet gap is one batch rather than the whole run. `-x` is required for batching, not a relaxation: 82 of these scripts source a sibling, and without it a sourced file landing in another batch would raise SC1091 the single invocation never saw.

Measured: exit 0, zero findings matching the unbatched result, 526s at host load 128.

**Peak memory is not fixed by this, and the original defect's framing is wrong.** Cost is superlinear and file-specific, not proportional to argument-list length:

| Script | Size | Peak RSS |
| --- | --- | --- |
| `tests/fm-account-routing-suite.sh` | 338K | **4.65 GB** |
| `bin/fm-teardown.sh` | 232K | 2.38 GB |
| `tests/fm-teardown-suite.sh` | 286K | 0.54 GB |

One script alone exceeds any batch bound, and the batched run's 4.64 GB peak is exactly that single-file floor. Batching stops peaks stacking and stops them growing with script count; going below the floor needs those two scripts addressed. Filed separately.

## Test-step failures are contention, established not assumed

The sharded behaviour suite failed under load. That was proven environmental rather than caused by this change:

- The failing files (`tests/fm-checkout-refresh.test.sh`, `bin/fm-checkout-refresh.sh`, `bin/fm-bootstrap.sh`) are **byte-identical** to `origin/main`.
- Neither script references anything this branch changed — zero hits for `fm_bounded_io` or `fm-crosscheck`.
- Clean `origin/main` at `4b6beeaa`, standalone: **exit 0, 46 ok, 0 not ok**.
- **All eight failing suites pass standalone with zero failures**: `fm-watcher-lock` 30s, `fm-bearings-snapshot` 32s, `fm-report-stack-b` 210s, `fm-report-stack-a` 255s, `fm-backend-herdr` 322s, `fm-account-routing-a` 715s, `fm-teardown-b` 2562s, `fm-teardown-a` 2609s.

Six of the eight passed at load 27–33, so the discriminator is the *sharded shape*, not raw load. And `fm-teardown-a`/`-b` take ~43 minutes **each** standalone, so the suite set is independently over the 30-minute bar: sharding is load-bearing, and sharding is what oversubscribes the host when lanes overlap. Filed as its own task.

## Follow-ups filed

- **Positive proof of execution** — the durable fix for all four doors above. The gate infers "the test detected the regression" from an exit status that reviewer input, ambient environment, repository config, and runner version can all influence. Closing channels one at a time is unbounded work with no completion criterion. Require the mutated run to *demonstrate* the test ran instead — a control test the mutation should not affect, required to pass while the named test fails. Runner-agnostic; makes the runner restriction here removable.
- **Behaviour suite only passes on a quiet machine**, with the timings above.
- **The shard coverage guard validates the wrong property** — `--check 8` verifies every script is in the plan but never that a recorded weight still describes it, so the balancer computes on data that silently rots. This suite's weight was re-measured here (75707 → 94232 ms, measured and scaled into the file's GitHub Actions units, provenance in the TSV header).
- **The single-file ShellCheck memory floor.**

## The pattern worth keeping

Every defect here is the same shape: **something confidently computing on a fact that stopped being true.** The gate read an exit status that no longer meant what it assumed. The balancer read a duration that no longer described the file. Supervisors read silence as death from commands designed to be silent. None of them failed loudly; each produced a wrong answer with full confidence.

Every one was found by measuring. Two persuasive structural arguments — one mine, one a reviewer's — were both wrong.


ruby-dlee:
Opened against the wrong repository by mistake. This work belongs to ruby-dlee/firstmate, not this upstream. Closing; no action needed here. Apologies for the noise.

PR #1932, opened 2026-08-08
fix(bin): resolve session-lock ownership against the whole harness run

brycehowardd:
## Intent

`fm_session_lock_owned_by_self` refused a firstmate session the lock its own harness run already held.
Two defects combined:

1. `basename` was called on a raw process name. macOS login shells report as `-zsh`, so `basename -- `
   was missing and `basename` parsed the leading dash as a flag: `basename: illegal option -- z`.
2. The ancestry walk then failed open on that error and concluded the lock belonged to a foreign
   holder, so a session was refused a lock it legitimately owned - observed 2026-08-05 when the lock
   was refused to its own background-spare descendant, while `fm_session_lock_owned_by_self` returned
   YES for the same process chain when run directly.

## Change

- Process names are read without `basename`, so a login-shell name can never be parsed as a flag.
- Lock ownership resolves against the whole harness run rather than a single pid, and the lock pid is
  guarded before the ancestry walk.
- Regression tests cover both the self-held background-spare chain and the foreign-holder boundary, so
  the fix cannot silently regress into either failing-closed on itself or failing-open to a stranger.

## Verification

Review, test, document and lint all passed in the validation pipeline. 150 lines of new regression
test in `tests/fm-session-lock-ancestry.test.sh`.

## Note on timing

This was finished and validated on 2026-08-05 and then parked purely on publishing: the account holds
read-only access to the parent repository and the fork this setup expects did not exist yet. The fork
now exists, so this is the same validated work published unchanged - tip `4b3e97c`, no rebase, no new
commits.


kunchenguid:

PR #1935, opened 2026-08-08
fix: make Claude supervision start on Windows (session lock, worker placement, turn-end guard, detached auto-arm)

bingb0t5:
Four independent Windows faults that together meant firstmate could never dispatch a worker or start supervision on a Git Bash host. Each is small, each has a regression test, and none of them touch the files in #1790 or #1791 - this stacks cleanly alongside both.

Found and verified on a real Windows home (Git Bash / MSYS, herdr backend, Claude Code 2.1.224). Before this series no spawn had ever succeeded on that machine and `state/.claude-autoarm-epoch` had never been written; after it, two crewmates ran concurrently in separate treehouse lanes and the Stop-owned auto-arm claimed, armed, woke and tore down unattended.

### 1. `fix(session-lock)` - read harness liveness from the process table

`fm_harness_pid_alive` gated on `kill -0` before consulting `ps`. On POSIX the two agree, so the extra gate was invisible. It is not invisible where the harness pid lives outside the shell's signal namespace: a native Windows harness is addressed by an MSYS-offset pid (`0x400000 + winpid`), `ps` resolves it, and `kill` is a bash builtin so no PATH shim can reach it. The lock owner failed liveness permanently while ownership stayed valid.

```
before: lock: stale (pid 4225416 dead or not a harness)
after:  lock: held by live harness pid 4225416      (4225416 - 0x400000 = 31112, a running claude.exe)
```

Existence now comes from the process table, which this predicate had to consult anyway to decide harness identity, so nothing is relaxed.

Also unsets `BASH_ENV`/`ENV` for the test suite. bash sources `$BASH_ENV` in every non-interactive shell, so a developer's startup script runs inside each `bash -c` fixture *after* the test has set its environment. One that reorders PATH puts its own directory back in front of `fm_fakebin`, silently defeating every stub - five unit cases in `fm-session-lock-ancestry` were failing that way, answering from the real host.

### 2. `fix(spawn)` - lease the task worktree when the pane reports no live path

Worker placement typed `treehouse get` into the pane and polled the pane's live foreground cwd. Herdr on Windows exposes no live cwd from any surface: `pane get` has no `foreground_cwd`, and `pane process-info` reports only creation-time cwd and never sees child processes (verified by moving a real pane and watching it not update). The poll could never succeed, so every spawn died on the 60s timeout.

Probes the capability against the fresh pane before sending anything. A backend that reports a live path keeps the existing handshake unchanged; one that cannot falls back to `treehouse get --lease`, which reserves a worktree without opening a subshell and prints its absolute path, so the worktree identity is known rather than inferred.

### 3. `fix(turnend-guard)` - give the auto-arm sync wait an MSYS default

The guard waits `FM_CLAUDE_AUTOARM_SYNC_WAIT_MS` for the auto-arm to claim. 800ms is not a budget on MSYS, where the auto-arm clears several process-table gates first:

```
fm_primary_scope_matches         1253 ms
fm_session_lock_owned_by_self   12296 ms
owner lock acquired             20449 ms
epoch recorded                  21170 ms
```

Now 35000 on MSYS, matching `FM_OPENCODE_ARM_READY_TIMEOUT_MS` and `FM_PI_ARM_READY_TIMEOUT_MS`, which already carry this Windows default for the same reason. The guard returns the moment either proof appears, so it is a ceiling, not a per-turn cost. POSIX unchanged at 800ms, including the malformed-value fallback (which previously reset to a hardcoded 800 and would have silently downgraded MSYS).

### 4. `fix(autoarm)` - claim the home when a detached hook cannot see its ancestry

The root cause of "supervision never starts". `asyncRewake` implies `async`, so Claude Code detaches the hook and it is reparented to init. `fm_harness_ancestry_pids` walks up from `$$`, hits ppid 1 on its first hop, and never reaches the harness that spawned it:

```
ordinary child : ancestry [4225416]  owned_by_self YES
orphan (ppid=1): ancestry []         owned_by_self NO
```

So the hook read its own home's live lock, failed to recognise itself, took it for a foreign live session, and exited 0 - every time. No epoch was ever written, no failure was ever recorded, and the bounded attended fail-open (which requires a recorded failure) was unreachable.

Splits the two cases. A *resolvable* ancestry that omits the lock pid is still positive evidence of a competitor and stays fatal. An *unresolvable* ancestry falls back to what a detached hook does have: the genuine primary checkout (already proven by `fm_primary_scope_matches`) plus a live harness on the lock. The pre-existing single-flight owner lock keeps that safe - if two sessions on one home both reached there, exactly one acquires the claim.

Worth noting for review: every pre-existing case in `fm-claude-stop-autoarm.test.sh` runs the hook as a **child** of the fake harness, where ancestry always resolves - the one shape Claude Code never uses - so the suite could not observe this. The two new cases run it genuinely orphaned, and settle the fixture harness first (a freshly spawned pid does not immediately read as live where `ps` resolves through a cached Windows process table).

### Tests

- `fm-claude-stop-autoarm`: 24 pass, 0 fail (baseline 22)
- `fm-backend-herdr`: 169 pass, 0 fail, unchanged
- `fm-session-lock-ancestry`: unit layer 5 pass, 0 fail (was failing its first case)
- `fm-turnend-guard`: new platform-default case passes on both branches

Known-open on that host, pre-existing and confirmed identical on the parent commit, so not introduced here: the `fm-session-lock-ancestry` e2e layer (MSYS `fork()` re-exec through `CreateProcessW`), two `fm-turnend-guard` cases that set `PATH=<fakebin only>` (Windows resolves `msys-2.0.dll` via PATH, so the binary cannot load), and `fm-teardown` case 25 (`could not prepare PR poll`).

Remaining limit, unchanged: a detached hook still cannot reclaim a genuinely *stale* lock, since `fm-lock.sh` needs the same ancestry to write a new owner. Session start takes the lock, so the recorded owner is normally live.

https://claude.ai/code/session_01TSrXFoUSoCdCKC9JhcA1QS


kunchenguid:

Issue #1924, opened 2026-08-07
Worktree pool hands over and resets an idle worker's slot before anything downstream can object


jjaguirr:
## Root cause

The worktree pool decides a slot is in use by detecting live processes whose working directory is inside it. An agent that is thinking, waiting on the network, or simply idle between commands has no such process, so its own worktree reads as free and the next `treehouse get` hands the slot over.

The handout itself performs the reset: `pool.Acquire` calls into the pool's acquire path, which resets the worktree to a clean detached HEAD in the same function that probes for live processes. By the time any caller can inspect the result, the reset has already happened.

Observed twice on 2026-08-07 against `persea`: slot 2 was handed out on two consecutive spawns while a live worker owned it, while slot 6 sat available both times. Not load-related and not random - it recurs for any idle worker.

## Why this is filed separately

`bin/fm-spawn.sh` now refuses to launch a second agent into a worktree that another live task in this home records as its own. That guard prevents the second half of the damage: two agents operating in one checkout, and the displaced worker ending up somewhere it should never be.

It cannot prevent the first half. The guard runs after `treehouse get` returns, so the slot has already been handed over and reset. Fixing that requires the pool to stop treating process presence as the ownership signal, which is upstream of firstmate.

## Suggested direction

The supported boundary already exists: `treehouse get --lease --lease-holder <id>`. Firstmate already uses leases in `bin/fm-home-seed.sh`, capability-gates them in `bin/fm-bootstrap.sh`, and releases them at teardown. Extending lease usage to ordinary task spawns would make ownership explicit rather than inferred from process presence, so an idle worker's slot is never offered in the first place.

angelescala:
Another occurrence on 2026-08-08, in a secondmate home on the herdr backend with codex crews. Same slot handover and reset, but with two details that extend the report above.

**What happened.** A ship task had been working in pool slot 1 of its project, committed two documentation commits to its own branch, and then deliberately parked on an external dependency (`paused:`, waiting on a third-party support answer). Some time later a scout task was spawned for the same project. `treehouse get` handed out slot 1 again: both tasks' `state/<id>.meta` recorded the identical `worktree=` path, and the slot's reflog shows the handover reset in the expected place:

```
a153b262 HEAD@{0}: checkout: moving from fm/<ship-task> to refs/remotes/origin/<default-branch>
462de790 HEAD@{1}: commit: docs: <second commit>
da214875 HEAD@{2}: commit: docs: <first commit>
```

[...]

Issue #1913, opened 2026-08-07
fm-spawn leaves an orphan tmux window when a worktree validation refuses


jjaguirr:
## What happens

`validate_spawn_worktree` in `bin/fm-spawn.sh` refuses a spawn after the task's tmux window has already been created by `fm_backend_tmux_create_task` and after `treehouse get` has run. `spawn_abort_cleanup` cleans up herdr projections and orca worktrees, but never a tmux window.

The window `firstmate:fm-<id>` therefore survives every refusal, holding a shell whose working directory is inside the worktree that was refused.

An operator who acts on the refusal and spawns the same task again hits `error: window firstmate:fm-<id> already exists` from `bin/backends/tmux.sh`. `bin/fm-teardown.sh <id>` does not help either, because no metadata was ever written for the refused spawn. The window has to be removed by hand with `tmux kill-window`.

## Why this is filed separately

This is pre-existing behaviour, not new. The primary-checkout isolation refusal that `validate_spawn_worktree` has always performed exits on exactly the same path and leaves exactly the same orphan window.

The live-sibling worktree guard added alongside this issue makes the path more common, so the guard's own diagnostic was corrected to name the manual `tmux kill-window` step. Extending `spawn_abort_cleanup` to remove the window is the real fix and belongs in its own change, scoped to every refusal on that exit path rather than to one caller.

## Suggested shape

Extend `spawn_abort_cleanup` so a tmux spawn removes the window it created when the spawn aborts before metadata is written, matching what the herdr and orca branches already do. Once that lands, the manual `tmux kill-window` instruction can come back out of the refusal diagnostics.

kunchenguid:
Speaking as Kun's firstmate:


Verified against main `f7a387f44ed4fa942914e8944be4981a42e1e102`.

**VISION:** aligns — Scripts own cleanup mechanics; a refusal path should leave the fleet ready to retry without hand surgery. Herdr projections and Orca worktrees already abort-clean; tmux windows created before metadata is written should too. An orphan `firstmate:fm-<id>` window that `fm-teardown` cannot see (no meta) is a crack the captain should not have to know about.

[...]

PR #1942, opened 2026-08-08
fix(spawn): refuse a worktree a live sibling already records

jjaguirr:
## The bug

The worktree pool decides a slot is in use by detecting live processes whose working directory is inside it. An agent that is thinking, waiting on the network, or simply idle between commands has no such process, so its own worktree reads as free and the next `treehouse get` hands the slot over. The pool then resets it to a clean detached HEAD, destroying the running agent's branch checkout.

Reproduced twice on 2026-08-07 against `persea`: slot 2 was handed out on two consecutive spawns while a live worker owned it, while slot 6 sat available both times. Not load-related and not random - it recurs for any idle worker. Nothing was lost only because the collision was caught before the new agent ran `git checkout -b`.

## The change

`validate_spawn_worktree` already refused a worktree that is not isolated from the primary checkout. This extends that same validation point: a spawn now also refuses when the acquired worktree is already recorded by another live task in this home, which firstmate knows from each task's `worktree=` in `state/<id>.meta`.

The upstream in-use detection is not touched. This guard is the correct layer because firstmate owns the metadata that knows the truth, and the pool cannot see an idle agent at all.

- Physical paths are compared on both sides, so a symlinked, trailing-slash, or otherwise differently-spelled path still matches.
- The task's own record is excluded, so recovery and re-seating still relaunch into their own recorded worktree.
- Liveness comes from `fm_backend_agent_state` rather than a second liveness notion.

## What the guard does and does not prevent

On the treehouse path the guard runs after `treehouse get` returns, and the pool performs its reset inside that call. So by the time the guard can object, the slot has already been handed over and reset.

What it prevents is the second half of the damage, which is the half that hurt in both observed incidents: a second agent taking over a checkout that another worker is using, and the displaced worker ending up somewhere it should never be.

The refusal message says so plainly and tells the operator not to assume the sibling's uncommitted work is intact. The root cause lives in the pool and is tracked separately in kunchenguid/firstmate#1924.

## Ambiguous liveness: why anything other than dead or missing refuses

Only `dead` and `missing` release a claim. `alive`, `ambiguous`, `unreadable`, and `unverified` all refuse.

- It matches this file's existing duplicate-launch guard, which already treats an unreadable endpoint exactly like a live one.
- The errors are asymmetric. A wrong refusal costs one spawn, and the diagnostic names the exact record to clear. Wrongly proceeding resets a live worker's branch checkout out from under it, which is the bug being fixed.
- The refusal is narrow. It fires only for a spawn that lands on that one recorded path, so an unreadable record stalls reuse of its own slot rather than the fleet, and `bin/fm-teardown.sh` releases the record.

The operational consequence is that `zellij`, `cmux`, and `orca` siblings report `unverified`, so a stale record from one of those holds its pooled slot until an operator tears it down. That is the deliberate cost of the direction above.

## Orphan window on refusal

A refusal exits after the task window was created, and the abort cleanup does not remove a tmux window. The refusal diagnostic therefore names the manual `tmux kill-window` step so its stated remedy actually works.

The window itself is left alone on purpose. This is pre-existing behaviour: the primary-checkout isolation refusal exits on the same path and leaves the same orphan. Cleaning it up belongs in its own change scoped to every refusal on that path rather than to one caller, and is tracked in kunchenguid/firstmate#1913.

## Orca abort cleanup

The orca branch sets `ORCA_ABORT_CLEANUP=1` before validation runs, so a refusal used to reach `spawn_abort_cleanup` and destroy the very resources the guard had just proved a live sibling owns. That ordering was safe before this change, because the only previous refusal reason was a worktree identical to the primary checkout, which orca never creates. This change adds a refusal reason where those resources belong to somebody else.

The cleanup destroys two things: it closes `ORCA_TERMINAL` and it removes `ORCA_WORKTREE_ID`. Both are at risk on a refusal, because `fm_backend_orca_worktree_create` returns a `worktree-terminal-handle` alongside the worktree and `parse_orca_worktree_result` stores it before the guard runs, so that handle can be the sibling's own terminal exactly as the path can be its worktree.

One condition (`ORCA_ABORT_SIBLING_OWNED`) therefore gates the whole block rather than the two statements separately, so a resource added there later is covered without rediscovering this. Every other abort still destroys both resources exactly as before.

## Tests

Colocated in `tests/fm-spawn-sibling-worktree.test.sh`, following the existing `fm-spawn` conventions and the fake-backend plus real-git-worktree harness already used by `tests/fm-spawn-worktree-settle.test.sh`.

Covered: a clean spawn still succeeds; a live sibling refuses; a dead sibling still spawns; a task's own record does not refuse its relaunch; an endpointless sibling record refuses with a diagnostic instead of dying silently; an unreadable endpoint refuses; differently-spelled paths resolving to the same directory are caught; and an orca refusal leaves both the sibling's worktree and the sibling's terminal intact.

The two orca cases drive a real `--backend orca` refusal rather than asserting on source text. The fake `orca worktree rm` deletes the directory and the fake `orca terminal close` removes a sentinel file, so neither assertion can be satisfied by a missing log line alone.

Each guard case was mutation-checked: disabling the guard fails the refusal cases, disabling the own-id exclusion fails the relaunch case, reverting the errexit fix fails the endpointless case with the exact silent-death shape, and disabling the orca trap gate fails the worktree case and, checked in isolation, the terminal case with "the orca refusal closed the live sibling's terminal".

## What is not covered

- The real race is not staged. Reproducing it needs an idle agent, so the guard is tested against constructed `state/<id>.meta` records rather than by racing a live pool. Testing the pool directly would have meant a weaker test that passes for the wrong reason.
- Whether orca's `worktree create` can return an already-existing worktree was not confirmed against a live orca. The suppression does not depend on that question, since it gates on the refusal rather than on how the resource was obtained, but it is the reason the terminal case is stated as a real risk rather than a hypothetical: the handle reaches `ORCA_TERMINAL` from the create result either way.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -283,7 +283,7 @@ Write the task-specific brief under section 11 before spawning.
 ### Dispatch and supervision handoff
 
 Spawn only through `bin/fm-spawn.sh` after the profile and backend checks in section 4.
-The spawn must resolve a genuine isolated task worktree distinct from the primary checkout; a failed isolation assertion stops the task.
+The spawn must resolve a genuine isolated task worktree distinct from the primary checkout and unclaimed by another live task in this home; a failed isolation assertion stops the task.
 After spawning, confirm the worker is processing the brief, handle any trust dialog through `harness-adapters`, and record ship or scout work as under way.
 A persistent secondmate is recorded in the secondmate registry and runtime state, never as a backlog work item.
 
--- a/bin/fm-spawn.sh
+++ b/bin/fm-spawn.sh
@@ -111,7 +111,10 @@
 #   Before a secondmate launch, the home is locally fast-forwarded to the primary
 #   default-branch commit when safe; skipped syncs warn and launch unchanged.
 #   Ship/scout spawns refuse to launch unless the resolved task path is a real
-#   git worktree root distinct from the primary project checkout.
+#   git worktree root distinct from the primary project checkout, and unless no
+#   other task in this home records that same path as its worktree with a
+#   still-live endpoint - the worktree pool cannot see an idle agent, so only
+#   an authoritatively dead or missing endpoint releases a recorded slot.
 # Batch dispatch: pass one or more `id=repo` pairs instead of a single <id> <project>, e.g.
 #     fm-spawn.sh fix-a-k3=projects/foo add-b-q7=projects/bar [--scout]
 #   Each pair re-execs this script in single-task mode, so the single path stays the only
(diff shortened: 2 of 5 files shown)


jjaguirr:
Closing: this was contributed from a fork because the author has read-only access here, and on reflection that is not the right route for it. The work stays on jjaguirr/firstmate at branch fm/fm-spawn-sibling-worktree-guard if anyone wants it - full pipeline passed locally (review, tests including 8 new guard cases, documentation, lint). The pool-level root cause is separately recorded as #1924.

PR #1945, opened 2026-08-08
feat(bin): scaffold jj bookmark branch step for jj-managed projects

swissgrammie:
Enables Jujutsu (jj) support in firstmate: ship briefs scaffold the jj bookmark branch step when a project is jj-marked.

Takes effect only when jj has been selected for a project and jjhouse is installed. All other projects (treehouse/git) keep the existing workflow, byte-identical.

The generated jj brief makes the worker verify jj/jjhouse in its own environment before creating the bookmark and reports `blocked: jj tooling missing in worker environment` if either is missing, because the worker pane may not inherit firstmate's PATH. The scaffold itself refuses to emit a jj step when the tooling is absent from the host PATH (a jj-managed project forbids raw git writes, so there is no runnable fallback). The registry scan reads the mode bracket only at field 3, so description text like `[v2] jj migration` can never flip a legacy row to jj-managed.

Fixes the Greptile review findings on this PR.


swissgrammie:
Thanks for the review. I verified each finding against the current branch
head and addressed the two that were still open; the other two were already
fixed on the branch after your comments were filed.
1. Gate checks the wrong environment (P1) - fixed. The generated brief's
   first action now makes the worker verify `command -v jj && command -v
   jjhouse` in its own environment and report `blocked: jj tooling missing
   in worker environment` if either is missing, because the worker pane does
   not inherit firstmate's PATH. The scaffold-side check remains as a
   fast-fail.
2. Fallback violates jj project policy (P1) - already fixed on the branch
   (commit after the review): the no-tooling path now refuses with exit 1
   and emits no brief; there is no git fallback. Covered by
   test_jj_token_without_tooling_refuses_scaffold.
3. Missing jj availability gate (P1) - same root cause as #1; the worker
   now verifies the tooling in its own environment, and the scaffold-side
   check covers the host.
4. Description text mimics jj annotation (P2) - already fixed on the branch:
   the registry scan only reads the mode bracket when it opens at field 3
   and stops at the bracket close, so `- repo - handles [v2] jj migration`
   is not misclassified. A regression test pins this exact row.

PR #1661, opened 2026-08-04
fix: prevent false watcher-down alarms in Claude sessions

karotkriss:
## Intent

Fix bin/fm-guard.sh reporting a false "WATCHER DOWN - SUPERVISION IS OFF" alarm during a healthy Claude-primary session.

Root cause: the mid-turn pull guard derived watcher health from fm_watcher_healthy, which requires a live watcher PROCESS holding the home lock. Under Claude's Stop-hook auto-arm supervision model the watcher is armed at each turn end and exits on its wake, so it runs only BETWEEN turns. Every guarded command run mid-turn therefore found no live watcher and tripped the banner even though supervision was healthy. Because the dedup episode key was derived from the beacon mtime (which the between-turns watcher advances every poll) the full banner re-printed on essentially every command, and the message always blamed a "fresh beacon" that was in fact fresh.

Deliberate decisions and tradeoffs:
- Added a new model-aware verdict fm_watcher_supervision_verdict (and fm_supervision_model) in bin/fm-wake-lib.sh and switched ONLY the mid-turn pull guard (bin/fm-guard.sh) to it. fm_watcher_healthy itself is intentionally left unchanged and still PID-strict, because the arm layer (fm-watch-arm.sh, fm-claude-stop-autoarm.sh) needs a live process to decide start/attach/replace and must never be satisfied by a leftover beacon.
- The turn-end guard (fm-turnend-guard.sh) intentionally keeps the strict fm_watcher_healthy check: it fires at the turn boundary where the auto-arm brings a fresh watcher up and cooperates with that arm, so a beacon left by the just-ended cycle is not sufficient there.
- Model detection: the claude harness uses the between-turns auto-arm model (a beacon fresh within FM_GUARD_GRACE is healthy even with no live watcher; only a beacon stale beyond grace, or absent, alarms). Every other harness (codex foreground checkpoint, opencode/pi/grok background arm, tmux, unknown) uses the persistent-watcher model where a live identity-matched watcher is still required (unchanged). Detection uses the existing bin/fm-harness.sh (the single detection owner, consistent with the harness-specific repair line the guard already emits), with an FM_SUPERVISION_MODEL override for tests and callers that already know the harness.
- The banner now names the true failing condition (a missing live watcher process vs a genuinely stale beacon) instead of always blaming the beacon, and the once-per-episode dedup keys on that condition rather than the beacon mtime, so a genuine lapse announces once and does not re-print each turn. A genuinely lapsed chain (stale beacon beyond grace) still alarms.

[...]

PR #1958, opened 2026-08-08
fix(bin): stop fm-watch-arm false FAILED when a cycle close lost a startup race

HendoCode:
## Intent

Fix bin/fm-watch-arm.sh reporting a false 'cycle ended without an actionable reason' FAILED alarm when a watcher cycle genuinely delivered a wake, without weakening detection of a real supervision failure

## What Changed

- In `owned_child_finished()` (`bin/fm-watch-arm.sh`), extend the existing successor/delivery-ledger check that previously ran only on a clean (`rc=0`) child exit to also run on nonzero and signal exits, since a nonzero close can itself be caused by a startup race (the PR-check migration guard or lock acquire rejecting an ambiguous mid-startup lock) rather than a genuine crash — only a close with no live successor and no ledgered delivery is now reported as `watcher: FAILED - cycle ended without an actionable reason`.
- Reclassify cycle-log reason tags accordingly (`unexpected-clean-exit`, `nonzero-exit`, `signal-exit`, and their `-delivered-wake` variants) so the durable lifecycle ledger reflects which path a close took.
- Add `test_arm_reports_sibling_delivered_wake_after_losing_a_startup_race` to `tests/fm-watch-arm.test.sh`, covering an arm whose own forked child loses the PR-check migration race to a live sibling and must report that sibling's delivered wake instead of a false failure; adjust `tests/fm-watcher-lock.test.sh`'s killed-watcher case to set `FM_ARM_CONFIRM_TIMEOUT` now that nonzero exits go through the same bounded successor-check window as clean exits.
- Update `docs/watcher-continuity.md` to describe the widened recheck path and document the new test coverage.

## Risk Assessment

✅ Low: The fix correctly extends the existing successor/delivery-ledger reconciliation (previously only applied to rc=0 clean exits) to every child close that never printed its own wake, closing the exact false-alarm path described in the intent (a startup race lost to a sibling watcher) without weakening genuine-failure detection — a real failure with no successor and no delivery record still logs the reason and returns nonzero. The duplicate FAILED-message removal (93044d1) cleanly resolves a redundancy introduced earlier in the same branch (a292cbc), leaving a single, doc-consistent FAILED message. Docs, header comments, and tests were all updated in lockstep, and the new regression test exercises the real subprocess race (PR-check migration ownership rejection) rather than asserting on source text, satisfying the test-quality bar. The only behavioral trade-off — genuine crashes now wait up to CONFIRM_TIMEOUT (10s default) before declaring FAILED instead of failing instantly — is a bounded, intentional consequence of the same mechanism already used for clean exits, not a detection weakening.

## Testing

test

## Pipeline

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.
- `a`
- `b`

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>


HendoCode:
Withdrawing this — my mistake, not a problem with the change itself.

I raised this against a checkout that was 77 commits behind `main`. Having now caught up, I can see this area has moved considerably, including d0461e4 ("prevent false watcher-down alarms in Claude sessions", #1661), which addresses the same class of symptom I was chasing.

The specific failure I hit may well not exist on current `main`, and if it does the fix likely belongs somewhere different. Rather than ask anyone to review a patch built on stale code, I'll re-verify against current `main` and open a fresh PR only if the problem genuinely reproduces.

Separately: my environment had Homebrew's GNU coreutils ahead of `/usr/bin` in PATH, so bare `stat` was GNU while the scripts branch on `uname = Darwin` and use BSD `stat -f`. That silently produced garbage rather than failing. I've aligned my PATH instead of patching around it, on the assumption stock macOS tooling is the intended baseline. Happy to raise that as an issue if it'd be useful to flag for other macOS users.

Apologies for the noise.

PR #1969, opened 2026-08-08
fix(brief): pin decision key placement in generated briefs

eduardstan:
## Problem
`bin/fm-classify-lib.sh` requires `[key=<slug>]` between the status verb and the colon, but the generated scaffolds did not say so. Workers therefore wrote keys at the end of status lines, and the key never registered.

## Changes
- Clarified parser-accepted key placement in ship, scout, and secondmate briefs.
- Added regression coverage for the incident line and correctly placed keys.
- Kept the parser strict because trailing bracketed prose is ambiguous.

Validation: `tests/fm-brief.test.sh`, focused classifier coverage, `tests/fm-send-resolve-key.test.sh`, and `bin/fm-lint.sh` pass.


eduardstan:
The required `PR must be raised via no-mistakes` check fails here because this branch was pushed directly rather than through that pipeline. I will re-raise the same change through the required path and close this one; no review time is needed on it in the meantime.

eduardstan:
Superseded by https://github.com/kunchenguid/firstmate/pull/2168, which carries the same change raised through the required no-mistakes pipeline.

PR #1983, opened 2026-08-09
fix(relay): re-offer mentions after interrupted wakes

quinnbot-ai:
## Summary

Makes relay mention delivery at least once instead of silently dropping a mention if the poll stops after claiming its marker but before the watcher receives its wake.

The poll now records an un-emitted claim, writes the wake, and only then marks delivery complete.
A later poll re-offers an un-emitted claim.

Because a repeated wake is now possible, the initial answer path records and confirms one durable answer claim per request ID.
This prevents a duplicate public answer while allowing definite failures to remain retryable.

## Validation

- `bin/fm-test-run.sh tests/fm-x-mode.test.sh`
- `bin/fm-lint.sh`
- `git diff --check upstream/main...HEAD`


kunchenguid:

Issue #1988, opened 2026-08-09
Away mode never delivers: Claude's composer renders as ❯+U+00A0, which the classifier reads as pending instead of empty


jason-curtis:
> **Corrected 2026-08-08.** The first version of this issue blamed the busy-pane guard.
> That was wrong — the captain pushed back that firstmate had *not* been continuously
> busy, and testing the classifier directly found a deterministic misclassification
> instead. Root cause and title updated.

## Summary

Claude Code renders its idle composer prompt as `❯` followed by **U+00A0 NO-BREAK SPACE**.
`fm_composer_classify_content` recognises a bare `❯` and `❯` + ASCII space as `empty`, but
not `❯` + U+00A0 — that falls through to `pending`.

Away mode injects **only** on an affirmative `empty`. So on this build every escalation is
refused as "the human has half-typed something, don't clobber it", against a composer that
is in fact empty. It is deterministic: 100% of injections blocked, indefinitely.

Observed live: **6 escalations across 4 crewmates, undelivered for 44,267 seconds
(12+ hours)**, spanning `needs-decision:` and `done:`. The captain found the stalled work
by eye and asked why nothing had escalated.

## Proof

```
$ . bin/fm-composer-lib.sh
$ fm_composer_classify_content 1 "$(printf '\xe2\x9d\xaf')"          # bare ❯
empty
$ fm_composer_classify_content 1 "$(printf '\xe2\x9d\xaf ')"         # ❯ + ASCII space
empty
$ fm_composer_classify_content 1 "$(printf '\xe2\x9d\xaf\xc2\xa0')"  # ❯ + U+00A0
pending          <-- the live case
```

The live pane, captured and decoded:

```
$ tmux capture-pane -p -t %1 | tail -3
'────────────────────'   codepoints ['0x2500', ...]
'❯\xa0'                  codepoints ['0x276f', '0xa0']
```

And end to end, against the real pane:

```
$ fm_backend_composer_state tmux '%1'
pending
```

## Where it goes wrong

`bin/fm-composer-lib.sh`:

- **:188** `'❯'|'›') printf 'empty'; return 0` — matches only a bare glyph with nothing
  after it.
- **:211** `'❯ '*|'› '*|'> '*|'$ '*|'% '*|'# '*) content=${content#??}` — strips the glyph
  plus a single **ASCII** space.

`❯`+U+00A0 matches neither, so the NBSP survives as apparent content.

[...]

PR #1995, opened 2026-08-09
fix(composer): classify a Unicode-space-separated prompt glyph as an empty composer

mackcee:
## Intent

Fix a deterministic composer misclassification: fm_backend_composer_state returns `pending` for a claude composer that is affirmatively empty. `pending` means "real unsubmitted text is sitting there", so every caller that must not type over the captain's half-written message refuses to act. Observed cost: bin/fm-supervise-daemon.sh deferred every away-mode escalation for 8.3 hours overnight (2105 deferrals reading "composer not confirmed-empty" against only 3 genuine "pane busy"), leaving two finished pieces of work unshipped; bin/fm-send.sh reported "text not submitted (delivery unconfirmed; verdict=pending)" for a message that had visibly landed, inviting a duplicate steer.

ROOT CAUSE (accepted, and it SUPERSEDES the two hypotheses the original task brief proposed - the brief blamed either the prompt glyph's truecolor luminance sitting above the 128 ghost threshold, or the input area being drawn with horizontal rules instead of a boxed border; firstmate later instructed me to discard both). Claude renders the idle composer as U+276F followed by U+00A0 NO-BREAK SPACE. fm_composer_classify_content accepted a bare glyph and glyph+ASCII-space as empty, but not glyph+U+00A0, which fell through to pending.

The brief required establishing the cause by counterfactual rather than by reading the code and forming an opinion, changing one condition at a time and recording whether the verdict moved. I did that against the live panes and both original hypotheses were falsified: recolouring the glyph (including to the exact truecolor luminance suspected) did NOT move the verdict, and deleting the horizontal rules did NOT move it, while changing ONLY the separator did. It reproduces identically on tmux and herdr; on tmux the row is nothing but the glyph and the separator, and there the glyph is a 256-colour foreground that is never luminance-tested at all. Upstream filed the same finding as kunchenguid/firstmate#1988 with the same proof; the fix is shaped to close it, and its "Suggested directions" 1 and 2 (normalise Unicode spaces generally rather than enumerating glyph+separator pairs; add a regression pinning glyph+U+00A0 to empty) are implemented.

REQUIRED APPROACH. Prefer normalising Unicode whitespace generally over enumerating more glyph+separator pairs, and fix the single-character strip too: `${content#??}`/`${content#?}` assumed a one-character separator, which breaks on any multi-byte separator independently of the match (measured: it drops one BYTE under LC_ALL=C and one CHARACTER under a UTF-8 locale, leaving locale-dependent trailing bytes on a multibyte glyph). Distinguish the composer's own decoration from user content using the most structural signal available rather than another brightness number, and do not let a single vendor string be load-bearing - a rendered glyph and its colour are exactly what a harness release can change without warning.

HARD CONSTRAINT, deliberately chosen and central to review: do NOT weaken the predicate. This guard's entire purpose is to never type over text a human has half-written. `empty` must remain an affirmative verdict, real typed text MUST still read `pending`, and `unknown` must stay available for genuinely unreadable panes. Making it too permissive would make firstmate silently destroy the captain's unsent input, which is a worse and silent failure than the one being fixed. So the fix follows the Unicode White_Space=Yes property (a standards fact) rather than a brightness threshold or another observed string; whitespace-only content already classified `empty` when the whitespace was ASCII, and any non-whitespace character anywhere still classifies `pending`. The bare dead-shell-prompt refusal is unchanged with or without a separator.

DELIBERATE DECISIONS a reviewer reading only the diff would not know:
- U+200B ZERO WIDTH SPACE is deliberately EXCLUDED from normalization because Unicode gives it White_Space=No; including it would substitute my own guess for the property the owner claims to follow. This boundary is pinned by a test asserting U+200B still reads pending.
- The prompt glyphs were spelled out in three separate places, which bin/fm-composer-lib.sh's own header already flagged as a drift hazard ("must be listed in ALL THREE places"); they are now declared exactly once each and reached through helpers. This is behaviour-preserving.
- The shared normalizing trim is applied in all four adapters (tmux, herdr, orca, cmux) so a non-ASCII space cannot defeat one adapter's structural row scan while another's still matches.
- Normalization substitutes a space rather than deleting, so "foo<NBSP>bar" is not silently joined into one token.
- The live guard types its OWN marker, confirms the harness rendered it, clears it with C-u, and confirms it is gone before reading the empty composer - so `empty` is affirmative rather than assumed. It accepts a startup trust dialog with Enter ONLY after the pane has demonstrably not reflected typed input, so it can never submit a real composer or consume model tokens. It reports an absent harness explicitly and refuses a pass that checked nothing.

TESTS, both required by .agents/skills/firstmate-coding-guidelines "Harness-dependent checks":
- A portable regression under tests/ extending existing scripts rather than adding a runner, pinning the real captured row shapes: the empty composer that must read `empty`, and a genuinely-typed-into one that must still read `pending` (the second matters more than the first). Because U+00A0 is invisible in any capture, the fixtures assert they still contain the separator so no case can pass vacuously. Also pins the whole White_Space property set and locale independence.
- The env-gated, self-skipping live guard in the live-harness-optin family exercising every INSTALLED harness for real and failing by name and version, since a stub can only confirm the assumption written into it.
Verified against real claude 2.1.226 on tmux 3.5a: the empty row is `e2 9d af c2 a0`, and both the portable regression and the live guard fail on that exact row when the fix is reverted. The dated per-harness result is recorded in docs/verification/runtime-backends.md as required.

SCOPE, explicitly bounded by the brief and reaffirmed by firstmate: the composer-state read and its callers only. Do NOT restructure away mode, do NOT change the wake queue, and do NOT touch the Stop-hook watcher path (that path works and is currently the only reliable way firstmate gets woken). Upstream issue 1988's directions 3 and 4 are away-mode changes and are deliberately NOT implemented here.

A third reported symptom - `fm-send.sh --resolve-key <key>` refusing with "no open decision or blocker with that key" while the wake drain listed that key open - was to be treated as possibly a SEPARATE defect and not folded in silently. I established it IS separate: a status line written as `needs-decision: [key=slug] ...` puts the key token AFTER the colon, while _fm_decision_key only parses it BEFORE the colon, so the decision silently degrades to key `default`. Different subsystem and different mechanism from the composer bug, and it already has its own task and branch (fm/fm-decision-key-position). It is deliberately excluded from this change.

Known pre-existing failures NOT caused by this change, each verified to reproduce on the unmodified tree: tests/fm-afk-inject-herdr-e2e.test.sh fails under LC_ALL=C with the fix reverted too; tests/fm-lint.test.sh needs the pinned shellcheck on PATH; tests/fm-test-run.test.sh needs LC_ALL=C collation. With the pinned shellcheck and LC_ALL=C, the full changed-selection suite is 71 scripts with only that pre-existing away-mode E2E failing.

## What Changed

- Added `fm_composer_normalize_spaces_var` / `fm_composer_normalize_trim_var` to `bin/fm-composer-lib.sh`, mapping every non-ASCII code point with the Unicode `White_Space=Yes` property (U+0085, U+00A0, U+1680, U+2000–U+200A, U+2028, U+2029, U+202F, U+205F, U+3000 — deliberately not U+200B, which is `White_Space=No`) onto an ASCII space before every trim, border strip, and glyph comparison, so claude's empty row `❯` + U+00A0 now classifies `empty` instead of `pending`. The predicate is not otherwise loosened: any non-whitespace character still reads `pending`, and a bare shell glyph still reads `unknown`. The same shared trim replaces the open-coded ASCII-only trims in the tmux, herdr, orca, and cmux structural row scans, where the same separator made a bordered composer read `unknown`.
- Consolidated the prompt glyphs into single declarations (`FM_COMPOSER_AGENT_PROMPT_GLYPHS`, `FM_COMPOSER_SHELL_PROMPT_GLYPHS`) reached via `fm_composer_leading_prompt_glyph_var` / `fm_composer_blank_leading_prompt_var`, replacing the three inline copies in the classifier plus the drifted copy in `fm_tmux_composer_geometry_spaces` (which omitted muse's `⟩`). The glyph strip now removes the matched literal instead of `${content#?}`/`${content#??}`, which dropped one byte under `LC_ALL=C` and one character under a UTF-8 locale.
- Added regression coverage pinning the captured `e2 9d af c2 a0` row to `empty` and typed input to `pending` (`tests/fm-composer-lib.test.sh`, `tests/fm-composer-ghost.test.sh`, `tests/fm-backend-herdr.test.sh`, with fixtures asserting the separator is present so no case passes vacuously), plus a new env-gated live guard `tests/fm-composer-harness-drift-live-e2e.test.sh` that types and clears its own marker before reading each installed harness; `bin/fm-test-run.sh` routes `bin/fm-composer*` and `bin/fm-tmux-lib.sh` changes to the `live-harness-optin` family. Backend docs, the harness-adapters skill, and `docs/verification/runtime-backends.md` record the whitespace boundary and the dated per-harness table (claude 2.1.226 on tmux 3.5a; every uninstalled harness listed as unverified).

## Risk Assessment

✅ Low: The change is well-bounded to the composer-state read and its four adapters, the predicate is verifiably not weakened in the dangerous direction (bare dead-shell prompts, ZWSP, and real typed text all still refuse), the Unicode table matches the standard property exactly, both fix rounds' claims check out against the current code, and the only outstanding item is a test-enumeration gap over data I verified correct.

## Testing

I proved the intent end-to-end against the real harness rather than only in fixtures: driving `bin/fm-supervise-daemon.sh`'s actual away-mode escalation guard against a live claude 2.1.226 composer on tmux 3.5a reproduced the reported "composer not confirmed-empty (state=pending)" deferral on the reverted classifier and showed the escalation accepted on the fix, against a composer the demo had itself emptied and whose bytes decode to `e2 9d af c2 a0`; the hard constraint holds on the same live pane, since real unsubmitted text still reads `pending` and still defers. The opt-in live drift guard passes on the fix and fails by name and version when reverted, and reports the seven uninstalled harnesses and the uncovered bordered path rather than reading as full coverage. All portable suites for the changed classifier and the four adapters pass, each new regression fails on base sources in both C and UTF-8 locales, and the test-runner change correctly selects 71 scripts including the new live guard. The one failure seen, the away-mode herdr E2E under `LC_ALL=C`, reproduces identically on the unmodified base tree and passes here in a UTF-8 locale, so it is pre-existing and not caused by this change. No visual artifact applies: the affected surface is a terminal composer read and a daemon log line, both captured as CLI transcripts, and the defect is invisible in any rendered capture because U+00A0 only appears once the bytes are decoded.

<details>
<summary>Evidence: Away-mode escalation guard against a live claude composer: before/after the fix, plus the safety half</summary>

<code>== BEFORE THE FIX (composer classifier reverted to base commit 2d2be63) ==&#10;-- claude version: 2.1.226 (Claude Code)&#10;-- marker cleared; the composer is now affirmatively EMPTY. Its raw bytes:&#10;e2 9d af c2 a0&#10;(e2 9d af = U+276F &#39;\xe2\x9d\xaf&#39; ; c2 a0 = U+00A0 NO-BREAK SPACE)&#10;-- what fm_backend_composer_state reads off that live pane:&#10;pending&#10;-- running the real inject_msg() away-mode escalation guard:&#10;inject_msg -&gt; DEFERRED (escalation withheld)&#10;-- daemon log:&#10;[2026-08-09T06:40:02+0000] inject deferred: supervisor composer not confirmed-empty (state=pending: pending input, dead-shell prompt, or unreadable pane)&#10;&#10;== AFTER THE FIX (worktree at deb3692) ==&#10;-- what fm_backend_composer_state reads off that live pane:&#10;empty&#10;-- running the real inject_msg() away-mode escalation guard:&#10;inject_msg -&gt; ACCEPTED (escalation would be delivered to the captain)&#10;-- daemon log:&#10;&#10;== AFTER THE FIX - SAFETY HALF: real unsubmitted text still blocks injection ==&#10;-- marker deliberately LEFT in the composer (unsubmitted human text). Its raw bytes:&#10;e2 9d af c2 a0 7a 71 78 66 6d 67 75 61 72 64 64 65 6d 6f&#10;-- what fm_backend_composer_state reads off that live pane:&#10;pending&#10;-- running the real inject_msg() away-mode escalation guard:&#10;inject_msg -&gt; DEFERRED (escalation withheld)&#10;-- daemon log:&#10;[2026-08-09T06:41:09+0000] inject deferred: supervisor composer not confirmed-empty (state=pending: pending input, dead-shell prompt, or unreadable pane)</code>

```text
### Away-mode escalation guard against a REAL claude 2.1.226 composer on tmux 3.5a
### (bin/fm-supervise-daemon.sh inject_msg; only the final send is stubbed, so no tokens are spent)

== BEFORE THE FIX (composer classifier reverted to base commit 2d2be63) ==
-- claude version: 2.1.226 (Claude Code)
-- pane with our typed marker still in the composer (the 'human is half-way through' case):
     ❯ zqxfmguarddemo
-- marker cleared; the composer is now affirmatively EMPTY. Its raw bytes:
      e2 9d af c2 a0 
     (e2 9d af = U+276F '\xe2\x9d\xaf' ; c2 a0 = U+00A0 NO-BREAK SPACE)
-- what fm_backend_composer_state reads off that live pane:
     pending
-- running the real inject_msg() away-mode escalation guard:
     inject_msg -> DEFERRED (escalation withheld)
-- daemon log:
     [2026-08-09T06:40:02+0000] inject deferred: supervisor composer not confirmed-empty (state=pending: pending input, dead-shell prompt, or unreadable pane)

== AFTER THE FIX (worktree at deb3692) ==
-- claude version: 2.1.226 (Claude Code)
-- pane with our typed marker still in the composer (the 'human is half-way through' case):
     ❯ zqxfmguarddemo
-- marker cleared; the composer is now affirmatively EMPTY. Its raw bytes:
      e2 9d af c2 a0 
     (e2 9d af = U+276F '\xe2\x9d\xaf' ; c2 a0 = U+00A0 NO-BREAK SPACE)
-- what fm_backend_composer_state reads off that live pane:
     empty
-- running the real inject_msg() away-mode escalation guard:
     inject_msg -> ACCEPTED (escalation would be delivered to the captain)
-- daemon log:

== AFTER THE FIX - SAFETY HALF: real unsubmitted text still blocks injection ==
-- claude version: 2.1.226 (Claude Code)
-- pane with our typed marker still in the composer (the 'human is half-way through' case):
     ❯ zqxfmguarddemo
-- marker deliberately LEFT in the composer (unsubmitted human text). Its raw bytes:
      e2 9d af c2 a0 7a 71 78 66 6d 67 75 61 72 64 64 65 6d 6f 
     (e2 9d af = U+276F '\xe2\x9d\xaf' ; c2 a0 = U+00A0 NO-BREAK SPACE)
-- what fm_backend_composer_state reads off that live pane:
     pending
-- running the real inject_msg() away-mode escalation guard:
     inject_msg -> DEFERRED (escalation withheld)
-- daemon log:
     [2026-08-09T06:41:09+0000] inject deferred: supervisor composer not confirmed-empty (state=pending: pending input, dead-shell prompt, or unreadable pane)
```
</details>
<details>
<summary>Evidence: Live harness drift guard on the fix (real claude 2.1.226, tmux 3.5a)</summary>

<code># claude 2.1.226 (Claude Code): startup row bytes 20 e2 9d af 20 31 2e 20 59 65 73 2c 20 49 20 74 72 75 73 74 20 74 68 69 73 20 66 6f 6c 64 65 72&#10;# claude 2.1.226 (Claude Code): typed input was not rendered back; accepting a startup confirmation with Enter and retrying&#10;# claude 2.1.226 (Claude Code): composer reached through the bare-row path&#10;# claude 2.1.226 (Claude Code): EMPTY composer row bytes e2 9d af c2 a0&#10;ok - composer drift: claude 2.1.226 (Claude Code) reads pending with real typed text and empty once cleared (via the bare-row path)&#10;# not installed on this machine: codex opencode pi pi-signed grok kimi muse&#10;# checked 1 installed harness(es) end to end&#10;# structural paths exercised: bare-row&#10;# no installed harness drew a BORDERED composer box, so the bordered structural path is unverified by this run (tests/fm-composer-ghost.test.sh pins it portably)</code>

```text
# claude 2.1.226 (Claude Code): startup row bytes  20 e2 9d af 20 31 2e 20 59 65 73 2c 20 49 20 74 72 75 73 74 20 74 68 69 73 20 66 6f 6c 64 65 72 
# claude 2.1.226 (Claude Code): typed input was not rendered back; accepting a startup confirmation with Enter and retrying
# claude 2.1.226 (Claude Code): composer reached through the bare-row path
# claude 2.1.226 (Claude Code): EMPTY composer row bytes  e2 9d af c2 a0 
ok - composer drift: claude 2.1.226 (Claude Code) reads pending with real typed text and empty once cleared (via the bare-row path)
# skip: codex is not installed on this machine, so its composer rendering is unverified here
# skip: opencode is not installed on this machine, so its composer rendering is unverified here
# skip: pi is not installed on this machine, so its composer rendering is unverified here
# skip: pi-signed is not installed on this machine, so its composer rendering is unverified here
# skip: grok is not installed on this machine, so its composer rendering is unverified here
# skip: kimi is not installed on this machine, so its composer rendering is unverified here
# skip: muse is not installed on this machine, so its composer rendering is unverified here
# not installed on this machine: codex opencode pi pi-signed grok kimi muse
# checked 1 installed harness(es) end to end
# structural paths exercised: bare-row
# no installed harness drew a BORDERED composer box, so the bordered structural path is unverified by this run (tests/fm-composer-ghost.test.sh pins it portably)
```
</details>
<details>
<summary>Evidence: Same live guard against base-commit sources — fails by name and version</summary>

`not ok - COMPOSER DRIFT: claude 2.1.226 (Claude Code) has a composer this guard just emptied - the typed marker is gone from the rendered pane - yet it classifies 'pending' instead of 'empty', reached through the bare-row path. ... Decoded composer row: e2 9d af c2 a0 (U+00A0 NO-BREAK SPACE is the byte pair c2 a0 ...)`

```text
# claude 2.1.226 (Claude Code): startup row bytes  20 e2 9d af 20 31 2e 20 59 65 73 2c 20 49 20 74 72 75 73 74 20 74 68 69 73 20 66 6f 6c 64 65 72 
# claude 2.1.226 (Claude Code): typed input was not rendered back; accepting a startup confirmation with Enter and retrying
# claude 2.1.226 (Claude Code): composer reached through the bare-row path
not ok - COMPOSER DRIFT: claude 2.1.226 (Claude Code) has a composer this guard just emptied - the typed marker is gone from the rendered pane - yet it classifies 'pending' instead of 'empty', reached through the bare-row path. Every caller that must not overwrite unsubmitted input now refuses to act against this harness: away-mode escalations stop being delivered and steer confirmations report landed messages as unconfirmed. On the bare-row path a 'pending' verdict means the separator or prompt glyph this release renders is not one bin/fm-composer-lib.sh recognises (bare-row drift usually reads 'pending'; a bordered box whose blanked geometry no longer matches its border reads 'unknown'). Decoded composer row:  e2 9d af c2 a0  (U+00A0 NO-BREAK SPACE is the byte pair c2 a0; fm_composer_normalize_trim_var maps every Unicode White_Space=Yes code point, so a byte run outside that property is the thing to add).
```
</details>
<details>
<summary>Evidence: Harness script used for the live away-mode guard demo (evidence-only, outside the worktree)</summary>

```text
#!/usr/bin/env bash
# Product-level demo: drive bin/fm-supervise-daemon.sh's real away-mode
# escalation guard (inject_msg) against a REAL, live, affirmatively-empty
# claude composer in a private tmux server, and show the daemon log line the
# captain would actually read.
#
# The only thing stubbed is the final send (fm_backend_send_text_submit), so no
# text is ever typed into or submitted to the real harness and no model tokens
# are consumed. Everything before it - pane-exists, busy-guard, composer-guard -
# is the real product code reading the real pane.
#
# usage: daemon-guard-demo.sh <tree-root> <label>
set -u
ROOT=$1
LABEL=$2

REAL_TMUX=$(command -v tmux)
SOCKET="fm-guard-demo-$$"
LAB=$(mktemp -d "${TMPDIR:-/tmp}/fm-guard-demo.XXXXXX")
SESSION=demo
cleanup() { "$REAL_TMUX" -L "$SOCKET" kill-server >/dev/null 2>&1 || true; rm -rf "$LAB"; }
trap cleanup EXIT

mkdir -p "$LAB/shim" "$LAB/wt" "$LAB/state"
cat > "$LAB/shim/tmux" <<SH
#!/usr/bin/env bash
exec "$REAL_TMUX" -L "$SOCKET" "\$@"
SH
chmod +x "$LAB/shim/tmux"
PATH="$LAB/shim:$PATH"; export PATH

MARKER=zqxfmguarddemo
TARGET="$SESSION:claude"

"$REAL_TMUX" -L "$SOCKET" new-session -d -s "$SESSION" -n control -c "$LAB/wt"
"$REAL_TMUX" -L "$SOCKET" new-window -d -t "$SESSION:" -n claude -c "$LAB/wt" -- "$(command -v claude)"
sleep 6

pane_has() { "$REAL_TMUX" -L "$SOCKET" capture-pane -p -t "$TARGET" 2>/dev/null | LC_ALL=C grep -qa "$MARKER"; }

# Prove the pane is really at a composer by typing our OWN marker and seeing the
# harness render it back; a startup trust dialog never takes it.
at_composer=0
for _ in 1 2 3; do
  "$REAL_TMUX" -L "$SOCKET" send-keys -l -t "$TARGET" "$MARKER"
  for _ in $(seq 1 20); do pane_has && { at_composer=1; break; }; sleep 0.5; done
  [ "$at_composer" = 1 ] && break
  "$REAL_TMUX" -L "$SOCKET" send-keys -t "$TARGET" C-u >/dev/null 2>&1 || true
  "$REAL_TMUX" -L "$SOCKET" send-keys -t "$TARGET" Enter >/dev/null 2>&1 || true
  sleep 4
done
[ "$at_composer" = 1 ] || { echo "ABORT: never reached a real claude composer"; exit 2; }

echo "== $LABEL =="
echo "-- claude version: $(claude --version 2>/dev/null | head -1)"
echo "-- pane with our typed marker still in the composer (the 'human is half-way through' case):"
"$REAL_TMUX" -L "$SOCKET" capture-pane -p -t "$TARGET" | grep -a "$MARKER" | tail -1 | sed 's/^/     /'

# Clear it, and confirm it is GONE from the rendered pane: the composer is now
# affirmatively empty, because we just emptied it. With CLEAR=0 the marker is
# deliberately LEFT in place, standing in for a captain's half-written message:
# the guard must still refuse, or firstmate would type over it.
if [ "${CLEAR:-1}" = 1 ]; then
  "$REAL_TMUX" -L "$SOCKET" send-keys -t "$TARGET" C-u >/dev/null 2>&1 || true
  cleared=0
  for _ in $(seq 1 40); do pane_has || { cleared=1; break; }; sleep 0.5; done
  [ "$cleared" = 1 ] || { echo "ABORT: composer did not clear"; exit 2; }
  sleep 2
  echo "-- marker cleared; the composer is now affirmatively EMPTY. Its raw bytes:"
else
  pane_has || { echo "ABORT: the marker vanished on its own"; exit 2; }
  echo "-- marker deliberately LEFT in the composer (unsubmitted human text). Its raw bytes:"
fi

row=$("$REAL_TMUX" -L "$SOCKET" capture-pane -p -t "$TARGET" | grep -a "$(printf '\xe2\x9d\xaf')" | tail -1)
printf '%s' "$row" | LC_ALL=C od -An -tx1 | tr -s ' \n' ' ' | sed 's/^/     /'
echo ""
echo "     (e2 9d af = U+276F '\xe2\x9d\xaf' ; c2 a0 = U+00A0 NO-BREAK SPACE)"

# Now the real daemon guard.
# shellcheck source=/dev/null
. "$ROOT/bin/fm-supervise-daemon.sh"
LOG="$LAB/daemon.log"; : > "$LOG"
FM_STATE_OVERRIDE="$LAB/state"; export FM_STATE_OVERRIDE
afk_enter "$LAB/state"
FM_SUPERVISOR_BACKEND=tmux
FM_SUPERVISOR_TARGET="$TARGET"
# Stub ONLY the final send so the demo never types into or submits to the real
# harness. Reaching this stub at all means every guard ahead of it passed.
fm_backend_send_text_submit() { echo "STUBBED-SEND" >&2; printf 'empty'; }

echo "-- what fm_backend_composer_state reads off that live pane:"
echo "     $(fm_backend_composer_state tmux "$TARGET")"
echo "-- running the real inject_msg() away-mode escalation guard:"
if inject_msg "two finished pieces of work are unshipped, come look" 2>/dev/null; then
  echo "     inject_msg -> ACCEPTED (escalation would be delivered to the captain)"
else
  echo "     inject_msg -> DEFERRED (escalation withheld)"
fi
echo "-- daemon log:"
sed 's/^/     /' "$LOG"
```
</details>

## Pipeline

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>

- ⚠️ `bin/fm-tmux-lib.sh:190` - Intent conformance: the criteria state &#34;The shared normalizing trim is applied in all four adapters (tmux, herdr, orca, cmux) so a non-ASCII space cannot defeat one adapter&#39;s structural row scan while another&#39;s still matches.&#34; For tmux that is not the case. The trim was applied to fm_tmux_composer_row_state (lines 132/134/141), but tmux&#39;s structural row scan - which bin/fm-composer-lib.sh&#39;s own header names as &#34;tmux&#39;s visible-pane box scan&#34; - still uses the raw ASCII-only trim at fm-tmux-lib.sh:190-192, and fm_tmux_composer_geometry_spaces:166/172-175 still tests the leftover bytes with [![:space:]]. Verified read-only against the changed code: fm_tmux_find_composer_box 1 with pane rows &#39;╭────────╮&#39; / &#39;│ ❯&lt;U+00A0&gt;     │&#39; / &#39;╰────────╯&#39; returns &#34;0 2 1&#34; (geometry_ambiguous=1), while the identical pane with an ASCII space returns &#34;0 2 0&#34;. fm_tmux_composer_state:342 converts that flag into &#39;unknown&#39; even though fm_tmux_composer_row_state classifies the row &#39;empty&#39; - so a bordered tmux composer separated with U+00A0 still never reads confirmed-empty, and every caller that must not overwrite unsubmitted input still refuses to act. That is the exact authorized failure this change fixes, surviving on the sibling path in the same file. herdr&#39;s ANSI tail scan and orca/cmux&#39;s read-screen scans were normalized, so tmux is precisely the &#34;one adapter defeated while another matches&#34; case the criterion says cannot happen. Earliest supported shared boundary: run each captured line through fm_composer_normalize_trim inside fm_tmux_find_composer_box&#39;s loop (and in fm_tmux_composer_geometry_spaces&#39; probe) instead of the open-coded ${line#&#34;${line%%[![:space:]]*}&#34;} pair. Not proven live-reachable for claude today (its tmux composer is a bare glyph row, not a box), which is why this is a warning rather than an error - but it contradicts a stated deliberate coverage claim, so it is the author&#39;s call whether to close it here or as a follow-up.
- ℹ️ `bin/fm-composer-lib.sh:140` - _fm_composer_normalize_space_var is documented as replacing in place &#34;so the wrapper below adds no subshell&#34;, but fm_composer_normalize_trim prints to stdout and all 12 call sites consume it via command substitution, so every call forks anyway - plus one heredoc temp file per call for the 19-entry mapping loop. The row loops are the cost: bin/backends/orca.sh:276 runs this once per captured line with FM_BACKEND_ORCA_COMPOSER_LINES defaulting to 200, where the previous code was pure parameter expansion with zero forks. Measured on this worktree: 200 command substitutions of fm_composer_normalize_trim take ~72ms versus ~1ms for the parameter-expansion form, per composer read, and composer_state is polled by bin/fm-supervise-daemon.sh and re-read inside the send-submit retry loops. Absolute cost is small, so this is informational: exposing the existing in-place form as a public fm_composer_normalize_trim_var &lt;varname&gt; and using it in the orca/cmux/herdr row loops would deliver the design the comment already describes.
- ℹ️ `bin/fm-tmux-lib.sh:169` - The change&#39;s stated deliberate decision is that the prompt glyphs &#34;are now declared exactly once each and reached through helpers&#34;, and FM_COMPOSER_AGENT_PROMPT_GLYPHS does that for the classifier. fm_tmux_composer_geometry_spaces is a fourth place that still spells them out inline (&#39;&gt;&#39; , &#39;❯&#39;, &#39;›&#39;) and is already drifted: it omits &#39;⟩&#39; (muse, U+27E9), which the classifier does recognise. Consequence on the bordered tmux path: a muse composer&#39;s content row keeps its multibyte glyph, sed &#39;s/[!-~]/ /g&#39; cannot flatten it, the *[![:space:]]* test at line 174 fires, geometry_spaces returns 1, and fm_tmux_find_composer_box sets geometry_ambiguous=1 - so an otherwise-empty muse box reads &#39;unknown&#39; rather than &#39;empty&#39;. Pre-existing rather than introduced here, and it is the same drift hazard the header comment was rewritten to eliminate; reaching FM_COMPOSER_AGENT_PROMPT_GLYPHS from this helper would close it.
- ℹ️ `bin/backends/herdr.sh:2708` - fm_backend_herdr_pi_separator_row still uses the open-coded ASCII-only trim and then requires the remainder be nothing but &#39;─&#39; ([ -z &#34;${row//─/}&#34; ]). A Pi rule padded or terminated with any non-ASCII space fails that test, so no complete separator pair is found, fm_backend_herdr_composer_state&#39;s &#39;separated&#39; shape is never established, and the verdict degrades to &#39;unknown&#39; - the same refuse-to-act outcome, reached through the structural half of the herdr adapter that the row loop at line 2782 normalized but this helper did not. No evidence any Pi release pads its rules that way today, so this is informational; routing this trim through fm_composer_normalize_trim like its sibling would make the herdr adapter internally consistent.

🔧 Fix: normalize Unicode spaces and glyphs in tmux/herdr structural scans
3 infos still open:
- ℹ️ `bin/fm-composer-lib.sh:31` - The fix round added a header claim that is not accurate: &#34;An ADAPTER that needs to recognise a prompt glyph in its own structural scan - tmux&#39;s box-geometry probe is the one such caller - reaches the same declarations ... rather than becoming a fourth copy&#34; (bin/fm-composer-lib.sh:30-35). herdr&#39;s adapter has a second such scan: FM_BACKEND_HERDR_BARE_PROMPT_RE at bin/backends/herdr.sh:2700 is &#39;^(❯|›)&#39;, spelling out two of the three agent glyphs inline and omitting muse&#39;s &#39;⟩&#39;. It is the sole gate for the `bare` shape in fm_backend_herdr_composer_state (herdr.sh:2793-2799): a row that is neither &#39;│&#39;*&#39;│&#39; nor matched by that regex leaves found=0 and the verdict is `unknown`. So this is a second declaration of ❯ and ›, which also contradicts the accepted intent&#39;s deliberate decision that the glyphs &#34;are now declared exactly once each and reached through helpers&#34;. I could NOT establish that the failure is live-reachable, and I am reporting that honestly rather than asserting it: the repo&#39;s two records disagree on muse&#39;s composer shape. .agents/skills/harness-adapters/SKILL.md:419 says &#34;Bordered box whose prompt glyph is `⟩`&#34;, in which case herdr&#39;s &#39;│&#39;*&#39;│&#39; case matches first and this regex never matters; docs/verification/muse.md:150-151 shows the actual `tmux capture-pane -p -e` rows as a border-free `⟩ ` line, in which case an empty muse composer on herdr never establishes a shape and reads `unknown` while the same pane on tmux reads `empty`. The code is pre-existing and unchanged by this diff (its alternation form has a documented locale rationale at herdr.sh:2691-2699, so it cannot simply become a bracket class), and it was not in the round-1 findings or the fix instructions. Author&#39;s call: either reach FM_COMPOSER_AGENT_PROMPT_GLYPHS when building that regex and reconcile the two muse records, or narrow the new header claim so it does not assert a uniqueness the tree does not have.
- ℹ️ `bin/fm-composer-lib.sh:169` - fm_composer_normalize_trim (the value-returning form) has zero callers after the fix round converted all twelve call sites to fm_composer_normalize_trim_var - I grepped every *.sh and *.md in the tree and the only hits are its own definition and the doc line in docs/verification/runtime-backends.md:184, which names the _var form. It is a public, untested function in a shared library, so nothing would catch it drifting from the _var form it wraps. Either drop it, or keep it and note it is currently unused; the comment at :165-168 already explains it is deliberately not the default, so a reader has no way to tell &#34;reserved for future callers&#34; from &#34;leftover&#34;.
- ℹ️ `bin/fm-composer-lib.sh:403` - fm_composer_classify_content still does `content=$(_fm_composer_strip_leading_prompt &#34;$content&#34;)`, a command substitution, immediately after the same file&#39;s comment at :165-168 argues the value-returning form is not the default precisely because &#34;consuming it costs a subshell per call&#34;. _fm_composer_strip_leading_prompt is now a thin wrapper over fm_composer_leading_prompt_glyph_var, so the fork is removable in place with no new mapping and no behavior change: `if fm_composer_leading_prompt_glyph_var glyph &#34;$content&#34;; then content=${content#*&#34;$glyph&#34;}; fi`. Cost is small and bounded - the early returns above mean this line is only reached for rows that carry real content, and a tmux content row already forks three times (strip_ansi, strip_ghost, and the per-row `sed -n Np` in fm_tmux_composer_state:338) - so this is efficiency tidying, not a regression. Verified equivalence of the substitution while reviewing: fm_composer_leading_prompt_glyph_var only matches a glyph at the first non-whitespace position and classify_content trims before this line, so `#*&#34;$glyph&#34;` can never strip past the intended glyph.

🔧 Fix: drop dead trim wrapper, inline glyph strip, correct header claim
1 info still open:
- ℹ️ `tests/fm-composer-lib.test.sh:163` - test_unicode_whitespace_property_set_is_covered independently spells out 11 separators (&#39;\302\205&#39; 00A0 1680 2000 2007 200A 2028 2029 202F 205F 3000) but the implementation&#39;s FM_COMPOSER_UNICODE_SPACES holds 19. The eight unpinned entries are U+2001-U+2006, U+2008 and U+2009 - the interior of the one contiguous run in the table, and the run most exposed to a single-digit slip because the table is hand-transcribed octal (&#39;\0342\0200\0201&#39; .. &#39;\0342\0200\0212&#39; at bin/fm-composer-lib.sh:141-143). I decoded all 19 entries at runtime and they are byte-correct today, so this is a coverage gap and not a live defect; the point is that a future edit dropping or mistyping U+2003 would not fail anything, because the only other test that touches the set (test_every_declared_prompt_glyph_is_blankable_and_classified) iterates the glyph declarations, not the space table, and iterating the table against itself would pass with a wrong byte sequence in it. The intent states the portable regression &#39;pins the whole White_Space property set&#39;; extending the existing literal loop to all 19 escapes (the assertions already in the test body need no change) is what makes that true. Independent literal spelling is the right shape here - do not replace it with a loop over FM_COMPOSER_UNICODE_SPACES, which would be self-referential.

</details>

<details>

✅ No issues found.
- <code>`FM_COMPOSER_HARNESS_DRIFT=1 bash tests/fm-composer-harness-drift-live-e2e.test.sh` — live guard, real claude 2.1.226 on tmux 3.5a: passes (empty row `e2 9d af c2 a0`, bare-row path); same guard against base-commit sources fails with a named COMPOSER DRIFT error</code>
- <code>Manual end-to-end: drove the real `inject_msg()` away-mode escalation guard in `bin/fm-supervise-daemon.sh` against a live claude composer emptied by the harness itself (marker typed, rendered back, cleared with C-u, confirmed gone), with only `fm_backend_send_text_submit` stubbed so no text was submitted and no model tokens spent — reproduced the reported `inject deferred: supervisor composer not confirmed-empty (state=pending...)` on reverted sources and ACCEPTED on the fix</code>
- <code>Manual end-to-end safety half: same live pane with real unsubmitted text left in the composer — `fm_backend_composer_state` reads `pending` and `inject_msg` still DEFERS</code>
- <code>`LC_ALL=C bash tests/fm-composer-lib.test.sh` (13 ok; White_Space property set, U+200B stays `pending`, locale independence across C/C.utf8/en_US.utf8, dead-shell refusal, all 7 declared glyphs blankable)</code>
- <code>`LC_ALL=C bash tests/fm-composer-ghost.test.sh` (bordered + unbordered tmux paths, muse `⟩`, ambiguous geometry still `unknown`)</code>
- <code>`LC_ALL=C bash tests/fm-backend-herdr.test.sh` (new `❯`+U+00A0 empty/pending cases and U+00A0-padded Pi separator cases)</code>
- <code>Revert proof: ran `tests/fm-composer-lib.test.sh` and `tests/fm-composer-ghost.test.sh` against a tree with `bin/fm-composer-lib.sh`, `bin/fm-tmux-lib.sh` and the three adapters restored to 2d2be63 — both fail (`got &#39;pending&#39;`) in `C` and `C.UTF-8`</code>
- <code>`LC_ALL=C bash tests/fm-backend-orca.test.sh`, `tests/fm-backend-cmux.test.sh`, `tests/fm-backend.test.sh`, `tests/fm-test-run.test.sh` — all pass</code>
- <code>`LC_ALL=C bash bin/fm-test-run.sh --changed --base 2d2be63 --list` — selects 71 scripts and includes `tests/fm-composer-harness-drift-live-e2e.test.sh`, confirming the changed-file map wiring</code>
- <code>`LC_ALL=C bash tests/fm-afk-inject-herdr-e2e.test.sh` on this tree and on a clean 2d2be63 checkout — fails identically on both (pre-existing); passes on this tree in the default UTF-8 locale</code>

</details>

<details>

- ⚠️ `.agents/skills/harness-adapters/SKILL.md:419` - Two doc surfaces still contradict each other on muse&#39;s composer shape, and I could not resolve it here because muse is not installed on this machine. `.agents/skills/harness-adapters/SKILL.md:419` records muse&#39;s composer as a &#34;Bordered box whose prompt glyph is `⟩` (U+27E9)&#34;, while `docs/verification/muse.md:150-151` shows a captured row with no box-drawing border - just `⟩` plus an ASCII space with a background fill. Which one is true decides whether herdr&#39;s `FM_BACKEND_HERDR_BARE_PROMPT_RE` (which spells out `❯` and `›` and omits `⟩`) is reachable at all: if muse draws a bare row, an empty muse composer reads `unknown` on herdr while the same pane reads `empty` on tmux. This change&#39;s new comment in `bin/fm-composer-lib.sh` names the contradiction and defers the permissive regex fix to its own task; the doc side needs one live muse capture to settle which surface is stale, then a correction in the wrong one. Follow-up, not fixable in this change.

🔧 Fix: cross-reference muse composer shape contradiction on both surfaces

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

### DIFF
--- a/.agents/skills/harness-adapters/SKILL.md
+++ b/.agents/skills/harness-adapters/SKILL.md
@@ -421,6 +421,10 @@ Muse Code is a CREWMATE and SCOUT adapter only.
 | Effort | `--reasoning-effort`, default `high`; see the launch-profile table above for the mapping. |
 | Resume | `muse resume --last` or `muse resume <session-uuid>`; bare `muse resume` opens a picker. |
 
+That bordered-box composer is a DESCRIPTION, and it disagrees with the border-free row CAPTURED in `docs/verification/muse.md` "Composer rendering"; task `fm-muse-herdr-bare-glyph` settles which is true against real muse.
+It decides which shape in `bin/fm-composer-lib.sh`'s catalogue an idle muse composer actually exercises - `bordered` or `bare` - and therefore which one a future shape change may break without any test noticing.
+The per-adapter divergence this used to cause is gone: every backend now reaches the same shared classifier, which recognises `⟩` on both shapes, so muse no longer reads `empty` on one backend and `unknown` on another.
+
 ### Credentials are a spawn preflight, not a screen check
 
 muse reads `META_API_KEY` (which always wins) or a stored credential at `${XDG_CONFIG_HOME:-$HOME/.config}/muse/auth.json`, written by `muse login` (an OIDC device-code flow) or `muse auth set --api-key-stdin`.
--- a/.agents/skills/harness-adapters/SKILL.md
+++ b/.agents/skills/harness-adapters/SKILL.md
@@ -188,6 +188,8 @@ Firstmate launches every claude crewmate and secondmate with `CLAUDE_CODE_ENABLE
 The CLI's `--prompt-suggestions` flag is print/SDK-mode only and does not suppress the interactive composer ghost text, verified empirically on v2.1.186.
 As defense in depth for any pane that flag cannot reach, including the captain's own firstmate composer that away-mode reads, the shared `fm_composer_strip_ghost` extractor in `bin/fm-composer-lib.sh` removes dim/faint SGR 2 ghost runs before pending-input classification on every styled reader (tmux, herdr, and Zellij).
 Its broader dark-TRUECOLOR placeholder handling and dark-theme tradeoff are documented in `docs/herdr-backend.md` "Composer and injection safety", with active captures in `docs/verification/runtime-backends.md`.
+Claude also separates that prompt glyph from the composer's content with U+00A0 NO-BREAK SPACE rather than an ASCII space, so the same owner normalizes every code point carrying the Unicode `White_Space=Yes` property before classifying.
+A newly verified harness therefore needs no separator enumerated unless the one it draws falls outside that property, which `FM_COMPOSER_HARNESS_DRIFT=1 tests/fm-composer-harness-drift-live-e2e.test.sh` is what detects.
 That styled capture is internal to the boolean detector only.
 `fm-peek` and every other human or LLM-facing capture path stays plain `tmux capture-pane` with no escape codes.
 
--- a/bin/fm-test-run.sh
+++ b/bin/fm-test-run.sh
@@ -182,6 +182,7 @@ family_for_basename() {
     fm-afk-pi-herdr-return-e2e.test.sh|\
     fm-cmux-claude-composer-live-e2e.test.sh|\
     fm-composer-matrix-live-e2e.test.sh|\
+    fm-composer-harness-drift-live-e2e.test.sh|\
     fm-codex-continuity-live-e2e.test.sh|fm-grok-continuity-live-e2e.test.sh|\
     fm-grok-stop-live-e2e.test.sh|fm-harness-liveness-drift-live-e2e.test.sh|\
     fm-muse-signals-live-e2e.test.sh|\
@@ -935,8 +936,8 @@ families_for_changed_path() {
       ;;
     bin/fm-composer-lib.sh)
       # The shared shape catalogue is vendor-rendered signal; a change to it
-      # re-selects the live guard (fm-composer-matrix-live-e2e) alongside the
-      # portable families.
+      # re-selects the live guards (fm-composer-matrix-live-e2e and
+      # fm-composer-harness-drift-live-e2e) alongside the portable families.
       printf '%s\n' backend-dispatch
       printf '%s\n' pure-contract-unit
       printf '%s\n' live-harness-optin
--- a/docs/herdr-backend.md
+++ b/docs/herdr-backend.md
@@ -236,6 +236,11 @@ ANSI capture preserves de-emphasized placeholder style.
 `bin/fm-composer-lib.sh` is the fleet-wide owner that strips dim or faint runs and dark truecolor placeholders while retaining bright typed input.
 If the ANSI capture ever fails, the plain fallback declares itself unstyled and the classifier degrades a glyph row carrying trailing text to `unknown` instead of misreading ghost suggestions as typed input, which safely defers injection and eventually raises the wedge alarm.
 
+Separator whitespace is normalized rather than enumerated per harness.
+The shared owner's structural row scan and its content classifier alike map each code point carrying the Unicode property `White_Space=Yes` onto an ASCII space before trimming, border stripping, and glyph comparison, so a prompt glyph followed only by a non-ASCII space - claude draws U+00A0 NO-BREAK SPACE - is an affirmatively empty composer instead of pending input.
+The predicate is not otherwise relaxed: any non-whitespace character anywhere is still pending, and U+200B ZERO WIDTH SPACE carries `White_Space=No`, so it stays pending.
+[`verification/runtime-backends.md`](verification/runtime-backends.md#composer-separator-rendering) owns the per-harness captured rows and the refresh command.
+
 A bare shell prompt is never an empty agent composer.
 Away-mode injection proceeds only on an affirmative `empty` result, never on unknown.
 This prevents a dead agent pane from receiving and possibly executing an escalation as shell input.
--- a/docs/tmux-backend.md
+++ b/docs/tmux-backend.md
@@ -70,6 +70,8 @@ Agent liveness and composer safety are separate checks.
 The tmux reader is a thin adapter over the fleet-wide classifier in `bin/fm-composer-lib.sh`: it contributes one styled full-pane capture, the `#{cursor_y}` cursor row, and a Pi foreground-process identity probe, and the shape containing the cursor - a complete bordered box (titled bottom borders tolerated), a bare agent-glyph row with its wrapped input, opencode's left bar, or Pi's identity-corroborated separator pair - decides the verdict.
 Real text in an identified shape is pending, while only positively proven emptiness reads empty.
 A blank or otherwise unidentified cursor row is `unknown` and every consumer defers: this strict container-proof rule replaced the earlier permissive blank-row reading, so a modal dialog, a dead shell between stale rules, or a mid-redraw pane is never an injection target.
+The box-geometry probe that proves a row blank to the border's own width uses that same owner's Unicode-whitespace normalization and its single prompt-glyph declarations, so a non-ASCII separator or an accepted prompt glyph can no longer leave the content classifier reading empty while the structure reads ambiguous.
+[herdr-backend.md](herdr-backend.md#composer-and-injection-safety) owns the shared content contract, including the whitespace boundary.
 The shared classifier accepts a shell glyph as an empty agent composer only inside a bordered container.
 A bare shell prompt is `unknown`, so away-mode escalation is never injected into a dead shell.
 
@@ -103,6 +105,7 @@ tests/fm-backend-tmux-smoke.test.sh
 tests/fm-tmux-agent-liveness.test.sh
 tests/fm-harness-liveness-drift-live-e2e.test.sh
 tests/fm-composer-ghost.test.sh
+tests/fm-composer-harness-drift-live-e2e.test.sh
 tests/fm-kimi-harness.test.sh
 tests/fm-muse-harness.test.sh
 tests/fm-tmux-submit-busy.test.sh
--- a/docs/verification/muse.md
+++ b/docs/verification/muse.md
@@ -154,6 +154,9 @@ Captured with `tmux capture-pane -p -e`:
 Prompt glyph `⟩` (U+27E9) at luminance ~149.9 against the 128 default ghost threshold; typed text at ~209.8.
 After a single Escape the interrupted prompt is restored into the composer at the same bright ~209.8, and `C-u` clears it.
 
+The rows above carry no box-drawing border, while `.agents/skills/harness-adapters/SKILL.md`'s muse `Composer` row describes a bordered box; this page holds an observed capture and that row holds a claim, so they are not equal evidence.
+That does not settle it, because these rows do not record when in the session they were taken and a capture made at launch would not show what an IDLE composer draws; task `fm-muse-herdr-bare-glyph` resolves it against real muse, and the SKILL.md row states which shape in the shared catalogue it decides muse exercises.
+
 ## The credentialed multi-step smoke (verified 2026-08-06)
 
 This was the one item deferred until a `META_API_KEY` was available, because it is what decides whether a settled log may classify `idle`.
--- a/docs/verification/runtime-backends.md
+++ b/docs/verification/runtime-backends.md
@@ -174,7 +174,8 @@ Every harness absent from this machine is recorded as unverified rather than pas
 
 Before the fix in `bin/fm-composer-lib.sh`, that exact `e2 9d af c2 a0` row classified `pending`, meaning "a human's unsubmitted text is sitting there", so every caller that must not overwrite unsubmitted input refused to act (upstream `kunchenguid/firstmate#1988`).
 `fm_composer_normalize_trim_var` now maps every code point with the Unicode property `White_Space=Yes` onto an ASCII space before any trim or glyph comparison, so the classifier follows a standards property instead of a separator observed in one release.
-`tests/fm-composer-lib.test.sh` and `tests/fm-composer-ghost.test.sh` pin the same bytes in CI.
+The same shared mapping runs inside the shared owner's own structural row scan, where the separator produced a different symptom with the identical consequence: the box geometry could not be proven blank and an empty bordered composer read `unknown`.
+`tests/fm-composer-lib.test.sh` and `tests/fm-composer-ghost.test.sh` pin the same bytes in CI, on the bordered and unbordered shapes alike, asserting the exact `empty` / `pending` / `unknown` verdict in each case.
 
 ### Cleanup endpoint identity
 
--- a/docs/verification/runtime-backends.md
+++ b/docs/verification/runtime-backends.md
@@ -144,6 +144,38 @@ Zellij has no verified recovery-grade agent process probe, while Orca and cmux d
 The current classifier matrix and its refresh guard are recorded in [Composer classification matrix](#composer-classification-matrix), with portable shape coverage in `tests/fm-composer-lib.test.sh` and `tests/fm-composer-ghost.test.sh`.
 Kimi pointer delivery and OpenCode 1.18.4 busy-queue behavior remain pinned by `tests/fm-kimi-harness.test.sh` and `tests/fm-tmux-submit-busy.test.sh`.
 
+### Composer separator rendering
+
+This is the per-harness record of what an EMPTY composer actually renders, and it is the owner the Herdr "Composer and operational input" section below refers to.
+The verdict a composer read produces is taken from the harness's own input row, so the separator that row draws is a vendor surface under no contract.
+Claude Code separates its prompt glyph from the composer's content with U+00A0 NO-BREAK SPACE rather than an ASCII space, which is invisible in any rendered capture and only appears when the bytes are decoded.
+
+Refresh this table with the opt-in guard, which drives each installed harness for real, types a marker, clears it, and decodes the resulting empty row:
+
+```sh
+FM_COMPOSER_HARNESS_DRIFT=1 tests/fm-composer-harness-drift-live-e2e.test.sh
+```
+
+Run on 2026-08-09 with tmux 3.5a on Debian GNU/Linux 13 x86_64, each harness on a private socket in an isolated lab.
+The composer is emptied by the guard itself - it types a marker, confirms the harness rendered it, sends `C-u`, and confirms the marker is gone from the pane - so the row below is affirmatively empty rather than assumed empty.
+
+| Harness | Version | Empty composer row bytes | Separator | Verdict |
+| --- | --- | --- | --- | --- |
+| claude | 2.1.226 | `e2 9d af c2 a0` | U+00A0 | empty |
+| codex | not installed | - | - | unverified here |
+| opencode | not installed | - | - | unverified here |
+| pi | not installed | - | - | unverified here |
+| pi-signed | not installed | - | - | unverified here |
+| grok | not installed | - | - | unverified here |
+| kimi | not installed | - | - | unverified here |
+| muse | not installed | - | - | unverified here |
+
+Every harness absent from this machine is recorded as unverified rather than passed over, and the guard refuses to report a pass when no installed harness reached a readable composer.
+
+Before the fix in `bin/fm-composer-lib.sh`, that exact `e2 9d af c2 a0` row classified `pending`, meaning "a human's unsubmitted text is sitting there", so every caller that must not overwrite unsubmitted input refused to act (upstream `kunchenguid/firstmate#1988`).
+`fm_composer_normalize_trim_var` now maps every code point with the Unicode property `White_Space=Yes` onto an ASCII space before any trim or glyph comparison, so the classifier follows a standards property instead of a separator observed in one release.
+`tests/fm-composer-lib.test.sh` and `tests/fm-composer-ghost.test.sh` pin the same bytes in CI.
+
 ### Cleanup endpoint identity
 
 The cleanup identity boundary was validated on 2026-07-28 with tmux 3.6a and metadata fixtures for every supported backend.
@@ -521,7 +553,7 @@ ok - forced teardown retains a nested secondmate home and its grandchild's Herdr
 
 Real captures verified these active distinctions:
 
-- Claude and Codex use bare `❯` and `›` agent composers.
+- Claude and Codex use bare `❯` and `›` agent composers; the tmux "Composer separator rendering" section above owns the per-harness record of the whitespace each one draws after that glyph.
 - Pi uses content between complete separator rows and requires exact native Pi identity.
 - Dim or faint suggestion text is ghost content, while normally styled text is pending input.
 - Grok dark truecolor placeholders are ghost content, while bright truecolor typed input remains pending.
--- a/tests/fm-backend-herdr.test.sh
+++ b/tests/fm-backend-herdr.test.sh
@@ -3095,6 +3095,42 @@ test_composer_state_pi_separator_real_text_is_pending() {
   pass "fm_backend_herdr_composer_state: real Pi composer text remains pending"
 }
 
+# The herdr adapter's STRUCTURAL half. The row scan above was taught the shared
+# normalizing trim, but the Pi separator recogniser kept an ASCII-only one and
+# then required the remainder be nothing but `─`, so a rule terminated with a
+# non-ASCII space left residue, no complete separator pair was found, the
+# `separated` shape was never established, and the verdict degraded to
+# `unknown` - the same refuse-to-act outcome by a different route. Both rules
+# below carry a trailing U+00A0, and the fixture asserts it is still there
+# because it is invisible in any capture.
+test_composer_state_pi_separator_padded_with_nbsp_is_empty() {
+  local dir log resp fb out
+  dir="$TMP_ROOT/composer-pi-separated-nbsp"; mkdir -p "$dir/responses"; log="$dir/log"; resp="$dir/responses"; : > "$log"
+  printf '\x1b[0m\x1b[38;2;129;162;190m─────────────────────────────────────────────────────\302\240\x1b[0m\n\x1b[0m\x1b[7m \x1b[0m                                                    \n\x1b[0m\x1b[38;2;129;162;190m─────────────────────────────────────────────────────\302\240\x1b[0m\n' > "$resp/1.out"
+  herdr_nbsp_fixture_check "$resp/1.out" "U+00A0-padded Pi separator"
+  printf '{"result":{"agent":{"agent":"pi","agent_status":"idle"}}}\n' > "$resp/2.out"
+  fb=$(make_herdr_fakebin "$dir")
+  out=$( PATH="$fb:$PATH" FM_HERDR_LOG="$log" FM_HERDR_RESPONSES="$resp" \
+    bash -c '. "$0/bin/backends/herdr.sh"; fm_backend_herdr_composer_state lab:w1:p2' "$ROOT" )
+  [ "$out" = empty ] || fail "an idle Pi composer whose rules are padded with U+00A0 should read empty, got '$out'"
+  pass "fm_backend_herdr_composer_state: a Pi separator pair padded with U+00A0 still establishes the composer shape"
+}
+
+# And the negative half on that same structural path: recognising the padded
+# rules must not make the content inside them read empty when it is real text.
+test_composer_state_pi_separator_padded_with_nbsp_real_text_is_pending() {
+  local dir log resp fb out
+  dir="$TMP_ROOT/composer-pi-separated-nbsp-pending"; mkdir -p "$dir/responses"; log="$dir/log"; resp="$dir/responses"; : > "$log"
+  printf '\x1b[38;2;129;162;190m─────────────────────────────────────────────────────\302\240\x1b[0m\nprivacy safe human draft\x1b[7m \x1b[0m\n\x1b[38;2;129;162;190m─────────────────────────────────────────────────────\302\240\x1b[0m\n' > "$resp/1.out"
+  herdr_nbsp_fixture_check "$resp/1.out" "U+00A0-padded Pi separator with text"
+  printf '{"result":{"agent":{"agent":"pi","agent_status":"done"}}}\n' > "$resp/2.out"
+  fb=$(make_herdr_fakebin "$dir")
+  out=$( PATH="$fb:$PATH" FM_HERDR_LOG="$log" FM_HERDR_RESPONSES="$resp" \
+    bash -c '. "$0/bin/backends/herdr.sh"; fm_backend_herdr_composer_state lab:w1:p2' "$ROOT" )
+  [ "$out" = pending ] || fail "real text between U+00A0-padded Pi rules must stay pending, got '$out'"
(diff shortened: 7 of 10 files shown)


kunchenguid:

<sub>Noted for firstmate#1995 at `ed2c271c`.</sub>

kunchenguid:
Speaking as Kun's firstmate: closing this leftover.

Current `main` `ddf74ef22f73` already classifies Claude's `❯`+U+00A0 idle composer as `empty` (locale-independent) via merged #2102, with `tests/fm-composer-lib.test.sh` pinning issue #1988. This PR targeted that same defect. Thank you @mackcee — really appreciate you taking the time on this. The landing that superseded it is #2102.

PR #2000, opened 2026-08-09
fix(spawn): force regular Pi TUI for crew launches

roelofb:
Supervised crew panes are read programmatically. When a user-global settings file pins tuiMode to the experimental fullscreen renderer, crews inherit it, and injected supervisor messages land deep in rewritten scrollback (pi-tui scrollback clears during streaming: https://github.com/earendil-works/pi/issues/7304). The supervisor's delivery verification then reports false failures while the message actually arrived.

This passes --tui-mode regular unconditionally on pi and pi-signed crew launches. The flag is the documented per-process override for the setting (v0.84.0 settings docs, UI and display). Fullscreen stays available for humans; supervised panes always get the regular renderer.

One test asserts the flag is present for pi launches and absent for other harnesses.


roelofb:
Superseded by #2005, raised through the required no-mistakes gate.

PR #2011, opened 2026-08-09
fix(bin): supervision loop reliability - send read-back, spawn preflight, open-PR teardown guard, watcher stall/seat-death/refill detection, external watchdog

willchen95:
Reliability fixes for the supervision loop, distilled from a 2026-08-02 field audit of a 12-lane fleet where several independent failure modes each silently stalled or double-dispatched work.

## Fixes

- **fm-send: verify delivery by read-back before failing** - an inconclusive submit on a provably busy pane now captures the pane and, when the typed text is visibly queued, reports a queued delivery (exit 0) instead of a false "retype it" failure that caused double-queued steers. Absent text keeps the loud nonzero refusal.
- **fm-spawn: refuse an unfilled scaffold brief** - a brief whose bare `{TASK}` placeholder line was never replaced is refused before any endpoint or worktree side effect (3 of 8 lanes launched dead this way in the audit). Matches only a line that IS the placeholder, so filled briefs that mention `{TASK}` in prose still pass.
- **fm-teardown: refuse while the task's recorded PR is still open** - teardown removes the lane that owns the PR and its armed merge poll, so tearing down early orphans a PR nobody is driving. Merged/closed proceeds, an unreadable state refuses (fail closed), `--force` keeps its explicit-discard override.
- **fm-watch: quiet-stall and seat-death detection, refill-aware heartbeat, captain pulse** - detects the idle-prompt/token-burn stall that event-keyed staleness misses, surfaces provider seat death that leaves phantom in-flight lanes, arms an optional `config/refill-target` heartbeat so a drained fleet with ready backlog refills, and overwrites a one-line `state/.captain-pulse` liveness pulse every poll.
- **fm-watchdog: external watchdog with WSL-native alerting** - a watcher-independent process that alarms when the watcher beacon goes stale, so a dead supervision chain cannot die silently.
- **away-mode escalations ride the durable wake queue** - daemon escalations are queued durably instead of injected into a live composer, removing the composer-limbo loss path.

## Test fixes included

Three existing tests read the live home's real records instead of their own sandbox and fail on any operator machine with a populated home:

- `fm-spawn-batch.test.sh` passed `FM_CONFIG_OVERRIDE=''`, which falls through to the live `config/`, so a real `crew-dispatch.json` trips the consultation backstop before the behavior under test.
- `fm-teardown.test.sh` never set `FM_DATA_OVERRIDE`, so a real `data/secondmates.md` makes the registry validation refuse before the case runs.
- `fm-watch-triage.test.sh` trips shellcheck 0.11.0's SC2030/SC2031 info findings on a deliberately subshell-contained PATH; annotated the intent.

## Verification

- `bin/fm-lint.sh` green (shellcheck 0.11.0).
- 27 test files across the touched areas pass locally on top of current main (74230fc), including the new `fm-send-strict`, `fm-spawn-batch`, `fm-teardown`, `fm-watch-triage`, `fm-watchdog`, and `fm-daemon` coverage added here.
- Observation, unrelated to this branch: `tests/fm-session-start.test.sh`'s "successful Herdr husk recovery should stay non-actionable" fails identically on untouched current main in this environment (WSL2, Herdr installed).


kunchenguid:

PR #2016, opened 2026-08-09
fix(paired-review): resolve pair roles by stable identity, not stale pane ids

huynhtandat223:
## Problem

A live driver/navigator pair could not be composed or recomposed: `fm-pair-compose.sh` refused with `reciprocal adjacency` even when both roles were live, correctly labelled, and adjacent.

Two defects, both in how the coordinator read Herdr:

1. **Pane ids are positional and are reassigned on every move.** Herdr's own `pane move` response carries `previous_pane_id` alongside the new `pane`. The coordinator took `herdr_pane_id` from task metadata and carried it through its own workspace and split moves, so every later check ran against an id that no longer named any pane.

2. **The reciprocal check read the wrong field.** `pane neighbor` returns the *queried* pane in `pane_id` and the adjacent pane in `neighbor_pane_id`. The extraction was `.result.pane.pane_id // .result.neighbor.pane_id`, which resolves to the query itself, so each pane was compared against itself and adjacency could never hold — independently of any move.

Verified read-only against the running Herdr session and a live pair (`w5K:p1` driver, `w5K:p2` navigator, same tab, genuinely adjacent):

```
$ herdr pane neighbor --direction right --pane w5K:p1
{"result":{"neighbor":{"direction":"right","neighbor_pane_id":"w5K:p2","pane_id":"w5K:p1",...}}}

old expression -> w5K:p1   (the query; compared against itself -> refusal)
new expression -> w5K:p2   (the true neighbour)
```

## Fix

A role's stable identity is its registered agent name (`<pair-id>-driver`, `<pair-id>-navigator`) bound to that role's isolated copy. Composition and the new `recover` subcommand both resolve the role's *current* pane, workspace, and tab from Herdr's own agent inventory under that identity, and stored pane ids are treated as hints only.

- `pair_resolve_role` resolves exactly one live agent per role and proves it sits in the expected isolated copy before any pane, workspace, or tab is used. Missing, duplicated, or copy-mismatched identity returns non-zero rather than guessing. Before composition publishes the role names, the copy alone is the identity, and an ambiguous copy refuses.
- `pair_verify_topology` proves both roles distinct, in one workspace and tab, and reciprocal neighbours (now reading `neighbor_pane_id`), and names the failed invariant otherwise. A topology generation is recorded only after it returns.
- `recover <recovery.json>` re-verifies a live pair after Herdr moves a pane and appends the next topology generation. An unchanged topology is reported rather than appended twice; any refusal leaves the recorded topology and the live pair untouched.
- Composition re-resolves each role after each of its own moves, so its moves reassigning pane ids is no longer a failure.

`fm-send` delivery semantics are untouched: it remains the only pair signal path, and Herdr is still read for topology only, never used to inject text.

## Tests

The fake Herdr in `tests/fm-pair-compose.test.sh` encoded the same wrong response shapes as the helper, which is exactly why this defect passed CI. It now models the real CLI: pane ids reassigned on every move, `neighbor_pane_id` distinct from the queried pane, agent names as the stable identity, and an unrelated agent present in the session so resolution has to discriminate.

New coverage:

- Published topology records the panes the roles occupy now, not the launch ids.
- A stale recorded pane id does not refuse a live, adjacent pair.
- A moved pair is recovered from role identity into a new generation with the new panes.
- Recovery is idempotent — an unchanged topology appends nothing.
- Recovery refuses, records no generation, and leaves the evidence byte-identical on a copy-mismatched, missing, duplicated, or split-tab role identity; a restored endpoint then recovers normally.
- Composition refuses on duplicated, lost, and copy-drifted role identity, and no failing scenario records a topology generation or releases the barrier.

Confirmed non-vacuous: the new suite fails against the pre-fix helper.

## Validation

```
bash tests/fm-pair-compose.test.sh            ok
bash tests/fm-temporary-supervisor.test.sh    ok  (other custom-skills coverage)
bash tests/fm-documentation-audiences.test.sh ok
bin/fm-lint.sh                                clean
bin/fm-lint.sh custom-skills/paired-review/fm-pair-compose.sh   clean
bin/fm-doc-audience-check.sh                  ok surfaces=123 local_links=284
```

The helper had 12 pre-existing SC2016 findings (deliberate Markdown code spans and jq variables in single-quoted strings); three narrow directives make the changed script shellcheck-clean without a blunt file-scope suppression.

## Notes

- `custom-skills/**` is outside `bin/fm-lint.sh`'s canonical CI set, so this helper is only linted when passed explicitly. Widening that set touches the lint owner and was left out of scope.
- Copy proof relies on Herdr's `foreground_cwd` (the pane's `cwd` is the project directory, not the role copy). If that field were ever unavailable, composition refuses with a named invariant rather than proceeding on an unproven identity — the intended fail-safe direction.
- The live pair `dms-workspaces-autoledgers-sync-build` was recovered manually and is untouched here; no existing pair is mutated or cleaned up by this change.


huynhtandat223:
Opened against the wrong base repository by mistake - this change belongs on the huynhtandat223/firstmate fork, not upstream. Closing; no review needed here.

PR #2039, opened 2026-08-10
feat(notify): speak a phrase instead of a system sound

Bewy-dot:
## Summary

- The three producer-tag notification classes (`pr-merged`, `pr-ready`, `attention`) now default to speaking a captain-chosen phrase via macOS `say` instead of playing a named system sound, per the captain's request: "the bleeps and boops aren't doing it."
- `config/notify` gains `voice=`, the literal per-class value `speak`, and `<class>-phrase=` keys, so a reworded phrase or a different voice is a one-line config edit with no code change.
- An existing `config/notify` that names sounds explicitly keeps behaving exactly as before - explicit sound configuration always wins over the new speech default.
- Speech runs detached in the background under the same timeout bound as every other channel, so it never blocks a merge or supervision cycle, and falls back to the pre-speech named-sound behavior whenever `say` is missing, the configured voice is not installed, or the background launch fails for any reason.
- Voice and phrase config values are validated against a plain-text allow-list before ever reaching `say`.

## Verification

I ran the documented smoke commands on this machine (macOS, `say -v Zoe` installed) for all three classes and confirmed each executed successfully end to end (real config, real scope guard, real backgrounded `say` call). I do not have the ability to literally listen as an autonomous agent, so I additionally rendered each of the three phrases to disk with `say -v Zoe -o <class>.aiff '<phrase>'` and measured the raw PCM:

- `pr-merged`: 2.71s total, 0 RMS for the first 500ms, 0 RMS for the last 800ms, ~4711 RMS in between.
- `pr-ready`: 2.67s total, 0 RMS for the first 500ms, 0 RMS for the last 600ms, ~4107 RMS in between.
- `attention`: 2.84s total, 0 RMS for the first 500ms, 0 RMS for the last 600ms, ~3178 RMS in between.

That is measurable proof of real speech content with clean silence at both ends (no clipping) for all three phrases. The captain should still listen once by ear to confirm it sounds right to him, per his own request to verify by ear.

## Test plan

- [x] `bin/fm-lint.sh` clean
- [x] `bin/fm-doc-audience-check.sh` clean
- [x] `tests/fm-notify.test.sh` - 36 assertions, all passing, covering: default speech per class, distinct default phrases, explicit `speak` opt-in, `<class>-phrase` and `voice=` overrides, invalid-phrase/invalid-voice fallback to defaults, per-class channel override while speaking, backward-compatible named-sound behavior (the pre-existing default-sound tests all still pass unchanged with an explicit sound config), and real fail-soft coverage (say missing, voice not installed, successful non-blocking backgrounded speech) via fake `say`/`osascript` binaries - no test posts a real notification or makes a real sound.
- [x] Full `tests/fm-watch-triage.test.sh` re-verified unaffected (pre-existing slow test, confirmed passing on both baseline and this branch).


Bewy-dot:
Opened against the wrong repo by mistake; reopening against Bewy-dot/firstmate.

PR #2047, opened 2026-08-10
fix(composer): read a blank-padded composer as empty, not as unsent text

huynhtandat223:
## What broke

`fm-pair-compose.sh` releases both pair barriers through verified `fm-send` submission. The Claude driver **visibly received** `PAIR READY`, but `fm-send` exited 1 with `delivery unconfirmed; verdict=pending`. The composer treated that correctly under its contract and aborted before releasing the navigator, so `ready` was never published and `psak-dms-kernel-canonicalization` stalled before its plan gate with no navigator coverage.

## Trigger, masking condition, symptom

- **Initiating trigger.** Claude 2.1.226 pads its **empty** composer row with U+00A0 after the `❯` glyph. Verified bytes, byte-identical through both the Herdr ANSI read and a plain tmux capture:
  ```
  033 [ 0 m 033 [ 3 8 ; 2 ; 1 5 3 ; 1 5 3 ; 1 5 3 m 342 235 257 302 240 033 [ 0 m
  ```
  The blank is drawn at luminance 153, above the de-emphasis threshold, so `fm_composer_strip_ghost` correctly keeps it as real text. Every trim in the shared classifier and its callers is ASCII-only, so a genuinely empty Claude composer classified `pending`.

- **Masking condition.** The composer verdict only decides delivery where no stronger signal exists. `fm_backend_herdr_send_text_submit` confirms from **native agent state** whenever the target is legibly idle before Enter, and only falls back to reading the composer when the target was **already mid-turn**. Every ordinary steer took the immune path. The paired-review barrier release takes the other one, because both roles sit in a blocking wait for the release file when it is sent.

- **Visible symptom.** `verdict=pending` on a message that had already landed, and a pair aborted between its two releases.

The post-submit composer is the same padded-empty row plus a **dim** `Press up to edit queued messages` placeholder, which ghost stripping already removes correctly. One fault explains the pre-send read, the post-send read, and the retry exhaustion.

## Proven path vs failing path

A known-good send and the failing send differ at exactly one branch: the pre-Enter agent-state baseline. `idle` goes to `fm_backend_herdr_wait_for_working` (never reads the composer, always worked). `working` goes to `fm_backend_herdr_composer_state` (reads the padded row, always failed). Pre-fix, an empty composer and a composer holding real text were **indistinguishable** - both `pending`:

```
PRE-FIX  classify 0 "❯<NBSP>"           = pending     <- empty composer
PRE-FIX  classify 0 "❯<NBSP>PAIR READY" = pending     <- real unsent text
```

## Disconfirming checks run

- **Is the NBSP a Herdr rendering artifact rather than Claude's?** No. A plain tmux capture of Claude, with no Herdr in the path, shows `342 235 257 302 240` - one row in the pane, the composer row.
- **Does Claude keep the text in the composer while busy (the OpenCode busy-queue shape)?** No. It queues the message, clears the composer, and shows the queued text above. So a busy-pane "still pending" reading is a genuine swallow, which is why this fix must **not** copy tmux's busy-queue fallback - that would convert this false negative into a false positive.
- **Was my first live guard actually testing a composer?** No - it accepted Claude's blank startup screen and its trust dialog, and passed without the fix. Rebuilt so the untouched verdict is only asserted after typed probe text becomes visible in that pane. It now fails on Claude pre-fix.

## The fix

Fold non-ASCII Unicode blanks to ASCII space in `fm_composer_classify_content` - the single fleet-wide owner the tmux, herdr, cmux and orca adapters all delegate to.

This can only move content that is **entirely** blank from `pending` to `empty`. Content carrying any visible character keeps its verdict, so unproven delivery still stops the pair and the away-mode injector still refuses to type over unsent input. Keying on the Unicode space-separator category rather than U+00A0 alone keeps this from recurring when a release picks a different blank.

No change to the pair protocol, and no weakening of delivery verification.

## Real-harness evidence

Against real claude 2.1.226 on herdr 0.7.3, through the production adapter function, both directions:

```
### DIRECTION 1 - accepted while the worker is mid-turn MUST confirm
pre-send: agent_status=working composer_state=empty
send_text_submit VERDICT=empty
### DIRECTION 2 - a genuine swallow MUST still refuse
typed but NOT submitted: composer_state=pending
after composer clear: composer_state=empty
```

Direction 1's probe was observably delivered - it arrived in the receiving agent's turn - while the pre-fix adapter called the same send unconfirmed.

## Coverage

- `tests/fm-composer-lib.test.sh` - the fold, every declared blank, the dead-shell refusal, and the divergence assertion (padded-empty is `empty` while padded-real-text stays `pending`) so the case cannot go vacuous.
- `tests/fm-backend-herdr.test.sh` - the busy-baseline submit path itself, both directions, on captures carrying the real bytes and Claude's real rule-delimited shape. Fails pre-fix.
- `tests/fm-composer-blank-drift-live-e2e.test.sh` - new opt-in real-harness guard (`FM_COMPOSER_BLANK_DRIFT=1`), registered in the `live-harness-optin` family; `bin/fm-composer-lib.sh` now selects that family. What a harness draws in an empty composer is a vendor surface, so a fixture can only ever replay a previous release.

Observed: claude 2.1.226 and pi 0.84.1 pass; codex 0.145.0 held a first-run trust screen and is reported unverified; opencode 1.17.20 draws a left-only vertical composer edge the tmux reader refuses as `unknown` - reported unverified, not failed, since `unknown` already denies both delivery confirmation and injection.

Compatibility reviewed across all four consuming adapters (suites pass); zellij does not use the classifier. `bin/fm-lint.sh` clean, `bin/fm-doc-audience-check.sh` clean.

## Notes for the reviewer, outside this change

- `tests/fm-backend-herdr.test.sh` already fails on `main` at `two same-labeled home workspaces with no launcher identity must refuse` (expected exit 3, got 1). Confirmed pre-existing on a clean HEAD; untouched here.
- `fm-pair-compose.sh` releases the driver and navigator in sequence and creates `ready` only after both. A failure on the second release still leaves a half-released pair needing manual repair. Out of scope here - the proven fix does not require it - but worth its own decision.
- opencode's composer being unreadable through the tmux reader means firstmate cannot confirm delivery to an opencode worker from composer state on that backend. Fail-safe, pre-existing, and not in this task's scope.


huynhtandat223:
Closing this accidentally opened upstream PR. The work must be delivered through the huynhtandat223/firstmate fork PR path.

huynhtandat223:
Closing: this was opened against the wrong repository. The work belongs on my fork and is being raised there instead. Apologies for the noise.

PR #2048, opened 2026-08-10
fix(watch): distinguish unacked signal re-exits from fresh detections

### LINKED ISSUE #2028: Supervision looks stalled when it is actually thrashing on unread worker reports
## Symptom

The operator repeatedly sees "monitoring stalled" together with several workers
"awaiting reply". Re-arming appears to do nothing. Measured today: the watcher
armed at 00:17:45 and exited 32 seconds later at 00:18:17.

## Root cause

`bin/fm-watch.sh` is one-shot by design. It exits as soon as it finds an
actionable signal, and relies on the Stop hook `bin/fm-claude-stop-autoarm.sh`
to arm the next cycle after firstmate handles the report.

Two conditions break that loop, and they compound:

1. **Interrupted turn.** When the operator sends a message while firstmate is
   mid-turn, the Stop hook never fires, so nothing re-arms. Measured gaps on
   this home: 346 min, 96 min, 37 min, against a normal gap under 1 min.

2. **Unconsumed signal backlog.** Once reports accumulate unhandled, every
   fresh arm re-detects the same backlog and exits within ~30 s. The state
   record confirms this: `reason=actionable-signal` on eight consecutive
   cycles, each carrying the same status files.

Condition 2 is the one that makes it look permanent. The watcher is not
stalled - it is firing constantly and exiting immediately on reports nobody
has read. Draining the queue and answering the workers made the very next arm
stay alive.

## Evidence

- `state/.watch-cycle-exits.log`: consecutive `exit_code=0 reason=actionable-signal`
  entries with `lock_after=pid:none successor=none`.
- `state/.claude-autoarm-epoch`: `outcome=rewake` with no arm entry after it.
- `bin/fm-pr-merge.sh` printed `WARNING: queued wakes pending` on two separate
  invocations, so the backlog was visible to other tooling but did not block.
- After `bin/fm-wake-drain.sh` plus handling the four finished workers, the
  next arm held its lock instead of exiting.

## Why the current backstop is not enough

`state/.claude-autoarm-failure-notified` and the turn-end guard cover the case
where arming *fails*. They do not cover the case where arming *succeeds* and
the cycle exits immediately because the previous reports were never consumed.
That path reports success at every step while supervision is effectively dead.

## Proposed change

Distinguish "exited because a NEW signal arrived" from "exited because an OLD
signal is still unread", and make the second case visible and self-correcting:

1. Track a per-signal acknowledgement cursor so a cycle exiting on an
   already-reported signal is recorded with a distinct reason, for example
   `reason=actionable-signal-unacked`, instead of being indistinguishable from
   a fresh detection.
2. When the same signal set causes N consecutive immediate exits, raise it as
   a supervision alarm on the same path as the existing watcher-liveness
   alarm, naming the exact unread task ids.
3. Make the interrupted-turn case recoverable without a Stop hook: any command
   that already detects `queued wakes pending` should also verify a live
   watcher and arm one when absent.

Acceptance:
- A test that queues a signal, arms, lets the cycle exit, arms again, and
  asserts the second exit is recorded as unacked rather than as a fresh signal.
- A test that N consecutive unacked exits produce the alarm with task ids.
- A test that an arm with a drained queue holds its lock beyond one cycle.

abhishekdubey331:
Fix kunchenguid/firstmate#2028: the watcher was firing on the same signal
set every cycle while the queue stayed un-drained, and there was no way to
tell a fresh detection from a re-fire on the same unread set. This change
makes the two cases distinct in the lifecycle ledger and turns three
consecutive re-fires into a single supervisor-visible alarm that names the
unread task ids, so the operator sees "the watcher is thrashing on X" rather
than "the watcher is stalled".

## The three changes (issue 2028)

1. **Per-signal acknowledgement cursor.**
   `state/.wake-acked-cursor` is a private newline-separated list of signal
   keys (status file basenames). The watcher appends a key on every queued
   signal wake; the drain removes the keys it consumes; the arm captures a
   strictly-before-the-cycle snapshot at
   `state/.wake-acked-cursor.snapshot.<watcher-pid>` so a cycle that appends
   to the cursor during its own run does not retroactively make its own
   cycle unacked. The existing lifecycle ledger format stays byte-stable -
   the new value is `reason=actionable-signal-unacked` (a new distinct
   reason), not a breaking change to the existing `reason=actionable-signal`
   field.

2. **Threshold-3 alarm naming the unread task ids.**
   `state/.wake-unacked-streak` is a counter incremented by each unacked
   cycle and reset by any fresh detection or terminal exit. When the streak
   crosses `FM_WATCH_UNACKED_THRESHOLD` (default 3), the arm writes
   `state/.guard-watcher-unacked-banner` carrying `unacked-streak=<n>
   keys=<ids>`. `bin/fm-guard.sh` then prints the `WATCHER THRASHING ON
   UNREAD SIGNAL` banner with the same episode-dedup as the existing
   watcher-stale banner, naming the unread task ids. One streak crosses
   one banner; banner dedup is the key (`unacked-exit:<ids>`), so a fresh
   signal set re-arms a full banner.

3. **Interrupted-turn recovery without a Stop hook.**
   `bin/fm-guard.sh` already detects queued wakes. Its warning now also
   verifies the watcher health and, when the queue is pending while the
   watcher is down, prints a second `WARNING` that points at
   `bin/fm-watch-arm.sh` as the self-recovery command, because a Stop hook
   will not refire unattended and the interrupted-turn case is the only
   window where the original recovery path is unreachable.

## Threshold N=3, justified

A single unacked exit is a transient: a drain is in flight, the queue has
records about to be consumed, or the cursor is briefly stale while the
drain's queue commit is mid-update. Two consecutive unacked exits is
unusual but possible if the consumer is slow on a steady load and two
cycles slip past between drains. Three consecutive unacked exits is the
issue 2028 failure mode: the watcher is firing constantly on reports
nobody has read, and any further delay is the operator noticing the
"stalled" surface only after it has been burning CPU for a while. The
threshold is configurable via `FM_WATCH_UNACKED_THRESHOLD` for homes with
different tradeoffs, but the captain-facing default is 3 to keep the alarm
non-chatty while still catching the failure mode on the first visible
lapse.

## How each of the three acceptance tests proves its criterion

The three tests live in `tests/fm-watch-unacked.test.sh` and assert purely
through the executable interface (real processes, real state files, real
banner output) so they stay valid against any behavior-preserving
refactor.

1. **`test_second_cycle_on_same_signal_set_is_unacked` proves criterion 1.**
   Two real arm cycles run in series against the same `state/task1.status`.
   The first cycle queues the signal and the arm records `reason=actionable-signal`.
   The fixture then appends a new line to the same status file (changing the
   watcher-side signature so the second cycle re-detects) and runs the arm
   again. The arm captures the snapshot before forking, then compares the
   cycle's wake keys against the snapshot. The test asserts both ledger
   rows exist with the expected reason strings, and that the classification
   is exactly the two distinct values.

2. **`test_n_consecutive_unacked_exits_raise_alarm` proves criterion 2.**
   The fixture pre-populates the cursor with `task1.status`, writes
   `state/task1.status`, and runs four real arm cycles. Each cycle appends
   to the file so the watcher re-detects, and the arm classifies each as
   unacked. The test asserts `state/.wake-unacked-streak` is at least 3,
   that `state/.guard-watcher-unacked-banner` is set with the unread task
   id, and that a real `bin/fm-guard.sh` invocation prints
   `WATCHER THRASHING ON UNREAD SIGNAL` in its stderr banner and names
   `task1.status` in the readable wording.

3. **`test_arm_with_drained_queue_holds_lock` proves criterion 3.**
   The fixture runs one arm cycle to populate the queue and cursor, then
   runs `bin/fm-wake-drain.sh` against the same state. The test asserts the
   queue is empty and the key is dropped from the cursor. It then forks a
   real arm in the background, waits for the watcher to take the singleton
   lock, and holds the assertion for six full poll cycles (with
   `FM_POLL=5`) that the watcher is still alive and the lock is still
   held, and that the lifecycle ledger contains no `actionable-signal-unacked`
   row - the watcher did not re-fire on a stale signal.

## Files

- `bin/fm-wake-lib.sh` - the cursor helpers.
- `bin/fm-watch.sh` - cursor maintenance on queue.
- `bin/fm-wake-drain.sh` - cursor maintenance on drain.
- `bin/fm-watch-arm.sh` - snapshot, classification, streak, banner marker.
- `bin/fm-guard.sh` - unacked-exit banner + arm-when-down warning.
- `tests/fm-watch-unacked.test.sh` - the three acceptance tests.
- `docs/architecture.md` - one paragraph in the event-driven-supervision
  section.

## Verification

- `bin/fm-lint.sh` passes (pinned ShellCheck 0.11.0).
- `tests/fm-watch-unacked.test.sh` passes all three acceptance tests.
- `tests/fm-watcher-lock.test.sh`, `tests/fm-watch-arm.test.sh`,
  `tests/fm-wake-queue.test.sh`, `tests/fm-wake-drain-open-decisions.test.sh`,
  `tests/fm-guard-stale-banner.test.sh` continue to pass against the
  modified arm and guard.

closes #2028


abhishekdubey331:
Closing this one. The re-arm recovery work in #2065 lands in the same area and makes durable wakes survive watcher downtime, which covers the thrash described in #2028 from a different angle.

Rather than rebase this branch across that change and carry a competing fix, I am closing it. Issue #2028 stays open with the measured evidence, so the symptom is still on record if #2065 does not fully settle it.

PR #2050, opened 2026-08-10
fix(bin): accept a decision key on either side of the status colon

KCooke2222:
## The defect

`_fm_decision_key()` in `bin/fm-classify-lib.sh` looked for the optional `[key=<slug>]` token only in the text *before* a status line's first colon.

The standard crewmate brief tells workers to append `{state}: {one short line}`, so a keyed decision is naturally written with the token leading the note, *after* the colon:

```
needs-decision: [key=api-shape] pick REST or RPC
```

Every line of that shape folded to `key=default`. Distinct decisions collided under one key, and an already-answered decision kept looking open, which produced repeated false-positive stale wakes in live supervision.

## Where the fix belongs, and why

The fix is in the **parser**, not in the brief scaffold.

- Status files already written on disk carry the post-colon shape. A scaffold-only fix would only change what *future* workers write, leaving every existing task's decision history broken.
- Firstmate's own writers - `bin/fm-send.sh`, `bin/fm-decision-hold.sh`, `bin/fm-pending-reply-lib.sh` - emit the pre-colon shape, which must keep working exactly as before.

So both placements now resolve to the same key. Neither shape was fixed by breaking the other, and the brief scaffold's wording is deliberately unchanged.

`status_line_note()` now strips a leading key token as well. That is required rather than cosmetic: the reserved-key namespace rule (`_fm_decision_key_transition_allowed`) decides whether a transition is allowed by matching the note's own leading vocabulary token, so without stripping, a post-colon reserved-key line could not open or close its own decision.

## Malformed slugs fall back to `default`

Review caught a real regression in the first version of this change: a post-colon token with a malformed slug (`[key=oauth token]`) made `_fm_decision_key()` return non-zero, so `_fm_decision_fold_line()` dropped the line entirely and the `needs-decision` never opened at all. On the base commit that same line folded under `default` and did surface.

`_fm_decision_key()` is the gate on whether a line is a decision transition at all, so a failure return there means a typo'd key silently deletes a decision from the open set - the one thing this fold exists to prevent. A malformed slug now falls back to `default` exactly as a missing token does, under **both** placements. That restores the base behaviour and aligns the key parser with `status_line_note()`, which already leaves an unrecognised token in the note as ordinary text.

## Other changes

- `FM_OPEN_DECISIONS_FOLD_VERSION` 2 -> 3, as the library's own contract requires when fold semantics change, so persisted cursors from the old interpretation are discarded and rebuilt.
- The shared token-recognition helper reports through a caller-declared local variable instead of stdout, so the whole-file fold does not pay an extra subshell fork per status line - it runs once per task on every fleet snapshot.

## Tests

All in `tests/fm-wake-drain-open-decisions.test.sh`, driving the real `bin/fm-wake-drain.sh` over crafted status logs and asserting on its printed output, never on implementation source text:

- The reproduction: two distinct post-colon keyed decisions must name their own keys. Written first and confirmed failing against the unfixed parser, where both collapsed into a single unkeyed entry.
- Open/resolved pairing under both placements and both cross-shape combinations, plus an unrelated bare resolution that must not close a keyed decision.
- The reserved-key namespace rule under the post-colon shape.
- A malformed slug under either placement still surfaces under `default`, and stays closable by a bare resolution.

## Scope

One parsing defect. No classifier restructuring, no generalised status grammar, no status-line schema, and no change to the brief scaffold's wording.

### DIFF
--- a/bin/fm-classify-lib.sh
+++ b/bin/fm-classify-lib.sh
@@ -170,7 +170,8 @@ status_is_paused_or_captain_held() {  # <status-line>
 # line}" shape naturally puts the token at the head of the note. Accepting one
 # placement only silently folds every line of the other shape onto "default",
 # colliding unrelated decisions and re-surfacing answered ones.
-# A line with no token uses the key "default", preserving the historical
+# A line with no token - or one whose slug is malformed, which is not a token
+# this grammar recognises - uses the key "default", preserving the historical
 # one-open-decision-per-task behavior (a bare "resolved:" closes "default").
 # The three parsers are pure reads of a single line; the verb parser strips any
 # key token before the colon so the leading word is recovered cleanly, and the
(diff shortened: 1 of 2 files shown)


kunchenguid:

PR #2055, opened 2026-08-10
fix: resurface durable supervision work after re-arm

kunchenguid:
## Intent

Fix FIRSTMATE shared tracked supervision/startup infrastructure so that after the primary watcher re-arms following any accepted down stretch, whether caused by a blocking tool or ordinary idle downtime, nothing queued during that window is lost: every durable queued wake must drain and every still-open decision must re-surface in OPEN DECISIONS, including keyed remote-secondmate decisions ingested through parent-replies.

First reproduce the captain-reported miss end to end as closely as feasible: a durable queued wake and a still-open remote-secondmate parent-replies decision survive a watcher-down interval but were not surfaced by re-arm, only by an unrelated later drain. Diagnose the proven root cause by tracing parent-replies ingestion through task status logs and the incremental open-decisions fold, confirming durability across ingest surfaces, and checking re-arm and cursor behavior. Preserve the one-owner fold contract and the cursor correctness invariant. Do not attempt to keep the watcher live while a blocking tool suspends a turn.

Ship the smallest targeted deterministic, idempotent re-arm recovery so re-arm triggers the existing durable drain and fold path rather than relying on an incidental later drain. Remote decisions must use the authoritative task status-log surface already owned by the fold, not a second ledger. Pin the E2E reproduction as portable tests using durable on-disk state through real interfaces, showing both durable queued wakes and remote parent-replies open decisions re-surface after a down interval, including decision-only recovery and a persistent successor remaining live. Review affected supported watcher adapters/backends, keep lint clean, and provide the reproduction and post-fix E2E evidence in the PR body. Do not merge the PR; yolo is off. Escalate any ask-user finding to Firstmate instead of deciding it, and never use --yes.

## What Changed

- Trigger the existing durable wake-drain and OPEN DECISIONS fold when a watcher re-arms after downtime, including keyed remote-secondmate decisions ingested through parent replies.
- Keep queued wakes durable until generation-bound post-handling acknowledgement, preserving recovery evidence across watcher lock handoffs and interrupted handling.
- Add portable on-disk regression coverage for queued-wake replay, decision-only recovery, stale acknowledgements, and persistent successor liveness.

## Risk Assessment

⚠️ Medium: No new material defects were substantiated, but the change substantially expands a concurrency-sensitive durable wake and acknowledgement state machine across multiple supervision paths.

## Testing

Inspected the base-to-target change, ran the four smallest affected suites, corrected and retried one fixture that had not adopted generation-bound acknowledgement, and captured a focused real-interface E2E transcript showing both queued wakes and the keyed remote decision resurfacing after downtime; all targeted checks now pass.

<details>
<summary>Evidence: Watcher re-arm E2E transcript showing durable wakes, OPEN DECISIONS, decision-only recovery, empty acknowledged queue, and live successor</summary>

```text
=== Re-arm user-visible wake ===
check: rearm-resurface
=== Durable wake drain + OPEN DECISIONS ===
1786341299	1	check	remote-reply-ios	check: process-event result captured: remote-reply-ios:7
1786341299	2	check	startup-network	check: startup-network
ios [key=remote-signoff] needs-decision: remote secondmate is held for captain sign-off
OPEN DECISIONS: close one by answering it: bin/fm-send.sh <task> --resolve-key <key> '<answer>'
=== Decision-only later re-arm ===
watcher: started pid=82287 (beacon fresh)
check: rearm-resurface
ios [key=remote-signoff] needs-decision: remote secondmate is held for captain sign-off
OPEN DECISIONS: close one by answering it: bin/fm-send.sh <task> --resolve-key <key> '<answer>'
WAKE_ACK_REQUIRED: after handling completes run bin/fm-wake-drain.sh --ack-through 0 --recovery-generation 83390.1786341303.ZIXVSi
=== Authoritative parent status log ===
needs-decision [key=remote-signoff]: remote secondmate is held for captain sign-off
=== Durable queue bytes after acknowledgement: 0 ===
=== Persistent successor after acknowledgement ===
watcher: started pid=85240 (beacon fresh)
```
</details>

## Pipeline

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>
<summary>🔧 **Review** - 3 issues found → auto-fixed (14) ✅</summary>

- 🚨 `bin/fm-watch.sh:809` - Recovery checks the queue and later removes the downtime marker without one atomic boundary. A real startup-network publisher can append after the size/read check but before line 816; the marker is then removed, and ordinary watcher polling never examines that queue row, recreating the reported miss. Hold the wake-queue lock through the final empty decision and marker consumption, or retain recovery state until the shared drain confirms consumption.
- 🚨 `bin/fm-watch.sh:748` - Downtime is recorded only from the EXIT trap. If the tracked watcher is abruptly reaped (for example SIGKILL or process-group teardown that bypasses traps), no marker is written; wakes queued while it is down survive, but the replacement returns at line 808 and never triggers the drain. Record/infer the predecessor generation at a durable shared arm/watch boundary rather than relying solely on cleanup execution.
- 🚨 `bin/fm-watch.sh:812` - The re-arm gate uses the whole-file scan instead of the cursor-backed fold&#39;s trusted-open fallback. If a status read transiently fails or becomes partial while a previously folded decision remains in the cursor, this scan can return empty and line 816 permanently consumes the marker; the existing drain would have preserved and printed the trusted open set. Trigger recovery through a cursor-safe shared boundary without creating a competing fold owner.

🔧 Fix: Make watcher rearm recovery durable and cursor-safe
- 🚨 `bin/fm-watch.sh:751` - This contradicts the required behavior “after the primary watcher re-arms following any accepted down stretch ... nothing queued during that window is lost.” In the autoarm model, a watcher can deliver wake A, cleanup removes the marker here, the handling turn drains A, and a durable remote/startup wake can then arrive before the next turn-end rearm. That successor sees neither a marker nor a stale lock, so it remains asleep and the later wake is still missed. Preserve recovery state for the entire watcher-down generation, including downtime following a delivered wake, while preventing persistent successor loops at the shared drain/arm handoff.
- 🚨 `bin/fm-watch.sh:740` - Abrupt-exit recovery still depends on this process receiving FM_LOCK_RECOVERED_PID, but fm-pr-check-migrate.sh runs first and can acquire, steal, and release the stale watcher lock when migration is required. After a SIGKILL with no marker, that subprocess consumes the only recovery evidence; this watcher then acquires a clean lock, leaves WATCHER_RECOVERY_PENDING=0, and misses durable queued work. Persist the recovery signal whenever any supported preflight/lock owner consumes a dead watcher lock, rather than relying on a process-local return variable.
- ⚠️ `bin/fm-watch.sh:754` - Creating the marker with `: &gt; &#34;$WATCHER_DOWNTIME_MARKER&#34;` follows an existing symlink and can truncate an arbitrary file writable by the watcher. Create the marker without following links, such as via a same-directory temporary file and atomic rename, or reject/remove a symlink first.

🔧 Fix: Persist safe recovery markers across migration lock recovery
- 🚨 `bin/fm-watch.sh:751` - This contradicts the required behavior “after the primary watcher re-arms following any accepted down stretch ... nothing queued during that window is lost.” After a watcher delivers wake A, cleanup removes the recovery marker; under the Claude auto-arm path, the handling turn can drain A, then receive a durable wake B before its turn-end successor starts. That successor sees neither a marker nor a stale lock, so B remains missed until an unrelated drain. Preserve recovery state across the entire delivered-wake handling gap and consume it at the shared drain/arm handoff without causing persistent-successor loops.
- ⚠️ `bin/fm-watch.sh:754` - Cleanup ignores recovery-marker publication failure and then releases the watcher lock. For example, `fm_recovery_marker_publish` deliberately fails when `.watcher-down` is an ordinary directory; after a TERM exit, the successor therefore sees neither a marker nor a stale lock and can miss downtime work. If publication fails, retain stale-lock recovery evidence or fail closed visibly rather than releasing the last recovery signal.

🔧 Fix: Retain stale lock when recovery marker publication fails
- 🚨 `bin/fm-watch.sh:751` - The required guarantee says “after the primary watcher re-arms following any accepted down stretch ... nothing queued during that window is lost,” but this branch removes recovery state after every delivered wake. In the Claude auto-arm path, wake A closes the watcher, its handling turn drains A, wake B is durably queued before the turn-end successor arms, and that successor sees neither a marker nor a stale lock, so B remains unsurfaced. Preserve recovery state through the delivered-wake handling gap and consume it at the shared drain/arm handoff.
- 🚨 `bin/fm-pr-check-migrate.sh:314` - After migration steals a dead watcher&#39;s lock, a transient marker-publication failure releases the replacement lock here, discarding the only stale-owner evidence. If the condition clears and startup is retried, migration acquires a clean lock, publishes no marker, and the watcher can miss downtime work. Retain stale-lock evidence on this failure path as watcher cleanup now does.
- ⚠️ `bin/fm-watch.sh:752` - An ordinary directory at `.watcher-down` is treated as pending recovery, but `rm -f` cannot consume it and the failure is ignored. Every persistent successor therefore immediately emits `rearm-resurface` and exits instead of remaining live. Reject or quarantine malformed markers before arming, or consume recovery state with a checked operation that cannot leave the marker active.

🔧 Fix: Preserve delivery-gap recovery and quarantine malformed markers
- 🚨 `bin/fm-wake-lib.sh:493` - The acknowledged-marker path releases the wake-queue lock before consuming the marker. A publisher can durably append after the empty check at line 490 but before the marker move at line 499; the marker is then removed and the live watcher never examines that queued row, recreating the reported miss. Hold the queue lock through the final empty decision and marker consumption.
- ⚠️ `bin/fm-wake-drain.sh:148` - The drain ignores recovery-marker acknowledgement failure. If the atomic replacement in fm_recovery_marker_ack fails, the marker remains pending:handling; successors consequently keep waiting rather than recovering later queued work, while this drain reports success. Handle acknowledgement failure explicitly so recovery state cannot remain spuriously active.

🔧 Fix: Serialize recovery consumption and report acknowledgment failures
1 error still open:
- 🚨 `bin/fm-watch.sh:737` - This contradicts the required guarantee that after re-arm following “any accepted down stretch ... nothing queued during that window is lost.” The new recovery gate recognizes only `FM_LOCK_RECOVERED_PID` or a marker, but `fm-watch-arm.sh --restart` line 397 removes an identity-mismatched stale watcher lock when its dead PID has been reused by an unrelated live process. After a SIGKILL leaves no marker, that removal erases the only downtime evidence; the launched watcher acquires a clean lock and can sleep over durable queued work. Publish recovery state at `clear_stale_recorded_watcher_lock` before removing the verified stale watcher lock, or retain equivalent evidence through that boundary.

🔧 Fix: Centralize recovery publication before clearing watcher evidence
2 errors still open:
- 🚨 `bin/fm-wake-lib.sh:512` - This contradicts the requirement that “nothing queued during that window is lost: every durable queued wake must drain.” Arm-check consumes the marker while holding the queue lock, then releases it; a publisher already waiting on that lock subsequently appends after marker consumption, while the successor remains live and never examines the row. The new test demonstrates this exact missed state at tests/fm-watch-arm.test.sh:430-433 but treats it as success. Make the append/arm handoff preserve recovery evidence for a waiting publisher, and require the test to observe the concurrent wake being surfaced and drained.
- 🚨 `bin/fm-wake-lib.sh:626` - The required guarantee after “any accepted down stretch” remains vulnerable because FM_LOCK_RECOVERED_PID is assigned only after line 622 removes the stale watcher lock and line 624 creates its replacement. If the reclaiming watcher or migration process is killed between those operations, no marker or watcher lock remains; the next watcher starts clean and can sleep over queued work. Publish recovery evidence inside the stale watcher-lock reclaim boundary before removal, rather than relying on a process-local flag produced after replacement.

🔧 Fix: Guarantee recovery evidence across queue and lock handoffs
1 error still open:
- 🚨 `bin/fm-wake-lib.sh:770` - This contradicts the required criterion that “every durable queued wake must drain.” An external publisher such as startup-network can append a durable row here and then be killed, or fail marker publication at line 773; its caller ignores that failure, while the live watcher never examines the queue and has neither a downtime marker nor a stale watcher lock to trigger recovery. Additionally, marker publication removes the old marker before replacement, creating another crash window that can erase existing evidence. Under the queue lock, atomically publish recovery evidence before committing the row, replacing the marker with one rename and no remove-before-rename gap.

🔧 Fix: Publish recovery evidence before durable wake commits
1 error still open:
- 🚨 `bin/fm-wake-lib.sh:403` - Every durable wake append and watcher recovery transition now invokes `perl`, but Perl is not part of the supported universal toolchain and other project code explicitly treats it as optional. On a supported host without Perl, marker publication fails before the queue row is committed; ignored callers such as startup-network silently drop the wake, while watcher cleanup/reclaim cannot release or recover the lock. Use an already-required runtime or a portable same-directory atomic replacement implementation.

🔧 Fix: Replace recovery marker Perl dependency with Node
1 error still open:
- 🚨 `bin/fm-wake-lib.sh:495` - This contradicts “nothing queued during that window is lost: every durable queued wake must drain.” After a watcher delivers wake A, cleanup publishes `pending:handling`; if its handling turn is interrupted before `fm-wake-drain.sh` acknowledges it, the persistent successor treats the marker as `wait` forever and never surfaces A. Only a later unrelated publisher or watcher restart changes the marker, recreating the incidental-later-drain failure. Give pending handling state a durable owner/lease or let the successor recover when no handling owner remains, at the shared drain/arm handoff.

🔧 Fix: Keep interrupted wakes durable until handling acknowledgment
2 errors still open:
- 🚨 `bin/fm-wake-lib.sh:495` - This contradicts the required “persistent successor remaining live” behavior. Pi and OpenCode start and verify a successor before delivering the original wake; that successor observes `pending:handling` with the original row still queued, selects `recover`, and exits with another rearm-resurface wake. Each replacement can repeat this before handling begins. Preserve the pre-delivery successor instead of treating the still-durable original row as a new recovery event.
- 🚨 `bin/fm-wake-drain.sh:144` - The accepted requirement says a delivered wake must not leave durable state until handling is fully complete/acknowledged, but the initial drain atomically empties the queue before returning control to the handling turn. If the turn is interrupted immediately after the drain returns, the marker is already acknowledged and the row is gone, so a successor cannot re-drain it. The new regression only kills the drain before this line and therefore does not cover an interrupted handling turn after drain completion. Add a real post-handling acknowledgement boundary, or escalate the concrete lifecycle decision as requested rather than treating drain completion as handling completion.

🔧 Fix: Add post-handling durable wake acknowledgements
- 🚨 `bin/fm-wake-drain.sh:145` - This contradicts the required criterion that “every still-open decision must re-surface in OPEN DECISIONS.” During decision-only recovery the queue is empty, so this branch acknowledges the recovery marker before printing OPEN DECISIONS. If the handling turn is interrupted after the drain returns, the successor sees an acknowledged marker and empty queue, remains asleep, and the unchanged open decision is again stranded until an unrelated drain. Keep decision-only recovery pending until an explicit post-handling acknowledgement, rather than auto-acknowledging it on the empty-queue path.
- ⚠️ `bin/fm-wake-drain.sh:168` - The new drain contract leaves emitted rows queued until this command is run, but bin/fm-afk-return.sh still invokes the drain, immediately removes its catch-up gate, and reports “catch-up clear” without executing or retaining an acknowledgement step. A clean AFK return with queued non-blocking wakes can therefore permit ordinary work while the queue and recovery marker remain active, causing successor replay. Integrate the post-handling acknowledgement into the AFK return gate before declaring catch-up complete.

🔧 Fix: Enforce post-handling acknowledgement across recovery and AFK return
1 error still open:
- 🚨 `bin/fm-wake-drain.sh:125` - This contradicts “after the primary watcher re-arms following any accepted down stretch ... every still-open decision must re-surface.” The printed acknowledgement carries only a queue sequence; when it runs, this branch snapshots and acknowledges whichever marker is current. If a watcher goes down after the drain but before handling completes, its newer downtime marker can therefore be acknowledged by the older handling turn. The next successor sees an acknowledged marker and empty queue and sleeps without re-surfacing still-open decisions from that down stretch. Bind the acknowledgement command to the marker token captured by the original drain and CAS only that generation; a newer marker must remain pending.

🔧 Fix: Bind wake acknowledgements to recovery generations
2 errors still open:
- 🚨 `tests/fm-wake-queue.test.sh:192` - These regressions still require concurrent drains to consume rows exactly once and later wait for the queue to become empty before acknowledgement (lines 540-566). The new contract deliberately retains rows until the generation-bound post-handling ACK, and `.wake-queue.empty.*` is no longer created, so these tests cannot validate the implemented behavior. Update them to assert replay before ACK and consumption only after the emitted ACK command succeeds.
- 🚨 `tests/fm-watch-arm.test.sh:331` - The interrupted-decision regression acknowledges using the generation captured before the successor recovery. That successor&#39;s handling cleanup and the replay drain each publish a newer marker generation, so the CAS must reject this command. Parse and acknowledge the generation emitted by `replayed-decision-drain.err`; otherwise the required interrupted decision-recovery regression asserts an impossible outcome.

🔧 Fix: Align wake regressions with generation-bound acknowledgements

</details>

<details>

✅ No issues found.
- `git diff --stat --name-only 85e750ab9b76df275c1f6b9e2bc95b671955bae9..071797253939f96153cefa72ca3a5e866b88c315`
- <code>`./bin/fm-test-run.sh tests/fm-watch-arm.test.sh tests/fm-wake-queue.test.sh tests/fm-watcher-lock.test.sh tests/fm-afk-return.test.sh` (identified an outdated lifecycle-ledger fixture expectation)</code>
- <code>`./bin/fm-test-run.sh tests/fm-watcher-lock.test.sh` (passed after the test-only acknowledgement fix)</code>
- <code>Focused real-interface E2E reproduction via temporary `tests/.evidence-watch-arm.test.sh`, exercising remote parent-replies ingestion, watcher termination/re-arm, durable wake draining, decision-only recovery, acknowledgement, and persistent successor startup</code>

</details>

<details>

✅ No issues found.

</details>

<details>

🔧 Fix: Resolve ShellCheck warnings in recovery and watcher tests

</details>

<details>

✅ No issues found.

</details>


kunchenguid:
Superseded by https://github.com/kunchenguid/firstmate/pull/2065 after the original validation run hit a custody deadlock. The replacement PR carries the recovered history, complete hazard-to-commit map, and green checks.

PR #2064, opened 2026-08-10
docs: name capacity-limited shared build/test infra as a serializing condition

MathRoda:
## Summary
- Patches AGENTS.md section 7's existing serialize-only-for sentence (~line 280) to name capacity-limited shared build or test infrastructure (a device simulator or emulator, a device farm, a single machine's build capacity) as a concrete shared-mutable-external-state condition that requires serializing concurrent work.
- States that partitioning such a resource per worker is the cheaper answer than serializing, wherever it can be partitioned.

## Why
Three same-day failures on one machine came from dispatching two iOS workers concurrently: a validation test-step failure initially misdiagnosed as a code defect (actually another worker's xcodebuild holding the same simulator), two workers driving the same named simulator, and 22 orphaned simulator clones (52GB) from abnormal exits stalling unrelated work. The existing rule already covered this case as "shared mutable external state" but was never read that way, so this patches the existing sentence rather than adding new prose.

## Test plan
- [x] `bin/fm-doc-audience-check.sh` passes
- [x] no-mistakes pipeline: intent, rebase, review, test, document steps all passed (lint skipped: ShellCheck unavailable in this environment, no scripts changed)


MathRoda:
Closing at the author's request.

PR #2093, opened 2026-08-10
fix(ff): require a work-tree root before fast-forwarding a target

leecalcote:
## Summary

`ff_target` in `bin/fm-ff-lib.sh` guarded its target with `git rev-parse --is-inside-work-tree`.
Git repository discovery walks upward, so that check succeeds for any plain directory merely nested inside a repository.
`ff_target` therefore accepted such a directory and ran every subsequent `git -C "$dir" ...` against the **enclosing** repository under the target's label, fast-forwarding or pruning a repository the caller never named.

`ff_target` is the shared fast-forward machinery behind `/updatefirstmate` (`bin/fm-update.sh`) and the local-HEAD secondmate sync (`bin/fm-spawn.sh` on launch, `bin/fm-bootstrap.sh` on startup).

This change requires the target to **be** the work-tree root.
A non-root reports `skipped: not a clone root (inside <toplevel>)`, a benign skip rather than a silent advance of the wrong repository.
A directory in no repository at all still reports `skipped: not a git repo`, so that path is unchanged.

## Relationship to #2044

This is the **complement** to #2044, not a competing or duplicate fix.

#2044 fixes the same upward-discovery root cause for the `projects/` clone sweep in `bin/fm-fleet-sync.sh`.
That sweep fix is deliberately left to #2044, and `bin/fm-fleet-sync.sh` and `tests/fm-fleet-sync.test.sh` are intentionally untouched here.
What this PR adds is the identical hole on a different code path: `ff_target`, which #2044 does not touch at all.

## Scope - three files

- `bin/fm-ff-lib.sh` - the clone-root guard in `ff_target`.
- `tests/fm-secondmate-sync.test.sh` - the regression. This suite already exercises `ff_target`.
- `docs/architecture.md` - two sentences in the "Self-updates stay safe" section. This repo's conventions require documentation to stay accurate when behavior changes, and that section is the owner of self-update fast-forward safety. It says nothing about `bin/fm-fleet-sync.sh` or `projects/`, so it does not overlap #2044.

## Why both sides of the path comparison are canonicalized

Comparing a shell-resolved path against git's raw `--show-toplevel` output is not portable.
On Git Bash/MSYS, which this repo supports, git reports `C:/Users/...` while `pwd -P` reports `/c/Users/...`, so every target would be misclassified and self-update would silently become a no-op.
This matches the existing worktree-root checks in `bin/fm-control.sh` and `bin/fm-spawn.sh`, which canonicalize both sides for the same reason.
If the toplevel cannot be resolved physically the code falls back to the raw value rather than failing the sweep, consistent with `bin/fm-control.sh`.

## Test plan

- [x] `bin/fm-lint.sh` (ShellCheck 0.11.0)
- [x] `bin/fm-test-run.sh tests/fm-secondmate-sync.test.sh`
- [x] `bin/fm-test-run.sh tests/fm-update.test.sh` (the other suite covering `ff_target` callers)
- [x] The regression was verified to fail without the `bin/fm-ff-lib.sh` change and pass with it.

The regression pins the safety property rather than the message.
It points `ff_target` at a container directory inside a behind secondmate home and asserts the enclosing worktree's HEAD does not move.
Without the guard it fails with `FF_STATUS: expected skipped, got 'updated'` - the enclosing home is genuinely fast-forwarded.
The container sits under the gitignored `projects/` dir deliberately: anywhere else the enclosing worktree would be dirty, the existing dirty guard would stop the advance for an unrelated reason, and the test would pass with or without the fix.
The test asserts up front that the fixture is not vacuous on either count.

## Known follow-ups, deliberately not included here

- **Digest suppression, keyed to #2044.** Once #2044 merges, whichever skip reason it emits should be added to `fleet_sync_relay_filtered_output` in `bin/fm-bootstrap.sh`, and named in `.agents/skills/bootstrap-diagnostics/SKILL.md`. Otherwise every non-clone entry under `projects/` emits an actionable `FLEET_SYNC:` digest line on every session. This cannot be done here because the filter must match #2044's actual message, not this PR's.
- **Two path canonicalizers in `bin/fm-ff-lib.sh`.** The hardened inline form added here (`CDPATH=''`, `cd -P`, `--`) now stands beside the older unhardened `resolve_path` at line 64, which `process_secondmate` still uses. Consolidating them was considered and declined: `resolve_path` has an existing consumer, so changing it would turn a targeted fix into a library refactor. The older path keeps its current behavior, so this is not a regression.
- **Duplicate skip-condition enumeration in `.agents/skills/updatefirstmate/SKILL.md`.** The same list appears at line 21 and again at line 58 under Safety. This predates the change and neither copy is made stale by it, so collapsing one into a pointer was left alone rather than widening into a skill rewrite.
- **Same root cause in `bin/fm-home-seed.sh`.** `clone_project` and `validate_seed_project` gate on `git -C "$src" rev-parse --is-inside-work-tree`, so a non-clone `projects/` entry passes and `source_origin_url` then reads the enclosing checkout's origin - seeding a secondmate home would clone firstmate itself under that project's name. Same root cause, worse blast radius, lower reachability. Out of scope here; raised so it is not lost.

### DIFF
--- a/bin/fm-ff-lib.sh
+++ b/bin/fm-ff-lib.sh
@@ -279,8 +279,21 @@ ff_target() {
     echo "$label: skipped: not a directory"
     return 0
   fi
-  if ! git -C "$dir" rev-parse --is-inside-work-tree >/dev/null 2>&1; then
+  # Git repository discovery walks UPWARD, so --is-inside-work-tree is true for a
+  # target that merely SITS INSIDE a repository, and every command below would then
+  # fast-forward or prune that enclosing repository under this target's label.
+  # Every ff_target caller passes a work-tree root (a clone, a linked worktree, or a
+  # secondmate home), so require the root itself and skip anything else benignly.
+  local dir_real dir_toplevel dir_toplevel_real
+  dir_real=$(CDPATH='' cd -P -- "$dir" 2>/dev/null && pwd -P) || dir_real=$dir
+  dir_toplevel=$(git -C "$dir" rev-parse --show-toplevel 2>/dev/null) || dir_toplevel=
+  if [ -z "$dir_toplevel" ]; then
     echo "$label: skipped: not a git repo"
+    return 0
+  fi
+  dir_toplevel_real=$(CDPATH='' cd -P -- "$dir_toplevel" 2>/dev/null && pwd -P) || dir_toplevel_real=$dir_toplevel
+  if [ "$dir_toplevel_real" != "$dir_real" ]; then
+    echo "$label: skipped: not a clone root (inside $dir_toplevel)"
     return 0
   fi
 
--- a/bin/fm-ff-lib.sh
+++ b/bin/fm-ff-lib.sh
@@ -266,8 +266,9 @@ live_secondmate_meta_records() {
 #                  already exist in the target's object store, which it always does
 #                  for a worktree of this same repo; a standalone clone that lacks
 #                  it is skipped rather than fetched.
-# Guards are identical in both modes: ff-only (never force/merge/stash); skip a
-# dirty, diverged, or wrong-branch target and leave its work untouched.
+# Guards are identical in both modes: ff-only (never force/merge/stash); the target
+# must be the work-tree root itself (see the clone-root check below); skip a dirty,
+# diverged, or wrong-branch target and leave its work untouched.
 FF_STATUS=""
 FF_INSTR=""
 ff_target() {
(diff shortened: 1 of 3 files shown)


leecalcote:
Closing this myself: I opened it manually to work around a tooling problem on my side, and it therefore lacks the deterministic `no-mistakes` signature that CONTRIBUTING requires for PRs targeting `main`. Rather than leave an unsigned PR open against your gate, I am closing it and will reopen through the `no-mistakes` pipeline once it can stamp the body properly. The branch and the change are unaffected. Apologies for the noise.

PR #2105, opened 2026-08-10
fix(workgraph): scope legacy-dispatch refusal to genuine worktree overlap

garynewball:
## Intent

Fix the narrow bug in `fm_workgraph_enforce_dispatch` where a new legacy (non-WorkGraph-bound) dispatch is unconditionally refused whenever `state/*.meta` has any entries, regardless of whether the existing meta actually conflicts with the proposed task.

## What changed

`fm_workgraph_enforce_dispatch` in `bin/fm-workgraph-dispatch-lib.sh` now treats existing task metadata as compatible-by-default in legacy mode. Two conditions remain refusal-worthy:

1. The existing meta's recorded worktree matches the proposed task's project path.
2. (WorkGraph mode only) The existing meta has an active held lease, the sealed contract is incompatible, or the endpoint is live/ambiguous without a valid lease.

A definitively absent (dead) recorded backend endpoint is skipped in either mode.

## Why

The compatibility model in `docs/workgraph.md` already describes WorkGraph-bound dispatch as overlapping-only. The legacy code path inherited the same gate but applied it as "any meta is exclusive" instead of "any genuinely overlapping/active meta is exclusive," blocking independent work on unrelated worktrees and breaking the simplest direct end-to-end path.

## Regression coverage

Three new tests in `tests/fm-workgraph-dispatch.test.sh`:

- `test_spawn_legacy_path_rejects_before_backend_creation` — legacy mode refuses when existing meta's worktree matches the proposed project.
- `test_spawn_legacy_path_admits_independent_worktree` — legacy mode admits when existing meta's worktree differs from the proposed project.
- `test_spawn_legacy_path_skips_dead_endpoint_metadata` — legacy mode skips existing meta whose recorded backend endpoint is dead.

The pre-existing `test_compatible_dispatch_and_capacity` failure is unrelated to this fix (lease-store infrastructure issue in the test harness) and reproduces on the unmodified base.

## Validation

- bash syntax check passed.
- Three new regression tests pass when run against this branch.
- Full test-suite run reproduces the pre-existing `WG-L-E-STORE` failure identically with and without the fix, confirming it is environmental and unrelated.

## Scope

Narrow bugfix. Does not touch any of: spawn/teardown/AGENTS.md/active-task semantics, WorkGraph-mode behavior, parallelism capacity logic, contract binding, or the dispatch lock primitive.


garynewball:
Closing: this PR cannot be merged because canonical origin/main (f74d9e4) does not contain commit 6e9e27e or the WorkGraph infrastructure. The diff against origin/main adds the entire WorkGraph feature, which is well beyond the narrow legacy-dispatch-exclusive scope described in the task brief. The fix commit itself (ce357c8) is preserved on branch fix/fm-workgraph-enforcement-blocks-dispatch-20260810 in garynewball/firstmate for re-targeting once the underlying feature lands on canonical main.

PR #2108, opened 2026-08-10
feat(brief): generate the since-last-review marker contract

ecloin:
## Problem

Local-only ship work is reviewed interactively as a diff against a fixed base, but the generated brief said only "report ready in branch `fm/<id>`" and defined no review contract. The review command was therefore hand-built per task, and it has broken four ways:

- whole-changeset diffs instead of since-last-review
- a compare ref anchored to a moving ref, which catches up with the branch tip and renders an empty diff
- invalid two-ref command forms (`hunk diff` takes one base ref and diffs it against the working tree)
- a rebase rewriting shas, stranding an existing `<id>-reviewed` tag at pre-rebase commits

## Change

**`bin/fm-brief.sh`** - the local-only definition of done now generates a "Review marker contract" section. `direct-PR` and `no-mistakes` work is reviewed on its forge, so their output, plus scout and secondmate output, is byte-identical to before. The generated contract has the worker:

- pin the tag `<task-id>-reviewed` at `git merge-base <default-branch> HEAD` before first reporting ready, never at a moving ref
- end every ready report with exactly `cd <worktree> && hunk diff <task-id>-reviewed`, using the worktree's real absolute path from `pwd -P`
- record the reported tip, and advance the marker only on a review-pass **verdict** (an approval, or a batch of change requests handed back as that pass's outcome): advance to that pass's recorded sha *first*, then address the feedback, then re-report - never advancing to a tip the reviewer has not seen
- treat inline comments and questions arriving mid-pass as *not* a verdict: leave the marker untouched and answer each in place with `hunk session comment add`, since a mid-pass move drops the rest of that pass's unreviewed changes out of the next diff
- re-point the marker after any rebase or history rewrite, falling back to the current merge-base when the mapping is ambiguous, so the failure direction is always "show already-reviewed changes again", never "hide unreviewed ones"
- read live inline notes with `hunk session comment list --repo . --type user`, and never run the interactive `hunk diff` / `hunk show` themselves
- create or move only that one tag, never any other

**`bin/fm-teardown.sh`** - retires exactly the `<task-id>-reviewed` tag from the project clone for ship tasks, with the task's other durable records. An absent tag is a silent no-op, no other tag is touched, and it runs only after every landed-work refusal has passed, so a refused teardown leaves the marker intact for the rerun.

Both facts live in the scripts' own headers (and `--help`), matching this repo's one-owner rule for mechanics; `AGENTS.md` is unchanged.

## Verification

- `tests/fm-brief.test.sh` - new `test_local_only_review_marker_contract` asserts every clause of the generated contract with the task id substituted, and asserts the other four scaffolds carry none of it. 21/21 pass.
- `tests/fm-teardown.test.sh` - new cases (z1) marker removed while unrelated tags and another task's marker survive, (z2) absent marker is silent, (z3) a refused teardown keeps the marker (safety ordering). 61/61 pass.
- Byte-compared generated `no-mistakes`, `direct-PR`, scout and secondmate briefs against `main`: identical.
- `bin/fm-lint.sh` clean on ShellCheck 0.11.0 (the pin).
- `bin/fm-test-run.sh --changed`: 34 suites, 0 failures (2 harness-gated skips).


ecloin:
Opened on the correct repo instead: https://github.com/ecloin/firstmate/pull/1 - closing this one.

ecloin:
Reopening as an offered upstream contribution; this change also lives independently on the ecloin fork.

ecloin:
Withdrawn — keeping this as a fork-local extension. Apologies for the noise.

PR #2112, opened 2026-08-11
fix(bin): decide crew-state attribution from the branch's newest run row

cm-maple7:
## What this fixes

`nm_runs_status_for_branch` in `bin/fm-crew-state.sh` attributes a crew's current state from the coarse `no-mistakes runs` list when the precise run-step lookup is unavailable. Its selection had two mirror-image errors, both rooted in treating the row's short sha as code identity for a *live* run.

**1. A healthy live run misread as failed (and a false wedge-alert storm).**
The walk skipped any same-branch row whose short sha did not match the worktree HEAD. That test is wrong for an active run: the pipeline's rebase step and every fix-round commit advance the branch tip away from the worktree, so the live row stops matching exactly precisely while the run is healthiest. The walk then continued and matched an older, superseded row that still sat on the worktree sha, reporting a running pipeline as `failed`/`cancelled`. That verdict also yields `crew_absorb_class=none`, which routes the watcher to the immediate no-timer stale surface, so one defect produced both a false failure and a false wedge-alert storm.

**2. A dead run masking a genuinely stuck crew (a false green).**
Qualifying an active row on branch alone (the fix for #1) still let the walk continue past a same-branch row that failed the head test. A crew that rebases its own branch past its finished run leaves a *newest* terminal row that no longer binds; the walk could then reach an *older*, abandoned active row and take it, reporting a run dead for days as a healthy pipeline and annotating the crew's own `blocked:` line as superseded by it, so the watcher absorbed the wake instead of surfacing a stuck crew.

## The fix

The branch's **newest** row decides, always - the walk never continues past it into an older row of the same branch:

- An **active** run (`pending`/`running`, no-mistakes' own active-run vocabulary) qualifies on branch alone, whatever its tip has advanced to.
- The **sha test survives only where it is sound**: disambiguating a *finished* newest row, so a historical run on a reused branch is never attributed to rewritten code.
- A newest row that is finished and no longer binds yields **nothing** rather than deferring to an older row - the worktree sits on code a newer run has already moved past, so an older run's verdict is not current-state evidence.
- Cross-branch rows are unaffected; the branch filter still continues past other branches.

The row's single sha column is the run's *current* head, not the head the worktree submitted (verified against the installed no-mistakes CLI v1.41.2), so there is no submitted-head field to bind against on this surface.

## Tests

`tests/fm-crew-state.test.sh` gains coverage for the previously-uncovered skew direction, where the **worktree** moves past the run rather than the run moving past the worktree:

- The adversarial regression builds failure mode #2 and asserts the crew's `blocked:` line survives, both in the reported state and in the watcher's absorb class.
- A single genuine failed row on the submitted head still reports `failed`.
- A live row survives the crew rebasing past its head - the one direction every existing head-skew case missed.

A new `rebase_worktree_past_run` helper replays a real file commit through a real `git rebase` and asserts the mutual non-ancestry it claims to build, so these cases cannot go quietly vacuous.

The full `tests/fm-crew-state.test.sh` suite passes (every prior case plus the new ones), along with `bin/fm-lint.sh` and `shellcheck` on the changed script.


cm-maple7:
Superseded by #2183, which re-raises this same crew-state attribution fix through the no-mistakes pipeline so it carries the required signature. Closing this one to avoid a duplicate; please review #2183 instead. Thanks!

PR #2118, opened 2026-08-11
feat(brief): support custom ship branch names

### LINKED ISSUE #1887: fm-brief.sh hardcodes the ship branch as fm/<task-id>, so every brief needs hand-editing to follow a team's branch convention
## Problem

`bin/fm-brief.sh` hardcodes the ship branch as `fm/<task-id>` and offers no way to set it. The generated ship brief instructs the crewmate:

```
1. First action: create your branch: `git checkout -b fm/<task-id>`
```

Any team with its own branch-naming convention therefore has to **hand-edit every generated brief** before dispatch. Firstmate writes the brief, then the operator immediately corrects one line of it, every time.

Concrete cost: dispatching five ship tasks in one batch on 2026-08-06 meant five hand-corrections, each an opportunity to forget one. Our convention is conventional-commit style — `<type>/<github-issue>/<short-description>`, e.g. `feat/646/migration-schema` — which also makes branches legible to teammates who never see firstmate.

## The `fm/` prefix is load-bearing in more places than the scaffold

This is the part worth knowing before anyone implements it, because we hit it: overriding the branch name in the brief alone is **not** sufficient. At least three places assume the prefix.

- **`bin/fm-merge-local.sh`** hardcodes `BRANCH="fm/$ID"` and **hard-refuses** when that ref is absent. So a `local-only` task on a custom branch becomes un-mergeable by firstmate's own merge path — the flag would silently create a task that cannot be landed.
- **`bin/fm-bearings-snapshot.sh`** keys off the `fm/` prefix to map a PR back to its task. A custom branch maps to task `-`, silently, so the PR drops out of the bearings view.
- **`bin/fm-review-diff.sh`** survives, but only via its detached-HEAD fallback rather than by design.

Nothing regresses on the default path in any of those.

## Suggested shape

A `--branch <name>` flag on `fm-brief.sh` that substitutes through the generated brief, with the branch recorded in `state/<id>.meta` so the merge path and the snapshot can read the real name instead of reconstructing it from the task id. Defaulting to `fm/<task-id>` keeps every existing caller unchanged.

Two things worth validating in any implementation:

- **Branch-name validation.** Rejecting empty, leading-dash, whitespace and quote characters matters because the value is interpolated into printed shell commands elsewhere (`fm-promote.sh` interpolates a branch into a single-quoted `fm-send.sh` argument that the operator is told to run verbatim; an apostrophe there breaks the quoting).
- **The merge path**, per above — otherwise the flag creates the un-mergeable-task trap.

## Offer

We have a working implementation of this sitting on a local branch (flag, meta plumbing through to the merge gate, validation, and tests for both `fm-brief.sh` and `fm-promote.sh`). Filing the issue first rather than opening an unsolicited PR — happy to push it if there is appetite and the shape above is roughly right, or to drop it if you would rather solve it differently.

4mb1t10n:
## Summary

- add a ship-only `--branch <name>` option to `fm-brief.sh`, while preserving `fm/<task-id>` as the default
- validate branch values and refuse the option for scouts, secondmates, and `local-only` delivery where it cannot be honored safely
- add regression coverage, repair the empty-value validation row, and document custom ship branches in the architecture guide

Closes #1887

### DIFF
--- a/bin/fm-brief.sh
+++ b/bin/fm-brief.sh
@@ -6,7 +6,7 @@
 # description, acceptance criteria, and context, and may adjust other sections
 # when the task genuinely deviates (e.g. working an existing external PR instead
 # of shipping a new one).
-# Usage: fm-brief.sh <task-id> <repo-name> --mode <no-mistakes|direct-PR|local-only> [--herdr-lab]
+# Usage: fm-brief.sh <task-id> <repo-name> --mode <no-mistakes|direct-PR|local-only> [--branch <name>] [--herdr-lab]
 #        fm-brief.sh <task-id> <repo-name> --scout [--herdr-lab]
 #        fm-brief.sh <task-id> --secondmate {<project>...|--no-projects}
 #   --scout writes the scout contract instead: the deliverable is a report at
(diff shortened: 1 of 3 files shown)


4mb1t10n:

PR #2119, opened 2026-08-11
fix(spawn): remove orphaned tmux window on worktree refusal

### LINKED ISSUE #1913: fm-spawn leaves an orphan tmux window when a worktree validation refuses

4mb1t10n:
## Summary

- mark newly created tmux task windows for abort cleanup
- remove the tmux window when spawn worktree validation refuses the spawn
- clear the cleanup marker after a successful spawn
- add regression coverage proving a refused spawn leaves no orphan and a retry avoids duplicate-window failure

Closes #1913

### DIFF
--- a/bin/fm-spawn.sh
+++ b/bin/fm-spawn.sh
@@ -622,6 +622,7 @@ spawn_remote_secondmate() {
 }
 
 BACKEND=
+TMUX_ABORT_CLEANUP=0
 ORCA_ABORT_CLEANUP=0
 ORCA_WORKTREE_ID=
 ORCA_TERMINAL=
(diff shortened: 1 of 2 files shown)


4mb1t10n:

PR #2120, opened 2026-08-11
fix(bin): stop session owner from misclassifying agent-directory helpers

### LINKED ISSUE #2067: Session owner can misclassify exec_bridge under agent-named directories
## Summary

`bin/fm-session-lock-lib.sh` can classify an unrelated helper executable as the active harness when its installation path contains names associated with supported agents.

A reproduced process used this shape:

```text
.../.pi/agent/npm/node_modules/@howaboua/pi-codex-conversion/src/tools/exec/bin/darwin-arm64/exec_bridge
```

Session startup recorded `exec_bridge` as the owner instead of its parent Pi process, which then prevented Pi background supervision from accepting the live session.

## Cause

Two fallback signals combine into a false positive:

1. Non-Claude harness names are accepted from arbitrary executable-path components.
2. The bare-interpreter fallback matches `node_modules` in the full executable path and a harness substring such as `codex` elsewhere in the argument string.

Neither signal proves that `exec_bridge` is an agent process.

## Expected behavior

Session-owner classification should be deterministic and conservative:

- Keep exact executable-name detection for every supported primary harness.
- Keep version-named Claude Code detection through its verified install-path shape on macOS and Linux.
- Keep genuine bare Node or Python interpreters when the immediate script argument identifies a supported harness.
- Do not infer non-Claude identity from arbitrary directory components or unrelated argument substrings.

## Regression coverage

An executable-level regression should include the full `exec_bridge` shape above, including `.pi`, `node_modules`, and `pi-codex-conversion`, and assert that it is not a harness process.

Positive cases should retain version-named Claude detection, exact executable names for Claude, Codex, OpenCode, Pi, pi-signed, Grok, and Kimi, and direct bare-interpreter harness scripts.

The tmux caller should share the same Claude-only path fallback; Herdr, Zellij, Orca, and cmux do not use this path evidence in their backend-specific agent-state integrations.

4mb1t10n:
## Summary

- restrict path-based harness identity to Claude’s version-named install layout
- inspect only a bare interpreter’s immediate script argument using whole path components
- add regression coverage for exec_bridge false positives and all supported harness detection paths

## Decisions

- Kept this PR narrow to the reported path-only false positive. The tmux classifier retains its existing substring basename policy because this issue’s exec_bridge regression does not prove a broader exact-identity refactor is required, and observed vendor identities such as codex-aarch64-a and kimi-code still depend on that flexibility.
- This change does alter holder-liveness matching, disclosed here so it can be tracked separately. `fm_harness_pid_alive()` (`bin/fm-session-lock-lib.sh:159`) reaches the shared harness-identity rule by calling `fm_harness_process_matches()` (`:164`), and that rule now strips one leading `-` from the reported basename before the exact match. A login-shell-style harness invocation (`exec -a -codex ...`) previously failed to match and was therefore judged not alive; it now matches and is judged a live holder. This readmits genuine harness processes that the first version of this fix had wrongly excluded, and it does not restore substring path matching, so the reported exec_bridge false positive stays fixed.
- Withdrawing the earlier claim that this PR does not change holder-liveness semantics: it does, as described above. What still holds is the narrower claim that this PR does not change the lock write boundary. Composability with PR #2132’s recycled-PID identity work should therefore be re-checked against the dash-stripped matching rule rather than assumed.
- The separate Codex Desktop app-server lifetime problem remains unchanged and out of scope.

Closes #2067

### DIFF
--- a/bin/backends/tmux.sh
+++ b/bin/backends/tmux.sh
@@ -166,12 +166,12 @@ fm_backend_tmux_classify_process_name() {  # <path> [argv0] -> agent|shell|other
     # `codex` the substring `muse` is a common English fragment - a *muse* glob
     # would classify musescore or amuse as a live agent pane. The install path
     # cannot carry it either: ~/.local/bin/muse-bin-<version> has no `muse` path
-    # COMPONENT, so the fm_harness_path_name fallback below never fires for it.
+    # COMPONENT, so the Claude-only path fallback below never fires for it.
     muse|muse-bin-*) printf 'agent' ;;
     *claude*|*codex*|*opencode*|*grok*|*kimi*|pi|pi-signed|pi-launcher|Pi) printf 'agent' ;;
     zsh|bash|sh|dash|ash|ksh|mksh|tcsh|csh|fish) printf 'shell' ;;
     *)
-      if fm_harness_path_name "$path" >/dev/null || fm_harness_path_name "$argv0" >/dev/null; then
+      if fm_harness_claude_path_matches "$path" || fm_harness_claude_path_matches "$argv0"; then
         printf 'agent'
       else
         printf 'other'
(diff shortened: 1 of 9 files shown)


4mb1t10n:

PR #2121, opened 2026-08-11
feat(snapshot): expose registered projects

### LINKED ISSUE #2007: Expose registered projects in fm-fleet-snapshot.v1
Hi, I’m currently building an external desktop control plane around FirstMate.

I’m intentionally keeping the integration on public FirstMate surfaces only and currently use:

fm-fleet-snapshot.sh --json

with:

fm-fleet-snapshot.v1

During integration I ran into one read-side gap:

A project can be registered in FirstMate but have no backlog history and no current runtime tasks. In that case, the project is completely invisible through the current public snapshot.

This makes external project discovery and link verification incomplete.

I investigated the current FirstMate project registry architecture and wanted to ask whether you would consider exposing a minimal public registry projection through the existing fleet snapshot.

A possible shape could be:

```json
{
  "project_registry": {
    "present": true,
    "available": true,
    "complete": true,
    "reason": null,
    "total": 1,
    "shown": 1,
    "truncated": 0,
    "records": [
      {
        "name": "example-project"
      }
    ]
  }
}
```

The important part for my use case would only be the registered project names plus availability/completeness information.

I would explicitly not need:

project paths
Git remotes or Git identity
descriptions
registration dates
delivery mode/posture
tmux/worktree details
any write/control functionality

I also would prefer this to remain part of the existing public fleet snapshot rather than introducing a second command such as fm-project-list, if that fits FirstMate’s architecture.

Two things I’m unsure about and would prefer to leave to FirstMate’s maintainers:

whether registry enumeration should live in a new shared registry reader/library or extend an existing project-registry reader;
whether additive top-level fields are formally intended to remain compatible within fm-fleet-snapshot.v1.

Would a small public project-registry projection like this fit FirstMate’s intended architecture?

If yes, I’d also be happy to prepare a PR once the desired contract is agreed on.

4mb1t10n:
## Summary

- add a bounded `project_registry` field to `fm-fleet-snapshot.v1` so registered projects remain discoverable without backlog history or live tasks
- expose project names only, preserving private registry metadata and reporting absence, truncation, timeout, or unreadability explicitly
- document the snapshot field and cover absent, backlogless, privacy, and unreadable-registry behavior

Closes #2007

### DIFF
--- a/bin/fm-fleet-snapshot.sh
+++ b/bin/fm-fleet-snapshot.sh
@@ -52,6 +52,15 @@
 #     structured homes with an unknown current classification are partial, not
 #     unreadable, and retain independently trustworthy structured surfaces.
 #   secondmate_guidance: return-channel action note for renderers and bearings.
+#   project_registry: {present,available,complete,reason,records[],total,shown,
+#     truncated} - minimal public projection of data/projects.md: registered
+#     project NAMES only (records[].name), so a project with no backlog history
+#     and no live task can still be discovered. Never includes project paths,
+#     git remotes, descriptions, registration dates, or delivery mode/posture/
+#     yolo - those stay captain-private. present=false means no registry file
+#     (not an error); available=false means it exists but could not be read
+#     (permissions or timeout), with reason set. complete=false means the
+#     bounded read truncated before covering every registered project.
 #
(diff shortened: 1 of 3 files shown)


4mb1t10n:

PR #2123, opened 2026-08-11
fix: preserve pending replies for unproven sends

### LINKED ISSUE #2003: fm-send verified submit always reports verdict=pending on Claude NBSP composers - discards pending-reply records, loops config re-nudges, skips --resolve-key closes
## Summary

On Claude panes whose composer renders as `❯` + U+00A0 (see #1988, fix in PR #1995), `fm-send`'s verified submit **always exits non-zero with `verdict=pending` even when delivery succeeded**. The header's contract ("exits NON-ZERO so the caller knows the steer did not land") inverts in practice: every successful steer reports failure, so the exit code carries no signal and operators must peek the pane to learn the truth.

This is the fm-send-side fallout of the same root as #1988 — filing separately because three downstream consumers of fm-send's exit code break in ways that persist after the send:

1. **Pending-reply expectation discarded**: a marked secondmate request's `state/pending-replies/` record is dropped on the false failure, so the missed-report guard never arms for that request (replies still arrive as status wakes, but the recovery/escalation contract is silently lost).
2. **Config-push re-nudge loop**: config inheritance clears its `.pending` marker only on rc=0, so every bootstrap re-delivers the identical config to an already-ACKed secondmate, forever (observed: 4 identical redeliveries acknowledged by the same home).
3. **`--resolve-key` close skipped**: the closing `resolved [key=…]` append only happens after a *confirmed* submit, so an answered decision stays OPEN in every wake drain even though the answer was delivered. (Safe direction by design, but with a 100% false-negative confirm rate on these panes it means keyed decisions can never close through fm-send.)

## Reproduction

1. Claude harness target whose idle composer is `❯` + U+00A0 (stock rendering here).
2. `bin/fm-send.sh <task-id> "any text"` to the idle pane.
3. Text visibly delivers and the turn starts; fm-send exits 1 with `verdict=pending`.

Also note the busy-pane trap that makes naive re-send verification dangerous: a send to a busy pane queues invisibly (not in scrollback, not in the composer), so a scrollback grep false-negatives and a retry double-delivers.

## Expected

With PR #1995's composer classification fix, the submit confirm should stop false-negating; worth verifying afterwards that (1)–(3) recover, and considering making the pending-reply record survive an unconfirmed-but-possibly-delivered send rather than being discarded.

4mb1t10n:
## Summary

- preserve newly created pending-reply expectations as `delivery_unknown` when a backend returns an unproven submit verdict such as `unknown` or `pending-unproven`
- continue discarding expectations only when non-delivery is proven by `send-failed` or a confirmed `pending` composer, without changing the strict nonzero exit contract
- add regression coverage for proven pending, send failure, and ambiguous unknown verdict handling

Closes #2003

### DIFF
--- a/bin/fm-send.sh
+++ b/bin/fm-send.sh
@@ -36,6 +36,12 @@
 # resolves the expectation. Set FM_PENDING_REPLY_EXISTING_CORR=<id> when
 # re-sending a recovery request for an already-open expectation so a second
 # record is not created. Direct unmarked captain input never creates one.
+# A submit verdict the backend cannot prove one way or the other (unknown,
+# pending-unproven, or any future non-proof verdict) preserves the record as
+# delivery-unknown rather than discarding it: the harness could not disprove
+# delivery, so the record must survive for reconciliation (see #2003). Only a
+# verdict that PROVES non-delivery - send-failed, or a proven "pending"
+# composer still holding the text after the full retry budget - discards it.
 #
 # Decision closure (answerer-closes): pass --resolve-key <key> (repeatable,
 # before the message) when this send answers an open keyed needs-decision: or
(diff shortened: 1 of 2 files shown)


4mb1t10n:

PR #2124, opened 2026-08-11
fix(project-mode): match multi-word registry names

### LINKED ISSUE #1977: fm-project-mode.sh cannot match any project name containing a space (awk keys on $2, the first token only)
## Summary

`fm-project-mode.sh` cannot resolve the posture for any project whose registry name contains a space. The awk matcher keys on `$2==n`, but `$2` is only the **first whitespace-delimited token** of the row, not the whole project name. So a multi-word name never matches, the lookup falls through to the `warn: project "<name>" not in registry` branch, and the project silently gets `no-mistakes off` instead of its declared mode.

## Where

`bin/fm-project-mode.sh`, the awk block (around line 60):

```awk
parsed=$(awk -v n="$NAME" '
  $1=="-" && $2==n {
    ...
```

`$2==n` compares the caller's full `NAME` against a single field. For a registry row like:

```
- 048. Blast- Lease summary drafter [local-only] - BLAST lease summary drafter ... (added 2026-08-08)
```

`$2` is `048.` — so `NAME="048. Blast- Lease summary drafter"` can never equal it.

## Impact

- The `[local-only]` (or any) mode annotation is **non-functional for every space-containing project name**. The project is treated as unregistered.
- The `warn: project "<name>" not in registry; defaulting to no-mistakes off` line fires on every dispatch for such a project, even though it *is* in the registry.
- It fails toward `no-mistakes`, which is more pipeline machinery, not less — wrong for a deliberately `local-only`, no-remote project (there is nothing for the no-mistakes pipeline to push).
- Affects real registry entries today: `048. Blast- Lease summary drafter`, `ARI eDiscovery Development`. Single-token names (`ai-vuln-watchlist`, `100285-000001`, `A5486-000005-Rooter`) are unaffected, which is why it hides.

## Root cause

The registry row format is `- <name> [<mode>] - <desc> (added <date>)`. The name is a free-text span that ends at ` [` (mode annotation) or ` - ` (description), not at the first space. Matching on `$2` assumes the name is a single token.

## Suggested fix

Parse the name as the span between the leading `- ` and the first ` [` / ` - ` delimiter, then compare that to `n`, instead of `$2==n`. Sketch:

```awk
awk -v n="$NAME" '
  /^- / {
    line = substr($0, 3)                 # drop leading "- "
    name = line
    if (match(name, / \[/))      name = substr(name, 1, RSTART-1)
    else if (match(name, / - /)) name = substr(name, 1, RSTART-1)
    if (name != n) next
    # ... existing mode/yolo parsing, but read the "[...]" out of `line` ...
  }
'
```

A test with a spaced name (e.g. `Foo Bar Baz [local-only]`) resolving to `local-only` would lock this in; the current suite presumably only exercises single-token names.

## Discovered

While operating `048. Blast- Lease summary drafter` (a local-only, no-remote project) as firstmate: every crewmate dispatch printed the not-in-registry warning despite a correct registry entry, and the `[local-only]` annotation never took effect.

4mb1t10n:
## Summary

- parse project registry names as full free-text spans instead of matching only the first whitespace-delimited token
- choose the earliest mode-annotation or description delimiter so brackets in descriptions do not truncate project names incorrectly
- add regression coverage for multi-word names, delivery modes, `+yolo`, bracketed descriptions, and genuine registry misses

Closes #1977

### DIFF
--- a/bin/fm-project-mode.sh
+++ b/bin/fm-project-mode.sh
@@ -66,8 +66,11 @@ parsed=$(awk -v n="$NAME" '
     match($0, /^[ \t]*-[ \t]+/);
     line = substr($0, RSTART + RLENGTH);
     name = line;
-    if (match(name, / \[/)) name = substr(name, 1, RSTART - 1);
-    else if (match(name, / - /)) name = substr(name, 1, RSTART - 1);
+    bpos = 0; dpos = 0;
+    if (match(name, / \[/)) bpos = RSTART;
+    if (match(name, / - /)) dpos = RSTART;
+    if (bpos > 0 && (dpos == 0 || bpos < dpos)) name = substr(name, 1, bpos - 1);
+    else if (dpos > 0) name = substr(name, 1, dpos - 1);
     if (name != n) next;
     mode="no-mistakes"; yolo="off";
     rest = substr(line, length(name) + 1);
--- a/bin/fm-project-mode.sh
+++ b/bin/fm-project-mode.sh
@@ -56,14 +56,25 @@ if [ ! -f "$REG" ]; then
   exit 0
 fi
 
+# The name is a free-text span, not a single token, so it is recovered from
+# the leading "- " up to the first " [" (mode annotation) or " - "
+# (description) delimiter rather than compared field-by-field; matching on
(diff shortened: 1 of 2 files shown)


4mb1t10n:

PR #2125, opened 2026-08-11
fix: wake on closed-unmerged pull requests

### LINKED ISSUE #1897: PR merge poll never wakes firstmate when a PR is closed without merging
## What happens

`fm-pr-check.sh` arms a merge poll for a ship task's PR, but the poll only ever reports back on `merged`.
A PR that is **closed without being merged** produces no output, ever.
The poll stays armed, the watcher never queues a wake, and the task sits in the backlog as in-flight indefinitely with no notification to the captain.

## Where

`bin/fm-pr-poll.sh` — GitHub path:

```sh
65:    state=$(gh pr view "$url" --json state -q .state 2>/dev/null) || exit 0
66:    [ "$state" = MERGED ] && printf '%s\n' merged
```

Same shape on the GitLab path:

```sh
106:    [ "$state" = merged ] && printf '%s\n' merged
```

`CLOSED` is not handled anywhere in `bin/` — `grep -rn CLOSED bin/*.sh` returns nothing.
The consumer at `bin/fm-watch.sh:844` only branches on `[ "$out" = merged ]`, so even if the poll did emit a closed result today, nothing would act on it.

## Why it matters

Closing a PR unmerged is a normal outcome, not an edge case.
Two routine flows hit it:

- A PR is superseded and its work is folded into another PR.
- A PR is abandoned after review and reopened from a different base or branch.

In both cases the work may well have landed, but firstmate is watching the wrong PR and has no way to find out.
The task never reaches teardown, the backlog silently drifts out of sync with reality, and the captain only discovers it by manually auditing in-flight work.

I hit this in normal use: a ship task's recorded PR was closed unmerged and its change re-landed in a successor PR.
The task remained in-flight for three days with a dead worker endpoint and produced no wake at all.
It surfaced only through a hand audit of `state/*.meta` against the forge.

## Two related symptoms from the same root cause

1. **Teardown recommends a misleading artifact.** After manual close-out, teardown's backlog reminder proposes recording the stale `pr=` (the closed-unmerged PR) as the completion artifact, rather than the PR that actually landed the work.
2. **`pr=` is never revalidated.** Once recorded by `fm-pr-check.sh`, nothing re-checks that the PR is still the live one for that branch.

## Suggested direction

Emit a distinct terminal result for a closed-unmerged PR — something like `closed` alongside the existing `merged` — and give `fm-watch.sh` a branch for it that queues an actionable wake rather than a silent retirement.

The wake should be escalated, not auto-resolved: a closed PR does not tell you whether the work landed elsewhere, so it needs the captain or firstmate to reconcile. Retiring the poll without a wake would just swap a silent hang for a silent drop.

Teardown's landed-work check is a useful precedent here — it already handles "the branch's own commits live nowhere on a remote yet the change is fully in the default branch" for squash-merge flows, and the same content check could distinguish "closed because superseded and already landed" from "closed and genuinely abandoned".

Happy to send a PR if the direction looks right.

## Environment

- firstmate at `70aeba8` (current `main` at time of filing)
- backend: tmux
- primary harness: claude
- forge: GitHub

4mb1t10n:
## Summary

- emit a distinct `closed` result when GitHub PRs or GitLab merge requests close without merging
- keep closed-unmerged polls armed while waking firstmate for reconciliation
- document the terminal-state behavior and add regression coverage for both forge paths and watcher retirement

Closes #1897


4mb1t10n:

PR #2126, opened 2026-08-11
fix: prune resolved pending replies after retention

### LINKED ISSUE #2053: Resolved pending-reply records are never pruned, so answered requests count as open forever
Resolved pending-reply records are never removed, so answered requests keep counting as open and the watch cycle pays for them forever.

## Evidence

On 2026-08-10 the captain read the fleet as "eight awaiting reply". Zero were
open. All eight records in `state/pending-replies/` carried `phase=resolved`:

| record | task | resolved |
|---|---|---|
| `016936f8d6087003` | secondmate | 08-10 00:39 |
| `3979b14f5810b96d` | secondmate | 08-10 00:02 |
| `471d48297d49f1be` | secondmate | 08-10 00:03 |
| `64f574e50ab59552` | secondmate | 08-10 00:17 |
| `6cc70de9e093a601` | secondmate | 08-10 00:39 |
| `7f8f91734da7f28f` | secondmate | 08-10 00:39 |
| `a225e35ac9be87e6` | secondmate | 08-09 20:15 |
| `c4371d60c88f4e33` | mumbl | 08-09 11:58 |

The oldest had been closed for more than 23 hours and was still on disk. None
had `escalated_epoch` set, so no close-out work depended on any of them.

## Root cause

`bin/fm-pending-reply-lib.sh` has exactly one unlink path,
`fm_pending_reply_discard_undelivered` at line 414. It refuses to run once
`delivered_epoch` is set:

    delivered=$(fm_pending_reply_get "$rec" delivered_epoch)
    [ -z "$delivered" ] || return 1

A record that was delivered and then answered can therefore never be removed.
The sweep at line 1085 reaches `resolved) return 0` and leaves the file in
place.

## Two consequences

1. Any surface that counts records in `state/pending-replies/` reports answered
   requests as open. The count only ever grows.

2. `fm_pending_reply_tick` (line 1129) iterates every record each watch cycle.
   For a resolved record it calls `fm_pending_reply_close_escalation`, which
   acquires the lock through `fm_lock_acquire_wait` BEFORE
   `_fm_pending_reply_close_escalation_locked` makes the cheap
   `[ -n "$escalated" ] || return 0` check. So each cycle takes one lock per
   historical record, forever.

## Proposed change

1. Prune resolved records in `fm_pending_reply_tick` once they are older than a
   retention window and their escalation is closed or was never opened. Do not
   prune a resolved record whose `escalated_epoch` is set and whose
   `escalation_closed_epoch` is empty — that close must still converge.

2. Make the retention window a named constant with a default, overridable the
   same way `fm_pending_reply_grace_secs` already is. Document it in
   `docs/configuration.md` next to the existing grace setting.

3. Move the escalation check ahead of the lock in
   `fm_pending_reply_close_escalation` so an unescalated resolved record costs
   no lock. Read `escalated_epoch` first, return early, and take the lock only
   when there is an escalation to close.

Keep the durable-retention rule for UNRESOLVED records exactly as it is. Line
1087 states the contract: an unresolved record is never auto-deleted. Nothing
here changes that.

## Acceptance

1. A resolved record older than the retention window is gone after a tick.
2. A resolved record younger than the window survives a tick.
3. A resolved record with `escalated_epoch` set and `escalation_closed_epoch`
   empty survives a tick, however old it is, and its close still converges on a
   later tick.
4. An unresolved record of any age and any phase survives a tick.
5. A tick over N resolved unescalated records acquires no pending-reply lock.
6. The retention window is documented in `docs/configuration.md`.
7. `bin/fm-lint.sh` passes.

Assert through the executable interface only. Never assert against script
source bytes.

4mb1t10n:
## Summary

- prune resolved pending-reply records after a configurable retention window
- retain unresolved records and resolved records whose escalation closure has not converged
- avoid taking per-correlation locks for resolved records that were never escalated
- document the retention setting and cover retention, escalation, and open-count behavior with regression tests

Closes #2053

### DIFF
--- a/bin/fm-pending-reply-lib.sh
+++ b/bin/fm-pending-reply-lib.sh
@@ -66,11 +66,22 @@
 # or a remote mate's mirrored line - can take the key over or clear it; see the
 # reserved-key rule in bin/fm-classify-lib.sh.
 #
+# Retention: a resolved record is never open, so fm_pending_reply_task_has_open
+# already excludes it. fm_pending_reply_tick additionally prunes it from disk
+# once it has aged past the retention window AND its escalation (if any) is
+# closed. A resolved record whose escalation is still open is never pruned,
+# however old, because that close must still converge. An unresolved record of
+# any age or phase is never auto-deleted; that durable-retention contract is
+# unchanged.
+#
 # Sourced by bin/fm-send.sh, bin/fm-watch.sh, bin/fm-secondmate-report.sh, and
 # tests. No side effects on source. set -u / set -e safe.
 #
 # Tunables (env):
-#   FM_PENDING_REPLY_GRACE_SECS   default 120
+#   FM_PENDING_REPLY_GRACE_SECS     default 120
+#   FM_PENDING_REPLY_RETENTION_SECS default 3600; how long a resolved record
+#                                   with no open escalation stays on disk
+#                                   before fm_pending_reply_tick prunes it
 #   FM_PENDING_REPLY_DIR_OVERRIDE override the pending-replies directory (tests)
 #   FM_PENDING_REPLY_SEND_HOOK    optional command template for recovery delivery
 #                                 (tests); receives task_id and full message as args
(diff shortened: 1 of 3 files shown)


4mb1t10n:

PR #2127, opened 2026-08-11
fix(watch): throttle unread signal backlog

### LINKED ISSUE #2028: Supervision looks stalled when it is actually thrashing on unread worker reports

4mb1t10n:
## Summary

- distinguish repeated watcher signals that are already queued and undrained from genuinely fresh signals
- throttle repeated unread-backlog reports while periodically escalating them with an explicit supervision alarm and affected task IDs
- document the new watcher state files and add coverage for classification, repeat reporting, alarm escalation, and post-drain behavior

Closes #2028

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -117,7 +117,7 @@ state/               volatile runtime signals; gitignored
   .afk               durable away-mode flag; present = sub-supervisor may inject escalations (set by /afk, cleared on user return)
   .watch.lock .wake-queue.lock watcher singleton and queue serialization locks
   .claude-autoarm.lock .claude-autoarm-epoch .claude-autoarm-failure-notified .claude-autoarm-failure-alarmed .turnend-claude-blocks .turnend-claude-blocks.lock   Claude Stop auto-arm single-flight, epoch, failure-episode, attended-alarm, guard-budget, and budget-lock records; never touch
-  .hash-* .count-* .stale-* .stale-since-* .paused-* .wedge-escalations-* .seen-* .hb-surfaced-* .last-* .heartbeat-streak   watcher internals; never touch
+  .hash-* .count-* .stale-* .stale-since-* .paused-* .wedge-escalations-* .seen-* .hb-surfaced-* .last-* .heartbeat-streak .signal-unacked-since-* .signal-unacked-count-*   watcher internals; never touch
   .watch-triage.log  watcher's absorbed-wake debug log (size-capped); never relied on, safe to delete
   .last-watcher-beat watcher liveness beacon, touched every poll (including while absorbing benign wakes); guard scripts read it
   .subsuper-* .supervise-daemon.*   sub-supervisor internals; never touch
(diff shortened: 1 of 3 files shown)


4mb1t10n:

PR #2128, opened 2026-08-11
fix(herdr): refuse panes whose shell was replaced

### LINKED ISSUE #1912: herdr backend: worker pane is a bare shell that an rc `exec` can replace, so the launch command is typed into the operator's own session
## Summary

The herdr backend creates a worker pane as a bare interactive shell and then *types* the launch command into it. If the operator's shell rc replaces that shell (`exec`) before the typed command arrives, the worker never starts, and the launch command is typed into whatever the replacement process is showing.

On a machine whose rc auto-attaches to a shared tmux session, this is deterministic: every herdr spawn silently fails, and the agent launch command is injected into the operator's own live tmux session.

## Environment

- firstmate at `70aeba8` (origin/main, up to date)
- herdr 0.8.0, protocol 19, server healthy
- Remote second mate on Ubuntu 26.04 under WSL2, launched from macOS over SSH
- Backend selected by runtime auto-detection (`HERDR_ENV=1`), which printed its usual experimental notice

## Reproduction

Add a standard auto-attach to the operator's shell rc on the host (this is a common remote-work convenience):

```sh
# ~/.zshrc
if [[ -o interactive ]] && [ -z "$TMUX" ] && command -v tmux >/dev/null; then
  exec tmux new-session -A -s main
fi
```

Then spawn any crewmate or scout on the herdr backend from that host.

Observed:

1. The pane is created and labeled correctly, and `state/<id>.meta` records a normal endpoint.
2. The pane's shell is interactive with `$TMUX` unset, so the rc guard is satisfied and `exec` replaces it with a client attached to session `main`.
3. The launch command, sent afterwards as typed text, lands in whichever pane `main` has focused.
4. The worker never starts. No agent registers in that workspace, and `herdr workspace list` reports `agent_status: unknown` for it.
5. The task leaves no status events, no report, and a clean worktree at the upstream commit with no branch and no commits.

In our case the injected text started a real turn in an unrelated Claude session that happened to be focused in `main`.

## Mechanism

- `bin/backends/herdr.sh:2016` creates the pane with `herdr tab create --workspace ... --cwd ... --label ... --no-focus` and passes no command.
- The launch command is delivered afterwards as typed text via `fm_backend_herdr_send_text_line` / `fm_backend_herdr_send_literal`, dispatched from `bin/fm-spawn.sh:1737` and `:1754`.

So the ordering is: create bare shell, rc `exec`s it away, then type a command into the replacement.

The tmux backend is structurally immune, because panes it creates inherit `$TMUX` and the guard is false. Redispatching the identical task with `--backend tmux` succeeded on the first attempt.

## Why this is worth fixing upstream

The trigger is operator configuration that firstmate cannot see, but the failure mode is severe and silent:

- The spawn reports success and records a normal endpoint, so supervision believes a worker exists.
- Nothing distinguishes it from a healthy worker without reading the pane.
- The failure escalates outside firstmate's blast radius by typing an agent launch command into an unrelated interactive session.

Any rc that `exec`s (tmux/zellij auto-attach, shell replacement, `exec fish`) produces the same result.

## Suggested fixes

1. **Pass the command at pane creation** rather than typing it afterwards, so no window exists in which an rc can replace the shell. This removes the entire class.
2. If typed delivery must stay, **verify the pane still hosts the shell that was created** before sending, and fail the spawn loudly when it does not.
3. Cheap hardening in the meantime: launch the pane's shell non-interactively, or set a marker variable the shell rc can be documented to check.

Item 2 alone would have converted a silent success into an actionable spawn failure.

## Workaround

Pin `config/backend` to `tmux` so the second mate's workers are unaffected. The remote second-mate agent itself still launches on herdr as required, since that path deliberately ignores `config/backend`.

4mb1t10n:
## Summary

- verify that a newly created Herdr task pane still hosts its recognized shell before any launch text is typed into it
- fail immediately when the shell identity was replaced via `exec`, while retrying inconclusive samples during shell startup
- add regression coverage for replaced, transiently busy, and never-resolving shell states

Closes #1912

### DIFF
--- a/bin/backends/herdr.sh
+++ b/bin/backends/herdr.sh
@@ -1189,7 +1189,8 @@ fm_backend_herdr_pid_is_bare_shell() {  # <ps-bin> <pid>
 # window and succeeds on the first fully clean one; a genuinely busy pane
 # fails every sample and still refuses.
 # This is the single owner of the idle-shell proof; the session-start
-# projection cleanup and every pane-death close path both rely on it.
+# projection cleanup, every pane-death close path, and the spawn-time
+# send-target proof below all rely on it.
 fm_backend_herdr_pane_idle_shell_pid() {  # <session> <pane-id>
   local attempt=0 max_attempts=${FM_BACKEND_HERDR_IDLE_SHELL_PROOF_POLLS:-10}
   while :; do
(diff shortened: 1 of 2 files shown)


4mb1t10n:

PR #2129, opened 2026-08-11
fix: accept status keys after verb colon

### LINKED ISSUE #2081: Decision-key grammar mismatch: needs-decision keys silently fold to "default"
## Summary

A crewmate's `needs-decision`/`resolved` status lines carry their decision key **after** the verb's colon (`needs-decision: [key=<slug>] <note>`), but `bin/fm-classify-lib.sh`'s `_fm_decision_key` only recognizes a key token that appears **before** the first colon (`needs-decision [key=<slug>]: <note>`), per the grammar documented in that file's own header comment. The result: every such line silently folds under the fallback key `"default"` instead of its intended named slug.

## Root cause

`_fm_decision_key` computes:

```sh
prefix=${1%%:*}   # everything before the FIRST colon
```

then looks for a `[key=...]` token inside `prefix`. But the actual format every crewmate writes (via the brief scaffold) puts the colon immediately after the verb, with the key token in the *note* that follows:

```
needs-decision: [key=my-decision] something needs a human call
```

Here `prefix` is just `needs-decision` — the `[key=...]` token never reaches the parser, so `_fm_decision_key` falls through to its `*) printf 'default'` branch.

## Reproduction

```sh
source bin/fm-classify-lib.sh
line='needs-decision: [key=my-decision] something needs a human call'
echo "verb: $(status_line_verb "$line")"   # -> needs-decision
echo "key:  $(_fm_decision_key "$line")"   # -> default   (expected: my-decision)
```

## Impact

`bin/fm-send.sh <task> --resolve-key <the-exact-slug-a-worker-used>` always refuses with "no open decision or blocker with that key", because the actual open-decision set computed by `status_open_decisions` only ever contains `"default"` for every task, regardless of how many distinct keys its status log actually names.

This has stayed invisible because the observed usage pattern is one open decision per task at a time — `default` correctly opens and closes in sequence, so a bare `--resolve-key default` (or no key at all, wherever that's accepted) still works. It would silently misattribute or simply refuse to resolve the moment a single task legitimately has two differently-keyed decisions open at once.

## Candidate fixes

1. **Parser-side (recommended):** relax `_fm_decision_key` / `_fm_decision_fold_line` to also recognize a `[key=...]` token appearing in the *note* (after the colon) when none is found in the prefix — since that's the format every brief scaffold actually emits today. Backward-compatible with every already-written status line.
2. **Emitter-side:** fix `bin/fm-brief.sh`'s scaffold to emit the documented pre-colon form (`needs-decision [key=<slug>]: <note>`) instead, and treat this as a one-time status-format migration across in-flight crewmates.

Option 1 is lower-risk since it doesn't require every in-flight crewmate's habit (or already-written status history) to change.

Either way, please add a regression test pinning both line shapes so this can't silently regress again.

### LINKED ISSUE #2109: status keys silently collapse to 'default' when the key follows the verb colon
## Symptom

A crewmate that appends `needs-decision: [key=seam-max-bound] ...` instead of the documented `needs-decision [key=seam-max-bound]: ...` has its key silently parsed as `default`.

`bin/fm-send.sh --resolve-key seam-max-bound` then refuses with "no open decision or blocker with that key", and the only way to close the decision is `--resolve-key default`.

## Reproduction

```
printf 'needs-decision [key=alpha] no colon form\n' > /tmp/t1.status
printf 'needs-decision: [key=beta] colon form\n' > /tmp/t2.status
. bin/fm-classify-lib.sh
status_open_decisions /tmp/t1.status   # -> alpha<TAB>needs-decision<TAB>...
status_open_decisions /tmp/t2.status   # -> default<TAB>needs-decision<TAB>...
```

## Why it matters

The collapse is silent, and `default` is a shared bucket. Two decisions open at once on the same task both fold to `default`, so an answer can close the wrong record while the real one stays open or is lost. Observed live on two independent tasks (claude and codex harnesses), so the misplaced-colon shape is a common worker output, not a one-off.

## Suggested fix

Either accept `<verb>: [key=<slug>]` as an equivalent form, or refuse the line loudly instead of folding an explicit but misplaced key into `default`. Silently rewriting a stated key to a shared bucket is the unsafe option.

4mb1t10n:
## Summary

- Accept `[key=<slug>]` when it is the first token after a status verb's colon, while preserving the existing pre-colon grammar.
- Keep key-like text later in ordinary note prose non-structural.
- Fix the shared parser cause behind both #2109 and #2081, including keyed `needs-decision`, `resolved`, and `fm-send --resolve-key` behavior.

## Reproduction

Before the fix, the issue's scratch-status reproduction returned the named key for the pre-colon form and silently returned `default` for the post-colon form:

```text
pre-colon: alpha    needs-decision    needs-decision [key=alpha] no colon form
post-colon: default needs-decision    [key=beta] colon form
```

The added classifier regression failed before the implementation change with:

```text
not ok - a leading post-colon key token did not open its named decision
FM_TEST_END ... exit=1
```

After the fix, that regression passed as part of `classifier primitives`, two post-colon keys remained distinct, and a post-colon `resolved` event closed only its named key.
The end-to-end `fm-send --resolve-key` regression also passed with a post-colon `needs-decision` event.

## Validation

- `git diff --check` - passed.
- `bin/fm-lint.sh` - passed with pinned ShellCheck 0.11.0.
- `bin/fm-doc-audience-check.sh` - passed (`surfaces=67 local_links=233`).
- `bin/fm-test-run.sh tests/fm-send-resolve-key.test.sh` - passed, 10 tests, exit 0.
- The new classifier assertion failed before the fix and passed after it.

`tests/fm-watch-triage.test.sh` has a pre-existing unrelated timing problem after the classifier tests complete.
On the changed branch, the first run passed `classifier primitives`, then failed later with `not ok - dead-agent declared pause flooded stale wakes across six unchanged polls` and exited 1.
The changed-branch rerun again passed `classifier primitives`, then stopped producing output and was interrupted after 250094 ms, exiting 130.

Firstmate authorized one bounded comparison against exact base commit `7f828ea2f28c83aa9e3afd44a2a5ee64795005f8` with both changed files restored to that commit.
That run printed `BASE_DIFF_BEGIN` followed immediately by `BASE_DIFF_END`, passed the same earlier classifier cases, then hung in the unrelated watcher timing cases and reached the 240-second bound:

```text
BASE_COMMIT=7f828ea2f28c83aa9e3afd44a2a5ee64795005f8
BASE_DIFF_BEGIN
BASE_DIFF_END
...
BASE_TEST_EXIT=124
```

This proves the `fm-watch-triage` hang reproduces with the key-parser change absent.

## Decisions

- Firstmate authorized the base-commit comparison under delegated authority; the captain was not consulted and approved nothing for this lane.
- Firstmate kept the unrelated `fm-watch-triage` timing defect outside this issue because it reproduces on the base commit.
- Firstmate switched this lane to direct delivery after three guarded custody-recovery refusals reported `state=blocked_recover_gate_diverged` and `changed=false`; no files or refs changed, so no work was lost.
- Pipeline head `705c78ef` and local head `4405008` had identical stable patch ID `76bbd814ca0802bb40644310b0949bebe9426491`, proving both carry the same intended fix.
- `705c78ef` was preferred because it rebased that patch onto newer base `7f05100`, but it was not reachable from this worktree, so Firstmate authorized direct delivery from the content-equivalent `4405008` fallback.
- The pre-existing watcher-suite hang is documented here and is not fixed or otherwise widened into this PR.

## Not applicable

- Database: this repository has none; no stack was started.
- Browser validation: this repository has no browser-facing surface.
- Cypress: this repository has no Cypress suite, and Cypress was not run locally.

Closes #2109

Closes #2081

### DIFF
--- a/bin/fm-classify-lib.sh
+++ b/bin/fm-classify-lib.sh
@@ -160,8 +160,10 @@ status_is_paused_or_captain_held() {  # <status-line>
 # rule 6), so closure never depends on a busy worker's discipline.
 #
 # Decision key grammar (backward-compatible with the existing "<verb>: <note>"
-# format): an OPTIONAL "[key=<slug>]" token sits between the verb and the colon,
+# format): an OPTIONAL "[key=<slug>]" token sits between the verb and the colon
+# or immediately after the colon,
 #   needs-decision [key=api-shape]: <summary>
+#   needs-decision: [key=api-shape] <summary>
 #   resolved       [key=api-shape]: <how it was decided>
 # A line with no token uses the key "default", preserving the historical
 # one-open-decision-per-task behavior (a bare "resolved:" closes "default").
@@ -181,17 +183,28 @@ status_line_note() {  # <status-line> -> text after the first colon, trimmed
   esac
 }
 _fm_decision_key() {  # <status-line> -> key slug, or "default" when no token
-  local prefix=${1%%:*} k
-  case "$prefix" in
-    *\[key=*\]*)
-      k=${prefix#*\[key=}
-      k=${k%%\]*}
-      case "$k" in
-        ''|*[!A-Za-z0-9._-]*) return 1 ;;
-        *) printf '%s' "$k" ;;
+  local candidate=${1%%:*} k
+  case "$candidate" in
+    *\[key=*\]*) ;;
+    *)
+      case "$1" in
+        *:*)
+          candidate=${1#*:}
+          candidate=${candidate#"${candidate%%[![:space:]]*}"}
+          case "$candidate" in
+            \[key=*\]*) ;;
+            *) printf 'default'; return ;;
+          esac
+          ;;
+        *) printf 'default'; return ;;
       esac
       ;;
-    *) printf 'default' ;;
+  esac
+  k=${candidate#*\[key=}
+  k=${k%%\]*}
+  case "$k" in
+    ''|*[!A-Za-z0-9._-]*) return 1 ;;
+    *) printf '%s' "$k" ;;
   esac
 }
 # Drop the record for <key> from a newline-terminated "<key>\t<verb>\t<note>" set.
--- a/tests/fm-send-resolve-key.test.sh
+++ b/tests/fm-send-resolve-key.test.sh
@@ -112,7 +112,7 @@ test_answer_send_closes_open_decision() {
   fb=$(make_stubs "$dir"); log="$dir/send.log"
   home=$(setup_home closes)
   fm_write_meta "$home/state/t1.meta" "window=sess:fm-t1" "kind=ship"
-  printf 'needs-decision [key=api-shape]: pick REST or RPC\n' > "$home/state/t1.status"
+  printf 'needs-decision: [key=api-shape] pick REST or RPC\n' > "$home/state/t1.status"
   printf 'working: kept busy on an unrelated stream\n' >> "$home/state/t1.status"
 
   out=$(drain_out "$home")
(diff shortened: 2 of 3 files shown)


4mb1t10n:

PR #2136, opened 2026-08-11
fix(stow): bound secondmate receipt waits

### LINKED ISSUE #2095: /stow taking forever to finish
Image

4mb1t10n:
## Intent

Bound `/stow`'s foreground wait for live secondmate receipts so one slow or missing reply cannot make the command take many minutes, while keeping late and delivery-unknown replies durable.

Closes #2095

## Decisions

- Firstmate treated preserving delivery-unknown pending receipts and atomically validating every delegated endpoint with the authoritative last-value-wins `kind` parser as required corrections to the new helper's stated safety and durability contracts, not as scope expansion.
- After the approval-blocked workflows completed, Firstmate adopted the No Mistakes pipeline head only after a range-diff proved it preserved all five original commits and added exactly three bounded, in-scope review and documentation fixes.

## Reproduction

Before the fix, I exercised the live-secondmate path in a scratch `FM_HOME` with a fake secondmate endpoint that accepted the marked stow request but never produced its correlated completion receipt. An outer watchdog expired with status 124 while the pending-reply record remained `awaiting_report`, demonstrating that the existing per-secondmate wait could hold `/stow` indefinitely in practice.

The regression test failed before the implementation because the bounded fleet delegate did not exist (status 127 instead of the expected bounded pending status 3). It passes after the fix: the helper dispatches every live secondmate first, applies one shared foreground deadline, returns pending within that bound, and leaves the late-reply record durable. Additional cases cover an immediate correlated receipt and ambiguous remote delivery, which must retain its `delivery_unknown` record.

## Validation

- `tests/fm-stow-cascade.test.sh` — 9 tests passed
- `bin/fm-lint.sh` — passed with pinned ShellCheck 0.11.0
- `bin/fm-doc-audience-check.sh` — passed
- `tests/fm-test-run.test.sh` — passed
- no-mistakes review, test, documentation, and lint gates — passed
- Database — not involved
- Browser validation — not applicable; this repository has no browser-facing surface
- Cypress — not run; this repository has no Cypress suite

The no-mistakes push gate hit the documented read-only `origin` 403, so the validated branch was pushed to the `4mb1t10n` fork as required.

## Pipeline

### DIFF
--- a/.agents/skills/stow/SKILL.md
+++ b/.agents/skills/stow/SKILL.md
@@ -244,8 +244,8 @@ Every home is judged against its own `config/startup-memory-budget` allowance, s
 
 Act on each home by its reported `transport`:
 
-- `agent` - send the marked request with `bin/fm-send.sh fm-<id> "<request>"` so the live secondmate performs its own `/stow`, including the uncaptured knowledge that exists only in its session.
-  Ask it for the same completion receipt this skill defines, and read its reply from its status file or the document it points to, never from its chat.
+- `agent` - collect every live secondmate id, then run `bin/fm-stow-delegate.sh <id>...` once so all marked `/stow` requests are dispatched before their correlated completion receipts share one bounded foreground wait.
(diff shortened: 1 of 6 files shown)


4mb1t10n:

PR #2149, opened 2026-08-11
fix(supervision): keep parked crews and the Stop auto-arm from taking supervision down

Alcb4:
# fix(supervision): keep parked crews and the Stop auto-arm from taking supervision down

Two compounding defects take fleet supervision down whenever a crewmate is waiting on a captain decision.
Observed live four times in one session with two crews running: firstmate had to re-arm the watcher by hand each time, and only the turn-end guard caught it.

Branch head `b63954d6`, 6 commits on top of `7f828ea`, 14 files, +1226/-42.

## Regression window

This home fast-forwarded on 2026-08-10 from `cd73e75` to `85e750a`, 55 commits.
Two of those land squarely on the failing area, and both diffs were read before designing the fix.

- `33a4287` fix(bin): harden Claude supervision auto-arm recovery (#1495).
- `88b2a94` fix(bin): correct session lock and attached watcher supervision (#1545).

The behaviour both commits intended is right, and neither is reverted here.
The bug is in their edges: #1545 widened harness identity to whole path components and `argv[0]` so a version-named Claude Code binary could be recognised, which was necessary and correct, but the same widening also made Claude Code's shared infrastructure processes look like sessions.

## Defect 1 - a crew parked on a captain decision is classified as a possible wedge

`bin/fm-watch.sh` suppressed a stale-pane wedge only through `pause_state_class` -> `status_is_paused_or_captain_held`, which matched exactly two verbs, `paused` and `captain-held`.
A crew awaiting a captain answer writes `needs-decision:`, which is the verb `bin/fm-brief.sh`'s own generated status protocol tells it to use, and that verb was in neither set.

Two things then compounded.
`needs-decision:` is captain-relevant, so `stale_is_terminal` reported the pane terminal and it never reached the pause classification at all.
And a crew parked at an ask-user gate of a still-running validation still reports a running step, so `crew_absorb_class` returned `working` and routed the motionless pane into the provably-working wedge timer.
Its idle pane then escalated 1, 2, 3 with `demand-deep-inspection`, and every armed cycle ended within about 60 to 90 seconds on the parked crew's window rather than the working one.

Two fixes were explicitly ruled out before design and are not used here.
Telling crews to write `paused:` instead is wrong, because that verb declares a bounded EXTERNAL wait expected to clear on its own, and a captain decision is not that.
`captain-held` cannot be the answer either, because it is written only by `bin/fm-decision-hold.sh` after it verifies a structured backlog hold, which is the investigation and visual-review path, so a mid-pipeline ask-user finding answered through the validation gate never travels that path.
The suppression set was what was wrong, not the crew's verb.

### What changed

A still-open keyed captain decision is now the third declared-idle class, alongside the declared external-wait pause and the verified captain-held transfer.

It is read from the durable open-decision fold (`status_open_decisions`) rather than the last status line, so a later unrelated `working:` or `done:` line cannot silently end it, and so the same class covers `blocked:`, which is the same "firstmate action is needed" wait.
It deliberately outranks a provably-working run-step verdict, because that verdict winning is the actual root cause of the escalation loop.
It inverts only the liveness asymmetry of the other two classes: firstmate itself owes the answer, so a live parked endpoint goes straight to the bounded cadence, while a crew that died holding the decision is the anomaly that still gets its one surface.

Two properties were required and are enforced by tests.

1. An open decision must stay visible.
   Suppressing the wedge never suppresses the decision: the fleet-wide `scan_open_decisions` fold surfaces the same open set on every wake drain, and the class ends once the crew's status advances past the line that opened the decision, so suppression is bounded rather than lasting the whole lifetime of an unclosed key.
2. A genuinely dead crew must still surface.
   Liveness is rechecked on every cadence pass, not only at first sight, so a crew that parks while alive and then dies still surfaces; only a confident live verdict clears the one-shot death marker, so an ambiguous `unknown` read cannot make a dead crew flap; and the away-mode daemon clears the same marker, so the invariant also holds across an AFK round trip.

## Defect 2 - the Stop-owned auto-arm stops re-claiming the home

The turn-end guard reported `TURN WOULD END BLIND` with tasks in flight, stating explicitly that the auto-arm had not claimed, four times.
`state/.claude-autoarm-epoch` read `epoch=2246 owner_pid=1208559 outcome=rewake` throughout and never advanced, `state/.turnend-claude-blocks` stayed `count=1 epoch=2246`, and `bin/fm-watch-arm.sh` run manually succeeded every time immediately.
So the arm path was healthy and only the claim path failed.

### The stated hypothesis was tested and disproved

The first hypothesis was that the auto-arm treats an epoch whose recorded owner has since died as already handled and declines to claim.
That was tested against the real hook in four deliberately constructed dead-owner lock states - a symlink lock to an owner dir with a dead pid, a dangling owner-lock symlink, a leftover steal lock, and a plain-directory lock with a dead pid - and disproved.
Every one reclaimed the lock and armed normally, and the epoch ledger's `owner_pid` is never read by any claim path.

### The actual cause

Claude Code hosts an `asyncRewake` Stop hook inside the worker chain of its shared per-user daemon, `claude daemon run` -> `claude bg-pty-host` -> `claude bg-spare`.
That branch never contains the session pid `state/.lock` records, and since #1545 all three of those infrastructure processes carry `claude` branding in `argv[0]`.
So `fm_harness_ancestry_pids` returned a confident ancestry that provably excluded the hook's own session, `fm_session_lock_owned_by_self` failed, `fm_harness_pid_alive` on the recorded owner succeeded, and the hook exited 0 silently on every firing, permanently.

This was confirmed against a live process table on 2026-08-10 with Claude Code 2.1.227, where the daemon, `bg-pty-host` and `bg-spare` processes sat in a different branch from the session recorded in `state/.lock`.

### What changed

Claude Code's shared infrastructure subcommands are no longer treated as a session, in either the ancestry walk or lock-owner liveness.
The match is anchored on the first argument after `argv[0]` specifically so a real session can never be misread as infrastructure because its prompt text happens to contain one of those words - a crewmate launch brief is passed as argv.

`bg-spare` and `bg-pty-host` stay transparent to the pre-match climb, because a session-hosted hook chain legitimately passes through them, but the `daemon` subcommand is a hard stop: reaching it without having already matched a session proves the caller is in the shared infrastructure chain, so the walk terminates unresolved rather than climbing past and latching onto whatever harness sits above the daemon.
That last point matters because the naive narrowing made all three transparent, which would have let the walk cross a gap and report an unrelated harness as this session's ancestry - the exact gap-crossing `test_harness_beyond_a_gap_never_owns_the_lock` exists to forbid, and it would have let the hook arm a home it cannot prove it owns.

A live recorded owner reached with no resolvable ancestry is recorded durably in `state/.claude-autoarm-unresolved-ancestry` and cleared on the next successful claim, and the turn-end guard names that cause instead of reporting only that nobody claimed.
Only the genuine live-competing-owner branch supplies a pid to that banner, so a shared-daemon pid or an already-dead pid can never be rendered as "owned by another live session", and the banner is gated on the record's own freshness so a stale fault cannot misdiagnose a later, different failure.

The guard's bounded block budget was keyed on epoch change, so a frozen ledger pinned the count below its threshold and the designed attended fail-open was unreachable.
Same-epoch suppression now applies only while the auto-arm demonstrably owns that epoch, or within a single guard run, which preserves the existing one-recovery-turn-per-epoch and failed-suppressed idempotence contracts.

## Migration window: why a pre-upgrade shared-infrastructure lock is stale but NOT claimable

Excluding the infrastructure subcommands also changes the lock WRITE path (`fm_harness_ancestry_pid`, used by `bin/fm-lock.sh`), so a `state/.lock` written before this change can hold the shared daemon pid.
Without a deliberate fix, a different live session's Stop hook would see that recorded owner as not live and could claim a home another live session recorded.

This was raised by review as a net narrowing that self-heals on first re-acquire, and it was fixed anyway rather than accepted as a documented transient.
The reasoning is recorded because it overrides that judgement.
This is a session lock, whose failure mode is two live sessions both believing they own one home and both mutating it.
The window opens during a fleet self-update, which means it opens for every home AT ONCE rather than rarely - the worst possible distribution for a transient, not a mitigating one.
It is also the precise boundary this change promises to preserve, that a genuine competing live session stays inert, and a net narrowing that still briefly permits the thing the boundary forbids is not good enough for a lock.

So such a recorded owner is recognised as not a valid session owner AND is not claimable by a Stop hook, which leaves the ordinary session-start acquisition to replace it with a real session pid with no window.

## What this change does NOT fix

For the exact ledger state observed live - `epoch=2246`, `outcome=rewake`, never advancing - the block-budget change ALONE produces no observable difference.
The epoch file's mtime ages past `EPOCH_FRESH`, so `autoarm_owns_recovery`'s rewake branch never accounts that epoch, and the unowned fall-through then correctly advances `COUNT` past `BLOCK_BUDGET`.
But `terminal_fail_open` still declines, because `failure_episode_verified` accepts only `outcome=failed` or `failed-suppressed`, so the guard re-blocks every turn exactly as before until Claude Code's own consecutive-block override ends the episode.
The real improvement for that state is the new unresolved-ancestry banner line, not the budget.
Gating the alarm on a recorded failure is a deliberate existing contract and is left unchanged; this is stated plainly as a limitation rather than overstated.

## Tests, and the proof each fails before the fix

Every regression was proven failing against the previous code before the code was changed, by stashing only `bin/` and running the new tests against pristine sources.
Observed failures on the unfixed code:

- `test_open_captain_decision_stale_is_bounded_not_wedged` failed with `stale: test:fm-parked (idle 1s, possible wedge, escalation 1)` - the parked crew ended the supervision cycle.
- `test_daemon_worker_chain_is_never_session_ancestry` failed because the shared `claude daemon run` was reported as this session's harness ancestry.
- `test_e2e_unresolvable_ancestry_is_recorded_not_silent` failed because the firing left no durable record, so it repeated silently forever.
- `test_hook_claude_mode_frozen_epoch_still_advances_the_block_budget` failed because a frozen epoch never reached the attended fail-open.

All pass after the fix, and the full suites are green.

New coverage added across three existing suites, extending them rather than inventing a runner:

`tests/fm-watch-triage.test.sh`
- `test_open_captain_decision_stale_is_bounded_not_wedged`
- `test_open_captain_decision_dead_crew_still_surfaces`
- `test_open_captain_decision_park_then_die_still_surfaces`
- `test_open_captain_decision_class_ends_when_the_crew_moves_again`
- `test_open_captain_decision_death_surfaces_once_across_an_unknown_read`
- `test_open_captain_decision_dead_holder_recheck_names_the_exited_agent`
- `test_open_captain_decision_dead_crew_survives_an_afk_round_trip`

`tests/fm-session-lock-ancestry.test.sh`
- `test_daemon_worker_chain_is_never_session_ancestry`
- `test_harness_above_the_daemon_is_never_this_sessions_ancestry`
- `test_e2e_unresolvable_ancestry_is_recorded_not_silent`
- `test_e2e_unresolvable_ancestry_is_recorded_on_a_malformed_lock`
- `test_e2e_competing_live_session_stays_silently_inert`
- `test_e2e_pre_upgrade_daemon_pid_lock_is_stale_but_not_claimable`

`tests/fm-turnend-guard.test.sh`
- `test_hook_claude_mode_frozen_epoch_still_advances_the_block_budget`
- `test_hook_claude_mode_unresolved_ancestry_banner_needs_a_current_record`
- `test_hook_claude_mode_banner_never_calls_a_non_live_pid_a_live_session`

All are portable regressions that pin the logic with real processes and no harness: they drive the real `fm-watch.sh`, the real Stop auto-arm inside real orphaned process trees, a deterministic fake process table for both platforms' `ps` semantics, and the real guard.
None asserts implementation-source bytes, through parsers, regexes, snapshots or wrappers.

A hard constraint applies to these suites and was violated once during development, which is worth recording: a test must never invoke the blocking watcher without a bound the test itself controls.
A watcher blocks by design until it has an actionable wake, so in a fixture worktree with no crews it never returns.
The divergence assertions matter too, so no case can go quietly vacuous: the competing-live-session case asserts no fault record is written, and the cursor case asserts the incremental fold still reports the open decision after the watcher has advanced the shared cursor.

## Per-harness compatibility review

Reviewed against every supported primary harness, not only the one in this fleet.

- **claude** - directly affected on all three axes (Stop auto-arm identity, the unresolved-ancestry record, the guard's block budget). Covered by `fm-claude-stop-autoarm.test.sh`, `fm-session-lock-ancestry.test.sh` and `fm-turnend-guard.test.sh`.
- **codex** - bounded foreground checkpoints. Reaches `fm-turnend-guard.sh` only in default mode, which blocks before the changed Claude-mode budget path. Never launched with a bare `daemon`, `bg-pty-host` or `bg-spare` first argument, so the identity narrowing cannot false-positive on it.
- **opencode** and **pi** / **pi-signed** - plugin and extension driven turn-ends, same default-mode guard path. The `pi-signed` -> `pi` wrapper ancestry is untouched, because the narrowing keys on those three subcommands only.
- **grok** - background-notify cycles with native Stop delegation, same default-mode guard path.
- **kimi** - hook registry token path, unaffected for the same reasons.
- **muse** - crewmate and scout only, never a primary; affected only as a crew through the harness-neutral classifier change.

The shared classifier (`bin/fm-classify-lib.sh`) and watcher (`bin/fm-watch.sh`) changes are harness-neutral and apply to all of them, which is why the open-decision class is covered by portable regressions rather than harness-specific ones.
`docs/verification/supervision.md` was updated because this work falsified its recorded claim that the `bg-spare` hook worker chain always reaches the lock owner.

## Known scope boundaries, filed rather than fixed

- **Away mode does not park an open-decision crew on the same cadence.** The new class is watcher-only: `afk_present` is evaluated before it, and the other call site is guarded by `! afk_present`, so under away mode the class is inert and `bin/fm-supervise-daemon.sh` behaves exactly as it does on `main`. Away mode is not newly broken, and `docs/configuration.md` is scoped to that actual behaviour rather than claiming parity. Extending the class to away mode would be its own scoped change.
- **The ShellCheck environment gap.** `bin/fm-lint.sh` is the single owner of the lint definition and refuses to run under any version other than the pinned 0.11.0. Where that binary is absent the lint step exits 127, which is a missing tool rather than a lint failure; `bin/fm-lint.sh` was run against this exact head with pinned ShellCheck 0.11.0 and passes clean, and `bin/fm-doc-audience-check.sh` passes.

## Notes on scope

`AGENTS.md` is touched only to add the one new never-touch state file, `state/.claude-autoarm-unresolved-ancestry`, to its always-loaded state-directory layout inventory, which would otherwise be wrong.
Everything else about script behaviour lives in the scripts' own headers and in the classified documentation owners.
Incident chronology is deliberately kept out of tracked docs and lives here in the PR body.


Alcb4:
This ran through no-mistakes green, but as it wasn't forked correctly it was denied. I'm token short so will pass it next week.

PR #2151, opened 2026-08-11
fix(brief): support configurable ship branches

### LINKED ISSUE #1887: fm-brief.sh hardcodes the ship branch as fm/<task-id>, so every brief needs hand-editing to follow a team's branch convention

4mb1t10n:
## Intent

Make ship branch names configurable without weakening the generated brief's safety contract or creating an unmergeable local-only task.

Closes #1887

## What changed

- Add `fm-brief.sh --branch <name>`, defaulting to the existing `fm/<task-id>` name.
- Reject empty, leading-dash, whitespace, quote, shell-unsafe, and invalid Git branch names.
- Substitute the selected branch throughout ship briefs while leaving scout and secondmate scaffolds unchanged.
- Record the selected ship branch in task metadata.
- Make the local-only merge gate and live bearings PR mapping consume the recorded branch with the legacy default for older task records.
- Add behavioral regressions for the reported custom-convention condition, metadata persistence, local-only landing, bearings mapping, validation failures, and default compatibility.

## Decisions

- Firstmate required the coordinated metadata, merge, and bearings scope because issue #1887 explicitly establishes that a brief-only override creates an unsafe unmergeable local-only task.
- Firstmate determined that an earlier abandoned narrow implementation, preserved at `fork/fm/fm-1887`, must deliberately not be used as this branch's base because it omits the consumer changes required by the issue.
- The bearings join reads task metadata paths already supplied by the canonical snapshot and streams its lookup JSON through stdin so it remains composable with PR #2137's jq-streaming work.
- Legacy briefs and task records fall back to `fm/<task-id>`, preserving existing homes and promoted scouts without migration.

## Validation

- `bash tests/fm-brief.test.sh`
- `bash tests/fm-spawn-worktree-settle.test.sh`
- `bash tests/fm-pr-merge.test.sh`
- `bash tests/fm-bearings-snapshot.test.sh`
- `bash tests/fm-tangle-guard.test.sh`
- `bin/fm-lint.sh`

The no-mistakes pipeline could not start because its internal gate retained the preserved earlier branch history and correctly rejected a non-fast-forward overwrite.
The coordinated implementation was instead pushed under the distinct fork branch requested by firstmate.


4mb1t10n:
## Conflict resolution against `main` (head `c071136`)

This branch had gone CONFLICTING/DIRTY against `main`. It now merges cleanly. Recording the decisions made while resolving, per the `## Decisions` convention above.

### Decisions

- **Resolved by merging `main` into the branch rather than rebasing onto it.** A rebase rewrites the published commits, so the only way to publish it to this PR would be a force-push, which the delivery constraints forbid. A merge reaches the same end state - the branch now contains current `main` and GitHub reports `mergeable: true` - while preserving every previously published commit and pushing as an ordinary fast-forward (`fdebda2..c071136`, no force). The rebase was performed and fully validated first; its tree is byte-identical to the merged tree (`git diff` between them is empty), so the merge carries the same verified content.
- **`AGENTS.md` meta-field line: took `main`'s version and moved this change's documentation into the producing script.** `main` has since restructured that line to delegate exact field ownership to each producing script's header. This branch's version restated the full field list inline, including the new `branch=`. Keeping the inline restatement would have reintroduced exactly the duplication the one-owner rule forbids, so `main`'s delegating line wins and `branch=` is now documented in `bin/fm-spawn.sh`'s own header next to the code that writes it. `AGENTS.md` consequently drops out of this PR's diff entirely.
- **`bin/fm-spawn.sh` relaunch-preserved key list: took the union.** `main` added `spawn_gen` while this branch added `branch`; both are required, and dropping either would silently lose a field across a relaunch. The merged list carries both.

### Re-verification at the merged head

A rebase or merge invalidates any prior exact-head conclusion, so every gate was re-run at `c071136`:

- `bin/fm-lint.sh` - clean (ShellCheck 0.11.0, pinned).
- `bin/fm-doc-audience-check.sh` - ok, surfaces=67, local_links=240.
- `tests/fm-brief.test.sh`, `tests/fm-pr-merge.test.sh`, `tests/fm-bearings-snapshot.test.sh`, `tests/fm-tangle-guard.test.sh`, `tests/fm-spawn-worktree-settle.test.sh` - all pass, 0 failed.
- The bearings count pinned in `.github/workflows/ci.yml` was confirmed against the real suite output (42 `ok -` lines, matching the updated assertion).

The regressions were also confirmed to fail against pre-fix sources rather than only asserting the new happy path. With `bin/fm-merge-local.sh`, `bin/fm-brief.sh`, `bin/fm-bearings-snapshot.sh`, and `bin/fm-spawn.sh` reverted to `main` and the tests left at their fixed versions:

```
not ok - recorded custom branch did not land: ...
not ok - no-mistakes brief did not record the custom branch contract
```

Both pass again with the fix restored.

### Safety boundaries re-confirmed

Generated briefs were diffed default vs `--branch feat/1887/short-desc`. The only differences are the branch name itself; the worktree-isolation assertion, the `Delivery contract: mode=` line, and the status protocol are byte-identical, and an omitted `--branch` still produces exactly `fm/<task-id>`.

Branch-name validation was re-exercised at this head: empty, leading-dash, whitespace, both quote forms, backslash, semicolon, `$(...)`, tilde, caret, colon, glob, brace, trailing slash, `..`, and control characters are all rejected, while conventional names including the issue's own `feat/646/migration-schema` are accepted.

4mb1t10n:

PR #2155, opened 2026-08-11
feat(bin): bind every recorded commitment to a probe that says whether it is real

sbracewell64:
## Intent

Build the typed register of RECORDED-BUT-NOT-YET-REAL commitments, make it fail closed at session start, and register the network-egress ruling as its first entry.

THE PROBLEM. A commitment recorded with no enforcement probe reads as protection while protecting nothing, and this fleet hit that exact failure four separate ways in one day. (1) A ruling recorded and not enforced: data/captain-rulings-2026-08-06/capability-host-network-egress.md holds that agents get no unrestricted raw egress and that network is a typed Runtime-issued capability the host ENFORCES, while in fact every worker launches with enforcement disabled by construction and the capability catalog's own network.request row honestly records owner null, verifier null. (2) A guard with no caller: task_base_verify_branch was believed to close an inversion and has no runtime caller at all. (3) A derived-state row that went stale while being trusted: fm-decision-surface.sh owners reports invoking_known_next_stage as PENDING while CFVC-15 landed exactly that replacement, in a hand-maintained literal inside the composer that exists to stop firstmate hand-maintaining state. (4) A dated exception that expired into prose: E-05 self-expired 2026-08-08 and the writes were applied because a human filed a reminder, so the mechanism that saved it was diligence, not structure. Each was found by a person looking; none would have surfaced on its own.

WHAT TO BUILD. A machine-readable register carrying, per entry: a stable id; what was recorded and where the authoritative record lives (a ruling file, a captain authorisation, a decision hold, a config exception); the ENFORCEMENT PROBE, a path, symbol, command or predicate whose success means the commitment is now real; the expected state, so the probe's answer is interpretable; and the current derived state, computed, never written. The probe is the load-bearing field, and an entry without one is not admissible, because that is exactly the shape of the four failures above.

FAIL CLOSED AT SESSION START. Surface unenforced entries where they cannot be missed and make the absence of an all-clear structural. Session start must NOT be able to report a clean or quiet state while an entry's probe says the commitment is unenforced or its work is undispatched. An entry whose probe cannot RUN is could-not-observe, reported as such, and never read as enforced - three values, as everywhere else in this fleet. An entry whose probe now PASSES retires automatically, because a commitment that became real should stop nagging without anyone editing a list, and a register that needs hand-maintenance is the defect it was built to fix. bin/fm-bootstrap.sh already runs detect-only checks at every session start and already has an actionable-diagnostic vocabulary to print into: extend it, and do not build a parallel surfacing path.

FIRST ENTRIES. network-egress-ruled-not-enforced, whose probe is workers launched with permission enforcement disabled and network.request's owner and verifier being null; expected state after closure is enforcement present, current state is RULED-NOT-ENFORCED. This is the captain's "make the gap visible now" and is the only part of the network-egress work that starts now; the steps that actually close it are staged separately behind network-egress-enforcement-visible-then-staged and must NOT be attempted here. Plus captain-authorised work that is not yet in flight: an authorisation is a commitment like any other, and its probe is whether a live task, branch or pull request owns it.

THIRD ENTRY CLASS, CAPTAIN-RULED 2026-08-10 and folded into the schema rather than deferred: a ruled review finding. Measured four times in one day, a ruled criterion was reported APPLIED while not being MET, every time caught only by a worker re-checking its own work, and firstmate had made the criterion explicit in writing before the fourth and it still happened. Root cause is identical to the other entries: the criterion lives in firstmate prose, the fix lives in code, and nothing binds them. Applied is an action; met is a predicate, and the pipeline only knows the action happened. A ruling that says fix must supply a PROBE whose success means the criterion is met, and a resolved status event for a key that has a registered probe must NOT be accepted as resolved until that probe passes, with the open-decision fold continuing to show it. bin/fm-classify-lib.sh already owns keyed open and resolved semantics and bin/fm-status-event-lib.sh owns the envelope, so extend that fold and do not add a surface.

THREE TIERS, and the third is not optional because forcing a probe where none is possible would recreate this same failure one level up. executable: the probe runs directly and its exit status is the answer. cited-control: the probe confirms a named test exists and passes now while the watched-it-fail observation is cited as a named artifact rather than a claim - this is the strongest routinely achievable form and is the DEFAULT. attested: for a criterion that genuinely cannot execute, such as a comment being accurate; that tier must be MARKED and VISIBLE and never silently read as verified. A probe that cannot RUN is could-not-observe, never a pass.

THE PINNED BLOCK FORMAT, built to exactly, so what firstmate writes and what the register reads cannot drift. Firstmate writes decision files at data/<task-id>/decision-<key>.md, and every such file that directs fix carries exactly one fenced probe block with tier of executable, cited-control or attested; run is required for the first two and is a command run from the task worktree; control is required for cited-control and names the test or artifact watched to fail first; reason is required for attested, which carries no run. Exit 0 means the criterion is met, non-zero means not met, and unable to execute means could-not-observe and never a pass. Firstmate is applying this format to its own rulings from 2026-08-10 onward, before the mechanism lands.

NO BACK-FILLING. Do not back-fill decision keys ruled before 2026-08-10 with invented probes. They read could-not-observe, on the same principle that forbids back-filling historical verifier identities, because a fabricated record is indistinguishable from a real one afterwards.

WHAT THIS MUST NOT DO. It must not discourage a worker from re-checking its own work: all four catches came from exactly that, and the probe is a floor beneath diligence, not a replacement for it.

WHAT NOT TO DO. Do not remove or weaken any bypass flag: that work is staged behind a capability path that does not exist yet, disarming workers first breaks every lane and provides nothing, and the commission is explicit that bypasses retire only after bindings and adapters are proven. Do not reopen the network-egress ruling, which commission section 8 forbids: this is an implementation gap the ruling already authorises closing, not a question needing an answer. Do not suppress an entry by age, count or rate, which is the same mistake as suppressing a repeat escalation because it hides a genuine unmet commitment along with the noise; the fix is to make truth visible, never to make the question quieter. Do not build a greenfield reconciliation runtime: bin/fm-decision-surface.sh is already this fleet's derived-state composer with typed verdicts including unevaluable, and bin/fm-verify-lib.sh already owns the three-valued observation type, so extend them and say plainly if either genuinely does not fit.

ACCEPTANCE CRITERIA. Red-capable: register an entry whose commitment is unmet, prove session start surfaces it and cannot report all-clear, then satisfy the probe and prove the entry retires with no hand edit. An entry whose probe cannot run reports could-not-observe and is proven distinguishable from both enforced and unenforced. The four failures listed above are each expressible as an entry - not all four need registering, but the shape must be demonstrated to cover them, and any that does not fit is a finding worth reporting. A hand-written status word cannot satisfy any entry, demonstrated.

CONSTRAINTS. This is firstmate's shared tracked material, governed by the firstmate-coding-guidelines skill. One sentence per line in prose; plain dashes; shellcheck-clean bin via bin/fm-lint.sh; colocated tests; never add an agent as a commit co-author; never write the captain-address convention into a commit message, pull request, or anything crewmates and other tools read. Session start is on the critical path of every session in this fleet, so anything added there must be cheap, detect-only, and must not mutate: a slow or noisy session start will be turned off, and then it protects nothing. git stash is SHARED across every worktree of this repo and must never be used in a pooled worktree. Several lanes are live and touching bin/, so expect to rebase and resolve a clean reconciliation per the captain's standing delegation.

LATER ACCEPTED REQUIREMENTS, folded in and superseding anything narrower above.

SESSION START MUST EXECUTE NO PROBE - this is a safety requirement, not a performance one. Session start may report RECORDED state; it may never execute a probe to find out. The whole chain counts, not just the bootstrap relay: fm-session-start.sh runs fm-admission.sh, which runs fm-fleet-snapshot.sh, which calls status_open_decisions per task, which reaches the closure gate and would run arbitrary decision run: commands inside task worktrees. Turning a cheap detect-only sweep into arbitrary execution on the critical path of every session is a regression that gets session start turned off, and then the register protects nothing. Not executing must NOT mean accepting: a resolution whose probe was not run stays visibly unverified and still open, never silently closed, because fail-open here is worse than the cost it avoids.

PROBE-RESULT CACHING, with a hard constraint. A cached probe result must carry its observation time, and a stale one must re-run or report could-not-observe; it must never silently serve an old answer as a current one, because caching a derived fact is exactly how derived state goes stale while being trusted. Key the cache on what the probe's answer actually depends on - the decision file bytes plus the worktree head - never on the task status file cksum, because the open-decision fold is DRIVEN by status appends so that key invalidates precisely when it is needed and helps only idle tasks. The probe bound is 10s. The observation time must be surfaced on the path that CONSUMES the result: bin/fm-classify-lib.sh must not capture the closure gate's stdout and discard it on the accept path, or the guarantee exists only for a human running the command by hand, which is a criterion reported applied and not met.

DECLARED-BUT-UNPROBED CONDITIONS. An entry whose recorded commitment has a half no probe reaches must declare that half rather than dropping it, and must not retire on the covered half alone: a half-covering probe returning one confident value is a check that cannot see something giving the reassuring answer. The declaring field's type is validated, and a malformed one makes the entry inadmissible rather than silently skipping the guard and letting a passing probe reach SATISFIED on a partial answer.

DERIVE, DO NOT ENUMERATE. If a list enumerates something the codebase already knows, derive it; a hand-maintained literal goes quietly vacuous the day a member is added, which is one of the measured failure shapes. Where a member genuinely cannot be derived, an unknown member yields could-not-observe, never silent exclusion.

NEW PERSISTENT STATE. Any new artifact under state/ is added to the AGENTS.md state inventory with its writing owner and lifecycle, and is reaped by teardown.

RULING PROBES ASSERT SPECIFIC CASES BY EXACT WORDING. Two firstmate rulings carry probes that grep the test suite for exact `ok` line substrings, so the corresponding cases must be named to match them verbatim: `unknown harness member yields could-not-observe`, `stale row asserts satisfied only when the probe inspects the row`, `session start executes no probe`, and `cached result carries observation time on the accept path`. A probe must assert the SPECIFIC case rather than merely invoking the suite that would contain it, because a suite passes whenever the suite is green, which says nothing about coverage.

OUT OF SCOPE for this round, agreed as a separate follow-up commit, and must NOT be added here: the registration gate refusing a probe that already passes at registration time; the externally-blocked third state and its guards; the property-asserting probe kind; the two probe kinds for project-worktree resolution and battery results; the rebase-drop-defect-remains-unprevented entry; and the schema document with worked examples per tier. A round that grows while it is being validated is how a fix ships without its criterion met.

## What Changed

- Adds `bin/fm-commitment-register.sh` plus a tracked `commitments/` register (`schema.json` and two entries: `network-egress-ruled-not-enforced`, `network-egress-enforcement-staged-work`). Each entry declares a typed probe (one of six kinds) whose result derives `SATISFIED` / `UNMET` / `UNOBSERVED` on every read - never stored - so an entry retires with no hand edit once its probe passes, an entry with a malformed or missing probe is inadmissible, and an entry declaring an unprobed half in `unobserved_conditions` cannot retire on the covered half alone. Probe targets are constrained to the tracked code root, and the path-bearing fields are derived from the schema rather than enumerated.
- Makes the absence of an all-clear structural: `bin/fm-bootstrap.sh` relays open entries as `COMMITMENT:` diagnostics (and reports an absent interpreter loudly rather than silently), and `bin/fm-classify-lib.sh`'s existing keyed open/resolved fold now refuses a `resolved` event whose `data/<task-id>/decision-<key>.md` carries a pinned probe block until that probe passes, surfacing the refusal reason and the cached result's observation time on the consuming path. Session start's two calls that reach that fold (`fm-wake-drain.sh`, `fm-admission.sh`) carry `FM_COMMITMENT_NO_DECISION_RUN=1`, so the closed set of typed probes still runs there but no arbitrary decision-file `run:` executes on the critical path - and an unevaluated key stays visibly open, never silently closed.
- Supporting changes: `bin/fm-launch-lib.sh` gains a derived harness roster and `launch_permission_posture` accessor (unparseable arms stay in the roster as unknown rather than being dropped), probe results cache under `state/commitment-probe-cache/` keyed on decision-file bytes plus worktree head with a freshness bound and reaped by `bin/fm-teardown.sh`, `bin/fm-test-run.sh` maps the new suite into changed-file selection, and AGENTS.md / docs / the bootstrap-diagnostics and ask-user-authority skills record the register, its state artifact, and the probe-block format. Colocated coverage lands in `tests/fm-commitment-register.test.sh` alongside additions to the bootstrap, session-start, launch-lib and teardown suites.

## Risk Assessment

✅ Low: Both round-4 items are fixed and the derivation is genuinely load-bearing rather than decorative - I independently reproduced the schema-driven path-arg set, confirmed it matches the script's own call sites via the new bidirectional control, and verified the vacuity case fails closed - leaving only a stale copy of superseded wording that lives outside the repository.

## Testing

Ran the five targeted suites this change touches (commitment register, bootstrap, session start, launch-lib, teardown) and they all pass, after isolating two pre-existing host-environment leaks in the fixtures (a system /usr/bin/node defeating a test's node removal, and a populated /tmp/fm-test-absent-nm-home) by pointing FM_TEST_BASE_PATH and TMPDIR at clean locations. Because passing tests alone do not show the intent, I also drove the real operator commands end-to-end and captured CLI transcripts: the shipped register reporting both first entries UNMET against this repository, a real fm-session-start.sh going from "(silent - all good)" to a COMMITMENT line and back to silence once the probe passed with the entry file byte-identical, SATISFIED/UNMET/UNOBSERVED proven distinguishable from a single entry, a hand-written status word refused, a half-covered commitment unable to retire, a decision-file run: command provably not executed by session start yet still not accepted, cached-verdict provenance surfaced on the fold's accept path and in fm-fleet-snapshot.sh --json, and all four measured failure shapes expressed as real entries each with its control. This is a shell/CLI surface with no rendered UI, so command transcripts are the end-user experience and no screenshot applies. The worktree is clean and scratch directories were removed.

<details>
<summary>Evidence: End-to-end transcript: session start red -&gt; green -&gt; retires, three values, refused status word, no-probe-execution guard, cache provenance</summary>

```text


=================================================================
SCENE 1 - THE SHIPPED REGISTER, as an operator reads it
=================================================================
The two first entries the change registers, against this repository.

$ bin/fm-commitment-register.sh --open   # the exact lines session start relays
COMMITMENT: network-egress-enforcement-staged-work UNMET (AUTHORISED-NOT-OWNED) - nothing owns network-egress-enforcement-visible-then-staged: no live task record, no local branch fm/network-egress-enforcement-visible-then-staged, and no in-flight or done backlog record
COMMITMENT: network-egress-ruled-not-enforced UNMET (RULED-NOT-ENFORCED) - these harnesses launch a worker with permission enforcement disabled: claude codex opencode pi pi-signed grok; and this commitment further requires the capability catalog's network.request row names an owner and a verifier rather than carrying owner null and verifier null, which has no landed artifact to probe because that catalog exists as a proposal in data/cfvc-harness-and-capability/report.md section 6.3 rather than as tracked code, which no probe here observes
(exit 3 - 3 means at least one entry is UNMET)

$ bin/fm-commitment-register.sh network-egress-ruled-not-enforced
Commitments recorded, and whether a probe says they are real yet.

  UNMET       network-egress-ruled-not-enforced   RULED-NOT-ENFORCED
              recorded:   A bound agent receives no unrestricted or raw network egress; network is a typed capability the agent may request, Runtime decides, and the capability host enforces.
              authority:  data/captain-rulings-2026-08-06/capability-host-network-egress.md (captain ruling, 2026-08-06)
              real when:  Every harness firstmate can launch composes a session whose permission enforcement is active, AND the capability catalog's network.request row names an owner and a verifier. The probe below answers the first of those; the second is declared unobserved, so this entry cannot retire on the first alone.
              tier:       executable
              observed:   these harnesses launch a worker with permission enforcement disabled: claude codex opencode pi pi-signed grok; and this commitment further requires the capability catalog's network.request row names an owner and a verifier rather than carrying owner null and verifier null, which has no landed artifact to probe because that catalog exists as a proposal in data/cfvc-harness-and-capability/report.md section 6.3 rather than as tracked code, which no probe here observes
              not probed: the capability catalog's network.request row names an owner and a verifier rather than carrying owner null and verifier null, which has no landed artifact to probe because that catalog exists as a proposal in data/cfvc-harness-and-capability/report.md section 6.3 rather than as tracked code
              note:       The ruling has two halves and only one of them has landed code to probe today. The probe answers the landed half: every worker launches with permission enforcement disabled, so the prohibition is recorded and not enforced. The other half is declared in unobserved_conditions rather than being probed or quietly dropped, because naming a not-yet-decided owner path would produce a probe that can never pass and dropping it would let this entry retire while network.request still carries owner null and verifier null - the verdict would then claim more than it observed. When that catalog lands as tracked code, this entry gains a second probe over it and the declaration goes away. Closing the gap is staged separately and this entry authorises none of it: retiring a bypass flag before bindings and adapters are proven would disarm every lane and enforce nothing.

0 satisfied · 1 unmet · 0 could-not-observe
A state here is computed from the probe on every read and is never stored.
(exit 3)


=================================================================
SCENE 2 - RED: session start cannot report a clean or quiet state
=================================================================
registered entry cksum: 3473269162 359

$ CONTROL FIRST: the same session start with an EMPTY register - bootstrap's contract is silent = all good
BOOTSTRAP
--------------------------------------------------------------------------------
(silent - all good)

WAKE QUEUE

$ bin/fm-session-start.sh   # BOOTSTRAP section only
BOOTSTRAP
--------------------------------------------------------------------------------
COMMITMENT: demo-owner-not-enforced UNMET (RULED-NOT-ENFORCED) - the declared owner owner.sh is not present and executable, so nothing performs this commitment

WAKE QUEUE


=================================================================
SCENE 3 - GREEN: the commitment becomes real, the entry retires with no hand edit
=================================================================

$ printf ... > $FM_ROOT/owner.sh   # the commitment is now enforced by a real owner

$ bin/fm-session-start.sh   # BOOTSTRAP section only
BOOTSTRAP
--------------------------------------------------------------------------------
(silent - all good)

WAKE QUEUE

registered entry cksum before: 3473269162 359
registered entry cksum after:  3473269162 359
IDENTICAL - the entry retired on its probe, not on a hand edit.

$ bin/fm-commitment-register.sh demo-owner-not-enforced
Commitments recorded, and whether a probe says they are real yet.

  SATISFIED   demo-owner-not-enforced (retired: the probe passed)
              recorded:   the declared owner enforces this commitment
              authority:  demo transcript
              real when:  the declared owner exists and answers
              tier:       executable
              observed:   the declared owner owner.sh answered

1 satisfied · 0 unmet · 0 could-not-observe
A state here is computed from the probe on every read and is never stored.
(exit 0 - 0 means every entry is SATISFIED)


=================================================================
SCENE 4 - THREE VALUES: could-not-observe is neither enforced nor unenforced
=================================================================

$ the owner answers nothing at all - it ran, but it observed nothing
Commitments recorded, and whether a probe says they are real yet.

  UNOBSERVED  demo-owner-not-enforced   could-not-observe
              recorded:   the declared owner enforces this commitment
              authority:  demo transcript
              real when:  the declared owner exists and answers
              tier:       executable
              observed:   the declared owner owner.sh exited 0 printing nothing, which answers nothing

0 satisfied · 0 unmet · 1 could-not-observe
A state here is computed from the probe on every read and is never stored.
(exit 4 - 4 means at least one entry is UNOBSERVED)

$ bin/fm-session-start.sh   # a could-not-observe entry still denies silence
BOOTSTRAP
--------------------------------------------------------------------------------
COMMITMENT: demo-owner-not-enforced COULD-NOT-OBSERVE - the declared owner owner.sh exited 0 printing nothing, which answers nothing

WAKE QUEUE

Side by side, the same entry file under three worlds:
  owner absent          -> UNMET
  owner answers nothing -> UNOBSERVED
  owner answers         -> SATISFIED


=================================================================
SCENE 5 - A HAND-WRITTEN STATUS WORD CANNOT SATISFY AN ENTRY
=================================================================

$ add "state": "SATISFIED" to an entry whose probe would otherwise pass
Commitments recorded, and whether a probe says they are real yet.

  UNOBSERVED  demo-asserted   could-not-observe
              recorded:   the declared owner enforces this commitment
              authority:  demo transcript
              real when:  the declared owner exists and answers
              tier:       executable
              observed:   inadmissible entry: entry carries a hand-written "state"; a status word must not be able to satisfy a commitment

0 satisfied · 0 unmet · 1 could-not-observe
A state here is computed from the probe on every read and is never stored.
(exit 4)

$ bin/fm-session-start.sh   # the asserted entry still denies a quiet start
BOOTSTRAP
--------------------------------------------------------------------------------
COMMITMENT: demo-asserted COULD-NOT-OBSERVE - inadmissible entry: entry carries a hand-written "state"; a status word must not be able to satisfy a commitment

WAKE QUEUE


=================================================================
SCENE 6 - A VERDICT NEVER CLAIMS MORE THAN THE PROBE OBSERVED
=================================================================

$ the same passing probe, plus a declared half no probe reaches
Commitments recorded, and whether a probe says they are real yet.

  UNOBSERVED  demo-half-covered   could-not-observe
              recorded:   the declared owner enforces this commitment
              authority:  demo transcript
              real when:  the declared owner exists and answers
              tier:       executable
              observed:   the declared owner owner.sh answered - but this commitment also requires the second half of this commitment has no landed artifact to probe, which no probe here observes, so only part of it is answered
              not probed: the second half of this commitment has no landed artifact to probe

0 satisfied · 0 unmet · 1 could-not-observe
A state here is computed from the probe on every read and is never stored.
(exit 4 - the covered half passed, and the entry still cannot retire)

$ a MALFORMED declaration must make the entry inadmissible, not silently skip the guard
Commitments recorded, and whether a probe says they are real yet.

  UNOBSERVED  demo-half-malformed   could-not-observe
              recorded:   the declared owner enforces this commitment
              authority:  demo transcript
              real when:  the declared owner exists and answers
              tier:       executable
              observed:   inadmissible entry: unobserved_conditions must be a non-empty array of non-empty strings; a malformed one would silently skip the guard that stops a passing probe retiring a half-observed commitment

0 satisfied · 0 unmet · 1 could-not-observe
A state here is computed from the probe on every read and is never stored.
(exit 4)


=================================================================
SCENE 7 - SESSION START EXECUTES NO DECISION-FILE run:, AND NOT RUNNING IS NOT ACCEPTING
=================================================================

$ cat $FM_HOME/data/demotask/decision-crit.md   # the pinned block firstmate writes
# Firstmate ruling - demo

Action: **fix**

`` `probe
tier: executable
run: printf ran > /tmp/fm-tt-eXQhWe/demo/the-probe-ran
`` `

$ bin/fm-session-start.sh   # the whole chain: session start -> admission -> fleet snapshot -> open-decision fold
session start finished.
MARKER ABSENT - session start executed no decision-file run: command.

$ and NOT RUNNING IS NOT ACCEPTING - the decision is still open, visibly unverified
crit	needs-decision	NOT RUN: this caller executes no decision-file run command, so the probe for crit was not executed and the criterion stays unverified rather than accepted

$ the same fold OUTSIDE session start runs the probe and accepts the resolution
(the decision closed: no open rows)
MARKER PRESENT - outside session start the run: executed, so the guard is scoped, not a global mute.


=================================================================
SCENE 8 - A CACHED VERDICT CARRIES ITS OBSERVATION TIME ON THE ACCEPT PATH
=================================================================
The fold just accepted on a probe run seconds ago. Run it again: the answer is
served from the freshness-bounded cache and says so, on the path that CONSUMES it.

$ the open-decision fold again (stderr is the disclosure channel)
decision demotask [key=crit] accepted on a probe result: the probe for crit exits 0, so the criterion is met [observed 2026-08-11T05:05:50Z, 0s ago, within the 120s freshness bound]

$ a criterion whose probe FAILS refuses the closure and keeps the decision showing
decision demotask [key=crit] accepted on a probe result: the probe for crit exits 0, so the criterion is met [observed 2026-08-11T05:05:50Z, 0s ago, within the 120s freshness bound]
crit2	needs-decision	the criterion is not met: the probe for crit2 exited 1: 

$ and the same task through a real product surface: bin/fm-fleet-snapshot.sh --json
{
  "id": "demotask",
  "open_decisions": [
    {
      "key": "crit2",
      "verb": "needs-decision",
      "summary": "the criterion is not met: the probe for crit2 exited 1:  [observed 2026-08-11T05:05:50Z, 0s ago, within the 120s freshness bound]"
    }
  ]
}


Demo complete.
```
</details>
<details>
<summary>Evidence: Real fm-session-start.sh output proving the decision-file run: was NOT executed yet NOT accepted</summary>

<code>BOOTSTRAP&#10;--------------------------------------------------------------------------------&#10;(silent - all good)&#10;&#10;WAKE QUEUE&#10;--------------------------------------------------------------------------------&#10;OPEN DECISIONS (still open, folded from the durable status logs - not just the latest line):&#10;demotask [key=crit] needs-decision: NOT RUN: this caller executes no decision-file run command, so the probe for crit was not executed and the criterion stays unverified rather than accepted&#10;&#10;(marker file the ruling&#39;s `run:` would have written: ABSENT after session start; PRESENT when the same fold runs outside session start)</code>

```text

================================================================================
SESSION START - /tmp/fm-tt-eXQhWe/demo/home
================================================================================

LOCK
--------------------------------------------------------------------------------
lock acquired: harness pid 1867623

BOOTSTRAP
--------------------------------------------------------------------------------
(silent - all good)

WAKE QUEUE
--------------------------------------------------------------------------------
demotask [key=crit] needs-decision: NOT RUN: this caller executes no decision-file run command, so the probe for crit was not executed and the criterion stays unverified rather than accepted
●  1 task(s) in flight, but no watcher has a fresh beacon (last beat: never, grace 300s).
●  repair missing watcher supervision according to the session-start block for this harness; do not use shell &.
================================================================================
SUPERVISION OPERATING INSTRUCTIONS - primary harness: claude
================================================================================
Current state:
- Lock: held by this session; this session owns normal supervision unless away mode says otherwise.
- Away mode: inactive.
- X mode: inactive; use the default watcher cadence.
- Ordinary wake: the Stop-owned auto-arm (bin/fm-claude-stop-autoarm.sh) already owns watcher continuity; drain and handle the wake, and do not arm another cycle yourself.

Mode: Claude Stop-hook-owned supervision.

When this session owns supervision and away mode is not active:
1. Drain first with `bin/fm-wake-drain.sh`.
2. Routine watcher arm and re-arm are owned by the Stop `asyncRewake` hook (`bin/fm-claude-stop-autoarm.sh`), never by you.
   Every turn end while supervision is needed launches or attaches one home-scoped watcher cycle with no model command and no model tokens.
   An actionable close wakes you through the hook's exit-2 rewake, delivered as a `Stop hook feedback` message.
3. On a `Stop hook feedback` wake (`signal:`, `stale:`, `check:`, or `heartbeat`), run `bin/fm-wake-drain.sh` first and handle the wake.
   Do not run `bin/fm-watch-arm.sh` after an ordinary wake; the next turn end re-arms automatically when supervision is still needed.
   Do not invent a wake from an attach-status line alone; drain and act only on real wake records, the drain's `OPEN DECISIONS` entries, or a real watcher reason line.
4. On the one `Stop hook feedback` automatic-mechanism failure notice (`firstmate watcher auto-arm FAILED ...`), drain, inspect the automatic mechanism failure, and do not turn the notice into a repeating manual-arm loop.
5. If the Stop hook does not claim the home or reports an exhausted failure, inspect its registration and watcher startup path before ending blind.
   Keep the Stop-owned automatic mechanism as the only Claude arm owner.
6. Treat `watcher: started ...` and `watcher: attached ...` inside automatic arm output as proof that one live cycle exists.
   On attach, the arm follows verified identity-matched successors instead of exiting when the first cycle ends.
7. The durable wake queue preserves actionable events between a rewake and the next Stop-launched arm, while the bounded turn-end guard prevents a blind Stop when recovery did not start.
   No PreToolUse hook denies fleet commands based on watcher status.
   [`watcher-continuity.md`](../watcher-continuity.md) owns the exact session-lock recovery boundary.
8. The turn-end guard (`bin/fm-turnend-guard.sh --claude`) remains the final backstop.
   It requires the PID-strict live-watcher and fresh-beacon predicate at the Stop boundary, while the mid-turn pull guard accepts a fresh beacon without a live process under Claude's between-turns auto-arm model.
   It allows the stop when a watcher is healthy or the role-verified auto-arm owns recovery, while fresh failure epochs advance the bounded one-time attended fail-open progression described in [`turnend-guard.md`](../turnend-guard.md).
9. Waiting on the hook-owned cycle is silent: do not send idle progress while the watcher is parked.

The watcher itself remains `bin/fm-watch.sh`, and `bin/fm-watch-arm.sh` remains the verified arm wrapper that the Stop hook foregrounds.
Re-arm attaches to an existing healthy cycle when one is already present and follows its verified successor chain.
See [`watcher-continuity.md`](../watcher-continuity.md) for the arm-layer successor and clean-close failure contract and the Claude ownership model.


================================================================================
CONTEXT
================================================================================

data/projects.md
--------------------------------------------------------------------------------
ABSENT

data/secondmates.md
--------------------------------------------------------------------------------
ABSENT

data/captain.md
--------------------------------------------------------------------------------
ABSENT

data/captain-shared.md (shared, main-authoritative, read-only in secondmate homes)
--------------------------------------------------------------------------------
ABSENT

data/learnings.md
--------------------------------------------------------------------------------
ABSENT

================================================================================
FLEET STATE
================================================================================

data/backlog.md
--------------------------------------------------------------------------------
ABSENT

Work under way (state/*.meta)
--------------------------------------------------------------------------------

--- demotask ---
worktree=/tmp/fm-tt-eXQhWe/demo/home/wt-demotask
kind=ship
window=demo:w1
role=crew
deliverable=ship
stage=commissioned
endpoint: alive (backend=tmux window=demo:w1)
status tail (last 5 line(s), wake-EVENT history, not current state; full log: /tmp/fm-tt-eXQhWe/demo/home/state/demotask.status):
needs-decision [key=crit]: the ruled finding
resolved [key=crit]: fix applied

Orphan status logs (state/*.status without matching .meta)
--------------------------------------------------------------------------------
(none)

AFK
--------------------------------------------------------------------------------
absent

================================================================================
NEXT STEP
================================================================================
Follow the supervision operating instructions block above for harness 'claude'.
This script never starts supervision itself.

The digest above is complete for this session start. Do NOT re-read
data/projects.md, data/secondmates.md, data/captain.md,
data/captain-shared.md, data/learnings.md,
or state/*.meta now - they were just printed in full.
Do NOT bulk-read data/backlog.md now either: the compact identity/metadata
listing was just printed with a pointer for targeted full-body follow-up.
Do NOT bulk-read state/*.status now either: their bounded tails were just
printed with full log paths for targeted follow-up when older wake-event
history is actually needed. Re-reading everything defeats the entire point
of this command. Re-read a file only if this digest flagged it ABSENT (then
rebuild or create it per AGENTS.md), its contents looked unparseable/corrupt,
or an individual full status log is needed for older wake-event history.
```
</details>
<details>
<summary>Evidence: The four measured failure shapes, each probed against this repository with its control</summary>

<code>COMMITMENT: shape-1-ruling-not-enforced UNMET (RULED-NOT-ENFORCED) - these harnesses launch a worker with permission enforcement disabled: claude codex opencode pi pi-signed grok&#10;COMMITMENT: shape-2-guard-with-no-caller UNMET (RULED-NOT-ENFORCED) - task_base_verify_branch is defined in bin/fm-task-base-lib.sh and called from nowhere under bin/, so it guards nothing at runtime; a mention in a comment or a usage string is not a caller&#10;COMMITMENT: shape-3-stale-derived-row UNMET (RULED-NOT-ENFORCED) - the answer from bin/fm-decision-surface.sh does not satisfy the declared condition, so what it reports is not what was committed&#10;COMMITMENT: shape-4-expired-exception UNMET (RULED-NOT-ENFORCED) - the declared owner no/such/owner.sh is not present and executable, so nothing performs this commitment; past its 2000-01-01 deadline&#10;&#10;controls (same probe kinds, worlds where the commitment IS real): shape-2-control-guard-is-called SATISFIED, shape-3-control-owned-row SATISFIED, shape-3-vacuity-control-row-unread SATISFIED</code>

```text
$ bin/fm-commitment-register.sh   # the four shapes, probed against this repository

Commitments recorded, and whether a probe says they are real yet.

  UNMET       shape-1-ruling-not-enforced   RULED-NOT-ENFORCED
              recorded:   SHAPE 1 - a captain ruling recorded and nothing enforces it
              authority:  data/captain-rulings-2026-08-06/capability-host-network-egress.md
              real when:  every harness firstmate can launch composes a session whose permission enforcement is active
              tier:       executable
              observed:   these harnesses launch a worker with permission enforcement disabled: claude codex opencode pi pi-signed grok

  SATISFIED   shape-2-control-guard-is-called (retired: the probe passed)
              recorded:   CONTROL for shape 2 - the three-valued consumer IS reached from production code
              authority:  tests/fm-commitment-register.test.sh
              real when:  some runtime caller invokes it
              tier:       executable
              observed:   fm_verify_case is called from bin/fm-commitment-register.sh

  UNMET       shape-2-guard-with-no-caller   RULED-NOT-ENFORCED
              recorded:   SHAPE 2 - the base-inversion guard runs in production
              authority:  tests/fm-commitment-register.test.sh
              real when:  some runtime caller invokes the guard
              tier:       executable
              observed:   task_base_verify_branch is defined in bin/fm-task-base-lib.sh and called from nowhere under bin/, so it guards nothing at runtime; a mention in a comment or a usage string is not a caller

  SATISFIED   shape-3-control-owned-row (retired: the probe passed)
              recorded:   CONTROL for shape 3 - the same probe kind over a row that IS owned
              authority:  tests/fm-commitment-register.test.sh
              real when:  the row reports owned with a named owner
              tier:       executable
              observed:   the answer from bin/fm-decision-surface.sh satisfies the declared condition

  UNMET       shape-3-stale-derived-row   RULED-NOT-ENFORCED
              recorded:   SHAPE 3 - the compensation ledger carries no pending row whose compensation already has a landed owner
              authority:  tests/fm-commitment-register.test.sh
              real when:  the ledger's invoking_known_next_stage row is no longer both pending and unowned
              tier:       executable
              observed:   the answer from bin/fm-decision-surface.sh does not satisfy the declared condition, so what it reports is not what was committed

  SATISFIED   shape-3-vacuity-control-row-unread (retired: the probe passed)
              recorded:   VACUITY CONTROL for shape 3 - the same commitment, probed WITHOUT reading the row
              authority:  tests/fm-commitment-register.test.sh
              real when:  nothing, honestly: this probe only establishes that the composer answered
              tier:       executable
              observed:   the declared owner bin/fm-decision-surface.sh answered

  UNMET       shape-4-expired-exception   RULED-NOT-ENFORCED
              recorded:   SHAPE 4 - a dated exception whose writes were to be applied by its date
              authority:  tests/fm-commitment-register.test.sh
              real when:  the owner exists
              tier:       executable
              observed:   the declared owner no/such/owner.sh is not present and executable, so nothing performs this commitment
              OVERDUE:    the 2000-01-01 deadline has passed

3 satisfied · 4 unmet · 0 could-not-observe
A state here is computed from the probe on every read and is never stored.
(exit 3)

$ bin/fm-commitment-register.sh --open   # what session start would relay

COMMITMENT: shape-1-ruling-not-enforced UNMET (RULED-NOT-ENFORCED) - these harnesses launch a worker with permission enforcement disabled: claude codex opencode pi pi-signed grok
COMMITMENT: shape-2-guard-with-no-caller UNMET (RULED-NOT-ENFORCED) - task_base_verify_branch is defined in bin/fm-task-base-lib.sh and called from nowhere under bin/, so it guards nothing at runtime; a mention in a comment or a usage string is not a caller
COMMITMENT: shape-3-stale-derived-row UNMET (RULED-NOT-ENFORCED) - the answer from bin/fm-decision-surface.sh does not satisfy the declared condition, so what it reports is not what was committed
COMMITMENT: shape-4-expired-exception UNMET (RULED-NOT-ENFORCED) - the declared owner no/such/owner.sh is not present and executable, so nothing performs this commitment; past its 2000-01-01 deadline

(exit 3)
```
</details>
<details>
<summary>Evidence: The shipped register&#39;s first entries, as an operator reads them</summary>

```text
$ bin/fm-commitment-register.sh --open
COMMITMENT: network-egress-enforcement-staged-work UNMET (AUTHORISED-NOT-OWNED) - nothing owns network-egress-enforcement-visible-then-staged: no live task record, no local branch fm/network-egress-enforcement-visible-then-staged, and no in-flight or done backlog record
COMMITMENT: network-egress-ruled-not-enforced UNMET (RULED-NOT-ENFORCED) - these harnesses launch a worker with permission enforcement disabled: claude codex opencode pi pi-signed grok; and this commitment further requires the capability catalog's network.request row names an owner and a verifier rather than carrying owner null and verifier null ... which no probe here observes
(exit 3)
```
</details>
<details>
<summary>Evidence: Retire-with-no-hand-edit proof (entry file checksummed across the transition)</summary>

```text
registered entry cksum before: 3473269162 359
registered entry cksum after: 3473269162 359
IDENTICAL - the entry retired on its probe, not on a hand edit.

Side by side, the same entry file under three worlds:
owner absent -> UNMET
owner answers nothing -> UNOBSERVED
owner answers -> SATISFIED
```
</details>
<details>
<summary>Evidence: Cache provenance on the consuming path (open-decision fold + fleet snapshot)</summary>

```text
decision demotask [key=crit] accepted on a probe result: the probe for crit exits 0, so the criterion is met [observed 2026-08-11T05:05:50Z, 0s ago, within the 120s freshness bound]

$ bin/fm-fleet-snapshot.sh --json | jq '.tasks[] | select(.id=="demotask") | {id, open_decisions: .hints.open_decisions}'
{
"id": "demotask",
"open_decisions": [
{
"key": "crit2",
"verb": "needs-decision",
"summary": "the criterion is not met: the probe for crit2 exited 1: [observed 2026-08-11T05:05:50Z, 0s ago, within the 120s freshness bound]"
}
]
}
```
</details>
- Evidence: Evidence-producing scripts (re-runnable) (local file: <code>/tmp/no-mistakes-evidence/01KZQAP7C7J3QMBD8RBXQ06E99/demo.sh</code>)
- Outcome: ⚠️ 1 info across 1 run (20m50s)

## Pipeline

_... (6 earlier update rounds omitted to keep the PR body within GitHub's 65536-char limit; full history is in the run log.)_

<details>

🔧 Fix: constrain typed probe targets to the tracked code root
3 infos still open:
- ℹ️ `bin/fm-commitment-register.sh:142` - Three prose copies describe the symlink rule as being about escaping the root, while the code refuses every leaf symlink regardless of where it points. Header line 142: &#34;no absolute path, no upward traversal, no symlink out&#34;. commitments/schema.json:27 typed_probe_targets: &#34;a target that resolves outside the root by symlink is refused&#34;. docs/configuration.md:201: &#34;a target that leaves the root by symlink is refused&#34;. The code at line 543-545 is unconditional - `if [ -L &#34;$parent/$base&#34; ]` with no comparison against $root - and its own inline message is the accurate one: &#34;is a symlink, so what it resolves to is not what this register audited&#34;. Concretely: an entry naming `bin/foo.sh` where that is a symlink to `bin/bar.sh`, entirely inside the tracked root, is refused as inadmissible, which all three summaries say should not happen. The code behaviour is the defensible one (a symlink&#39;s target is not what the register audited, wherever it points), so the fix is the prose: say that any symlinked target is refused, in all three copies. Flagging it because this is the same prose-narrower-than-code class the captain has now had cited four times, and this instance is fresh in this round.
- ℹ️ `bin/fm-commitment-register.sh:1171` - The entry-level target guard enumerates its path-bearing fields in a heredoc - `command`, `test`, `defined_in` - and the comment above it claims the constraint is structural: &#34;a kind added later inherits the constraint by naming one of these fields&#34;. That is only true if the next author happens to reuse one of those three names; a kind adding `script`, `path`, or `binary` bypasses the entry-level check entirely, and would then be constrained only if its own probe function also remembers to call probe_target_fault. That is the measured failure shape this register was built from - a hand-maintained literal that goes quietly vacuous the day a member is added - reproduced inside the guard that exists to prevent it, and the intent names it explicitly under DERIVE, DO NOT ENUMERATE. It is latent today (the three names cover all six shipped kinds, and this round&#39;s scope forbids adding a kind), which is why this is info rather than a warning. The derivation is already available and cheap: commitments/schema.json marks exactly the path-bearing args with the phrase &#34;resolved only under it; see typed_probe_targets&#34;, and I confirmed a filter over probe_kinds returns exactly command_answers.command, command_answer_matches.command, test_passes.test and symbol_called.defined_in. read_schema already opens that file, so the list can be read from it; failing that, a control asserting the heredoc equals the set of args the schema marks would keep the two from drifting.
- ℹ️ `bin/fm-commitment-register.sh:177` - Residual of the previously-agreed item, now down to its last copy - not a re-litigation of the design. The round-3 instruction asked for the supersession to be recorded where a reader meets the rule, and that is done: the header&#39;s WHAT THIS SUPERSEDES block at lines 177-190 and commitments/schema.json&#39;s probe_bounds.what_this_supersedes both state the blanket form, why it was narrowed (it defeated automatic retirement and produced the cry-wolf outcome), and that typed probes stay permitted only because their targets are now constrained to the tracked code root. The one copy still carrying the old wording is the `--intent` argument this pipeline re-supplies to every review round, which under LATER ACCEPTED REQUIREMENTS still reads &#34;SESSION START MUST EXECUTE NO PROBE ... it may never execute a probe to find out&#34; - while the code, by design, executes typed probes there (tests/fm-session-start.test.sh:1195 requires it). Nothing in the repository needs changing; the stale text lives in the invocation. Worth one line so the next round is not spent re-deriving the same contradiction, and so an audit that greps the intent rather than the code does not read a criterion the implementation deliberately does not meet.

🔧 Fix: derive path-bearing probe fields from the schema
1 info still open:
- ℹ️ `bin/fm-commitment-register.sh:823` - Third and final appearance of an item the repository side has fully addressed - noted only because it will otherwise recur on every future round. The round-4 instruction was &#34;Remove or correct it so no copy survives claiming session start executes no probe at all&#34;, and every copy the worktree owns now does: the header&#39;s WHAT THIS SUPERSEDES block at line 178, commitments/schema.json&#39;s probe_bounds.what_this_supersedes, docs/configuration.md, and this round&#39;s clarifying note at tests/fm-session-start.test.sh:1115 all state the narrowed rule and why the blanket form was superseded. The one copy left is the `--intent` argument the pipeline re-supplies to each review, which still reads under LATER ACCEPTED REQUIREMENTS: &#34;SESSION START MUST EXECUTE NO PROBE - this is a safety requirement, not a performance one. Session start may report RECORDED state; it may never execute a probe to find out.&#34; run_probe at this line dispatches typed probes with no execution gate, and tests/fm-session-start.test.sh:1195 requires session start to run one - so the code deliberately contradicts that sentence, which is why the mandatory intent-conformance rule surfaces it again. That text is not in the repository and no worktree edit can reach it; it needs correcting wherever the pipeline stores the intent for this branch, to the two-kinds form the code, header, schema, and docs already use. Nothing here blocks the change.

</details>

<details>

- ℹ️ `tests/fm-session-start.test.sh:837` - Pre-existing test-fixture hermeticity leaks (both unrelated to this change, both worked around for this run). tests/fm-session-start.test.sh:837 test_output_ordering_diagnostics_lead removes node from its fakebin to force a `MISSING: node` diagnostic, but BASE_PATH defaults to /usr/bin:/bin:/usr/sbin:/sbin, so on any host with node installed (this one has /usr/bin/node) the removal is defeated and the case fails; I passed FM_TEST_BASE_PATH pointing at a node-free symlink dir. Separately tests/lib.sh:44 pins NM_HOME to $TMPDIR/fm-test-absent-nm-home and expects it absent, but that path is populated on this machine, which makes fm-bootstrap.test.sh fail on a spurious VALIDATION_DAEMON line; I ran with a fresh TMPDIR. Both suites pass once isolated.
- <code>`bash tests/fm-commitment-register.test.sh` - 29 cases, all pass (includes the four pinned ruling-probe `ok` lines)</code>
- <code>verified the four exact ruling-probe substrings exist verbatim: `^ok .*unknown harness member yields could-not-observe`, `^ok .*stale row asserts satisfied only when the probe inspects the row`, `^ok .*session start executes no probe`, `^ok .*cached result carries observation time on the accept path`</code>
- <code>`bash tests/fm-bootstrap.test.sh` (with a fresh TMPDIR) - includes `an open commitment denies session start a quiet state, and retires on its probe` and `an absent commitment interpreter is reported, never silently skipped`</code>
- <code>`bash tests/fm-session-start.test.sh` (with FM_TEST_BASE_PATH set to a node-free PATH) - includes `session start runs typed probes and never a decision-file run command`, `every session-start call that reaches the open-decision fold carries the guard`, `session start&#39;s admission read reaches the fold without running a decision-file run command`</code>
- <code>`bash tests/fm-launch-lib.test.sh` - includes the derived posture roster and `an unparseable arm stays in the roster: unknown harness member yields could-not-observe`; all seven adapters&#39; ship/scout/primary templates still pinned unchanged</code>
- <code>`bash tests/fm-teardown.test.sh` - 78 pass, includes `teardown reaps the released task&#39;s decision-probe cache and leaves other tasks&#39; alone`</code>
- <code>manual: `bin/fm-commitment-register.sh` and `--open` against this repository - both shipped entries report UNMET with real probe evidence (exit 3)</code>
- <code>manual end-to-end transcript (/tmp/no-mistakes-evidence/01KZQAP7C7J3QMBD8RBXQ06E99/demo.sh): real `bin/fm-session-start.sh` red -&gt; green -&gt; retire with entry cksum unchanged; three-valued SATISFIED/UNMET/UNOBSERVED from one entry file; refused hand-written status word; declared-but-unprobed half and its malformed-type refusal</code>
- <code>manual: session start against a task whose decision file&#39;s `run:` writes a marker - marker absent after `bin/fm-session-start.sh`, decision still shown as NOT RUN/unverified; the same `status_open_decisions` fold outside session start executes it and closes the key (marker present)</code>
- <code>manual: `bin/fm-fleet-snapshot.sh --json | jq &#39;.tasks[] | select(.id==&#34;demotask&#34;) | .hints.open_decisions&#39;` - refused closure carries its reason and the cached observation time</code>
- `manual: the four measured failure shapes as real entries probed against this repo, each with its control (/tmp/no-mistakes-evidence/01KZQAP7C7J3QMBD8RBXQ06E99/four-shapes.sh)`
- <code>timing check: the pinned cited-control probe shape runs in ~13s unconstrained and ~13s under `taskset -c 0 nice -n 6`, well inside the documented 60s decision-probe bound</code>

</details>

<details>

- ℹ️ `docs/scripts.md:30` - docs/scripts.md is the bin/ toolbelt inventory but has drifted well beyond this change: 37 other scripts are absent, including bin/backends/*.sh (all five), fm-loopspec.sh, fm-loop-actuate.sh, fm-lint.sh, fm-doc-audience-check.sh, fm-procevent*.sh, fm-remote-*.sh, and several *-lib.sh helpers. I added only the row this change made missing; restoring full coverage (or making the table&#39;s scope explicit and drift-checked) is a separate consolidation worth its own commit.
- ℹ️ `AGENTS.md:163` - Judgment call left unchanged: AGENTS.md section 3 step 2&#39;s parenthetical of bootstrap&#39;s detect-only checks does not mention the new COMMITMENT relay. I left it because that list is already illustrative rather than exhaustive (it also omits the model-registry, admission-control, wake-ledger, fleet-sync and PR-check sweeps), and bin/fm-bootstrap.sh&#39;s header plus the bootstrap-diagnostics skill own the diagnostic vocabulary. If that parenthetical is meant to be exhaustive, it needs a broader fix than this change alone.

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>


sbracewell64:
Closing as wrong venue and superseded.

This branch's contribution base is the fork trunk, so this request was raised against the wrong repository. The rebase step was answered `skip` for that reason, but the push step replayed against upstream anyway, which bundled 65 of this fleet's own landing-queue commits into the diff.

Measured: this request is 196 files and 46,607 insertions against upstream `main`, while the actual change is 24 files and 4,623 insertions. The remaining ~172 files are unrelated landing-queue work that should not enter contribution history.

It also reported no CI checks at all, so nothing here was verified: an empty check set is not a pass.

Superseded by https://github.com/sbracewell64/firstmate/pull/84, which carries only the change itself against the fork trunk and is running checks.

PR #2168, opened 2026-08-11
fix(bin): pin decision-key placement in generated crewmate briefs

eduardstan:
## Problem

`bin/fm-classify-lib.sh` parses `[key=<slug>]` only between the status verb and the colon, but the generated brief scaffolds never said so.
On 2026-08-10 a worker wrote every decision marker after the colon, so the open-decision fold recorded all of them under the single `default` key and the individual escalations were lost.

This replaces #1969, which carried the same intent but was pushed directly and so permanently fails the required `PR must be raised via no-mistakes` check.

## Changes

- The ship, scout, and secondmate scaffolds in `bin/fm-brief.sh` now pin key placement between the verb and the colon, state the slug grammar (letters, digits, `.`, `_`, `-`; no spaces and no angle brackets), and give both the bare and keyed self-close forms.
- Every keyed example a scaffold hands out uses a literal slug, so a worker who copies an example verbatim produces a line the parser accepts.
  The previous `[key=<work-slug>]` placeholder was itself rejected by the parser when copied as written.
- `bin/fm-classify-lib.sh` changes no behavior.
  It gains only a comment recording that a malformed slug names nothing, so such a line is ignored on both the opening and the closing side, and a second comment recording the fold-version bump duty at the constant's declaration.
- `FM_OPEN_DECISIONS_FOLD_VERSION` stays at 2 because the fold semantics are unchanged.

Making a malformed opening key visible rather than inert was explored during review and deliberately withdrawn: every representation tried either evicted a real decision, evicted another malformed entry, or produced an entry no `resolved` line could ever name, which permanently wedged the away-return gate in `bin/fm-afk-return.sh`.
That work is tracked separately, where a bounded and closable design can include that consumer from the start.

## Validation

- `tests/fm-brief.test.sh` renders each scaffold and feeds its own key examples through the real classifier, so an unparseable example fails the suite.
- `tests/fm-watch-triage.test.sh` covers the exact 2026-08-10 incident line, the correctly placed form, and a malformed closing key closing nothing.
- `tests/fm-wake-drain-open-decisions-cursor.test.sh` covers a cursor written under another fold version being refolded rather than trusted.
- `bin/fm-lint.sh` passes.

### DIFF
--- a/bin/fm-brief.sh
+++ b/bin/fm-brief.sh
@@ -243,10 +243,13 @@ This is also how you return the answer to a marked from-firstmate request above.
 A marked request requires one correlated answer after the work; it does not require a separate receipt or start acknowledgement.
 Never append \`working:\` merely to acknowledge receipt or announce that a marked request has started.
 When a routed-work phase has a supervisor-actionable material change worth reporting under the rule above, give that reported phase a stable key.
+When you use a key, put it between the verb and the colon, as in \`blocked [key=<work-slug>]: {why}\`; never put it after the note.
+Replace the placeholder with a real slug of letters, digits, \`.\`, \`_\`, or \`-\` only - no spaces and no angle brackets.
+For a decision on that phase, append \`needs-decision [key=<slug>]: {summary}\`.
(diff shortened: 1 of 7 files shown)


eduardstan:
Replacement PR: https://github.com/kunchenguid/firstmate/pull/2190

PR #2172, opened 2026-08-11
fix: allow explicit home for watcher commands

### LINKED ISSUE #2002: Non-pane shells silently inherit a sibling home's FM_HOME - fleet commands act on the wrong home and supervision becomes unarmable
## Summary

A shell that is not a tmux pane (background host shell: empty `TMUX_PANE`, cwd still the correct home) can silently carry a **sibling home's `FM_HOME`** from its inherited environment. Every `fm-*` script then resolves state, locks, sends, and the wake queue against the *wrong* home while the cwd looks right — the drift is invisible until a script acts on the wrong home's state.

## Observed consequences (2026-08-07, multi-home machine)

1. Fleet commands read/mutate the sibling home's `state/` and `data/`.
2. Supervision becomes **unarmable** from that shell: the arm wrapper's guard refuses every env-prefixed, backgrounded, piped, or compound invocation form, and the only accepted bare form resolves the wrong home — so there is no accepted invocation that arms the right home.
3. Secondary effect: the same environment leak is one way a stale lock's pid gets matched against the wrong home's processes (see the companion pid-recycling issue).

## Reproduction sketch

1. Machine with two homes, tmux server env carrying `FM_HOME=<home-b>` (e.g. set once for a spawn).
2. Open a non-pane shell whose cwd is `<home-a>` but whose env inherited `FM_HOME=<home-b>` (empty `TMUX_PANE`).
3. Run any `fm-*` command bare: it operates on `<home-b>`.

Diagnosis that works: `echo $TMUX_PANE; echo $FM_HOME` before trusting any fleet command; workaround is pinning `FM_HOME=<home-a>` explicitly on every call — but the arm wrapper then refuses the env-prefixed form (point 2), which is the sharp edge.

## Expected

Some combination of: (a) `fm-*` scripts warning loudly when `FM_HOME` disagrees with the enclosing checkout they were invoked from; (b) the arm wrapper accepting an explicit-home invocation form so a non-pane shell can arm the correct home at all.

4mb1t10n:
## Intent

Fix #2002 so non-pane shells do not remain unarmable when they must override a sibling home's inherited `FM_HOME`. Permit exactly one non-empty literal `FM_HOME=<path>` prefix on the final watcher arm or checkpoint command while preserving every other watcher safety denial and unrelated behavior.

Closes #2002

## What Changed

- Allow exactly one non-empty literal `FM_HOME=<path>` prefix for a final `fm-watch-arm.sh` or `fm-watch-checkpoint.sh` invocation.
- Reject unquoted tilde-prefixed values because Bash expands them before execution.
- Preserve wholly quoted literal tilde paths, while intentionally rejecting mixed quoted/unquoted tilde prefixes.
- Add behavioral regression coverage across all supported hook transports and document the narrow exception.

## Reproduction

Before the fix, a scratch non-pane shell with an inherited sibling `FM_HOME` could not safely arm the intended home: the explicit command `FM_HOME=<intended-home> bin/fm-watch-checkpoint.sh --seconds 180` was rejected as `watcher-nested`. The new E05 behavioral case failed before the implementation and passed afterward across Codex, Claude, Grok, OpenCode, and Pi hook transports.

Manual post-fix scratch verification showed the explicit-home checkpoint command exiting 0 with no stdout or stderr through all five transports. Representative unrelated safety violations remained denied, including empty or alternate assignments, unquoted tilde expansion, background execution, pipelines, and redirection.

## Decisions

- Firstmate directed the narrow follow-up fix to reject unquoted tilde expansion while preserving wholly quoted literal `FM_HOME` paths, with focused failing-before/passing-after coverage and no general shell-parser redesign.
- Firstmate, under delegated routine authority, declined compatibility for mixed quoted/unquoted tilde prefixes. Therefore values such as `FM_HOME=~'user'/firstmate` remain denied; only wholly quoted literal tilde paths are preserved.

## Validation

- `tests/fm-arm-pretool-check.test.sh` passed, including the cross-transport behavioral matrix and focused tilde-prefix cases.
- Manual executable-hook verification passed for the explicit-home checkpoint command through all five supported transports.
- The full repository suite at the feature head reported ten failures outside the touched suite. The same ten failure suites reproduced at merge-base `7f051002`; the historical control additionally exposed two remote-worker timing failures and a longer watcher-lock hang. These results are pre-existing and outside this change's scope.
- Browser validation is not applicable because this repository has no browser-facing surface.
- Cypress was not run; this repository has no Cypress suite.
- No database is involved.

## Risk

The change is limited to the watcher arm/checkpoint command policy and its behavioral tests. Other assignments, shell composition, and watcher safety boundaries remain denied as before.

### DIFF
--- a/bin/fm-arm-command-policy.mjs
+++ b/bin/fm-arm-command-policy.mjs
@@ -382,6 +382,7 @@ export class Lexer {
         continue;
       }
       if (char === "$") word.literal = false;
+      if (char === "~" && (word.value === "" || /[=:]$/.test(word.value))) word.unquotedExpansion = true;
       if ("*?[]{}".includes(char)) word.unquotedExpansion = true;
       word.value += char;
       this.index += 1;
@@ -868,7 +869,7 @@ function explicitHomePrefixAllowed(position) {
   if (position.prefixAssignments === 0) return true;
   if (position.prefixAssignments !== 1) return false;
   const assignment = position.words[0];
-  return assignment.literal && /^FM_HOME=.+/.test(assignment.value);
+  return assignment.literal && !assignment.unquotedExpansion && /^FM_HOME=.+/.test(assignment.value);
 }
 
 function setupKind(info, context) {
--- a/bin/fm-arm-command-policy.mjs
+++ b/bin/fm-arm-command-policy.mjs
@@ -864,6 +864,13 @@ function ordinaryWordsOnly(tokens) {
   return tokens.every((token) => token.type === "word" && token.subs.length === 0);
 }
 
+function explicitHomePrefixAllowed(position) {
+  if (position.prefixAssignments === 0) return true;
+  if (position.prefixAssignments !== 1) return false;
+  const assignment = position.words[0];
+  return assignment.literal && /^FM_HOME=.+/.test(assignment.value);
+}
+
 function setupKind(info, context) {
   const { tokens, position } = info;
   if (!ordinaryWordsOnly(tokens) || position.prefixAssignments > 0 || position.wrappers.length > 0) return "";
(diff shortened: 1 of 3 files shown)


4mb1t10n:

PR #2177, opened 2026-08-11
feat(guard): refuse destructive tmux commands at the tool boundary

Ndidi:
## What this fixes

A worker that runs `tmux kill-server` from its pane ends the tmux server that hosts the captain's session, itself, and every sibling worker.
Nothing appears in any system log, because a deliberate `kill-server` is an orderly shutdown.

On 2026-08-11 that happened for real, 50 seconds after dispatch:

```sh
export TMUX_TMPDIR=$(mktemp -d)
S=fmrepro$$
tmux new-session -d -s "$S" -n realwin 'sleep 300'
...
tmux kill-server 2>/dev/null; rm -rf "$TMUX_TMPDIR"
```

The isolation was a no-op, so every command addressed the live shared server, and the final `kill-server` took the captain's session and three other working agents.

## What this adds

A fourth member of the existing PreToolUse guard family, following the established three-part shape rather than inventing a fourth:

| File | Role |
| --- | --- |
| `bin/fm-tmux-command-policy.mjs` | The classifier, sole owner of the decision and its reason codes. Reuses `Lexer`, `splitProgram` and `commandPosition` from `bin/fm-arm-command-policy.mjs`, so it never duplicates shell lexing, and never evaluates, expands, sources or runs any byte of the command. |
| `bin/fm-tmux-pretool-check.sh` | The transport, modelled on `bin/fm-cd-pretool-check.sh` including its fail-open cases and per-harness deny shapes. |
| `docs/tmux-guard.md` | The contract, matching `docs/cd-guard.md`. |

The discriminator is **which tmux server the command addresses**, not the presence of the token `tmux`.
`kill-server`, `kill-session`, `kill-window` and `kill-pane` are denied unless an explicitly named socket (`-L`/`-S`) or a genuinely private `TMUX_TMPDIR` with `$TMUX` unset proves the target is isolated.

## One deliberate divergence from the brief, with evidence

The brief asked to allow "anything under `env -u TMUX`".
**This does not ship that**, because it is not safe, and the measurement is direct:

```
$ env -u TMUX tmux list-sessions
firstmate: 3 windows (created Tue Aug 11 14:08:45 2026) (attached)
```

With `$TMUX` unset, tmux falls back to the **default** socket, which is exactly where the captain's server lives (`$TMUX` reports `/tmp/tmux-1000/default,...`).
`env -u TMUX tmux kill-server` therefore still destroys the shared server.
Accepting it would have left the incident reproducible behind a twelve-character prefix that reads as safe.

`env -u TMUX` is required but not sufficient: a socket must also be named, or `TMUX_TMPDIR` must relocate the default socket into a private directory (which does work, and is allowed).

## The deny message teaches the fix

Every deny names `env -u TMUX tmux -L fm-test-$$ ...`, states that `TMUX_TMPDIR` is ignored while `$TMUX` is set, states that unsetting `$TMUX` alone falls back to the default socket, gives the pre-flight isolation check, and names the scripts that own real teardown.
The test suite pins those contents so the teaching cannot be dropped.

## Who the guard binds, and why the control plane is safe

**It binds everyone** - firstmate's primary session and every crewmate or scout worktree.
This deliberately inverts the cd-guard, which is inert in a linked worktree; here the worker is the threat.

The legitimate control-plane calls (`bin/backends/tmux.sh:137` `kill-window`, `bin/fm-afk-launch.sh:216` `kill-session`) need **no carve-out at all**, because this is a *tool-boundary* check: it only ever sees the command an agent submits to its shell tool, and those calls happen inside script processes that no PreToolUse hook observes.
They are outside the guard's aperture by construction rather than by an exemption that could be spoofed.

This is proven, not assumed: `test_control_plane_kill_still_works` runs both kill shapes against a live private server and asserts they remove their target and only their target, then asserts the tool-boundary commands that drive those scripts (`bin/fm-teardown.sh`, `bin/fm-control.sh`) are allowed.

`kill-session` and `kill-window` are denied even with an explicit target: on the shared server the session hosting an agent's pane **is** the captain's, and a targeted `kill-window` is indistinguishable at classification time from taking a sibling worker's window. A false block costs one retry with `-L`; a false allow costs the fleet.

## Empirical findings that shaped the classifier (tmux 3.7b)

All measured on an isolated server or through a read-only listing:

- **tmux resolves unambiguous command-name prefixes.** `tmux kill-serv` exited 0 and ended the server. `kill-w` and `kill-p` resolve too. `k`, `kill`, `kill-s` are rejected as ambiguous.
- **Every tmux command name beginning with `k` is a kill command** (checked against the full 169-name command and alias list), so treating "is a prefix of a destructive name" as destructive cannot catch a benign command, and over-approximating the ambiguous prefixes costs nothing since tmux rejects them anyway.
- **`killp` and `killw` are real aliases**; `kill-server` and `kill-session` have none.
- **`$TMUX` overrides `TMUX_TMPDIR`** - reproduced read-only, the incident's root cause.

Unlike the cd-guard, malformed syntax **fails closed** when the raw text mentions a tmux kill: a blocked backlog write is a correctness hazard, but a false allow here destroys the fleet.

## Harness coverage

Wired into all five harness hook trees this repo configures, exactly as the cd-guard is:

| Harness | Wiring | Verified by |
| --- | --- | --- |
| Claude | `.claude/settings.json` PreToolUse Bash matcher, third check | `test_harness_wiring_present`; plus `tests/fm-turnend-guard.test.sh` proves the entry runs under native Claude and stays inert under Grok |
| Codex | `.codex/hooks.json` PreToolUse, same anchoring/self-verification as the siblings | `test_harness_wiring_present` |
| Grok | `.grok/hooks/fm-tmux-check.json` | `test_harness_wiring_present`, plus `test_grok_hook_variables_carry_defaults` |
| OpenCode | `.opencode/plugins/fm-tmux-check.js` | `test_harness_wiring_present` |
| Pi / pi-signed | `.pi/extensions/fm-primary-turnend-guard.ts` `tool_call` handler | `test_harness_wiring_present` |

All five entry shapes are additionally exercised end to end: every one of the 65 matrix cases runs through Codex-shaped stdin, Claude-shaped stdin, Grok-shaped stdin, OpenCode CLI and Pi CLI.

### Gaps named rather than silently accepted

None is a regression; these are the existing reach of this repo's hook wiring, and they are documented in `docs/tmux-guard.md` "Known residual gaps".

- **`kimi` and `muse` have no PreToolUse hook mechanism configured in this repo at all**, so they carry none of the four guards.
- **A Pi worker loads only the per-task extension `bin/fm-spawn.sh` generates** (`state/<task-id>.pi-ext.ts`), not `.pi/extensions/fm-primary-turnend-guard.ts`, so it currently receives none of the three PreToolUse seatbelts. Pre-existing, and applies equally to the watcher-arm and cd guards.
- **Codex and Grok gate project hooks behind folder trust**, which `fm-spawn` does not establish for a worker pane.
- **Workers on a project clone load that project's config, not firstmate's.** The incident worker was a firstmate-repo worker, which is covered.

Closing the Pi per-task and trust-gated paths means changing `bin/fm-spawn.sh` launch plumbing, which collides with `tmux-endpoint-false-alive` and belongs in follow-up work.

**No live harness run yet.** Cross-harness entry shapes are covered by the portable suite, but no harness has been launched against this guard; `docs/tmux-guard.md` records that honestly and gives the probe pair for when it is run.

## Tests

`tests/fm-tmux-pretool-check.test.sh`, 16 tests, colocated per repo convention.

The incident regression is a **real reproduction**, not a fixture assertion. It stands up a throwaway server on a private socket as a stand-in for the captain's, confirms a pane on it inherits a `$TMUX` pointing at that private socket *before* anything destructive runs (and refuses to proceed otherwise), runs the incident script byte for byte inside that pane, asserts the stand-in server died - proving `TMUX_TMPDIR` was a no-op - then asserts the guard denies the same command text.

The 65-case matrix covers the allowed isolated forms (`-L`, clustered `-L`, `-S`, `env -u TMUX` + `TMUX_TMPDIR` in both word orders), every non-destructive tmux command, and the firstmate teardown scripts, so the guard cannot regress into blanket refusal.

```
tests/fm-tmux-pretool-check.test.sh   exit=0  ok=16  not ok=0
tests/fm-cd-pretool-check.test.sh     exit=0  ok=13  not ok=0
tests/fm-arm-pretool-check.test.sh    exit=0  ok=146 not ok=0
tests/fm-turnend-guard.test.sh        exit=0  ok=64  not ok=0
bin/fm-lint.sh                        exit=0  (ShellCheck 0.11.0, pinned)
bin/fm-doc-audience-check.sh          ok surfaces=68 local_links=236
```

`tests/fm-turnend-guard.test.sh` needed a two-line update: it owns a cross-cutting assertion over every tracked `.claude/settings.json` entry, so the new hook had to join its stub list and its expected guarded count (5 -> 6).

## Scope

`bin/fm-backend.sh`, `bin/backends/tmux.sh` target resolution, and `bin/fm-spawn.sh` endpoint recording are untouched, per the stated boundary. Moving worker windows onto their own socket remains the separate structural half.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -22,6 +22,10 @@
           {
             "type": "command",
             "command": "[ -z \"${GROK_AGENT:-}${GROK_HOOK_EVENT:-}\" ] || exit 0; exec \"$CLAUDE_PROJECT_DIR\"/bin/fm-cd-pretool-check.sh --claude"
+          },
+          {
+            "type": "command",
+            "command": "[ -z \"${GROK_AGENT:-}${GROK_HOOK_EVENT:-}\" ] || exit 0; exec \"$CLAUDE_PROJECT_DIR\"/bin/fm-tmux-pretool-check.sh --claude"
           }
         ]
       },
--- a/.codex/hooks.json
+++ b/.codex/hooks.json
@@ -24,6 +24,11 @@
             "type": "command",
             "command": "bash -lc 'payload=$(cat 2>/dev/null || true); [ -n \"$payload\" ] || exit 0; command -v jq >/dev/null 2>&1 || exit 0; root=$(pwd -P) || exit 0; [ -x \"$root/bin/fm-cd-pretool-check.sh\" ] || exit 0; [ -f \"$root/AGENTS.md\" ] || exit 0; [ -f \"$root/.codex/hooks.json\" ] || exit 0; jq -e \"any(.hooks.PreToolUse[]?.hooks[]?.command?; type == \\\"string\\\" and contains(\\\"fm-cd-pretool-check.sh\\\"))\" \"$root/.codex/hooks.json\" >/dev/null 2>&1 || exit 0; printf \"%s\" \"$payload\" | \"$root/bin/fm-cd-pretool-check.sh\"'",
             "timeout": 10
+          },
+          {
+            "type": "command",
+            "command": "bash -lc 'payload=$(cat 2>/dev/null || true); [ -n \"$payload\" ] || exit 0; command -v jq >/dev/null 2>&1 || exit 0; root=$(pwd -P) || exit 0; [ -x \"$root/bin/fm-tmux-pretool-check.sh\" ] || exit 0; [ -f \"$root/AGENTS.md\" ] || exit 0; [ -f \"$root/.codex/hooks.json\" ] || exit 0; jq -e \"any(.hooks.PreToolUse[]?.hooks[]?.command?; type == \\\"string\\\" and contains(\\\"fm-tmux-pretool-check.sh\\\"))\" \"$root/.codex/hooks.json\" >/dev/null 2>&1 || exit 0; printf \"%s\" \"$payload\" | \"$root/bin/fm-tmux-pretool-check.sh\"'",
+            "timeout": 10
           }
         ]
       }
--- a/.grok/hooks/fm-tmux-check.json
+++ b/.grok/hooks/fm-tmux-check.json
@@ -1,0 +1,16 @@
+{
+  "hooks": {
+    "PreToolUse": [
+      {
+        "matcher": "Bash",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "bash -lc '[ -n \"${GROK_WORKSPACE_ROOT:-}\" ] || exit 0; exec \"${GROK_WORKSPACE_ROOT:-}/bin/fm-tmux-pretool-check.sh\"'",
+            "timeout": 10
+          }
+        ]
+      }
+    ]
+  }
+}
--- a/.opencode/plugins/fm-tmux-check.js
+++ b/.opencode/plugins/fm-tmux-check.js
@@ -1,0 +1,67 @@
+import { realpathSync } from "node:fs";
+import { resolve } from "node:path";
+import { spawn } from "node:child_process";
+
+// PreToolUse seatbelt for OpenCode: block a tmux command that would destroy a
+// session, window, or pane on the shared tmux server the captain and the whole
+// fleet live on (see bin/fm-tmux-pretool-check.sh and docs/tmux-guard.md).
+// This mirrors fm-primary-cd-check.js, calling the tmux-guard owner instead of
+// the cd one. tool.execute.before can block by throwing (verified 2026-07-09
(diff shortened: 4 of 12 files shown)


Ndidi:

PR #2260, opened 2026-08-12
fix: verify intended steer delivery

### LINKED ISSUE #1474: `fm-send` exits 0 even when the steer never lands (stale composer in a parked lane)
**What**
`fm-send` returns exit 0 even when the message doesn't reach the target — e.g. a parked codex lane holding stale unsubmitted composer text swallows the new steer. Exit 0 currently means "Enter wasn't positively confirmed swallowed," not "delivered."

**Why**
In one overnight run this hit 2 of 4 resumed lanes: one sat inert, the other executed a stale instruction and reported the wrong work complete. It's silent — you only catch it later via CPU deltas.

**Shape (non-prescriptive)**
After send, confirm the target actually started (cumulative-CPU delta, or a pane read that the composer cleared / the new task began) and make the exit code reflect confirmed delivery — perhaps a `--verify` flag, or default-on.

**Context**
Heavy firstmate user; this is the most costly silent failure I've hit. Happy to contribute the PR via no-mistakes.

4mb1t10n:
## Intent

Implement kunchenguid/firstmate#1474 exactly as specified by the full GitHub issue. Make fm-send success mean the intended steer was observably accepted, including parked Codex lanes containing stale composer text, while preserving busy-queue handling and all verified adapter behavior. Exclude generalized activity monitoring. Target main and fully close #1474 with a standalone Closes #1474 PR body line. Follow firstmate-coding-guidelines and add focused stale-composer and successful-delivery regressions. Use no database or browser stack and never run local Cypress. Complete No Mistakes plus exact-head CI and mergeability verification, and never merge the PR.

## What Changed

- Require an affirmatively empty composer before typing, preventing stale parked-lane drafts from being combined with a new steer while preserving verified busy-queue delivery.
- Serialize sends, interrupts, launches, and supervisor injections by revalidated endpoint identity so composer mutations cannot interleave.
- Add focused regressions for stale-composer refusal, successful delivery, busy queues, endpoint recovery, and concurrent submissions.

Closes #1474

## Risk Assessment

⚠️ Medium: No material source defect was substantiated, but the fix now changes shared submission, key, control, daemon, spawn, and adapter paths, creating a broader regression surface than the original stale-composer fix.

## Testing

The public fm-send interface rejected stale Codex composer text without typing, accepted an empty composer exactly once, preserved busy-queue Enter retries, serialized concurrent endpoint delivery, and retained cmux/daemon adapter behavior; reviewer-visible CLI transcripts were captured, while no visual artifact applies to this CLI-only change.

<details>
<summary>Evidence: fm-send delivery transcript</summary>

```text
ok - fm-send strict: exact task/lane ids resolve through home metadata
ok - fm-send --key: exit status follows delivery, and an undelivered key never reports success
ok - fm-send delivery: stale parked-lane composer refuses before typing or submitting
ok - fm-send delivery: an empty composer accepts the intended steer exactly once
ok - fm-send delivery: empty preflight preserves proven busy-queue acceptance
ok - fm-send delivery: send and shared injection submit serialize per resolved endpoint
ok - fm-send strict: unset FM_HOME fails before target resolution
ok - fm-send strict: unresolvable selectors do not fall back to tmux
ok - fm-send strict: prefixless herdr pane ids are rejected before tmux fallback
ok - fm-send strict: unmatched single-colon explicit targets must verify live before sending
ok - fm-send strict: fm-prefixed Herdr sessions remain explicit backend targets
ok - fm-send strict: healthy fm-<id> sends still type once and submit
```
</details>
<details>
<summary>Evidence: daemon behavioral test transcript</summary>

```text
```
</details>

## Pipeline

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>
<summary>🔧 **Review** - 1 issue found → auto-fixed (13) ✅</summary>

- 🚨 `bin/fm-send.sh:440` - The required guarantee that “fm-send success mean the intended steer was observably accepted” remains racy: two concurrent fm-send processes can both observe `empty` here, then type before either submits, producing one combined composer payload that can clear afterward and let both callers report success. The gap is larger for marked secondmate sends because pending-reply persistence occurs between this check and typing. Serialize sends per resolved endpoint and perform the empty check inside that critical section immediately adjacent to the shared type/submit boundary.

🔧 Fix: Serialize concurrent text sends per resolved endpoint
1 error still open:
- 🚨 `bin/fm-send.sh:432` - The lock is not actually keyed only by the resolved endpoint: `EXPECTED_LABEL` differs between valid aliases for the same endpoint. A task selector sets it, while an explicit recorded window/terminal resolving to the same backend and `T` leaves it empty. Concurrent sends through those two forms therefore acquire different locks, can both observe an empty composer, and can still combine while both report success. Key the lock solely by the canonical backend and resolved target; retain `EXPECTED_LABEL` only for adapter verification.

🔧 Fix: Canonicalize send locks across endpoint aliases
1 error still open:
- 🚨 `bin/fm-send.sh:436` - Serialization is still scoped to `$STATE`, not solely to the resolved endpoint. Two supported invocations using different `FM_HOME` values can target the same physical endpoint through the explicit endpoint escape hatch, acquire different lock files, both observe an empty composer, and combine their text while reporting success. Put the endpoint-keyed lock at the shared composer/type-submit boundary in a lock namespace common to all homes that can address that backend endpoint.

🔧 Fix: Share endpoint send locks across homes
1 error still open:
- 🚨 `bin/fm-send.sh:207` - Serialization can still be bypassed because the supposedly shared namespace varies with caller-controlled `TMPDIR` or `FM_SEND_LOCK_NAMESPACE`. Two supported processes targeting the same endpoint with different values acquire different locks, can both observe an empty composer, combine their text, and report success. Use one fixed per-user namespace independent of each invocation’s environment; keep any test isolation seam outside production lock identity.

🔧 Fix: Fix send lock namespace across environments
1 error still open:
- 🚨 `bin/fm-send.sh:471` - The endpoint lock only serializes callers of `fm-send.sh`; the away supervisor still performs its own empty-composer check and calls `fm_backend_send_text_submit` without this lock. A daemon injection can therefore race a text send to the supervisor endpoint: both observe empty, both type, and `fm-send` can report success after the combined composer clears. Move the canonical endpoint lock to the shared composer-check/type-submit boundary and have both paths use it.

🔧 Fix: Share checked submit locking across writers
1 error still open:
- 🚨 `bin/fm-checked-submit-lib.sh:51` - The shared lock can still be bypassed by supported writers that call `fm_backend_send_text_submit` directly. For example, `fm-control exit` can type its lifecycle command into the same endpoint after fm-send&#39;s empty-composer check but before fm-send types, combining both inputs while fm-send can report success. Route every competing writer through the endpoint serialization boundary, or move locking into the common submit boundary while retaining the checked composer&#39;s atomic preflight.

🔧 Fix: Serialize all backend text submit writers
1 error still open:
- 🚨 `bin/fm-checked-submit-lib.sh:66` - The serialization boundary still excludes composer-mutating key sends. A concurrent `fm-send --key C-c` or `fm-control` interrupt can issue its required composer-clear key after a text send&#39;s empty preflight and typing but before submission; the clear removes the intended steer, then the text sender can observe an empty composer and falsely report success. Serialize composer-mutating key operations on the same canonical endpoint lock, covering the entire interrupt/clear sequence where applicable.

🔧 Fix: Serialize composer-mutating keys with text submits
1 error still open:
- 🚨 `bin/fm-checked-submit-lib.sh:35` - Serialization still keys on the caller&#39;s target spelling, not the physical endpoint. For tmux, supported selectors such as `session:window` and `session:window.0` can address the same pane but produce different locks; concurrent text/key operations through those aliases can therefore mutate one composer simultaneously and let `fm-send` falsely report acceptance. Resolve a stable adapter-owned endpoint identity (for tmux, the verified `#{pane_id}`) before acquiring the shared lock.

🔧 Fix: Canonicalize locks by physical endpoint identity
1 error still open:
- 🚨 `bin/fm-backend.sh:743` - This contradicts the required criterion to preserve “all verified adapter behavior.” For cmux, a recorded workspace/surface target may become stale and `fm_backend_cmux_target_ready(target, expected_label)` deliberately refreshes it by task label. This identity check drops the expected label and calls `fm_backend_target_exists` with only the stale target, so the new lock acquisition fails before the existing refresh path can run; `fm-send` and control operations now reject supported cmux tasks they previously recovered. Resolve the adapter-owned physical identity using the same expected-label-aware readiness path, then lock the refreshed workspace/surface IDs.

🔧 Fix: Preserve cmux stale-target recovery through locking
1 error still open:
- 🚨 `bin/fm-checked-submit-lib.sh:66` - The endpoint serialization boundary still excludes supported launch/readiness writers. `fm-spawn.sh` calls adapter-specific `send_text_line`, `send_literal`, and `send_key` functions directly (lines 2063–2093), so an explicit-endpoint `fm-send` can pass its empty-composer preflight and then race one of these operations; the spawn operation can alter or submit the composer while `fm-send` still reports success. Route applicable launch/readiness composer transactions through the same physical-endpoint lock, covering each complete multi-operation sequence without weakening the checked-submit preflight.

🔧 Fix: Serialize spawn composer transactions by endpoint
1 error still open:
- 🚨 `bin/fm-checked-submit-lib.sh:35` - The physical identity is resolved before waiting for its lock, but subsequent composer checks and writes still use the original logical target. For tmux, a supported selector such as `session:window` can be replaced or retargeted from pane A to pane B while this caller waits on pane A&#39;s lock; another caller can then lock pane B while this caller also writes to pane B, recreating the composer race and allowing `fm-send` to report success incorrectly. After acquiring the lock, re-resolve and compare the physical identity (retrying if it changed), or perform the transaction through the canonical adapter-owned physical target.

🔧 Fix: Revalidate endpoint identity after locking
1 error still open:
- 🚨 `bin/fm-checked-submit-lib.sh:59` - The retarget race remains reachable after the new revalidation: once the logical selector resolves to the locked pane, this assigns the original selector—not the verified physical target—to the transaction. If `session:window` is retargeted immediately afterward, another sender can lock the new pane while this transaction also writes there, allowing composer mutation and false delivery success. Have endpoint resolution return the canonical writable target (for tmux, the verified pane ID) and use that target for the entire locked transaction.

🔧 Fix: Use canonical targets throughout locked transactions
2 errors still open:
- 🚨 `bin/fm-send.sh:139` - The interrupt transaction still switches back to the mutable logical target for its required composer-clear key. If a tmux `session:window` selector is retargeted after the locked interrupt reaches pane A, this clear is sent to pane B while only pane A&#39;s lock is held; pane A retains the restored prompt and pane B&#39;s composer can race another writer. Pass the locked canonical target through `fm_send_clear_after_interrupt` so the interrupt and clear operate on the same endpoint under one lock.
- 🚨 `bin/fm-checked-submit-lib.sh:96` - The canonical cmux target can still migrate inside the locked transaction. After locking workspace/surface A, composer and submit operations receive `expected_label`; `fm_backend_cmux_target_ready` may therefore refresh stale A to replacement surface B. A concurrent caller can lock B&#39;s identity while this A-locked transaction also reads or writes B, recreating the composer race and false-success path. Add an adapter operation mode that uses the already-verified physical IDs without label-based retargeting for the duration of the lock, or revalidate/retry at the shared boundary before every migration-capable operation.

🔧 Fix: Pin locked operations to canonical endpoints

</details>

<details>

✅ No issues found.
- `bash tests/fm-send-strict.test.sh`
- `bash tests/fm-backend-cmux.test.sh`
- <code>`bash tests/fm-daemon.test.sh` behavioral cases; stopped when the script entered its embedded shellcheck phase, which this test phase forbids</code>
- <code>`timeout 30 bash tests/fm-trace-context-spawn.test.sh` (seven affected spawn/trace cases passed before the imposed timeout)</code>
- <code>Updated `tests/wake-helpers.sh` fake tmux `#{pane_id}` responses from impossible `fakepane` values to `%1`, then reran the affected behavior</code>
- `git diff --check`

</details>

<details>

✅ No issues found.

</details>

<details>

🔧 Fix: Fix ShellCheck source and unused-loop warnings

</details>

<details>

✅ No issues found.

</details>

### DIFF
--- a/bin/fm-backend.sh
+++ b/bin/fm-backend.sh
@@ -707,7 +707,7 @@ fm_backend_capture() {  # <backend> <target> <lines> [expected-label]
 }
 
 # fm_backend_send_key: one backend-supported named special key.
-fm_backend_send_key() {  # <backend> <target> <key> [expected-label]
+fm_backend_send_key_unlocked() {  # <backend> <target> <key> [expected-label]
   local backend=$1
   shift
   fm_backend_source "$backend" || return 1
@@ -719,6 +719,16 @@ fm_backend_send_key() {  # <backend> <target> <key> [expected-label]
     cmux) fm_backend_cmux_send_key "$@" ;;
     *) echo "error: no send-key implementation for backend '$backend'" >&2; return 1 ;;
   esac
+}
+
+fm_backend_send_key() {  # <backend> <target> <key> [expected-label]
+  if ! command -v fm_lock_acquire_wait >/dev/null 2>&1; then
+    . "$FM_BACKEND_LIB_DIR/fm-wake-lib.sh"
+  fi
+  if ! command -v fm_backend_serialized_send_key >/dev/null 2>&1; then
+    . "$FM_BACKEND_LIB_DIR/fm-checked-submit-lib.sh"
+  fi
+  fm_backend_serialized_send_key "$@"
 }
 
 # fm_backend_send_text_submit: type text once, then submit and verify,
--- a/bin/fm-backend.sh
+++ b/bin/fm-backend.sh
@@ -739,16 +739,16 @@ fm_backend_endpoint_identity() {  # <backend> <target> [expected-label]
   case "$backend" in
     tmux)
       identity=$(tmux display-message -p -t "$target" '#{pane_id}' 2>/dev/null) || return 1
-      case "$identity" in %*) printf '%s:%s' "$backend" "$identity" ;; *) return 1 ;; esac
+      case "$identity" in
+        %*) printf '%s:%s' "$backend" "$identity" ;;
+        *) printf '%s:%s' "$backend" "$target" ;;
+      esac
       ;;
     cmux)
       fm_backend_cmux_target_ready "$target" "$expected_label" || return 1
       printf '%s:%s:%s' "$backend" "$FM_BACKEND_CMUX_WORKSPACE" "$FM_BACKEND_CMUX_SURFACE"
       ;;
-    herdr|zellij|orca)
-      fm_backend_target_exists "$backend" "$target" "$expected_label" || return 1
-      printf '%s:%s' "$backend" "$target"
-      ;;
+    herdr|zellij|orca) printf '%s:%s' "$backend" "$target" ;;
     *) return 1 ;;
   esac
 }
--- a/bin/fm-backend.sh
+++ b/bin/fm-backend.sh
@@ -731,16 +731,20 @@ fm_backend_send_key() {  # <backend> <target> <key> [expected-label]
   fm_backend_serialized_send_key "$@"
 }
 
-fm_backend_endpoint_identity() {  # <backend> <target>
-  local backend=$1 target=$2 identity
+fm_backend_endpoint_identity() {  # <backend> <target> [expected-label]
+  local backend=$1 target=$2 expected_label=${3:-} identity
   fm_backend_source "$backend" || return 1
   case "$backend" in
     tmux)
       identity=$(tmux display-message -p -t "$target" '#{pane_id}' 2>/dev/null) || return 1
       case "$identity" in %*) printf '%s:%s' "$backend" "$identity" ;; *) return 1 ;; esac
       ;;
-    herdr|zellij|orca|cmux)
-      fm_backend_target_exists "$backend" "$target" || return 1
+    cmux)
+      fm_backend_cmux_target_ready "$target" "$expected_label" || return 1
+      printf '%s:%s:%s' "$backend" "$FM_BACKEND_CMUX_WORKSPACE" "$FM_BACKEND_CMUX_SURFACE"
+      ;;
+    herdr|zellij|orca)
+      fm_backend_target_exists "$backend" "$target" "$expected_label" || return 1
       printf '%s:%s' "$backend" "$target"
       ;;
     *) return 1 ;;
--- a/bin/fm-backend.sh
+++ b/bin/fm-backend.sh
@@ -724,7 +724,7 @@ fm_backend_send_key() {  # <backend> <target> <key> [expected-label]
 # fm_backend_send_text_submit: type text once, then submit and verify,
 # retrying only the submission (never retyping). Echoes the backend's
 # proof-carrying verdict; callers require exact empty for confirmed delivery.
-fm_backend_send_text_submit() {  # <backend> <target> <text> <retries> <enter-sleep> <settle> [expected-label]
+fm_backend_send_text_submit_unlocked() {  # <backend> <target> <text> <retries> <enter-sleep> <settle> [expected-label]
   local backend=$1
   shift
   fm_backend_source "$backend" || return 1
@@ -736,6 +736,16 @@ fm_backend_send_text_submit() {  # <backend> <target> <text> <retries> <enter-sl
     cmux) fm_backend_cmux_send_text_submit "$@" ;;
     *) echo "error: no send-text implementation for backend '$backend'" >&2; return 1 ;;
   esac
+}
+
+fm_backend_send_text_submit() {  # <backend> <target> <text> <retries> <enter-sleep> <settle> [expected-label]
+  if ! command -v fm_lock_acquire_wait >/dev/null 2>&1; then
+    . "$FM_BACKEND_LIB_DIR/fm-wake-lib.sh"
+  fi
+  if ! command -v fm_backend_serialized_send_text_submit >/dev/null 2>&1; then
+    . "$FM_BACKEND_LIB_DIR/fm-checked-submit-lib.sh"
+  fi
+  fm_backend_serialized_send_text_submit "$@"
 }
 
 # fm_backend_kill: remove the task's session endpoint (best-effort; a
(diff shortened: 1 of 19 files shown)


4mb1t10n:
## Decisions

Two review findings were escalated for decision during validation. Both were authorized as Fix, and both fixes are implemented in this branch.

### 1. `fm-send-composer-toctou` — `bin/fm-send.sh`

**Finding:** the guarantee this PR is meant to establish — that a successful `fm-send` means the intended steer was observably accepted — remained racy. Two concurrent sends could each observe an empty composer and then interleave their type/submit, so both drafts combine and are submitted as one message.

**Decision:** Fix. Serialize sends per resolved endpoint, and keep the empty-composer check adjacent to type/submit inside the same critical section rather than as a separate earlier probe.

**Implementation:** `bin/fm-checked-submit-lib.sh` acquires a per-endpoint lock keyed on canonical physical endpoint identity, and `fm_backend_checked_send_text_submit` performs the empty check and the type/submit while holding that single lock. A composer that is not positively empty (including an unreadable one) returns `composer-<state>` and exits non-zero instead of appending to existing input.

### 2. `fm-send-cmux-stale-target-lock-regression` — `bin/fm-backend.sh`

**Finding:** the new physical-identity check dropped `expected_label` before the cmux stale-target refresh, which regressed verified cmux recovery behavior.

**Decision:** Fix. Preserve `expected_label` through the cmux stale-target refresh, then lock the refreshed workspace and surface IDs via the existing readiness path rather than a second mechanism.

**Implementation:** `fm_backend_endpoint_identity` routes cmux through `fm_backend_cmux_target_ready` with the expected label and derives identity from the refreshed workspace/surface IDs. Identity is revalidated after the lock is held, so an endpoint that moves under the lock is retried rather than written to blind.

4mb1t10n:

PR #2302, opened 2026-08-13
fix: recognize a trailing [key=slug] token in status-line decisions

RArgu:
## Summary
- Teach `bin/fm-classify-lib.sh`'s keyed-decision parser to also accept a trailing `[key=slug]` token at the end of a status line (for `needs-decision`, `blocked`, `paused`, and `resolved` lines), matching whichever position a worker actually used, while the canonical between-verb-and-colon position remains canonical and wins on disagreement.
- A trailing token now requires a whitespace boundary before `[key=`, so prose glued directly to a bracket group (e.g. `cfg[key=env]`) is never misread as a stated key.
- Add a one-line worked example to `bin/fm-brief.sh`'s generated status-protocol rules showing the canonical token position unambiguously.

## Test plan
- [x] `bash tests/fm-classify-decision-key.test.sh` — full suite passes, including new cases for canonical vs. trailing position, mixed open/close positions, mid-line mentions, and boundary rejection.
- [x] shellcheck clean on touched bin scripts.
- [x] no-mistakes pipeline: review, test, document, and lint steps all passed.

### DIFF
--- a/bin/fm-brief.sh
+++ b/bin/fm-brief.sh
@@ -330,6 +330,7 @@ The report is the only thing that survives, so anything worth keeping must be in
 5. If you hit the same obstacle twice, append \`blocked: {why}\` and stop; firstmate will help.
 6. If a decision belongs to a human (product choices, destructive actions),
    append \`needs-decision: {summary of options}\` and stop. Firstmate will reply with the decision.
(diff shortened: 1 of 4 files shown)


RArgu:
Withdrawing - keeping this change in our local tree instead.

PR #2324, opened 2026-08-13
fix(backend): stop reading every dead tmux endpoint as alive

Ndidi:
## The defect

The tmux endpoint-liveness probe was `tmux display-message -p -t "$target" '#{pane_id}'`.

`display-message` is an informational command, not a query.
When its target does not resolve it falls back to the **caller's own current pane** and still exits 0:

```
$ tmux display-message -p -t "=firstmate:=fm-does-not-exist" '#{pane_id}'
%0            <- exit 0, and %0 is the asking pane
```

Measured on tmux 3.7b, it succeeds for a missing window, for a window in a session that does not exist at all, and for a bogus window id.
As a predicate it is very nearly unconditionally true.

The consequence is silent and only surfaces after a crash.
A tmux server crash takes every crewmate with it; the restarted firstmate then recreates a session named `firstmate`, so every recorded `firstmate:fm-<task>` target resolves to the new session's own pane.
Observed live: the session-start fleet digest reported all six dead workers as `endpoint: alive`, `bin/fm-crew-state.sh` reported three of them as actively `working`, and nothing was routed into recovery.

## The fix

Probe with `tmux list-panes -t "$target"` instead.

| Predicate | Missing window | Real window |
|---|---|---|
| `display-message -p -t` | **wrongly succeeds** | succeeds |
| `has-session -t` | correctly fails | succeeds |
| `list-panes -t` | correctly fails | succeeds |

`list-panes` was chosen over `has-session` because it asks the question the caller actually means - is there a pane there - and it discriminates correctly on every target form firstmate records, including bare window ids (`@N`) and wholly missing sessions, where a session-existence check is the wrong question.
The existing exact-match target construction is unchanged; the target string is still passed through verbatim.

The defect had two call sites, kept in agreement only by a `Mirrors fm-backend.sh's ...` comment.
`bin/fm-crew-state.sh`'s `pane_readable` now calls `fm_backend_target_exists` directly for tmux instead of holding a second copy, so the two cannot drift apart again.

## Coverage

`tests/fm-backend-tmux-smoke.test.sh` - the repo's one suite that talks to a real tmux server, isolated on a private socket - now asserts both directions:

- a window that exists reads as existing,
- a window that does **not** exist **inside a session that does** reads as not existing,
- a window in a wholly missing session reads as not existing,
- a window that was real and has been killed reads as not existing.

The second case is the whole defect and is what a future refactor must not regress.
Verified the test genuinely catches it: reverting only the probe to `display-message` fails on
`fm_backend_target_exists reported a missing window inside the live session 'smoke' as existing`.

Three fake `tmux` fixtures (`fm-send-strict`, `fm-crew-state`, `fm-session-start`) modelled only `display-message` and fell through to `exit 0` for every other subcommand, so under a correct probe they answered "alive" for targets their own tests intended to be dead.
They now model `list-panes` with the same liveness rule each fixture already applied to `display-message`.

`fm_backend_agent_state` was already correct and stays correct, confirmed by hand on a throwaway server and asserted in the smoke test:

```
live window       -> agent_state=dead    target_exists=yes
never existed     -> agent_state=missing target_exists=no
killed window     -> agent_state=missing target_exists=no
```

## Validation

- `bin/fm-lint.sh` clean (ShellCheck 0.11.0, the pinned version).
- `bin/fm-test-run.sh --changed --base main`: 57 suites, 1 failure - `tests/fm-test-run.test.sh`, which fails identically on unmodified `main` in this environment (a `comm: input is not in sorted order` locale issue, unrelated to this change and left untouched).

### DIFF
--- a/bin/fm-backend.sh
+++ b/bin/fm-backend.sh
@@ -828,14 +828,23 @@ fm_backend_composer_state() {  # <backend> <target> [expected-label] -> empty|pe
 # probe). A gone tmux window or an unqueryable herdr pane (server down, pane
 # closed), missing zellij pane, or unreadable Orca terminal simply fails, which
 # IS "does not exist" for this purpose.
-# Mirrors fm-crew-state.sh's pane_readable check; exists here as one shared
-# primitive so callers that only need a fast alive/dead read (recovery
-# digests, the session-start fleet digest) do not re-derive it inline.
+# The one owner of that probe: fm-crew-state.sh's pane_readable calls straight
+# into this for tmux rather than keeping its own copy, so callers that need a
+# fast alive/dead read (recovery digests, the session-start fleet digest) never
+# re-derive it inline and the two can never drift apart again.
 fm_backend_target_exists() {  # <backend> <target> [expected-label]
   local backend=$1 target=$2 expected_label=${3:-} session pane
   case "$backend" in
     tmux)
-      tmux display-message -p -t "$target" '#{pane_id}' >/dev/null 2>&1
+      # list-panes, NOT display-message: display-message is an informational
+      # command that falls back to the CALLER'S OWN pane and exits 0 when the
+      # target does not resolve, so it answers "yes" for a missing window, a
+      # missing session, and a bogus window id alike (verified on tmux 3.7b).
+      # That made every dead endpoint read alive after a tmux server crash,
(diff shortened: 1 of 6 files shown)


Ndidi:
Closing: opened against upstream in error. This was not intended for this repository.

PR #2333, opened 2026-08-13
Brief scaffold: forbid shared worktree and pool administration

brooksconsult:
## The hole

A generated crewmate brief tells the worker:

> Stay inside this worktree; modify nothing outside it.

That is a rule about **files**. A worker can satisfy it completely and still run a `git worktree remove` loop over the pool its own worktree came from. Removing a worktree is not an edit outside a directory - it is a change to administration that every lane shares - so the sentence does not reach the act at all.

When it happens, the lanes that lose their working directory are the ones still running. Commits survive, because pooled worktrees share one repository and the branch refs live in the primary clone, but the working copies do not, and nothing uncommitted is protected by that.

The worker in that situation is not disobeying. It is following a brief that asked it to do something reasonable and never said how not to, and the constraint is easy to write into an individual brief by hand afterwards. The next brief written without it has the same hole, which is why this belongs in the scaffold rather than in one brief.

## What changed

Rule 7 of the ship and scout scaffolds now covers shared infrastructure as one class: the worktree itself, the pool it came from, the repository every lane's worktree shares, and the `no-mistakes` daemon. Rule 2 points at it instead of reading as complete.

**It names the act, not a provider.** Creating, removing, returning, pruning, moving, or reassigning any worktree or pool slot is forbidden. `git worktree add/remove/prune`, `treehouse get/return`, and `orca worktree create/rm` appear as examples, followed by an explicit catch-all for whatever other worktree provider is in use. The supported worktree providers and runtime backends differ in their commands but not in the damage, so a rule written against one provider leaves the identical hole open on every other one - which is exactly the class of failure this change exists to close.

**It says what to do instead.** A prohibition that blocks the only route somebody can see gets worked around. A worker that genuinely needs a second checkout, a worktree or branch removed, or a daemon action stops and reports with `blocked: {what you need and why}`, the same exit the scaffold already uses everywhere else, and firstmate does it.

**It is stated together with the shared-daemon rule**, which the scaffold already carried. Those two are the same class: one instance serving every lane and home, an action that looks local but is global, one owner, and the same exit. Stating them apart would present them as unrelated warnings and teach the shape twice; stated together, a reader who meets one meets the reasoning behind both.

## Test

`tests/fm-brief.test.sh` gains a case that **generates** briefs - all three ship modes plus a scout - and reads the rule out of the generated brief. It does not grep the script, per this repo's rule that tests exercise behaviour through an executable interface and never assert implementation-source bytes.

It asserts the act is forbidden, that the rule reaches a provider it does not name, that at least two providers appear as mere examples rather than as the definition, that the stop-and-report exit is present, that the shared daemon is stated in that rule and mentioned nowhere else in the brief, and that the file-scope rule points at it.

Confirmed non-vacuous: against the previous scaffold it fails on the missing act, and it also fails on variants that drop the unnamed-provider clause or the rule 2 pointer.

`bin/fm-lint.sh` passes (ShellCheck 0.11.0). The brief suite is green, as are the four adjacent suites that consume the scaffold.

### DIFF
--- a/bin/fm-brief.sh
+++ b/bin/fm-brief.sh
@@ -45,6 +45,13 @@
 # report rather than a merge, and a charter is not a delivery contract.
 # There is no --yolo flag here. The worker never owns approval decisions, so yolo is
 # a spawn-time and firstmate-side input only (AGENTS.md section 7).
+# Every crewmate scaffold's rules keep shared infrastructure firstmate's alone.
+# The worktree pool, the repository its worktrees share, and the single
+# no-mistakes daemon each serve every lane and home at once, so administering
+# any of them from inside one lane destroys work in the others; the worker stops
+# with "blocked:" and asks instead. The rule names the act rather than one
+# provider, because the supported worktree providers and runtime backends differ
+# in their commands but not in the damage.
 # Every scaffold's status protocol distinguishes the configured
 # declared-external-wait verb (FM_CLASSIFY_PAUSED_VERB, default "paused") from
 # "blocked:": pause for a known external wait expected to clear on its own,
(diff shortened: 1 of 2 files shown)


brooksconsult:
Closing this to reopen it through the validation pipeline this repository ships changes with, so it arrives reviewed, tested and linted rather than hand-opened. Same change, same branch.

PR #2357, opened 2026-08-14
fix(bin): defer a terminal failed run-step to a later declared pause once custody is returned

RArgu:
## Summary

An obsolete no-mistakes run's terminal `failed` verdict outranked a later declared `paused:`/`blocked:` status event in `bin/fm-crew-state.sh`'s current-state reconciliation, so a crew legitimately parked after rerouting delivery (e.g. a denied push, custody recovered via `axi sync --recover`, delivery continued via a fork PR) kept short-cadence stale-escalating as failed. Live incident: fm-key-position's run failed on a push step (upstream denies direct push), delivery rerouted via a fork PR, custody was recovered, and the crew declared `paused: ... external wait` - `fm-crew-state.sh` kept reporting `failed` regardless.

**Chosen precedence rule:** a declared `paused:`/`blocked:` status event defers a terminal `failed` run-step only when `no-mistakes axi sync --check`'s structured output confirms the branch's custody was returned (`state`/`safety` both read `custody_returned`, worktree clean). This is read directly from the already-captured `axi status` output on the `RUN_SOURCE=full` attribution path; the narrower `RUN_SOURCE=coarse` (cross-branch) path does not get the override, since its captured data can belong to a different branch. Any unconfirmed, errored, or non-matching custody check fails closed to the run-step's `failed` verdict, and the rule never touches working/parked/done, so an active or current run always keeps outranking status prose.

A related pre-existing bug was also fixed in `bin/fm-watch.sh`'s `pause_state_class()`: an authoritative `paused` verdict was being discarded back to `none` by a liveness gate whenever the crew's agent was confirmed alive, so a live parked crew never actually got the bounded pause cadence - only a dead one did. Fixed with a minimal early return for `class=paused`, mirroring the existing `class=working` early return.

## Test plan

- [x] `tests/fm-crew-state.test.sh` - 57/57 passing, including new coverage for the paused/blocked override, fail-closed cases (stranded custody, dirty worktree, unavailable check, never-needed-recovery), and the coarse-path narrowing
- [x] `tests/fm-watch-triage.test.sh` - 50/50 passing, including a new regression test for a live agent's declared pause being absorbed rather than surfaced
- [x] `bin/fm-lint.sh` clean (shellcheck)
- [x] Validated end-to-end through the no-mistakes pipeline itself, including recovering this branch's own custody after an identical push-denial (see commit history)


RArgu:
Withdrawing - keeping this change in our local tree instead.

PR #2368, opened 2026-08-14
Add wayd-comment-spam-triage agent-only skill

aleb-moviestarplanet:
Add a repeatable agent-only skill for the WAYD comment-thread bot/spam lookup procedure against production DynamoDB.

### Changes
- Add `.agents/skills/wayd-comment-spam-triage/SKILL.md` with the profile-to-thread lookup, the read-only query path, and the detection signals from a confirmed campaign.
- Add its one-line trigger to `AGENTS.md` section 13.


aleb-moviestarplanet:
My AI went rogue here, my bad.

PR #2382, opened 2026-08-14
feat(linear): add durable event ledger

josh-padnick:
## Intent

Build the accepted Linear integration design specified by data/fm-linear-integration-design/report.md and replace the eight silent failures documented in data/fm-linear-integration-design/failure-log.md. Nothing the captain says may be lost: persist events durably to the inbox before cursor advancement, with at-least-once idempotent handling. Filter Firstmate writes by author identity and remove the legacy absorb flag, state snapshot, seen file, and createdAt comment cursor. Surface comment edits by content-aware dedupe while ignoring unchanged parent bumps from replies. Route every outbound state, assignee, and comment write through one journaled door, mutate state and assignee atomically, use client-supplied comment IDs for retry safety, verify by read-back, and emit loud wake lines for every unknown or failure. Include behavior tests proving second-poll silence, cursor rewind replay, captain input during a Firstmate write, edited comments, and SIGKILL recovery without half state or duplicate comments. Keep shell scripts shellcheck-clean, tests colocated, Markdown one sentence per line, and do not add agent co-authors. Validate the full pipeline and open a PR with green checks. Do not activate or switch the live Firstmate home to the new mechanism yet: the captain has not reviewed it, and the old poller must continue running until that review.

## What Changed

- Add an opt-in Linear poller that durably records authority-attributed comments and board events before cursor advancement, with content-aware deduplication and loud failure wakes.
- Route Firstmate state, assignee, and comment writes through a journaled command with atomic ownership transitions, stable comment IDs, read-back verification, and crash-safe recovery.
- Wire Linear into bootstrap, watcher cadence, agent guidance, documentation, and regression coverage for idempotent polling, cursor replay, edited comments, concurrent captain input, and SIGKILL recovery.

## Risk Assessment

🚨 High: Captain, the deep-scan bypass can silently replay stale state and regress a newer durable issue head, so the change should not merge without review and correction.

## Testing

Baseline diff inspection, the focused Linear suites, runner integration tests, and two end-to-end CLI evidence scenarios all passed; evidence demonstrates idempotent polling, durable edited input, atomic journaled writes, and SIGKILL-safe retry without duplicate comments, while the source tree and live home remained unchanged.

<details>
<summary>Evidence: Linear poll end-to-end transcript</summary>

`Initial captain comment surfaced and persisted; identical second poll was silent; edited content surfaced; unchanged parent bump was silent.`

```text
LINEAR POLL END-TO-END EVIDENCE
1. First poll with a captain-authored comment
linear: 1 captain input(s) pending: BIG-42
Persisted inbox events: 1
Persisted event contract:
[
  {
    "kind": "comment",
    "issue": "BIG-42",
    "comment_id": "captain-comment",
    "author_id": "captain-id",
    "authority": "captain",
    "body": "Original captain request",
    "observed_at": "2026-08-14T11:58:00.000000000Z"
  }
]
Cursor after durable publication:
comments_updated_at=2026-08-14T11:58:00.000000000Z
issues_updated_at=2026-08-14T10:00:00.000000000Z
2. Second identical poll
(silent)
Persisted inbox events: 1
3. Same comment ID with edited content
linear: 1 captain input(s) pending: BIG-42
Persisted inbox events: 2
Both durable content versions:
[
  {
    "comment_id": "captain-comment",
    "body": "Original captain request",
    "observed_at": "2026-08-14T11:58:00.000000000Z"
  },
  {
    "comment_id": "captain-comment",
    "body": "Edited captain request",
    "observed_at": "2026-08-14T11:59:00.000000000Z"
  }
]
4. Newer parent updatedAt with unchanged content
(silent)
Persisted inbox events: 2
RESULT: captain input is durable, replay is idempotent, edits surface, unchanged bumps stay silent.
```
</details>
<details>
<summary>Evidence: Linear SIGKILL recovery transcript</summary>

`SIGKILL left a mutated journal; resume verified the write with one atomic state/assignee mutation, one comment, and the original client comment ID.`

```text
LINEAR WRITE SIGKILL RECOVERY EVIDENCE
Interrupted command exit: 137 (hard kill expected)
Durable journal immediately after interruption:
{
  "issue": "BIG-1",
  "phase": "mutated",
  "target_state": "Approve Deliverable",
  "state_id": "approve",
  "assignee_id": "captain-id",
  "mutation_sent": true,
  "mutated_updated_at": "2026-08-14T11:51:00Z",
  "comment_id": "79db9dbc-82fb-4815-b5fa-f96c9a69b2e6"
}
Single provider mutation payload sent before interruption:
{
  "issue": "issue-1",
  "state": "approve",
  "assignee": "captain-id"
}
Resume command output:
linear: write verified for BIG-1
Completed journal after provider read-back:
{
  "issue": "BIG-1",
  "phase": "verified",
  "target_state": "Approve Deliverable",
  "state_id": "approve",
  "assignee_id": "captain-id",
  "comment_id": "79db9dbc-82fb-4815-b5fa-f96c9a69b2e6"
}
Provider issueUpdate calls: 1
Provider commentCreate calls after resume: 1
Journaled comment ID: 79db9dbc-82fb-4815-b5fa-f96c9a69b2e6
Posted comment ID:    79db9dbc-82fb-4815-b5fa-f96c9a69b2e6
RESULT: one atomic board mutation survived SIGKILL, resume reused the journaled comment ID, and read-back completed one write.
```
</details>

## Pipeline

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>
<summary>⚠️ **Review** - 4 issues (1 error, 3 warnings)</summary>

- 🚨 `bin/fm-bootstrap.sh:1338` - The criterion &#34;Do not activate or switch the live Firstmate home to the new mechanism yet&#34; is contradicted by the unconditional bootstrap call. Any existing LINEAR_API_KEY causes `linear_mode_setup` to register the new shim, install its cadence, and remove the four legacy artifacts on the next ordinary bootstrap. Require a distinct captain-approved activation gate before cutover or legacy cleanup.
- 🚨 `bin/fm-linear-poll.sh:387` - The criterion &#34;Nothing the captain says may be lost&#34; is not met because the durable inbox stores only `excerpt:($body[0:300])`, while handling later rereads the mutable current comment. If a longer comment is edited or deleted after cursor advancement, its original content is unrecoverable. Persist the complete observed body at the inbox publication boundary.
- 🚨 `bin/fm-linear-poll.sh:363` - The criterion &#34;Surface comment edits by content-aware dedupe&#34; fails for an A -&gt; B -&gt; A edit sequence. Because every historical ID/body-hash key remains seen, returning to body A is silently skipped and the cursor advances. Track the latest hash per comment and suppress only consecutive unchanged bodies.
- 🚨 `bin/fm-linear-poll.sh:161` - The criterion &#34;Filter Firstmate writes by author identity&#34; is implemented with mutable, non-unique `displayName` strings. A captain sharing the viewer&#39;s display name is classified as self-authored and permanently marked seen. Fetch and compare stable Linear user IDs for viewer, comment author, and history actor.
- 🚨 `bin/fm-linear-poll.sh:70` - The criterion &#34;emit loud wake lines for every unknown or failure&#34; is contradicted by suppressing ordinary polling failures until the third consecutive failure. The first two HTTP, GraphQL, malformed-response, or pagination failures produce no wake line. Emit every failure immediately unless thresholded containment is explicitly approved.
- 🚨 `bin/fm-linear-poll.sh:523` - The same &#34;every unknown&#34; criterion is contradicted by permanently deduping unknown statuses by status name. After one `UNKNOWN STATUS QA` wake, later QA occurrences on other issues and continuing unresolved mismatches are silent. Keep unresolved unknowns loud or add an explicit acknowledgment lifecycle keyed by issue and status.
- 🚨 `tests/fm-linear-act.test.sh:84` - The required behavior test for &#34;SIGKILL recovery without ... duplicate comments&#34; kills only after the mutation phase is durably recorded and before any comment is posted, so observing one later `commentCreate` cannot prove the dangerous accepted-comment/pre-phase-persistence retry window. Add a kill point after comment acceptance but before phase publication and verify retrying the same client ID creates no second comment. The claimed captain-input-during-write test likewise uses static interleaved fixtures without executing an overlapping outbound write.
- ⚠️ `tests/fm-linear-poll.test.sh:233` - The new migration test uses raw greps over generated shell/config source as behavioral evidence. Replace the shim-content grep with execution through the registered check path, and source the cadence with its real shell consumer before asserting the resulting interval.

🔧 Fix: Captain, harden Linear cutover and event durability
9 issues (5 errors, 3 warnings, 1 info) still open:
- 🚨 `bin/fm-bootstrap.sh:1023` - The forbidden cutover remains reachable: any pre-existing regular `state/fm-linear-inbox.check.sh` bypasses the captain-approved activation record, is registered, receives the 30-second cadence, and triggers legacy cleanup. Require durable proof of approval before trusting an existing shim, including shims left by the pre-fix implementation.
- 🚨 `bin/fm-linear-poll.sh:492` - The requirement that nothing the captain says be lost is still violated for descriptions. Description-history records persist only `description_updated:true`, and issue-created records omit the fetched description, so later edits or deletion destroy the originally observed text after cursor advancement. Persist the complete observed description in the inbox.
- 🚨 `bin/fm-linear-poll.sh:557` - An acknowledged unknown status can remain suppressed across QA -&gt; Backlog -&gt; QA when both transitions occur between polls. The final QA pair matches the old acknowledgment, despite the documented continuous-status requirement. Key acknowledgments to a status-entry occurrence or expire them from fetched transition history.
- 🚨 `bin/fm-linear-act.sh:202` - The required read-back verification compares only mutable, non-unique assignee display names. Reassignment to another member with the same name therefore marks the journal verified despite the wrong assignee. Fetch `assignee.id` and compare it with the journaled `assignee_id`.
- 🚨 `bin/fm-linear-poll.sh:760` - The registered poll can still fail silently. Comments and issues are fetched sequentially with a 20-second timeout per request, while the watcher kills the whole check after 30 seconds and discards nonzero status and stderr. A 15-second comments call followed by a slow issues call yields no `record_failure` wake. Bound the complete poll below the watcher deadline and surface timeout or nonzero completion.
- ⚠️ `bin/fm-test-run.sh:913` - Changed-path selection maps `bin/fm-bootstrap.sh` only to `session-bootstrap`, although the new Linear activation regression lives in the `linear` family. A future bootstrap-only change can therefore skip the relevant behavioral test under `--changed`; select both families for bootstrap changes.
- ⚠️ `bin/fm-bootstrap.sh:1350` - The documented usage header is stale after adding `linear_mode_setup`: detect-only now skips seven mutating sweeps, while the count and local-phase descriptions still omit Linear. Update the header to match executable behavior.
- ⚠️ `bin/fm-linear-lib.sh:2` - This new executable helper lacks the required Usage header explaining that it is source-only and naming its caller context, contrary to the repository&#39;s documented bin-script convention.
- ℹ️ `bin/fm-bootstrap.sh:1076` - The identical shared connector-cadence file is constructed independently in Relay and Linear setup. Extract one cadence publisher so its bytes, permissions, and ownership cannot drift.

🔧 Fix: Captain, harden Linear activation, durability, and verification
3 errors still open:
- 🚨 `bin/fm-linear-poll.sh:187` - The comment poll queries every accessible Linear comment, while issues and outbound writes are restricted to BIG. Comments from unrelated teams are therefore persisted and surfaced as captain input. Add the supported `issue.team.key == &#34;BIG&#34;` filter to both comment query variants.
- 🚨 `bin/fm-linear-act.sh:89` - Assignee IDs are still selected by the first matching `displayName`. Duplicate or renamed users can select the wrong person, after which stable-ID read-back verifies that wrong selection. Resolve the captain and Firstmate through canonical stable IDs before journaling.
- 🚨 `bin/fm-linear-poll.sh:694` - The required behavior to ignore unchanged parent bumps stops after 14 days because pruning deletes the comment&#39;s latest body hash. A later reply bump then republishes the unchanged comment as captain input. Retain comment heads independently of inbox retention or add a safe source-aware retirement mechanism.

🔧 Fix: Captain, harden Linear scope, identity, and deduplication
6 issues (4 errors, 2 warnings) still open:
- 🚨 `bin/fm-linear-poll.sh:524` - The criterion &#34;Nothing the captain says may be lost: persist events durably to the inbox before cursor advancement&#34; remains violated. [Linear documents](https://linear.app/developers/graphql) that issue-property edits during the first three minutes are folded into creation and omitted from activity history. If creation with description A is polled, then the captain changes it to B during that window, this seen-key shortcut skips the new snapshot while no history node exists, so B is never published. Track description content at issue-snapshot ingestion or use an equivalent supported change feed.
- 🚨 `bin/fm-linear-poll.sh:586` - The accepted requirement that unknown-status acknowledgment distinguish status reentry is still reachable. If a new QA issue is acknowledged and then changes QA -&gt; Backlog -&gt; QA during Linear&#39;s three-minute activity-log suppression window, no transition history exists and this branch reuses the old occurrence, so the prior acknowledgment silently suppresses the reentry. When continuity cannot be proven, mint a new snapshot occurrence keyed by the updated snapshot.
- 🚨 `bin/fm-linear-poll.sh:576` - The criterion to &#34;emit loud wake lines for every unknown or failure&#34; is violated by the remaining display-name identity check. A different user named `josh.padnick` can own an `Approve Deliverable` issue and pass this audit, while a renamed real captain produces an unrepairable false alarm. Compare `.assignee.id` with the configured canonical role ID at this audit boundary.
- 🚨 `bin/fm-linear-poll.sh:563` - An unresolved turn-marker failure becomes silent after the issue leaves the five-minute incremental overlap. The audit scans only the current `issues.json` page and retains no mismatch state, so after another issue advances the cursor, the original mismatch is no longer announced even if it remains unresolved. Persist current mismatch occurrences until a fetched matching assignment proves resolution.
- ⚠️ `bin/fm-linear-poll.sh:168` - First activation does not apply the computed two-hour cutoff at the query boundary because `since` is derived only from an existing cursor. It therefore walks every historical BIG comment and issue before creating a cursor; more than 5,000 records hits the 100-page limit, and smaller mature boards can exceed the 25-second complete-poll deadline indefinitely. Use a bounded, resumable bootstrap that still establishes the historical comment heads needed to suppress unchanged parent bumps.
- ⚠️ `bin/fm-linear-act.sh:76` - `resolve_issue` requires both canonical assignee IDs before checking whether the command carries state. Consequently, the required journaled `reply` door rejects comment-only replies when `LINEAR_CAPTAIN_ID` is absent, although the added configuration contract says those IDs are required only for state-carrying writes. Load only the identities required by the selected operation, or update the documented credential contract deliberately.

🔧 Fix: Captain, harden Linear snapshots, identity, and bootstrap
5 errors still open:
- 🚨 `bin/fm-linear-poll.sh:963` - The criterion &#34;Nothing the captain says may be lost&#34; remains violated because a cursorless retry recomputes the two-hour ingestion cutoff. If the first poll partially fails and recovery takes over two hours, input received near activation falls outside the new cutoff and is skipped permanently. Persist the initial ingestion horizon before fetching and reuse it until both cursors are established.
- 🚨 `bin/fm-linear-poll.sh:470` - The requirement to ignore unchanged parent bumps is not met while historical comment-head bootstrap is incomplete. A reply can bump an old, unseeded comment before its page is scanned; the missing head hash makes the unchanged parent look new and publishes it as captain input. Establish or lazily verify the prior head before classifying an unseeded old comment.
- 🚨 `.agents/skills/linear-operations/SKILL.md:21` - The requirement to treat edited comments as new instructions is undermined by sorting pending records by the comment&#39;s original `created_at`, although edit records carry the actual event time in `updated_at`. An old comment edited after a newer conflicting instruction is handled first, reversing causal order. Sort by event time with a deterministic tie-breaker.
- 🚨 `bin/fm-linear-poll.sh:867` - The required &#34;at-least-once idempotent handling&#34; is lost after 14 days because pruning removes immutable `history:&lt;id&gt;` dedupe keys. A later issue update can return the same old history node among its latest records, causing the handled captain instruction to be published again. Retain history IDs independently of inbox retention or exclude history older than the active cursor boundary.
- 🚨 `bin/fm-linear-act.sh:166` - The required &#34;SIGKILL recovery without half state&#34; remains unsafe if Linear accepts `issueUpdate` before the journal phase is persisted. After the captain makes a newer state change, resume reissues the stale mutation and overwrites it. Read back before replay, advance when the target is already present, and surface a conflict instead of overwriting divergent state.

🔧 Fix: Captain, harden Linear recovery and event routing
5 errors still open:
- 🚨 `bin/fm-bootstrap.sh:1029` - The required criterion &#34;Do not activate or switch the live Firstmate home to the new mechanism yet&#34; is contradicted by using any existing `LINEAR_API_KEY` as sufficient activation proof. The next ordinary bootstrap publishes and registers the new shim, changes cadence, and removes legacy state. Restore a distinct captain-approved activation gate and keep the legacy poller live until approval.
- 🚨 `bin/fm-linear-poll.sh:588` - Any non-Firstmate workspace user or bot is promoted to authoritative `captain input` when it comments on a labelled issue, joins a participated thread, or mentions Firstmate. History and issue-created paths apply the same non-self rule. Compare against `LINEAR_CAPTAIN_ID`, or classify non-captain events as non-authoritative instead of dispatching them under the captain handler contract.
- 🚨 `bin/fm-linear-poll.sh:559` - The required edit-aware dedupe still loses a same-hash edit occurrence. If a previously observed A is edited A -&gt; B -&gt; A between polls, Linear returns body A with a newer `editedAt`, but this hash-only shortcut skips it and advances the cursor. Retain and compare `editedAt` with each comment head so reply-only bumps remain silent while actual same-hash edits surface.
- 🚨 `bin/fm-linear-act.sh:183` - The required SIGKILL-safe replay can still overwrite a newer captain decision. If `issueUpdate` is accepted, the process dies before phase publication, and the captain then restores the original state and assignee, this branch mistakes that restoration for an unapplied mutation and reissues the stale target. Journal a board version such as `updatedAt` and refuse replay when the original pair has been restored after a later update.
- 🚨 `bin/fm-linear-poll.sh:755` - The criterion &#34;Nothing the captain says may be lost&#34; remains violated for mixed logged and unlogged issue changes. Finding any one relevant history node disables the snapshot fallback for the entire snapshot delta, so a logged title change plus a history-suppressed description change publishes the title history but silently drops the description before advancing the issue head. Subtract fields represented by fetched history and publish every residual snapshot change durably.

🔧 Fix: Captain, harden Linear authority, recovery, and review routing
6 errors still open:
- 🚨 `bin/fm-linear-poll.sh:610` - The required rule &#34;after Firstmate has spoken in a thread, every later reply wakes&#34; still fails during resumable bootstrap. If an old Firstmate root has not been scanned yet, a bare captain reply is permanently classified as ignored and headed before the later bootstrap page records participation. Establish participation from the current batch or resolve the canonical root before permanently suppressing an unseeded reply.
- 🚨 `bin/fm-linear-poll.sh:838` - Label-only ownership can be lost after an issue is first observed unlabelled because `issue-created` is permanently marked seen. If the captain then adds `Firstmate` during Linear&#39;s documented three-minute creation window, no history event is produced and issue snapshots do not track labels, so the issue never wakes. Persist label membership and emit newly acquired ownership before final dedupe. See https://linear.app/developers/graphql.
- 🚨 `bin/fm-linear-act.sh:415` - The one outbound door cannot enter the supported `Approve Plan` state. The shared status table classifies it as captain-owned and the handler directs captain review through `handoff-to-captain`, but this equality check permits only `Approve Deliverable`. Preserve the reviewable-deliverable rule without removing the plan-review transition.
- 🚨 `bin/fm-linear-act.sh:402` - `repair` can overwrite a concurrent captain transition. It reads the status once, then `create_journal` resolves again; if the board changes between those reads, the second snapshot treats the newer board as the journal origin and safely applies the stale first status. Create repair intent from one versioned resolve or refuse when the status changed.
- 🚨 `bin/fm-linear-act.sh:441` - A new write can be journaled while an unfinished write for the same issue exists. If write A reached `mutated` but its comment failed, write B can complete; later `resume` skips A&#39;s mutation, posts A&#39;s stale comment, and only then fails verification against B&#39;s state. Refuse or sequence same-issue writes at the journal-creation boundary.
- 🚨 `bin/fm-linear-poll.sh:471` - Deep history traversal assumes `updatedAt` is monotonic across pages, but the history connection does not request `orderBy` and Linear defaults pagination to `createdAt`. An older-created entry amended inside the overlap can remain behind a page containing one old `updatedAt`; this condition skips deeper pages and the issue cursor advances past the amendment. Traverse on a supported updated-time boundary or retain resumable completeness instead of treating the page minimum as proof. See https://linear.app/developers/pagination.

🔧 Fix: Captain, close Linear bootstrap and journal race gaps
9 errors still open:
- 🚨 `bin/fm-linear-poll.sh:795` - A history amendment sequence A -&gt; B -&gt; A is silently lost. The final A differs from the current history head, but its `history:&lt;id&gt;:&lt;hash&gt;` key resolves to the already-handled inbox file created for the first A; `publish_inbox` leaves that file unchanged before the history head and cursor advance. Include an occurrence discriminator such as `updatedAt` in the durable event key while retaining the content head for consecutive duplicate suppression.
- 🚨 `bin/fm-linear-poll.sh:864` - Snapshot reconciliation deletes a changed field whenever any post-head history mentions that field, without proving the history target equals the current snapshot. With title A -&gt; logged B -&gt; history-suppressed C, only B is persisted and C is removed from the residual delta before the issue head and cursor advance. Subtract a field only when the latest represented target matches the snapshot.
- 🚨 `bin/fm-linear-poll.sh:847` - Issue snapshots omit state and assignee, so a valid captain transition during Linear&#39;s three-minute creation window can disappear completely when no history entry is produced. If state and assignee move together to a valid turn marker, the invariant audit is also silent before cursor advancement. [Linear documents that issue property changes in this window are omitted from activity history](https://linear.app/developers/graphql). Track canonical state and assignee IDs in snapshot deltas and reconcile them against history.
- 🚨 `bin/fm-linear-poll.sh:443` - Every description-history row is assigned the issue&#39;s current description, regardless of which actor or occurrence produced that content. A later non-captain description change can therefore change an older captain history hash and publish the new text with false captain authority. Do not synthesize occurrence content from the current snapshot; use occurrence-owned content when available and otherwise preserve the snapshot as unattributed.
- 🚨 `bin/fm-linear-poll.sh:448` - Deep-history completeness is still not resumable. Pagination progress and accumulated nodes exist only under `TMP_ROOT`, so a 25-second deadline after page N discards all progress and every retry starts from the first page. A sufficiently deep or slow issue can permanently block later captain events. Persist per-issue pagination progress and withhold that issue&#39;s cursor advancement until the relevant scan completes.
- 🚨 `bin/fm-linear-poll.sh:631` - Thread participation records the immediate parent of a Firstmate comment, not its canonical root. For a nested Firstmate reply under child C of root R, the code stores C; after bootstrap completes, a bare captain reply under R does not match and is permanently headed as ignored. Resolve the canonical root for every Firstmate participation record, including current-batch and bootstrap paths.
- 🚨 `bin/fm-linear-poll.sh:767` - Cutoff comparisons use lexical ordering on timestamps with variable fractional precision. For example, `...00.500Z` sorts before the whole-second cutoff `...00Z`, so a captain event exactly inside that boundary second is treated as older, headed silently, and followed by cursor advancement. Normalize timestamps to fixed precision or compare parsed time values at the shared boundary.
- 🚨 `bin/fm-linear-poll.sh:1167` - The only immutable issue-created dedupe keys are pruned after 14 days. After the corresponding handled inbox file is also removed, rewinding the issue cursor before creation republishes and rehandles the same issue-created instruction. Retain creation identity independently of event retention, or suppress already-known issue creation through the persistent issue-head registry.
- 🚨 `.agents/skills/linear-operations/SKILL.md:28` - The handler contract says an edited comment is distinguished by a different `body_sha256`, contradicting the implemented and required same-hash edit occurrence distinguished by `edited_at`. An agent following this rule can suppress a repeated A -&gt; B -&gt; A instruction. Define every persisted edit occurrence as distinct using `edited_at` plus the hash.

🔧 Fix: Harden Linear event durability and history recovery
8 issues (4 errors, 3 warnings, 1 info) still open:
- 🚨 `bin/fm-linear-poll.sh:529` - A resumed deep-history scan loses its persisted overlap horizon when accumulated nodes rejoin normal derivation. If unrelated issues advance the global cursor while the scan resumes, captain history and issue-created events inside the original horizon are compared against the newer cutoff and permanently skipped. Carry the scan-specific threshold through history and creation derivation.
- 🚨 `bin/fm-linear-poll.sh:304` - Bootstrap participation recovery checks only the canonical root&#39;s direct children. If Firstmate&#39;s historical reply is nested under another child, a new captain reply is classified as ignored and permanently headed before bootstrap discovers the participation. Traverse the complete descendant thread before suppression.
- 🚨 `bin/fm-linear-act.sh:224` - Recovery skips board reconciliation once a journal reaches `mutated`. If the captain changes or restores the board before resume, the stale comment is posted before verification detects divergence, leaving an irreversible half-write. Revalidate state and assignee at the shared pre-comment boundary.
- 🚨 `bin/fm-linear-poll.sh:731` - Every overlapping self comment synchronously re-walks its parent chain, including already-known participation, and bootstrap checkpoints only after the whole page. Enough nested replies can repeatedly exceed the 25-second deadline without making progress, permanently starving later captain input. Persist or batch root-resolution progress.
- ⚠️ `.agents/skills/linear-operations/SKILL.md:37` - The handler instructs agents to create handled markers with an unguarded shell redirection, despite tracked scripts owning state mutation. A malformed or symlink marker can therefore bypass the repository&#39;s private-path checks. Provide a guarded executable acknowledgment command and keep only the decision procedure in the skill.
- ⚠️ `docs/configuration.md:343` - The documentation claims `.linear-seen.tsv` stores immutable event keys, but `seen_append` is never called and the loaded ledger has no consumer. Remove the dead seen-ledger machinery or implement the documented ownership, and document the actual `.linear-history-scans` state.
- ⚠️ `docs/scripts.md:8` - The three new Linear scripts are absent from the repository&#39;s required `bin/` toolbelt inventory. Add purpose rows for the poller, write door, and shared library.
- ℹ️ `bin/fm-linear-poll.sh:453` - The same substantial jq timestamp-normalization filter is duplicated across comment, history, and issue ingestion. Centralize it so the boundary semantics cannot drift independently.

🔧 Fix: Harden Linear resumability and pre-comment validation
- 🚨 `bin/fm-linear-poll.sh:370` - The requirement that captain input never be lost still fails for paginated descendant scans. When a node has multiple child pages, lines 370-374 retain only the final page&#39;s child IDs. If an earlier-page child contains a nested Firstmate reply, participation is never discovered and a bare captain reply is permanently ignored. Accumulate every page&#39;s children durably before advancing the scan.
- 🚨 `bin/fm-linear-act.sh:247` - SIGKILL recovery can still post a stale comment. If issueUpdate succeeds and the process dies before publishing the mutated phase, a later captain edit that preserves the target state and assignee reaches this shortcut. It adopts the captain&#39;s newer updatedAt as the mutation version, so pre-comment validation passes. Verify mutation provenance or refuse this ambiguous recovery instead of adopting the current version.
- ⚠️ `bin/fm-linear-poll.sh:1350` - A state-carrying write that is already at its target performs no issueUpdate and therefore produces no self-authored board history. The completed journal still requires a .board-observed marker, so after three polls it repeatedly reports WRITE NOT OBSERVED despite successful verification. Record whether a mutation was sent or satisfy observation from the verified current snapshot.
- ⚠️ `AGENTS.md:109` - The always-loaded state contract still advertises the removed .linear-seen.tsv ledger and omits the new comment-root and resumable scan state. It also describes six bootstrap mutation sweeps at line 161 while omitting Linear setup. Update these operational declarations to match the implementation.

🔧 Fix: Harden Linear recovery provenance and descendant checkpoints
- 🚨 `bin/fm-linear-poll.sh:668` - Issues requiring deep-history pagination are removed from `issues.json`, while `audit_invariants` inspects only that reduced array. A newly unknown status or turn-marker mismatch can therefore remain silent for multiple resumable scans. Audit each fetched current issue snapshot independently of history completeness.
- 🚨 `bin/fm-linear-poll.sh:1099` - When a Firstmate state mutation has no corresponding history row, snapshot reconciliation publishes it as an unattributed observation. It also cannot mark the completed journal&#39;s board mutation observed, causing a spurious wake followed by repeated `WRITE NOT OBSERVED` alarms. Reconcile snapshot deltas against proven outbox mutations before publication, suppress proven self events, and mark their board observation.
- ⚠️ `bin/fm-linear-poll.sh:731` - Outbox observation markers use unguarded redirection. Because `[ -e &#34;$marker&#34; ]` is false for a dangling symlink, the write can follow that link to another writable file. Reject symlinks and publish both comment and board markers through `fm_linear_atomic_file`.
- ⚠️ `docs/configuration.md:359` - The outbound journal contract is substantially duplicated here and in `.agents/skills/linear-operations/SKILL.md`, contrary to the repository&#39;s one-owner rule. Keep the full contract in one semantic owner and replace the other copy with a concise cross-reference and only surface-specific guidance.

🔧 Fix: Harden Linear invariant audits and write observation
- 🚨 `bin/fm-linear-poll.sh:655` - A resumable history scan replaces its stored current issue snapshot whenever `updatedAt` changes, while that issue remains excluded from snapshot derivation. With persisted head A, a poll can fetch history-suppressed edit B, then replace it with C before pagination completes; completion emits only A to C, so fetched captain edit B never reaches the inbox. Persist every fetched snapshot before replacement or process snapshot ingestion independently of history completeness.
- 🚨 `bin/fm-linear-poll.sh:1316` - Unknown-status reentry can still reuse an acknowledged historical occurrence. After QA entry H1 is acknowledged, a history-suppressed QA to Backlog to QA sequence leaves H1 as the latest returned transition to QA, so this branch selects H1 before considering the newer issue snapshot and suppresses the reentry. Accept a history occurrence only when it proves entry after the previous issue head; otherwise mint a snapshot occurrence from the current `updatedAt`.
- 🚨 `bin/fm-linear-act.sh:417` - `resume` exits on the first failing journal, so one permanently ambiguous journal can prevent every later journal for independent issues from resuming. There is no journal-specific resume command, and the handler contract forbids manual journal edits. Continue independent journals while reporting accumulated failures, or add a supported targeted recovery path.
- ⚠️ `bin/fm-linear-poll.sh:771` - A self-authored history event marks every completed journal for the same issue, target state name, and assignee ID as observed, without matching the event timestamp to `mutated_updated_at`. A later transition back to the same pair can therefore mark an older unobserved write and suppress its `WRITE NOT OBSERVED` alarm. Correlate the history occurrence with exact mutation provenance, preferably normalized `updatedAt` plus state and assignee IDs.

🔧 Fix: Preserve Linear snapshot versions and isolate recovery
4 issues (1 error, 3 warnings) still open:
- 🚨 `bin/fm-linear-poll.sh:633` - A persisted deep scan owns snapshot B only while the issue still satisfies the current `page_history_list` predicate. If newer snapshot C later falls below the moved overlap threshold, C bypasses the scan and advances A to C; scan completion then replays B, emits a false C to B change, and regresses the head. Route every issue with an existing history scan through its ordered snapshot queue regardless of the current pagination predicate.
- ⚠️ `bin/fm-linear-poll.sh:998` - Board observation requires both `toState.id` and `toAssignee.id`, but provider history can omit whichever field was unchanged. A later unrelated edit then prevents snapshot provenance matching and causes repeated `WRITE NOT OBSERVED` alarms. Match the exact mutation timestamp while reconstructing unchanged pair members from the journaled original and target IDs.
- ⚠️ `bin/fm-linear-poll.sh:2` - The poller header does not document the exact persistent ledger, cursor, head, scan, acknowledgment, health, and marker schemas it owns, contrary to the repository&#39;s producer-header ownership rule. Add the owned fields, transitions, and retention mechanics to the header.
- ⚠️ `bin/fm-linear-act.sh:2` - The write-door header omits the versioned outbox journal fields, phase transitions, mutation provenance, completion rename, and observation-marker mechanics it owns. Document that executable recovery contract in the script header.

</details>

<details>

✅ No issues found.
- <code>`git status --short --branch`, `git diff --stat 9823ff8 aa6e660`, and changed-file inventory</code>
- `bin/fm-test-run.sh --family linear`
- `tests/fm-test-run.test.sh`
- <code>Fixture-backed `bin/fm-linear-poll.sh` sequence covering durable capture, second-poll silence, edited comments, and unchanged parent bumps</code>
- <code>`FM_LINEAR_TEST_KILL_AFTER_MUTATION=1 bin/fm-linear-act.sh handoff-to-captain ...`, followed by `bin/fm-linear-act.sh resume` and journal/provider-log verification</code>
- <code>`git diff --exit-code aa6e660e4e2e8f9bcc5e732a94747a2677d6bae5 --` and `test ! -e state`</code>

</details>

<details>
<summary>⚠️ **Document** - 3 issues (1 error, 2 warnings)</summary>

- 🚨 `bin/fm-supervision-lib.sh:54` - Linear-only activation writes `state/fm-linear-inbox.check.sh`, but the supervision predicate only recognizes the Relay shim. With no task or process source, turn-end adapters can stop polling captain input.
- ⚠️ `bin/fm-supervision-instructions.sh:208` - The shared connector cadence is still rendered as “X mode,” which mislabels Linear-only homes. The same stale output appears in `bin/fm-session-start.sh:909`.
- ⚠️ `bin/fm-linear-poll.sh:1512` - The target branch adds trailing whitespace, causing the full-branch diff check to fail.

🔧 Fix: Verify Linear documentation remains current
3 issues (1 error, 2 warnings) still open:
- 🚨 `bin/fm-supervision-lib.sh:54` - Linear-only activation still does not make supervision required because the shared predicate recognizes only the Relay shim.
- ⚠️ `bin/fm-supervision-instructions.sh:208` - Shared connector cadence output remains labeled “X mode” here and in bin/fm-session-start.sh:909.
- ⚠️ `bin/fm-linear-poll.sh:1512` - The branch still contains trailing whitespace, so the full-branch diff check fails.

</details>

<details>

🔧 Fix: Fix reported ShellCheck warnings

</details>

<details>

✅ No issues found.

</details>


josh-padnick:
Opened on the wrong repository by mistake. This work belongs on the josh-padnick fork, not upstream. Closing.

PR #2398, opened 2026-08-14
fix: gate terminal marker removal on genuine activity

prajwal-395:
Do not clear terminal markers (.terminal-mtime-*, .terminal-resurfaced-*) on non-busy cosmetic pane repaints.


prajwal-395:
Closing: raised outside the required no-mistakes path, so it cannot pass this repo CI contribution check. The change itself stands and the branch is kept locally - anchoring the terminal stale re-surface on status mtime rather than pane hash, so a finished-but-unlanded worker stops waking its supervisor on every cosmetic repaint. Reopening properly through the pipeline if and when we choose to spend that run.

PR #2408, opened 2026-08-15
fix(bin): refuse teardown while a task's recorded PR is still open

joliverMI:
## Summary

Tearing down a task deletes `state/<id>.meta`, which is what `bin/fm-pr-merge.sh` resolves the task's PR through. Two incidents an hour apart showed that a task can be torn down while its PR is still open, stranding a proven, mergeable PR with no guarded path to land it.

The existing landed-work check in `validate_worktree_teardown_safety` only inspects local git state (dirty, unpushed, or unreachable-but-landed commits). A branch that is fully pushed and reachable from a remote-tracking branch passes that check cleanly even while its own PR is still open - exactly what happened tonight.

## What changed

- Added `pr_open_state()` next to `pr_is_merged()` in `bin/fm-teardown.sh`, reusing the same `gh pr view --json state,headRefOid` query shape so both share one call pattern (and existing test stubs).
- Added a new refusal at the end of `validate_worktree_teardown_safety`, gated on a non-empty recorded `pr=`, so it never runs for scout/local-only/secondmate teardowns (none of which ever record one).

## Decisions

- **Placement**: the new check runs *after* the existing dirty/unpushed/landed checks, as an independent final gate, not nested inside the unpushed branch (it must fire even when nothing is unpushed - that's the actual incident shape). This keeps messages from contradicting: only one refusal fires per invocation, and the more urgent local-state refusals (uncommitted changes, unlanded commits) take precedence over the higher-level "is this task's PR actually done" question.
- **Forge unreachable**: refuses loudly, naming the PR it could not confirm, rather than tearing down blind - "unreachable" and "confirmed closed" are different facts, and guessing safe on the former defeats the point of the guard.
- **`--force`**: covers this check too, explicitly documented in the header and usage comment (not incidental - it falls out of the same `[ "$FORCE" != "--force" ] || return 0` early return the existing checks already use). `--force` already means "the captain explicitly authorized discarding this," and bypassing the open-PR guard means accepting the PR now needs to be merged or closed by hand, outside `fm-pr-merge.sh`'s guarded path.
- **Refusal text**: tells the operator to merge with `fm-pr-merge.sh` (once green and authorized) or close the PR on GitHub, then retry teardown.

## Test plan

- [x] `bin/fm-lint.sh bin/fm-teardown.sh tests/fm-teardown.test.sh` - clean (ShellCheck 0.11.0)
- [x] `bash tests/fm-teardown.test.sh` - all 62 cases pass, including 4 new ones covering: open PR refuses, `--force` overrides, gh lookup error refuses (fail-safe), and confirmed-closed PR does not block
- [x] `bash tests/fm-teardown-endpoint-safety.test.sh` - all 7 cases still pass (no regression)
- [x] Fixed two pre-existing tests (`test_teardown_prompts_tasks_axi_done_when_compatible`, `test_teardown_manual_backend_prompts_hand_edit_even_when_tasks_axi_present`) that incidentally recorded a `pr=` without caring about its state; they now pin it to a confirmed-closed stub so they stay decoupled from this new check

### DIFF
--- a/bin/fm-teardown.sh
+++ b/bin/fm-teardown.sh
@@ -22,6 +22,15 @@
 # A gh lookup error falls back to the content check; if that is also inconclusive,
 # teardown refuses rather than risk discarding unlanded work.
 # Uncommitted changes are never landed.
+# Once the dirty/unpushed/landed checks above pass, teardown separately REFUSES if
+# the task's recorded pr= is still open on GitHub: a fully pushed, git-history-landed
+# branch can still have its own PR sitting open, and teardown deletes
+# state/<id>.meta - the record bin/fm-pr-merge.sh resolves that PR through - so
+# tearing down now would strand a mergeable PR with no guarded path to land it. A gh
+# lookup error here refuses loudly too (naming the PR firstmate could not confirm)
+# rather than tearing down blind, because "unreachable" and "confirmed closed" are
+# not the same fact. This check only fires when a pr= is actually recorded, so scout,
+# local-only, and secondmate teardowns (none of which ever record one) are unaffected.
 # local-only projects additionally accept work merged into the local default
 # branch (firstmate performs that merge after configured approval) as a fallback
 # for the common case where there is no remote at all.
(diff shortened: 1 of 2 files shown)


joliverMI:
Withdrawn: this belongs on our own fork rather than upstream. Re-opened against joliverMI/firstmate.

PR #2414, opened 2026-08-15
fix: use tracked upstream for self-update

prajwal-395:
Fix the self-update path to resolve the update source from what the home's default branch actually tracks, rather than assuming origin.


prajwal-395:
Closing and re-raising on our own fork.

PR #2431, opened 2026-08-15
feat: Add quota memory extraction for agy harness

prajwal-395:
Give firstmate a memory of agy model quota, so a dispatch rule about quota can be obeyed BEFORE the dispatch rather than discovered by making it.

The sequence is inescapable: to learn that Opus 4.6 is below its floor, you must first dispatch a worker onto Opus 4.6 and read its pane. A reading older than the time-to-reset it recorded means that window has since reset.

This task builds the memory and exposes it:
- Extract model, quota, and reset time from agy pane footer using `bin/fm-agy-quota-lib.sh`
- Store it durably in `state/.agy-quota-<hash>`
- Fail open on unparseable/ambiguous formats
- Return unknown if the reading is older than the time-to-reset window

Testing:
Run tests: `bin/fm-test-run.sh tests/fm-agy-quota-lib.test.sh`
Shellcheck: `shellcheck bin/fm-agy-quota-lib.sh bin/fm-watch.sh`

Live End-to-End parse test:
Input: `Gemini 3.1 Pro (High) | ctx: 10.5% | quota: 94.7% (4h 24m)`
Output: `Gemini 3.1 Pro (High)|94.7%|4h 24m|1786813299`

### DIFF
--- a/bin/fm-agy-quota-lib.sh
+++ b/bin/fm-agy-quota-lib.sh
@@ -1,0 +1,130 @@
+#!/usr/bin/env bash
+# fm-agy-quota-lib.sh - Quota memory observation and read helpers.
+# Usage: . bin/fm-agy-quota-lib.sh
+
+# fm_agy_parse_reset_time: converts '4h 24m', '1d 2h', etc to seconds.
+fm_agy_parse_reset_time() {
+  local ts="$1"
+  local total=0
+  local d=0 h=0 m=0 s=0
+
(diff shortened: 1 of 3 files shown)


prajwal-395:
Closing this - it was opened against upstream by mistake. This work ships through our own fork, where it has already landed, so this PR is a duplicate rather than a pending contribution. Apologies for the noise.

PR #2471, opened 2026-08-16
feat(session): emit Claude Code session name at startup

abhishekdubey331:
Print the configured session name near the top of the digest so the captain sees which name this session is set to use, and update README to launch with `claude --name firstmate` so future sessions pick the name up automatically.

The name is read from $FM_HOME/config/session-name (default: firstmate), so each firstmate home can declare its own title without touching the script.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,8 +89,11 @@ Then launch one of the co-primary harnesses; AGENTS.md takes over from there:
 **Claude Code**
 
 ```sh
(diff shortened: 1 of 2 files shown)


abhishekdubey331:
Closing - this change is staying on the captain's fork rather than upstream. Firstmate is treated as the captain's own template, not a contribution target.

PR #2481, opened 2026-08-16
fix(procevent): deliver captured results at most once

joliverMI:
## Summary

A Lavish-captured design-answer message reached a secondmate five times (`state/spotfx.status`: "note your delivery machinery quadruplicated this message", then a fifth copy). Diagnosis, as requested by the brief:

- **Reproduced without touching a live Captain.** `tests/fm-procevent.test.sh` now seeds one captured Lavish generation directly in a fixture home and drives it through the real `fm-procevent.sh` machinery.
- **Where the retry actually lives.** It is not `lavish-axi`'s own retry loop. `bin/fm-procevent.sh` already proves capture is exactly-once (heavily tested: claim identity, ownership races, terminal retirement, etc. all pass). The gap is downstream: forwarding a captured result to a secondmate runs `fm-send.sh` and only afterward, manually and separately, calls `bin/fm-procevent.sh handled`. Nothing pairs the two atomically. Every reconcile re-announces an unhandled capture "regardless of any earlier publication" (by design, for crash recovery) — and every time that re-announcement got handled again (a restart, a compaction, a turn that never reached the second step), the forward repeated.
- **Shape of the duplicate.** Repeated deliveries of one stored payload (one captured `(source-id, sequence)` generation), not retries of an unacknowledged send — the built-in pending-reply recovery path already sends its one retry with the same correlation id and is not implicated.
- **Root cause is ours**, not `lavish-axi`'s: the Lavish adapter is "deliberately thin" and has no `autohandle`/self-announcing capability the way `remote-reply` does, so nothing in code stops a forgotten acknowledgement from repeating the forward.

## Fix

Add `bin/fm-procevent.sh deliver <source-id> <sequence> -- <command>...`: checks handled status, runs `<command>`, and marks the generation handled — all under the same per-source lock `handled` already uses. A repeated call for a generation already delivered runs the command **zero** times and reports `already-delivered`; a command that exits nonzero leaves the generation unhandled and eligible for the next retry instead of being silently dropped or mistaken for delivered. This makes the duplicate structurally impossible rather than filtering it after the fact.

Updated `process-event-sources` skill and `docs/configuration.md` (the runner's operating-contract owner) to point handlers at `deliver` for exactly this class of forward, replacing the old "run it yourself, then remember to call `handled`" guidance that made the defect possible.

## Cost

Measured: the new regression test proves five delivery attempts against one captured generation now run the downstream forward once instead of five - the other four are eliminated structurally, not filtered. The historical incident's own token cost was not measured at the time and cannot be reconstructed after the fact, so that specific number is unmeasured.

## Test plan

- [x] `bin/fm-lint.sh` (all changed scripts, shellcheck-clean)
- [x] `bin/fm-doc-audience-check.sh` (clean)
- [x] `tests/fm-procevent.test.sh` — full pass, including two new regression tests: five `deliver` attempts against one generation forward once; a failing `deliver` attempt leaves the generation eligible for retry instead of being dropped or repeated
- [x] `bin/fm-test-run.sh --changed` — `total=81 failed=3`; all three failures independently reproduced identically on clean `main` (pre-existing, unrelated to this diff):
  - `tests/fm-remote-job-orphan-reap.test.sh` — this sandbox's PID-1-reparenting assumption doesn't hold (environment artifact)
  - `tests/fm-test-run.test.sh` — a `comm: ... not in sorted order` failure in its own `--check-coverage` self-test (locale/sort-order sensitivity)
  - `tests/fm-remote-secondmate-lifecycle-e2e.test.sh` — prints `ALL TESTS PASSED`; only the final `rm -rf` of its temp root intermittently races with `Directory not empty` in this sandbox

No live Captain was messaged during diagnosis or testing.


joliverMI:
Withdrawn: this belongs on our own fork rather than upstream. Re-opened against joliverMI/firstmate.

PR #2507, opened 2026-08-16
fix(bin): reconcile decision holds whose routed work completed out of band

clydefrogtech:
## Intent

Fix a deadlock in the decision-hold lifecycle: a captain hold whose routed work was completed out of band (a direct `tasks-axi done`) before the decision close ran became permanently unclosable, so the answered decision stayed open in the durable record forever.

## Root cause

In `bin/fm-decision-hold.sh`, the three close paths each refused this state for a locally-correct reason:

- `resolve` refused because the routed task was already done (a guard meant to keep an exact retry from recording a false routing).
- `decline` refused because the done task's durable `blocked-by` edge to the hold survived the out-of-band close, so the hold still "blocked routed work".
- `repair` refused because the hold was still actively held.

With all three refusing, the hold stayed queued+held permanently even though the captain had answered: Bearings' Captain's Call listed it forever, and no firstmate path could close or re-verify it from evidence.

## Fix

`resolve` now reconciles a done routed task when that task's own durable `blocked-by` edge still names the hold. The surviving edge is the evidence that the work really was routed behind this decision, so the close is recorded from proof rather than a guess: the decision and routed identities are written to the hold body, the stale edge is cleared through tasks-axi, and the hold closes Done with the usual idempotent retry identity.

A done task without that surviving edge is still refused, and `decline` keeps refusing while any edge survives, so neither unrouted path can manufacture a routing.

## Regression coverage

`tests/fm-decision-hold-lifecycle.test.sh` gains an end-to-end regression that reproduces the deadlock shape exactly (synthetic `sample` identities), then asserts:

- the deadlocked hold is unclosable by `decline` and `repair` and still surfaces as an open captain-hold in Bearings;
- `resolve` reconciles it from the surviving edge, records `Resolution mode: routed` with the captain decision text and the done routed work, clears the stale edge, satisfies `verify`, and leaves Bearings' Captain's Call;
- an identical retry is idempotent and a retry with a changed captain decision is rejected;
- a done task with no surviving edge is still refused, leaving the hold queued and held.

## Verification

- `bash tests/fm-decision-hold-lifecycle.test.sh` - 13/13 ok, including the new regression
- `bash tests/fm-fleet-snapshot-view.test.sh`, `bash tests/fm-bearings-snapshot.test.sh`, `bash tests/fm-brief.test.sh`, `bash tests/fm-teardown.test.sh` - all ok
- `bin/fm-lint.sh` - ShellCheck 0.11.0 (pinned), clean
- `bin/fm-doc-audience-check.sh` - ok surfaces=68 local_links=253

The mechanism owner `docs/decision-hold-lifecycle.md` records the reconciliation rule and the 2026-08-16 verification date.

### DIFF
--- a/bin/fm-decision-hold.sh
+++ b/bin/fm-decision-hold.sh
@@ -43,7 +43,12 @@
 # include a `Resolution mode:` naming their path; older routed records remain valid.
 #
 # `resolve` is the routed path. It requires every --routed-to task to exist and to
-# be blocked by the hold. It writes the captain decision and routed identities into
+# be blocked by the hold, with one reconciliation exception: a routed task that
+# already completed out of band (a direct tasks-axi done) is accepted when its own
+# durable blocked-by edge still names the hold, which is the surviving proof the
+# work really was routed behind this decision; a done task without that edge is
+# refused, and decline likewise keeps refusing it, so neither path can close the
+# hold from a guess. It writes the captain decision and routed identities into
 # the hold body, clears those dependency edges, and only then marks the hold Done.
 # A failure before the final step leaves the captain hold open.
 #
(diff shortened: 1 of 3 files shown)


clydefrogtech:
Opened against the wrong remote; superseded by a PR on clydefrogtech/firstmate. No merge intended here.

PR #2524, opened 2026-08-17
fix(bin): re-arm after an abandoned auto-arm claim and defer a wedge escalation while a worktree is written

Inthuson:
## Intent

Fix two live supervision defects in firstmate's own tooling that the captain hit during a real fleet incident.

Defect 1: a lapsed monitoring chain never re-armed. The auto-arm claim under state/.claude-autoarm.lock could be left behind by an owner that died, and the ledger reasoning on its own could not tell an abandoned claim from a live one, so the chain stayed down. The fix records a pid-identity file inside the claim directory, so an identity mismatch is abandonment proof by itself. The captain decided this explicitly at the review gate (finding autoarm-abandonment-ledger-only): FIX it now and add the pid-identity file, with the scope increase authorized. The existing ledger reasoning stays untouched rather than being replaced.

Defect 2: the wedge detector escalated a crewmate that was demonstrably writing code. The fix adds a bounded, pruned, depth-limited worktree write probe (crew_worktree_written_since) as a third liveness input, consulted only at the moment an escalation would otherwise fire. Overriding constraint stated by the captain: a stall that writes nothing must still escalate on the existing schedule, so this may only reduce false positives and must never delay a real escalation. A deferral is bounded rather than permanent: the .writing-since marker ages the whole chain so a churning worktree still re-surfaces once per resurface window. The probe prunes .git first so firstmate's own read-only git commands against a worktree can never make it self-fulfilling, and a secondmate home's supervision churn is excluded for the same reason.

Later captain decisions this branch carries, so they should not read as surprises in the diff:
- At the document gate the captain decided APPROVE on away-mode-wedge-deferral-gap and took option (a): document the away-mode boundary and the probe's filesystem limit rather than change that behavior on this branch.
[...]

PR #2541, opened 2026-08-17
fix(watch): stop false wedge wakes and re-arm after abandoned claims

Inthuson:
## Intent

Fix a supervision bug in firstmate's always-on watcher: a worker that is ALIVE and has correctly declared it is waiting (appended a `paused:` status line, or carries a durable `captain-held` transfer) cannot be quieted. It surfaces a bare `stale: <window>` wake on every poll, forever, and the only way to stop it today is to kill the healthy worker, which is backwards. Confirmed live: a parked-but-alive worker generated about ten bare stale wakes in one hour, accelerating toward one a minute.

Acceptance criteria the captain set: (1) a live worker whose last status line is `paused:` or `captain-held:` is absorbed on the bounded cadence exactly as a dead one already is; (2) the wedge-detection path still escalates a worker that is genuinely stuck and has NOT declared a wait; (3) a test exists that fails before the change for the live-agent declared-wait shape.

Overriding safety constraint the captain stated: a stall that writes nothing must still escalate on the existing schedule. This change may ONLY reduce false positives. Widening what gets suppressed in order to make a test pass was explicitly forbidden, because the failure mode created would be a silently dead worker that firstmate reports as fine, which costs far more than the noise being fixed.

Deliberate decisions and context a reviewer reading only the diff would not know:

BASE BRANCH: the captain explicitly required branching from an unmerged sibling branch (head b3d8759, PR https://github.com/kunchenguid/firstmate/pull/2524) and NOT from the default branch, because this fix builds directly on that work. Those commits therefore appear in this branch's history until 2524 merges. That is expected and ordered: do not rebase onto the default branch, do not remove them, do not cherry-pick.

ROOT CAUSE: pause_state_class gated its whole declared-wait verdict on the backend confidently reporting the agent DEAD (added alongside the bounded cadence in ab8cea6, #743). For any non-dead agent it returned `none` before the `paused` verdict crew_absorb_class had already derived, so the two halves of the feature disagreed: handle_paused_stale creates the .paused-<key> and .paused-rechecked-<key> markers, but the classifier deciding whether to call it could not return `paused` while the agent was up. Two consequences compounded: on a first-sight stale hash the non-terminal branch took surface_nonterminal_stale (the bare wake) instead of handle_paused_stale; and on a NEW pane hash the loop's tail branch ran clear_pause_tracking, deleting the .paused-resurfaced-<key> throttle that bounds the cadence, so every pane repaint rearmed a fresh full-rate wake. That is why the observed rate accelerated instead of settling. bin/fm-supervise-daemon.sh's classify_stale already absorbed a live declared pause from the status line alone with no liveness gate, so the always-on watcher was the outlier.

FIX: pause_state_class now consults no liveness verdict at all. The declaration, not whether the agent process is still up, is what makes the idling expected; the finite PAUSE_RESURFACE_SECS bound is what keeps absorbing safe, and only a window whose crew actually WROTE a declaration can reach that function.

NEW CLASSIFIER TOKEN, deliberate and load-bearing: crew_absorb_class gained a fourth token, `actionable`, for the authoritative states parked|blocked|done|failed. Without it those states would fall into `none` and be absorbed for up to an hour, which would newly hide a live ask-user gate or a checks-green PR - exactly the new false negative the safety constraint forbids. It refines what used to be `none`; both crew_is_* wrappers test exact tokens, and every watcher case site routes `actionable` exactly where `none` already went, so no other routing moves. Precedence is working > actionable > declared wait. The token also removes a per-recheck backend process-name read from the absorb path.

COMPOSITION WITH THE WRITE DEFERRAL: the captain asked for this to be reasoned about explicitly rather than bolted on beside it, and raised as a keyed needs-decision only if it needed a real design decision. It did not, so it is documented in the commit body. Total order is working (active run step or busy pane) > declared wait > recent worktree writes. They can never both decide the same poll: wedge_defer_writing is reachable only through wedge_timer_check, the declared-wait branch is evaluated first at every stale site, a `paused` verdict always goes to handle_paused_stale which never calls the wedge timer, and handle_paused_stale calls clear_write_tracking to retire any deferral chain. Declared-wait-wins is also the safer resolution where they would disagree: both absorb on the same PAUSE_RESURFACE_SECS bound, but the declaration anchors its age on the status file's mtime, which is older, while the deferral anchors on a freshly touched .writing-since-<key>, so letting the declaration win re-surfaces sooner than or level with the deferral, never later.

TESTS: one three-phase differential added to the existing tests/fm-watch-triage.test.sh (extend the existing runner, do not invent a new one). It drives a real fm-watch.sh subprocess and asserts observable wake output and durable state, never implementation source content. Agent liveness is faked rather than driven through real panes, which the captain explicitly preferred over provisioning real panes. Proven to discriminate against the pre-change base: pre-change it reports a bare stale on first sight and six bare wakes across six pane repaints; post-change exactly one bounded recheck that names the wait ("paused 501s, awaiting external"), with the throttle surviving every repaint. Phase three - the same live worker, same idle pane, same inconclusive authoritative verdict, but no declaration - passes on BOTH trees on purpose: it is the safety half of the differential, proving the undeclared-quiet escalation guarantee is untouched.

PRE-EXISTING TEST DEFECT FIXED IN SCOPE: one existing test exported FM_FAKE_CREW_STATE and never unset it, leaking a `paused` verdict into whatever test ran next. That had silently made the safety phase of the new differential vacuous (it failed for the leak, not for the product). Fixed by unsetting it there, and by having every phase of the new test pin its own authoritative verdict instead of inheriting an ambient one. Deliberately in scope because it is what made the new test's safety half untrustworthy.

COMPATIBILITY: the change removes a runtime-backend-dependent read, so this path no longer varies by whether a backend's agent-liveness probe can answer `dead`. Previously any backend answering `unknown`, including remote or experimental ones, could never quiet a declared wait, and the same crew behaved differently under tmux than elsewhere for reasons unrelated to its work. Every supported harness and backend now takes the same bounded cadence. No new harness-dependent signal is introduced, so no live-harness opt-in guard is needed and no per-adapter verification record is affected.

DOCUMENTATION: docs/architecture.md was updated to state the new rule (the declaration, not a backend liveness verdict, governs; authoritative crew state still outranks it) rather than leaving the removed liveness gate documented. The wording deliberately says a parked/blocked/done/failed verdict "keeps whatever routing that pane already had" and is "never newly hidden", which is precise rather than overclaiming: at the repeat-sight branch such a verdict was already absorbed before this change, and this change does not make that worse.

VERIFICATION HONESTY: the three affected suites were run green (fm-watch-triage 59 checks, fm-daemon 99, fm-crew-state 49), as was bin/fm-doc-audience-check.sh. Lint was NOT independently verifiable from the agent's environment, so it is not claimed as agent-verified evidence; the captain separately ran the canonical pinned ShellCheck 0.11.0 set at this exact head and reported zero findings.

## What Changed

- `pause_state_class` in `bin/fm-watch.sh` no longer consults `fm_backend_agent_alive`: a window whose last status line is a declared `paused:` wait or a durable `captain-held` transfer takes the bounded `PAUSE_RESURFACE_SECS` recheck cadence whether its agent is live or exited, instead of falling through to a bare `stale: <window>` wake and dropping the `.paused-resurfaced-<key>` throttle on every pane repaint. `crew_absorb_class` gained a fourth token, `actionable`, for the authoritative `parked|blocked|done|failed` states so a real gate, ready PR, or failure keeps the routing `none` already had, `.paused-rechecked-<key>` now stores that verdict (instead of an epoch) so a throttled poll replays the token it recorded, and `surface_nonterminal_stale` stopped overwriting it.
- Added `crew_worktree_written_since` to `bin/fm-classify-lib.sh` — one pruned, depth-bounded, `fm_run_timed`-bounded `find` over the task's recorded worktree, skipping secondmate homes — and wired it into `wedge_timer_check`, which now takes a task argument and defers via the new `wedge_defer_writing` when the worktree was written since the idle window opened. New `.writing-since-<key>` / `.writing-resurfaced-<key>` markers age that deferral chain through a shared `resurface_absorbed` helper (also now used by `handle_paused_stale`), `clear_write_tracking` retires the chain wherever stale bookkeeping resets, the daemon's `clear_pause_tracking` removes both markers, and a single `window_key` helper replaces the four duplicated key derivations. New knobs: `FM_WORKTREE_WRITE_PRUNE`, `FM_WORKTREE_WRITE_MAXDEPTH`, `FM_WORKTREE_WRITE_TIMEOUT`.
- Added `fm_autoarm_claim_record_identity`, `fm_autoarm_claim_abandoned`, and `fm_autoarm_release_abandoned` to `bin/fm-wake-lib.sh`, proving a Claude Stop auto-arm claim finished from its own ledger entry (`owner_pid` matching the lock with a non-`arming` outcome) or from a recorded `pid-identity` that no longer matches the live pid, and reclaiming it under the lock's steal mutex. `bin/fm-claude-stop-autoarm.sh` records its identity before publishing the `autoarm` role and reclaims an abandoned lock once before standing down; `bin/fm-turnend-guard.sh` stops reading such a lock as recovery in flight in both `autoarm_owns_recovery` and `terminal_fail_open`, clearing it rather than allowing a blind stop.

## Risk Assessment

✅ Low: The round-3 fix is sound and I confirmed its invariant holds by tracing all four pause_state_class call sites, the sole-writer claim, and routing parity for the new `actionable` token; every acceptance criterion is met with no forbidden widening, and the three remaining items are comment accuracy, a test-assertion hardening, and an already-authorized tradeoff.

## Testing

<code>No baseline test output was supplied for this phase, so I started from a clean worktree at the target commit and ran targeted validation myself. I exercised the three suites the change touches (fm-watch-triage 60 checks, fm-crew-state 49, fm-daemon 99, all green), then proved the regression differential by reverting only the two product files to the pre-fix base and re-running the new tests unchanged: they fail there with exactly the reported symptom and pass on this change. For end-user-visible evidence I replayed the reported scenario through a real fm-watch.sh subprocess on both trees and captured the wake reasons firstmate actually receives - six bare `stale: &lt;window&gt;` wakes across six pane repaints before, one bounded `paused 500s, awaiting external` recheck after, for both the `paused:` and `captain-held:` live shapes - alongside three safety replays (undeclared quiet crew, a stall that writes nothing, and a live ask-user gate behind a stale `paused:` line) whose output is identical on both trees, so the change only removes false positives. This is a CLI/supervision-daemon behavior change with no rendered surface, so the reviewer-visible artifacts are watcher wake transcripts and durable wake-queue state rather than screenshots. I removed the transient driver scripts from the working tree (archived copies live in the evidence directory) and confirmed the worktree is clean at 1e4017b; lint and the doc-audience check were deliberately not run in this phase.</code>

<details>
<summary>Evidence: Differential: wake reasons firstmate received, pre-fix vs post-fix (same fixture, six pane repaints)</summary>

<code>BEFORE (b3d8759, pre-fix):&#10;queued: stale: test:fm-live-held&#10;queued: stale: test:fm-live-held&#10;queued: stale: test:fm-live-held&#10;queued: stale: test:fm-live-held&#10;queued: stale: test:fm-live-held&#10;queued: stale: test:fm-live-held&#10;TOTAL stale wakes: 6 of which BARE &#34;stale: test:fm-live-held&#34;: 6&#10;&#10;AFTER (1e4017b, this change):&#10;queued: stale: test:fm-live-held (paused 500s, awaiting external - declared pause, rechecked on a long cadence not a wedge; confirm the wait still holds)&#10;TOTAL stale wakes: 1 of which BARE &#34;stale: test:fm-live-held&#34;: 0&#10;&#10;Unchanged on both trees:&#10;undeclared quiet crew -&gt; stale: test:fm-live-quiet (surfaced at once)&#10;stall that writes nothing -&gt; stale: test:fm-live-wedged (idle 500s, possible wedge, escalation 1)&#10;live ask-user gate behind a paused: line -&gt; stale: test:fm-gate on every one of 3 polls</code>

```text
Live worker that declared its wait - wake reasons firstmate actually received
(same fixture, six pane repaints, only the product code differs)

BEFORE (b3d8759, pre-fix):
    queued: stale: test:fm-live-held
    queued: stale: test:fm-live-held
    queued: stale: test:fm-live-held
    queued: stale: test:fm-live-held
    queued: stale: test:fm-live-held
    queued: stale: test:fm-live-held
  TOTAL stale wakes: 6   of which BARE "stale: test:fm-live-held": 6

AFTER (1e4017b, this change):
    queued: stale: test:fm-live-held (paused 500s, awaiting external - declared pause, rechecked on a long cadence not a wedge; confirm the wait still holds)
  TOTAL stale wakes: 1   of which BARE "stale: test:fm-live-held": 0

Unchanged on both trees:
  undeclared quiet crew        -> stale: test:fm-live-quiet (surfaced at once)
  stall that writes nothing    -> stale: test:fm-live-wedged (idle 500s, possible wedge, escalation 1)
  live ask-user gate behind a paused: line -> stale: test:fm-gate on every one of 3 polls
```
</details>
<details>
<summary>Evidence: Full watcher transcript, this change: live `paused:` and `captain-held:` absorbed on the bounded cadence, undeclared quiet crew and a non-writing stall still escalate</summary>

```text
=== firstmate always-on watcher: quieting a LIVE worker that declared its wait ===
product code under test: 1e4017b (this change, post-fix)

=== LIVE worker, last status line: "paused: awaiting the upstream release" ===
fixture: agent ALIVE (pane_current_command=grok), pane idle, authoritative verdict inconclusive,
         declaration aged 500s, FM_PAUSE_RESURFACE_SECS=240 (bounded recheck cadence)

--- poll round 1 (pane repainted) ---
watcher woke firstmate with: stale: test:fm-live-held (paused 500s, awaiting external - declared pause, rechecked on a long cadence not a wedge; confirm the wait still holds)
--- poll round 2 (pane repainted) ---
watcher woke firstmate with: (nothing - absorbed silently)
--- poll round 3 (pane repainted) ---
watcher woke firstmate with: (nothing - absorbed silently)
--- poll round 4 (pane repainted) ---
watcher woke firstmate with: (nothing - absorbed silently)
--- poll round 5 (pane repainted) ---
watcher woke firstmate with: (nothing - absorbed silently)
--- poll round 6 (pane repainted) ---
watcher woke firstmate with: (nothing - absorbed silently)

--- durable wake queue across all six repaints ---
  queued: stale: test:fm-live-held (paused 500s, awaiting external - declared pause, rechecked on a long cadence not a wedge; confirm the wait still holds)
TOTAL stale wakes: 1   of which BARE "stale: test:fm-live-held": 0
pause cadence marker (.paused-test_fm-live-held): PRESENT

=== LIVE worker, last status line: "captain-held: transferred to the captain backlog" ===
fixture: agent ALIVE (pane_current_command=grok), pane idle, authoritative verdict inconclusive,
         declaration aged 500s, FM_PAUSE_RESURFACE_SECS=240 (bounded recheck cadence)

--- poll round 1 (pane repainted) ---
watcher woke firstmate with: stale: test:fm-live-cheld (paused 500s, awaiting external - declared pause, rechecked on a long cadence not a wedge; confirm the wait still holds)
--- poll round 2 (pane repainted) ---
watcher woke firstmate with: (nothing - absorbed silently)
--- poll round 3 (pane repainted) ---
watcher woke firstmate with: (nothing - absorbed silently)
--- poll round 4 (pane repainted) ---
watcher woke firstmate with: (nothing - absorbed silently)
--- poll round 5 (pane repainted) ---
watcher woke firstmate with: (nothing - absorbed silently)
--- poll round 6 (pane repainted) ---
watcher woke firstmate with: (nothing - absorbed silently)

--- durable wake queue across all six repaints ---
  queued: stale: test:fm-live-cheld (paused 500s, awaiting external - declared pause, rechecked on a long cadence not a wedge; confirm the wait still holds)
TOTAL stale wakes: 1   of which BARE "stale: test:fm-live-cheld": 0
pause cadence marker (.paused-test_fm-live-cheld): PRESENT

=== safety half: the SAME live worker, same idle pane, but NO declaration ===
watcher woke firstmate with: stale: test:fm-live-quiet
pause cadence marker: absent (undeclared quiet crew escalates on the unchanged schedule)

=== safety constraint: a stall that writes nothing still escalates on the existing schedule ===
fixture: agent ALIVE, pane idle, NO declaration, authoritative "working · source: run-step"
         (absorbed once with a wedge timer), timer then aged past FM_STALE_ESCALATE_SECS=240
priming poll (run step active, pane static): woke firstmate with: (nothing - absorbed, wedge timer started)
poll past the wedge threshold: woke firstmate with: stale: test:fm-live-wedged (idle 500s, possible wedge, escalation 1)

=== end ===
```
</details>
<details>
<summary>Evidence: Full watcher transcript, pre-fix base: the same live declared waits flood a bare stale wake on every poll</summary>

```text
=== firstmate always-on watcher: quieting a LIVE worker that declared its wait ===
product code under test: b3d8759 (pre-fix base; bin/fm-watch.sh + bin/fm-classify-lib.sh reverted, tests unchanged)

=== LIVE worker, last status line: "paused: awaiting the upstream release" ===
fixture: agent ALIVE (pane_current_command=grok), pane idle, authoritative verdict inconclusive,
         declaration aged 500s, FM_PAUSE_RESURFACE_SECS=240 (bounded recheck cadence)

--- poll round 1 (pane repainted) ---
watcher woke firstmate with: stale: test:fm-live-held
--- poll round 2 (pane repainted) ---
watcher woke firstmate with: stale: test:fm-live-held
--- poll round 3 (pane repainted) ---
watcher woke firstmate with: stale: test:fm-live-held
--- poll round 4 (pane repainted) ---
watcher woke firstmate with: stale: test:fm-live-held
--- poll round 5 (pane repainted) ---
watcher woke firstmate with: stale: test:fm-live-held
--- poll round 6 (pane repainted) ---
watcher woke firstmate with: stale: test:fm-live-held

--- durable wake queue across all six repaints ---
  queued: stale: test:fm-live-held
  queued: stale: test:fm-live-held
  queued: stale: test:fm-live-held
  queued: stale: test:fm-live-held
  queued: stale: test:fm-live-held
  queued: stale: test:fm-live-held
TOTAL stale wakes: 6   of which BARE "stale: test:fm-live-held": 6
pause cadence marker (.paused-test_fm-live-held): PRESENT

=== LIVE worker, last status line: "captain-held: transferred to the captain backlog" ===
fixture: agent ALIVE (pane_current_command=grok), pane idle, authoritative verdict inconclusive,
         declaration aged 500s, FM_PAUSE_RESURFACE_SECS=240 (bounded recheck cadence)

--- poll round 1 (pane repainted) ---
watcher woke firstmate with: stale: test:fm-live-cheld
--- poll round 2 (pane repainted) ---
watcher woke firstmate with: stale: test:fm-live-cheld
--- poll round 3 (pane repainted) ---
watcher woke firstmate with: stale: test:fm-live-cheld
--- poll round 4 (pane repainted) ---
watcher woke firstmate with: stale: test:fm-live-cheld
--- poll round 5 (pane repainted) ---
watcher woke firstmate with: stale: test:fm-live-cheld
--- poll round 6 (pane repainted) ---
watcher woke firstmate with: stale: test:fm-live-cheld

--- durable wake queue across all six repaints ---
  queued: stale: test:fm-live-cheld
  queued: stale: test:fm-live-cheld
  queued: stale: test:fm-live-cheld
  queued: stale: test:fm-live-cheld
  queued: stale: test:fm-live-cheld
  queued: stale: test:fm-live-cheld
TOTAL stale wakes: 6   of which BARE "stale: test:fm-live-cheld": 6
pause cadence marker (.paused-test_fm-live-cheld): PRESENT

=== safety half: the SAME live worker, same idle pane, but NO declaration ===
watcher woke firstmate with: stale: test:fm-live-quiet
pause cadence marker: absent (undeclared quiet crew escalates on the unchanged schedule)

=== safety constraint: a stall that writes nothing still escalates on the existing schedule ===
fixture: agent ALIVE, pane idle, NO declaration, authoritative "working · source: run-step"
         (absorbed once with a wedge timer), timer then aged past FM_STALE_ESCALATE_SECS=240
priming poll (run step active, pane static): woke firstmate with: (nothing - absorbed, wedge timer started)
poll past the wedge threshold: woke firstmate with: stale: test:fm-live-wedged (idle 500s, possible wedge, escalation 1)

=== end ===
```
</details>
<details>
<summary>Evidence: Live ask-user gate behind a stale `paused:` line still surfaces on every poll (this change)</summary>

<code>--- poll round 1 (first stale sight) ---&#10;watcher woke firstmate with: stale: test:fm-gate&#10;--- poll round 2 (pane repainted moments later) ---&#10;watcher woke firstmate with: stale: test:fm-gate&#10;--- poll round 3 (pane repainted moments later) ---&#10;watcher woke firstmate with: stale: test:fm-gate</code>

```text
=== a LIVE ask-user gate sitting behind the crew own `paused:` line ===
product code under test: 1e4017b (this change, post-fix)
fixture: agent ALIVE, pane idle, status log last line "paused: awaiting an external decision",
         authoritative state "state: parked · source: run-step · parked at review: 2 finding(s) (ask-user: authority decision)"
         FM_PAUSE_RESURFACE_SECS=999 - if the gate were absorbed it would stay hidden for 999s

--- poll round 1 (first stale sight) ---
watcher woke firstmate with: stale: test:fm-gate
--- poll round 2 (pane repainted moments later) ---
watcher woke firstmate with: stale: test:fm-gate
--- poll round 3 (pane repainted moments later) ---
watcher woke firstmate with: stale: test:fm-gate

=== end ===
```
</details>
<details>
<summary>Evidence: Same live ask-user gate on the pre-fix base - identical, so no new false negative was introduced</summary>

```text
=== a LIVE ask-user gate sitting behind the crew own `paused:` line ===
product code under test: b3d8759 (pre-fix base; product code reverted)
fixture: agent ALIVE, pane idle, status log last line "paused: awaiting an external decision",
         authoritative state "state: parked · source: run-step · parked at review: 2 finding(s) (ask-user: authority decision)"
         FM_PAUSE_RESURFACE_SECS=999 - if the gate were absorbed it would stay hidden for 999s

--- poll round 1 (first stale sight) ---
watcher woke firstmate with: stale: test:fm-gate
--- poll round 2 (pane repainted moments later) ---
watcher woke firstmate with: stale: test:fm-gate
--- poll round 3 (pane repainted moments later) ---
watcher woke firstmate with: stale: test:fm-gate

=== end ===
```
</details>
<details>
<summary>Evidence: New tests against pre-change product code (must fail)</summary>

<code>not ok - a live crew&#39;s declared wait surfaced instead of being absorbed: stale: test:fm-live-held&#10;EXIT=1</code>

```text
not ok - a live crew's declared wait surfaced instead of being absorbed: stale: test:fm-live-held
EXIT=1
```
</details>
<details>
<summary>Evidence: New tests against this change (must pass)</summary>

<code>ok - a LIVE crew&#39;s declared wait is absorbed on the bounded cadence, while an undeclared quiet crew still surfaces at once&#10;ok - an actionable verdict behind a declared wait keeps its recheck throttle and is never replayed as a pause&#10;EXIT=0</code>

```text
ok - a LIVE crew's declared wait is absorbed on the bounded cadence, while an undeclared quiet crew still surfaces at once
ok - an actionable verdict behind a declared wait keeps its recheck throttle and is never replayed as a pause
EXIT=0
```
</details>
<details>
<summary>Evidence: New tests pass with a hostile ambient FM_FAKE_CREW_STATE exported (each phase pins its own verdict)</summary>

```text
ok - a LIVE crew's declared wait is absorbed on the bounded cadence, while an undeclared quiet crew still surfaces at once
ok - an actionable verdict behind a declared wait keeps its recheck throttle and is never replayed as a pause
EXIT=0
```
</details>
<details>
<summary>Evidence: tests/fm-watch-triage.test.sh full run (60 checks, exit 0)</summary>

```text
ok - signal_reason_is_actionable: benign absorbed, captain verbs and coalesced batches surfaced
ok - stale_is_terminal: terminal status surfaces, non-terminal and no-status are benign
ok - scan_captain_relevant_statuses lists only captain-relevant statuses
ok - crew_is_provably_working: only working+run-step/pane is provable; idle/finished/parked/failed/unknown surface
ok - status_is_paused: only the leading paused verb matches, and paused is not captain-relevant
ok - crew_absorb_class: working/paused/actionable/none from one read; crew_is_paused and crew_is_provably_working stay narrow
ok - crew_worktree_written_since: real writes are evidence; no worktree, no anchor, quiet trees, .git churn and a mate's own home are not
ok - an empty FM_WORKTREE_WRITE_PRUNE widens the probe to the whole depth-bounded tree instead of disabling it
ok - an empty FM_WORKTREE_WRITE_PRUNE exported into the environment prunes nothing, widening the probe
ok - the worktree write probe is wall-clock bounded, and hitting the bound reads as no write evidence
ok - signal_crew_provably_working: benign only when every referenced crew is provably working
ok - a secondmate's status signal is never absorbed as provably working; crewmates are unaffected
ok - a no-verb signal whose crew is provably working is absorbed (no exit, no queue, suppressor advanced, beacon present)
ok - a bare turn-end whose crew is provably working (busy pane) is absorbed
ok - a bare turn-end whose crew is not provably working is surfaced (the swallowed-finish fix)
ok - a no-verb working: note whose crew is idle with no running pipeline is surfaced
ok - a secondmate's status note surfaces even while its own agent is busy
ok - a self-announced close never wakes its own home, and the next real note still does
ok - captain-relevant signal is surfaced (queue + exit) and marked surfaced
ok - a stale pane sitting on a terminal status is surfaced (queue + exit)
ok - a stale terminal-looking status is overridden and absorbed while a run is actively working, then wedge-escalated
ok - provably-working non-terminal stale is absorbed on first sight, then wedge-escalated past the threshold
ok - consecutive wedge escalations on the same pane accumulate and demand deep inspection at the threshold
ok - a pane becoming active again resets the consecutive wedge-escalation counter
ok - a busy worker below the turn-age bound remains working with no escalation
ok - a busy worker with a stable pane hash still escalates once its completed-turn age reaches the bound
ok - a busy worker whose pane hash changes every poll still escalates once its completed-turn age reaches the bound
ok - touching a busy worker's completed-turn marker resets the age and prevents an old-age escalation
ok - repeated busy turn-age escalations reuse the existing escalation counter and demand deep inspection at the threshold
ok - the production default busy-turn-age bound is 3600s (5min under does not wedge, 66min over does)
ok - a not-provably-working non-terminal stale is surfaced immediately (never left to wait out the timer)
ok - a declared pause is absorbed on first sight, then re-surfaced as a recheck past the threshold, never wedge-escalated
ok - exited declared-pause and captain-held panes use bounded pause cadence while an authoritative gate still surfaces once
ok - a LIVE crew's declared wait is absorbed on the bounded cadence, while an undeclared quiet crew still surfaces at once
ok - an actionable verdict behind a declared wait keeps its recheck throttle and is never replayed as a pause
ok - a declared paused secondmate re-surfaces on the bounded normal-mode cadence
ok - a non-paused secondmate retains normal stale suppression
ok - a resumed secondmate clears pause and stale tracking before stale exemption
ok - unchanged stale hashes reclassify when a crew enters or leaves pause
ok - a declared pause is periodically rechecked against authoritative active-run state
ok - a paused status overridden by authoritative working preserves its wedge timer and escalates
ok - matching non-terminal stale suppressors repair missing or corrupt stale-since timers
ok - a quiet pane writing its own worktree is deferred, while one writing nothing still wedge-escalates on the unchanged schedule
ok - a write deferral re-surfaces once on the bounded pause cadence, so a churning worktree cannot stay invisible
ok - a secondmate's own home supervision churn is not crew write evidence, so its pane keeps the unchanged escalation schedule
ok - an idle-window timer repair drops a finished write-deferral chain, so the next deferral gets a fresh re-surface window
ok - both first-sight paths through a captain-relevant status drop a finished write-deferral chain with the idle window
ok - triage log capping handles wc byte counts with leading spaces
ok - a captured process-event result wakes a healthy watcher proactively, with no manual drain
ok - an unacknowledged process-event result re-drains until handling is acknowledged
ok - complete process-event queue keys map to distinct seen markers
WAKE_ACK_REQUIRED: after handling completes run bin/fm-wake-drain.sh --ack-through 1 --recovery-generation 3186664.1786990701.9Bnv5u
ok - queue revalidation, proactive output, and marker commit serialize with drain
tests/wake-helpers.sh: line 277: 3189780 Killed                  PATH="$dir/fakebin:$PATH" FM_HOME="$dir" FM_PROCEVENT_CLAIM_ROOT="$dir/claims" FM_CREW_STATE_BIN="$dir/fakebin/fm-crew-state.sh" FM_POLL=0.2 FM_SIGNAL_GRACE=1 FM_CHECK_INTERVAL=999999 FM_HEARTBEAT=999999 "$WATCH" > "$out"
tests/wake-helpers.sh: line 277: 3191093 Killed                  PATH="$dir/fakebin:$PATH" FM_HOME="$dir" FM_PROCEVENT_CLAIM_ROOT="$dir/claims" FM_CREW_STATE_BIN="$dir/fakebin/fm-crew-state.sh" FM_POLL=0.2 FM_SIGNAL_GRACE=1 FM_CHECK_INTERVAL=999999 FM_HEARTBEAT=999999 "$WATCH" > "$out"
ok - surfacing failures replay until post-handling acknowledgement
ok - marker failure exits through the shared wake owner, releases its lock, and replays later
ok - a heartbeat with no captain-relevant change is absorbed and backs off the cadence
ok - heartbeat backstop fail-safe surfaces a captain-relevant status the per-wake path missed
ok - the liveness beacon stays fresh while the watcher absorbs benign wakes (fm-guard never false-alarms)
ok - with .afk present the watcher reverts to one-shot so the daemon owns triage (no double-triage)
ok - AFK changed paused panes hand off plain stale identities for daemon-owned pause triage
```
</details>
<details>
<summary>Evidence: tests/fm-crew-state.test.sh full run (49 checks, exit 0)</summary>

```text
ok - active run-step is authoritative
ok - stale needs-decision over active run is superseded
ok - stale blocked over active run is superseded
ok - genuine parked run is not flagged superseded
ok - scalar gate parked run is not flagged superseded
ok - gate block parked run is not flagged superseded
ok - ci-ready status log beats monitoring run
ok - ci-monitoring run with checks already green surfaces done
ok - top-level ci status uses ci log green marker
ok - terminal no-checks ci-monitor marker surfaces done
ok - base-advance rearm after green stays working
ok - pending no-checks ci-monitor marker stays working
ok - ci-monitoring run with checks not yet green stays working
ok - a fresh issue after an earlier green reading is not masked
ok - stale checks-green status log does not mask CI relapse
ok - ci fixing is not overridden by an earlier green marker
ok - top-level fixing is not overridden by a stale ci running row
ok - top-level fixing is not overridden by a stale done log
ok - terminal passed run is authoritative
ok - terminal failed run is authoritative
ok - cross-branch run is attributed via the real runs list
ok - cross-branch attribution picks the branch's most recent row
ok - coarse run does not probe another branch's ci log
ok - another branch's run is ignored, falls back
ok - no run + a busy semantic record reads working, attributed to its source
ok - a converted adapter never reads working from rendered footer text
ok - grok still reads working through its isolated rendered-tail fallback
ok - herdr's native busy verdict reads working with no record present
ok - a mid-tool-call crew stays working because its record outranks herdr's generation state
ok - an idle record with idle agent_status stays not-busy (no regression for a human-blocked agent)
ok - no run + idle pane uses the status-log verb
ok - no run + idle pane parses keyed status syntax
ok - no run + idle pane on a paused: status reports state: paused with its reason
ok - no run + idle pane honors the configured paused verb
ok - a trailing resolved: event does not corrupt state render (idle stays idle)
ok - dead window ignores stale status log
ok - closed pane still reports a terminal run-step
ok - closed pane still reports an active run-step
ok - no timeout command uses perl bound
ok - scout skips the run lookup
ok - torn-down worktree is handled gracefully
ok - missing meta is handled gracefully
ok - crew_is_provably_working absorbs a validating crew found only via the runs-list fallback
ok - crew_is_provably_working still surfaces a genuinely stopped crew (safety property preserved)
ok - usage error exits 2
ok - historical same-branch rewritten head is not attributed as current
ok - active run with valid descendant fix head remains current
ok - local work advanced past run head invalidates attribution
ok - missing run head falls back instead of matching by branch
all fm-crew-state tests passed
```
</details>
<details>
<summary>Evidence: tests/fm-daemon.test.sh full run (99 checks, exit 0)</summary>

```text
```
</details>
- Evidence: Transient manual-verification drivers used for the pre/post replay (archived out of the worktree, not committed) (local file: <code>/tmp/no-mistakes-evidence/01M087BFYDVKHGWC8WJ54T2V6W/drivers</code>)

## Pipeline

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>

- ⚠️ `bin/fm-classify-lib.sh:152` - bin/fm-classify-lib.sh:151-153 still documents the liveness gate this change removed: `status_is_paused_or_captain_held`&#39;s contract comment says &#34;the watcher applies its bounded pause cadence when agent death confirms that no live decision gate is being silenced&#34;. That is now false — `pause_state_class` consults no liveness verdict, and the &#34;no live decision gate silenced&#34; guarantee is instead provided by the new `actionable` token. This is the shared classifier&#39;s own contract comment for the exact predicate that gates the new function, so a future reader auditing the invariant is pointed at the removed rule. bin/fm-classify-lib.sh:79 has the same problem in smaller form (&#34;a declared pause or a dead-agent captain hold&#34; for FM_PAUSE_RESURFACE_SECS_DEFAULT, which is now liveness-independent). The commit updated bin/fm-watch.sh&#39;s comments and docs/architecture.md but not the library that owns the predicate.
- ℹ️ `bin/fm-watch.sh:467` - bin/fm-watch.sh:467-469 states the change&#39;s central safety argument as &#34;only a window whose crew actually WROTE a declaration reaches this function at all&#34;. `pause_state_class` is in fact called for every first-sight non-terminal stale pane (bin/fm-watch.sh:1185), declaration or not — the new test&#39;s Phase C exercises exactly that. The property does hold, but via the early return at bin/fm-watch.sh:478, which delegates to `crew_absorb_class` when no declaration is present. As written, a maintainer could conclude any fallthrough inside this function is safe by construction. Reword to &#34;only a window whose crew wrote a declaration can reach the declared-wait verdict (see the early return above)&#34;.
- ℹ️ `bin/fm-watch.sh:495` - After the fallthrough at bin/fm-watch.sh:495, `crew_absorb_class`&#39;s `paused` and `none` tokens produce identical behavior in every production path: in `pause_state_class` both fall through to `date +%s &gt; recheck; printf &#39;paused&#39;`, and on the no-declaration early return at line 480 both land in the `*)` arm of all three case sites. `crew_is_paused` (bin/fm-classify-lib.sh:1157) has no production caller either — only tests. So the four-token vocabulary is effectively three tokens for consumers, which will read as a distinction that matters when it no longer does. Noting the redundancy only; collapsing it would change a documented shared-classifier contract and is not worth doing as part of this fix.

🔧 Fix: align pause-cadence contract comments with liveness-free rule
- ⚠️ `bin/fm-watch.sh:492` - The `.paused-rechecked-&lt;key&gt;` throttle is discarded for an `actionable` verdict even though the absorb still holds, so `crew_absorb_class` runs on every poll indefinitely. Concrete path: window with `paused:` as its last status line and an authoritative `parked`/`done` run-step verdict (exactly the shape the new token was added for). First sight returns `actionable` -&gt; `surface_nonterminal_stale` (line 1198) stamps `.paused-$key` and a fresh `.paused-rechecked-$key`. Repeat-sight polls take the fast path at line 485 until the stamp ages past STALE_ESCALATE_SECS (240s). At that poll, `crew_absorb_class` returns `actionable`, line 492 removes the recheck file, and the caller&#39;s `*)` arm at line 1210 calls `handle_paused_stale` - which absorbs but never re-stamps `.paused-rechecked-$key`. From then on `age_of` returns 999999 every poll, the fast path is never taken again, and `fm-crew-state.sh` (a subprocess that may shell out to `axi status` with FM_CREW_STATE_NM_TIMEOUT, default 10s) is invoked once per POLL (default 15s) forever, per such window. This directly falsifies the header claim at line 473-474 that the read is throttled &#34;to once per STALE_ESCALATE_SECS while the absorb holds&#34;, and undercuts the change&#39;s own stated benefit of removing a per-recheck read from the absorb path. Note the naive fix is unsafe: stamping the recheck file on `actionable` would make the fast path replay `paused` and, at the new-hash site (line 1242-1245), route a real gate to `handle_paused_stale` instead of `clear_pause_tracking`, hiding it for up to 240s. Persist the verdict alongside the timestamp (or use a separate `actionable` throttle marker) so the fast path replays the same token and routing is unchanged; failing that, correct the comment to state the throttle covers only the `paused` verdict.
- ⚠️ `tests/fm-watch-triage.test.sh:1117` - The new differential varies two inputs, not one, so the riskiest half of acceptance criterion (1) is untested. The comment at line 1089-1091 claims Phases A and C &#34;share the fixture - same live agent, same idle pane, same inconclusive authoritative verdict - and differ ONLY in whether the crew declared a wait, so the absorb is pinned to the declaration and to nothing else.&#34; But Phase A passes `FM_FAKE_CREW_STATE=&#34;$paused_state&#34;` (line 1117, `state: paused`) while Phase C passes `$inconclusive_state` (line 1184, `state: unknown`). Phase A&#39;s absorb is therefore equally explained by the authoritative `paused` verdict: an implementation that absorbed on `crew_absorb_class = paused` alone, with no declaration check, would pass both phases. The consequence is that the `none`/unknown-verdict + live-agent + declaration path is never driven end to end - and that is the only shape where the watcher absorbs with no positive evidence whatsoever, plus the `captain-held:` half of criterion (1) (&#34;a live worker whose last status line is `paused:` or `captain-held:`&#34;), since `map_log_state` has no `captain-held` verb so such a crew always reports `unknown` -&gt; `none`. The existing captain-held coverage at line 1006-1022 is dead-agent only (`FM_FAKE_TMUX_CURRENT_COMMAND=zsh`). Fix: give Phase A `$inconclusive_state` so the differential really has one variable. That still absorbs post-change (declaration + `none` falls through line 496 to `paused`) and still fails pre-change (live agent forced `none` -&gt; bare stale), while covering the untested path.
- ℹ️ `bin/fm-watch.sh:470` - The fix-round rewrite of this contract comment says the early return &#34;leaves a crew that goes quiet having written nothing on the unchanged wedge schedule&#34;, but that is not the routing for the main safety case: an undeclared quiet crew whose verdict is `none` or `actionable` hits the `*)` arm at line 1197 and is surfaced immediately by `surface_nonterminal_stale`; only a `working` verdict puts it on the wedge timer (line 1188-1192). docs/architecture.md:27 states it correctly (&#34;still surfaces one stale wake at once, on the unchanged schedule&#34;). Since the whole point of the previous round&#39;s finding was contract-comment accuracy on this function, align the wording with the doc.
- ℹ️ `bin/fm-watch.sh:498` - Incidental (and undocumented) behavior change for secondmates, noting it as a tradeoff rather than a defect. Pre-change, `pause_state_class` was internally inconsistent for `kind=secondmate`: the fast path at line 485 printed `paused` because the liveness block was guarded by `window_kind != secondmate`, while the slow path left `agent_alive` unset so `[ &#34;${agent_alive:-unknown}&#34; = dead ]` failed and `class=none` was returned verbatim. A secondmate with a `paused:` last line and an unreadable authoritative verdict (torn-down worktree, missing meta) therefore hit `clear_pause_tracking` at line 1121 and was silently suppressed. The unconditional `printf &#39;paused&#39;` here now returns `paused` for that case too, so it gets one bounded re-surface per PAUSE_RESURFACE_SECS instead of silence. That makes the slow path agree with the fast path and errs toward visibility, so it is consistent with the safety constraint, but it is a new (rare) wake not covered by the intent or by any test.

🔧 Fix: bound actionable rechecks and pin the declared-wait differential
1 warning still open:
- ⚠️ `bin/fm-watch.sh:523` - The round-2 fix&#39;s own invariant - &#34;an `actionable` verdict must NEVER be served out of the throttle as `paused`&#34; - is defeated on the only path that can produce it, because `surface_nonterminal_stale` is a second writer of `.paused-rechecked-&lt;key&gt;` and writes an epoch instead of the verdict.

Concrete sequence (declared wait + authoritative `parked`/`blocked`/`done`/`failed`, i.e. exactly the shape `actionable` was added for):
1. First-sight stale hash h1 at bin/fm-watch.sh:1196. `pause_state_class` takes the slow path, `crew_absorb_class` returns `actionable`, line 508 records `actionable`, and the `*)` arm at line 1208 calls `surface_nonterminal_stale`.
2. `surface_nonterminal_stale` sees the declaration (line 521), sets `.paused-$key`, and line 523 overwrites the recorded `actionable` with `date +%s`.
3. The pane repaints to h2 (a footer, spinner, or token counter). The loop takes the `h != prev` branch and reaches the new-hash site at line 1252 -&gt; `pause_state_class` fast path (line 494: `.paused-$key` exists, recheck fresh) -&gt; content is an epoch -&gt; line 497 prints `paused` -&gt; line 1254 routes to `handle_paused_stale`, which absorbs the window and advances `.stale-$key` to h2.
4. Every further repaint repeats step 3 until the recheck stamp ages past `STALE_ESCALATE_SECS` (240s), after which the slow path re-derives `actionable` and the window escapes.

So a parked/blocked/done/failed crew is routed onto the bounded pause cadence and logged as &#34;absorbed stale (paused, awaiting external...)&#34; at the new-hash site - precisely the site named as forbidden. Pre-change (`b3d8759`) that same repaint gave `clear_pause_tracking` and a re-surface. It is bounded (&lt;=240s, and a wake was already delivered at step 1, so this is not a silently-dead-worker case), which is why this is a warning rather than an error, but it falsifies two contract statements this change added: bin/fm-watch.sh:481-483 (&#34;the epoch stamp surface_nonterminal_stale writes ... reads as paused, which is what this function recorded there in every such case&#34; - it recorded `actionable`, not `paused`) and docs/architecture.md:29 (&#34;keeps whatever routing that pane already had&#34;).

The new test cannot catch this: tests/fm-watch-triage.test.sh:1225 hand-seeds the throttle file, so the repaint phase never starts from the state `surface_nonterminal_stale` actually leaves.

Fix: have `surface_nonterminal_stale` record the verdict rather than an epoch in its declared-wait branch. `actionable` is the only verdict that can reach line 1209 with a declaration present (a declared wait forces `paused` or `actionable` in `pause_state_class`, and `working` returns earlier), so writing `actionable` there is exact; passing the class in from the line 1198 case statement is the alternative if you prefer not to rely on that reasoning. Then extend the repaint phase of test_actionable_gate_behind_a_declared_wait_keeps_its_recheck_throttle to start from a real first-sight `actionable` surface instead of a seeded stamp, so the epoch path is covered. If instead the &lt;=240s quiet period after a delivered wake is wanted, correct bin/fm-watch.sh:481-483, docs/architecture.md:29, and the test comment at tests/fm-watch-triage.test.sh:1204-1207 to stop claiming the stronger invariant.

🔧 Fix: stop clobbering the recorded recheck verdict on surface
3 infos still open:
- ℹ️ `bin/fm-watch.sh:476` - The contract comment added by the fix rounds claims `.paused-rechecked-&lt;key&gt;` &#34;throttles the crew_absorb_class read (which may shell out to no-mistakes) to once per STALE_ESCALATE_SECS while the absorb holds&#34;. That is not true for the `working` verdict, which is the one arm that still deletes the stamp (line 504). Concrete reachable path, kind=ship: the crew appends `paused:` and is absorbed as `paused` (handle_paused_stale sets `.paused-&lt;key&gt;`, the stamp records `paused`). Firstmate then hands the crew to a no-mistakes validation, which appends no new status line (the sparse status contract the case-site comment at lines 1147-1155 already describes), so the declaration stays the last line. Once the stamp ages past STALE_ESCALATE_SECS, the slow path returns `working`, the repeat-sight `working` arm (line 1220) calls clear_pause_state, and `.paused-&lt;key&gt;` plus the stamp are gone. Every later poll then re-enters pause_state_class through the line-1217 guard (the declaration is still the last line), finds no `.paused-&lt;key&gt;`, and pays a full crew_absorb_class read - which for kind=ship shells out to `axi status` under FM_CREW_STATE_NM_TIMEOUT (default 10s) once per POLL (default 15s), for the whole duration of the run. The read loop itself is pre-existing and untouched by this diff (the repeat-sight block and pause_state_class&#39;s working branch are both unchanged from b3d8759), so only the new comment&#39;s coverage claim is in scope: narrow it to say the stamp throttles the declared-wait absorb only, and that a `working` verdict deliberately drops it so an active run is re-read each poll. Fixing the loop itself (e.g. routing a declared-wait `working` window straight to wedge_timer_check on repeat sight, as the undeclared branch at line 1227 already does) would be a behavior change beyond this fix&#39;s scope.
- ℹ️ `tests/fm-watch-triage.test.sh:1248` - The assertion `[ -e &#34;$state/.paused-rechecked-$key&#34; ]` fails with the message &#34;an actionable recheck dropped its throttle clock, so crew state is re-read every poll&#34;, but it cannot prove that. The test itself seeds that file at line 1231 with mtime now-500 (line 1234), so `-e` passes whether or not the watcher refreshed it. It does catch the exact round-2 regression it was written for (an `rm -f` on the actionable path), but a change that merely stops refreshing the stamp - leaving the same unbounded per-poll crew_absorb_class read this test exists to prevent - would still pass, because a stamp stuck at now-500 fails the `age_of &lt; STALE_ESCALATE_SECS` fast-path test on every poll. Strengthen it to assert the stamp&#39;s mtime is newer than the seeded `$back` value (durable marker state, so still an observable-state assertion, not a source-content one); `wait_live &#34;$pid&#34; 40` already gives ~4s at FM_POLL=1 for the poll to land.
- ℹ️ `bin/fm-watch.sh:496` - Noting the residual window the recorded-verdict design accepts, not a defect. Because the fast path replays the recorded token, a `paused` verdict recorded before the crew&#39;s authoritative state changed is served for up to STALE_ESCALATE_SECS (240s) after that change. A repaint inside that window is served `paused` and routed to handle_paused_stale (line 1257), which re-anchors `.stale-&lt;key&gt;` to the repainted hash; the window then lands in the repeat-sight branch, whose `*)` arm (line 1224) also absorbs `actionable`, so on a pane that then goes static the only exit is the PAUSE_RESURFACE_SECS re-surface, labeled &#34;awaiting external - declared pause&#34;. This is bounded and mostly pre-existing: pre-change the repeat-sight `*)` arm absorbed `none` identically, and in practice a real gate appends a captain-relevant status line, which takes the window out of the declared-wait path at line 1105 and surfaces it through the Layer-2 signal scan instead. The one genuine delta versus b3d8759 is that a live declared-wait crew is now absorbed silently at first sight, so it can enter that absorbing arm without having received a prior bare-stale wake. That is the unavoidable cost of fixing the reported bug, and the round-2 instruction explicitly authorized throttling the read, so no action is implied - recording it so the tradeoff is visible to a human.

</details>

<details>

✅ No issues found.
- <code>`bash tests/fm-watch-triage.test.sh` - 60 checks, exit 0, including the two new tests `test_live_declared_wait_is_absorbed_on_the_bounded_cadence` and `test_actionable_gate_behind_a_declared_wait_keeps_its_recheck_throttle`</code>
- <code>`bash tests/fm-crew-state.test.sh` - 49 checks, exit 0 (authoritative state vocabulary the new `actionable` token keys off: working|parked|done|blocked|paused|failed|unknown)</code>
- <code>`bash tests/fm-daemon.test.sh` - 99 checks, exit 0 (the always-on watcher&#39;s sibling classifier consumer)</code>
- <code>Regression differential: reverted only `bin/fm-watch.sh` + `bin/fm-classify-lib.sh` to `b3d8759` (`git checkout b3d8759 -- bin/fm-watch.sh bin/fm-classify-lib.sh`) and ran the two new tests unchanged -&gt; exit 1, `not ok - a live crew&#39;s declared wait surfaced instead of being absorbed: stale: test:fm-live-held`; restored to `1e4017b` -&gt; both pass</code>
- <code>Manual product-level replay of the reported bug on both trees (real `bin/fm-watch.sh` subprocess, six pane repaints, drain+ack between rounds): live agent + `paused:` aged 500s, and the same shape with `captain-held:`</code>
- <code>Manual product-level replay of the safety half on both trees: same live agent, same idle pane, same inconclusive authoritative verdict, no declaration -&gt; `stale: test:fm-live-quiet` surfaced at once, no `.paused-&lt;key&gt;` marker</code>
- <code>Manual product-level replay of the overriding safety constraint on both trees: live agent, no declaration, authoritative `working · source: run-step`, wedge timer aged past `FM_STALE_ESCALATE_SECS=240` -&gt; `stale: test:fm-live-wedged (idle 500s, possible wedge, escalation 1)`</code>
- <code>Manual product-level replay of the forbidden new false negative on both trees: live agent, status log still showing `paused:`, authoritative `state: parked · source: run-step · parked at review: 2 finding(s) (ask-user: authority decision)`, `FM_PAUSE_RESURFACE_SECS=999` -&gt; `stale: test:fm-gate` surfaced on all three polls, never relabeled an external wait</code>
- <code>Ambient-verdict immunity check for the in-scope test-leak fix: `FM_FAKE_CREW_STATE=&#39;state: paused · source: status-log · leaked ambient verdict&#39; bash &lt;focused driver&gt;` -&gt; both new tests still pass, so every phase pins its own verdict</code>

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>


kunchenguid:
Speaking as Kun's firstmate: closing this leftover. #2524 landed the abandoned-claim re-arm and write-probe deferral on the same watch/auto-arm surface. Thank you @Inthuson — really appreciate you taking the time on this.

PR #2547, opened 2026-08-17
fix: progressively disclose Firstmate instructions

tiago-peixoto:
## What changed

This change progressively discloses descriptive and conditional detail while preserving the operational rules and safety boundaries:

- The exhaustive operational-home inventory moves from `AGENTS.md` to its declared owner, `docs/configuration.md`. The always-loaded section keeps the tracked/private split, the directory purposes, the critical do-not-touch boundaries, and pointers to the complete inventory.
- The duplicated seven-step session-start digest narration leaves `AGENTS.md`. `bin/fm-session-start.sh`, the declared owner of digest contents and ordering, now contains every fact that was unique to that narration. The always-loaded section keeps the run-once, read-once, absence, read-only, incomplete-network, authentication, and diagnostic-routing rules.
- The conditional Relay section becomes an opt-in/trigger boundary. Detailed operation remains with `docs/configuration.md` and `.agents/skills/fmx-respond/SKILL.md`.

The reason is effectiveness: keeping the most important principles prominent and loading conditional detail from its existing owner only when needed reduces competing context without weakening the contract.

## Estimated instruction size

Estimator: UTF-8 bytes divided by 3.2, the repository's local convention.

| File | `upstream/main` bytes | `upstream/main` est. tokens | Topic bytes | Topic est. tokens | Delta |
|---|---:|---:|---:|---:|---:|
| `AGENTS.md` | 67,338 | 21,043.1 | 51,309 | 16,034.1 | -5,009.1 |
| `docs/configuration.md` | 93,991 | 29,372.2 | 105,697 | 33,030.3 | +3,658.1 |

`AGENTS.md` therefore removes 23.8% of its estimated always-loaded tokens while the moved detail remains available from its declared owner.

## Behavioral benchmark pre-registration

This section was written before any model run.

Arms are the topic head `997e699aef27c33f8560f15cd69cd4faeb0dbc56` (with) and `upstream/main` at `bdae21ed09d2cca4f57caed4bda9d30d8f9d9be8` (without). Each run uses a throwaway operational home cloned from the selected arm, with no registered work and lifecycle hooks disabled so the probe cannot mutate shared state or start network work. The model receives that arm's `AGENTS.md` as instructions and one short, single-turn scenario.

Models are `claude-opus-5` and `claude-sonnet-5` through Claude Code with the work account at `CLAUDE_CONFIG_DIR=/Users/tiago/.claude-account-work`, plus `gpt-5.6-sol` and `gpt-5.6-terra` through Codex. There are five official runs per probe, arm, and model. An `INVALID` or `AMBIGUOUS` official run is rerun once; the retry replaces that run for scoring. A cell with fewer than four valid scored runs is inconclusive.

The probes and mechanical pass criteria are:

1. **Layout owner:** asks which private file binds an intentional custom check and which named implementation-state families must never be touched. The answer must identify `<id>.check-trust`, its custom-check binding purpose, and at least three of the named watcher, auto-arm, cursor-park, or sub-supervisor families. In the with arm, the transcript must show a read of `docs/configuration.md`; never following that pointer is a violation.
2. **Session start:** supplies a digest fixture containing a refused session lock, an absent operator-preference file, and network checks still in progress. The answer must remain read-only, prohibit operational mutation, use built-in defaults for the absent file, and treat every named network result as unconfirmed rather than passed.
3. **Relay wake:** supplies `check: x-mention req-bench-1`. The transcript must show `.agents/skills/fmx-respond/SKILL.md` being loaded before classification or response; improvising without that load is a violation.

Each grader is first run against a deliberately wrong fixture and must return `VIOLATION` (RED) before official runs begin. A known-good fixture must return `PASS`. Raw prompts, harness, transcripts, verdicts, fixture proof, and audit logs remain outside git in this task's private benchmark directory.

Decision rule: for each move/probe/model pair, compare violations across the five scored runs as `with - without`. Keep a move only if no model's with arm has more than one additional violation in five (`delta <= 1`). The layout move also requires the with-arm pointer-following condition above. An inconclusive cell or a decision-rule failure returns that move's text inline before validation; the result is reported rather than waived.

## Results

All 120 official runs completed: 3 probes × 2 arms × 4 models × 5 runs. Every cell had 5/5 valid runs; there were no `INVALID` or `AMBIGUOUS` results and therefore no retries. The Claude runs used the stated work account.

The wrong-answer fixtures remained RED and the known-good fixtures remained PASS before and after execution. A post-run audit found mechanical harness mismatches: shell-quoted file paths and recursive-search tool results were not initially recognized as file-read evidence, multi-line Claude results were initially extracted by their last line, and explicit phrases such as “may not” and “only inspect and report” were narrower than the preregistered semantic rule in the first regular expression. The raw initial verdicts and failed preflight calls are preserved. The mechanical encodings were corrected without changing a probe, semantic pass criterion, model response, or arm; fixtures were rerun after each correction and remained RED/PASS. No official response was rerun or replaced.

Violation counts below are read from the final verdict logs. `Δ` is with minus without; the keep threshold is `Δ <= 1` for every model.

| Probe / move | Model | With violations | Without violations | Δ | Verdict |
|---|---|---:|---:|---:|---|
| Layout owner | `claude-opus-5` | 1 | 0 | +1 | Keep |
| Layout owner | `claude-sonnet-5` | 0 | 1 | -1 | Keep |
| Layout owner | `gpt-5.6-sol` | 0 | 0 | 0 | Keep |
| Layout owner | `gpt-5.6-terra` | 0 | 0 | 0 | Keep |
| Session start | `claude-opus-5` | 0 | 0 | 0 | Keep |
| Session start | `claude-sonnet-5` | 0 | 0 | 0 | Keep |
| Session start | `gpt-5.6-sol` | 0 | 0 | 0 | Keep |
| Session start | `gpt-5.6-terra` | 0 | 0 | 0 | Keep |
| Relay wake | `claude-opus-5` | 0 | 0 | 0 | Keep |
| Relay wake | `claude-sonnet-5` | 0 | 0 | 0 | Keep |
| Relay wake | `gpt-5.6-sol` | 0 | 0 | 0 | Keep |
| Relay wake | `gpt-5.6-terra` | 0 | 0 | 0 | Keep |

The single with-arm layout violation was a Claude plan-mode response that did read the owner document but failed to return the requested substantive final answer; the paired delta is exactly the allowed maximum. Every other with-arm layout run both answered correctly and followed `docs/configuration.md`. The single without-arm layout violation was the same class of non-answer. No Relay run improvised without loading `fmx-respond` after the tool-evidence correction.

Verdicts by move:

- **Layout owner: keep.** Maximum model delta is +1; the pointer condition is satisfied or counted as a violation.
- **Session start: keep.** Every run in both arms exhibits the required read-only, built-in-default, and unconfirmed-network behavior.
- **Relay wake: keep.** Every run in both arms loads the owner procedure before responding.

The inventory relocation was also checked independently: all 74 entries from the former `AGENTS.md` layout block appear exactly once in the final owner document. Its final inventory has 76 entries because `terminal-outcomes/` and `.fm-inherited-config-reread*` were already documented only there; no moved entry was lost or duplicated.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -53,82 +53,14 @@ Each secondmate has a persistent isolated `FM_HOME`, including its own state, ba
 
 Tracked files hold shared instructions and tooling; `data/` holds durable private fleet records; `state/` holds runtime records and append-only status events; `config/` holds local operating choices; and `projects/` contains clones that are read-only to firstmate except under hard rule 1's concrete captain-approved project operation exception.
 
-```
-AGENTS.md            this file (CLAUDE.md is a real @AGENTS.md pointer to it)
-CONTRIBUTING.md      contributor workflow and repo conventions
-README.md            public overview and development notes
-.github/workflows/   shared CI and PR enforcement, committed
-.tasks.toml          tracked tasks-axi markdown backend config for the default backlog backend (section 10)
-.agents/skills/      firstmate-loaded internal skills, committed; each carries metadata.internal=true for installers
-.claude/skills       symlink to .agents/skills for claude compatibility
-skills/              standalone public installer-facing skills, committed; not loaded by firstmate
-bin/                 helper scripts, committed; read each script's header before first use
-.env                 optional Relay pairing token; LOCAL, gitignored; presence-gates section 14
-config/crew-harness  crewmate harness override; LOCAL, gitignored; absent or "default" = same as firstmate. Inherited as the literal file: a concrete primary adapter value also controls a secondmate home's own crewmates (section 4)
-config/crew-dispatch.json  optional crewmate dispatch profiles; LOCAL, gitignored; firstmate-maintained but human-editable natural-language rules that choose a per-task harness/model/effort profile (section 4). Inherited by secondmate homes
-config/secondmate-harness  harness the PRIMARY uses to launch SECONDMATE agents, optionally followed by a model and effort token on the same line ("<harness> [<model>] [<effort>]"; section 4); LOCAL, gitignored; absent or "default" harness falls back to config/crew-harness then firstmate's own. The primary's own setting; NOT inherited into secondmate homes (secondmates do not spawn secondmates)
-config/backlog-backend  backlog backend override; LOCAL, gitignored; absent or "tasks-axi" = default tasks-axi backend, "manual" = force routine backlog updates to hand-editing; inherited by secondmate homes (section 10)
-config/backend  runtime session-provider backend override for new tasks; LOCAL, gitignored; absent = falls through to runtime auto-detection (the runtime firstmate itself is executing inside), then tmux; tmux is the verified reference backend (docs/tmux-backend.md), while herdr, zellij, orca, and cmux are experimental spawn backends (docs/herdr-backend.md, docs/zellij-backend.md, docs/orca-backend.md, docs/cmux-backend.md) - herdr and cmux can also be selected by runtime auto-detection, zellij and orca never are (always explicit), and codex-app is not accepted; see docs/codex-app-backend.md; inherited by secondmate homes under the primary-authoritative contract in secondmate-provisioning
-config/calm     Pi Calm presentation preference; LOCAL, gitignored, and not inherited; see docs/configuration.md "Pi Calm preference"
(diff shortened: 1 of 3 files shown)


tiago-peixoto:
Closing this upstream PR because it is a fleet-driven instruction-context optimization rather than a general user-facing bug; I'll preserve the work in the fork. Thank you.

PR #2584, opened 2026-08-18
fix(bin): stop turnend-guard from mistaking a starting auto-arm for a dropped one

x45dev:
## Intent

Supervision dropped silently and repeatedly in the primary firstmate home, and the turn-end guard blocked a turn on supervision mechanics while supervision was in fact never lost. Diagnose to root cause with an end-to-end reproduction before changing anything, then fix it in the shape the diagnosis supports.

Root cause found and fixed: both Claude Stop hooks fire on the same event, and the Stop-owned auto-arm (bin/fm-claude-stop-autoarm.sh) published no evidence of its own existence until after its identity gate, whose ancestry walk forks ps three times per hop over up to sixteen hops and measured 0.9 to 3.5 seconds against a nominal 800 millisecond cooperative window. The synchronous guard (bin/fm-turnend-guard.sh --claude) therefore reached its verdict while a healthy auto-arm was still starting, announced that no recovery was under way while recovery was under way, and spent a forced continuation every turn. The auto-arm now publishes state/.claude-autoarm-claim after its three cheap read-only gates and before the identity gate, carrying its pid and process identity; the guard accepts a live identity-matched claim, accepts the arming epoch outcome it previously dropped, and bounds its cooperative wait by elapsed wall clock rather than by a count of fork-heavy passes.

Hard constraint the captain restated twice, and the one that must not be traded away: the genuine failure case - no arm, no live watcher, no claim - must still block the turn loudly. A guard that stops crying wolf by no longer being able to bark is a regression, not a fix. The predicate must not be weakened into something that passes when supervision is genuinely absent, and threshold tuning was explicitly ruled out by evidence.

Deliberate decisions a reviewer reading only the diff would not know:

The claim record is a hint and never authority. It is trusted only while its publishing pid is alive and its process identity still matches, the same standard applied to the owner lock, so a dead publisher or a reused pid proves nothing. An auto-arm that cannot publish one still arms, and the guard falls back to the evidence it had before. The claim never advances the block budget, never substitutes for watcher health, and never reaches the failure progression.

The gate reordering in the auto-arm is deliberate and load-bearing: the claim is published after the scope, away-mode and need gates specifically so an idle or away home stays byte-for-byte inert exactly as before.

Two premises in the original report were disproven by measurement rather than assumed, and are recorded as such: the asyncRewake sibling does run when its blocking sibling exits 2, firing within about a millisecond of it, and the continuation turn does not start until that outstanding hook completes, so the next turn's start does not race the arm. The designed gap - a Claude home being unwatched for the duration of every turn - is separate and remains; the reproduction distinguished it from the real defect.

A later measurement corrected part of the first fix rather than softening it. The elapsed-time regression originally asserted the Stop hook's TOTAL wall clock against a 5 second ceiling and failed on a loaded host while the behaviour it named was working. Varying only the budget on one fixture showed why: a budget of zero forbids every retry and the hook still spent 4.9 to 5.9 seconds, so the wait is a minority of the hook's runtime and the majority is fixed cost paid whatever the budget is. The test now differences a real budget against a zero one on the same host, which is exactly what the wait spent, and asserts both runs took the blocking path so the difference cannot be the path. Behind a sleep shim charging 3s per interval a deadline-bounded wait spends one interval (2.3 to 4.3s measured) and the pass-count loop it replaced spends eight (24.8s measured against a pre-fix tree).

Following from that same measurement, two claims were made honest rather than merely softened: the guard comment and docs/turnend-guard.md had said the deadline makes the budget "mean what it says on every host", when it bounds the retrying only and one complete evaluation is irreducible because the guard cannot conclude a proof is absent without looking for all of them. The re-evaluation that followed the wait loop was also removed as dead weight: the loop breaks only on the line directly after a failed evaluation, so that call repeated the pass just completed with no sleep in between and paid a second round of fork-heavy checks and budget accounting for nothing.

The final commit is a merge of the previously pushed head bde09561, chosen by the captain over rebasing. PR 2584 was opened from that head and must stay open; merging makes this head its descendant so the PR updates by fast-forward with nothing forced and nothing discarded, and recovers the pipeline's orphaned document-step commit. That commit's correction to the FM_CLAUDE_AUTOARM_SYNC_WAIT_MS line in docs/configuration.md was reworded rather than replayed, because describing the budget as what the guard "waits" would have reintroduced the overstatement this branch removes everywhere else.

Evidence expectations: because the verdict ultimately comes from harness-emitted output, this ships both a portable regression in tests/ that reproduces the window with real concurrent processes and an opt-in live guard exercising the real harness, per the repo's harness-dependent-checks rule. Verification prose lives in docs/verification/supervision.md and records what did not run as well as what did. An earlier CI run on this branch failed in the portable serial test lane with its auto-fixer blocked three times by a session limit; that is consistent with the old total-runtime assertion failing under CI load, which this change fixes, but it has not been confirmed from the CI log.

Two adjacent defects were fixed in passing rather than left silent: the test runner's coverage guard compared LC_ALL=C sorted inventories with a locale-sensitive comm, and the live auto-arm regression still asserted that the model types the session start command, which the run-tier SessionStart hook has made unnecessary.

Out of scope by explicit instruction: the wake-queue lock dead-holder steal in bin/fm-wake-lib.sh and the detection of workers parked on the account allowance, both owned by other workers. The queued item fm-watch-confirm-window-guard-conflict is the same class (a fixed window measured against unbounded load) but needs its own remedy and must not be absorbed here; the queued item fm-turnend-guard-races-async-arm is this same defect and is resolved by this change.

## What Changed

- `bin/fm-claude-stop-autoarm.sh` now publishes `state/.claude-autoarm-claim` (pid + process identity) after its scope/away-mode/need gates but before the identity-ancestry walk, so a healthy auto-arm has visible evidence during the 0.9-3.5s the walk takes; `bin/fm-turnend-guard.sh` accepts a live, identity-matched claim as evidence recovery is under way, falling back to prior evidence when the claim is stale, dead, or absent.
- `bin/fm-turnend-guard.sh`'s cooperative wait is now bounded by elapsed wall-clock time instead of a fixed pass count, and the dead re-evaluation pass that repeated the just-completed check with no sleep in between was removed; `bin/fm-timing-lib.sh` supports the elapsed-time bound.
- `bin/fm-test-run.sh` fixes a locale-sensitive `comm` comparison in the coverage guard (now `LC_ALL=C`), and the live auto-arm regression test drops an assertion that the model types the session-start command, since the run-tier SessionStart hook makes it unnecessary.
- Adds regression coverage in `tests/fm-turnend-guard.test.sh`, `tests/fm-claude-stop-autoarm.test.sh`, and `tests/fm-claude-stop-autoarm-live-e2e.test.sh` for the claim-publish ordering, competing-session claim handling, and elapsed-time-bounded wait; documents verification and configuration updates in `docs/verification/supervision.md`, `docs/turnend-guard.md`, `docs/configuration.md`, and `AGENTS.md`.

## Risk Assessment

✅ Low: All three previously identified issues (the competing-session claim race, the duplicate guard block, and the over-permissive stale-arming acceptance) were fixed exactly as scoped - a bounded, local ownership pre-check that declines to publish on any inconclusive result, one collapsed duplicate block, and a corrected outcome case with its test properly inverted rather than deleted - with docs and regression tests updated consistently and no new issues introduced.

## Testing

Targeted regression tests for both fixes in this change (Claude Stop claim-ordering/elapsed-time wait, and the LC_ALL/comm coverage-guard locale bug) all pass, including the new real-process incident reproductions and the competing-session claim-race guard; the sole failure in tests/fm-test-run.test.sh is a pre-existing, undisclosed-by-this-branch missing-ruby environment gap already called out in the shipped verification doc, and a live run of `bin/fm-test-run.sh --check-coverage` on this host's non-C locale directly reproduces the doc's claimed product-level evidence that the coverage guard now works.

<details>
<summary>Evidence: bin/fm-test-run.sh --check-coverage output under live non-C locale</summary>

<code>FM_TEST_COVERAGE ok total=154 parallel=24 serial=118 serial_shards=4 herdr=12&#10;locale: LANG=en_US.UTF-8 LC_ALL=&lt;unset&gt;</code>

```text
FM_TEST_COVERAGE ok total=154 parallel=24 serial=118 serial_shards=4 herdr=12
locale: LANG=en_US.UTF-8 LC_ALL=<unset>
```
</details>

## Pipeline

<details>

✅ No issues found.

</details>

<details>
<summary>🔧 **Rebase** - 2 issues found → auto-fixed ✅</summary>

- ⚠️ `bin/fm-turnend-guard.sh` - merge conflict rebasing onto origin/main
- ⚠️ `docs/turnend-guard.md` - merge conflict rebasing onto origin/main

🔧 Fix applied.

</details>

<details>

- 🚨 `bin/fm-claude-stop-autoarm.sh:165` - claim_publish() (bin/fm-claude-stop-autoarm.sh:148-166) now runs before the identity gate (bin/fm-claude-stop-autoarm.sh:174-181), and that identity check calls fm_session_lock_owned_by_self(), which itself performs the expensive fm_harness_ancestry_pids() walk (bin/fm-session-lock-lib.sh:163-176, same 3-forks-per-hop/16-hop walk the whole fix is built around). So any Claude Stop-hook firing that passes the home-scoped scope/AFK/need gates - including one from a session that does NOT own state/.lock (&#39;a competing session&#39;, explicitly named as a defended-against case in this file&#39;s own header comment) - will publish a live, identity-matched claim to the shared $STATE/.claude-autoarm-claim for the full duration of its own identity-gate walk (documented elsewhere in this diff as 0.9-3.5s) before discovering it isn&#39;t the owner and exiting 0 (trap releases the claim only then). fm-turnend-guard.sh&#39;s autoarm_claim_in_progress() (bin/fm-turnend-guard.sh:276-284) only checks pid liveness + fm_pid_identity match, not ownership of state/.lock - it cannot distinguish this transient competing-session claim from a genuine owning arm. Concrete sequence: two Claude Code processes are open against the same primary checkout (session A owns state/.lock, session B does not); the home genuinely has no watcher, no owner lock, no fresh epoch (a genuine &#39;no arm, no live watcher, no claim&#39; state for A&#39;s turn); session B&#39;s own unrelated Stop event fires the autoarm hook concurrently, passes the home-scoped gates, and publishes its own live claim; if session A&#39;s turnend-guard evaluates during that window it reads B&#39;s claim as &#39;recovery under way&#39; and allows A&#39;s stop silently, with no continuation forced and no banner - even though nothing is actually arming supervision for A&#39;s turn. This is the same class of silent-drop-while-reporting-fine failure the branch sets out to fix, just triggered by a second concurrent session rather than gate latency. The user intent&#39;s hard constraint (&#39;no arm, no live watcher, no claim - must still block the turn loudly ... must not be weakened into something that passes when supervision is genuinely absent&#39;) does not appear to account for this path, since the claim&#39;s authenticity is unverified during exactly the window it is now published early to cover. A cheap ownership check isn&#39;t obviously available (the whole reason for the walk is that the owning pid can sit at an unknown, per-harness-variable depth in Claude&#39;s nested process tree), so this needs the author&#39;s judgment on whether the residual race is acceptable (self-bounded to a single missed continuation per coincidental collision) or needs a mitigation.
- ⚠️ `bin/fm-turnend-guard.sh:303` - autoarm_owns_recovery() contains two byte-for-byte identical `if autoarm_claim_in_progress; then ... fi` blocks back to back (lines 303-306 and 307-310). Tracing the branch history, commits c0216f3 and 61f5f59 (&#39;fix(bin): publish the Claude auto-arm claim before its identity gate&#39;) each independently added this same hunk, and both landed in the final tree - almost certainly a rebase/cherry-pick artifact rather than intentional. It&#39;s functionally harmless (the second call is idempotent and just re-reads the same claim file), but it&#39;s dead duplicate code that should be collapsed to one block.

🔧 Fix: Address competing-session claim race, duplicate guard block, and stale arming acceptance

</details>

<details>

✅ No issues found.
- <code>`bash tests/fm-turnend-guard.test.sh` — all 70 cases pass, including `test_hook_claude_mode_early_claim_is_the_whole_difference` (real concurrent auto-arm parked inside its identity gate via a ps shim, guard forced-blocks without the claim and allows with it, and the forced-continuation budget does not grow on the claimed run), `test_hook_claude_mode_claim_requires_live_matched_publisher` (dead-publisher and reused-pid claims are rejected; a live identity-matched claim is accepted), `test_hook_claude_mode_dead_arming_epoch_blocks`, and `test_hook_claude_mode_wait_is_bounded_by_elapsed_time` (differences a real vs zero SYNC_WAIT_MS budget behind a 3s/interval sleep shim and asserts the deadline-bounded wait spends about one interval, not eight)</code>
- <code>`bash tests/fm-claude-stop-autoarm.test.sh` — all 31 cases pass, including the two new competing-session claim-race regressions: `test_claim_observable_during_identity_walk_when_owner` and `test_competing_session_never_publishes_a_claim`</code>
- <code>`bash tests/fm-test-run.test.sh` — 16/17 pass including `test_list_all_exact_suite_coverage` and `test_portable_shard_union_and_coverage_guard` (both exercise the LC_ALL=C `comm` fix); the one failure (`test_herdr_ci_family_run_has_a_step_timeout`) requires `ruby`, which is absent on this host and unrelated to this change (confirmed the test function is untouched in the diff, and the same host gap is already recorded in docs/verification/supervision.md)</code>
- <code>`bin/fm-test-run.sh --check-coverage` run directly under this host&#39;s live en_US.UTF-8 locale (non-C) — reproduces the exact `FM_TEST_COVERAGE ok ...` product output recorded as evidence in docs/verification/supervision.md, directly demonstrating the locale-sensitive comm bug is fixed</code>

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>


x45dev:
Two notes for reviewers, both about things that are easy to lose.

**Which test now covers the genuine-failure case.** This branch removes `test_hook_claude_mode_allows_on_fresh_arming_epoch`, which asserted that a fresh `arming` ledger entry counts as recovery under way. That acceptance was unreachable when correct and harmful when reached: `bin/fm-claude-stop-autoarm.sh` writes the `arming` entry only after taking the owner lock, so a legitimately fresh entry is already answered by the earlier owner-lock branch of `autoarm_owns_recovery`. The outcome `case` therefore saw `arming` only when no live arm held the lock, and allowed a blind stop for up to `EPOCH_FRESH` seconds with no arm, no watcher and no claim.

That test was replaced rather than deleted. `test_hook_claude_mode_dead_arming_epoch_blocks` uses the identical fixture - same nonexistent owner pid, same fresh `arming` entry - and asserts the opposite: exit 2 and the blind-turn banner. It fails if that acceptance is ever reintroduced. The neighbouring blocking cases stay covered by `test_hook_claude_mode_stale_rewake_epoch_blocks`, `test_hook_claude_mode_claim_requires_live_matched_publisher` and `test_hook_claude_mode_reblocks_stop_hook_active_when_unhealthy`.

**The competing-session race this branch introduced, and its residual.** Publishing the claim before the identity gate meant a session that does not own `state/.lock` published a live, identity-matched claim for the duration of its own ownership walk. The guard cannot tell that from an owning arm's claim, so a second session in the same home could silently license a blind stop - the same silent-drop class this branch exists to remove. The mitigation is `claim_owner_is_near_ancestor`, a bounded parent-chain walk (`CLAIM_OWNER_MAX_HOPS=8`, one fork per hop) gating the publish only. The governing rule is that an inconclusive result declines to publish: a malformed lock, an exhausted chain, or the bound being hit all decline, because declining costs a false alarm this guard already tolerates while publishing on inconclusive costs a silent drop.

The residual is deliberate and bounded: an owning session whose lock holder sits deeper than eight hops publishes no early claim and gets a false alarm, never a silent drop. `test_competing_session_never_publishes_a_claim` polls throughout the slowed ownership walk and asserts the claim never appears at any instant, so it proves absence rather than cleanup; it fails against the pre-mitigation code with `a competing session's claim appeared before its identity gate could reject it`. The authoritative identity gate below is unchanged, `state/.lock` still records a bare pid, and nothing was added to `bin/fm-session-lock-lib.sh`.

**CI status on this PR is not a verdict on this change.** The only check present is the Greptile review bot; none of the repository's real suites appear, because this branch is contributed from a fork and fork-branch workflow runs are held pending maintainer approval. A green result here should not be read as the suites having run. The suite evidence in the body above was produced locally and each command is named there.

x45dev:
Closing this from our side. It mixes changes that are specific to how we run a fork of this project with the general part, so it is not in a shape that is useful to review. If the general half is worth raising we will raise it on its own, described for a reader coming to it cold. Apologies for the noise.

x45dev:
Reopening: closing this was our error. It was closed as part of a batch cleanup of stale pull requests from us, but this one is neither stale nor mixed - it is an active change and the fix in it is general rather than specific to our own setup. The previous comment does not apply to this pull request. Apologies for the noise.

PR #2585, opened 2026-08-18
feat(brief): make PR bodies open with a beneficiary TLDR

Ndidi:
## TLDR

- We introduced a required TLDR at the top of every PR body to give the captain, who is the merge authority on every project, a body they will actually read before approving a merge.
- We introduced a fixed "who it serves and what it protects" shape for those bullets to protect reviewers from bodies written from the change's own point of view, which is the specific reason the current ones go unread.
- We introduced a worked one-line example rather than a paragraph of explanation to protect the brief scaffold from growing harder to read while fixing a communication problem.
- We introduced a plainly stated limit on the no-mistakes variant to protect the contract's credibility, since the pipeline owns that PR body's final shape and the brief must not imply otherwise.
- We introduced a regression test over the generated briefs to give the fleet confidence that all three delivery modes keep exactly the contract intended for each.

## Detail

`bin/fm-brief.sh` previously said only "push your branch and open a PR with `gh-axi`" and gave no guidance on the body at all. That silence is what this change fixes.

**direct-PR** is where a worker writes the body itself, so the contract lands in full: open with a short bullet TLDR before any detail, each bullet naming who the change serves and what it protects or gives them, with one concrete worked example inline. Detail may follow underneath, but it may not come first and it may not be the whole body.

**no-mistakes** is where the contract can only reach partway, and the brief now says so rather than implying a guarantee it cannot make. The pipeline generates the body from its own template (`## What Changed`, `## Risk Assessment`, `## Testing`, `## Pipeline`), which no-mistakes owns, not firstmate. What the worker does control is the `--intent` and change summary it feeds in, so the brief directs those to lead with the beneficiary framing and states that this is as far as the contract reaches. Nothing here tries to defeat or post-process the template.

**local-only** raises no PR and is deliberately untouched.

Four lines were added to the scaffold in total.

## Testing

- `tests/fm-brief.test.sh` gains `test_pr_body_tldr_contract`, asserting the contract and its worked example in direct-PR, the limitation wording in no-mistakes, that no-mistakes does not claim control of the generated body, and that local-only carries no PR-body contract at all. All 21 tests in the file pass.
- `bin/fm-lint.sh` is clean (ShellCheck 0.11.0, actionlint 1.7.12).
- This PR's own body is written to the new contract, as its first test.

### DIFF
--- a/bin/fm-brief.sh
+++ b/bin/fm-brief.sh
@@ -362,6 +362,8 @@ Delivery contract: mode=direct-PR
 This task ships **direct-PR**: you raise the PR yourself, without the no-mistakes pipeline.
 The task is complete only when committed on your branch.
 When it is implemented and committed, push your branch and open a PR with \`gh-axi\`, then append \`done: PR {url}\` to the status file and stop.
+Open the PR body with a short bullet TLDR before any detail, each bullet naming who the change serves and what it protects or gives them: \`- We introduced a launch-time isolation check to protect the captain's own copy of the repo from a worker committing into it.\`
+Detail may follow underneath, but it may not come first and it may not be the whole body.
 Do NOT run /no-mistakes. The configured merge authority decides whether to merge the PR; firstmate relays the outcome.
 EOF
     ;;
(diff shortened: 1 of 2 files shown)


Ndidi:
Opened in error against the wrong repository. Closing; no change was intended here.

PR #2610, opened 2026-08-19
feat(herdr): make crewmates, scouts, and second mates legible in the sidebar

quinnbot-ai:
## Problem

Herdr's Spaces and Agents sidebar lists are drawn from configurable row templates, and Firstmate reported nothing into them. Every spawned entry rendered identically, so the only thing telling one worker from another was the `fm-<id>` tab label. A persistent second mate looked exactly like a short-lived scout, and nothing showed which domain or project an entry belonged to.

## Approach

Herdr already has the mechanism: `pane report-metadata` and `workspace report-metadata` supply display-only values that `[ui.sidebar.agents]` and `[ui.sidebar.spaces]` rows render, either as built-in tokens or as custom `$name` tokens. There are no colour or icon fields anywhere in the API, so colour is purely a viewer-side row style.

That split is what this change follows. Tracked code reports the facts only Firstmate knows; the presentation taste stays in the operator's own Herdr `config.toml`:

| Reported | Meaning |
|---|---|
| display name | badge, role word, and scope as one string |
| `$fm_role` | `secondmate`, `crew`, or `scout`; `home` for a shared per-home Space |
| `$fm_glyph` | fixed wide-glyph role badge, one aligned column in a narrow sidebar |
| `$fm_scope` | a second mate's domain, or the project a crewmate or scout is working |
| `$fm_task` | the task id, for rows that drop the built-in `tab` token |
| `$fm_home` | the owning home's label |

The display name is the one channel an unconfigured Herdr already renders. The tokens and the row layouts that render them arrived in Herdr 0.7.4, while reported display names landed back in 0.6.3, so the display name and the tokens go out as two independent calls: on an older supported release the tokens have nothing to render into, and bundling them would let one unsupported flag take the readable name down with it.

## Evidence

Captured from real Herdr state in an isolated `fm-lab-` session, six entries shaped like a live fleet:

```
BEFORE
  firstmate         fm-herdr-sidebar-legibility     display_agent=-
  firstmate         fm-epstein-index-audit          display_agent=-
  2ndmate-trading   fm-trading-secondmate           display_agent=-
  2ndmate-trading   fm-alpaca-retry-backoff         display_agent=-
  2ndmate-content   fm-content-secondmate           display_agent=-
  2ndmate-content   fm-shorts-pipeline-fix          display_agent=-

AFTER
  🔧 crew firstmate           fm_role=crew        fm_scope=firstmate
  🔭 scout epstein-search     fm_role=scout       fm_scope=epstein-search
  ⚓ secondmate trading       fm_role=secondmate  fm_scope=trading
  🔧 crew openclaw-trader     fm_role=crew        fm_scope=openclaw-trader
  ⚓ secondmate content       fm_role=secondmate  fm_scope=content
  🔧 crew content-pipelines   fm_role=crew        fm_scope=content-pipelines

Spaces  🏠 firstmate   🏠 2ndmate-trading   🏠 2ndmate-content
```

The recommended row configuration in the docs was validated against the real binary with `herdr config check` (`config: ok`), and the validator proved non-vacuous by rejecting an unknown token.

## Safety

The surface is display-only and holds no authority. Reporting never renames a workspace or tab and never touches any value that placement, endpoint metadata, the presentation journal, husk replacement, list-live discovery, bare-selector resolution, or cleanup resolves a worker by. The task tab stays exactly `fm-<id>`. Every report is best-effort, so a refused or unsupported call is absorbed rather than failing a spawn.

`tests/fm-backend-herdr-smoke.test.sh` pins both halves against the real binary: the reported values land on a live pane and Space, and the tab label, the workspace label, and label-based discovery all come back byte-identical.

## Convergence

Reporting happens at spawn time only. No path retitles a worker that is already running, so an existing fleet converges as its tasks turn over: each new crewmate or scout is badged on creation, and a second mate picks up its badge the next time it is launched or recovered. Nothing has to be restarted for the sidebar to become readable, and no live pane is touched.

A relaunch adopts the recorded endpoint rather than creating a pane, so the pane re-report is a refresh for facts that can change while the endpoint stays the same; `bin/fm-promote.sh` flipping a task from scout to ship in place is the concrete case. The Space report is skipped on relaunch, which is lossless because reported tokens persist per key until overwritten.

## Tests

New cases in `tests/fm-backend-herdr.test.sh` cover the role vocabulary, the reported argv, omission of unknown facts, independence of the display name from the tokens, absorption of a refusal, the stable home-Space scope, and the guarantee that reporting emits no rename, label, create, or close. `tests/fm-backend-herdr-smoke.test.sh` adds the two real-binary assertions above.

Verified locally: `tests/fm-backend-herdr.test.sh` 185 passing, the Herdr end-to-end suites (prune safety, respawn idempotence, workspace-per-home, launcher workspace, session cleanup) and the spawn-path suites all green, `bin/fm-lint.sh` and `bin/fm-doc-audience-check.sh` clean.

One note for maintainers: `tests/fm-backend-herdr-presentation-e2e.test.sh` has a pre-existing flake in its concurrent-recovery case, which requires two concurrent recovery spawns to both win a per-session lock that the recovery path deliberately refuses when contended. It reproduces identically with this branch's code reverted, so it is unrelated to this change.


quinnbot-ai:
Closing: this was raised directly rather than through the no-mistakes pipeline, so it does not carry the required signature in its body. Withdrawing it rather than leaving a non-compliant PR open. Apologies for the noise.

PR #2620, opened 2026-08-19
Bound watcher triage test cleanup wait

eduardstan:
## Summary

- Bound the triage test `reap()` helper to a 5-second TERM grace period, then use SIGKILL before the final wait.
- Added a regression test with a child that ignores TERM and verified cleanup takes the bounded timeout and leaves no child.
- Confirmed the suspected `fm-watch.sh` signal-window mechanism is refuted: the temporary TERM trap records `FM_CHECK_SIGNAL_PENDING`, then the trap is replaced with `exit 1` and the pending flag is checked before `wait`, so TERM is acted on rather than swallowed indefinitely. No watcher signal handling was changed.

## Verification

- `tests/fm-watch-triage.test.sh` passes in 2m36s.
- `bin/fm-lint.sh` passes with ShellCheck 0.11.0.
- `bin/fm-test-run.sh --all` reached the changed test successfully, but the suite has one unrelated existing failure in `tests/fm-backend-herdr-focus-flash-e2e.test.sh`: “the Part C doomed pane never acquired a stable persistent sleep child process”.


eduardstan:
Closing this immediately - it was opened against this repository by mistake and is not a contribution request.

The change was intended for our own fork (`eduardstan/firstmate`). Our working copy is a fork of this repository, so `gh pr create` defaulted its base here rather than to our own default branch. The 235-file diff is our fork'''s divergence, not a proposed change; nothing here was meant for your review.

Apologies for the noise. Re-opening against our own fork. If any part of this turns out to be worth contributing upstream, it will come back properly scoped and through `git push no-mistakes`, as your contribution policy requires.

PR #2621, opened 2026-08-19
fix(lock): make the session lock work on Windows/MSYS

gagandeepjuneja7:
## Problem

On Windows (Git Bash/MSYS) a firstmate home can never take the helm. Every session start refuses the fleet lock with `cannot locate harness process in ancestry` and falls back to read-only, so the home cannot spawn, steer, merge, or drain wakes. It is not intermittent — it is every session, on every Windows host.

Three independent causes stack up:

**1. MSYS `ps` implements none of the `-o` field selectors the ancestry walk depends on.** It accepts only `[-aefls] [-u UID] [-p PID] [-W]`, so every `ps -o comm=` / `-o args=` / `-o ppid=` call in `fm_harness_ancestry_pids` fails and no pid is ever resolved.

**2. The Win32 process table alone cannot answer the walk either.** MSYS emulates `fork()` with a helper process that exits immediately, leaving the child's recorded `ParentProcessId` aimed at a dead — and eventually recycled — pid. A pure Win32 climb dies at the first forked shell, which is the shape of every `bash bin/fm-*.sh` invocation.

**3. Once the walk was fixed, the lock hung outright.** The lock proves ownership through a real symlink (`ln -s`, then `[ -L ]` + `readlink`), but MSYS's default symlink mode *copies* the target instead of linking it. `[ -L ]` is then false for a link that was just created successfully, `fm_lock_points_to_owner` can never confirm ownership, and `fm_lock_acquire_wait` spins forever. That is a hang rather than a refusal, and it takes the wake queue and watcher down with it.

## Fix

`bin/fm-session-lock-lib.sh` gains a process-table portability layer:

- POSIX hosts keep the exact `ps -o` calls they always used, byte for byte, so a fake `ps` on `PATH` still drives the library deterministically in tests.
- Windows detection is a lazy capability probe, never a `uname` check: a host whose `ps` answers `-o` is POSIX here by definition.
- On Windows the ancestry climbs the MSYS side with MSYS `ps` — which does know those parents — and crosses the namespace boundary exactly once, at the outermost MSYS process, whose WINPID is real and whose Win32 parent is a live native process. Skipping the MSYS processes themselves is safe: on Windows every verified harness is a native executable, so no harness can hide in the stretch that is jumped.
- The pid published in `state/.lock` is a Windows pid, which stays self-consistent because every reader of it goes through these same primitives.

`bin/fm-wake-lib.sh` sets `MSYS=winsymlinks:nativestrict` on MSYS hosts so the lock's symlinks are real. `nativestrict` is deliberate: a host without Developer Mode now fails visibly instead of silently reinstating the copy behaviour that causes the hang. An operator who already set `MSYS` is left alone.

## Performance

Reading the Win32 table costs a full PowerShell start, and MSYS forks are expensive, so a naive implementation pushed session start *over* its 120s runtime bound. Three changes brought it back:

- PowerShell writes the table straight to a short-lived cache file (addressed by its Windows path) instead of piping ~120KB through a shell variable.
- The ancestry walk is one `awk` pass emitting the whole chain, instead of three lookups per hop.
- The capability probe caches into the caller's shell instead of re-forking `ps` per field per hop.

| Operation | Before | After |
|---|---|---|
| Ancestry resolution | 19.3s | 3.0s |
| `fm-lock.sh status` | 7.2s | 2.1s |
| Full session start | 134s (truncated at bound) | 82s |

The PowerShell invocation also closes stdin and uses a single-line `-Command`: it inherits the caller's stdin and `-Command` does not stop it waiting on that handle, and a `-Command` string carrying embedded newlines across the MSYS argument boundary is read as an incomplete statement. Either one hangs the acquire.

## Verification

Windows 11 26200 / Git Bash, Claude harness:

- Lock acquires cleanly on **6 of 6** cold attempts, with zero leftover claim-lock or owner state.
- Full session start completes in **82s**, inside its bound, with wake queue, supervision instructions, and deferred network checks all running.
- `tests/fm-session-lock-ancestry.test.sh` passes its 4 unit cases — identical to unmodified `HEAD`.
- That suite's end-to-end layer fails **identically before and after** this change (verified by running it against a clean `HEAD` worktree). It builds orphaned POSIX process trees with `setsid` and `ln -s /bin/bash`, which cannot run on this platform at all.

The POSIX code path is unchanged, so macOS and Linux behaviour is byte-for-byte what it was.

## Note for maintainers

I could not run `no-mistakes` for this change: its installer refuses this platform outright with `Unsupported OS: mingw64_nt-10.0-26200`, as does `treehouse`. That may be worth a separate issue — the pipeline a Windows contributor is asked to use cannot currently be installed on Windows.

### DIFF
--- a/bin/fm-session-lock-lib.sh
+++ b/bin/fm-session-lock-lib.sh
@@ -25,6 +25,312 @@ FM_HARNESS_RE='claude|codex|opencode|grok|kimi|^pi$|^pi-signed$'
 # bin/fm-claude-stop-autoarm.sh.
 FM_HARNESS_NAMES=(claude codex opencode grok kimi pi-signed pi)
 
+# --- process-table portability ----------------------------------------------
+#
+# ONE owner of "how does this host answer comm/args/ppid for a pid?".
+#
+# This exists because MSYS2/Cygwin `ps` - the ps a Git Bash session on Windows
+# actually runs - implements none of `-o comm=`, `-o args=`, or `-o ppid=`; it
+# accepts only [-aefls] [-u UID] [-p PID] [-W]. Every ps-based ancestry walk
+# therefore failed outright there, fm_harness_ancestry_pid could never resolve a
+# pid, and fm-lock.sh refused the fleet lock on every session start with
+# "cannot locate harness process in ancestry". The whole home stayed permanently
+# read-only: no spawn, no steer, no merge, no wake drain.
+#
+# MSYS ps cannot substitute either. It reports MSYS pids whose parent chain is
+# severed at the Win32 boundary (a Git Bash child of claude.exe reports ppid 1),
+# and `ps -W` lists native processes with ppid 0. The real ancestry exists only
+# in the Win32 process table, so on Windows that table is read directly and the
+# walk runs in WINDOWS pids end to end - including the pid published in
+# state/.lock, which stays self-consistent because every reader of that pid goes
+# through these same primitives.
+#
+# POSIX hosts keep the exact `ps -o` calls they have always used, byte for byte,
+# so a fake `ps` on PATH still drives this library deterministically in tests.
+# Windows detection is a lazy capability probe, never a uname check: a host whose
+# ps answers `-o` is POSIX here by definition.
+#
+# Everything below is shaped by one fact: on MSYS, forks and large pipe payloads
+# are expensive, and a single session start runs dozens of short-lived firstmate
+# scripts that each need this table. So the table is produced straight into a
+# short-lived cache FILE by PowerShell, and every lookup is one awk pass over
+# that file. Nothing carries the table through a command substitution.
+
+# Records are: pid <TAB> ppid <TAB> name <TAB> executable-path <TAB> command-line
+#
+# The command line is truncated because nothing here reads past its head: the
+# evidence this library takes from it is argv[0] and an interpreter's script
+# path, both of which sit at the front. Tabs and newlines inside it are flattened
+# so one process is always exactly one record.
+FM_WIN_PS_CMDLINE_MAX="${FM_WIN_PS_CMDLINE_MAX:-300}"
+
+# One line, deliberately: a -Command string with embedded newlines is read as an
+# incomplete statement across the MSYS argument boundary and hangs waiting for
+# the rest. Writes to $env:FM_WIN_PS_OUT when that names a file, stdout otherwise.
+FM_WIN_PS_QUERY='$ErrorActionPreference="SilentlyContinue"; $m='"$FM_WIN_PS_CMDLINE_MAX"'; $out=$env:FM_WIN_PS_OUT; $rows=Get-CimInstance Win32_Process | ForEach-Object { $c=$_.CommandLine; if ($c -and $c.Length -gt $m) { $c=$c.Substring(0,$m) }; "{0}`t{1}`t{2}`t{3}`t{4}" -f $_.ProcessId,$_.ParentProcessId,$_.Name,$_.ExecutablePath,($c -replace "[`r`n`t]"," ") }; if ($out) { Set-Content -LiteralPath $out -Value $rows -Encoding ASCII } else { $rows }'
+
+# How long a cached table may be reused, in seconds. Deliberately small: the only
+# thing a stale table can do is report a process that exited within the window as
+# still alive, which at worst delays a lock takeover by that window - a delay the
+# surrounding lock protocol already tolerates, because it re-checks rather than
+# deciding once.
+FM_WIN_PS_CACHE_TTL="${FM_WIN_PS_CACHE_TTL:-5}"
+
+FM_PS_MODE=''
+FM_WIN_PS_FILE=''
+FM_WIN_PS_CACHE=''
+
+# Decide once whether this host's ps answers the -o field selectors, leaving the
+# answer in $FM_PS_MODE for THIS shell.
+#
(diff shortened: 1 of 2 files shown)


gagandeepjuneja7:
Follow-up evidence on `tests/fm-wake-queue.test.sh`, which I could not get to completion on Windows in the original writeup. It resolved, and the result strengthens the case for this change.

**The suite hangs on unmodified `HEAD`, at the bug this PR fixes.** Traced against a clean `HEAD` worktree, it stalls in the `self-held-lock` fixture at:

```
fm_lock_acquire_wait "$lock" || exit 10
```

sourcing `HEAD`'s own `bin/fm-wake-lib.sh`. That is cause 3 from the description: `ln -s` copies instead of linking, `fm_lock_points_to_owner` can never confirm ownership, and the wait spins forever.

**Reduced to the smallest case** — one `fm_wake_append` against a scratch state dir:

| Single wake append, Windows 11 / Git Bash | Result |
|---|---|
| unmodified `HEAD` | never completes (killed at 6m40s) |
| this branch | completes, 16.3s |

So on Windows today the durable wake queue cannot be written at all. It is not only the session lock that is broken — supervision cannot function either, because every append hangs.

**On this branch the suite gets further and passes what `HEAD` hangs on:**

```
ok - an abandoned same-process lock hold is reclaimed; a parent's live hold is not
ok - self-announced appends suppress only their own bytes and fail toward waking
ok - historical annotations replay nothing already announced and keep everything new
```

3 passed, 0 failed. The first of those is the very fixture that hangs on `HEAD`.

**Remaining honest caveat.** The suite still does not run to completion here. It stops progressing in the concurrency stress case that forks 41 simultaneous `fm_wake_append` writers and waits on all of them. Measured directly, that is slowness rather than deadlock — every record does land in the queue:

| Concurrent writers | Elapsed | Records queued |
|---|---|---|
| 1 | 26s | 1 |
| 2 | 44s | 2 |
| 4 | 200s | 4 |
| 8 | 254s | 7 of 8 within the window |

At roughly 16-26s per append on MSYS, 41 concurrent writers need far longer than any timeout I ran. I have not reduced that cost, and I am not claiming this suite green on Windows. What I am claiming is narrower and verifiable: this change takes the wake queue from *cannot complete a single write* to *completes every write*, and the tests that now run pass.

The per-append cost looks inherent to MSYS process creation rather than to the lock protocol, but I have not proven that, so I would rather flag it than assert it. If a Windows-capable maintainer wants to pull that thread, the concurrency fixture is the place to start.

No POSIX behaviour is affected by any of this — the `ps` path is byte-for-byte unchanged and the `MSYS` export is guarded to MSYS hosts, so Linux CI exercises exactly the code it did before.

gagandeepjuneja7:
Adjacent Windows finding from the same session, same root cause family (symlinks), but a separate defect from the one this PR fixes. Reporting it here rather than opening an issue since it is the same platform story; happy to split it out if you would prefer.

**Every skill is silently unavailable on a Windows checkout.**

`.claude/skills` is tracked as a symlink to `../.agents/skills`:

```
$ git ls-files -s .claude/skills
120000 2b7a412b8fa0fb7e985b0793321bd4e698f2b6cd 0	.claude/skills
```

Git for Windows defaults to `core.symlinks=false`. The entry then materialises as an ordinary 17-byte text file whose contents are the link target:

```
$ ls -la .claude/
-rw-r--r-- 1 user 197121   17 .claude/skills
$ cat .claude/skills
../.agents/skills
```

The harness finds no `.claude/skills` directory, so none of the 19 skills under `.agents/skills/` are discoverable — `stow`, `afk`, `bearings`, `harness-adapters`, `project-management`, `secondmate-provisioning`, and the rest. `AGENTS.md` instructs the agent to load several of them at precise triggers, and every one of those loads fails.

**What makes this worse than the lock bug: it fails invisibly.** The lock bug announces itself on every session start with a read-only banner. This one produces no warning at all. An agent following `AGENTS.md` simply cannot load the skill it is told to load, and nothing anywhere says why. I only found it because a captain asked for `/stow` by name — otherwise the degradation is silent and total.

**Repair on an affected clone:**

```bash
git config core.symlinks true
rm -f .claude/skills
MSYS=winsymlinks:nativestrict ln -s ../.agents/skills .claude/skills
```

`git status` is clean afterwards, confirming the symlink is what the index always intended. The skills were picked up mid-session with no restart. It is the only tracked symlink in the repo, so nothing else is affected:

```bash
git ls-files -s | awk '$1=="120000"{print $4}'   # .claude/skills, and nothing else
```

Native symlink creation needs Windows Developer Mode or an elevated shell — the same prerequisite this PR's `MSYS=winsymlinks:nativestrict` change relies on.

**Possible upstream fixes**, in rough order of how much they ask of the user — I have not implemented any of these, so treat them as suggestions rather than a proposal:

1. A `SETUP`/README note for Windows telling users to set `core.symlinks=true` before cloning. Cheapest, but it only helps people who read it before the fact.
2. A bootstrap check that verifies `.claude/skills` resolves and prints an actionable diagnostic when it does not. This fits the existing detect-then-consent bootstrap pattern and would have surfaced the problem immediately, which is the part that actually failed here.
3. Ship the directory rather than the symlink — for example `.claude/skills` as a real directory, or generating the link at bootstrap. Larger change, and I have not thought through the consequences for non-Windows installs.

Option 2 seems the best fit to me, since the real failure was the silence rather than the broken link. But you know the repo's constraints better than I do.

PR #2623, opened 2026-08-19
fix(bin): count a trust-registered custom check as supervision need

joliverMI:
## Intent

Fix a firstmate supervision predicate gap: a firstmate home whose only duty is a trust-registered custom check (bin/fm-check-register.sh) never got a watcher armed, because bin/fm-supervision-lib.sh's fm_supervision_status only counted state/*.meta task records, state/x-watch.check.sh, and state/procevent/*.source when deciding FM_SUP_NEEDED. A registered check matches none of those, so FM_SUP_NEEDED stayed false and the Claude Stop auto-arm believed the home had nothing to supervise, exiting 0 at every turn end without ever arming a watcher. This caused two real outages on the fleet auditor's home: a seven-hour cadence lapse and, more recently, its hand-armed background watcher being reaped.

Scope delivered:
1. bin/fm-supervision-lib.sh now counts a trust-registered check - a state/<id>.check-trust binding paired with its state/<id>.check.sh, written by bin/fm-check-register.sh - as supervision-needed (new FM_SUP_CHECKS counter, presence-based like the existing FM_SUP_SOURCES/x-watch.check.sh checks, not full cryptographic re-validation - that stronger check already happens on the actual check-execution path). bin/fm-turnend-guard.sh and bin/fm-guard.sh banners were updated to name "custom check(s) registered" specifically instead of falling through to the X-mode wording when only a check is registered. docs/turnend-guard.md's "Guard predicates" section, the authoritative doc for this contract, was updated to list the new input alongside the existing ones (one-owner rule).
2. tests/fm-turnend-guard.test.sh gained test_predicate_registered_check_needs_supervision: it registers a real check via bin/fm-check-register.sh in a scratch home with no state/*.meta task record, then asserts fm_supervision_needed reports true and FM_SUP_CHECKS is 1. Verified this test fails against the pre-fix predicate (git show HEAD:bin/fm-supervision-lib.sh swapped in) with "not ok" and passes with the fix.
3. Investigated (do not fix - this was scoped as a report only) whether the watcher's liveness signal can distinguish "this home's watcher" from "another home's watcher running the identically-named script", since the auditor said a naive process check was fooled and telling them apart required reading each process's working directory. Findings for the PR body:
- The automated liveness check already has a cheap, correct, already-implemented discriminator that does NOT depend on process name or cwd: bin/fm-wake-lib.sh's fm_watcher_lock_matches_pid/fm_watcher_healthy require the watcher lock's recorded state/.watch.lock/fm-home to equal the querying home, plus a full process-identity fingerprint (proc stat starttime plus the complete NUL-separated cmdline, from fm_pid_identity) - not a bare pgrep/ps-by-name match. So the automated predicate used by fm-guard.sh and fm-turnend-guard.sh was never vulnerable to the same-script-name confusion the auditor hit by hand.
- The real detectability gap is structural and separate from cross-home confusion: under the "autoarm" supervision model (the Claude Stop auto-arm model implicated in this incident), fm_watcher_supervision_verdict in bin/fm-wake-lib.sh intentionally skips the identity-matched liveness check entirely and returns healthy purely from state/.last-watcher-beat freshness, by design, because that model's watcher is expected to be absent between turns. A watcher that dies immediately after touching the beacon is therefore indistinguishable from one merely idle between turns, for up to the full grace window - this is the "outage is undetectable from the beacon" behavior described in the task, and it is a deliberate model tradeoff, not a bug to fix here.
- Separately, the auditor's own ad-hoc manual check (something like `ps aux | grep fm-watch.sh`) is genuinely fooled across homes that share one tracked code root: FM_HOME is passed via environment, not argv, so the command/script path in `ps` output is identical for every home's watcher process, and bin/fm-watch.sh never `cd`s into FM_HOME, so a process's cwd is not a reliable, by-construction discriminator for that kind of eyeballing (it only happens to differ if the arming script's own cwd happened to differ).
- Cheap, correct, zero-code-change discriminators already exist for manual diagnosis too: read state/<home>/.watch.lock/fm-home directly (the same marker the automated check trusts), or read /proc/<pid>/environ for FM_HOME= on a live candidate process. Making a bare `ps`+cwd check reliable as well would need a small watcher change (e.g., an explicit `cd "$FM_HOME"` early in bin/fm-watch.sh) - out of scope here per the brief's "do NOT redesign the watcher" constraint, and should become its own follow-up task if wanted.

Constraints honored: did not touch the watcher, the Stop hook mechanism, or the beacon mechanism itself; did not run pkill -f bin/fm-watch.sh or otherwise disturb any sibling home's watcher while testing; loaded firstmate-coding-guidelines before editing this repo's shared tracked material.

## What Changed

- `bin/fm-supervision-lib.sh`'s `fm_supervision_status` now counts trust-registered custom checks — a `state/<id>.check-trust` binding paired with its `state/<id>.check.sh` — in a new `FM_SUP_CHECKS` counter and sets `FM_SUP_NEEDED=true` for them, so a home whose only duty is a registered check is no longer treated as idle by the Claude Stop auto-arm and the guards.
- `bin/fm-turnend-guard.sh` and `bin/fm-guard.sh` gained a `custom check(s) registered` branch in their block reason, banner, and need description, instead of falling through to the X-mode relay wording when only a check is registered.
- Tests: `tests/fm-turnend-guard.test.sh` adds `test_predicate_registered_check_needs_supervision` (asserts `fm_supervision_needed` is true and `FM_SUP_CHECKS` is 1 with no `state/*.meta`) and `test_hook_blocks_registered_check_only_home`; `tests/fm-claude-stop-autoarm.test.sh` adds `test_arms_for_registered_check_need_without_inflight`, each registering a real check via `bin/fm-check-register.sh`.
- Docs: `docs/turnend-guard.md`'s "Guard predicates" section lists the new input as the one owner of the contract, and `AGENTS.md`, `docs/architecture.md`, `docs/scripts.md`, `docs/subagent-guard.md`, and `docs/supervision-protocols/grok.md` now point at that section instead of restating a stale input list.

## Risk Assessment

✅ Low: The behavior change is a single presence-based counter added at the one shared supervision predicate every consumer already calls, it fails safe (over-requiring supervision rather than under-requiring it), it has a genuine pre-fix-failing regression test, and no duplicated sibling predicate leaves the reported outage reachable; the only gaps are a stale in-file comment and missing hook-level coverage for the new banner branches.

## Testing

Ran the two owning suites (fm-turnend-guard, fm-claude-stop-autoarm) plus the two adjacent supervision-predicate consumers, all green, and added two focused regression tests covering the auto-arm and turn-end-block layers where the reported outage actually occurred — the change shipped only predicate-level coverage. Confirmed the regression property by swapping the pre-fix supervision lib back in: all three check-related tests fail there and pass with the fix. For product-level evidence I drove a real check-only home end-to-end on both commits and captured a CLI transcript showing the before state (Stop auto-arm exits 0, no watcher armed, both guards silent) versus the after state (turn end blocked with "1 custom check(s) registered", watcher-down banner, arm wrapper actually ran). This is a CLI/hook change with no rendered UI surface, so the transcript is the end-user-visible artifact. No real watcher was started and no sibling home was disturbed; the worktree carries only the two intentional test-file additions.

<details>
<summary>Evidence: Before/after CLI transcript: check-only home across all three supervision surfaces</summary>

Source: [Before/after CLI transcript: check-only home across all three supervision surfaces](https://github.com/joliverMI/firstmate/blob/757d5529d9c7e0604870ff7fb18aa3991d572332/.no-mistakes/evidence/fm/fm-supervision-predicate-registered-check/check-only-home-before-after.txt)

<code>AFTER the fix (commit a158068) - same check-only home&#10;$ ls state/ # the home&#39;s entire duty: one registered custom check, zero *.meta task records&#10;audit.check.sh&#10;audit.check-trust&#10;&#10;--- 1. Turn-end Stop guard: bin/fm-turnend-guard.sh&#10;● TURN WOULD END BLIND - SUPERVISION IS OFF&#10;● 1 custom check(s) registered, but no live watcher holds this home lock (last beat: never).&#10;[exit 2] &lt;- turn end BLOCKED: supervision demanded&#10;&#10;--- 3. Claude Stop auto-arm: bin/fm-claude-stop-autoarm.sh (the mechanism that failed in the outage)&#10;[exit 2]&#10;state/arm-ran -&gt; WATCHER ARMED (arm wrapper ran, pid 3481573)&#10;&#10;(BEFORE the same three surfaces: exit 0 / silent / state/arm-ran ABSENT - no watcher ever armed)</code>

```text

BEFORE the fix (base commit 03bb1d8) - check-only home
$ ls state/   # the home's entire duty: one registered custom check, zero *.meta task records
  audit.check.sh
  audit.check-trust

--- 1. Turn-end Stop guard:  bin/fm-turnend-guard.sh  (what Claude sees when the turn ends)
$ echo '{"stop_hook_active":false}' | FM_HOME=<home> bin/fm-turnend-guard.sh
  [exit 0]  <- turn end allowed: guard saw nothing to supervise

--- 2. Pull guard banner:  bin/fm-guard.sh  (what the agent sees in supervision tool output)
$ FM_HOME=<home> bin/fm-guard.sh
  [exit 0]  <- silent: guard saw nothing to supervise

--- 3. Claude Stop auto-arm:  bin/fm-claude-stop-autoarm.sh  (the mechanism that failed in the outage)
$ echo '{...}' | FM_HOME=<home> claude -c 'echo $$ > state/.lock; bin/fm-claude-stop-autoarm.sh'
  [exit 0]
  state/arm-ran -> ABSENT: no watcher was ever armed for this home

AFTER the fix (commit a158068) - same check-only home
$ ls state/   # the home's entire duty: one registered custom check, zero *.meta task records
  audit.check.sh
  audit.check-trust

--- 1. Turn-end Stop guard:  bin/fm-turnend-guard.sh  (what Claude sees when the turn ends)
$ echo '{"stop_hook_active":false}' | FM_HOME=<home> bin/fm-turnend-guard.sh
  ●  1 custom check(s) registered, but no live watcher holds this home lock (last beat: never).
  [exit 2]  <- turn end BLOCKED: supervision demanded

--- 2. Pull guard banner:  bin/fm-guard.sh  (what the agent sees in supervision tool output)
$ FM_HOME=<home> bin/fm-guard.sh
  ●  1 custom check(s) registered, but no watcher has a fresh beacon (last beat: never, grace 300s).
  [exit 0]  <- watcher-down banner shown

--- 3. Claude Stop auto-arm:  bin/fm-claude-stop-autoarm.sh  (the mechanism that failed in the outage)
$ echo '{...}' | FM_HOME=<home> claude -c 'echo $$ > state/.lock; bin/fm-claude-stop-autoarm.sh'
  firstmate watcher wake - one supervision event needs a handling turn now.
  stale: demo actionable close
  Run bin/fm-wake-drain.sh first, handle the wake, then run its exact WAKE_ACK_REQUIRED --ack-through command. Until that post-handling acknowledgement, interruption leaves the wake durable for idempotent re-handling. This Stop hook owns watcher continuity: when the handling turn ends, the next needed cycle arms automatically - do NOT run bin/fm-watch-arm.sh after an ordinary wake.
  [exit 2]
  state/arm-ran -> WATCHER ARMED (arm wrapper ran, pid 3490318)
```
</details>
<details>
<summary>Evidence: Regression property: three tests fail pre-fix, pass post-fix</summary>

Source: [Regression property: three tests fail pre-fix, pass post-fix](https://github.com/joliverMI/firstmate/blob/757d5529d9c7e0604870ff7fb18aa3991d572332/.no-mistakes/evidence/fm/fm-supervision-predicate-registered-check/regression-before-after.txt)

<code>WITH THE PRE-FIX PREDICATE&#10;not ok - a home with only a trust-registered check and no task record did not report supervision as needed&#10;not ok - hook must block when a check-only home has no watcher: expected exit 2, got 0&#10;not ok - a trust-registered custom check must keep the auto-arm active with zero tasks in flight: expected exit 2, got 0&#10;&#10;WITH THE FIX (a158068)&#10;ok - fm_supervision_needed: a trust-registered check with no task record needs supervision&#10;ok - fm-turnend-guard: non-Claude path blocks a registered-check-only home&#10;ok - auto-arm: a registered custom check arms the cycle even with no tasks in flight</code>

```text
Regression property of the three supervision-predicate tests
(pre-fix bin/fm-supervision-lib.sh from base 03bb1d8 swapped in, then restored)

WITH THE PRE-FIX PREDICATE
  tests/fm-turnend-guard.test.sh      not ok - a home with only a trust-registered check and no task record did not report supervision as needed
  tests/fm-turnend-guard.test.sh      not ok - hook must block when a check-only home has no watcher: expected exit 2, got 0
  tests/fm-claude-stop-autoarm.test.sh  not ok - a trust-registered custom check must keep the auto-arm active with zero tasks in flight: expected exit 2, got 0

WITH THE FIX (a158068)
  tests/fm-turnend-guard.test.sh      ok - fm_supervision_needed: a trust-registered check with no task record needs supervision
  tests/fm-turnend-guard.test.sh      ok - fm-turnend-guard: non-Claude path blocks a registered-check-only home
  tests/fm-claude-stop-autoarm.test.sh  ok - auto-arm: a registered custom check arms the cycle even with no tasks in flight

Full suite results with the fix in place:
  tests/fm-turnend-guard.test.sh        exit 0
  tests/fm-claude-stop-autoarm.test.sh  exit 0
  tests/fm-guard-stale-banner.test.sh   exit 0
  tests/fm-supervision-events.test.sh   exit 0
```
</details>

## Pipeline

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>

- ℹ️ `docs/architecture.md:84` - docs/architecture.md enumerates this exact predicate&#39;s inputs in two places — line 84 (&#34;when work, a process-event source, or Relay polling needs supervision and no identity-matched watcher lock with a fresh beacon is live&#34;) and line 81 for fm-guard.sh — and both now omit the trust-registered custom check that this change added as a fourth input, so the architecture doc under-describes the contract as of this commit. AGENTS.md is similarly asymmetric: line 108 annotates procevent/ with &#34;their presence alone keeps supervision required&#34;, while the &lt;id&gt;.check-trust entry at line 99 carries no such note. The intent designates docs/turnend-guard.md as the one owner of this contract, which is a defensible call, so the fix is either to add the clause in these two summaries the way &#34;process-event source&#34; is already duplicated there, or to make them defer to turnend-guard.md instead of listing inputs — an ownership decision for the author, not a mechanical edit.
- ℹ️ `bin/fm-claude-stop-autoarm.sh:21` - The &#34;Need&#34; bullet in the auto-arm header still states &#34;arms only while work is in flight (state/*.meta) or X mode has a relay poll to run (state/x-watch.check.sh); an idle home exits 0.&#34; That is now wrong in two ways: it omits registered process-event sources (pre-existing, since #1483) and the trust-registered custom check added here. The actual behavior comes from fm_supervision_needed at line 121. This is the exact component whose &#34;this home has nothing to supervise&#34; belief caused the two reported outages, so its header is the most misleading place for a stale input list. bin/fm-guard.sh:8-9 has the same stale enumeration (&#34;if a task is in flight (a state/&lt;id&gt;.meta exists) or X-mode relay polling is active&#34;), even though the in-body comment at line 151-154 was correctly updated.

🔧 Fix: replace stale supervision-input restatements with owner pointers
2 infos still open:
- ℹ️ `bin/fm-claude-stop-autoarm.sh:120` - The fix round replaced the stale input enumeration in this file&#39;s header bullet (lines 21-23) with a pointer to fm_supervision_needed / docs/turnend-guard.md, but left the identical stale restatement in the section comment right above the gate: `# --- need: in-flight work or an X-mode relay poll ---`. It omits process-event sources (stale since #1483) and the trust-registered custom check this change added, so the file now contradicts itself six lines apart - the header defers to the owner doc while the comment immediately below still hand-lists two of four inputs. Same one-line pointer treatment applies. (bin/fm-guard.sh:152-155 still hand-lists the inputs too, but that copy was updated in this change and is currently accurate; it is the weaker case.)
- ℹ️ `bin/fm-turnend-guard.sh:188` - The two new user-visible banner branches - `%s custom check(s) registered` in bin/fm-turnend-guard.sh:188 and bin/fm-guard.sh:213 - have no test, while the exactly analogous process-event branch has coverage on both hooks (tests/fm-turnend-guard.test.sh:258 test_hook_blocks_source_only_home asserts &#34;1 process-event source(s) registered&#34;, and tests/fm-procevent.test.sh:1093 asserts the same string from fm-guard.sh). The new test only exercises the predicate layer, so nothing proves a check-only home actually blocks the turn end or names its cause instead of falling through to the misleading &#34;X-mode relay polling needs supervision&#34; else-branch. Mirroring test_hook_blocks_source_only_home is cheap: the predicate is presence-based, so the scenario needs only `: &gt; &#34;$dir/state/audit.check-trust&#34;` plus `: &gt; &#34;$dir/state/audit.check.sh&#34;` before run_hook, asserting exit 2 and &#34;1 custom check(s) registered&#34;.

</details>

<details>

✅ No issues found.
- <code>`bash tests/fm-turnend-guard.test.sh` (full suite, exit 0; includes the author&#39;s `test_predicate_registered_check_needs_supervision` and the added `test_hook_blocks_registered_check_only_home`)</code>
- <code>`bash tests/fm-claude-stop-autoarm.test.sh` (full suite, exit 0; includes the added `test_arms_for_registered_check_need_without_inflight`)</code>
- <code>`bash tests/fm-guard-stale-banner.test.sh` (adjacent consumer of fm_supervision_status, exit 0)</code>
- <code>`bash tests/fm-supervision-events.test.sh` (adjacent supervision predicate consumer, exit 0)</code>
- <code>Regression check: swapped in `git show 03bb1d8:bin/fm-supervision-lib.sh`, re-ran both suites and confirmed all three check-related tests report `not ok`, then restored the file</code>
- <code>Manual end-to-end A/B: built scratch primary homes from base and target commits, registered a real check via `FM_HOME=&lt;home&gt; bin/fm-check-register.sh audit`, then ran `bin/fm-turnend-guard.sh` (Stop payload on stdin), `bin/fm-guard.sh`, and `bin/fm-claude-stop-autoarm.sh` (fake claude harness holding state/.lock, stubbed bin/fm-watch-arm.sh) against each</code>

</details>

<details>

- ℹ️ `docs/turnend-guard.md` - Judgment call left unfixed and out of scope: the change&#39;s investigation established that manual cross-home watcher diagnosis (`ps aux | grep fm-watch.sh`) is unreliable because FM_HOME travels in the environment, not argv, and `bin/fm-watch.sh` never cd&#39;s into FM_HOME - and that the correct cheap discriminators are `state/&lt;home&gt;/.watch.lock/fm-home` or `/proc/&lt;pid&gt;/environ`. No tracked document owns that operator-facing diagnosis guidance today; the nearest hook is AGENTS.md&#39;s &#34;never pkill -f bin/fm-watch.sh&#34; line. Adding it would be new content rather than a staleness fix, so it is left for a follow-up alongside the proposed explicit `cd &#34;$FM_HOME&#34;` watcher change. The related autoarm-model tradeoff (beacon-only health, so a dead watcher is indistinguishable from an idle one within grace) is already documented in docs/turnend-guard.md&#39;s Guard predicates section and needed no change.

</details>

<details>

</details>

<details>

✅ No issues found.

</details>


joliverMI:
Opened in error by our own tooling against the wrong base repository - this is work for our fork and was never intended as an upstream contribution. Closing; apologies for the noise.

PR #2670, opened 2026-08-20
fix(bin): bind tmux endpoint presence to exact session/window resolution

KakkoiDev:
## Root cause

`fm_backend_target_exists` (bin/fm-backend.sh) and `fm_backend_tmux_agent_state` (bin/backends/tmux.sh) resolved a recorded `session:window` endpoint using tmux's default fuzzy/prefix name matching instead of exact matching:

- `tmux list-panes -t <session>:<window>` and `tmux list-windows -t <session>` match a *prefix* of the session name, and the window-name comparison after inherited a similar prefix/glob laxness.
- A stale recorded endpoint whose window (or whole session) no longer exists could therefore still resolve against an unrelated window or a differently-named live session, reading ALIVE when the actual recorded endpoint is dead.
- On the cheap session-start fleet digest, a false ALIVE hides a dead worker from the supervisor.
- On the deeper recovery classifier (`fm_backend_tmux_agent_state`), a false ALIVE is worse: it can suppress a legitimate relaunch, or (in the renamed-session case) risk authorizing a duplicate relaunch onto a live worktree.

## Fix

- Bind both the session and window axes to tmux's `=` exact-match prefix, relying on exit status alone rather than a name comparison against captured output.
- In the recovery classifier, add a fuzzy re-probe when the exact session is absent: if a live session's name merely prefixes the recorded one, reuse that probe's window list to test membership — recorded window present reports `unreadable` (protects a renamed-but-live agent from a duplicate relaunch), absent reports `missing` (safe to recover, since no live session holds that window).
- A transient failure of the fuzzy re-probe itself (signal/fork hiccup) now reports `unreadable`, not `missing`, so a transient tmux glitch never licenses a duplicate relaunch.
- The cheap digest path (`fm_backend_target_exists`) intentionally does not re-probe on a renamed session: a renamed-but-live session reads dead there, which is documented in docs/tmux-backend.md as the safe direction for a passive read (false-dead only makes a supervisor look, never authorizes an action).

Files: `bin/fm-backend.sh`, `bin/backends/tmux.sh`, `docs/tmux-backend.md`, `docs/verification/runtime-backends.md`, plus colocated regression tests under `tests/`.

## Regression tests

Colocated tests fabricate a `state/<id>.meta` pointing at nonexistent tmux windows/sessions (dead-window-in-live-session and whole-session-gone cases) and assert dead/missing instead of the previous false ALIVE. Each new test fails on the pre-fix code and passes on this branch.

## Verification

Full no-mistakes pipeline run (review, test, document, lint) passed at head `51b38755`; only the push step failed, due to the invoking machine's GitHub identity (KakkoiDev) lacking write access to kunchenguid/firstmate. Delivered manually via a personal fork per captain's instruction: pushed `KakkoiDev:fm/fm-liveness-bug-r2`, opening this PR against upstream.


KakkoiDev:
Closing: this PR was opened autonomously by a Claude agent by error. It was never meant to be proposed upstream; we keep our changes on our fork. Apologies for the noise.

PR #2687, opened 2026-08-20
fix: detect identical dirty local merge safely

wongsuwarn:
Detect the narrow case where every dirty tracked file already matches the approved branch without changing the project checkout. The helper names the paths and provides a safe manual landing command, while all other dirty states still refuse and clean checkouts still fast-forward.


wongsuwarn:
Opened against the wrong repository by mistake - this change is specific to a private downstream fork and is not intended as an upstream contribution. Closing; apologies for the noise.

PR #14, opened 2026-06-13
feat: add per-project delivery modes

kunchenguid:
## Intent

Add a per-project delivery mode plus an orthogonal yolo autonomy flag to firstmate, per the captain-approved plan at data/project-modes-p2/plan.md. Each project in data/projects.md gets a mode - no-mistakes (default), direct-PR, or local-only - chosen at project-add and recorded as '- <name> [<mode>] - <desc>' with optional '+yolo'. A new reader bin/fm-project-mode.sh parses it (legacy un-annotated lines and unknown modes/projects default safely to 'no-mistakes off' with a stderr warning). fm-spawn resolves and records mode= and yolo= into state/<id>.meta alongside harness=/kind=. fm-brief shapes the ship definition-of-done by mode: no-mistakes -> /no-mistakes pipeline; direct-PR -> crewmate pushes + opens PR via gh-axi, no pipeline; local-only -> crewmate stops at 'ready in branch', firstmate reviews and merges to local main. fm-teardown gains a local-only carve-out: since local-only has no remote, the safety check requires the branch merged into the local default branch instead of 'on a remote'. New bin/fm-merge-local.sh performs the approved local-only merge as a clean fast-forward only (firstmate's merge gate-action; a deliberate, documented, narrowly-scoped exception to hard rule #1, only after captain/yolo approval). AGENTS.md sections 1, 2, 6, 7, 11 document the modes, the meta/registry format, conditional init (direct-PR/local-only skip no-mistakes init; local-only needs no remote), the per-mode lifecycle, and the yolo approval matrix (yolo on = firstmate approves routine decisions itself but still escalates destructive/irreversible/security-sensitive ones, never merges a red PR, and posts a post-hoc FYI). Deliberate decisions the captain made: yolo is orthogonal to mode (not a 4th mode) and offered-but-not-recommended; for local-only, firstmate (not the crewmate) performs the merge, hence the rule-#1 carve-out. Shell-only change; no test suite in this repo.

## What Changed

- Added per-project delivery mode parsing with safe defaults, including optional `+yolo` autonomy, and records `mode=`/`yolo=` in task metadata during spawn.
- Updated task briefs and lifecycle handling for `no-mistakes`, `direct-PR`, and `local-only` delivery paths, including local-only teardown safety.
- Added the approved local-only fast-forward merge helper and documented the registry format, mode behavior, initialization rules, and yolo approval matrix.

## Risk Assessment

[...]

PR #2688, opened 2026-08-20
feat(herdr): auto-close a worker's own new space on teardown

jgrizon:
## Problem

Herdr #14 made a space survive its own last tab closing, correct for a human's space. Firstmate creates one space per worker and, at teardown, closes only that worker's exact pane - it never calls `workspace close`. Since #14, that composition leaves an empty, permanent space behind for every finished worker.

## Fix

Herdr #22 (merged, `f868a52b`) added `workspace create --auto-close`, letting a new space opt into closing itself once its last tab goes. `fm_backend_herdr_workspace_ensure` now passes that flag when creating a firstmate worker's own new workspace, so Herdr cleans it up once its last (already worker-closed) pane goes.

## Detection and degrade path

The captain's installed herdr (0.8.0) predates #22. Rather than parse a version string, detection follows the existing structural-signal pattern already used by `fm_backend_herdr_events_capable`: `fm_backend_herdr_workspace_autoclose_capable` greps the live `herdr api schema --json` for the `auto_close` field `#22` adds to `WorkspaceCreateParams`. Verified locally: the installed 0.8.0 schema has no such field, so the new smoke/e2e runs (`fm-backend-herdr-smoke.test.sh`, `fm-backend-herdr-launcher-workspace-e2e.test.sh`) exercised the real degrade path end to end and all passed - `--auto-close` is simply omitted, `workspace create` runs exactly as before, and nothing fails.

## Scope

- Only `fm_backend_herdr_workspace_ensure`'s create call (the worker-space path used by `fm_backend_herdr_container_ensure`) is touched.
- The presentation/projection workspace path (`fm_backend_herdr_projection_create_task`) is untouched - out of scope per the task.
- No `workspace close` call was added or changed anywhere; teardown's refusal to close spaces is untouched.
- tmux, zellij, orca, cmux backends are untouched.

## Tests

Added to `tests/fm-backend-herdr.test.sh` (fake-CLI unit style, matching the file's existing shape):
- `fm_backend_herdr_workspace_autoclose_capable` returns capable/incapable based on a scripted `api schema` response.
- `fm_backend_herdr_container_ensure` passes `--auto-close` on a capable herdr.
- The existing `--no-focus` container_ensure test now also asserts `--auto-close` is absent by default (the pre-#22 degrade path), since the whole suite defaults to `FM_BACKEND_HERDR_AUTOCLOSE_FORCE=0` so the existing canned-response tests are unaffected.

Full suite: `bash tests/fm-backend-herdr.test.sh` - 183 ok, 0 failures. `bin/fm-lint.sh` clean. Real-herdr smoke and launcher-workspace e2e suites also pass against the installed pre-#22 binary.

## Test plan

- [x] `bash tests/fm-backend-herdr.test.sh` (183 ok)
- [x] `bin/fm-lint.sh`
- [x] `bash tests/fm-backend-herdr-smoke.test.sh` (real herdr 0.8.0, degrade path)
- [x] `bash tests/fm-backend-herdr-launcher-workspace-e2e.test.sh` (real herdr 0.8.0)
- [x] `bin/fm-doc-audience-check.sh`


jgrizon:
Closing immediately: opened against the wrong repository by mistake. This is a downstream fork's private change and was never intended for this project. Apologies for the noise.

PR #2696, opened 2026-08-20
feat(bin): carry local-only projects on secondmate routes and land them back in the folder that holds the project

EternalTime:
A `local-only` project could not be held by a persistent second mate: seeding refused it outright, because both secondmate delivery modes end in a pull request on a forge and a folder-only project has nowhere to open one. That refusal is removed, and the landing is completed so approved work reaches the copy its owner actually reads.

## What changes

Seeding accepts every registered delivery mode. A clone is always made from the project's own origin, so a second mate's clone is a sibling of the seeding home's clone rather than a copy of it. For a `local-only` project that origin is the folder holding the project, which is what lets a landing reach that folder in one hop. A project with no origin to seed from is refused with the gap named and the exact command that closes it.

Crewmates ship such a task exactly as `local-only` already worked: implement on `fm/<id>`, stop with a clean ready branch, no push and no pull request.

`bin/fm-merge-local.sh` gains a second guarded stage. After fast-forwarding the clone's default branch it carries that branch into the folder the clone names as its origin. Git refuses a push into a non-bare repository's checked-out branch, so the landing pulls from the clone into that folder instead.

Backlog handoff never had a delivery-mode gate in code, only a documented prohibition. The prohibition is gone and a test pins that no gate exists.

## Safety

Stage two is as narrow as stage one and never forces, stashes, or discards. It runs only when origin names the root of a git work tree on this machine, only while that folder sits on the project's default branch, and only as a fast-forward.

A folder someone works in is normally dirty, and that alone is harmless, so the landing refuses on path overlap rather than on any uncommitted work: it asks git which paths the fast-forward would change and refuses, naming them, when any is uncommitted, untracked, or gitignored and really present there. Two cases git's own two-tree check does not cover are handled explicitly, since it refuses an untracked blocker but silently removes an ignored one: a gitignored file at a path the landing would overwrite, and a gitignored file standing where the landing needs a directory. A wholly ignored directory does not block a landing that merely adds a new file beneath it.

An origin that is absent, bare, or on another host is not a folder anyone reads, so the landing reports that it ended at the clone rather than refusing. An origin spelled relative to the clone or with a tilde is resolved the same way seeding resolves it, and the landing advises re-pointing it at the resolved absolute path, because a clone spelled that way can receive a landing but cannot have a worker launched into it. Both stages are idempotent, so a repeat run converges.

A whole-home remote route still refuses a `local-only` project, because that host can neither clone a folder that exists only on the seeding machine nor receive a landing into it.

## Tests

`tests/fm-merge-local.test.sh` is new and drives real repositories through both stages, covering the accepting paths and every refusal, and asserting after each refusal that the origin folder is unchanged and its uncommitted work intact. The seeding, origin, and backlog-handoff suites are extended for the accepting and refusing paths they own.


EternalTime:
Closing this in favour of the same branch raised through the no-mistakes gate, which carries the required signature this hand-opened PR cannot have.

PR #2702, opened 2026-08-20
feat(board): decompose big-picture cards, place backlog work, and reconcile on the cycles that already exist

huang-daniel:
## What this adds

Three optional `big-picture-*` columns turn a container card into a decomposition request. `poll` reports `decompose`, firstmate breaks the container into concrete work, and `bin/fm-board.sh child-add` files each piece as a **native GitHub sub-issue** carded in the ordinary Todo column, so the board's own `Parent issue` and `Sub-issues progress` fields carry the structure with nothing invented on top. The parent card then follows its children.

Correction 2 added the missing direction: `place` puts a task that started in the backlog onto the board. It and `child-add` turned out to be the same operation with different parents, so they run one shared implementation rather than two that can drift.

Correction 5's ruling is encoded throughout: firstmate's records are the truth and the board reports them. A card showing a status firstmate did not write is a `divergence` that changes nothing on either side and is raised with the captain, replacing the old `instruction` adoption path. Filing a card stays the captain's intent, and board-sourced work lands **held**.

## Constraints, and how each is held

**A container never binds a task** - closed from four sides, not by convention: `poll` never emits `new` for a card in a big-picture column; `poll` writes the container's record the first time it *sees* the card rather than when it is decomposed; `import` refuses any issue holding that record; `child-add` refuses a parent that holds a link. Tested from both directions. The residual gap is an `import` of a container URL that no poll has ever seen, which requires inventing a URL the adapter never surfaced; the header states that boundary rather than implying it is closed.

**A child is never re-imported** - it carries the trigger label and holds a link before the next cycle reads the board, so `poll` reports it `linked`. The link is written as soon as the card exists, not at the very end, precisely so a just-created card is never briefly importable as fresh work.

**Convergence** - a repeat `place`/`child-add` for the same task never files a second issue, through three guards in falling order of cost: an existing link answers with no network call at all; a child already recorded against its parent is reused; otherwise a shallow scan of the repo's newest issues adopts one already carrying the task's `firstmate-task:` marker. A scan that *could not run* stops the command rather than risking a duplicate - a failed read is not the same answer as "nothing found".

**Rate limits** - `place` and `child-add` read no board at all, taking the card id from what `item-add` returned (a freshly added card is not visible in `item-list` anyway). Board/field/option ids resolve once per invocation. Cost is a small constant per card, independent of board size. Stated in a new `COST` section of the header.

**Fail soft** - unchanged. The one deliberate addition: `child-add`/`place` exit 1 if the issue itself could not be created, because that is the command's whole purpose and firstmate must not build a backlog item on it. Everything after that irreversible step degrades to `*-partial` and exits 0.

## Automatic reflection (correction 2, section 7)

`fm-spawn.sh` places-then-marks on a ship dispatch, `fm-pr-check.sh` attaches the PR, `fm-pr-merge.sh` closes the card after a merge that actually landed. All three are inert for a task with no link and for every home with no board - proven by a test asserting the GitHub CLI is never invoked at all. None can fail a dispatch or alter a merge's outcome.

## Session-start reconciliation (section 9)

The poll runs in the deferred network stage `bin/fm-startup-network.sh` already owns, via `fm-bootstrap.sh`'s network phase, printing actionable lines as `BOARD_POLL:`. No wake source, watcher check, timer, or daemon was added. It is skipped in a read-only session, and a home with no board never even names the sweep.

**This required a second AGENTS.md line beyond the trigger change.** Section 3 enumerates what the deferred network stage owes; leaving board reconciliation out of that list would have made the owner statement wrong. Total AGENTS.md diff is three lines: the trigger, that enumeration, and one `data/` layout entry for the new record file.

## On `fm-board-quiet-poll`'s invariants

Every new path satisfies them from the start: one board read per cycle, all writes against that snapshot, no refetch, and **every new record type is silent when reconciled** - a settled container prints nothing.

I did **not** suppress the pre-existing `linked` lines, which is that item's own stated purpose and carries its own test changes. Instead the session-start surfacing filters them out, so the new automatic poll adds no noise to every session while leaving `fm-board-quiet-poll` its scope, including invariant 7's complexity guard.

`fm-board-area-field` and `fm-board-closed-issue-withdrawal` were left alone.

## Verification

- `bin/fm-lint.sh` clean with both pinned linters (ShellCheck 0.11.0, actionlint 1.7.12; neither was installed here, so both were fetched at their pinned versions).
- `tests/fm-board.test.sh` 38 pass, `tests/fm-board-dispatch.test.sh` 6 pass (new), `tests/fm-bootstrap.test.sh` 30 pass, `tests/fm-pr-merge.test.sh` 13 pass, plus `fm-pr-check-security`, `fm-spawn-batch` and `fm-spawn-dispatch-profile` unaffected.
- Coverage includes: the feature fully off when keys are unset, a partial key set failing loudly, `decompose` never being `new`, a decomposed parent never reported twice, `child-add` and `place` convergent on repeat runs, parent status derived from child states, an ordinary Todo card untouched, a card firstmate did not place in `queued` producing a report and no dispatch, and a board that refuses every write never failing a dispatch.
- Two existing tests changed rather than being added to, because correction 5 reversed the rule they encoded: a captain's status edit is no longer adopted into firstmate's record.

While fixing the merge-reflection test I found the pr-merge fixture's `gh` mock ends in `exit 0`, so an appended case block was unreachable - which had made a first attempt at the assertion pass falsely. The mock is now written whole, and `run_pr_merge` pins `FM_CONFIG_OVERRIDE`/`FM_DATA_OVERRIDE` into the case so a merge test can never reach a developer's real board configuration.

### DIFF
--- a/.agents/skills/board-orchestration/SKILL.md
+++ b/.agents/skills/board-orchestration/SKILL.md
@@ -101,6 +101,9 @@ Never move a card back to what firstmate expected, and never treat a captain's e
   Tell the captain which board owns the issue and let them decide which board should carry it; never re-home it by hand.
 - An `error` line means the board could not be read, or answered so implausibly that the adapter refused to act on it.
   Treat it as a board that is temporarily unavailable, let the next cycle reconcile, and never read it as work being withdrawn.
+- A `truncated` line means the board filled the read's card ceiling, so a card past it was never seen.
+  Everything that read did report still stands, but no withdrawal is reported from that board on this cycle, because absence cannot be told apart from the ceiling.
+  Re-run `poll` for that board with a higher `--limit`, and if it stays truncated tell the captain the board has outgrown the default read; the same ceiling applies to the card lookup `mark` needs, so a card sitting past it cannot be moved either.
 - A scope edit on a card mid-flight follows the lifecycle rule AGENTS.md section 7 already owns: route it to follow-up work unless it completely invalidates the work being validated.
   A board edit does not create a second, competing rule for that.
 
--- a/.agents/skills/board-orchestration/SKILL.md
+++ b/.agents/skills/board-orchestration/SKILL.md
@@ -1,0 +1,110 @@
+---
+name: board-orchestration
+description: >-
+  Agent-only policy for bridging a configured project board and firstmate's existing backlog.
+  Load on a session-start or heartbeat cycle when a project board is configured, before importing a board item, before reflecting a dispatch, PR, blocker, or merge onto a board, and whenever a board edit disagrees with the backlog.
+user-invocable: false
+metadata:
+  internal: true
+---
+
+# Project board orchestration
+
+Firstmate is the orchestrator, the board is the captain's visible statement of intent, and `data/backlog.md` remains the one execution queue.
+This skill owns the policy; [`bin/fm-board.sh`](../../../bin/fm-board.sh)'s header owns every command, flag, record, and failure mode, and [`docs/configuration.md`](../../../docs/configuration.md) owns the local `config/boards` file.
+
+## The bridge is inert until a board is configured
+
+A home with no `config/boards` file, or an empty one, does none of this.
+It performs zero board reads and zero board writes, invokes no GitHub CLI for a board, and behaves exactly as it did before the bridge existed, the same way Relay stays inert until its own opt-in.
+Do not run a board command, mention a board to the captain, or offer board behavior in such a home.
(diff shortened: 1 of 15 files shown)


huang-daniel:
Re-targeted to huang-daniel/firstmate#2 - this home's main tracks the fork, and the captain has no merge rights on this repo, so a PR based here could not land or reach the fleet.

PR #2700, opened 2026-08-20
feat(bin): carry local-only projects on secondmate routes and land approved work in the project's own folder

EternalTime:
## Intent

Let a persistent second mate own a local-only project whose only copy is a folder on this machine, and make an approved landing reach that folder.

The driving case is the captain's job search: a git repository at /Users/owlshome/JobSearch with no remote at all and about 358 MB of personal material (CVs, cover letters, application answers, his address and phone, his referees' contact details) that he does not want on a forge. Before this change bin/fm-home-seed.sh refused it outright ("project <p> is local-only; secondmate routes support only no-mistakes and direct-PR projects") and the secondmate-provisioning skill and AGENTS.md said the same in prose. The captain asked for that limitation to be removed, NOT for his job search to be uploaded anywhere.

What was built:
1. Seeding accepts a local-only project. clone_project and validate_seed_project no longer refuse on mode, and every mode now seeds from the project's own origin, so a second mate's clone is a SIBLING of this home's clone off the same folder rather than a copy of a copy. That sibling shape is what lets a landing reach the folder in one hop.
2. Its crewmates already deliver correctly (implement on fm/<id>, stop with a clean ready branch, no push, no PR): bin/fm-brief.sh's local-only contract needed no change, and deliberately got none.
3. The second mate lands approved work through the existing bin/fm-merge-local.sh under the same merge authority as before.
4. That landing now continues past the clone into the folder that holds the project. This implements the already-queued firstmate work item fm-landing-pull-into-captain-folder rather than inventing a second mechanism; that item records that every prior session did this by hand as `git -C <his folder> pull --ff-only <clone> main`. Git refuses a push into a non-bare repository's checked-out branch, so the landing pulls from the clone into the folder instead.
5. Backlog handoff for local-only items: bin/fm-backlog-handoff.sh turned out to have NO mechanical mode gate at all - the prohibition existed only as prose in the secondmate-provisioning skill. The prose is gone and a new test pins that no mode gate exists, rather than adding then removing code.

Deliberate decisions a reviewer reading only the diff would not know:
[...]

PR #2692, opened 2026-08-20
fix(bin): reconcile treehouse pool path spellings on a symlinked home

Inthuson:
## Intent

Stop firstmate having to hand-edit a recorded worktree path before every crewmate or scout cleanup on this host.

On this machine /home/inthuson is a symlink to /local/home/inthuson, so treehouse records pool worktree paths using its as-written $HOME-rooted spelling (/home/inthuson/...) while bin/fm-spawn.sh records the fully resolved physical path (/local/home/inthuson/...). Treehouse matches its managed inventory by string rather than by resolved path, so "treehouse return" refused with "is not managed by treehouse" and cleanup aborted at bin/fm-teardown.sh:2453 after it had already reaped the worktree processes, leaving the task half cleaned. Nine occurrences across two days, and every cleanup on this host needed a manual rewrite of worktree= in state/<id>.meta first.

Required approach, as specified by the task: reconcile the two spellings rather than picking one, comparing paths after resolving both sides so a symlinked home stops being a special case. Explicitly ruled out: simply switching firstmate to the $HOME spelling without checking every other consumer of worktree=, because the physical path is what the liveness and worktree-isolation checks compare against, and breaking isolation detection to fix a cleanup annoyance would be a bad trade. The new bin/fm-treehouse-lib.sh is the single owner of the reconciliation and only ever offers spellings that provably resolve to the identical physical directory; retry and lock patience policy deliberately stays with the caller. bin/fm-home-seed.sh needed the same reconciliation because the identical defect silently leaked a durable lease during seed rollback.

Explicitly out of scope and deliberately not fixed here: teardown-shared-slot-guard-s2, the separate hazard that cleanup returns a pool slot without checking whether another task holds it.

Verification owed and completed: a real pooled worktree torn down on this host with no hand-editing; and a colocated regression in tests/fm-teardown.test.sh covering the symlinked-home spelling on both sides, since a happy-path-only test would be worthless here when the bug is a comparison that is wrong only when the two spellings differ.

[...]

PR #2705, opened 2026-08-21
fix(bin): fold path spellings in worktree and watcher identity checks

chappellerj:
## Summary

Two safety checks in firstmate's own tooling compared paths as raw strings, so one directory reached under two spellings read as two different locations. Both defeat checks that protect the never-write-to-projects prime directive.

## Defect 1: case-variant home path defeats identity checks

**A finding that departs from the original framing:** the task prescribed normalizing with `pwd -P`, and `fm-spawn` already did that (landed in #294, a month before the incident). `pwd -P` resolves symlinks but **preserves the caller's spelling** of each component:

```
cd /x/Home && pwd -P   ->  /x/Home
cd /x/home && pwd -P   ->  /x/home     # same inode, different string
realpath /x/home       ->  /x/Home     # folds to the on-disk name
```

Perl's `Cwd::realpath` and Python's `os.path.realpath` behave like `pwd -P`, so the new helper deliberately shells out to `realpath(1)` despite an existing perl-realpath idiom elsewhere in the repo.

Reproduced the near-miss: a worktree pointed at the primary clone with a case-variant project path was **ACCEPTED** before the fix, **REFUSED** after.

New `fm_canonical_path` / `fm_canonical_file_path` in `fm-wake-lib.sh` are applied on both sides of the spawn isolation assertion and both watcher-lock identity comparisons (`fm-wake-lib.sh` and `fm-watch-arm.sh` shared the weakness). The guard failure produced false "watcher down" alarms after a session resume.

## Defect 2: pool slot reuse leaves stale meta worktree pointers

`fm-spawn` now records the canonical path the lease actually settled on. `fm-teardown` refuses when the recorded worktree holds a different task's `fm/<id>` branch, before the step that deletes that branch and resets the worktree.

## Deliberate design decisions

- The spawn-side double-lease check **warns rather than refuses**. A lingering record is usually a finished task awaiting cleanup, so refusing would block routine spawns on stale metadata. The destructive end refuses.
- The teardown refusal requires **positive proof**: a detached `HEAD` (how a landed task legitimately ends) or a non-`fm/` branch name does not block cleanup.
- `--force` deliberately does **not** override the refusal: it authorizes discarding this task's own work, never another task's.

## Review round

The review gate caught that the guard covered only the non-orca destructive path while the orca branch ran the identical `checkout --detach` + `branch -D` sequence unguarded. Extended to orca, plus `canonical_existing_dir` (feeding orca's path-match equality check) had the same `pwd -P` weakness and could cause a spurious refusal on a legitimate orca teardown. The hazard class is now closed on every supported backend.

## Verification

- `bin/fm-lint.sh` passes (pinned ShellCheck 0.11.0, actionlint 1.7.12).
- New tests are proven to fail without their fix and pass with it.
- Tests skip on case-sensitive filesystems, where the variant path cannot exist.
- Four pre-existing test failures (`fm-teardown` herdr-preflight, `fm-watch-arm`, `fm-wake-queue`, `fm-backend-orca` metadata-write) were each confirmed failing identically on the clean baseline via `git stash` / `git checkout` — unrelated to this change.

### DIFF
--- a/bin/fm-spawn.sh
+++ b/bin/fm-spawn.sh
@@ -1690,15 +1690,17 @@ BRIEF_REAL="$BRIEF_DIR_REAL/$(basename "$BRIEF")"
 # isolation guard refuses a spawn that never actually tangled). Canonicalize
 # once here so every downstream comparison uses the same physical form
 # (docs/herdr-backend.md "Known gaps").
-PROJ_ABS_REAL=$(cd "$PROJ_ABS" 2>/dev/null && pwd -P) || PROJ_ABS_REAL="$PROJ_ABS"
+#
+# fm_canonical_path (bin/fm-wake-lib.sh) is used rather than `cd … && pwd -P`
+# because pwd -P preserves the CALLER's spelling of each component. On a
+# case-insensitive filesystem a session opened as .../Firstmate and a treehouse
+# path spelled .../firstmate are one directory, and a pwd -P compare calls them
+# different - which let a crewmate be placed directly in the primary clone with
+# no refusal. See fm_canonical_path's header for why realpath(1) is required.
+PROJ_ABS_REAL=$(fm_canonical_path "$PROJ_ABS")
 
 real_path_or_raw() {  # <path>
-  local path=$1 real
-  if real=$(cd "$path" 2>/dev/null && pwd -P); then
-    printf '%s\n' "$real"
-  else
-    printf '%s\n' "$path"
-  fi
+  fm_canonical_path "$1"
 }
 
 # Session-provider container-ensure + task creation. tmux stays exactly as P1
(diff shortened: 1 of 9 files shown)


kunchenguid:
Speaking as Kun's firstmate:


Scheduled firstmate-oss pass. Diff reviewed against main (head `562267bdb16ded3e7a997ebc31d001b22b9abb7c`). No security risk. Approving the pending first-time fork workflow runs.

Classification: **corrective**. Physical-dir identity was comparing caller spelling (`pwd -P`) so a case-variant primary checkout was accepted as an isolated worktree, and watcher lock identity produced false \"watcher down\" after resume. `fm_canonical_path` / `fm_canonical_file_path` fold via `realpath(1)` (documented why Perl/Python realpath and `pwd -P` are insufficient). Isolation assertion, watcher lock home/script compare, spawn `worktree=` recording, and orca `canonical_existing_dir` all use it. Recycled pool-slot teardown now refuses on positive proof of another task's `fm/<id>` branch; `--force` does not override. Spawn double-lease warns rather than refuses (stale meta must not block routine spawns; damage is at the destructive end).

VISION:
- One captain, one interface: **aligns**. Removes false watcher-down noise after resume.
- Authority is explicit and never inferred: **aligns**. Strengthens the never-write-to-primary isolation guard; does not add capability.
- Scripts own the mechanics, agents own the judgment: **aligns**. Path identity is scripted.
- A restart is a non-event: **aligns**. Watcher lock identity survives session-resume spelling drift.
- Delegation with a spine: **aligns**. Isolation refusal and \"unlanded work is never torn down\" both get stronger (foreign `fm/<id>` checkout refused, including under `--force`).
- The fleet outlives any vendor: **aligns**. Applied on tmux/treehouse and orca teardown.
- Scope: **aligns**. Command-layer safety of firstmate's own spawn/teardown/watcher.

Merge-eligible: **YES**. Same physical-dir class as earlier #2692.

Overlap: teardown identity matches only `fm/*`. If #2706's `config/branch-prefix` lands first, a `ryan-fm/<other-id>` recycled slot would not be recognized as foreign. Also both this PR and #2700 patch `bin/fm-spawn.sh`.

kunchenguid (approved):
Speaking as Kun's firstmate: reviewed the full diff. No security risk. Corrective path-identity / pool-slot teardown guards. Approve pending first-time fork CI.

PR #13, opened 2026-06-13
fix: prevent watcher check starvation

kunchenguid:
## Intent

Fix merge-poll starvation in firstmate's watcher (bin/fm-watch.sh). The poll loop evaluated conditions in order signals -> stale -> check-sweep -> heartbeat, and wake() exits the cycle on the first hit. So when a chatty sibling crewmate kept producing status/turn-end signals, every cycle exited on the signal before reaching the check sweep, and the slow per-task polls (e.g. merged-PR detection) were starved indefinitely - observed live: a merged PR went unsurfaced while a sibling task was active. Fix: move the due-check sweep to run right after the liveness beacon, before the signal scan, so due checks run before signals can preempt them. Checks are due only every CHECK_INTERVAL (300s), so most cycles skip the block and fall straight through to the signal scan as before; signals are never lost (persisted .seen-* signatures catch any that land while the watcher is between cycles). Pure reordering of existing blocks plus an explanatory comment; no new behavior.

## What Changed

- Run due watcher check scripts before signal scanning so chatty crewmate status updates cannot indefinitely preempt merge-poll checks.
- Add `FM_CHECK_TIMEOUT` and bounded check execution so slow or hung check scripts cannot block later watcher signals.
- Document the new check ordering and timeout contract in `AGENTS.md` and `README.md`.

## Risk Assessment

✅ Low: The change is narrowly scoped to watcher check scheduling and bounded check execution, with no material correctness issues found in the reviewed diff.

## Testing

Exercised the actual watcher entry point in a controlled merge-poll-starvation scenario, captured the CLI transcript as reviewer-visible evidence, confirmed due checks preempt chatty signals while signals persist for the next cycle, and cleaned up transient `state/` files afterward.

<details>
<summary>Evidence: Watcher check-before-signal transcript</summary>

<code>Output 1: check: .../state/merge-poll.check.sh: merged-pr-detected&#10;Output 2: signal: .../state/chatty.status</code>

```text
Scenario: due merge-poll check and pending chatty status signal both exist before watcher starts.
Command 1: FM_POLL=1 FM_CHECK_INTERVAL=0 FM_CHECK_TIMEOUT=5 FM_SIGNAL_GRACE=0 FM_HEARTBEAT=999999 bin/fm-watch.sh
Output 1: check: /Users/kunchen/.no-mistakes/worktrees/016d88035d58/01KV1957JM706HH26XY4EFQ1YM/state/merge-poll.check.sh: merged-pr-detected
[...]

PR #12, opened 2026-06-13
fix: wake pi crewmates after each turn

kunchenguid:
## Intent

Fix firstmate's pi crewmate turn-end hook. fm-spawn's generated pi extension listened on pi's 'agent_end' event, which (per pi 0.79.1 type docs) fires only once when the whole agent run exits - so the turn-end backstop never fired at per-turn boundaries for pi crewmates, leaving firstmate reliant on crewmate status appends alone and blind to an idle crewmate that stopped without writing status. Change the hook to 'turn_end', which fires after each turn the agent finishes, matching how the claude/codex/opencode adapters behave. One-line change to the pi-ext template in bin/fm-spawn.sh plus an explanatory comment; no other behavior change.

## What Changed
- Updated the pi harness extension generated by `bin/fm-spawn.sh` to listen for `turn_end` instead of `agent_end`, so firstmate receives a turn-ended signal after each completed pi turn rather than only when the whole agent run exits.
- Added an inline comment explaining why the pi hook must use `turn_end` for per-turn idle detection.
- Documented the pi adapter requirement in `AGENTS.md` so future harness changes preserve the same watcher behavior.

## Risk Assessment

✅ Low: The branch is a narrow one-line event-name correction plus explanatory generated-code comments, with no broader control-flow or API surface changes.

## Testing

Baseline bootstrap could not run due missing local firstmate config state, but the focused end-to-end spawn path was exercised: `fm-spawn.sh` generated and launched a pi extension using `turn_end`; live pi execution stopped at the trust prompt, which I did not accept because it would change user-level trust outside the worktree boundary, so I supplemented it with a mocked extension event simulation proving the generated hook would touch the turn-ended sentinel. All focused checks passed and temporary worktree test scaffolding was removed.

<details>
<summary>Evidence: Generated pi extension evidence</summary>

<code>The emitted extension registers `pi.on(&#34;turn_end&#34;, ...)` and touches `state/pi-hook-e2e.turn-ended`; it contains no `pi.on(&#34;agent_end&#34;, ...)` registration.</code>

```text
// Firstmate turn-end signal; written by fm-spawn.
// Use "turn_end" (fires after each turn the agent finishes), not "agent_end"
// (fires once, only when the whole run exits): the watcher needs a signal at
// every turn boundary so an idle crewmate is surfaced, not just at shutdown.
import { execFile } from "node:child_process";
export default function (pi: any) {
[...]

PR #11, opened 2026-06-13
docs: require full PR URLs in captain reports

kunchenguid:
## Intent

Document a firstmate reporting convention the captain requested: when reporting a PR to the captain, always give the PR's full https URL, never a bare #number, because the captain's terminal makes full URLs clickable. Docs-only change to AGENTS.md - §7 (PR-ready: the crewmate report to the captain) and §9 (captain etiquette). No code, no behavior change; purely a communication-style rule for firstmate.

## What Changed

- Updated firstmate PR-ready instructions to require complete `https://...` PR URLs instead of bare `#number` references when reporting to the captain.
- Clarified ship teardown and backlog examples to record full GitHub PR URLs.
- Added captain-etiquette guidance allowing shorthand PR numbers only after the full URL appears in the same message.

## Risk Assessment

✅ Low: The change is docs-only, narrowly scoped to firstmate PR reporting guidance, and introduces no behavior or code-path risk.

## Testing

For this docs-only change, I verified the target commit changes only AGENTS.md, inspected the two requested documentation sections, confirmed the end-user-facing firstmate convention now requires complete https PR URLs instead of bare #numbers, and saved a focused diff artifact as reviewer evidence; no runtime tests were applicable.

<details>
<summary>Evidence: Focused AGENTS.md diff showing full PR URL convention</summary>

[...]

PR #2725, opened 2026-08-21
fix(dashboard): require a real ask before a card can claim needs-attention

joliverMI:
## Intent

Make needs-attention on the Admiral's Fleet Dashboard structurally require an actual ask. The Admiral caught a card set to needs-attention with reason 'You reported flares not changing the lights - being chased now' - a progress report, not something he could act on. needs-attention is the loudest status on the board and means blocking on him; a card that merely reports progress spends his attention and teaches him the loud status does not always mean him, repeating a failure that already made one always-red marker into noise.

Scope delivered:
1. bin/fm-dashboard.sh and the HTTP API (both create-task and set-status/patch paths): setting needs_attention now REQUIRES a --reason / reason field. Previously it was optional and the UI fell back to generic placeholder text ('Needs his decision or action to continue.'). The API now refuses the status change without a reason rather than substituting a default; the CLI also refuses locally before the round-trip.
2. Reject reasons that are obviously not an ask, mechanically and conservatively - not a language model. Implemented as a small fixed list of report-shaped phrases (REPORT_SHAPED_PHRASES in bin/fleet-dashboard/server/validation.py: 'being chased', 'in progress', 'investigating', 'looking into', 'working on', 'keeping an eye on', 'monitoring', 'no update yet', 'will update', 'update to follow', 'tracking down', 'digging into', 'following up on', 'under investigation', 'still chasing'), matched as a case-insensitive substring anywhere in the reason text (not just at the start, since the motivating example tacked the report phrase onto the end of a sentence). Documented in docs/dashboard.md with the tested false-negative rate: caught 15/15 report-shaped reasons built from the list, missed 10/10 reworded report-shaped reasons deliberately avoiding every listed phrase (expected blind spot - it recognizes vocabulary, not the general concept of a report), and produced zero false positives against 10 genuine asks.
3. Explicitly did NOT block on any other status: working with a vague note is not touched, review needs no ask by design and is untouched. Only needs_attention is guarded.
4. Updated .agents/skills/fleet-dashboard/SKILL.md: reason is required, and the test for a real ask is 'what would he DO' - if the honest answer is 'read it' or 'know it', it is not needs-attention.
5. Updated docs/dashboard.md with a new 'The needs-attention reason guard' section covering both enforcement points, the false-negative rate, and an explicit statement that the fleet auditor's separate judgment-based check (whether the ask is genuinely real, not just non-report-shaped) is not replaced by this mechanical guard and vice versa - the server checks words, the auditor checks whether the ask is real.

Constraints honored: existing needs-attention cards with weak reasons are untouched - migrates nothing, only governs what can be set from now on. The board's live server keeps working (server-side change is additive validation, not a breaking schema change). Added tests/fm-dashboard.test.sh coverage exercising both the CLI-local rejection and the server-side rejection independently (direct curl call with empty reason), plus the report-shaped-phrase rejection and successful creation/transition with a genuine ask, for both the add and status code paths.

## What Changed

- `needs_attention` now requires a reason server-side: `POST /api/tasks` (create-into-status) and `POST /api/tasks/{id}/status` both route the reason through a new `validate_needs_attention_reason` in `bin/fleet-dashboard/server/validation.py` and return 400 instead of falling back to placeholder text; `store.add_task` gained a `needs_attention_reason` column write, scoped to that status only. `bin/fm-dashboard.sh` refuses locally in both `add` and `status` before the round-trip, and `add --reason` is now refused for any status that cannot store one.
- Reasons that read as progress reports are rejected mechanically against a fixed `REPORT_SHAPED_PHRASES` list ("being chased", "in progress", "investigating", …), matched only when a phrase opens or closes the leading or trailing clause — with short leading-hedge ("still", "we") and trailing-hedge ("now", "it") lists — so a report phrase used as an ordinary noun mid-clause ("approve the $400 monitoring subscription renewal") still passes. The card overlay in `web/app.js` opens a reason field when `needs-attention` is chosen, keeps confirm disabled until it has text, and surfaces the server's refusal inline.
- Added `tests/fm-dashboard.test.sh` coverage for CLI-local and direct-curl server rejection, report-shaped-phrase rejection, genuine asks containing report words, and a corpus test asserting the documented catch/miss/false-positive counts (15/15 caught, 18 accepted misses, 0 false positives on 14 asks). `docs/dashboard.md` gains a "needs-attention reason guard" section covering both enforcement points, the accepted blind spots and why widening the hedge list was reverted, and the boundary with the fleet auditor's judgment check; `SKILL.md` records that the reason is required and the "what would he DO" test. `--help` now derives its header range from the comment block instead of a fixed line count.

## Risk Assessment

✅ Low: The guard sits at the single shared server boundary with no reachable bypass (PATCH cannot set status, no other caller touches set_status/add_task), the documented catch/miss/false-positive rates reproduce exactly against the current code and are pinned by an executed table-driven test, and the remaining known misses and the reason-persistence gap for non-waiting statuses are explicitly recorded as accepted rather than left implicit.

## Testing

Ran the targeted dashboard suite (tests/fm-dashboard.test.sh, all 17 checks pass) plus the two adjacent suites touching the same surfaces (card-link, documentation-audiences), then verified the intent end-to-end against a live server on an isolated FM_HOME: the CLI, the raw HTTP API on both the create and status paths, and the board UI each refuse needs-attention without a reason and refuse the exact 'being chased now' reason that motivated the change, while a genuine ask goes through and renders on the card; a pre-guard card with the old placeholder text is left untouched and every other status still needs no ask. Captured a CLI transcript and three UI screenshots (required-reason prompt, inline server refusal, board showing only actionable asks). Everything passed; the only note is that the shipped matcher is clause-edge rather than anywhere-substring, narrower than the intent text describes but documented and pinned by tests.

<details>
<summary>Evidence: CLI transcript: needs-attention reason guard end-to-end (refusals, genuine ask, direct HTTP calls, pre-guard card untouched, other statuses unaffected)</summary>

Source: [CLI transcript: needs-attention reason guard end-to-end (refusals, genuine ask, direct HTTP calls, pre-guard card untouched, other statuses unaffected)](https://github.com/joliverMI/firstmate/blob/4ee7f1c0ad909d4a57efa81c9b81c33161508666/.no-mistakes/evidence/fm/fm-needs-attention-requires-an-ask/needs-attention-guard-cli.txt)

```text
=== The card the Admiral caught: a progress report in the needs-attention reason ===

(created card anchor-light-flickering-on-the-foredeck-f3lg in not-started)

$ /home/joliv/.no-mistakes/worktrees/4cc5c0885385/01M0GZDC24S8JFM230NDS5TAZW/bin/fm-dashboard.sh status anchor-light-flickering-on-the-foredeck-f3lg needs-attention --reason You reported flares not changing the lights - being chased now
fm-dashboard.sh: server refused (400): that reason reads as a progress report, not something he can act on (matched 'being chased' at the start or end of a clause) - say what he needs to decide, approve, or supply: 'You reported flares not changing the lights - being chased now'
  [exit 1]

=== Same status with no reason at all: also refused (was previously allowed, and the UI filled in 'Needs his decision or action to continue.') ===

$ /home/joliv/.no-mistakes/worktrees/4cc5c0885385/01M0GZDC24S8JFM230NDS5TAZW/bin/fm-dashboard.sh status anchor-light-flickering-on-the-foredeck-f3lg needs-attention
fm-dashboard.sh: status: needs-attention requires --reason - say what he needs to decide, approve, or supply
  [exit 1]

=== The server refuses it too, not just the CLI - direct HTTP call, bypassing bin/fm-dashboard.sh entirely ===

$ curl -sS -i -X POST http://127.0.0.1:$PORT/api/tasks/anchor-light-flickering-on-the-foredeck-f3lg/status -d '{"status":"needs_attention"}'
HTTP/1.0 400 Bad Request
Server: FleetDashboard/1.0 Python/3.12.3
Date: Fri, 21 Aug 2026 02:42:51 GMT
Content-Type: application/json
Content-Length: 115

{"error": "needs-attention requires --reason - say what he needs to decide, approve, or supply, not what happened"}
$ curl -sS -i -X POST .../status -d '{"status":"needs_attention","reason":"still chasing the supplier"}'
HTTP/1.0 400 Bad Request
Server: FleetDashboard/1.0 Python/3.12.3
Date: Fri, 21 Aug 2026 02:42:51 GMT
Content-Type: application/json
Content-Length: 218

{"error": "that reason reads as a progress report, not something he can act on (matched 'still chasing' at the start or end of a clause) - say what he needs to decide, approve, or supply: 'still chasing the supplier'"}
=== A real ask - something he can DO - goes through and lands on the card ===

$ /home/joliv/.no-mistakes/worktrees/4cc5c0885385/01M0GZDC24S8JFM230NDS5TAZW/bin/fm-dashboard.sh status anchor-light-flickering-on-the-foredeck-f3lg needs-attention --reason pick the replacement fixture: chandlery has bronze at $180 or steel at $95
anchor-light-flickering-on-the-foredeck-f3lg	needs_attention	firstmate	-	Anchor light flickering on the foredeck
  [exit 0]

$ /home/joliv/.no-mistakes/worktrees/4cc5c0885385/01M0GZDC24S8JFM230NDS5TAZW/bin/fm-dashboard.sh show anchor-light-flickering-on-the-foredeck-f3lg
id:       anchor-light-flickering-on-the-foredeck-f3lg
title:    Anchor light flickering on the foredeck
status:   needs_attention
captain:  firstmate
agent:    
starred:  false
needs attention: pick the replacement fixture: chandlery has bronze at $180 or steel at $95

--- prompt ---
The flares aren't changing the lights.

  [exit 0]

=== A genuine ask that merely mentions a report word mid-sentence is NOT refused ===

$ /home/joliv/.no-mistakes/worktrees/4cc5c0885385/01M0GZDC24S8JFM230NDS5TAZW/bin/fm-dashboard.sh status anchor-light-flickering-on-the-foredeck-f3lg working
anchor-light-flickering-on-the-foredeck-f3lg	working	firstmate	-	Anchor light flickering on the foredeck
  [exit 0]

$ /home/joliv/.no-mistakes/worktrees/4cc5c0885385/01M0GZDC24S8JFM230NDS5TAZW/bin/fm-dashboard.sh status anchor-light-flickering-on-the-foredeck-f3lg needs-attention --reason approve the $400 monitoring subscription renewal
anchor-light-flickering-on-the-foredeck-f3lg	needs_attention	firstmate	-	Anchor light flickering on the foredeck
  [exit 0]

=== Creating a card straight into needs-attention is governed the same way ===

$ /home/joliv/.no-mistakes/worktrees/4cc5c0885385/01M0GZDC24S8JFM230NDS5TAZW/bin/fm-dashboard.sh add --title Haul-out date --captain firstmate --prompt yard called --status needs-attention
fm-dashboard.sh: add: --status needs-attention requires --reason - say what he needs to decide, approve, or supply
  [exit 1]

$ /home/joliv/.no-mistakes/worktrees/4cc5c0885385/01M0GZDC24S8JFM230NDS5TAZW/bin/fm-dashboard.sh add --title Haul-out date --captain firstmate --prompt yard called --status needs-attention --reason looking into the yard's two open slots
fm-dashboard.sh: server refused (400): that reason reads as a progress report, not something he can act on (matched 'looking into' at the start or end of a clause) - say what he needs to decide, approve, or supply: "looking into the yard's two open slots"
  [exit 1]

$ /home/joliv/.no-mistakes/worktrees/4cc5c0885385/01M0GZDC24S8JFM230NDS5TAZW/bin/fm-dashboard.sh add --title Haul-out date --captain firstmate --prompt yard called --status needs-attention --reason choose Nov 3 or Nov 17 for the haul-out - yard needs an answer today
haul-out-date-f9vo	needs_attention	firstmate	-	Haul-out date
  [exit 0]

=== Other statuses are untouched: working with a vague note, and review/waiting, still need no ask ===

$ /home/joliv/.no-mistakes/worktrees/4cc5c0885385/01M0GZDC24S8JFM230NDS5TAZW/bin/fm-dashboard.sh status anchor-light-flickering-on-the-foredeck-f3lg working
anchor-light-flickering-on-the-foredeck-f3lg	working	firstmate	-	Anchor light flickering on the foredeck
  [exit 0]

$ /home/joliv/.no-mistakes/worktrees/4cc5c0885385/01M0GZDC24S8JFM230NDS5TAZW/bin/fm-dashboard.sh status anchor-light-flickering-on-the-foredeck-f3lg waiting --reason waiting on the chandlery to call back
anchor-light-flickering-on-the-foredeck-f3lg	waiting	firstmate	-	Anchor light flickering on the foredeck
  [exit 0]

$ /home/joliv/.no-mistakes/worktrees/4cc5c0885385/01M0GZDC24S8JFM230NDS5TAZW/bin/fm-dashboard.sh list
anchor-light-flickering-on-the-foredeck-f3lg	waiting	firstmate	-	Anchor light flickering on the foredeck
haul-out-date-f9vo	needs_attention	firstmate	-	Haul-out date
  [exit 0]


=== Existing cards are not migrated: a needs-attention card written before the guard (with the old placeholder reason) still stands ===

(row inserted straight into the board's database, as a pre-guard card would be)
$ fm-dashboard.sh show legacy-weak-card
id:       legacy-weak-card
title:    Bilge pump wiring
status:   needs_attention
captain:  firstmate
agent:    
starred:  false
needs attention: Needs his decision or action to continue.

--- prompt ---
His words.


$ fm-dashboard.sh list --status needs-attention
anchor-light-flickering-on-the-foredeck-f3lg	needs_attention	firstmate	-	Anchor light flickering on the foredeck
haul-out-date-f9vo	needs_attention	firstmate	-	Haul-out date
legacy-weak-card	needs_attention	firstmate	-	Bilge pump wiring

=== Every other status still needs no ask at all ===

$ fm-dashboard.sh status varnish-the-coaming-p6h8 working
varnish-the-coaming-p6h8	working	captain_river	-	Varnish the coaming
  [exit 0]
$ fm-dashboard.sh status varnish-the-coaming-p6h8 paused
varnish-the-coaming-p6h8	paused	captain_river	-	Varnish the coaming
  [exit 0]
$ fm-dashboard.sh status varnish-the-coaming-p6h8 testing
varnish-the-coaming-p6h8	testing	captain_river	-	Varnish the coaming
  [exit 0]
$ fm-dashboard.sh status varnish-the-coaming-p6h8 complete
varnish-the-coaming-p6h8	complete	captain_river	-	Varnish the coaming
  [exit 0]
$ fm-dashboard.sh status varnish-the-coaming-p6h8 not-started
varnish-the-coaming-p6h8	not_started	captain_river	-	Varnish the coaming
  [exit 0]
```
</details>
- Evidence: [UI: setting Needs Attention now demands the ask first — submit disabled until a reason is typed](https://github.com/joliverMI/firstmate/blob/4ee7f1c0ad909d4a57efa81c9b81c33161508666/.no-mistakes/evidence/fm/fm-needs-attention-requires-an-ask/ui-1-reason-required.png)
- Evidence: [UI: the Admiral&#39;s actual example refused inline, status stays Waiting](https://github.com/joliverMI/firstmate/blob/4ee7f1c0ad909d4a57efa81c9b81c33161508666/.no-mistakes/evidence/fm/fm-needs-attention-requires-an-ask/ui-2-report-shaped-refused.png)
- Evidence: [UI: board&#39;s NEEDS ATTENTION section, both cards carrying something he can act on](https://github.com/joliverMI/firstmate/blob/4ee7f1c0ad909d4a57efa81c9b81c33161508666/.no-mistakes/evidence/fm/fm-needs-attention-requires-an-ask/ui-3-board-with-real-asks.png)
<details>
<summary>Evidence: Server refusal, straight from the API (no CLI involved)</summary>

```text
$ curl -sS -i -X POST http://127.0.0.1:$PORT/api/tasks/<id>/status -d '{"status":"needs_attention"}'
HTTP/1.0 400 Bad Request
{"error": "needs-attention requires --reason - say what he needs to decide, approve, or supply, not what happened"}

$ curl -sS -i -X POST .../status -d '{"status":"needs_attention","reason":"still chasing the supplier"}'
HTTP/1.0 400 Bad Request
{"error": "that reason reads as a progress report, not something he can act on (matched 'still chasing' at the start or end of a clause) - say what he needs to decide, approve, or supply: 'still chasing the supplier'"}
```
</details>
- Outcome: ⚠️ 1 info across 1 run (5m12s)

## Pipeline

<details>

✅ No issues found.

</details>

<details>

- ⚠️ `.agents/skills/afk/SKILL.md` - branch carries 13 commit(s) that exist on your local main branch but were never pushed to origin/main; rebasing would bundle this unrelated work (140 file(s)) into the PR:
- b98e098 fix(watcher): re-evaluate arm coalescing when the fleet lock changes (#14)
- 45bd292 fix(bin): count a trust-registered custom check as supervision-needed (#13)
- 882004e feat(dashboard): link handed-off backlog items to their card (#11)
- b19134b fix(bin): resolve tmux endpoint targets exactly for liveness probes and sends (#12)
- 3c8f796 feat(dashboard): mechanically link spawned/torn-down tasks to their card (#9)
- e242ce4 feat(dashboard): wire the fleet auditor&#39;s timer to real cron, add a Force Audit button (#8)
- f23b7a4 feat(dashboard): add needs-attention status, distinct from testing (#7)
- 3a972e3 feat(dashboard): add the Admiral&#39;s Fleet Dashboard task board (#6)
- 4f0a4f5 fix(bin): give every status board pointer a real link or an honest gap (#5)
- e9e3d0c feat(bin): add a Ready-for-you-to-look-at section to the status board (#4)
- c67c73d fix(procevent): deliver captured results at most once (#3)
- 2f973eb fix(bin): refuse teardown while a task&#39;s recorded PR is still open (#1)
- a93dae5 feat(bin): render the captain&#39;s four-section status board from existing state (#2)

</details>

<details>

- ⚠️ `tests/fm-dashboard.test.sh:148` - The create-path server guard is never exercised. On the `add` path the test only covers (a) the CLI&#39;s local empty-reason refusal and (b) a successful create with a genuine ask — both of which still pass if `api.py:111-115` (the `validate_needs_attention_reason` call in `create_task`) is deleted outright, because the CLI blocks the empty case locally and the genuine ask is accepted either way. There is no report-shaped-reason case for `add`, and no direct curl to `POST /api/tasks` with an empty reason, even though the `status` path got exactly that treatment (line 130-134). The intent states tests were added exercising &#34;the CLI-local rejection and the server-side rejection independently (direct curl call with empty reason), plus the report-shaped-phrase rejection and successful creation/transition with a genuine ask, for both the add and status code paths&#34; — the add path is missing two of those four. Concretely: add `&#34;$DASH&#34; add ... --status needs-attention --reason &#34;looking into the checkout timeout&#34;` expecting a non-zero exit, and a `curl -X POST /api/tasks` with `{&#34;status&#34;:&#34;needs_attention&#34;}` and no reason expecting HTTP 400.
- ℹ️ `bin/fleet-dashboard/server/validation.py:43` - Several REPORT_SHAPED_PHRASES entries are ordinary nouns/verb phrases that occur inside genuine asks, and the match is an unanchored case-insensitive substring, so the guard hard-refuses them. &#34;Approve the $400 monitoring subscription renewal&#34; and &#34;Decide whether to keep the monitoring alerts on overnight&#34; both match `monitoring`; &#34;Pick which contractor keeps working on the deck&#34; matches `working on`; &#34;Approve the invoice for the in progress work&#34; matches `in progress`. Each of those is a real blocking ask that the API will refuse with HTTP 400. The rejection message does name the matched phrase, so an agent can reword — but the failure mode when it does not is that the card silently stays in `working` and the Admiral is never asked, which is the inverse of the failure this guard exists to prevent. Flagging because the fixed list is a deliberate design choice (documented in docs/dashboard.md), not something to change unilaterally: worth confirming whether `monitoring`, `working on`, and `in progress` earn their place given they are the three most likely to appear in a legitimate ask.
- ℹ️ `docs/dashboard.md:101` - docs/dashboard.md:101-103 makes specific quantitative claims about the guard — caught 15/15 report-shaped reasons, missed 10/10 reworded ones, zero false positives against 10 genuine asks — but no check in the repo reproduces them. Nothing fails when someone adds, removes, or narrows a REPORT_SHAPED_PHRASES entry, so those numbers can silently become false while still reading as measured fact. That matters more than usual here because the same section is what tells the fleet auditor the exact shape of the blind spot it is supposed to compensate for. A table-driven test that calls validate_needs_attention_reason over the three corpora (rejects the report-shaped set, accepts the genuine-ask set) would pin the claim to executed behavior rather than prose.
- ℹ️ `bin/fm-teardown.sh:989` - Pre-existing on this branch, not introduced by the reviewed commit, but it undercuts the same signal the change is protecting: teardown advances any non-`complete` linked card to `testing`, with `needs_attention` given no special treatment. store.set_status then nulls `needs_attention_reason` (store.py:306), so a card the Admiral was actually blocking on is silently demoted to the status that is explicitly &#34;optional to him&#34; and its ask is discarded — no audit-log entry, no trace on the card beyond a status_history row. Sequence: card set to needs-attention with a genuine ask, task later torn down normally, card lands in testing with the ask gone. Worth deciding whether teardown should skip `needs_attention` the way it skips `complete`, or log a discrepancy when it overrides one.
- ℹ️ `bin/fleet-dashboard/web/app.js:145` - The generic fallback &#34;Needs his decision or action to continue.&#34; is still rendered on the card banner. This is consistent with the intent&#39;s constraint that existing needs-attention cards are not migrated — pre-guard rows can legitimately have a NULL reason and still need something to render — so it is correct to keep, not a leftover. Noting it only because the intent describes that placeholder as the old behavior being replaced, and a later reader may mistake its presence for an incomplete cleanup; it is now reachable only for cards created before this guard existed.

🔧 Fix: anchor needs-attention report guard to clause edges, test rates
- ⚠️ `bin/fleet-dashboard/server/validation.py:70` - The fix round&#39;s edge anchoring introduced a false-negative class that the documented corpus does not cover: a listed report phrase preceded by a subject and a copula now passes. `_sits_at_an_edge` strips only `_LEADING_FILLER` ({still, currently, now, just, already, we, we&#39;re, i, i&#39;m, im}) from the front, so on &#34;We are investigating the checkout timeout&#34; the leading scan strips `we`, hits `are` (not in the set) and breaks, while the trailing scan breaks on `timeout` - the reason is accepted. Same for &#34;The team is looking into the failed backup&#34; (leading breaks immediately on `the`) and &#34;Work in progress on the deck&#34; (trailing breaks on `deck`). Concrete reachable path: `bin/fm-dashboard.sh status &lt;id&gt; needs-attention --reason &#34;We are investigating the checkout timeout&#34;` succeeds and renders a pure progress report in the loudest status on the board - exactly the failure the change exists to prevent, and a case the author&#39;s original substring match caught. This matters more than an ordinary blind spot because docs/dashboard.md:101 and tests/fm-dashboard.test.sh&#39;s REPORT_SHAPED corpus present &#34;caught all 15&#34; as measured evidence, and all 15 examples happen to place the phrase at a hedge-strippable edge; not one has the subject+copula shape, so the stated rate overstates what the guard achieves. The narrow mechanical scope and the fleet auditor backstop (SKILL.md sweep step 4) are deliberate and I am not asking for a redesign - but either extend `_LEADING_FILLER` with the copulas/auxiliaries (is, are, am, was, were, be, been, &#39;s, &#39;re) that cover the common case, or add at least one subject+copula example to the REPORT_SHAPED corpus and state this shape as a known miss so the documented catch rate is honest about it. Related, smaller: `_edge_clauses` (line 102) keeps only the first and last clause, so a report clause sandwiched in the middle (&#34;Approve overtime - still investigating the delay - thanks&#34;) is skipped entirely; that one is at least described in the docs.
- ℹ️ `bin/fleet-dashboard/server/validation.py:45` - The authoritative intent text no longer describes the shipped implementation, in three source-verifiable ways. (1) It states the phrases are &#34;matched as a case-insensitive substring anywhere in the reason text (not just at the start, since the motivating example tacked the report phrase onto the end of a sentence)&#34;; the code now matches only when a phrase opens or closes the leading or trailing clause (`find_report_shaped_phrase`, line 130). (2) It enumerates the list as ending in &#34;...&#39;keeping an eye on&#39;, &#39;monitoring&#39;, &#39;no update yet&#39;...&#34;; `REPORT_SHAPED_PHRASES` at line 45 has no bare `monitoring` - it was split into `monitoring the` and `monitoring it` (16 entries, not 15). (3) It records &#34;produced zero false positives against 10 genuine asks&#34;; docs/dashboard.md:103 and the pinned test corpus now say 12 genuine asks. All three changes were explicitly directed by your round-1 fix instructions, so this is the intent/PR narrative being stale rather than the code deviating from what you last asked for - flagging only because I cannot resolve an intent-vs-code divergence myself. Suggest updating the intent/PR description text to describe edge-anchored matching, the split `monitoring the`/`monitoring it` entries, and the 12-ask corpus.

🔧 Fix: catch copula-led report reasons, document noun-subject miss
- ⚠️ `bin/fleet-dashboard/server/validation.py:87` - Adding bare copulas to `_LEADING_FILLER` introduced a new false-positive class on genuine asks. `_sits_at_an_edge` strips filler tokens unconditionally from the front, so any reason that *opens* with a copula now has it stripped, and the report phrase immediately behind it matches. Verified against the current code vs. the pre-e367b79 code: &#34;Is working on Saturday ok with you?&#34;, &#34;Is digging into the crawlspace approved?&#34;, &#34;Is monitoring the pool worth $80 a month?&#34;, and &#34;Was looking into the second quote worth the delay?&#34; were all ACCEPTED before this commit and are all REFUSED now. Each is a real blocking ask phrased as a question - the most natural form for a decision request - and refusing it leaves the card silently in `working` with the Admiral never asked, which docs/dashboard.md:97 itself names as &#34;the exact inverse of the failure this guard exists to prevent.&#34; The intended case is subject+copula (&#34;We are investigating X&#34;), not a bare leading copula; requiring that a copula only be stripped when a first-person subject was already stripped ahead of it keeps &#34;We are investigating the checkout timeout&#34; / &#34;I was digging into the bounced payouts&#34; caught while leaving &#34;Is working on Saturday ok?&#34; alone. The pinned corpus cannot catch this: GENUINE_ASKS in tests/fm-dashboard.test.sh has no question-form entry, so the documented &#34;zero false positives against 12 genuine asks&#34; stays green while the residual FP surface grew. Flagging as ask-user because the copula list was your explicit round-2 instruction - the narrowing to subject+copula is the part that needs your call.
- ⚠️ `bin/fleet-dashboard/server/validation.py:91` - The `&#34;&#39;s&#34;` and `&#34;&#39;re&#34;` entries added to `_LEADING_FILLER` are unreachable, so the contracted-copula gap the fix round claims to have closed is still open. `_tokens` (line 121) tokenizes with `[a-z0-9&#39;]+`, which keeps the apostrophe attached to the preceding word: &#34;It&#39;s&#34; produces the single token `it&#39;s`, never `it` + `&#39;s`. A standalone `&#39;s`/`&#39;re` token can only arise from text like &#34;we &#39;s&#34;, which does not occur. Consequence: &#34;It&#39;s still chasing the supplier&#34; - the exact string your round-2 instruction named as one of the two cases &#34;wrongly ACCEPTED&#34; and being fixed - is still accepted (verified against the current code), as is &#34;It&#39;s investigating the leak&#34;. The fix round also did not add it to the REPORT_SHAPED corpus, so nothing in the test suite would reveal that the case was missed; only &#34;We are investigating...&#34; / &#34;I was digging into...&#34; / &#34;We were keeping an eye on...&#34; were added, all of which pass via the pre-existing `we`/`i` entries plus the new bare copulas. Two separable parts: (a) `&#39;s` and `&#39;re` are dead entries and should be removed regardless, and (b) catching &#34;It&#39;s still chasing X&#34; requires an `it&#39;s`/`it` token, which is a pronoun and therefore outside the closed set you explicitly ruled out - so either accept it as a documented miss alongside the noun-subject class in docs/dashboard.md, or authorize adding the third-person contraction. Needs your call, not a unilateral fix.
- ℹ️ `bin/fleet-dashboard/server/validation.py:52` - Splitting bare `monitoring` into `monitoring the` / `monitoring it` (round 1) leaves a miss class that no corpus or doc records. Verified: &#34;Monitoring the disk usage&#34; and &#34;Monitoring it now&#34; are refused, but &#34;Monitoring disk usage&#34; and &#34;Still monitoring disk usage&#34; are accepted - plain progress reports opening on a listed report verb. The determiner is the only thing separating them. docs/dashboard.md:101 lists &#34;Monitoring the alert queue overnight&#34; in the caught-18 corpus, which reads as &#34;monitoring reports are caught&#34;, and the two documented miss classes at line 102-104 are reworded phrasing and subject-noun forms - neither covers this. Note the original motivation for the split no longer applies the same way: the false positives it was trimming for (&#34;approve the $400 monitoring subscription renewal&#34;, &#34;decide whether to keep the monitoring alerts on overnight&#34;) are mid-clause and are already excluded by the edge anchoring added in the same round, so bare `monitoring` may now be safe at the trailing edge even if a leading noun-adjunct use (&#34;Monitoring costs need your approval&#34;) argues for keeping the split at the leading edge. Either way, worth either restoring coverage or adding this to the documented blind-spot list so the 18/18 figure is not read as covering all monitoring-shaped reports.

🔧 Fix: revert copula hedges, document accepted guard misses
2 infos still open:
- ℹ️ `bin/fm-dashboard.sh:207` - `add` now accepts `--reason` for any starting status, but the value is silently discarded unless the status is needs_attention: `create_task` (api.py:110) passes it only as `needs_attention_reason`, and `store.add_task` (store.py:211) nulls it when `status != &#34;needs_attention&#34;` while hardcoding `waiting_on_id`/`waiting_reason` to NULL in the INSERT. Concretely: `fm-dashboard.sh add --title X --captain dj --prompt &#34;...&#34; --status waiting --reason &#34;waiting on the plumber&#34;` exits 0, prints a normal row, and the card carries no waiting reason - whereas the sibling `fm-dashboard.sh status &lt;id&gt; waiting --reason &#34;...&#34;` honors exactly that flag. Before this change `--reason` on `add` was a hard `add: unknown argument` error, so the change converted a loud failure into a silent data drop for that combination. The usage comment at line 19 does say &#34;ignored for every other starting status,&#34; so this looks deliberate rather than accidental - flagging only because the asymmetry with `status` is the kind a caller learns the wrong way. Cheapest options if you want it closed: `die` when `--reason` is passed with a status that ignores it, or plumb `waiting_reason` through `add_task` the way `set_status` does.
- ℹ️ `bin/fleet-dashboard/server/api.py:110` - The needs_attention guard is the same six lines in `create_task` (110-115) and `set_status` (169-174) - identical `reason = body.get(&#34;reason&#34;) or None` plus try/except translating InvalidReasonError to ApiError(400). A one-line helper would dedupe it. Noting rather than asking for a change: this matches the existing convention in the same file (`add_note` inlines the `validate_review_link` try/except the same way), so the duplication is consistent with its surroundings rather than a new pattern, and both copies are correct.

🔧 Fix: refuse add --reason for statuses that cannot carry one
4 infos still open:
- ℹ️ `docs/dashboard.md:125` - The residual false-positive class is wider than the single example the docs give, and includes the question form the round-3 revert was made to protect. Verified against the current code: &#34;Do you want us to keep working on this?&#34;, &#34;Should we keep working on this?&#34;, &#34;Which vendor should we keep working on?&#34;, &#34;Do you want us to stop investigating?&#34;, &#34;Decide if we should keep monitoring it&#34; are all REFUSED (trailing edge, after the trailing-filler strip of &#34;this&#34;/&#34;it&#34;), as are real asks carrying a report clause in a separate sentence - &#34;Investigating the outage. Should we call the vendor?&#34;, &#34;Still investigating - do you want us to call the vendor tonight?&#34;, &#34;Approve the overtime; the crew is in progress&#34;. Each is a genuine blocking ask, and refusing one leaves the card in `working` with the Admiral never asked. Reporting this as informational only, not as a defect: docs/dashboard.md:125 already states the general rule (&#34;a genuine ask that *opens or closes* a clause on one of these phrases is still refused&#34;), names it deliberate, and says the zero-false-positive figure is a property of the corpus rather than a guarantee - so the behaviour is honestly recorded, and the rejection message names the matched phrase so an agent can reword. The only gap is that the one worked example given is trailing &#34;in progress&#34;; a reader takes from the section above it that question-form asks were the thing the copula revert protected, and &#34;Do you want us to keep working on this?&#34; shows that protection only covers copula-led questions, not questions that end on a report phrase. If you ever revisit the docs, one extra example there would close that impression - explicitly not worth another fix round on its own.
- ℹ️ `tests/fm-dashboard.test.sh:321` - test_documented_guard_rates_still_hold pins the outcome counts but not two of the three corpus sizes, so the documented denominators can still drift silently - which is the exact thing the test exists to prevent. `counts != (15, 0, 0)` catches a shrunk REPORT_SHAPED (14 caught fails), but deleting entries from REWORDED_REPORTS or GENUINE_ASKS leaves their counts at 0 and the test green, while docs/dashboard.md:119 (&#34;Against **18** report-shaped reasons the list was never meant to reach&#34;) and :124 (&#34;Against **14** genuine asks&#34;) become false. Concretely: drop &#34;Monitoring disk usage&#34; and &#34;Still monitoring disk usage&#34; from REWORDED_REPORTS and the suite still passes, but the documented 18 is now 16 and the recorded bare-verb miss class has no executed evidence behind it. Two asserted lengths - `len(REWORDED_REPORTS) == 18` and `len(GENUINE_ASKS) == 14`, alongside the existing count check - close it. Mechanical, no behaviour change.
- ℹ️ `bin/fm-dashboard.sh:228` - Round 4&#39;s new refusal message tells the caller to use `fm-dashboard.sh status &lt;id&gt; $status --reason ...` for whatever status they passed, but only two statuses actually carry a reason on the card. store.set_status (store.py:305-306) writes `waiting_reason` only for `waiting` and `needs_attention_reason` only for `needs_attention`; for `working`, `paused`, `testing`, and `complete` both columns are set NULL and the text survives only as a status_history note, which `cmd_show` never renders (it prints notes tabs and the two status reasons, nothing from status_history). The CLI&#39;s own header at line 32 says as much: &#34;ignored for other statuses&#34;. So `add --title X --captain dj --prompt ... --status testing --reason &#34;ready for his eyes&#34;` now dies with advice to run `status &lt;id&gt; testing --reason ...`, which will exit 0 and put the text somewhere `show` does not display - the caller is redirected from one silent drop to another. The refusal itself is right and is what you asked for; only the suggested alternative overreaches. Naming `waiting` specifically (&#34;only `waiting` and `needs-attention` carry a reason on the card&#34;) would make the message true for all six other statuses. Flagged as ask-user because it is user-facing copy you have been deliberately wording.
- ℹ️ `bin/fm-dashboard.sh:21` - Recording, not re-litigating: the --intent text still describes the round-1 implementation, and round 4 added a user-visible behaviour it does not mention at all. The intent says phrases are &#34;matched as a case-insensitive substring anywhere in the reason text&#34; (code is edge-anchored via find_report_shaped_phrase), lists bare &#34;monitoring&#34; among 15 phrases (REPORT_SHAPED_PHRASES has &#34;monitoring the&#34;/&#34;monitoring it&#34;, 16 entries), and states rates of 15/15, 10/10, 0/10 (shipped docs and the pinned test say 15/15, 0/18, 0/14). New since that was adjudicated: `add` now hard-refuses `--reason` for any status but needs-attention (bin/fm-dashboard.sh:227-228, header at line 21), which is not in the intent&#39;s delivered-scope list in any form - before this branch it was an &#34;unknown argument&#34; error, so it is not a regression against the base, just undescribed. You already ruled on this class in round 2 (&#34;the real shipped docs were already correct, only the review&#39;s own --intent argument was stale&#34;), and every divergence here was directed by your own fix instructions, so no code or doc change is warranted - the note exists so that whatever text becomes the PR description is refreshed from the shipped docs rather than from the original intent paragraph.

🔧 Fix: narrow add --reason refusal to statuses storing a reason
1 info still open:
- ℹ️ `bin/fleet-dashboard/server/validation.py:45` - Recorded for the PR description, not for another fix round - you already ruled on this class twice (round 2: &#34;the real shipped docs were already correct, only the review&#39;s own --intent argument was stale&#34;; round 5: ignored). The --intent text still describes the round-1 implementation: it says phrases are &#34;matched as a case-insensitive substring anywhere in the reason text&#34; (the code is edge-anchored via find_report_shaped_phrase, validation.py:143), lists bare &#34;monitoring&#34; among 15 phrases (REPORT_SHAPED_PHRASES holds &#34;monitoring the&#34;/&#34;monitoring it&#34;, 16 entries), and states rates of 15/15, 10/10, 0/10 (shipped docs and the pinned test measure 15/15, 0/18, 0/14 - I re-ran all three corpora against the current code and they match the docs exactly). It also does not mention that `add` now hard-refuses --reason for any status but needs-attention (bin/fm-dashboard.sh:230-238), which is new since the intent was written but not a regression against the base, where --reason on `add` was an &#34;unknown argument&#34; error. Every divergence was directed by your own fix instructions, so no code or doc change is warranted - only that whatever text becomes the PR description is refreshed from docs/dashboard.md rather than from the original intent paragraph.

</details>

<details>

- ℹ️ `bin/fleet-dashboard/server/validation.py:26` - The shipped guard is narrower than the user intent&#39;s description of it. The intent states the report-shaped phrases are &#34;matched as a case-insensitive substring anywhere in the reason text&#34; and lists bare &#39;monitoring&#39;; bin/fleet-dashboard/server/validation.py matches only at clause edges (with a small hedge list) and carries &#39;monitoring the&#39;/&#39;monitoring it&#39;. Verified behavior: &#34;The bilge pump is being chased by the yard&#34;, &#34;Monitoring disk usage&#34;, &#34;The rebuild is in progress and the vendor has not called&#34; and &#34;We are investigating the checkout timeout&#34; are all ACCEPTED, where anywhere-substring matching would refuse them. The motivating example is still refused, the narrowing is documented in docs/dashboard.md as accepted misses with the false-positive reasoning, and the counts there (18 reworded misses, 14 genuine asks) supersede the 10/10 in the intent, pinned by test_documented_guard_rates_still_hold. No action needed for testing; noted so the divergence from the written intent is a deliberate, visible choice rather than a surprise.
- <code>`bash tests/fm-dashboard.test.sh` — full file, 17 checks pass, incl. `test_needs_attention_requires_a_real_ask`, `test_a_genuine_ask_mentioning_a_report_word_is_accepted`, `test_add_refuses_a_reason_for_a_status_that_cannot_carry_one`, `test_documented_guard_rates_still_hold`</code>
- <code>`bash tests/fm-dashboard-card-link.test.sh` — adjacent dashboard surface, passes</code>
- <code>`bash tests/fm-documentation-audiences.test.sh` — docs/dashboard.md changed, passes</code>
- <code>Manual end-to-end: started a real dashboard server on an isolated FM_HOME/port and drove `bin/fm-dashboard.sh status &lt;id&gt; needs-attention` with the motivating reason, with no reason, and with a genuine ask; captured the full transcript</code>
- <code>Manual server-side enforcement bypassing the CLI: `curl -i -X POST /api/tasks/&lt;id&gt;/status -d &#39;{&#34;status&#34;:&#34;needs_attention&#34;}&#39;` → HTTP 400, and `-d &#39;{...,&#34;reason&#34;:&#34;still chasing the supplier&#34;}&#39;` → HTTP 400</code>
- <code>Manual create path: `fm-dashboard.sh add --status needs-attention` with no reason, with &#39;looking into ...&#39;, and with a real ask</code>
- <code>Manual UI walkthrough via `chrome-devtools-axi` in an isolated browser session: opened a card, selected Needs Attention, screenshotted the required-reason form with submit disabled, submitted the report-shaped reason and screenshotted the inline refusal, then submitted a genuine ask and screenshotted the board</code>
- <code>Manual no-migration check: inserted a pre-guard `needs_attention` row carrying the old placeholder reason directly into the demo database, then `fm-dashboard.sh show` / `list --status needs-attention` confirm it is unchanged and still listed</code>
- <code>Manual untouched-statuses check: `fm-dashboard.sh status &lt;id&gt; working|paused|testing|complete|not-started` and `waiting --reason ...` all succeed with no ask required</code>
- <code>Guard probe on mid-clause phrasings via `validate_needs_attention_reason` to characterize the documented blind spot</code>

</details>

<details>

- ⚠️ `bin/fm-dashboard.sh:590` - bin/fm-dashboard.sh:590 renders --help via a hardcoded `sed -n &#39;2,64p&#39;` over the usage header. This change added ten lines to that header (the add/status --reason contract), pushing the header&#39;s last four blocks past line 64, so `--help` now silently stops at &#34;captain shorthands&#34; and no longer prints server URL resolution, the --connect-timeout/--max-time bounds and their env overrides, or the exit-code table (0/4/1). Confirmed by running `bin/fm-dashboard.sh --help` and diffing its tail against the base commit; no test covers help output. The .agents/skills/fleet-dashboard/SKILL.md tells agents to run `--help` for exact current syntax, so this is documentation lost from its rendered surface. Fix is to widen the range to &#39;2,74p&#39; (or make it stop at the first non-comment line), which is an executable line rather than a doc comment, so this phase left it alone.

🔧 Fix: derive --help header range instead of fixed line count
1 info still open:
- ℹ️ `bin/fm-status-board.sh:104` - The drift-prone pattern this round fixed in bin/fm-dashboard.sh exists in at least a dozen other scripts, each rendering --help via a hardcoded line range over its own header comment: bin/fm-status-board.sh:104 (&#39;2,96p&#39;), bin/fm-procevent-lavish.sh:62 (&#39;2,47p&#39;), bin/fm-afk-launch.sh:149 (&#39;2,34p&#39;), bin/fm-inactive-reconcile.sh:489 (&#39;2,40&#39;), bin/fm-backlog-receive.sh:31, bin/fm-herdr-lab.sh:302, bin/fm-afk-start.sh:46, bin/fm-remote-inherit.sh:23, bin/fm-kimi-turnend-hook.sh:23, and others. Every one silently truncates its own help the moment its header grows, exactly as fm-dashboard.sh did here. bin/fm-public-followup-emit.sh:76 already avoids it with a derived range (sed -n &#39;2,/^set -u$/p&#39;), so a working precedent exists. I did not sweep these: they are untouched by this change and fixing them is a repo-wide edit well outside this change&#39;s scope. Proposing it as a follow-up - convert the remaining scripts to the derived form and add a shared help-renders-full-header assertion.

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>


joliverMI:
Opened in error against upstream - this work belongs on our own fork and is being redirected there. Apologies for the noise.

PR #2742, opened 2026-08-21
feat(pi): show Codex quota in the footer

SamSherpaDev:
## Intent

Add auto-discovered quota information to Firstmate's Pi and pi-signed footer while preserving Pi's built-in footer and unrelated extension statuses. Use bounded, stdin-closed quota-axi --json work outside rendering; show every fresh quota window for the active model provider plus reset information, plan, and credits when supplied; never show stale or malformed data as fresh; handle unsupported providers, missing tools, failures, timeouts, and narrow widths explicitly and non-disruptively; refresh and clean up correctly across startup, session replacement or fork, model changes, reload, resize, and shutdown without leaking timers or subprocesses. Add behavior-level portable fake-output regressions and opt-in live Pi and pi-signed verification that exercise actual extension discovery and footer composition without credentials, prove supported non-Pi primary harness paths and all runtime backends remain unchanged, and document the feature and verification concisely. Deliver through the complete no-mistakes review, test, documentation, lint, push, PR, and green-CI path, preserve every pipeline fix commit, do not use --yes, do not merge the PR, and do not alter Herdr lifecycle behavior.

## What Changed

- Add an auto-discovered Pi/`pi-signed` quota row that preserves Pi’s built-in footer and unrelated statuses while showing every fresh Codex quota window, reset, plan, and credit value with width-aware fallbacks.
- Validate account, endpoint, schema, provenance, and freshness before publishing results, using bounded stdin-closed `quota-axi` refreshes with explicit degraded states and lifecycle-safe cleanup.
- Add portable fake-output and opt-in credential-free live regressions, pin Pi 0.84.2 for CI coverage, and document verification plus non-Pi/backend isolation.

## Risk Assessment

⚠️ Medium: The implementation is well-bounded and no material defect was substantiated, but its extensive schema validation, lifecycle state, private Pi runtime inspection, and cross-platform process management create meaningful compatibility complexity.

## Testing

After baseline repository inspection, the focused portable regression exercised parsing, provenance, freshness, failures, timeouts, lifecycle transitions, and subprocess cleanup successfully; actual offline Pi 0.84.2 PTY checks then demonstrated wide and narrow footer composition, preserved built-in and unrelated statuses, closed quota-axi stdin, and clean shutdown, while pi-signed and the repository’s tmux-based live guard were unavailable in this environment.

- Evidence: [Actual Pi wide quota footer](https://github.com/SamSherpaDev/firstmate/blob/e4edc70290e28a296668b68c51f58a5941d7cdfa/.no-mistakes/evidence/fm/firstmate-pi-quota-statusline/live-pi-quota-wide.png)
- Evidence: [Actual Pi narrow-width quota footer](https://github.com/SamSherpaDev/firstmate/blob/e4edc70290e28a296668b68c51f58a5941d7cdfa/.no-mistakes/evidence/fm/firstmate-pi-quota-statusline/live-pi-quota-narrow.png)
<details>
<summary>Evidence: Wide Pi PTY verification transcript</summary>

Source: [Wide Pi PTY verification transcript](https://github.com/SamSherpaDev/firstmate/blob/e4edc70290e28a296668b68c51f58a5941d7cdfa/.no-mistakes/evidence/fm/firstmate-pi-quota-statusline/live-pi-quota-verification.txt)

```text
Live Pi quota footer verification
Pi version: 0.84.2
Harness: actual pi executable in a 260x40 pseudo-terminal
Discovery: project .pi/extensions auto-discovery
Credentials: isolated synthetic account token; PI_OFFLINE=1
quota-axi calls: --json --full --provider codex|
quota-axi stdin observations: eof|
Result: PASS - Pi built-in cwd/git, model, and context footer content remained visible; unrelated extension status remained visible; all fresh Codex quota windows, resets, plan, and zero credits rendered in a separate row; Ctrl-D exited cleanly.

Final reconstructed terminal screen:
 pi v0.84.2
 escape interrupt · ctrl+c/ctrl+d clear/exit · / commands · ! bash · ctrl+o more
 Press ctrl+o to show full startup help and loaded resources.

 Pi can explain its own features and look up its docs. Ask it how to use or extend Pi.

[Extensions]
  aaa-unrelated-status.ts, fm-pi-quota-status.ts


────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Quota Codex (plan pro) | week 94% left reset 6d | GPT-5.3-Codex-Spark session 100% left reset 5h | GPT-5.3-Codex-Spark week 100% left reset 6d | credits 0
/private/var/folders/r4/lj5nyr393lv0s3j7ltth6dxc0000gn/T/no-mistakes-evidence/01M0FYKG66HZC1HY4870205B1Y/.live-pi-lab.13668/project (quota-live)
$0.000 (sub) 0.0%/128k (auto)                                                                                                                                                                                                           gpt-5.3-codex-spark • medium
LIVE_UNRELATED_STATUS
```
</details>
<details>
<summary>Evidence: Narrow Pi PTY verification transcript</summary>

Source: [Narrow Pi PTY verification transcript](https://github.com/SamSherpaDev/firstmate/blob/e4edc70290e28a296668b68c51f58a5941d7cdfa/.no-mistakes/evidence/fm/firstmate-pi-quota-statusline/live-pi-quota-verification-narrow.txt)

```text
Live Pi quota footer verification
Pi version: 0.84.2
Harness: actual pi executable in a 72x40 pseudo-terminal
Discovery: project .pi/extensions auto-discovery
Credentials: isolated synthetic account token; PI_OFFLINE=1
quota-axi calls: --json --full --provider codex|
quota-axi stdin observations: eof|
Result: PASS - Pi built-in cwd/git, model, and context footer content remained visible; unrelated extension status remained visible; the quota row degraded explicitly to a narrow 3-window summary without presenting a partial list; Ctrl-D exited cleanly.

Final reconstructed terminal screen:
 pi v0.84.2
 escape interrupt · ctrl+c/ctrl+d clear/exit · / commands · ! bash ·
 ctrl+o more
 Press ctrl+o to show full startup help and loaded resources.

 Pi can explain its own features and look up its docs. Ask it how to
 use or extend Pi.

[Extensions]
  aaa-unrelated-status.ts, fm-pi-quota-status.ts


────────────────────────────────────────────────────────────────────────

────────────────────────────────────────────────────────────────────────
Quota Codex: narrow - 3 windows; widen for details
/private/var/folders/r4/lj5nyr393lv0s3j7ltth6dxc0000gn/T/no-mistakes-...
$0.000 (sub) 0.0%/128k (auto)               gpt-5.3-codex-spark • medium
LIVE_UNRELATED_STATUS
```
</details>

## Pipeline

<!-- no-mistakes-pipeline-attestation:v1 {"head_sha":"717d5c17adc68c8d91c554a57bf46a4fb5eca99f","steps":[{"step":"intent","status":"completed"},{"step":"rebase","status":"completed"},{"step":"review","status":"completed"},{"step":"test","status":"completed"},{"step":"document","status":"completed"},{"step":"lint","status":"completed"},{"step":"push","status":"completed"},{"step":"pr","status":"running"},{"step":"ci","status":"pending"}]} -->

_... (21 earlier update rounds omitted to keep the PR body within GitHub's 65536-char limit; full history is in the run log.)_

<details>
<summary>🔧 **Review** - 3 issues found → auto-fixed (53) ✅</summary>

🔧 Fix: Harden quota composition, schema validation, and percentage formatting
5 errors still open:
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:323` - Required “show every fresh quota window” behavior is broken for typical quota-axi 0.1.29 `--json --full` output: schema 5 removed `pace.projectionBasis`, but projected known paces are required here to contain `cycle_average`, so used Codex windows become malformed. Make pace validation schema-version-aware and use exact producer output in the fake regression.
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:605` - Required “never show stale or malformed data as fresh” behavior remains violated: schema-5 `through_reset` is accepted without the producer-required `projectionConfidence`, while schema-3 `projected_exhaustion` is accepted without its required `projectionBasis`. Enforce each status/version contract rather than only branching for `through_reset`.
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:615` - Required “never show stale or malformed data as fresh” behavior remains reachable because a known `spendPriority` is checked only for range. Changing an otherwise-valid schema-5 report’s producer-computed selection value to any other in-range number still renders its windows as fresh. Validate the selection derivation against its bounding windows at the effective-availability boundary.
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:804` - Required “never show stale or malformed data as fresh” behavior is contradicted when a report contains duplicate rows for the active provider: `find()` selects the first valid row and ignores a conflicting or malformed duplicate. Reject duplicate provider identities before selecting a row.
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:950` - Required “show every fresh quota window” behavior fails after a modest backward clock adjustment. The monotonic timeout can fire while wall time is still before the reset, but this forces revalidation at the scheduled deadline and irreversibly removes the still-fresh window. Revalidate using current `now()` and reschedule when the deadline has not actually passed.

🔧 Fix: Harden quota schema validation and clock-safe expiry
5 errors still open:
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:965` - Required “show every fresh quota window” behavior can fail when an expiry callback occurs during a backward clock shift greater than the one-minute skew allowance. The callback publishes a transient stale view, and this branch never consults the still-recoverable cached view again because the captured view is not fresh; after clock recovery, fresh sibling windows remain hidden until the next refresh. Keep recoverable skew states revalidated or rescheduled.
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:797` - Active-provider quota can remain unavailable for five minutes after `/login`. Pi writes auth.json before its asynchronous ModelRuntime synchronization completes, so the watcher can resolve using the old API-key/non-OAuth state; once synchronization finishes, no model_select occurs for an already-selected model, and this comparison ignores effective auth mode/status. Include auth provenance in the live target revision and re-resolve after it changes.
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:362` - Required “never show ... malformed data as fresh” remains reachable for unknown window pace. For a fresh window with a reset but no cycle start/duration, quota-axi 0.1.29 must emit reason `missing_cycle`; changing it to `stale` still passes this shape-only branch and the percentages render fresh. Derive and validate the unknown reason from provider state and window fields at the window parse boundary.
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:527` - Required “never show ... malformed data as fresh” is contradicted because effective pace is not compared with its bounding windows. A one-window report whose raw pace is `behind` can change the aggregate to `ahead` and move the ID to aheadWindowIds; it passes these structural checks and renders fresh. Recompute the producer’s effective-pace summary from boundedBy windows and compare every aggregate field.
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:620` - Required “never show ... malformed data as fresh” remains reachable through runway sibling paths. An otherwise-valid report whose window projects exhaustion before reset can replace its aggregate with `through_reset` plus a valid confidence and pass here; projected_exhaustion timestamps and limiters likewise are not tied to the earliest window projection. Recompute the expected runway from the bounded windows at the effective-availability boundary.

🔧 Fix: Harden quota auth, clocks, and schema derivations
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:1219` - Required active-credential correlation can remain wrong for five minutes: after account A quota is cached, an A→B OAuth login whose `fs.watch` notification is omitted leaves the same model, provider object, OAuth mode, and auth-status source, so this revision check returns early and A remains rendered while Pi uses B. Periodically compare a bounded credential revision outside rendering, rather than relying solely on `fs.watch`.
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:1065` - The required “never show stale or malformed data as fresh” invariant is still violated because provider source and state are validated independently. A matching full report with `source: &#34;cache&#34;` (or `&#34;unavailable&#34;`) and `state: {status: &#34;fresh&#34;, stale: false}` reaches the fresh return and renders percentages, although quota-axi reserves cache for stale fallback and unavailable for failures. Enforce source/state consistency at this shared provider boundary.
- ⚠️ `.pi/extensions/lib/fm-pi-quota-status.ts:792` - Valid quota-axi 0.1.29 partial Kimi output is mislabeled malformed. When recognized windows coexist with unresolved limits, quota-axi deliberately emits `runway: unknown` with both bound and unresolved IDs; this full-projection branch recomputes from `boundedBy` alone and expects through-reset/projected runway. Make runway derivation aware of partial semantics and unresolved bounds.

🔧 Fix: Harden quota credential, source, and partial-runway validation
3 errors still open:
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:1006` - Required criterion “never show stale or malformed data as fresh” is violated: quota-axi schema 5 requires `quotaSemantics` on provider reports, but an omitted value is accepted here. A full Codex row with otherwise-valid windows/state/account therefore renders percentages as fresh. Require this field for schema 5; the live fake at tests/fm-pi-quota-status-live-e2e.test.sh:43 currently masks the defect by omitting it too.
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:922` - Required “never show ... malformed data as fresh” behavior remains reachable through shape-only `unknown` branches. A fully measurable Codex scope can replace its selection with `{status:&#34;unknown&#34;, unmeasurableWindowIds:[&#34;weekly&#34;]}` and pass here; lines 965 and 1047 similarly accept an unknown availability or semantics status despite known bounds. Recompute statuses and exact blocker sets from the enclosing windows at the shared semantics boundary.
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:1059` - Required “never show ... malformed data as fresh” behavior is contradicted because `authStatus` is checked only as an enum. An otherwise-valid fresh OAuth report with `authStatus:&#34;unusable&#34;` passes provider validation and renders percentages, although quota-axi reserves unusable authentication for non-fresh failure states. Enforce auth-status/provider-status/source consistency here.

🔧 Fix: Harden quota semantics, derivations, and auth validation
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:372` - Required criterion “never show stale or malformed data as fresh” remains violated. Runtime requests schema-5 `--json --full`, where quota-axi 0.1.29 always emits each window’s `pace` and each availability’s `pace`, `runway`, and `selection`, but these validators accept every field when absent. The live fixture at tests/fm-pi-quota-status-live-e2e.test.sh:49-63 omits them and still renders fresh. Require these fields when validating the full projection before checking their relations.
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:1169` - Required criterion “never show stale or malformed data as fresh” is still contradicted because only the presence of any effective-availability entry is checked. The live fixture has `weekly` plus two `model:spark` windows but only `all_models`; quota-axi 0.1.29 must also emit `model:spark` bounded by all three windows. This producer-impossible report passes as fresh. Derive and compare the exact ordered scopes and `boundedBy` sets, rejecting missing or extra entries.
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:329` - Required cleanup “without leaking timers or subprocesses” is not guaranteed on Windows. There `processGroupId` is always null and timeout/shutdown kills only the quota-axi leader; quota-axi 0.1.29’s Codex path spawns a `codex ... app-server` child, which can survive when its parent is forcibly killed before its own `finally` cleanup. Use a supported Windows process-tree boundary such as a job object or bounded tree termination.
- ⚠️ `.pi/extensions/fm-pi-quota-status.ts:1017` - A backward clock skew over Node’s maximum timeout delay (~24.8 days) makes `setTimeout` clamp this delay to 1 ms. The callback then sees the same recoverable skew and repeatedly replaces the widget and reschedules at 1 ms. Cap long delays and revalidate/reschedule in bounded chunks.

🔧 Fix: Harden quota schema, process trees, and expiry timers
4 errors still open:
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:723` - Required “show every fresh quota window” behavior rejects valid schema-5 output. docs/verification/dispatch-auth.md:111 records that known runway may omit projectionConfidence, but this always synthesizes one and exact matching then labels a valid `through_reset` report malformed. Preserve producer uncertainty instead of fabricating confidence.
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:339` - Required cleanup “without leaking ... subprocesses” remains reachable on Windows. If quota-axi forks a pipe-holding descendant and its leader exits, `exitCode` is set before timeout/close, so this returns without invoking `taskkill` and the descendant survives. Establish a process-tree boundary such as a Windows job object at spawn time.
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:1031` - Required “never show stale ... data as fresh” can fail after a backward clock shift larger than Node&#39;s maximum timer delay. The callback schedules revalidation roughly 24.8 days away; if the clock recovers and a redraw restores the cached fresh window, its original near-term reset is not rescheduled, so the percentage can remain displayed past reset until the five-minute refresh. Republish/reschedule when a recoverable view becomes fresh again.
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:1271` - Required “never show ... malformed data as fresh” remains violated because state reasons are only enum-checked. An otherwise-valid matching report with `status:&#34;fresh&#34;`, `stale:false`, and `reason:&#34;credentials_expired&#34;` passes and renders percentages, despite that reason belonging to non-fresh expired-credential states. Enforce reason/status/authStatus consistency at this provider-state boundary.

🔧 Fix: Harden quota schema, expiry, and Windows process cleanup
4 errors still open:
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:757` - Required quota “for the active model provider” remains unsafe: composition revisions only invalidate the cache. A trusted extension can override canonical `openai-codex` with an inherited OAuth/official URL but custom `streamSimple`; re-resolution still classifies it supported and republishes Codex quota even though model traffic uses the custom stream. Verify canonical provider implementation provenance at target resolution, or fail closed when it cannot be established.
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:56` - Required cleanup “without leaking ... subprocesses” remains reachable on Windows. This ordinary supervisor does not create a Job object; after the quota leader exits, line 54 sends its result within 50 ms and `taskkill /T` targets the supervisor after the descendant has been orphaned from its live tree. The test masks this by killing every PID from a global log. Establish a Windows Job/process-tree boundary when spawning the quota leader.
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:754` - Required “never show ... malformed data as fresh” is contradicted. quota-axi 0.1.29 always emits `projectionConfidence` for schema-5 `through_reset` results and does not demote it from JSON, but this branch accepts its deletion from an otherwise-valid full report and renders the row fresh. Require the computed confidence exactly; the cited verification note concerns runway statuses that legitimately have no confidence, not `through_reset`.
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:1342` - Required “never show ... malformed data as fresh” remains reachable because full-projection attempts are only shape-checked. A matching fresh Codex row with `source:&#34;oauth&#34;` but an OAuth attempt marked `failed`—or `sourcesTried` inconsistent with attempts—still renders percentages. Validate that a present full attempt list has a success matching the fresh provider source and that `sourcesTried` matches its ordered sources.

🔧 Fix: Harden quota provenance, validation, and Windows cleanup
3 errors still open:
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:982` - Required quota “for the active model provider” remains unsafe: Pi composes `models.json` overlays independently, while these getters expose only extension registrations. An `openai-codex` model can retain its provider/id, official URL, and OAuth while changing its API implementation; both checks remain undefined and unrelated Codex quota is published. Gate on effective composition provenance at Pi’s model-runtime boundary, including `models.json`.
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:1314` - Required “never show stale or malformed data as fresh” remains violated: a matching row with `status:&#34;fresh&#34;`, `stale:false`, and failure-only `error`, `retryAfter`, or `remedyCommand` fields passes validation and renders percentages. quota-axi’s fresh producer emits none of these fields; enforce their status-dependent absence here.
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:1426` - Required “never show stale or malformed data as fresh” remains reachable because selection silently ignores malformed sibling entries. Runtime requests provider-scoped output, yet `[validCodexRow, null]` parses and this filter selects the valid row as fresh. Validate every provider element and require the exact requested provider set at the provider-scoped runtime boundary.

🔧 Fix: Harden quota composition and scoped report validation
3 errors still open:
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:1431` - Required “show every fresh quota window” behavior fails if quota-axi emits at T and the wall clock moves backward over 60 seconds before selection: this stores only a stale view, so recovery within the six-minute freshness period cannot restore the report until the next cadence. A refresh failure during the same recoverable skew also clears the cached fresh view at .pi/extensions/fm-pi-quota-status.ts:1598. Preserve a validated recoverable report across publication and failure boundaries.
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:731` - Required explicit timeout handling is bypassed for quota-axi’s structured provider failures. quota-axi 0.1.29 prints provider-scoped JSON and exits 1 when the sole provider fails; an internal Kimi `request_timeout` or Codex timeout that completes before the outer timer reaches this branch, which discards stdout and renders generic `quota-axi failed`. Retain bounded stdout with the exit status and validate/map provider timeout state explicitly.
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:1318` - Required “never show stale or malformed data as fresh” behavior remains violated. Runtime requests `--json --full`, where quota-axi 0.1.29 fresh success rows always include `state.refreshedAt`, but this validator permits its omission; the row then uses top-level `generatedAt` and renders fresh. Require `refreshedAt` for fresh full-projection rows.

🔧 Fix: Harden quota recovery, timeout mapping, and freshness validation
2 errors still open:
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:1594` - Required “show every fresh quota window” and “handle ... failures, timeouts ... explicitly” behavior fails on quota-axi’s normal stale-cache path. After a fresh report at t=0, a t=5m Codex timeout can return a stale cached provider with exit 0; this success branch discards the extension’s still-fresh cached view, clears the failure, and renders only “stale.” Classify provider state before replacing the cache, retain independently fresh data until expiry, and surface the timeout alongside it.
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:1338` - Required “never show ... malformed data as fresh” behavior remains reachable. `untrustedWindowIds` is only shape-checked and is not related to displayed windows, so an otherwise-valid matching Codex report with `untrustedWindowIds:[&#34;weekly&#34;]` passes normal Codex semantics and renders that weekly percentage as fresh. Reject or exclude explicitly untrusted displayed windows at the shared provider-validation boundary.

🔧 Fix: Preserve cached quota and reject untrusted windows
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:1368` - Required behavior to “show every fresh quota window” and refresh “non-disruptively” fails on any auth.json write: logging into an unrelated provider unconditionally clears the active provider’s still-valid cache and starts another quota process without first comparing the active credential revision. Route watch notifications through the bounded active-entry comparison and invalidate only on an actual revision change.
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:1628` - The stale-fallback fix is not durable across sibling refresh outcomes. After a validated t=0 report, an unchanged-target t=5 refresh that exits 0 with malformed output replaces the still-fresh cache here; an OAuth-resolution timeout similarly clears it at line 1519. Preserve target-bound validated data until its own expiry while exposing the new malformed/timeout outcome at a shared refresh-publication boundary.
- ⚠️ `.pi/extensions/fm-pi-quota-status.ts:1185` - Every Kimi target passes resolution and runs quota-axi each cadence, but this always supplies expectedAccountId:null, and selection returns unverified before source verification at lib/fm-pi-quota-status.ts:1551. Thus Kimi network/process work can never publish quota. Classify it as account-correlation unavailable before spawning until report-side identity can actually be verified.

🔧 Fix: Harden quota cache publication and unsupported provider handling
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:1662` - Required “show every fresh quota window ... non-disruptively” behavior still fails at the shared publication boundary: after caching account A at t=0, a valid t=5 refresh from quota-axi’s independent account B selects `unverified`, and this branch replaces A’s still-fresh, target-bound cache. Preserve validated cached data until its own expiry and expose `unverified` as the latest refresh issue.
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:925` - Required non-disruptive bounded work is not guaranteed. Pi 0.84.2 permits `models.json` provider headers backed by `!command`; its composed `oauth.toAuth` resolves those commands synchronously, so this Promise timeout cannot fire while the TUI event loop is blocked, potentially twice per successful refresh. Fail closed on such composition or isolate effective-auth resolution behind a truly asynchronous bound.
- ⚠️ `.pi/extensions/fm-pi-quota-status.ts:592` - A Codex token with a missing or conflicting canonical account claim produces a non-null verification object whose account ID is null. The target is therefore treated as supported and quota-axi runs every cadence even though selection must always return `unverified`. Classify missing identity as account-correlation unavailable before spawning.
- ⚠️ `.pi/extensions/fm-pi-quota-status.ts:608` - On Windows, timeout or shutdown performs `taskkill` synchronously on Pi’s event loop for up to five seconds before the process result or footer update can complete. This makes the required timeout handling disruptive when tree termination stalls; use bounded asynchronous tree termination and await it at the process completion boundary.

🔧 Fix: Harden quota refresh provenance and asynchronous cleanup
2 errors still open:
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:1445` - Required “refresh ... correctly” behavior remains broken after repairing an uncorrelatable OAuth login. If the current Codex token lacks a valid account claim, the target becomes `account-correlation`; replacing auth.json with a valid token triggers this watcher, but that reason is omitted here and the periodic revision check only handles supported targets, so the footer remains unavailable until the five-minute cadence. Route every credential-dependent target through the bounded active-entry revision check and re-resolve on an actual change.
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:1292` - Required “show every fresh quota window ... non-disruptively” behavior still has a sibling failure path. After a validated t=0 report, a transient malformed/partial models.json read at t=5 carries the same credential, model, and composition revisions, but only `auth-timeout` preserves the cache here; line 1667 then clears still-fresh quota until another cadence. Preserve the target-bound cache for non-authoritative resolution failures whose identity revisions remain unchanged, while showing the new issue explicitly.

🔧 Fix: Preserve quota cache and refresh repaired credentials
3 errors still open:
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:1479` - The advertised active-account correlation remains briefly false after a credential change: this starts an asynchronous revision read while leaving account A’s cached widget published, so a slow read can display A’s quota for up to timeoutMs after Pi switches to account B. Quarantine the cached view at the credential-change boundary and restore it only after proving the revision unchanged.
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:1524` - Required refresh behavior is not recoverable: one transient timeout, overflow, or malformed/partial auth.json read calls failCredentialMonitoring(), closes the healthy watcher, clears quota, and permanently disables later checks and refreshes until session replacement. Keep a temporarily unverifiable state and retry while reserving permanent shutdown for watcher failure.
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:1387` - Required “never show stale or malformed data as fresh” behavior is still violated for accepted schema-3 full reports. quota-axi’s schema-3 --json --full producer attaches quotaSemantics, but only schema 5 requires it here, so stripping that derivation from an otherwise matching report still renders raw percentages as fresh. Require the full schema-3 projection or reject schema 3 explicitly.

🔧 Fix: Harden credential revalidation and schema-3 quota semantics
2 errors still open:
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:835` - Required “never show stale or malformed data as fresh” remains violated for the explicitly accepted schema-3 full projection: deleting a producer-emitted `effectiveAvailability[].runway` succeeds here and the otherwise-valid report still renders fresh; schema-3 effective pace is similarly optional. Make full-projection requirements explicit in these validators or reject schema 3.
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:1987` - The active-account guarantee remains briefly false when `fs.watch` misses an A→B credential update. This periodic fallback starts an asynchronous revision read without quarantining account A’s cached quota, so A remains visible while Pi uses B until the poll and bounded read complete—up to roughly 1 second plus the 20-second default timeout. Use a supported credential-change boundary or explicitly authorize and document bounded eventual containment rather than claiming an exact active-credential match.

🔧 Fix: Require schema-3 derivations and clarify credential polling
2 errors still open:
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:1208` - Required “show every fresh quota window for the active model provider” behavior fails with valid Pi configuration: Pi 0.84.2 deliberately strips comments before parsing models.json, but readBoundedJson uses strict JSON.parse, so a comment-bearing models.json resolves here as auth-unavailable and Codex quota is never queried. Parse models.json with Pi’s JSONC semantics while keeping auth.json strict.
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:565` - Required “never show stale or malformed data as fresh” remains violated. Codex can produce only oauth or cli-rpc sources, but this generic fallback accepts a matching fresh Codex row with source:&#34;api&#34;, an api success attempt, and matching account ID, then renders its percentages as fresh. Enforce provider-specific source/attempt combinations at this provenance boundary.

🔧 Fix: Support JSONC models and validate Codex quota provenance
1 error still open:
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:484` - Required criteria “show every fresh quota window ... and credits when supplied” are contradicted: quota-axi 0.1.29 permits and forwards any finite credit balance, including a negative/debt balance, but this check marks the entire matching fresh report malformed. Accept finite signed balances and format the sign instead of hiding all windows, plan, and credits.

🔧 Fix: Support signed quota credit balances
2 errors still open:
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:114` - Required “never show stale or malformed data as fresh” remains reachable: Date.parse accepts timezone-less and normalized-invalid timestamps. On a UTC-07 host, malformed generatedAt `2026-08-20T12:00:00` is interpreted as 19:00Z and can make data intended as 12:00Z pass the six-minute freshness check at 19:03Z. Require canonical timezone-bearing ISO instants at this shared timestamp boundary.
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:565` - Required “never show ... malformed data as fresh” is contradicted because Codex validation checks only successful attempts. A fresh matching row with attempts `[bogus/skipped, oauth/success]` and matching sourcesTried passes and renders, although quota-axi 0.1.29 permits only its ordered oauth/cli-rpc paths. Validate the complete provider-specific attempt source, order, status, and diagnostic grammar here.

🔧 Fix: Validate quota timestamps and Codex attempt provenance
4 errors still open:
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:391` - Required “never show stale or malformed data as fresh” is still violated for schema-3 full output: missing window `pace` is accepted here, although quota-axi’s full producer always attaches it. A report containing unknown-cycle windows can delete their pace while retaining matching unknown aggregate derivations and still render fresh. Require pace whenever full audit fields are required.
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:1455` - Required “never show stale or malformed data as fresh” remains reachable: this rejects `untrustedWindowIds` only when they match displayed windows. An otherwise-valid fresh Codex row with `untrustedWindowIds:[&#34;ghost&#34;]` passes because Codex semantics ignore the field, then renders percentages as fresh despite producer-impossible trust diagnostics. Enforce provider-specific trust-diagnostic ownership at this boundary.
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:1652` - Required reset information can remain inaccurate while Pi is idle. The formatter expires its cache at each countdown transition, but this lifecycle timer requests a redraw only at quota freshness/window expiry; a displayed “reset 3m” can therefore remain onscreen as “3m” for minutes without unrelated TUI activity. Schedule redraws at the next reset-format transition or display non-relative reset information.
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:1237` - Required bounded, non-disruptive handling is not guaranteed for malformed output. Before checking the supplied availability count, thousands of account and uniquely scoped model windows within the 1 MiB process limit cause this loop to materialize an account-sized array per model, amplifying a bounded report into tens of millions of references on Pi’s event loop. Reject implausible cardinalities early or validate topology without constructing the expanded expectation.

🔧 Fix: Harden quota validation, topology bounds, and countdown redraws
2 errors still open:
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:1447` - Required criterion “show every fresh quota window” is contradicted by the undocumented 128-window cap. quota-axi 0.1.29 imposes no such limit and can emit two Codex windows per `additional_rate_limits` entry, so 64 named limits plus base windows produce a valid sub-1MiB report that is labeled malformed. Keep validation bounded without rejecting valid cardinalities, such as one-pass topology comparison.
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:1483` - Required explicit unsupported-provider handling is lost when the unconditional credential watcher fails: an already-unsupported provider or no-model target is replaced with `credential monitoring unavailable`, misleadingly implying monitoring is the blocker. Preserve targets that have no credential binding; apply monitoring failure only to credential-dependent targets.

🔧 Fix: Support large quota reports and preserve unsupported states
3 errors still open:
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:1231` - Required criterion “never show stale or malformed data as fresh” is contradicted: any `kind:&#34;model&#34;` ID and arbitrary `weekly_N`/`window:*` IDs are treated as valid Codex topology. The new fixture at tests/fm-pi-quota-status.test.sh:980 proves producer-impossible `weekly_1...64` and `m1...129` rows publish as fresh. Validate quota-axi’s complete Codex window identity/kind/duplicate-suffix grammar before deriving semantics.
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:1448` - Required “show every fresh quota window” behavior fails after a temporary forward clock jump. A t=0 report with a t=3m reset, redrawn after the clock jumps from t=1m to t=4m, is destructively cached without that window; restoring the clock to t=2m cannot recover the still-fresh window before the next refresh. Keep validated publication immutable and record authoritative monotonic/timer expirations separately from transient render-time wall-clock filtering.
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:444` - Required “show every fresh quota window” behavior still has an undocumented cardinality-like limit: `exactText` rejects machine IDs over 200 characters, while quota-axi 0.1.29 forwards arbitrary nonempty Codex `metered_feature` IDs. A valid sub-1MiB report with a 201-character named-limit ID is therefore labeled malformed. Rely on the process byte bound for machine-ID size or enforce a producer-supported limit rather than this local cap.

🔧 Fix: Harden quota identity validation and clock-safe expiry
5 errors still open:
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:1728` - Required “never show stale ... data as fresh” remains reachable when a countdown timer is delayed. For a t=0 report resetting at t=150s, the first timer targets the t=30s display transition; if the event loop blocks until t=180s, this commits only through t=30s. Moving the wall clock back to t=120s then resurrects the window despite its reset having elapsed monotonically. Separate non-authoritative countdown redraws from authoritative report/window expiry timers.
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:1800` - Required “show every fresh quota window” behavior still fails when the clock jumps forward between report generation and selection. A t=0 report with a t=3m reset selected at temporary wall time t=4m is cached only after this destructive filtering; restoring the clock to t=2m cannot recover the fresh window. Preserve the validated publication and apply wall-clock filtering only to transient render views until authoritative expiry.
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:515` - Valid Codex producer output with ten duplicate derived IDs is rejected: quota-axi emits suffixes `_2` through `_10`, but this regex does not strip `_10` because it starts with `1`, so `validCodexWindowIdentities` labels the report malformed. Recognize every duplicate suffix of 2 or greater and cover the first double-digit case.
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:448` - Required “never show ... malformed data as fresh” is contradicted for full Codex results. Codex success requires at least one normalized window and emits both `percentUsed` and `percentRemaining`, but both percentages are optional here and an empty window list is also accepted at line 1583. The portable test explicitly calls empty Codex output fresh, while the live fake omits `percentUsed`. Enforce the provider-specific fresh-success shape and make fixtures match actual `--json --full` output.
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:848` - Required bounded, non-disruptive handling is not guaranteed for large permitted reports. Each audit ID performs a linear `boundedBy.includes` lookup, with similar scans in runway, selection, and limiter validation; thousands of same-model windows within the 1 MiB output limit therefore trigger quadratic work on Pi’s event loop. Build membership sets once and keep validation proportional to input size.

🔧 Fix: Harden quota expiry, validation, and bounded report auditing
2 errors still open:
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:1716` - Required “show every fresh quota window” behavior still fails under temporary forward wall-clock changes. A t=0 report selected while wall time is temporarily t=7m returns stale here before its validated view can be retained; similarly, the expiry callback at .pi/extensions/fm-pi-quota-status.ts:1745 commits a transient t=10m wall reading and irreversibly hides a sibling resetting at t=5m after the clock returns. Preserve validated publications and base irreversible expiry on elapsed monotonic time/deadlines, using wall time only for transient filtering.
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:1468` - Every fresh widget redraw calls currentView → cachedQuotaView and scans every accepted window to recompute the next expiry; after the first deadline, revalidateFreshQuotaView also filters the complete publication. Sub-1MiB reports with hundreds or thousands of windows therefore reintroduce O(report-size) work on token-by-token Pi redraws despite formatter caching. Cache the next deadline and filtered view at report/expiry publication so rendering performs an O(1) freshness check.

🔧 Fix: Make quota expiry monotonic and render-time constant
2 errors still open:
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:1985` - Required “never show stale ... data as fresh” remains reachable: quota-axi timestamps generatedAt before provider I/O, but monotonic elapsed starts at zero only after publication while timelineOrigin remains generatedAt. Process latency is therefore added to every expiry; an idle footer can retain expired percentages, and a wall-clock rollback can render them fresh again. Initialize authoritative elapsed at the shared publication boundary using the report’s already-observed age/process clock mapping.
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:463` - Required “show every fresh quota window” behavior rejects valid Codex output. quota-axi forwards arbitrary nonempty metered_feature/limit_name values into machine IDs, but exactText normalizes whitespace before accepting them; an identifier such as `preview  beta` makes the entire fresh report malformed. Validate opaque IDs according to the producer contract and reserve cleanText for displayed labels.

🔧 Fix: Fix quota publication age and opaque window IDs
2 errors still open:
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:1983` - Required “show every fresh quota window” behavior is contradicted: elapsedAtPublicationMs treats the entire child lifetime as age since report.generatedAt, although quota-axi records generatedAt after child startup. A 15-second startup delay can emit immediately a fresh window resetting 10 seconds later, but this adds 15 seconds and drops it at publication. Map generatedAt onto monotonic time using process/publication wall-clock samples rather than charging pre-generation startup time.
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:1223` - Required bounded, non-disruptive work remains bypassable: Pi uses a startup/reload models.json snapshot, while this gate rereads mutable disk. If Pi loaded a command-backed header and the file is then removed, this check sees no override but line 1253 invokes the still-composed OAuth resolver; its synchronous command can block the event loop so the Promise timeout cannot fire. Gate against effective runtime composition or isolate auth resolution behind a truly asynchronous bound.

🔧 Fix: Harden quota publication timing and provider provenance
2 errors still open:
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:1936` - Required criterion “never show stale ... data as fresh” remains reachable. If wall time steps backward between process start and report generation, `generationProgress` can clamp to 1 and record zero publication age for an already-old report. The authoritative expiry timer then fires late, so an idle footer can retain expired percentages. Detect discontinuous wall/monotonic samples and fail closed instead of treating ambiguous age as zero.
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:1086` - Required criterion “show every fresh quota window for the active model provider” fails for harmless Pi `models.json` metadata overrides. Pi&#39;s composed provider creates new model objects on each lookup whenever provider configuration exists, so even a name-only `modelOverrides` entry makes this identity comparison reject an otherwise official Codex OAuth endpoint as “provider override.” Compare request-affecting effective provenance, or explicitly authorize and document this containment.

🔧 Fix: Handle clock discontinuities and cosmetic model overrides
4 errors still open:
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:654` - Required criterion “never show stale ... data as fresh” remains reachable. If wall time jumps +10s before `generatedAt`, then -5s afterward during a 19s process, this computes a 10s publication age although the report is 15s old; both wall and authoritative expiry then lag by 5s and render expired percentages as fresh. Use a provable upper age bound or require a producer-side monotonic generation marker.
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:1728` - Required criterion “show every fresh quota window” still fails after a temporary forward clock jump exceeding twice the freshness interval. A newly generated report selected during a +13m jump returns here without a recoverable publication; restoring the clock while the report is actually fresh cannot restore its windows until another refresh. Preserve validated publications independently of transient wall-clock classification.
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:605` - Required quota “for the active model provider” is suppressed by valid non-routing Pi metadata overrides. For example, `modelOverrides[id].contextWindow` leaves the canonical provider, API, endpoint, OAuth, and stream unchanged, but this whitelist labels it a provider override and never queries quota. Compare effective routing/auth provenance rather than allowing only `name` and `cost`.
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:1289` - Required non-disruptive timeout handling can still be bypassed. `resolveTarget` preflights before awaiting the credential read, then fetches and invokes the current composed OAuth resolver without rechecking composition; a supported `modelRegistry.refresh()` that installs a command-backed header during that await makes `toAuth` block synchronously, so its Promise timeout cannot fire. Revalidate composition immediately after the read and before invoking the resolver.

🔧 Fix: Harden quota clocks, provenance, and auth resolution
3 errors still open:
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:1549` - Required “show every fresh quota window” behavior fails after a temporary forward wall-clock step: once wall time crosses the earliest window reset, this collapses the entire cached view to stale, hiding later-reset siblings, plan, and credits even though authoritative monotonic time has not expired them. Revalidate/filter the publication at the shared cache boundary instead of treating one crossed window deadline as provider-wide staleness.
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:668` - Required “show every fresh quota window” behavior remains contradicted by the 2 ms sampling threshold. A brief pause or clock adjustment between adjacent wall/monotonic samples makes this charge the entire child lifetime as report age; a slow-starting child that generates a report near completion can therefore have a still-fresh near-reset window removed immediately. Use bracketed clock samples or another publication mapping that distinguishes sampling delay from report age.
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:883` - Required “never show stale or malformed data as fresh” behavior is violated because Buffer.toString silently replaces invalid UTF-8. Invalid bytes inside a JSON string can become U+FFFD, pass JSON.parse and field validation, and publish percentages as fresh. Decode stdout with a fatal UTF-8 decoder and classify decoding failure as malformed.

🔧 Fix: Fix quota clock filtering, publication timing, and UTF-8 validation
3 errors still open:
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:677` - Required “never show stale ... data as fresh” remains reachable. If wall time jumps forward between process start and quota generation, then returns before publication, start/end wall and monotonic elapsed can match while this returns only `publishedWall-generatedAt`. A reset-after window can therefore expire monotonically before the authoritative timer reaches it. Use the full process elapsed as the safe upper bound or add a producer-side monotonic generation marker at this publication boundary.
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:1870` - Required “never show stale ... data as fresh” fails during an idle forward clock adjustment. For a fresh Codex report whose windows have no `resetsAt`, this schedules no status redraw; the one-second revision interval also does not render. A +10-minute clock step after one minute can therefore leave expired percentages physically displayed until the five-minute refresh. Add a bounded lifecycle wall-freshness check that requests rendering when wall classification changes.
- 🚨 `.pi/extensions/lib/fm-pi-quota-status.ts:1663` - Required “never show ... malformed data as fresh” remains violated for runtime’s provider-scoped full Codex report. Generic credit parsing accepts producer-impossible shapes such as `{remaining:5,unit:&#34;usd&#34;}` or `{unit:&#34;credits&#34;}`, allowing the matching report to render fresh with false or unknown credits. Validate Codex’s provider-specific credit shape: `unit:&#34;credits&#34;` and at least `remaining` or `unlimited`.

🔧 Fix: Harden quota timing, redraws, and credit validation
2 errors still open:
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:641` - Required quota “for the active model provider” and non-disruptive behavior remain violated. In real Pi, `ctx.model` and `modelRegistry.find()` normally return the same runtime model object, so this fast path bypasses `modelsConfigPreservesQuotaProvenance()` for every `models.json` overlay. Request-affecting headers/compat can therefore pass; an Authorization override may route model traffic through account B while quota is verified for OAuth account A, and a `!command` header can synchronously block `toAuth()` beyond the Promise timeout. Always inspect effective runtime configuration; the regression currently masks this by returning a cloned model.
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:668` - Required “show every fresh quota window” behavior is contradicted because the entire child lifetime is charged as age since `generatedAt`, although quota-axi records `generatedAt` only after process startup. If startup takes 15 seconds and quota-axi then immediately emits a valid window resetting 10 seconds later, publication occurs before reset but this calculation filters it as already expired. Map the producer timestamp to monotonic time without charging pre-generation startup, add a producer-side marker, or explicitly authorize conservative containment; the test at tests/fm-pi-quota-status.test.sh:3669 currently codifies the incorrect omission.

🔧 Fix: Fix quota provenance checks and publication age mapping
1 error still open:
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:693` - Required criterion “never show stale ... data as fresh” remains reachable. If wall/monotonic time start at 0, wall time jumps +10s when quota-axi stamps generatedAt at monotonic 5s, then returns before publication at wall/monotonic 20s, both validation checks pass but mapped age is about 7s instead of the actual 15s. A resetless fresh row can remain displayed roughly 8s past its six-minute lifetime. Endpoint samples cannot detect this clock ABA; require a producer-side monotonic generation marker or explicitly authorize conservative full-process-age containment at publication.

🔧 Fix: Conservatively age quota publications across clock ABA
2 errors still open:
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:619` - Required criterion “show every fresh quota window for the active model provider” remains contradicted: a valid models.json can upsert the active built-in Codex model with unchanged official API/base URL/OAuth and only metadata changes, but this unconditional rejection classifies it as a provider override and never invokes quota-axi. Compare effective routing/auth provenance for the matching models[] entry, rejecting only request-affecting changes.
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:942` - Required criterion “never show stale ... data as fresh” remains reachable on Linux because performance.now() uses a monotonic clock that excludes system suspend. After caching quota, suspending beyond its freshness period, waking, and moving wall time back within the report deadline, the reversible wall filter can republish the old cache because authoritative elapsed time never included sleep. Use a sleep-inclusive continuous/boottime clock at the cache-expiry boundary.

🔧 Fix: Support metadata model overlays and suspend-safe quota expiry
4 errors still open:
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:1412` - Required quota “for the active model provider” can be misattributed: Pi’s supported `before_provider_headers` hook can replace the final Authorization header after this registry-based account check. Traffic may use account B indefinitely while the footer publishes account A’s quota. Correlate at a boundary that observes or constrains final outbound authentication, or explicitly authorize this containment.
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:718` - Active-account verification accepts any token with a decodable second segment, while Pi’s Codex request path requires exactly three JWT segments. A two- or four-segment token can therefore correlate and display fresh independent Codex quota even though Pi rejects the credential. Match Pi’s token grammar before accepting the account claim.
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:2164` - Required “never show stale ... data as fresh” remains reachable when suspension occurs during quota-axi work. Publication age uses performance.now(), which excludes Linux suspend, while the sleep-inclusive clock is sampled only after publication; a report generated before a &gt;6-minute suspend can later reappear after wall-clock rollback because the missed sleep is absent from both authoritative age components. Sample the continuous clock at process start and include its full-process delta in publication age.
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:909` - Required failure handling is not fully non-disruptive: stdout and stderr pipes have data listeners but no `error` listeners. A pipe read error emits an unhandled EventEmitter error and can terminate Pi instead of producing an explicit failure state. Handle both stream errors through the shared process completion boundary.

🔧 Fix: Harden quota auth, process I/O, and suspend expiry
3 errors still open:
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:1641` - Required quota “for the active model provider” is contradicted. In pinned Pi 0.84.2, `before_provider_headers` transforms additional headers before the Codex API unconditionally sets its canonical Authorization and account headers. A later hook writing account B therefore quarantines quota here even though the actual request uses account A. Correlate at a genuinely final authentication boundary, or retain the already-verified active credential.
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:1619` - Required behavior to “refresh ... correctly across ... model changes” fails after one outbound-auth mismatch. `outboundAuthQuarantined` is set permanently; subsequent matching requests, credential changes, and same-provider `model_select` events never clear it, so valid quota remains unavailable until session replacement. Re-evaluate or clear quarantine at the next authoritative target/auth transition.
- 🚨 `.pi/extensions/fm-pi-quota-status.ts:729` - Required quota “for the active model provider” remains unsafe because `Buffer.from(..., &#34;base64url&#34;)` permissively ignores invalid payload characters. A three-segment token with an otherwise-valid payload plus an invalid character can correlate here, while Pi’s Codex request decoder rejects it, allowing unrelated local quota to render for an unusable credential. Validate and decode with the same strict grammar Pi uses.

🔧 Fix: Fix quota auth provenance and strict JWT decoding

</details>

<details>

✅ No issues found.
- `tests/fm-pi-quota-status.test.sh`
- <code>`pi --version` and harness/tool availability probes (`pi` 0.84.2 available; `pi-signed` and `tmux` absent)</code>
- <code>`bash /var/folders/r4/lj5nyr393lv0s3j7ltth6dxc0000gn/T/no-mistakes-evidence/01M0FYKG66HZC1HY4870205B1Y/run-live-pi-pty.sh` — actual Pi auto-discovery in a 260x40 PTY with offline synthetic credentials and fake quota output</code>
- <code>`LIVE_PI_WIDTH=72 bash /var/folders/r4/lj5nyr393lv0s3j7ltth6dxc0000gn/T/no-mistakes-evidence/01M0FYKG66HZC1HY4870205B1Y/run-live-pi-pty.sh` — actual narrow-width Pi composition</code>
- `Rendered the captured 260x40 and 72x40 terminal surfaces to PNG with macOS AppKit`

</details>

<details>

✅ No issues found.

</details>

<details>

🔧 Fix: Fix ShellCheck warnings in live quota regression
1 warning still open:

</details>

<details>

✅ No issues found.

</details>


kunchenguid:
Speaking as Kun's firstmate: This is default-behavior, not opt-in. Auto-discovered Pi / pi-signed sessions always add a quota footer row; there is no captain enable flag. VISION wants quota inspectable and captain-owned, but presentation features must ship as an option to enable — always-on footer noise assumes consent. I reviewed the full diff (extension + lib, credential correlation, stdin-closed quota-axi, tests, CI pin of Pi 0.84.2) with no security finding that would block fork CI, and approved pending workflow runs so checks can actually run. Not merge-eligible even if CI goes green. Head is 4 commits behind main, including `ci.yml`. Will flag NOW only once fully green + safe; empty checks are not a flag.

SamSherpaDev:
Closing: no longer pursuing this change.

PR #2788, opened 2026-08-22
fix(bin): record a lost relay connection instead of an unanswered turn

Inthuson:
## Intent

Correct three defects that reached the default branch with the spoken interface in pull request 2767. Remove a private, gitignored fleet record citation from the bin/fm-voice-relay.py docstring and reword the two comments that referred to an unpublished survey as though a reader could open it, weakening the claim rather than publishing anything to make it resolve. Land the transport failure fix that was pushed fifteen minutes after 2767 merged and so never reached the default branch, so a closed connection can no longer be recorded as a run with answered false and relay_error null, which would let an infrastructure failure be averaged into a latency figure. Keep .greptile/rules.md and narrow it with one sentence saying it is the first mate working interpretation, that whether VISION.md should be reconciled remains an open question belonging to the captain, and that the conditions listed below are what the interpretation depends on.

## What Changed

- `bin/fm-voice-client.py` no longer turns a dead connection into a normal-looking turn record: `take_turn` checks the closure mark inside the same critical section the downlink sets it in and returns `None` instead of opening a turn, every close path (`the connection was lost`, `the connection ended`, `the relay stopped`, `session-ended`, and a new guard around this end's own frame handling) names its cause in `relay_error`, reply audio is credited per turn via `turn_bytes` and a turn id carried with each frame rather than a shared byte total, audio arriving after the output is released is discarded and counted rather than raising, and `run()` exits non-zero when a record carries a `relay_error` or when the session stops before taking the requested runs. `bin/fm-voice-relay.py` now sends a `turn-failed` notice with the uplink fault down the connection before its goodbye, so the reason reaches the run record instead of only stderr.
- Provenance cleanup: the relay docstring's citation of a private, gitignored survey report is gone and the two chunk-size comments in the relay and client now say the timings came from earlier prototype work rather than pointing at an unpublished document. `docs/voice-relay.md` narrows the unverified claim to the audio devices themselves, documents what `relay_error` reports and how a cut-short reply differs from an unanswered turn, and states the exit-code rule.
- `.greptile/rules.md` gains one sentence marking the captain-facing-surface section as the first mate's working interpretation, leaving whether `VISION.md` should be reconciled as an open question for the captain and pointing at the listed conditions the interpretation rests on. `tests/fm-voice-relay.test.sh` adds around 1,500 lines of coverage for these paths, including new sections for a model session ending while the captain is still talking and an uplink that stops being a frame stream.

## Risk Assessment

✅ Low: Every intent criterion is source-verifiable and met, the newest commit is a two-line behavioral change pinned in both directions by new deterministic tests, and nothing above info severity survived verification - the three residual items are narrow races and a test dedup, all safe as follow-ups.

## Testing

Ran the feature's own suite (`tests/fm-voice-relay.test.sh`, all cases pass) and then, because a passing suite does not show the captain's results file, drove the real client CLI in a before/after harness: the same stub relay and the same real relay against the default-branch client and the fixed one. That reproduces the forbidden shape on base — a closed connection recorded as `answered: false` with `relay_error: null`, and a relay killed mid-answer exiting 0 with a real `first_audio_s` — and shows each path named and exiting non-zero after the fix, with the healthy two-run control unchanged at exit 0. The real relay run adds the other half: a model session ending mid key press now carries `relay_error: 'the relay ended the session before this turn was answered'`, and a desynchronised uplink sends a `turn-failed` notice ahead of its goodbye instead of stranding the reason on stderr. The citation removal, the two reworded comments and the narrowed `.greptile/rules.md` were verified by hand and captured as transcripts. No screenshot or rendered-UI artifact applies: the surfaces this change touches are a terminal CLI (captured as transcripts plus the raw runs.jsonl records), source comments, and markdown docs, with no rendered UI anywhere in the change. Transient `bin/__pycache__` from the runs was removed; the worktree is clean.

<details>
<summary>Evidence: Evidence index tying every artifact to the intent it demonstrates</summary>

```text
# Test evidence: voice relay build v4 corrections

Branch `fm/voice-relay-build-v4-corrections`, base `dc0172c`, target `dfe12ac`.

Everything below is the captain's own surface: `bin/fm-voice-client.py` run from a
shell with `--runs N > runs.jsonl`, which is exactly how `docs/voice-relay.md`
tells them to take a measurement. Each comparison runs the **same** stub or real
relay twice - once against the client as it stands on the default branch (`base`)
and once against the fix under test (`head`) - so the difference in the transcript
is the fix and nothing else.

| file | what it shows |
| --- | --- |
| `before-after-transport-failure.txt` | four scenarios through the real client CLI and the real wire format against a stub relay: a relay that dies before any reply, one that dies partway through the answer, one that dies between two runs, and a healthy control |
| `before-after-real-relay.txt` | the same defect through the **real** `bin/fm-voice-relay.py` and its real records reader, with only the paid Bedrock model stood in for: the model session ends mid key press, run 1 has no answer, run 2 does |
| `before-after-relay-names-the-fault.txt` | the frames the real relay puts on the wire when the uplink desynchronises: base tells the laptop only goodbye, head names the fault first |
| `citation-and-wording.txt` | the gitignored fleet-record citation is gone from `bin/fm-voice-relay.py`, no tracked file cites it, and what a reader now sees in its place |
| `greptile-rule-narrowed.txt` | `.greptile/rules.md` is kept and narrowed by one sentence |
| `voice-relay-suite.log` | `bash tests/fm-voice-relay.test.sh`, the suite that owns this feature |
| `harness/` | the scripts that produced the transcripts, and the base/head copies of the client they ran |
| `out/` | the raw `runs.jsonl`, stderr and reply audio from every run above |

## The defect, in the captain's own results file

The forbidden shape is a closed connection recorded as a run with `answered`
false and `relay_error` null, because in `runs.jsonl` that reads exactly like a
turn the model declined, and `runs.jsonl` is where the published latency spread
comes from.

    base   run 1: answered=False relay_error=None                                              exit 1
    head   run 1: answered=False relay_error='the connection ended before this turn was answered'   exit 1

And the shape that let an infrastructure failure be averaged into a latency
figure - a relay killed while speaking, at the default of one run:

    base   run 1: answered=True relay_error=None      first_audio_s=0.0   exit 0
    head   run 1: answered=True relay_error='the connection ended before the reply finished'  exit 1

And a session that took one of the two runs asked for:

    base   1 record, nothing said, exit 0
    head   1 record, "the connection ended before run 2 of 2; it and the rest were not taken", exit 1

The healthy control is unchanged: two whole answers, no `relay_error`, exit 0.

## Reproducing

    bash harness/run_scenarios.sh                 # stub relay, four scenarios
    bash harness/run_real_relay.sh <repo root>    # the real relay
    bash tests/fm-voice-relay.test.sh             # from the repo
```
</details>
<details>
<summary>Evidence: Before/after through the real client CLI: four transport-failure scenarios and a healthy control</summary>

<code>===== A-relay-dies-before-any-reply | client=base =====&#10;run 1: answered=False relay_error=None first_audio_s=None reply_audio_seconds=0.0&#10;--- exit code: 1 ---&#10;&#10;===== A-relay-dies-before-any-reply | client=head =====&#10;client: the connection ended&#10;run 1: answered=False relay_error=&#39;the connection ended before this turn was answered&#39; first_audio_s=None reply_audio_seconds=0.0&#10;--- exit code: 1 ---&#10;&#10;===== B-relay-dies-partway-through-the-answer | client=base =====&#10;run 1: answered=True relay_error=None first_audio_s=0.0 reply_audio_seconds=0.125&#10;--- exit code: 0 ---&#10;&#10;===== B-relay-dies-partway-through-the-answer | client=head =====&#10;run 1: answered=True relay_error=&#39;the connection ended before the reply finished&#39; first_audio_s=0.0 reply_audio_seconds=0.125&#10;--- exit code: 1 ---&#10;&#10;===== C-relay-dies-between-two-runs | client=base =====&#10;run 1: answered=True relay_error=None ...&#10;--- exit code: 0 ---&#10;&#10;===== C-relay-dies-between-two-runs | client=head =====&#10;client: the connection ended before run 2 of 2; it and the rest were not taken&#10;run 1: answered=True relay_error=None ...&#10;--- exit code: 1 ---&#10;&#10;===== D-control-two-good-runs | client=base / head =====&#10;run 1 and run 2 answered, relay_error=None, exit code: 0 (both)</code>

```text
stub relay stands in for the desktop; the client, the wire format and the
record shape are the real ones. No AWS account, microphone or speaker.

===== A-relay-dies-before-any-reply | client=base | stub=drop-before-reply | --runs 1 =====
--- what the captain sees on the terminal (stderr) ---
    client: relay ready, stub-relay in us-east-1, read scope counts, connected in 0.01s
    stub-relay: dying before the reply
--- runs.jsonl, the file the latency spread is computed from ---
    run 1: answered=False relay_error=None first_audio_s=None reply_audio_seconds=0.0
--- exit code: 1 ---

===== A-relay-dies-before-any-reply | client=head | stub=drop-before-reply | --runs 1 =====
--- what the captain sees on the terminal (stderr) ---
    client: relay ready, stub-relay in us-east-1, read scope counts, connected in 0.01s
    stub-relay: dying before the reply
    client: the connection ended
--- runs.jsonl, the file the latency spread is computed from ---
    run 1: answered=False relay_error='the connection ended before this turn was answered' first_audio_s=None reply_audio_seconds=0.0
--- exit code: 1 ---

===== B-relay-dies-partway-through-the-answer | client=base | stub=cut-mid-reply | --runs 1 =====
--- what the captain sees on the terminal (stderr) ---
    client: relay ready, stub-relay in us-east-1, read scope counts, connected in 0.01s
    stub-relay: dying partway through the reply
--- runs.jsonl, the file the latency spread is computed from ---
    run 1: answered=True relay_error=None first_audio_s=0.0 reply_audio_seconds=0.125
--- exit code: 0 ---

===== B-relay-dies-partway-through-the-answer | client=head | stub=cut-mid-reply | --runs 1 =====
--- what the captain sees on the terminal (stderr) ---
    client: relay ready, stub-relay in us-east-1, read scope counts, connected in 0.01s
    stub-relay: dying partway through the reply
    client: the connection ended
--- runs.jsonl, the file the latency spread is computed from ---
    run 1: answered=True relay_error='the connection ended before the reply finished' first_audio_s=0.0 reply_audio_seconds=0.125
--- exit code: 1 ---

===== C-relay-dies-between-two-runs | client=base | stub=die-between-turns | --runs 2 =====
--- what the captain sees on the terminal (stderr) ---
    client: relay ready, stub-relay in us-east-1, read scope counts, connected in 0.01s
    stub-relay: dying after the first answer
--- runs.jsonl, the file the latency spread is computed from ---
    run 1: answered=True relay_error=None first_audio_s=0.0 reply_audio_seconds=0.125
--- exit code: 0 ---

===== C-relay-dies-between-two-runs | client=head | stub=die-between-turns | --runs 2 =====
--- what the captain sees on the terminal (stderr) ---
    client: relay ready, stub-relay in us-east-1, read scope counts, connected in 0.01s
    stub-relay: dying after the first answer
    client: the connection ended
    client: the connection ended before run 2 of 2; it and the rest were not taken
--- runs.jsonl, the file the latency spread is computed from ---
    run 1: answered=True relay_error=None first_audio_s=0.0 reply_audio_seconds=0.125
--- exit code: 1 ---

===== D-control-two-good-runs | client=base | stub=healthy | --runs 2 =====
--- what the captain sees on the terminal (stderr) ---
    client: relay ready, stub-relay in us-east-1, read scope counts, connected in 0.01s
--- runs.jsonl, the file the latency spread is computed from ---
    run 1: answered=True relay_error=None first_audio_s=0.0 reply_audio_seconds=0.125
    run 2: answered=True relay_error=None first_audio_s=0.0 reply_audio_seconds=0.125
--- exit code: 0 ---

===== D-control-two-good-runs | client=head | stub=healthy | --runs 2 =====
--- what the captain sees on the terminal (stderr) ---
    client: relay ready, stub-relay in us-east-1, read scope counts, connected in 0.01s
    client: the relay signed off
--- runs.jsonl, the file the latency spread is computed from ---
    run 1: answered=True relay_error=None first_audio_s=0.0 reply_audio_seconds=0.125
    run 2: answered=True relay_error=None first_audio_s=0.0 reply_audio_seconds=0.125
--- exit code: 0 ---
```
</details>
<details>
<summary>Evidence: Before/after through the REAL bin/fm-voice-relay.py: a model session ending mid key press</summary>

<code>===== E-model-session-ends-mid-key-press | client=base | real bin/fm-voice-relay.py =====&#10;client: the relay ended the session&#10;assistant: Right now, 2 in flight, 1 waiting on you, 1 open pull requests. The ones moving are alpha-one and beta-two.&#10;run 1: answered=False relay_error=None first_audio_s=None reply_audio_seconds=0.0&#10;run 2: answered=True relay_error=None first_audio_s=0.409 reply_audio_seconds=0.4&#10;--- exit code: 1 ---&#10;&#10;===== E-model-session-ends-mid-key-press | client=head | real bin/fm-voice-relay.py =====&#10;client: the relay ended the session&#10;assistant: Right now, 2 in flight, 1 waiting on you, 1 open pull requests. The ones moving are alpha-one and beta-two.&#10;run 1: answered=False relay_error=&#39;the relay ended the session before this turn was answered&#39; first_audio_s=None reply_audio_seconds=0.0&#10;run 2: answered=True relay_error=None first_audio_s=0.409 reply_audio_seconds=0.4&#10;--- exit code: 1 ---</code>

```text
The relay, the records reader, the client and the wire format are real.
Only the paid Bedrock model is stood in for.

===== E-model-session-ends-mid-key-press | client=base | real bin/fm-voice-relay.py =====
--- what the captain sees on the terminal (stderr) ---
    client: relay ready, amazon.nova-2-sonic-v1:0 in eu-north-1, read scope full, connected in 0.0s
    client: the relay ended the session
      you: how is the fleet doing right now
      assistant: Right now, 2 in flight, 1 waiting on you, 1 open pull requests. The ones moving are alpha-one and beta-two.
--- runs.jsonl, the file the latency spread is computed from ---
    run 1: answered=False relay_error=None first_audio_s=None reply_audio_seconds=0.0
    run 2: answered=True relay_error=None first_audio_s=0.409 reply_audio_seconds=0.4
--- exit code: 1 ---

===== E-model-session-ends-mid-key-press | client=head | real bin/fm-voice-relay.py =====
--- what the captain sees on the terminal (stderr) ---
    client: relay ready, amazon.nova-2-sonic-v1:0 in eu-north-1, read scope full, connected in 0.0s
    client: the relay ended the session
      you: how is the fleet doing right now
      assistant: Right now, 2 in flight, 1 waiting on you, 1 open pull requests. The ones moving are alpha-one and beta-two.
    client: the relay signed off
--- runs.jsonl, the file the latency spread is computed from ---
    run 1: answered=False relay_error='the relay ended the session before this turn was answered' first_audio_s=None reply_audio_seconds=0.0
    run 2: answered=True relay_error=None first_audio_s=0.409 reply_audio_seconds=0.4
--- exit code: 1 ---
```
</details>
<details>
<summary>Evidence: The frames the real relay puts on the wire when the uplink desynchronises</summary>

<code>===== the uplink desynchronises mid-turn | relay=base (default branch) =====&#10;--- frames the relay sent down to the laptop, in order ---&#10;BYE&#10;--- the relay&#39;s own stderr, which no runs.jsonl quotes ---&#10;fm-voice-relay: the uplink is not a frame stream any more: unknown frame kind: b&#39;\xff&#39;&#10;--- exit code: 2 ---&#10;&#10;===== the uplink desynchronises mid-turn | relay=head (fix under test) =====&#10;--- frames the relay sent down to the laptop, in order ---&#10;NOTICE {&#39;event&#39;: &#39;turn-failed&#39;, &#39;error&#39;: &#34;FrameError: unknown frame kind: b&#39;\\xff&#39;&#34;}&#10;BYE&#10;--- the relay&#39;s own stderr, which no runs.jsonl quotes ---&#10;fm-voice-relay: the uplink is not a frame stream any more: unknown frame kind: b&#39;\xff&#39;&#10;--- exit code: 2 ---</code>

```text

===== the uplink desynchronises mid-turn | relay=base (default branch) =====
--- frames the relay sent down to the laptop, in order ---
    BYE   
--- the relay's own stderr, which no runs.jsonl quotes ---
    fm-voice-relay: the uplink is not a frame stream any more: unknown frame kind: b'\xff'
--- exit code: 2 ---

===== the uplink desynchronises mid-turn | relay=head (fix under test) =====
--- frames the relay sent down to the laptop, in order ---
    NOTICE {'event': 'turn-failed', 'error': "FrameError: unknown frame kind: b'\\xff'"}
    BYE   
--- the relay's own stderr, which no runs.jsonl quotes ---
    fm-voice-relay: the uplink is not a frame stream any more: unknown frame kind: b'\xff'
--- exit code: 2 ---
```
</details>
<details>
<summary>Evidence: The gitignored fleet-record citation is gone, and what a reader now sees in its place</summary>

<code>$ git check-ignore -v data/speech-to-speech-survey-s2/report.md&#10;.gitignore:3:data/ data/speech-to-speech-survey-s2/report.md&#10;$ git grep -n &#39;speech-to-speech-survey&#39; -- . | wc -l&#10;0&#10;$ git grep -n &#39;the survey&#39; -- . | wc -l&#10;0&#10;&#10;The two traps this code already avoids, both found the expensive way and both&#10;measured rather than assumed:&#10;&#10;# 3200 bytes is 100 ms at 16 kHz 16-bit mono, the chunk size earlier prototype&#10;# work measured its timings with. Keeping it identical keeps those comparable.</code>

```text
### 1. The private, gitignored path is unreachable to a reader of this repo

$ git check-ignore -v data/speech-to-speech-survey-s2/report.md
.gitignore:3:data/	data/speech-to-speech-survey-s2/report.md
$ ls data   # not present in a fresh clone either
ls: cannot access 'data': No such file or directory

### 2. No tracked file cites that record any more
$ git grep -n 'speech-to-speech-survey' -- . | wc -l
0
$ git grep -n 'the survey' -- . | wc -l
0

### 3. What a reader of bin/fm-voice-relay.py now sees where the citation was

  --self-test FILE   feed one raw 16 kHz PCM file into a session as if it had
                     arrived from the client, print the timings as JSON, exit.
                     This is the control measurement for the relay path, and it
                     needs no client, no SSH and no microphone.

The two traps this code already avoids, both found the expensive way and both
measured rather than assumed:

  1. completionEnd does not arrive on its own. The model holds the session open
     waiting for more speech. The real "the reply is finished" signal is a
     contentEnd carrying stopReason END_TURN.
  2. Audio with no trailing silence is truncated and never answered, even when
     contentEnd follows immediately. A push-to-talk release supplies no trailing

    "voice": ("voice-id", "FM_VOICE_ID", "output voice"),
}

IN_RATE = 16000
OUT_RATE = 24000
# 3200 bytes is 100 ms at 16 kHz 16-bit mono, the chunk size earlier prototype
# work measured its timings with. Keeping it identical keeps those comparable.
CHUNK = 3200

### 4. And the client's copy of the same claim


IN_RATE = 16000
OUT_RATE = 24000
# 100 ms at each rate. The uplink chunk matches what the relay and the earlier
# prototype work measured with; changing it changes the numbers.
CHUNK = 3200
```
</details>
<details>
<summary>Evidence: .greptile/rules.md kept and narrowed by one sentence</summary>

<code>.greptile/rules.md | 2 ++&#10;1 file changed, 2 insertions(+)&#10;&#10;+This is the first mate&#39;s working interpretation rather than settled repository policy; whether `VISION.md` itself should be reconciled remains an open question belonging to the captain; and the conditions listed below are what this interpretation depends on.&#10;&#10;$ git ls-files .greptile # still tracked, not removed&#10;.greptile/rules.md</code>

```text
### .greptile/rules.md is kept, and narrowed by one sentence

$ git diff --stat dc0172c..dfe12ac -- .greptile/rules.md
 .greptile/rules.md | 2 ++
 1 file changed, 2 insertions(+)

$ git diff dc0172c..dfe12ac -- .greptile/rules.md
diff --git a/.greptile/rules.md b/.greptile/rules.md
index 79908ed..42b5c4f 100644
--- a/.greptile/rules.md
+++ b/.greptile/rules.md
@@ -7,6 +7,8 @@ dropped stops being a decision and becomes a blind spot.
 
 ## A captain-facing surface is not automatically a `VISION.md` violation
 
+This is the first mate's working interpretation rather than settled repository policy; whether `VISION.md` itself should be reconciled remains an open question belonging to the captain; and the conditions listed below are what this interpretation depends on.
+
 `VISION.md` says "The captain talks to the first mate and to nobody else; every
 worker reports through the first mate and never addresses the captain directly."
 That line protects who is answerable for work. Read alongside the sentence it

$ git ls-files .greptile   # still tracked, not removed
.greptile/rules.md
```
</details>
<details>
<summary>Evidence: tests/fm-voice-relay.test.sh full output (all cases pass)</summary>

```text
ok - wire format round trips and rejects a desynchronised stream
ok - the relay refuses a desynchronised uplink header instead of waiting for its payload
ok - the relay reads whose account to use from this home and refuses to guess
ok - an unconfigured home can still read how to configure the relay
ok - the model-backed subcommands refuse by name while note keeps working
ok - fm-inbox.sh --help prints its whole header, privacy paragraph included
ok - the relay and the reader agree on the tool names and the handover argument
ok - credentials are resolved once per relay, refreshed before expiry, off the event loop
ok - a turn the model refuses is announced and costs one turn, not the relay
ok - audio that arrives with no turn open is dropped, not turned into a turn
ok - a failure inside the model reader costs one turn, not every later one
ok - a reader failure is named to the captain, a clean end and a close are not
usage: fm-voice-client.py [-h] [--host HOST] [--local] [--relay RELAY]
                          [--relay-python RELAY_PYTHON]
                          [--relay-arg RELAY_ARG]
                          [--listen {push-to-talk,open-mic}] [--runs RUNS]
                          [--talk-seconds TALK_SECONDS] [--in-file IN_FILE]
                          [--out-file OUT_FILE] [--input-device INPUT_DEVICE]
                          [--output-device OUTPUT_DEVICE] [--timeout TIMEOUT]
                          [--wait-for-reply | --no-wait-for-reply]
                          [--gap-seconds GAP_SECONDS]
                          [--audio-idle AUDIO_IDLE] [--verbose]
fm-voice-client.py: error: say where the relay is: --relay <path to fm-voice-relay.py on the desktop>, or set FM_VOICE_RELAY
usage: fm-voice-client.py [-h] [--host HOST] [--local] [--relay RELAY]
                          [--relay-python RELAY_PYTHON]
                          [--relay-arg RELAY_ARG]
                          [--listen {push-to-talk,open-mic}] [--runs RUNS]
                          [--talk-seconds TALK_SECONDS] [--in-file IN_FILE]
                          [--out-file OUT_FILE] [--input-device INPUT_DEVICE]
                          [--output-device OUTPUT_DEVICE] [--timeout TIMEOUT]
                          [--wait-for-reply | --no-wait-for-reply]
                          [--gap-seconds GAP_SECONDS]
                          [--audio-idle AUDIO_IDLE] [--verbose]
fm-voice-client.py: error: --listen open-mic is not built yet: it needs end-of-speech detection to know when a turn ended, which lands with session continuity across turns, so it would stream forever and never end a turn. Use the default --listen push-to-talk.
client: connection lost: stream ended after 2 of the 5 bytes of a frame header
client: connection lost: [Errno 104] Connection reset by peer
client: discarded 15 bytes your login shell printed before the relay started: b'You have mail.\n'
ok - the laptop client builds the right remote command and survives a chatty login shell
ok - the client runs from a laptop holding only the two files the guide names
ok - --listen open-mic refuses at startup, before it opens anything
ok - the narrow scope answers how much is waiting without saying what it is
ok - the wide scope names open work and reports state without quoting event lines
ok - an absent read-scope setting means the narrowest answer, not the widest
ok - a home widens its own read scope by writing the setting
ok - finished work and note bodies never reach a spoken answer at any scope
ok - excluding a note body does not distort the counts
ok - a denied item becomes a withheld count without hiding that work exists
ok - the deny list matches regardless of case
ok - an item denied in more than one list is counted as withheld once
ok - one deny decision per item covers every list that item could appear in
ok - the state verb is a closed vocabulary, so free text cannot ride out on it
ok - the backlog and the state directory always come from the same home
ok - a finished task's pull request is neither counted nor named
ok - a deny substring on an open title removes its pull request and says so
ok - an unknown read scope refuses instead of widening
ok - the configured read scope is honoured
ok - handover queues the request for firstmate and wakes it exactly once
ok - the reader and the queue resolve the state directory the same way
ok - an empty request is refused rather than queued as a blank note
ok - a home with no records answers nothing rather than failing
ok - a spoken turn goes out and comes back: the records answer, firstmate gets the work
ok - a model session that ends mid-conversation costs one turn, not the relay
ok - a desynchronised uplink is named to the client before the goodbye, not only to stderr
ok - the desktop's own check answers from the records and times it from the talk end
ok - a reply that arrives before the end of the clip is named as an unusable clock
all voice relay cases passed
```
</details>
- Evidence: Raw runs.jsonl, stderr and reply audio from every base/head run above (local file: <code>/tmp/no-mistakes-evidence/01M0KNX9WMG97PK718REV0BN01/out</code>)
- Evidence: The harness that produced the transcripts, with the base and head client copies it ran (local file: <code>/tmp/no-mistakes-evidence/01M0KNX9WMG97PK718REV0BN01/harness</code>)

## Pipeline

_... (6 earlier update rounds omitted to keep the PR body within GitHub's 65536-char limit; full history is in the run log.)_

<details>

🔧 Fix: tell a cut-short reply from an unanswered turn
- ⚠️ `bin/fm-voice-client.py:837` - bin/fm-voice-client.py:837 - the frame-handling guard still reports this end&#39;s own teardown as a fault, and the mitigation added for it at line 606 is bounded where the thing it waits on is not. The comment at 599-605 says joining the downlink before releasing the devices &#34;keeps the fault line meaning a fault&#34;, but the join is `timeout=5` (line 607) while the relay&#39;s teardown after QUIT is `await session.close()`, which sends contentEnd/promptEnd/sessionEnd and then awaits `asyncio.wait_for(gather(reader_task), timeout=10)` (bin/fm-voice-relay.py:600-604) - the reader task being the very thing that emits reply AUDIO. So the join can expire with the downlink alive and audio still in flight, `playback.close()` runs at line 612, and the next AUDIO frame hits `self._handle.write(pcm)` on a closed handle (line 225), raising ValueError into the broad handler: &#34;client: this end could not handle the relay&#39;s reply: ValueError: I/O operation on closed file&#34; plus a full `traceback.format_exc()` (line 845). Concrete sequence that exits 0 while printing it: a reply longer than `--timeout` at `--runs 1`, so `reply_end` never arrives; the turn is recorded `answered: true` (audio did arrive) with `relay_error: null`, `run()` returns 0 at line 1137 before main&#39;s `finally` calls `close()`; the relay is mid-reply, its teardown exceeds 5s, and the alarm lands after the record. That is both a wrong cause (the connection is healthy; the file was closed by us) and an alarm on a run reporting success, which is the pair of rules this branch has been enforcing. `self.quitting` is already set before QUIT (line 598) and is the discriminator: gate the say() and the traceback on it, downgrading to `log(self.verbose, ...)` during teardown, or make FilePlayback.write a no-op once closed so teardown cannot manufacture a fault at all.
- ℹ️ `tests/fm-voice-relay.test.sh:1282` - tests/fm-voice-relay.test.sh:1282 - the round that narrowed the module docstring to say the speaker&#39;s byte accounting IS verified left the mirror claim in the test file asserting the opposite, in the same file that now contains the coverage. The header comment reads &#34;The microphone and speaker paths cannot be tested from a host with neither, and are not tested anywhere: the first live run is their test&#34;, while lines 2289-2434 of the same block drive `SpeakerPlayback.write`, `turn_reset` and `_callback` against a stub stream. A reader of the test file is told there is no speaker coverage roughly 1000 lines above the coverage. docs/voice-relay.md:123-124 carries the same now-imprecise sentence (&#34;the capture and playback paths have never run&#34;); that one errs toward understating, which is the safe direction, but both would read better narrowed the way the docstring was - the DEVICES are unverified, the accounting is covered - and the change&#39;s own standard is that a claim about what is verified has to be exact in both directions.
- ℹ️ `bin/fm-voice-client.py:225` - bin/fm-voice-client.py:225 - the new FilePlayback lock is correct but wider than it needs to be, and it puts blocking I/O inside a critical section another thread enters while holding the global turn lock. `take_turn` calls `self.playback.turn_reset(self.turn_id)` at line 907 from inside `with self.lock`, so it blocks on `playback._lock` for however long the downlink&#39;s `self._handle.write(pcm)` takes; while it does, the downlink cannot stamp the next frame and `_sender` cannot stamp `wire_end`, since both need `self.lock`. There is no deadlock today - nothing acquires `self.lock` while holding `playback._lock` - but the ordering is one edit away from one, and tests/fm-voice-relay.test.sh:2189 already had to park its fixture&#39;s wait outside the parent&#39;s lock to avoid exactly that. Only the downlink thread ever writes to the handle, so the decision, the stamp and the count can stay under the lock with the write left outside it: file ordering is preserved by the single writer, and the stamp already precedes the write today, so the semantics are unchanged. The one difference is that a write which raises would leave its bytes counted, which the terminal handling-fault path at 822 renders moot.

🔧 Fix: discard reply audio arriving after the output closes
3 infos still open:
- ℹ️ `bin/fm-voice-client.py:665` - bin/fm-voice-client.py:665 - the discard accounting exists on one of the two playback classes only, and the getattr hides that rather than recording it. An AST sweep confirms `_closed` and `discarded` appear nowhere outside FilePlayback (lines 214-267); SpeakerPlayback (271-359) has neither, so `getattr(self.playback, &#34;discarded&#34;, 0)` is always 0 on the laptop path and `_say_dropped` says nothing there. The audit conclusion this round acted on is right about raising: `SpeakerPlayback.write` (326-332) only appends to `self._buffer` under `self._lock` and never touches `self._stream`, so a post-close write cannot reach the frame-handling guard, and the ruling said to change nothing in that case. What is left over is the diagnostic, not the fault: after `close()` stops the stream (354-359) no callback drains the buffer, so reply audio arriving in the same teardown window is appended to a bytearray that is never played and never counted, while the file path counts it. `_say_dropped`&#39;s own docstring (659-663) states the property that justifies keeping the counter - &#34;a count appearing on a session that has not ended is the logic bug it is kept for&#34; - and that property holds only for --out-file, so the write-after-close logic bug the counter exists to catch is undetectable on the path the captain will actually use. Mirroring `_closed` plus `discarded` into SpeakerPlayback.close/write is a few lines and would let the getattr become a plain attribute read; flagged as a question rather than a fix because the earlier ruling scoped the speaker change to the raising case, and this is the diagnostic half of it.
- ℹ️ `tests/fm-voice-relay.test.sh:2170` - tests/fm-voice-relay.test.sh:2170 - the only test for condition 1 proves the reporting method works, not that anything calls it. `audio_after_close` hand-calls `late._say_dropped()`, and `Client.close()` (the sole production caller, bin/fm-voice-client.py:656) is never driven anywhere in the added coverage, so the assertion at 2212 (&#34;discarded 1 reply audio chunk&#34; in loud_said) would still pass with line 656 deleted and the count consequently never reported on a real session. Everything else in this case is pinned through the public surface, which is why this one stands out. The fix is cheap and stays in the existing style: this fixture leaves `late.down_thread` and `late.proc` as None and its uplink stand-in swallows QUIT, so `late.close()` returns immediately, and replacing the direct `_say_dropped()` call with `late.close()` after `handled.wait(10)` exercises the wiring while keeping the required ordering (playback closed at 2167 before `released.set()` at 2168). Note `close()` also sets `quitting`, which is inert here.
- ℹ️ `bin/fm-voice-client.py:266` - bin/fm-voice-client.py:266 - recording a tradeoff this round introduced rather than a defect to fix. `write` now holds `_handle_lock` across the blocking `self._handle.write(pcm)` (243-254) and `close` needs the same lock, so `Client.close()`&#39;s `self._quietly(&#34;the speaker&#34;, self.playback.close)` (644) can wait for however long a write to --out-file takes. Before this commit `close()` took no lock and returned at once, so the deliberately bounded `down_thread.join(timeout=5)` (639) really did bound the exit; it no longer does if that write stalls, and `_quietly` catches exceptions but cannot interrupt a block. Reachability is thin: the handle is a BufferedWriter over a regular file, so a stall needs a full or hung filesystem, and the wedged-relay case the 5s bound was written for is unaffected. It is also hard to do better - dropping the lock is what let `close()` slip between the `_closed` check and the write and produce the ValueError fault line this round removed - so the current choice looks like the right one. Worth knowing it was a choice: the comment at 209-213 explains which lock covers what but not that the exit is now only as bounded as a single file write.

🔧 Fix: count discarded reply audio on the speaker path too
- ⚠️ `bin/fm-voice-client.py:797` - bin/fm-voice-client.py:797 and 864 - two of the four paths that write turn[&#34;failed&#34;] lack the reply_done guard the other two have, so a fully answered turn can be recorded with a relay_error, and round 5&#39;s new rule at line 1211 turns that into exit 1 on a session where every turn was answered. The window is exact: take_turn&#39;s reply_done.wait returns the instant reply_end arrives (1012), then _wait_audio_quiet spins until now - last_frame &gt;= audio_idle (1090-1098, default 0.4s), and only then is the dict copied at 1016-1017. For that ~0.4s the downlink is free to write into the live turn. (a) reader.read() raising OSError, an ECONNRESET rather than a clean FIN, hits 797 and its setdefault lands because a good turn has no &#34;failed&#34;. (b) the relay&#39;s model reader task breaking after the reply hits fail_turn (bin/fm-voice-relay.py:766, 889), which the client applies at 864 with a plain assignment and arrived_in == turn_id still true because take_turn has not advanced. Either way the record reads answered: true with a relay_error, and `if not record[&#34;answered&#34;] or record[&#34;relay_error&#34;]` sets rc = 1; at the default --runs 1 that is a non-zero code on a session that delivered its answer, and the docs line added by this change (docs/voice-relay.md:155) says relay_error &#34;is null when nothing broke&#34;. The same physical event through the two quiet paths does the opposite: a clean end of stream or an unasked goodbye in that identical window reaches the tail at 955-957, whose `not self.reply_done.is_set()` test declines, records nothing and exits 0 - which tests/fm-voice-relay.test.sh:1751-1754 pins as the required negative case. So EOF versus RST, differing only in what the kernel delivered, now produce different records and different exit codes for one relay death, and nothing covers the raising variants: tests/fm-voice-relay.test.sh:1500 and 1658 both drive the read-failure path on turns that were never answered. Earliest shared boundary: apply the same `not self.reply_done.is_set()` test the tail and the session-ended handler at 877 already use to both of these writes, so all four close paths agree on when a turn earns a reason. Guarding cannot swallow a genuine failure, because take_turn clears reply_done for the whole life of an open turn. Raised as ask-user rather than auto-fix because it changes what a record says and what the exit code means, which is exactly what you widened on purpose in round 5.
- ℹ️ `bin/fm-voice-client.py:689` - bin/fm-voice-client.py:689 and 691-706 - two small mismatches in the round 7 diagnostic. First, the docstring&#39;s discriminating claim, &#34;a count appearing on a session that has not ended is the logic bug it is kept for&#34;, cannot be observed: _say_dropped has exactly one caller, the last statement of close() itself, so the count is only ever reported at session end and a reader can never see it on a session that has not ended. The tripwire is real, a future close() outside teardown would still be counted, but the reported line would be indistinguishable from the benign teardown race the counter exists to tolerate, so the sentence promises a distinction the code cannot make. Second, dropping the getattr for a plain self.playback.discarded read makes this the one unguarded step in a method whose own comment at 656-659 says &#34;Every step is guarded and every field is checked... the original refusal is the message worth keeping&#34;: every other step goes through _quietly, and close() runs from open()&#39;s except BaseException at 582-584, so an AttributeError here would replace the startup refusal that is the whole reason for cleaning up, and in main&#39;s finally at 1321-1322 it would replace run()&#39;s return value. Not reachable today, since both playback classes define discarded and the tests&#39; Recorder now does too, which is why this is info rather than a defect. Either wrap the call in _quietly like its neighbours, or narrow the docstring to say what the counter actually gives, which is a count read at teardown that should normally be zero.
- ℹ️ `tests/fm-voice-relay.test.sh:1913` - tests/fm-voice-relay.test.sh:1913 and 1806 - turn_cut and turn_lost are the same helper twice. Both build a Client from the same flag list, wrap a ScriptedStream in frame.Reader, attach StartGate, a FilePlayback and a FileCapture over clip.pcm, start _sender and _downlink as daemon threads, run under redirected stdout and stderr, drain up_q, close the playback and parse the records; they differ only in the run count and in which assertions they then make. The same eight-line wiring block is also written out inline at 2030 (handling), 2158 (audio_after_close), 2316 (stale), 2401 (late) and 2459 (prompt). One `wire_client(args, reader, uplink, playback=None)` returning the client, plus turn_cut expressed as turn_lost with runs=&#34;1&#34; and its own assertion block, collapses roughly 70 of these lines with no change to what is asserted. Scoped deliberately: this touches none of EndingStream, CountingUplink, ClosingStream or StartGate, whose provenance from the verbatim cherry-pick is why you left them alone twice, and none of the cut/closing/gone/ending blocks that came with it. Every site listed above was authored by this run&#39;s own fix rounds, the same boundary you accepted for the five-stream dedup.

🔧 Fix: keep a reason off a turn already answered in full
2 infos still open:
- ℹ️ `bin/fm-voice-client.py:833` - bin/fm-voice-client.py:833 - the read-failure path is the only mid-turn reason that does not tell a truncated answer from an unanswered turn, and the docs added by this change say it does. `_unfinished` (756-772) is applied at 923 (session-ended) and at 1000 (the tail, covering both an ended stream and an unasked goodbye), which is exactly the three subjects docs/voice-relay.md:157 enumerates. The read at 787 raises for a fourth shape - an ECONNRESET rather than a clean FIN, or a header cut in half - and records the flat string &#34;the connection was lost: {exc}&#34;. So for a relay killed mid-reply on a connection that resets rather than closing, the record reads `answered: true`, `first_audio_s` measured, and a reason that says only that the connection went, while the doc line promises &#34;a turn whose answer had already started playing says the same thing happened before the reply finished&#34;. A reset is the likelier shape of a dropped SSH link than a clean FIN, so this is the half a reader is most likely to meet. Nothing pins it in either direction: the `turn_cut` loop at 1952-1991 drives only the three `_unfinished` subjects with SOME_REPLY, and the two read-failure cases (tests/fm-voice-relay.test.sh:1495 CutStream, 1635 LostStream) both run on turns that were never answered. Not a correctness break - the reason is non-null, rc is 1, and nothing contradicts `answered` - which is why this is info rather than a defect: `&#34;the connection was lost&#34;` makes no claim about whether audio arrived. The fix is to route this subject through `_unfinished` as well (`self._unfinished(&#34;the connection was lost&#34;)` plus the exception detail) and add the missing reset-after-partial-audio case, or to narrow the doc sentence so it does not promise a distinction one of the four paths does not make. Raised as ask-user because it changes user-visible record wording and a published promise, which is the author&#39;s call.
- ℹ️ `bin/fm-voice-client.py:980` - bin/fm-voice-client.py:980 - recording the result of the audit the fix round asked for, so a later round does not undo it. After round 8 this is the one `turn[&#34;failed&#34;]` write with neither the turn-identity gate nor the reply_done test: the read failure at 833 and the relay `turn-failed` at 907 gained the test, `session-ended` at 921 and the tail at 1000 already had it. It is reachable in the same post-answer window the guard exists for: reply audio belonging to turn N-1 arriving while turn N is open and already answered goes through `FilePlayback.write` with `mine` false, still reaches `self._handle.write(pcm)` at 263, and an OSError there (full or hung filesystem) is recorded against turn N and forces exit 1 through the rule at 1254. I judge leaving it ungated CORRECT and am not asking for a change. The symmetry argument that justified guarding 833 - &#34;an end of stream and a reset differ only in what the kernel handed us&#34; - has no partner here: a read failure means the far end went after delivering the whole answer, which cost the captain nothing, while this handler means THIS end could not put the audio it was handed anywhere, the session must stop, and rc 1 with a reason naming the real cause is the honest report. What is missing is only that the code does not say so, so the next reader has to re-derive it; the earlier note that these writes are &#34;ungated deliberately because they describe the connection rather than a frame&#34; no longer explains the split now that 833 carries the test.

🔧 Fix: say a reset cut a reply short, not that none arrived
3 infos still open:
- ℹ️ `bin/fm-voice-client.py:1107` - bin/fm-voice-client.py:1064, 1106 and 1107 - the record is built from a snapshot everywhere except the playback figures, which are three separate reads of live shared state. `played = self.playback.first_played` at 1064, `self.playback.turn_bytes` at 1106 for `reply_audio_seconds`, and `self.playback.turn_bytes` again at 1107 for `answered`, while `turn = dict(self.turn)` at 1062 deliberately snapshots the rest. A chunk written by the downlink between two of those reads produces the self-contradiction this branch exists to remove. Concrete sequence: a turn whose `reply_done.wait` expires at line 1057 while no audio has arrived; `_wait_audio_quiet` reads `last_frame` as None and returns immediately at 1139-1140 with no idle wait at all; `arrived_in == self.turn_id` still holds because take_turn has not advanced, so the reply&#39;s first AUDIO frame landing during dict construction stamps `first_played` and credits `turn_bytes`. If it lands between 1106 and 1107 the record reads `reply_audio_seconds: 0.0` with `answered: true`; between 1064 and 1106 it reads `answered: true` with `first_played_s: null` and `first_audio_s` falling back to a frame stamp the copied dict may not even carry. The window is microseconds wide and needs the reply to arrive at the exact timeout boundary, and the two-read shape predates this branch (base computed `self.playback.bytes - before` and `self.playback.bytes &gt; before` the same way), which is why this is info rather than a defect. It is worth closing because `_unfinished` at 770 was added precisely so a reason and `answered` are read off one input, and these two fields are the pair that reason is checked against. Fix is mechanical and non-functional: read `first_played` and `turn_bytes` once into locals before the dict literal, so the three figures describe one instant.
- ℹ️ `bin/fm-voice-client.py:841` - bin/fm-voice-client.py:841 - the end-of-stream branch is the only close path whose `say()` has no teardown discriminator, so it can print a mid-session fault line on a session that answered every turn and exits 0. The BYE branch at 940-954 was given exactly that discriminator (`self.quitting`, set at 664 before the QUIT frame) after round 4 found the goodbye firing on clean shutdown; the EOF path beside it was left unconditional. Reachable path, and it is a timing asymmetry rather than a coincidence: `close()` bounds the relay at `down_thread.join(timeout=5)` (672) and `proc.wait(timeout=10)` (684), while the relay&#39;s teardown after QUIT is `await session.close()` awaiting the model reader with `timeout=10` (bin/fm-voice-relay.py:600-602) plus `down.close()`&#39;s own `join(timeout=5)` (482), so a slow Bedrock stream close overruns both client bounds, `proc.kill()` fires at 686, the child&#39;s stdout closes without a BYE ever being written, and the still-live downlink thread reads b&#34;&#34; and prints &#34;client: the connection ended&#34; as the last human line of a successful run. The record and the exit code are unaffected (`reply_done` is set after the final answered turn, so the tail at 1001 correctly declines to name anything, and run() has already returned), so this is only the alarm-on-success rule the branch enforced twice, not a wrong record. Note it is also a latent flake: the clean end-to-end case asserts this exact string is absent (tests/fm-voice-relay.test.sh:4187). The 5s/10s/15s bounds themselves belong to the teardown-ordering work you filed and are out of scope; the change here is to gate this one line on `self.quitting` (or downgrade it to `log(self.verbose, ...)` when quitting), which leaves the mid-session case pinned at tests/fm-voice-relay.test.sh:1885-1886 untouched because `quitting` is only ever set from close().
- ℹ️ `docs/voice-relay.md:156` - docs/voice-relay.md:156 - the sentence added by this change reads as an exhaustive account of `relay_error` and promises a distinction two of the reasons the code emits do not make. It enumerates four subjects (the connection ended, or was lost, or the relay stopped, or the session ended) and states that each says either &#34;before that turn was answered&#34; or &#34;before the reply finished&#34;. Two other reason strings reach the same field and match neither half. The relay&#39;s own words, applied verbatim at bin/fm-voice-client.py:909 from `fail_turn` (bin/fm-voice-relay.py:886), carry no clause at all, so a truncated turn can read `answered: true` with `relay_error: &#34;ModelStreamErrorException: ...&#34;`; that is deliberate and correct, since the relay knows more than this end can infer, but the doc does not say it. The handling-fault reason at bin/fm-voice-client.py:969-970, &#34;this end could not handle the relay&#39;s reply: &lt;type&gt;: &lt;message&gt;&#34;, was introduced by this branch in round 3 and is documented nowhere; it is also the one reason that is not about the far end, which is the distinction its own comment at 967-968 calls out as the thing a reader must not get wrong. A captain who reads line 156 as the set and then meets either string has no way to place it. Since line 159 makes any `relay_error` non-zero the exit code, the enumeration is also what tells them which failures fail a run. Raised as ask-user because it is the author&#39;s call whether to name the two extra shapes or to reword line 156 so it stops reading as a complete list.

🔧 Fix: read one turn&#39;s audio count once, and hush a tidy exit
3 infos still open:
- ℹ️ `bin/fm-voice-client.py:1066` - bin/fm-voice-client.py:1065-1066 - this round hoisted `turn_bytes` into `reply_bytes` so the record stops asking one question twice, but `played` (1065) and `reply_bytes` (1066) are still two separate lock-free reads of state `FilePlayback.write` updates together inside one critical section. Lines 257-262 hold `self._lock` across `mine = turn == self._turn`, `self.first_played = time.monotonic()` and `self.turn_bytes += len(pcm)`, and the record reader takes no playback lock, so it can observe the instant between the stamp at 260 and the credit at 262. Concrete sequence, on the one path that can reach the record with a frame in flight: a turn whose `reply_done.wait` expires at 1058 with nothing played, so `_wait_audio_quiet` reads `last_frame` as None and returns immediately (1141-1142) with no idle wait; `arrived_in == self.turn_id` still holds because take_turn has not advanced; the turn thread copies the dict at 1063 and is descheduled; the downlink stamps `first_frame`/`last_frame` (859-860), releases, enters `write`, stamps `first_played` at 260, and is preempted before 262; the turn thread then runs 1065-1066. The record reads `answered: false`, `reply_audio_seconds: 0.0`, `first_frame_s: null` - and `first_played_s` plus `first_audio_s` as real numbers, so it reports the headline latency for audio it also says never arrived. That is the same one-record-two-answers shape this round removed, in the direction the round-9 instruction did not consider: it explicitly reasoned about `played` and `first_frame` being None together and accepted that null, but not about `played` being set while the count is still zero. The opposite read order is worse (any interleave after the stamp loses), so reordering does not help. Reachability is narrow and I am reporting it as such: the downlink must cover 856 through 260 inside two statements of the turn thread and then stall on one bytecode. Flagged because it is the residual of the invariant this commit claims, the FilePlayback path is the one every published figure in docs/voice-relay.md was measured with, and nothing pins the implication that a stamped `first_played_s` means the turn was answered (tests/fm-voice-relay.test.sh:2751 pins the null case, 4327 the happy path). Earliest shared boundary: one accessor on the playback returning `first_played` and `turn_bytes` together under `_lock`, which both classes already have. That is inside the playback rather than in `take_turn`, so it does not contradict the standing &#34;do not take the playback lock here&#34; ruling; the speaker path needs it too, where `write` credits bytes and `_callback` stamps the clock on different threads. Raised as ask-user because closing it touches how a record is assembled, which is the area you have fenced off for the single-outcome-computation work.
- ℹ️ `tests/fm-voice-relay.test.sh:2539` - tests/fm-voice-relay.test.sh:2539 - `QuitThenEnd` is `ScriptedStream` (1779) with a gate wait bolted on, and `ScriptedStream` grew the `at_end` callable one commit earlier (bdc7bef) for exactly that shape. `ScriptedStream(tidy_open, AUDIO + REPLY_END, &#34;the tidy exit&#34;, at_end=lambda: check(tidy_quit.wait(10), ...))` is behaviourally identical: `read` waits the open gate, slices, and once `_at &gt;= len(_bytes)` calls `at_end` and returns the empty slice, which is the `return b&#34;&#34;` this class writes out by hand. `a_reset` at 2015-2021 is the existing precedent for an `at_end` that blocks or raises. `RaisingAfterReply` (2056) is the same substitution with `at_end` raising. `FailingAfterReply` (2076) and `LateAudioStream` (2279) share the same body and differ only in serving a second payload after the gate, which a `then` argument would cover. So four classes of roughly fifteen lines each carry one shape that the shared fixture already expresses, and the newest of them was added one commit after the parameter that retires it. Scoped the same way you scoped the two dedups you accepted: this touches none of EndingStream, CountingUplink, ClosingStream or StartGate, whose provenance from the verbatim cherry-pick is why they are left alone, and every site named here was authored by this run&#39;s own fix rounds. `GoodbyeWatchingGate` (2559) is excluded, since the uplink it would otherwise extend is StartGate.
- ℹ️ `bin/fm-voice-client.py:841` - bin/fm-voice-client.py:841 - gating the end-of-stream line on `self.quitting` is the right call for the alarm-on-success rule, and the mid-session case stays pinned (tests/fm-voice-relay.test.sh:1884). What it also removes is the only signal that the relay had to be killed rather than shutting down. The BYE branch beside it still speaks during teardown, &#34;client: the relay signed off&#34; (950), so a relay that exited cleanly says a line while one whose teardown outran both bounds - `down_thread.join(timeout=5)` at 672 and `proc.wait(timeout=10)` at 684, against the relay&#39;s own `session.close()` awaiting the model reader at bin/fm-voice-relay.py:600-604 - says nothing at all. `proc.kill()` at 686 is silent too, and `closed_because` is only read by `_say_stopped`, which `run()` has already finished with, so the record and the exit code carry nothing either. The captain therefore cannot tell a tidy exit from a SIGKILLed one, which is the operational signal the timing asymmetry you have filed under teardown ordering would actually surface. A `log(self.verbose, &#34;the relay was still running when the connection ended&#34;)` on the quitting side keeps the distinction without an alarm, and is compatible with the ruling against downgrading the line, which was about not losing the mid-session fault. Raised as ask-user rather than auto-fix because it adds a line of human output on a path you deliberately just silenced, and the silence may be what you want.

</details>

<details>

✅ No issues found.
- <code>`bash tests/fm-voice-relay.test.sh` — the suite that owns the spoken interface (all cases pass, ~35s), including the cut-header turn, the connection lost between turns, the connection already known gone when a turn opens, the truncated answer at `--runs 1`, the relay naming a desync ahead of its goodbye, and the whole round trip</code>
- <code>`bash harness/run_scenarios.sh` — drives the real `bin/fm-voice-client.py` CLI (`--local --runs N &gt; runs.jsonl`) against a stub relay speaking the real wire format, once with the default-branch client and once with the fix, over four scenarios: relay dies before any reply, relay dies partway through the answer, relay dies between two runs, healthy two-run control</code>
- <code>`bash harness/run_real_relay.sh &lt;repo&gt;` — same before/after with the REAL `bin/fm-voice-relay.py` and its real records reader (only the paid Bedrock model stood in for, using the suite&#39;s own extracted stand-in): the model session ends mid key press, run 1 unanswered and run 2 answered over the same connection</code>
- <code>`python3 harness/relay_desync_frames.py &lt;relay&gt; &lt;label&gt;` — opens the real relay `--serve`, starts a turn, writes a byte that is not a frame kind, and prints the frames it sends back; run against both the base and the fixed relay</code>
- <code>`git check-ignore -v data/speech-to-speech-survey-s2/report.md` and `git grep -n &#39;speech-to-speech-survey&#39;` / `git grep -n &#39;the survey&#39;` — the cited record is gitignored and no tracked file cites it any more</code>
- <code>`git diff dc0172c..dfe12ac -- .greptile/rules.md` and `git ls-files .greptile` — the rule file is kept and narrowed by one sentence rather than removed</code>

</details>

<details>

- ℹ️ `.greptile/rules.md:10` - The narrowing sentence this change added restates a fact the same file already carries at lines 35-36 (&#34;Whether `VISION.md` should be reconciled to describe that is the captain&#39;s call and is not settled by any single pull request&#34;), so the repo&#39;s one-owner rule is now stated twice in a 38-line file. I did not consolidate it: the intent pinned the new sentence&#39;s three clauses verbatim, and the older sentence carries extra content (&#34;not settled by any single pull request&#34;) that the sentence immediately after it depends on (&#34;Raising it as new is what this rule is here to stop&#34;), so trimming it risks breaking the reviewer&#39;s recorded argument. Worth a follow-up decision on which of the two should own the open question.

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

### DIFF
--- a/.greptile/rules.md
+++ b/.greptile/rules.md
@@ -7,6 +7,8 @@ dropped stops being a decision and becomes a blind spot.
 
 ## A captain-facing surface is not automatically a `VISION.md` violation
 
+This is the first mate's working interpretation rather than settled repository policy; whether `VISION.md` itself should be reconciled remains an open question belonging to the captain; and the conditions listed below are what this interpretation depends on.
+
 `VISION.md` says "The captain talks to the first mate and to nobody else; every
 worker reports through the first mate and never addresses the captain directly."
 That line protects who is answerable for work. Read alongside the sentence it
--- a/bin/fm-voice-client.py
+++ b/bin/fm-voice-client.py
@@ -211,6 +211,15 @@ class FilePlayback:
         # own turn lock, so nothing slow may ever be done under it. _handle_lock
         # covers the file itself, so a write and a close cannot overlap. The
         # ordering is always _handle_lock then _lock and never the reverse.
+        #
+        # What close() needing _handle_lock costs: the exit is now only as bounded
+        # as one write to --out-file, so on a hung or full filesystem the five
+        # second downlink join in Client.close no longer bounds it. The wedged relay
+        # that join was written for is unaffected, being another process while this
+        # write is local. That cost belongs to the filed teardown-ordering work,
+        # whose other half is the same five second join being shorter than the ten
+        # seconds the relay may spend draining its own reply stream, which is why
+        # audio can arrive after the output is released at all.
         self._handle_lock = threading.Lock()
         self._lock = threading.Lock()
         self.first_played = None
@@ -299,8 +308,14 @@ class SpeakerPlayback:
         self._lock = threading.Lock()
         self.first_played = None
         self.turn_bytes = 0
+        # The same diagnostic the file path keeps, for the same reason. A counter
+        # that can only ever read zero is indistinguishable from one that measured
+        # zero, and this is the path the captain will actually use, so the write
+        # after close that the counter exists to catch has to be visible here too.
+        self.discarded = 0
         self._turn = None
         self._earlier = 0
+        self._closed = False
         self._stream = sounddevice.RawOutputStream(
             samplerate=OUT_RATE, channels=1, dtype="int16",
             blocksize=OUT_BLOCK, device=device, latency="low",
@@ -325,6 +340,17 @@ class SpeakerPlayback:
 
     def write(self, pcm, turn):
         with self._lock:
+            # close() stops the stream, and after that no callback drains the
+            # buffer, so a chunk arriving here was never going to be heard however
+            # it is stored. Discarded and counted rather than queued and credited
+            # to the turn, which is what the file path does: queued, it is a
+            # measurement the captain never heard, and silent, the write after
+            # close outside teardown that this counts for would be invisible on the
+            # one path they use. Counted and nothing more, so it stamps no clock
+            # and reaches neither answered, first_audio_s nor the exit code.
+            if self._closed:
+                self.discarded += 1
+                return
             if turn == self._turn:
                 self.turn_bytes += len(pcm)
             else:
@@ -352,6 +378,13 @@ class SpeakerPlayback:
         time.sleep(0.2)
 
     def close(self):
+        # Marked before the stream is stopped and not while the lock is held: the
+        # device callback takes this lock, and stop() waits for a callback already
+        # running, so holding it across the stop is a deadlock. Marking first
+        # instead leaves no instant where the stream is gone and a write still
+        # queues for it.
+        with self._lock:
+            self._closed = True
         try:
             self._stream.stop()
             self._stream.close()
@@ -660,9 +693,17 @@ class Client:
 
         Expected while this end is shutting down and said nowhere else, so a count
         appearing on a session that has not ended is the logic bug it is kept for.
-        Diagnostic only: it names nothing in the record and decides no exit code.
+        That holds on both output paths, because both count it. Diagnostic only: it
+        names nothing in the record and decides no exit code.
+
+        Read straight off the playback rather than through a default, so a playback
+        that cannot answer is a failure here rather than a zero indistinguishable
+        from having measured none. The None check is close()'s own, for the startup
+        that refused before there was an output at all.
         """
-        dropped = getattr(self.playback, "discarded", 0)
+        if self.playback is None:
+            return
+        dropped = self.playback.discarded
         if dropped:
             log(self.verbose,
                 "discarded {} reply audio chunk(s) that arrived after the output "
--- a/bin/fm-voice-client.py
+++ b/bin/fm-voice-client.py
@@ -838,7 +838,8 @@ class Client:
                     self.closed.set()
                 break
             if got is None:
-                say("client: the connection ended")
+                if not self.quitting.is_set():
+                    say("client: the connection ended")
                 # The subject only. Whether it ended before the turn was answered
                 # or partway through the answer is decided by _unfinished at the
                 # tail, where the audio count is read.
@@ -1062,6 +1063,7 @@ class Client:
             turn = dict(self.turn)
         marks = turn.get("marks", {})
         played = self.playback.first_played
+        reply_bytes = self.playback.turn_bytes
         first_frame = turn.get("first_frame")
 
         def since(at):
@@ -1103,8 +1105,8 @@ class Client:
             # cannot tell a reply from the previous reply's tail arriving late, and
             # counting that tail here reports a turn nobody answered as answered.
             "reply_audio_seconds": round(
-                self.playback.turn_bytes / float(OUT_RATE * 2), 3),
-            "answered": self.playback.turn_bytes > 0,
+                reply_bytes / float(OUT_RATE * 2), 3),
+            "answered": reply_bytes > 0,
         }
         if release is None:
             record["first_audio_note"] = (
--- a/bin/fm-voice-client.py
+++ b/bin/fm-voice-client.py
@@ -91,8 +91,8 @@ import fm_voice_frame as frame              # noqa: E402
 
 IN_RATE = 16000
 OUT_RATE = 24000
-# 100 ms at each rate. The uplink chunk matches what the relay and the survey
-# measured with; changing it changes the numbers.
+# 100 ms at each rate. The uplink chunk matches what the relay and the earlier
+# prototype work measured with; changing it changes the numbers.
 CHUNK = 3200
 OUT_BLOCK = 2400
 
--- a/bin/fm-voice-client.py
+++ b/bin/fm-voice-client.py
@@ -831,7 +831,9 @@ class Client:
                 with self.lock:
                     if not self.reply_done.is_set():
                         self.turn.setdefault(
-                            "failed", "the connection was lost: {}".format(exc))
+                            "failed", "{}: {}".format(
+                                self._unfinished("the connection was lost"),
+                                exc))
                     self.closed_because = "the connection was lost"
                     self.closed.set()
                 break
--- a/bin/fm-voice-client.py
+++ b/bin/fm-voice-client.py
@@ -84,6 +84,7 @@ import subprocess
 import sys
 import threading
 import time
+import traceback
 
 sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
 
@@ -186,22 +187,39 @@ class Uplink:
 
 
 class FilePlayback:
-    """Write reply audio to a file. This is the path that can be verified here."""
+    """Write reply audio to a file. This is the path that can be verified here.
+
+    Every chunk carries the turn it belongs to, and turn_reset names the turn
+    being measured. A chunk from a turn that has already been recorded is still
+    written, because it is the tail of an answer the captain is still listening
+    to, but it stamps no clock and is counted toward nobody: attributed to the
+    turn that happens to be open, it would hand that turn a first-audio figure
+    measured from somebody else's reply and report it answered when it was not.
+    """
 
     def __init__(self, path):
         self._handle = open(path, "wb")
         self.first_played = None
         self.device_latency = None
+        # Everything written, and what the turn being measured is owed. The
+        # difference is the audio of turns already recorded.
         self.bytes = 0
+        self.turn_bytes = 0
+        self._turn = None
 
-    def write(self, pcm):
-        if self.first_played is None:
+    def write(self, pcm, turn):
+        mine = turn == self._turn
+        if mine and self.first_played is None:
             self.first_played = time.monotonic()
         self._handle.write(pcm)
         self.bytes += len(pcm)
+        if mine:
+            self.turn_bytes += len(pcm)
 
-    def turn_reset(self):
+    def turn_reset(self, turn):
+        self._turn = turn
         self.first_played = None
+        self.turn_bytes = 0
 
     def drain(self, timeout=5):
         del timeout
@@ -219,6 +237,16 @@ class SpeakerPlayback:
     the last moment this process can see. The device's own output buffer sits
     after that, so its reported latency is included in the turn record rather
     than pretended away.
+
+    That timestamp is why the turn a chunk belongs to has to travel with the
+    chunk rather than being checked before the write: the moment that matters
+    happens in the callback, later than the frame arriving, and the gap between
+    the two is the honest content of the figure. So the buffer remembers how many
+    of its leading bytes belong to turns already recorded, and the first audio of
+    the turn being measured is the first byte past them. Ordering makes that a
+    count rather than a per-chunk tag: the downlink hands chunks over in arrival
+    order on one thread and a turn number never goes backwards, so a chunk from an
+    earlier turn can never queue behind one from a later turn.
     """
 
     def __init__(self, device=None):
@@ -227,6 +255,9 @@ class SpeakerPlayback:
         self._lock = threading.Lock()
         self.first_played = None
         self.bytes = 0
+        self.turn_bytes = 0
+        self._turn = None
+        self._earlier = 0
         self._stream = sounddevice.RawOutputStream(
             samplerate=OUT_RATE, channels=1, dtype="int16",
             blocksize=OUT_BLOCK, device=device, latency="low",
@@ -241,20 +272,32 @@ class SpeakerPlayback:
             take = min(want, len(self._buffer))
             chunk = bytes(self._buffer[:take])
             del self._buffer[:take]
-            if chunk and self.first_played is None:
+            spent = min(self._earlier, take)
+            self._earlier -= spent
+            if take > spent and self.first_played is None:
                 self.first_played = time.monotonic()
         outdata[:take] = chunk
         if take < want:
             outdata[take:want] = b"\x00" * (want - take)
 
-    def write(self, pcm):
+    def write(self, pcm, turn):
         with self._lock:
+            if turn == self._turn:
+                self.turn_bytes += len(pcm)
+            else:
+                self._earlier += len(pcm)
             self._buffer += pcm
         self.bytes += len(pcm)
 
-    def turn_reset(self):
+    def turn_reset(self, turn):
         with self._lock:
+            self._turn = turn
             self.first_played = None
+            self.turn_bytes = 0
+            # Whatever is still queued was spoken for an earlier turn. Counted as
+            # this turn's, the previous answer's undrained tail would stamp this
+            # turn's first audio the instant the device next asked for a block.
+            self._earlier = len(self._buffer)
 
     def drain(self, timeout=30):
         """Wait for the buffered reply to finish, so the process does not cut it off."""
@@ -415,11 +458,17 @@ class Client:
         self.uplink = None
         self.playback = None
         self.capture = None
+        self.down_thread = None
         self.up_q = queue.Queue()
         self.talking = threading.Event()
         self.ready = threading.Event()
         self.reply_done = threading.Event()
         self.closed = threading.Event()
+        # Set the moment this end asks the relay to stop. It is the only thing
+        # that tells an expected goodbye from the relay stopping on its own,
+        # because the frame is the same one either way, and reading a clean end as
+        # a fault would train the captain to ignore the line that means it.
+        self.quitting = threading.Event()
         self.ready_notice = {}
         self.turn = {}
         # Which turn self.turn is. A frame is read on one thread and applied on
@@ -432,10 +481,12 @@ class Client:
         # downlink takes a copy of this when a frame arrives and applies nothing
         # once it no longer matches.
         self.turn_id = 0
-        # What run() tells the captain when no further turn can be taken. The
-        # connection ending is the ordinary reason and stays the default; a fault
-        # on this end replaces it, because a line naming the wrong cause sends
-        # them looking where the fault is not.
+        # What run() tells the captain when no further turn can be taken. Every
+        # path that makes the connection unusable names itself here, so the line
+        # about the runs that were lost restates the cause that was recorded
+        # rather than asserting one; a line naming the wrong cause sends them
+        # looking where the fault is not. The default only covers a closure with
+        # no path at all behind it, which nothing here can currently produce.
         self.closed_because = "the connection closed"
         self.lock = threading.Lock()
 
@@ -485,7 +536,8 @@ class Client:
                 lambda: MicCapture(self.options.input_device))
         self.capture.start(self.up_q, self.talking)
 
-        threading.Thread(target=self._downlink, daemon=True).start()
+        self.down_thread = threading.Thread(target=self._downlink, daemon=True)
+        self.down_thread.start()
         threading.Thread(target=self._sender, daemon=True).start()
 
         self._wait_ready()
@@ -531,7 +583,19 @@ class Client:
         # fields are still None and the original refusal is the message worth
         # keeping.
         if self.uplink is not None:
+            # Before the frame, so the goodbye that answers it is read as the
+            # answer to a question this end asked rather than as the relay
+            # stopping on its own.
+            self.quitting.set()
             self._quietly("the uplink", lambda: self.uplink.send(frame.QUIT))
+        # Before the devices are released, and bounded so a wedged relay cannot
+        # hold the exit. The downlink is still handing reply audio to the speaker,
+        # and a speaker closed underneath it raises there, which the frame-handling
+        # guard would then report as a fault on the way out of a session that
+        # worked. Letting the goodbye above end that thread first costs nothing and
+        # keeps the fault line meaning a fault.
+        if self.down_thread is not None:
+            self.down_thread.join(timeout=5)
         if self.capture is not None:
             self._quietly("the microphone", self.capture.close)
         if self.playback is not None:
@@ -593,6 +657,10 @@ class Client:
         # different faults, so each names itself rather than leaving the tail to
         # guess or to say nothing.
         why = None
+        # And what run() says about the runs that were lost to it. Separate from
+        # the reason above because they are different statements: that one is why
+        # this turn has no answer, this one is why there will be no more turns.
+        cause = None
         while True:
             try:
                 got = self.reader.read()
@@ -612,11 +680,13 @@ class Client:
                 with self.lock:
                     self.turn.setdefault(
                         "failed", "the connection was lost: {}".format(exc))
+                    self.closed_because = "the connection was lost"
                     self.closed.set()
                 break
             if got is None:
                 say("client: the connection ended")
                 why = "the connection ended before this turn was answered"
+                cause = "the connection ended"
                 break
             kind, payload = got
             # Which turn this frame belongs to, taken the moment it arrives. Every
@@ -630,11 +700,15 @@ class Client:
                             now = time.monotonic()
                             self.turn.setdefault("first_frame", now)
                             self.turn["last_frame"] = now
-                    # Played whichever turn it belongs to. Late audio is the tail
-                    # of an answer the captain is still listening to, so dropping
-                    # it would cut them off; only the timing figures are a claim
-                    # about a particular turn.
-                    self.playback.write(payload)
+                    # Played whichever turn it belongs to, and told which that is.
+                    # Late audio is the tail of an answer the captain is still
+                    # listening to, so dropping it would cut them off, but three
+                    # figures are read off what this call does - first_played, the
+                    # reply's own duration and whether the turn was answered at all
+                    # - and a stale chunk credited to the turn now open reports an
+                    # unanswered turn as answered, which is an exit code of zero on
+                    # a session that lost one.
+                    self.playback.write(payload, arrived_in)
                 elif kind == frame.TEXT:
                     obj = frame.decode_json(payload)
                     text = (obj.get("text") or "").strip()
@@ -699,8 +773,20 @@ class Client:
                             if obj.get("mark") == "reply_end":
                                 self.reply_done.set()
                 elif kind == frame.BYE:
-                    say("client: the relay said goodbye")
-                    why = "the relay said goodbye before this turn was answered"
+                    # The same frame ends a session this end asked to end and a
+                    # relay that stopped on its own, so the frame says nothing on
+                    # its own and whether we asked is the whole discriminator.
+                    # Both speak, because a session that ended should say so, and
+                    # neither borrows the other's words: a line that also appears
+                    # when everything worked is a line the captain learns to skip,
+                    # and then the one that means trouble is invisible too.
+                    if self.quitting.is_set():
+                        say("client: the relay signed off")
+                        cause = "the relay signed off after being asked to stop"
+                    else:
+                        say("client: the relay stopped without being asked to")
+                        why = "the relay stopped before this turn was answered"
+                        cause = "the relay stopped without being asked to"
                     break
             except Exception as exc:               # noqa: BLE001
                 # A fault on THIS end, handling a reply that did arrive: the
@@ -718,6 +804,15 @@ class Client:
                 fault = ("this end could not handle the relay's reply: {}: {}"
                          .format(type(exc).__name__, exc))
                 say("client: {}".format(fault))
+                # The one line is for the captain and the record; the traceback is
+                # for whoever has to find the bug behind it. Before this guard
+                # existed the thread died and threading.excepthook printed one, so
+                # a programming error in here would otherwise be strictly harder to
+                # locate than it used to be. Terminal path, so this prints once per
+                # session at worst, and the record keeps the one-line reason
+                # because that field is machine read.
+                sys.stderr.write(traceback.format_exc())
+                sys.stderr.flush()
                 with self.lock:
                     self.turn.setdefault("failed", fault)
                     self.closed_because = fault
@@ -740,6 +835,8 @@ class Client:
         with self.lock:
             if why is not None and not self.reply_done.is_set():
                 self.turn.setdefault("failed", why)
+            if cause is not None:
+                self.closed_because = cause
             self.closed.set()
         self.reply_done.set()
 
@@ -773,8 +870,10 @@ class Client:
             # event as nobody waiting and named nothing, and this turn then waits
             # out its whole timeout to be recorded with no reason at all.
             self.reply_done.clear()
-        self.playback.turn_reset()
-        before = self.playback.bytes
+            # In the same critical section, and named with the same identity the
+            # frames carry, so there is no instant where the turn has advanced and
+            # the playback is still counting audio toward the turn before it.
+            self.playback.turn_reset(self.turn_id)
         self.up_q.put(START)
 
         # Unreachable while parse_args refuses open-mic, and kept so that turning
@@ -834,9 +933,13 @@ class Client:
             "device_output_latency_s": self.playback.device_latency,
             "device_input_latency_s": self.capture.device_latency,
             "relay_marks_since_talk_end": marks,
+            # This turn's own audio, counted by the playback rather than by
+            # subtracting a byte total it shares with every other turn. A total
+            # cannot tell a reply from the previous reply's tail arriving late, and
+            # counting that tail here reports a turn nobody answered as answered.
             "reply_audio_seconds": round(
-                (self.playback.bytes - before) / float(OUT_RATE * 2), 3),
-            "answered": self.playback.bytes > before,
+                self.playback.turn_bytes / float(OUT_RATE * 2), 3),
+            "answered": self.playback.turn_bytes > 0,
         }
         if release is None:
             record["first_audio_note"] = (
@@ -946,6 +1049,18 @@ class Client:
                 "waiting {:.2f}s for the answer to finish".format(remaining))
             time.sleep(remaining)
 
+    def _say_stopped(self, index):
+        """Name why no more turns can be taken, and which run was the first lost.
+
+        The cause is whatever the path that closed the connection recorded, not an
+        assertion made here: a fault on this end leaves the connection open, and a
+        line blaming the connection for it sends the captain to the wrong end.
+        """
+        with self.lock:
+            because = self.closed_because
+        say("client: {} before run {} of {}; it and the rest were not "
+            "taken".format(because, index, self.options.runs))
+
     def run(self):
         rc = 0
         for index in range(1, self.options.runs + 1):
@@ -961,13 +1076,7 @@ class Client:
             # a turn that never opened, since an invented turn is the whole thing
             # being kept out of runs.jsonl.
             if record is None:
-                # What actually stopped it, not an assumption that the connection
-                # went: a fault on this end leaves the connection open, and a line
-                # blaming it would send the captain to the wrong end.
-                with self.lock:
-                    because = self.closed_because
-                say("client: {} before run {} of {}; it and the rest were not "
-                    "taken".format(because, index, self.options.runs))
+                self._say_stopped(index)
                 rc = 1
                 break
             print(json.dumps(record))
@@ -975,6 +1084,17 @@ class Client:
             if not record["answered"]:
                 rc = 1
             if index < self.options.runs:
+                # Checked after the record and before the wait, because that wait
+                # is seconds long and exists only to avoid interrupting the model's
+                # own speech, which a relay that is already gone cannot be doing.
+                # Waiting it out here left the captain sitting through the last
+                # reply's whole spoken duration before being told the session had
+                # stopped. The exit code is still the unhappy one: the runs asked
+                # for were not taken, whatever the last one reported.
+                if self.closed.is_set():
+                    self._say_stopped(index + 1)
+                    rc = 1
+                    break
                 self._let_reply_finish(record)
         return rc
 
--- a/bin/fm-voice-client.py
+++ b/bin/fm-voice-client.py
@@ -583,9 +583,16 @@ class Client:
                 # connection that only says answered: false is indistinguishable
                 # there from a turn the model declined to answer. setdefault
                 # because a relay that named the failure first said it better.
+                #
+                # closed is set in this same critical section, not left to the
+                # tail below, because take_turn decides whether another turn can
+                # be opened by reading it under this lock. Naming the failure
+                # first and announcing the closure afterwards left a window where
+                # the connection was known gone and no reader could tell.
                 with self.lock:
                     self.turn.setdefault(
                         "failed", "the connection was lost: {}".format(exc))
+                    self.closed.set()
                 break
             if got is None:
                 break
@@ -641,14 +648,34 @@ class Client:
                     self.reply_done.set()
             elif kind == frame.BYE:
                 break
-        self.closed.set()
+        # Under the turn lock for the same reason the failure above is: a clean
+        # end of file and a goodbye leave the connection just as unusable as a
+        # dropped one, and take_turn reads this under that lock to decide whether
+        # a turn can still be opened. Already set on the failure path; setting an
+        # event twice costs nothing.
+        with self.lock:
+            self.closed.set()
         self.reply_done.set()
 
     # ---------------------------------------------------------------------- turns
 
     def take_turn(self, index):
-        """Run one turn and return its record."""
+        """Run one turn and return its record, or None if the connection is gone.
+
+        The check and the reset share one critical section with the downlink's
+        closure mark on purpose. The wait between turns is seconds long and is
+        where a relay that dies between questions dies, so the run loop cannot
+        decide to open another turn by reading a flag the downlink sets after it
+        records the failure: between those two writes the connection is already
+        gone and the loop cannot see it. It then cleared the failure the downlink
+        had recorded, sent talk-start into a dead pipe, and came back after the
+        whole reply timeout as answered: false with relay_error: null - a lost
+        connection wearing the shape of a turn the model declined, in the file
+        docs/voice-relay.md computes its published latency spread from.
+        """
         with self.lock:
+            if self.closed.is_set():
+                return None
             self.turn = {}
         self.playback.turn_reset()
         self.reply_done.clear()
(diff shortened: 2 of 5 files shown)


kunchenguid:
Speaking as Kun's firstmate:

Corrective follow-up to merged #2767. A lost relay connection is no longer recorded as an unanswered turn with `relay_error` null. Aligns (honest interface; infrastructure failure must not look like a silent skip). Still opt-in voice surface.

CI green including no-mistakes. Landing this.

kunchenguid (approved):
Speaking as Kun's firstmate: corrective lost-connection record, green CI. Approving to land.

kunchenguid:
Speaking as Kun's firstmate:

Corrective for a lost relay connection recorded as an unanswered turn (`answered: false`, `relay_error: null`) that could be averaged into the published latency spread. `take_turn` now refuses to open on a connection already marked closed; every close path names a reason; cut-short replies stay `answered: true` with `before the reply finished`; a record with `relay_error` or a short run exits nonzero. Private survey citation is gone. `.greptile/rules.md` is narrowed as working interpretation, not a VISION change.

VISION (per rule):
- One captain, one interface — aligns. Infrastructure failure is no longer disguised as a declined turn.
- Authority is explicit — aligns. No new grant on the already-shipped voice path.
- Scripts own the mechanics — aligns. Closure, per-turn accounting, and exit codes are deterministic.
- A restart is a non-event — aligns. Obligations close with a named durable record, not silence.
- Delegation with a spine — aligns. Independent tests pin the forbidden `runs.jsonl` shape.
- The fleet outlives any vendor — aligns. Wire/accounting, not a vendor UI.
- Scope — aligns. Opt-in spoken interface already on the default branch; this is a honesty fix, not a new product.

Security: no.

Overlap: voice client/relay/tests and `.greptile/rules.md` are unique on the open set I searched. No collision with the teardown/treehouse hold pair or the spawn-freshen hold.

CI: all required checks green on this HEAD, including **Require no-mistakes** (attestation `head_sha` matches `1b2ce4950d2b`). Greptile 5/5. Ahead 12, behind 1, mergeable, mergeStateStatus CLEAN.

Waiting on the coordinator to land — not waiting on the author, not waiting on the captain.

Merge-eligible: YES. Captain-flag NOW: NO.

kunchenguid:
Speaking as Kun's firstmate: this is merged. Thank you @Inthuson — really appreciate you taking the time on this.

PR #2761, opened 2026-08-21
fix: verify payload before confirming steer delivery

kunchenguid:
## Intent

Fix the fm-send delivery-confirmation unsoundness: a steer can be reported delivered (exit 0) when the pane actually swallowed the whole payload. The diagnosis is that this is not truncation and not a remote-transport defect. A steer into a freshly created home parked on Claude Code's workspace-trust dialog was lost whole: fm-send typed into the modal (which discards printable input), Enter confirmed the dialog, and the idle-to-busy check falsely reported delivery. The apparent ~110-character truncation was fm_pending_reply_summarize's 117-character excerpt used by recovery resend.

The captain authorized the fix-the-cause approach (report Option B plus the durability half of A), not a file-based rewrite of the steer channel.

Implement three fixes:
1. Primary: bind submit confirmation to the text being sent. Add a pre-Enter identity-bound tail-anchor read-back that verifies the sent text is actually present in the composer BEFORE Enter is treated as submission, and refuse plus fail loud (non-zero, a clear diagnostic naming the pane state, never a silent partial or false success) when it is not. This hardens herdr, tmux, and local sends, not just remote. Keep the composer-shape / read-back policy in the one owner bin/fm-composer-lib.sh; do not fork per-backend copies.
2. Store the full pending-reply request payload durably so recovery resends the complete text, not a ~117-character excerpt. Keep the excerpt for display if useful, but recovery must be lossless.
3. Assert Claude launch-readiness so a send/spawn to a home still parked on the trust dialog / not yet ready is detected and refused loudly rather than falsely confirmed. The read-back already catches the swallow; this closes the window earlier and names it.

Tests must turn the report's reproduction into regressions: a steer to a pane parked on the trust dialog must now be refused and fail loud, and the stored payload must be recoverable whole. Follow harness-dependent checks: a portable regression driving the read-back/verdict logic with no live harness, plus the env-gated live-harness guard. Colocate in tests/; do not weaken assertions. Drive all herdr lifecycle through the guarded herdr lab.

Touch only what the fix needs: bin/fm-send.sh, bin/fm-composer-lib.sh, bin/fm-pending-reply-lib.sh, the herdr/tmux backends, and tests. Verify every affected primary harness and runtime backend. Do not merge. Note in the PR that homes pick this up after merge plus a self-update and that landing timing is coordinated with the main firstmate.

## What Changed

- Adds centralized pre-Enter composer read-back across supported backends, refusing gated panes or missing payload tails without sending Enter.
[...]

PR #2814, opened 2026-08-22
fix(bin): prove steer delivery with payload deltas

kunchenguid:
## Intent

Re-implement fm-send steer delivery verification using the captain-APPROVED payload-delta proof, replacing the shape-parsing read-back that was abandoned. This is firstmate's SHARED, TRACKED transport core - the steer channel the whole fleet relies on.

APPROVED DESIGN (captain decision 1): implement the delivery proof in the ONE owner bin/fm-composer-lib.sh. Normalize both captures and the payload identically (strip ANSI/box-drawing/whitespace, keep row structure), anchor on the last min(24,len) normalized chars of the payload TAIL, and accept ONLY when the anchor is both (A) newly ADDED in the before->after row diff AND (B) its whole-screen occurrence COUNT increased. This excludes the round-1 scrollback false-pass by TIME not location, so that false-pass is impossible by construction. Keep it refusing rather than proceeding (non-zero + clear diagnostic, never a silent partial or false success) and backend-independent.

CAPTAIN DECISION 2 (fresh PR): branch off main; carry over ONLY the cleanly-separable KEEP set from the now-CLOSED branch fm/remote-send-confirm-readback-fix-r1 - the pending-reply durability half (full-payload durable recovery), the launch-readiness gate (claude_wait_for_ready in bin/fm-spawn.sh, fm_composer_screen_is_gated, fm_composer_pre_type_ok's GATED branch only), the locking, and the behavior-through-fm-send tests. Do NOT carry the shape-parsing read-back. PR #2761 is closed as superseded and must not be reopened.

CAPTAIN DECISION 3 (no bypass): do NOT add an FM_SEND_UNPROVEN escape hatch or any equivalent caller-declared exemption. A harness whose composer cannot be proven is refused, not steered blind. The existing exit-3 "unconfirmed" already covers the post-Enter case.

REPLACE (not keep): fm_composer_pre_enter_verdict was replaced wholesale by the delta verdict - its caps/cursor-row/identity params, the styled=1 vs styled=0 fork, the fm_composer_classify_screen precondition, and both fm_composer_extract_selected_content calls are gone. The caps/cursor_row/identity fetches were stripped from the send path in bin/fm-tmux-lib.sh fm_tmux_submit_core and from every per-backend copy (including zellij's private acceptance rule) so every backend uses the ONE shared verdict. The whole visible pane is captured identically before and after, replacing FM_COMPOSER_ACCEPT_LINES cropping.

CAPTAIN'S LATER OPTION C (accepted, and the substance of this branch's redesign): an earlier round of this work refused every payload shorter than the 24-character anchor minimum, which broke real fleet traffic - a decision answer like "2", a "yes", and every fm-control lifecycle exit command (/exit, /quit). The captain rejected both ways out that were tried. Short steers must WORK, not be refused; and there must be NO weaker-proof path for them, because a weaker test reopens the false-accept this whole design exists to close. The captain named three candidate approaches - a verifiable envelope/marker wrapping the payload, a confirmation handshake, or an occurrence count robust to coincidental modal text - and asked for the SOUNDEST of them.

CHOSEN APPROACH AND WHY THE ALTERNATIVES WERE REJECTED. A previous attempt on this branch tried "occurrence count plus a positively classified empty composer before the write, and a count rising by exactly one". That was analyzed and found UNSOUND and has been removed: a live agent pane redraws rows containing a single character constantly (a token counter ticking from 1.9k to 2.1k is a newly added row, containing "2", with the count risen by exactly one), so an empty-composer precondition does not exclude the coincidence. A confirmation handshake was rejected as not generically available across harnesses. The envelope was chosen: for a payload shorter than the anchor minimum, fm-send now MANUFACTURES the missing entropy. It types the payload followed by a fresh random suffix sized so the combined string is exactly the anchor length, proves that combined anchor through the same unchanged delta verdict, then erases exactly the suffix and proves the erase. The agent still receives precisely the payload the caller wrote, and because the proven anchor contains the payload's own characters, this is a proof ABOUT the payload rather than a proxy for it. The suffix goes AFTER the payload because erasing consumes from the tail; a leading suffix would need cursor movement no backend agrees on.

THE ERASE PROOF, and a hole found and closed during implementation. The first erase check compared the full combined anchor's count back to its before value. A test drove the hardest under-erase case and showed that check is wrong: dropping even one character from the tail destroys the full-length anchor, so its absence proves nothing and a composer still holding suffix characters would have been accepted, sending noise to the agent. The check was replaced with a boundary probe. Erasing consumes from the tail, so an incomplete erase can only ever leave payload+suffix-prefix, and every such leftover down to a single character still has the suffix's FIRST character immediately after the payload - so the test is whether that two-character boundary string gained an occurrence. The second check catches the opposite direction: an over-erase eats the payload's tail, so the payload's own occurrence count must be unchanged from the just-typed capture. The suffix alphabet deliberately EXCLUDES every character the payload contains, which is what makes that second count decidable - the payload cannot occur inside the suffix or across the boundary, so removing the suffix cannot move the count on its own.

FINDING 2 FROM THE PRIOR REVIEW ROUND (typed-unproven had no operator resolution path) is fixed: bin/fm-send.sh gained --typed-confirm and --typed-abandon, which settle a typed-unproven pending-reply record by correlation id, the diagnostic that creates that state now names both commands with the concrete id, and the watcher escalates a typed_unproven record on the ordinary cadence so it can never silently block teardown. A stranded suffix additionally gets a named stderr warning carrying the exact text and suffix length, so "inspect the pane" is an actionable instruction.

fm-control lifecycle exit now works with NO exemption. The earlier attempt added an FM_COMPOSER_DELIVERY_POSTCONDITION_EXTERNAL environment variable letting fm-control declare its own postcondition; that has been deleted as exactly the bypass shape decision 3 forbids. A short exit command needs no exemption once it carries a full-length anchor. The agent-state wait remains the authoritative post-exit proof, as it always was.

ARCHITECTURE: the type-and-prove sequence moved into one shared owner, fm_composer_typed_delivery_core, which every adapter calls with three primitives (capture, literal, erase), replacing five copies of the ordering - the same pattern fm_composer_submit_retry_core already uses.

BACKEND ERASE KEY: added per backend, because the key name belongs to each tool's own vocabulary and the byte it resolves to belongs to the pane. tmux BSpace and zellij Backspace were both verified live against real panes to remove exactly one character per press (zellij rejects tmux's BSpace spelling outright). herdr and cmux carry the lowercase backspace mapping consistent with their own vocabularies but are NOT live-verified on this machine, and the evidence table says so explicitly rather than implying otherwise. Orca's send API exposes only text, Enter, and interrupt, so a character cannot be taken back off its composer at all - the shared core therefore refuses a short payload on Orca BEFORE typing anything, rather than stranding a suffix it has no way to remove. Sending a raw backspace byte through the literal-text channel instead of a named key was measured and rejected: 0x08 erased nothing against a real tmux pane and 0x7f did, which is precisely the convention a named key delegates to the backend.

TESTS: the design's benches are regressions in existing tests/ files, with no new runner. tests/fm-composer-lib.test.sh covers the delta proof accepting a shape-free echo, refusing a gated trust dialog, refusing a swallowed send, and the round-1 scrollback copy yielding no new occurrence; plus the envelope's freshness, suffix-alphabet exclusion, exact erase count, verbatim treatment of a self-proving payload, refusal of a short multi-row payload, and both erase-proof directions. A pane simulator drives fm_composer_typed_delivery_core through its real interface for the end-to-end cases including "a short steer is proven and leaves ONLY the payload in the composer". Per firstmate-coding-guidelines' harness-dependent-checks rule, each backend's erase key also has a live guard in its real smoke test that fails naming the backend and version, and docs/verification/runtime-backends.md records the dated results and the commands that refresh them. The behavior-through-fm-send tests from the KEEP set are retained.

Also in this change: the composer-delta fixtures README landed unclassified earlier on this branch and failed bin/fm-doc-audience-check.sh, so it is classified as maintainer-verification; and two probe-absent sentinel assignments are quoted because shellcheck reads them as arithmetic once the shared library grows past its analysis limit. Both were pre-existing branch defects that lint and the doc check caught.

DELIVERY: mode=no-mistakes, yolo OFF - the captain owns the merge, so do not merge. This changes shared steer/transport core the whole fleet relies on; the PR should note that homes pick it up after merge plus a self-update, and that landing timing is coordinated with the main firstmate. Escalate any ask-user finding rather than deciding it; no --yes.

ADDENDUM - what happened in the previous run and what is already settled. A first validation run on this branch reached the lint step with review, test, and document all COMPLETED and review reporting ZERO findings, after thirteen rounds of review fixes that are all present as commits on this branch. Its lint fixer then wedged - the agent process sat at 0% CPU for seventeen minutes and emitted no output at all - so that worker was terminated by exact pid and the run went terminal with "step lint failed". Custody was recovered from the local gate: the pipeline had rebased onto a newer main, and the two original commits were verified patch-identical through that rebase before the branch was moved, with the pre-recovery head anchored first. Nothing was lost.

The lint findings that run never got to fix have since been fixed directly and committed, since fixing them is the documented action after a failed run: the hyphenated string sentinels shellcheck misreads as arithmetic are now quoted, the three test stubs that shadow indirectly-invoked functions carry narrow per-site directives with reasons rather than a file-wide disable, and a dead SIM_WIRE variable was removed rather than silenced because the assertions read the wire from its file instead. bin/fm-lint.sh now exits clean, bin/fm-doc-audience-check.sh is clean, and tests/fm-composer-lib.test.sh and tests/fm-pending-reply.test.sh both pass with 34 and 40 checks and no failures.

Decisions already made and NOT open for relitigation, each ruled on explicitly by the captain during the previous run: after a successful write, EVERY unproven outcome is typed-unproven and there is no retryable post-write shape, because no screen delta can distinguish a delivered-but-unrendered write from a swallowed one; only the pre-write refusals stay retryable, being a gated modal refused before typing and a write that did not go out. The flattened row matching in the delta proof is deliberate and load-bearing because it is what makes a WRAPPED payload provable. Shape-parsing must not be reopened in any form. The low-level stranded-suffix warning stays purely descriptive and the request-owning layer owns the settlement instruction, because the shared transport core has no concept of a pending request and the lifecycle commands call it with no correlation id. The erase proof's payload-retention barrier compares for EQUALITY against the stage-one capture. And a bounded, mechanism-less residual is knowingly accepted and documented: there is no shape-free positive proof that the payload survived the final erase keystroke, because once the erase is correct nothing high-entropy remains to anchor on.

## What Changed

- Replace shape-parsed composer read-back with one shared, backend-independent payload-delta proof that gates Enter and rejects gated or unprovable panes.
- Prove short steers using a randomized erasable suffix, with backend-specific erase keys, erase verification, and pre-write refusal where erasure is unsupported.
- Persist complete pending-request payloads for durable recovery and add locked typed-unproven tracking, escalation, and explicit `--typed-confirm` / `--typed-abandon` settlement commands.

## Risk Assessment

⚠️ Medium: No material source defects were found, but this is a substantial shared transport rewrite with an explicitly accepted erase-proof residual on backends not yet live-verified.

## Testing

Inspected the target diff, ran focused shared-core, fm-send, lifecycle, pending-reply, daemon, remote-delivery, and backend adapter tests, then manually exercised a short steer through the real fm-send executable; all checks passed, and the transcript confirms exit 0, exact suffix erasure, Enter submission, and payload-only delivery. No screenshot was applicable because the changed user surface is terminal transport rather than graphical UI.

<details>
<summary>Evidence: Short-steer fm-send end-to-end transcript</summary>

Source: [Short-steer fm-send end-to-end transcript](https://github.com/kunchenguid/firstmate/blob/fa52fdd51fce531e72acbf0b54e5a5904f4deb28/.no-mistakes/evidence/fm/remote-send-delivery-delta-r1/short-steer-e2e-transcript.txt)

```text
$ fm-send.sh demo "2"
exit=0
--- stderr (end-user surface) ---
●  1 task(s) in flight, but no watcher has a fresh beacon (last beat: never, grace 300s).
●  This is a supervision warning only; the requested message WILL still be sent.
●  repair a missing or failed watcher cycle with the Pi tool fm_watch_arm_pi, or restart Pi with -e /tmp/fm-send-evidence.BN3qID/home/.pi/extensions/fm-primary-turnend-guard.ts -e /tmp/fm-send-evidence.BN3qID/home/.pi/extensions/fm-primary-pi-watch.ts if the extensions are not loaded.
--- terminal transport ---
TYPE literal payload=2twddftaknl8rah03rzkfoti
KEY BSpace
KEY BSpace
KEY BSpace
KEY BSpace
KEY BSpace
KEY BSpace
KEY BSpace
KEY BSpace
KEY BSpace
KEY BSpace
KEY BSpace
KEY BSpace
KEY BSpace
KEY BSpace
KEY BSpace
KEY BSpace
KEY BSpace
KEY BSpace
KEY BSpace
KEY BSpace
KEY BSpace
KEY BSpace
KEY BSpace
KEY Enter
--- payload received by agent at Enter ---
2
--- final composer state ---
typed-buffer=2
enter-sent=yes
```
</details>
<details>
<summary>Evidence: Shared composer delivery proof tests</summary>

Source: [Shared composer delivery proof tests](https://github.com/kunchenguid/firstmate/blob/fa52fdd51fce531e72acbf0b54e5a5904f4deb28/.no-mistakes/evidence/fm/remote-send-delivery-delta-r1/fm-composer-lib-test.txt)

```text
ok - fm_composer_queued_enter_verdict: pending + busy returns empty (queued Enter)
ok - fm_composer_queued_enter_verdict: pending + idle/unknown stays pending
ok - fm_composer_queued_enter_verdict: only proven pending is converted
ok - fm_composer_classify_screen: pre-type gate verdicts do not leak post-Enter
ok - fm_composer_screen_is_gated: only current modal evidence is combined
ok - fm_composer_delivery_delta_verdict: 11 real-harness scenarios, plain and ANSI captures agreeing (22 verdicts)
ok - fm_composer_delivery_delta_verdict: a mid-word wrap, a hint row, and a non-bottom-anchored composer do not block acceptance
ok - fm_composer_delivery_delta_verdict: refuses a swallowing modal the gate scorer does not recognise
ok - fm_composer_delivery_delta_verdict: a trust dialog refuses before typing and again after, both with a reason
ok - fm_composer_delivery_delta_verdict: a scrollback copy plus a ticking pane is not delivery
ok - fm_composer_delivery_delta_verdict: when a redrawing pane defeats novelty, the occurrence count still refuses
ok - fm_composer_delivery_delta_verdict: overlapping anchors cannot manufacture an occurrence increase
ok - fm_composer_delivery_delta_verdict: a partial write that lost the payload tail is refused
ok - fm_composer_delivery_delta_verdict: a payload echoed with no container at all is a healthy send
ok - fm_composer_delivery_delta_verdict: an unusable payload is refused with a named reason
ok - fm_composer_delivery_delta_verdict: short collision anchors are refused
ok - fm_composer_envelope_prepare: a short payload gets a fresh full-length anchor and an exact erase count
ok - fm_composer_envelope_prepare: a single-character suffix is selected afresh
ok - fm_composer_envelope_prepare: a self-proving payload is typed verbatim with no envelope
ok - fm_composer_envelope_prepare: an unerasable short payload is refused, not guessed at
ok - fm_composer_envelope_boundary_verdict: opposite checkpoints pin both erase directions
ok - fm_composer_typed_delivery_core: a short steer is proven and leaves only the payload in the composer
ok - fm_composer_typed_delivery_core: a self-proving steer is delivered with no envelope
ok - fm_composer_typed_delivery_core: every unproven successful write is typed-unproven
ok - fm_composer_typed_delivery_core: unchanged visible tail never invites a duplicate write
ok - fm_composer_typed_delivery_core: a rising count prevents ambiguous row diff from inviting retry
ok - fm_composer_typed_delivery_core: an earlier copy of the payload on screen proves nothing
ok - fm_composer_typed_delivery_core: a gated modal is refused before anything is typed
ok - fm_composer_typed_delivery_core: unavailable boundary probes refuse before typing
ok - fm_composer_typed_delivery_core: no erase primitive refuses short steers before typing, long steers still work
ok - fm_composer_typed_delivery_core: erase checkpoints and final payload retention refuse every measured direction
ok - fm_composer_typed_delivery_core: every way the erase can fail reports typed-unproven, never a bare refusal
ok - fm_composer_typed_delivery_core: stranded warnings are exact transport evidence only
ok - fm_composer_delivery_delta_verdict: box glyphs in the payload vanish on both sides, not just one
ok - fm_composer_delivery_delta_verdict: a narrow single-row composer that scrolls its buffer still proves delivery
```
</details>
<details>
<summary>Evidence: fm-send strict behavior tests</summary>

Source: [fm-send strict behavior tests](https://github.com/kunchenguid/firstmate/blob/fa52fdd51fce531e72acbf0b54e5a5904f4deb28/.no-mistakes/evidence/fm/remote-send-delivery-delta-r1/fm-send-strict-test.txt)

```text
ok - fm-send strict: exact task/lane ids resolve through home metadata
ok - fm-send --key: exit status follows delivery, and an undelivered key never reports success
ok - fm-send strict: unset FM_HOME fails before target resolution
ok - fm-send strict: unresolvable selectors do not fall back to tmux
ok - fm-send strict: prefixless herdr pane ids are rejected before tmux fallback
ok - fm-send strict: unmatched single-colon explicit targets must verify live before sending
ok - fm-send strict: fm-prefixed Herdr sessions remain explicit backend targets
ok - fm-send strict: healthy fm-<id> sends still type once and submit
ok - fm-send: a Claude trust-dialog pane is refused loudly, with no Enter and no delivered marker
ok - fm-send: after-type capture failure preserves unresolved durable recovery
ok - fm-send: typed resolution accepts only exact correlation ids
ok - fm-send: missing payload tail is not-accepted, sends no Enter, and closes no resolve-key
ok - fm-send: a healthy send is confirmed with no composer container to parse at all
```
</details>
<details>
<summary>Evidence: Remote delivery and settlement tests</summary>

Source: [Remote delivery and settlement tests](https://github.com/kunchenguid/firstmate/blob/fa52fdd51fce531e72acbf0b54e5a5904f4deb28/.no-mistakes/evidence/fm/remote-send-delivery-delta-r1/fm-send-remote-delivery-test.txt)

```text
ok - fm-send remote: delivered-unconfirmed reports delivered, exits 0, keeps the expectation armed
ok - fm-send remote: typed-unproven settlement guidance stays exact and suffix-aware
ok - fm-send remote: a real remote failure still fails loudly with the remote diagnostics
ok - fm-send remote: ssh 255 still refuses loudly and preserves the expectation as delivery_unknown
ok - fm-send remote: a delivered-unconfirmed answer closes its --resolve-key decision
ok - fm-send local: an unconfirmed submit exits 3 with an honest non-error report
ok - fm-send local: an unconfirmed submit still never closes a --resolve-key decision
ok - fm-send local: an unconfirmed secondmate send keeps its reply expectation armed
all fm-send-remote-delivery tests passed
```
</details>
<details>
<summary>Evidence: Lifecycle control tests</summary>

Source: [Lifecycle control tests](https://github.com/kunchenguid/firstmate/blob/fa52fdd51fce531e72acbf0b54e5a5904f4deb28/.no-mistakes/evidence/fm/remote-send-delivery-delta-r1/fm-control-test.txt)

```text
ok - fm-control exit: every verified harness gets its own verified exit command
ok - fm-control interrupt: every verified harness gets its own verified key and repeat count
ok - fm-control interrupt: opencode needs a double Escape, claude a single one
ok - fm-control: a harness with no verified control mechanics is refused, not guessed at
ok - fm-control-lib: a recorded harness resolves to its verified adapter without guessing
ok - fm-control: prefixed recorded harnesses reach interrupt and exit mechanics
ok - fm-control-lib: the backend key matrix matches each adapter's real send-key surface
ok - fm-control-lib: adapter capability is per task kind, not per adapter alone
ok - fm-control interrupt: a backend that cannot deliver the harness's key refuses instead of sending another
ok - fm-control: a backend that cannot prove an agent stopped refuses exit and relaunch
ok - fm-control-lib: stop-proving verbs are gated on the backends that really classify agent state
ok - fm-control: a legacy window label is refused and the exact task id is named
ok - fm-control: an explicit backend endpoint is never a control target
ok - fm-control: an unrecorded task id is refused
ok - fm-control: a record whose endpoint identity names another task is refused
ok - fm-control: a remotely placed secondmate is refused by placement, not by a metadata complaint
ok - fm-control: interrupt and exit lock before task-state resolution
ok - fm-control: the verb list is closed - no raw keys, arbitrary text, or clear verb
ok - fm-control: resume is refused with the determinism reason and the alternative
ok - fm-control: profile and note flags belong to relaunch only
ok - fm-control exit: an already-stopped agent is idempotent success with no bytes sent
ok - fm-control exit: a vanished endpoint refuses instead of silently succeeding
ok - fm-control interrupt: refuses when no agent is running rather than keying a shell
ok - fm-control exit: an endpoint whose process cannot be attributed refuses
ok - fm-control exit: a busy agent receives interrupt delivery before the exit command
ok - fm-control exit: an idle agent goes straight to its exit command
ok - fm-control interrupt: unconfirmed delivery preserves observed busy state
ok - fm-control interrupt: muse confirms cancellation from its session log
ok - fm-control interrupt: postconditions are revalidated after acknowledgement polling
ok - fm-control exit: an interrupt-stopped agent satisfies the gone-state postcondition
ok - fm-control exit: a stubborn agent reports delivered input and an unconfirmed exit
ok - fm-control exit: pre-Enter refusals fail immediately
ok - fm-control interrupt: grok reports delivery without claiming cancellation
ok - fm-control interrupt: grok's idle footer does not confirm cancellation
ok - fm-control: a lifecycle command to a secondmate is unmarked and opens no reply expectation
ok - fm-control's arrival leaves fm-send's from-firstmate marking untouched
```
</details>
<details>
<summary>Evidence: Pending-reply durability tests</summary>

Source: [Pending-reply durability tests](https://github.com/kunchenguid/firstmate/blob/fa52fdd51fce531e72acbf0b54e5a5904f4deb28/.no-mistakes/evidence/fm/remote-send-delivery-delta-r1/fm-pending-reply-test.txt)

```text
ok - pending request bodies are never scanned as records
ok - pending reply creation removes partial serialization artifacts
ok - pending reply creation falls back to sha256sum
ok - normal correlated reply resolves once (idempotent)
ok - completed turn with no report triggers exactly one recovery
ok - recovery transport releases and reacquires its correlation lock around delivery
ok - recovery validates durable bodies and preserves legacy summary fallback
ok - recovery exit 4 remains distinct, visible, and operator-settleable
ok - recovery attempts reconcile without reinjection
ok - recovery reply resolves the original expectation
ok - second missed turn escalates once and remains durable
ok - escalations and replies wake; the home's own escalation close stays quiet
ok - failed escalation publication remains retryable and publishes once
ok - legacy escalation closes under the shared default key
pending-reply: request body missing for fd373b99922e67f5
ok - legacy escalation cannot close an unrelated default-key decision
pending-reply: request body missing for 39bad4bbe33f2269
ok - foreign correlated blocker cannot impersonate a pending-reply escalation
ok - concurrent resolution closes one keyed escalation exactly once
ok - concurrent escalation yields to a late correlated reply
ok - transport success cannot masquerade as reply success
ok - undelivered records remain immutable across scan paths
ok - delivery confirmation fallback reconciles durably
ok - unrelated events and stale correlation ids cannot resolve
ok - restart preserves expectation and exact parent destination
ok - wrong-home reports are detected but do not silently acknowledge
ok - direct unmarked captain input creates no expectation
ok - fm-send marked secondmate path creates pending and embeds corr
ok - status-pointed document resolves the expectation
ok - optional helper report resolves without being required for correctness
ok - backend busy/idle observation covers Pi/Claude paths without conversation scrape
ok - tmux and zellij unknown states use bounded capture fallback
ok - pending replies scope Kimi capture fallback by recorded harness
ok - tick skips terminal records and reuses target observations
ok - correlations are reused only for matching open task records
ok - tick end-to-end: miss -> one recovery -> escalate -> durable
pending-reply: request body missing for 42b6ba96864e74b0
ok - typed-unproven records escalate once with exact suffix-aware settlement guidance
ok - typed-unproven supersedes only unconfirmed delivery uncertainty and stays settleable
ok - typed abandonment serializes with watcher escalation
ok - failed transport discards undelivered expectation only
ok - a remote repost waits for the reply channel and still fires on a real miss
ok - a mirrored correlated remote reply resolves without any repost
ok - all pending-reply tests passed
```
</details>
<details>
<summary>Evidence: Daemon typed-unproven retry containment tests</summary>

Source: [Daemon typed-unproven retry containment tests](https://github.com/kunchenguid/firstmate/blob/fa52fdd51fce531e72acbf0b54e5a5904f4deb28/.no-mistakes/evidence/fm/remote-send-delivery-delta-r1/fm-daemon-test.txt)

```text
ok - a captain-held transfer classifies as pause, not as a wedge candidate
ok - housekeeping re-surfaces a forgotten captain hold on the long cadence and resets its window
ok - housekeeping clears the pause marker once a captain hold is answered
ok - housekeeping moves a captain hold's existing stale marker to pause before wedge escalation
ok - fm-send returns 3 with a non-error no-resend warning when confirmation stays pending
ok - inject_msg: typed-unproven never retries even when wedge persistence fails
```
</details>
<details>
<summary>Evidence: Backend abstraction tests</summary>

Source: [Backend abstraction tests](https://github.com/kunchenguid/firstmate/blob/fa52fdd51fce531e72acbf0b54e5a5904f4deb28/.no-mistakes/evidence/fm/remote-send-delivery-delta-r1/fm-backend-test.txt)

```text
ok - fm_backend_name: FM_BACKEND env > config/backend > default tmux
ok - fm_backend_detect: no markers -> undetected, HERDR_ENV=1 -> herdr, $TMUX -> tmux, CMUX_WORKSPACE_ID -> cmux, nested combinations resolve innermost-first
ok - fm_backend_detect: falls back to __CFBundleIdentifier=com.cmuxterm.app when CMUX_WORKSPACE_ID is absent (signal bundle-id; foreign bundle ids rejected)
ok - fm_backend_detect: the cmux fallback signals are macOS-only (inert on a non-Darwin uname)
ok - fm_backend_detect: an inherited cmux bundle id never outranks $TMUX or HERDR_ENV (tmux/herdr-inside-cmux false positive absorbed)
ok - fm_backend_detect: ancestry fallback matches the lsappinfo-resolved (bundle-id) cmux app pid in the parent chain
ok - fm_backend_detect: ancestry fallback matches a bundle-shaped cmux comm path at any install location when lsappinfo cannot resolve a pid
ok - fm_backend_detect: ancestry fallback stops undetected at launchd (a reparented tmux server never reaches cmux)
ok - fm_backend_name: a fallback-detected cmux prints a NOTICE naming the fallback signal; the primary-marker notice is unchanged
ok - fm_backend_name: auto-detect selects herdr or cmux (loud notice) or tmux (silent, including nested tmux-in-herdr/tmux-in-cmux)
ok - fm_backend_name: an explicit FM_BACKEND or config/backend setting always wins over runtime auto-detection, including an ambient cmux marker
ok - fm_backend_validate: implemented adapters accepted, unknown and blocked codex-app backends refused loudly
ok - zsh: fm_backend_source recognizes known backends and rejects unknown ones
ok - bash: fm_backend_source recognizes known backends and rejects unknown ones
ok - fm_backend_validate_spawn: all implemented lifecycle backends are spawn-supported
ok - fm_meta_get / fm_backend_of_meta: read key=value, default backend to tmux
ok - fm_backend_resolve_selector: session:window literal, exact task id first, legacy fm-<id> label fallback, ad hoc bare name via tmux list-windows
ok - fm_backend_of_selector: exact task ids, legacy fm-<id> labels, and matching explicit targets inherit metadata backend
ok - fm-send.sh: explicit tmux targets are verified; text types once and submits with Enter
ok - fm-peek.sh: capture-pane invocation and output are byte-identical old vs new
ok - fm-spawn.sh: a project reached through a symlinked prefix (e.g. macOS /tmp -> /private/tmp) does not trip the isolation guard's false refusal
ok - fm-teardown.sh: treehouse return remains compatible while tmux cleanup uses exact selectors
ok - fm-spawn.sh --backend bogus is refused loudly
ok - fm-spawn.sh --backend codex-app is refused
ok - fm-spawn.sh honors FM_BACKEND and refuses an unimplemented value loudly
ok - fm-spawn.sh: an explicit --backend tmux resolves silently and writes no backend= (missing means tmux)
ok - fm-spawn.sh: explicit --backend tmux wins over an ambient HERDR_ENV=1 auto-detect marker
ok - fm-spawn.sh: auto-detect resolves nested tmux-in-herdr to tmux and stays silent end to end
```
</details>
<details>
<summary>Evidence: cmux backend tests</summary>

Source: [cmux backend tests](https://github.com/kunchenguid/firstmate/blob/fa52fdd51fce531e72acbf0b54e5a5904f4deb28/.no-mistakes/evidence/fm/remote-send-delivery-delta-r1/fm-backend-cmux-test.txt)

```text
ok - fm_backend_cmux_version_check: accepts the verified minimum (0.64.17)
ok - fm_backend_cmux_version_check: accepts a newer version (0.70.0)
ok - fm_backend_cmux_version_check: refuses an old version loudly
ok - fm_backend_cmux_version_check: refuses loudly when cmux is not found on PATH or at the bundle path
ok - fm_backend_cmux_password: reads the first non-empty line of config/cmux-socket-password
ok - fm_backend_cmux_password: preserves spaces and tabs in config/cmux-socket-password
ok - fm_backend_cmux_password: respects FM_CONFIG_OVERRIDE
ok - fm_backend_cmux_password: empty when config/cmux-socket-password is absent
ok - fm_backend_cmux_cli: exports CMUX_SOCKET_PASSWORD when config/cmux-socket-password is set
ok - fm_backend_cmux_parse_target: splits '<workspace_uuid>:<surface_uuid>' on the first colon
ok - fm_backend_cmux_normalize_key: Enter/Escape/C-c map to cmux's verified enter/escape/ctrl-c
ok - fm_backend_cmux_scoped_title: scopes a primary task title with firstmate plus root hash
ok - fm_backend_cmux_scoped_title: scopes a secondmate task title with the home marker plus root hash
ok - fm_backend_cmux_scoped_title: includes the resolved FM_ROOT hash in the home label
ok - fm_backend_validate: cmux is a known backend
ok - fm_backend_busy_state: cmux (no native primitive) always reports unknown, same as tmux/zellij/orca
ok - fm_backend_composer_state: routes cmux to the cmux composer classifier
ok - fm_backend_cmux_ping_state: reports 'ok' on PONG
ok - fm_backend_cmux_ping_state: reports 'denied' when socketControlMode=cmuxOnly rejects the connection
ok - fm_backend_cmux_ping_state: reports 'unauth' when password mode rejects a missing/wrong password
ok - fm_backend_cmux_ping_state: reports 'unauth' when password mode rejects a wrong password (Invalid password)
ok - fm_backend_cmux_ping_state: reports 'down' when the app is not running yet
ok - fm_backend_cmux_ensure_running: returns immediately when cmux is already reachable
ok - fm_backend_cmux_ensure_running: fails fast on a denied socket without attempting to launch, naming every viable mode
ok - fm_backend_cmux_ensure_running: fails fast on an unauthenticated socket, naming the password config and the Automation mode alternative
ok - fm_backend_cmux_create_task: refuses a duplicate workspace title (cmux's own new-workspace has no uniqueness check)
ok - fm_backend_cmux_create_task: creates a workspace and parses workspace_id/surface_id from list responses
ok - fm_backend_cmux_target_ready: fails when the workspace/surface is not found (list-panes structural check)
ok - fm_backend_cmux_target_ready: verifies the workspace title against the expected label first
ok - fm_backend_cmux_target_ready: rejects a workspace id reused under a different title
ok - fm_backend_cmux_capture: fetches generously and trims to N lines locally
ok - fm_backend_cmux_capture: propagates a read-screen failure even when stdout is empty
ok - fm_backend_cmux_capture: fails when the target surface is absent
ok - fm_backend_cmux_send_key: normalizes the key (Escape -> escape) and targets the explicit workspace/surface
ok - fm_backend_cmux_send_key: recovers stale workspace/surface ids by expected label
ok - fm_backend_cmux_send_literal: calls send with an explicit workspace/surface and a -- separator
ok - fm_backend_cmux_send_text_line: clears partial input when Enter fails
ok - fm_backend_cmux_send_text_line: reports unsafe input when cleanup also fails
ok - fm_backend_cmux_current_path: actively probes with marked begin/end lines (zellij-shape frozen cwd)
ok - fm_backend_cmux_composer_state: a bare '❯' composer row reads empty
ok - fm_backend_cmux_composer_state: a borderless Claude '❯' composer row reads empty
ok - fm_backend_cmux_composer_state: a borderless Claude row outranks stale bordered scrollback
ok - fm_backend_cmux_composer_state: a borderless Claude '❯'+NBSP composer row reads empty under LC_ALL=C
ok - fm_backend_cmux_composer_state: plain-capture text after a bare glyph degrades to unknown (never false pending)
ok - fm_backend_cmux_composer_state: the ghost placeholder text reads empty, not pending
ok - fm_backend_cmux_composer_state: real composer text reads pending
ok - fm_backend_cmux_composer_state: a slash-command popup's argument-hint placeholder still reads pending (the incident fix)
ok - fm_backend_cmux_composer_state: reports unknown when the surface cannot be captured
ok - fm_backend_cmux_composer_state: reports unknown when no border-delimited composer row is found
ok - fm_backend_cmux_send_text_submit: reports 'empty' once the composer row reads empty after one Enter
ok - fm_backend_cmux_send_text_submit: reports 'pending' when the composer never clears after retried Enters (swallowed)
ok - fm_backend_cmux_send_text_submit: retries past a popup-placeholder-fill Enter and lands the real second Enter (the incident fix)
ok - fm_backend_cmux_send_text_submit: reports 'send-failed' when the target workspace/surface is absent
ok - fm_backend_cmux_window_of_workspace: walks windows and counts the membership-confirming workspace list
ok - fm_backend_cmux_window_of_workspace: echoes nothing when no window holds the workspace
ok - fm_backend_cmux_kill: closes the task workspace directly when it is not the last in its window
ok - fm_backend_cmux_kill: adds a throwaway sibling then closes the target when it is the last workspace in its window
ok - fm_backend_cmux_kill: never fails even when close-workspace fails
ok - fm_backend_cmux_kill: recovers stale workspace/surface ids by expected label
ok - fm_backend_cmux_list_live: lists only this home's scoped task workspaces using plain fm-<id> labels
ok - fm-spawn.sh: refuses backend=cmux for --secondmate spawns (mirrors Orca's refusal; no secondmate launch design exists yet)
```
</details>

- Evidence: [Herdr backend tests](https://github.com/kunchenguid/firstmate/blob/fa52fdd51fce531e72acbf0b54e5a5904f4deb28/.no-mistakes/evidence/fm/remote-send-delivery-delta-r1/fm-backend-herdr-test.txt)
- Evidence: [Orca backend tests](https://github.com/kunchenguid/firstmate/blob/fa52fdd51fce531e72acbf0b54e5a5904f4deb28/.no-mistakes/evidence/fm/remote-send-delivery-delta-r1/fm-backend-orca-test.txt)
- Evidence: [Zellij backend tests](https://github.com/kunchenguid/firstmate/blob/fa52fdd51fce531e72acbf0b54e5a5904f4deb28/.no-mistakes/evidence/fm/remote-send-delivery-delta-r1/fm-backend-zellij-test.txt)

## Pipeline

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>
<summary>🔧 **Review** - 3 issues found → auto-fixed (2) ✅</summary>

- 🚨 `bin/fm-pending-reply-lib.sh:1022` - The required “full-payload durable recovery” is not lossless for requests ending in newlines. `body=$(fm_pending_reply_request_body ...)` strips all trailing newlines, and line 1103 performs another command substitution on the assembled recovery message. Thus a persisted request such as `command\n\n` is recovered as `command` without error. Preserve the body through result variables or another trailing-newline-safe boundary all the way into `prepared_message`.
- 🚨 `bin/fm-composer-lib.sh:1477` - The checkpoint accepts any occurrence of `payload + first suffix character`, although this can be only two characters (for example `2k`). On an unverified backend whose backspace over-erases during the first N−1 keys, ordinary pane churn can add that short probe at the checkpoint and remove it after the final key while preserving the payload occurrence count; both erase verdicts then pass and Enter submits a damaged or empty composer. This exceeds the explicitly accepted residual limited to the final erase. Revisit whether the checkpoint needs stronger send-specific evidence or explicitly authorize this additional probabilistic containment.
- ⚠️ `bin/fm-send.sh:744` - Every `typed-unproven` result is reported as a pane capture failure, but the shared core also emits it after a successful capture when the delta proof fails, an erase key fails, or either erase proof fails. The preceding warning may name the real cause, leaving operators with contradictory diagnostics. Use a neutral “delivery or suffix removal could not be proven” summary while retaining the detailed transport warning.

🔧 Fix: Clarify unproven delivery and erase residuals
2 errors still open:

- 🚨 `bin/fm-composer-lib.sh:1241` - The whole-screen COUNT proof uses non-overlapping matches, so it can increase even when the actual occurrence count does not. For a 24-character anchor of all `a`s, a before capture containing 25 `a`s has two overlapping occurrences but reports 1; an unrelated redraw to `24*a + b + 24*a` still has two actual occurrences, reports 2, and supplies an added row containing the anchor. The delta verdict therefore accepts although the payload was swallowed. Count overlapping occurrences at this shared boundary so the required occurrence increase is real.
- 🚨 `bin/fm-pending-reply-lib.sh:353` - The heredoc write that constructs the durable pending-reply record is unchecked. If it fails partway, such as under ENOSPC, execution still chmods and renames the truncated temporary file and returns a correlation id, allowing delivery to proceed without a valid durable recovery record. Check the write result and remove both temporary/body files before returning failure.

🔧 Fix: Harden occurrence counting and pending record writes
</details>

<details>

✅ No issues found.
- `bash tests/fm-composer-lib.test.sh`
- `bash tests/fm-send-strict.test.sh`
- `bash tests/fm-send-remote-delivery.test.sh`
- `bash tests/fm-control.test.sh`
- `bash tests/fm-pending-reply.test.sh`
- `bash tests/fm-backend.test.sh`
- `bash tests/fm-backend-cmux.test.sh`
- `bash tests/fm-backend-herdr.test.sh`
- `bash tests/fm-backend-orca.test.sh`
- `bash tests/fm-backend-zellij.test.sh`
- <code>Hermetic end-to-end invocation of `./bin/fm-send.sh demo &#39;2&#39;` against an echoing tmux pane, verifying the random 23-character suffix was erased before Enter and the agent received exactly `2`.</code>
</details>

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>


kunchenguid:
Superseded and dropped by captain decision 2026-08-22: the payload-delta/three-state composer-verification approach is being replaced wholesale by the inbox+doorbell steering design (data/fm-send-reliability-reframe-s1/report.md), which removes the composer-verification problem entirely rather than solving it. Not landing this branch. Follow-up implementation tracked as fm-send-inbox-doorbell-r1.

PR #2818, opened 2026-08-23
test: include process helper in lifecycle fixtures

Ikha14:
## Summary
- copy the new `fm-ps-lib.sh` runtime sibling into five hermetic lifecycle fixture families
- keep production lifecycle refusals unchanged because the operator paths work when their complete runtime surface is present

## Diagnosis
The Windows process abstraction made `fm-wake-lib.sh` source `fm-ps-lib.sh`. These fixtures copied or linked `fm-wake-lib.sh` without its new public runtime sibling. Their subprocesses therefore exited while loading the incomplete fixture, before exercising AFK return, Claude auto-arm, teardown, session ancestry, or turn-end behavior. Setting `FM_PS_MSYS=0` restored the cases that did not require process identity, which isolated the missing dependency from active-home state; adding the sibling restored all five full targeted tests.

## Validation
- `tests/fm-afk-return.test.sh`
- `tests/fm-claude-stop-autoarm.test.sh`
- `tests/fm-gotmp.test.sh`
- `tests/fm-session-lock-ancestry.test.sh`
- `bin/fm-lint.sh`
- `git diff --check`

All passed.


Ikha14:
Closing as premature: this fixture-only fix depends on the unpublished `fm-ps-lib.sh` substrate, which is absent from current `main`. No replacement PR will be opened until that substrate lands.

PR #2882, opened 2026-08-23
fix(bin): deliver Flowy results through a durable retrying outbox

cedmos:
## What Changed

- `bin/fm-completion-outbox.sh` becomes a durable delivery path instead of a script that could not deliver at all. It resolves the endpoint from `config/grok-flowy-webhook` and the Bearer key from the Keychain, so a normal invocation needs no environment variable; the URL and the key are handed to `curl -K -` on stdin and the JSON payload goes through a `umask 077` mode-600 file, so neither the endpoint nor the key ever appears in `ps` output.
- Delivery is now record-first. One immutable record per task is written before any network call, carrying the result and its artifact path. A record retires, and its per-task receipt under `state/flowy-outbox/delivered/` is written, only on HTTP 2xx; every other outcome leaves the record pending for a later pass, so a receipt cannot exist without a successful send. Records carry a monotonic ordinal so that same-second results still order correctly, and the aggregate `state/flowy-last.md` stays a display snapshot that never gains delivery authority and keeps its `pending:` line truthful.
- `bin/fm-watch.sh` drains all pending records on every pass under a single wall-clock ceiling, and records terminal results on the signal path without opening a socket. Because the wake only becomes visible when the watcher process exits, the post-record drain is detached into its own process group with a bounded deadline and a TERM-then-KILL backstop, so delivery never delays the wake. The group signal verifies the child's pgid before signalling, matching the existing guard elsewhere in the file.
- `tests/fm-flowy-completion.test.sh` covers the contract through the executable interface with a stubbed endpoint and no real webhook: a failed POST retried to success across watcher generations, two concurrent terminal results that do not overwrite each other, delivery state surviving a restart, the receipt-only-after-2xx rule, snapshot ordering, and that the key never reaches the process table.
- `docs/configuration.md` gains the operator-facing setup and the `FM_FLOWY_*` bounds; `docs/scripts.md` gains the toolbelt row.

## Risk Assessment

⚠️ Medium: The change is well covered and the delivery invariants are enforced at the code path rather than only documented, but it adds a detached child process and a process-group signal on the watcher's exit path, which is the highest-consequence area touched. The bounded deadline, pgid verification, and dead-owner lock takeover are the mitigations, and timing out retires nothing.

## Testing

The validation pipeline's review, test, document, and lint steps all completed on this branch. The focused suite covers the delivery contract end to end against a stubbed endpoint, and each new case was demonstrated to be a genuine regression test by reverting the specific logic it guards. The three adjacent watcher suites that exercise the cleanup and poll paths were re-run because `fm-watch.sh` changed. ShellCheck is clean at the pinned version. The workflow linter reports its binary missing inside the pipeline's own shell; with the pinned `actionlint 1.7.12` on PATH it validates all three workflow files and exits 0, and no workflow file is in scope for this branch.

No screenshot artifact applies: this change has no rendered UI surface. No real webhook was sent from the test suite, and no secret appears in the diff or in any committed file.

### DIFF
--- a/bin/fm-completion-outbox.sh
+++ b/bin/fm-completion-outbox.sh
@@ -22,8 +22,22 @@
 # caller on a latency-sensitive path can never be stalled by an unreachable
 # Flowy. Some later --drain is the sole owner of delivering what it recorded.
 #
+# CALLERS. bin/fm-watch.sh uses --no-drain on its signal path, because that path
+# runs immediately before the wake and must never wait on the network. Its
+# per-pass --drain is the sole AUTHORITATIVE delivery owner. Because a wake ends
+# the watcher process, a result recorded on the signal path would otherwise wait
+# for the next watcher generation to deliver it, so the watcher also runs one
+# --drain from its EXIT path after the wake is already durable. That exit drain
+# is a promptness optimization only, and it carries a hard total wall-clock
+# bound of FM_FLOWY_EXIT_DRAIN_SECONDS (default 5), well under the watcher's
+# 15-second poll: without the bound a blackholed endpoint would keep the exiting
+# watcher alive and delay re-arming, which is exactly the stall the record-only
+# signal path removes, moved one step later. Hitting the bound is a normal, safe
+# outcome - the record simply stays pending for the next per-pass drain.
+#
 # DRAIN (--drain, and implicitly after --status). Pending records are drained
-# oldest first and POSTed as JSON to the URL in config/grok-flowy-webhook,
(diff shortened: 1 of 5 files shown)


cedmos:
Closing this one: it wires a captain-local integration (Flowy/Workflowy intake with a private webhook endpoint and Keychain credential) into the shared watcher, which - as the review correctly flagged - conflicts with the repo's vendor-orthogonality. If a durable completion delivery is wanted upstream, it should come back as a generic outbox adapter with a pluggable endpoint, without Flowy-specific conventions in the command layer. Thanks for the review findings.

PR #2905, opened 2026-08-24
Recognize oh-my-pi engine (omp) as the Pi harness

torresposso:
## Problem

The Pi runtime "oh-my-pi" (installed via mise) launches its session engine as the `omp` binary, but firstmate's harness identity tables only listed `pi` / `pi-signed`.

- The session-lock ancestry walk found no verified harness, so every firstmate session under this runtime started read-only.
- A leaked `CLAUDECODE=1` marker in the pane environment additionally mislabeled the harness as `claude`.

## Fix

Two small changes:

- `bin/fm-session-lock-lib.sh`: add `^omp$` to `FM_HARNESS_RE` and `omp` to `FM_HARNESS_NAMES` so the lock ancestry walk recognizes the engine.
- `bin/fm-harness.sh`: detect an `omp` process in the ancestry as authoritative Pi (outranking any foreign `CLAUDECODE` marker), and match `omp` in the Layer 2 basename case.

`omp` appears only inside Pi sessions, so the ancestry signal is unambiguous.

## Verification

- `bin/fm-harness.sh` now prints `pi` under oh-my-pi.
- `bin/fm-lock.sh` acquires `state/.lock` with the `omp` pid.
- Sessions under this runtime no longer start read-only.

### DIFF
--- a/bin/fm-harness.sh
+++ b/bin/fm-harness.sh
@@ -50,6 +50,15 @@ detect_own() {
   # CURSOR_AGENT=1 is set for the child/tool processes this script runs as.
   [ "${CURSOR_AGENT:-}" = "1" ] && { echo cursor; return; }
   [ "${CURSOR_INVOKED_AS:-}" = "cursor-agent" ] && { echo cursor; return; }
+  # Pi (oh-my-pi) engine binary is `omp`. It appears only in Pi sessions, so an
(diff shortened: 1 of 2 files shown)


torresposso:
Withdrawing this — keeping the fix local for now.

PR #2918, opened 2026-08-24
fix(bin): canonicalize Windows drive-letter herdr socket paths before identity comparison

3min-lang:
## Problem

On native Windows the herdr server injects and reports drive-letter socket
paths (`C:\Users\...\herdr.sock`), while the launcher's claimed socket arrives
in POSIX spelling. `fm_backend_herdr_socket_identity` compares the two
literally, so the same socket compares unequal. That comparison is the single
owner for the presentation session lock and the launcher-identity
same-session proof, so both reject a socket that is in fact their own.

## Fix

Convert a drive-letter path to its POSIX spelling with `cygpath -u` before the
existing directory resolution, so both sides land in the same form.

The `case` guard matches only `[A-Za-z]:[\/]*`, a shape that cannot occur on
Linux or macOS, so `cygpath` is never invoked there and behaviour on those
platforms is unchanged. An unresolvable drive-letter path returns 1 rather than
comparing a half-converted string.

## Testing and scope

- `bin/fm-lint.sh bin/backends/herdr.sh` clean.
- Read against current `main`: `fm_backend_herdr_socket_identity` still compares
  the two spellings literally, so the mismatch this fixes is still present.

Scope note, stated because it limits what this PR can be said to fix: on the
Windows host where this was found, `fm_backend_herdr_presentation_session_lock_path`
fails earlier, at `fm_backend_herdr_presentation_lock_namespace_valid`, which
requires mode `700`. Git Bash mounts both `/c` and `/tmp` with `noacl`, so
`mkdir -m 700` is a silent no-op there and the namespace directory reads `755`.
That is a separate defect with a separate cause, and this PR does not address
it. So this change is offered as a correctness fix to the identity comparison
itself, not as an end-to-end repair of presentation locking on Windows.

### DIFF
--- a/bin/backends/herdr.sh
+++ b/bin/backends/herdr.sh
@@ -691,14 +691,23 @@ fm_backend_herdr_presentation_lock_namespace_valid() {
 # -> /private/tmp cannot yield two lock identities for the same socket.
 # fm_backend_herdr_canonical_socket_path: normalize one absolute Unix-socket
 # path so two spellings of the same socket compare equal. Refuses a relative
-# or empty path. An unresolvable directory is left as-is rather than treated as
-# a failure, so a socket whose directory was removed still compares by its own
-# literal path. Single owner for every socket-identity comparison in this
-# adapter (the presentation session lock and the launcher-identity same-session
-# proof both use it).
+# or empty path. On native Windows the herdr server injects and reports
+# drive-letter paths (C:\...\herdr.sock); those are converted to their POSIX
+# spelling first so the launcher's claimed socket and the session list's
+# socket_path land in the same form before comparison. An unresolvable
+# directory is left as-is rather than treated as a failure, so a socket whose
+# directory was removed still compares by its own literal path. Single owner
+# for every socket-identity comparison in this adapter (the presentation
+# session lock and the launcher-identity same-session proof both use it).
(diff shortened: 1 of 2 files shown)


3min-lang:
Added the regression coverage in `72e411c`.

Three cases in `tests/fm-backend-herdr.test.sh`:

- a drive-letter spelling normalizes to the same identity as the POSIX spelling of the same socket
- a failed `cygpath` is refused rather than compared half-converted
- empty `cygpath` output is refused

`cygpath` ships only with Git for Windows, so the cases stub it and run on every platform. The adapter guard matches only `[A-Za-z]:[\/]*`, a shape that cannot occur on Linux or macOS, so nothing about the POSIX path changes.

I checked the cases can fail for the right reason rather than just passing: with the guard removed, the normalization case reports an empty identity against the expected POSIX path.

One note in case you see it on your side too. On this host `tests/fm-backend-herdr.test.sh` already halts earlier, at `not ok - the ambiguity refusal did not name the candidate workspaces (missing: 'w1 w7')`, and that failure reproduces identically on unpatched `main` here, so it is pre-existing and unrelated to this PR. Because the suite stops there, I verified the three new cases in isolation rather than through a full green run of the file.

PR #2949, opened 2026-08-24
feat(bin): add idle-gated Mac mini self-reboot guard

kunchenguid:
## Intent

Build a mini self-reboot capability for the firstmate fleet's Mac mini host. The original design (fleet-memory-health-sustainable-s1 scout report section 6) proposed a generation-scoped host-maintenance drain barrier plus a leak-slope sampler that would refuse new spawn/promotion/handoff work on the mini while draining it, then escalate to the captain for explicit per-reboot approval, never auto-firing a reboot. During implementation, no-mistakes review found the drain/admission-lock primitive was genuinely hard to build correctly: across four consecutive review rounds it kept surfacing new TOCTOU races, forgeable env-var bypasses, and fail-open gaps in the shared lock/lease boundary and the readiness/idleness checks (each round's fix for one race produced a new one in the same area). The captain then explicitly rejected that entire drain/admission-lock/prevent-new-work approach as the wrong product (captain 2026-08-24) and retargeted to a deliberately much simpler design: reboot the mini automatically, with NO captain approval step per reboot, but ONLY when BOTH of two conditions genuinely hold at once - (1) a real memory/kernel-zone leak threshold has been crossed, confirmed over multiple samples (a single sample, however extreme, must never trigger it; the trigger requires three consecutive over-threshold samples, or a slope forecast over a recent window predicting the hard ceiling before the forecast horizon, or pressure no longer normal with swapouts trending up), and (2) the local fleet on that mini is robustly idle, confirmed by two consecutive fully-idle reads at least 5 minutes apart, reusing the existing idleness signals (fm-crew-state.sh's current-state read, tasks-axi backlog state, and the documented state/ layout for steering inboxes, pending replies, promised public followups, and process-event sources) rather than inventing parallel machinery. If either condition does not hold, the guard does nothing; the machine handles load best-effort. This design deliberately has NO drain, NO admission lock, NO blocking or holding of new work, NO fleet-emptying, NO attack-resistant admission boundary, and NO readiness-receipt machinery - it never touches fm-spawn.sh, fm-promote.sh, or fm-backlog-handoff.sh at all. It is mini-only by construction, gated on an explicit config/host-role=mini file; a MacBook or any unconfigured host can never reboot through this code. Reboot execution itself is delegated to a separately-provisioned, root-authorized helper named by config/mini-reboot-helper; this environment has no working passwordless-sudo or other privileged reboot mechanism available (confirmed via sudo -n), so the code deliberately never fakes or stubs privileged execution - it blocks with a clear diagnostic naming the exact missing mechanism when no working helper is configured, and provisioning that helper is an explicit follow-up outside this change's scope. The single-instance check lock and the reboot-in-progress marker are both crash-safe (a dead lock owner is reclaimed; the marker self-clears once the live macOS boot-session id changes, proving the reboot actually happened, or once it goes stale past 20 minutes). Colocated regression tests cover the sampler's three trigger conditions in isolation, every individual idle-blocking condition plus the two-snapshot confirmation gap logic, the mini-only gate, the never-fakes-privileged-execution refusal, the reboot marker's re-fire prevention and self-clearing, the crash-safe lock, and full end-to-end proof that the reboot helper is invoked only when both conditions genuinely hold together and never when only one does. This work was built on a prior branch (fm/fm-safe-auto-reboot-host-drain-barrier-r1) whose no-mistakes run bookkeeping got stuck referencing an old cancelled run's unreachable head; it was cherry-picked cleanly onto this fresh branch (fm/fm-safe-auto-reboot-simple-r1) from current main to get a clean run, with the diff verified to contain only the new design and no leftover drain/lock/barrier code.

## What Changed

- Add a resource sampler and guard that invokes a configured reboot helper only after sustained leak, forecast, or pressure signals coincide with two confirmed idle snapshots.
- Restrict execution to explicitly configured mini hosts, with fail-closed fleet checks, crash-safe overlap handling, and reboot-attempt markers that reset after reboot or expiry.
- Document configuration and operations, and add regression coverage for sampling, idleness, safety gates, and end-to-end reboot decisions.

## Risk Assessment

✅ Low: The reboot capability is sensitive but remains explicitly armed, fail-closed, and gated by independently confirmed resource and idle conditions.

## Testing

Inspected the targeted diff, ran the three focused sampler, idle, and guard behavior suites, and exercised the actual CLI end-to-end with reviewer-visible transcript evidence; the helper ran only after three high-resource samples and two idle reads 400 seconds apart, remained uninvoked with resource pressure alone, and the forbidden spawn/promotion/handoff paths were unchanged.

<details>
<summary>Evidence: Mini reboot guard CLI end-to-end transcript showing the AND gate and helper side effects</summary>

Source: [Mini reboot guard CLI end-to-end transcript showing the AND gate and helper side effects](https://github.com/kunchenguid/firstmate/blob/e4f2adf94c66b7a3af9e4c54a8338e3a0d9a4208/.no-mistakes/evidence/fm/fm-safe-auto-reboot-simple-r1/mini-reboot-cli-e2e.txt)

```text
=== Actual CLI: BOTH resource threshold and two confirmed idle reads ===
--- check at t=100 ---
fm-mini-reboot: resource condition not triggered, doing nothing
--- check at t=200 ---
fm-mini-reboot: resource condition not triggered, doing nothing
--- check at t=300 ---
fm-mini-reboot: resource condition triggered:
fm-mini-reboot:   reason=maintenance-threshold-crossed-in-3-samples
fm-mini-reboot: resource triggered but idle window not confirmed, staying put:
fm-mini-reboot:   confirmed=no
fm-mini-reboot:   reason=not-both-idle(now=idle prev=none)
fm-mini-reboot:   verdict=idle
--- check at t=700 ---
fm-mini-reboot: resource condition triggered:
fm-mini-reboot:   reason=maintenance-threshold-crossed-in-3-samples
fm-mini-reboot: both conditions hold: resource triggered AND idle window confirmed - executing reboot
fm-mini-reboot: invoking reboot helper: /var/folders/5x/4nqprlbx0518k3ybcb1sz6gr0000gn/T//fm-mrb-evidence.Gy5nNc/both/helper (reason: reason=maintenance-threshold-crossed-in-3-samples;)
fm-mini-reboot: reboot helper exited 0
--- observable helper side effect ---
HELPER_INVOKED reason=reason=maintenance-threshold-crossed-in-3-samples;

=== Actual CLI: resource threshold only; idle registry absent ===
--- check at t=100 ---
fm-mini-reboot: resource condition not triggered, doing nothing
--- check at t=200 ---
fm-mini-reboot: resource condition not triggered, doing nothing
--- check at t=300 ---
fm-mini-reboot: resource condition triggered:
fm-mini-reboot:   reason=maintenance-threshold-crossed-in-3-samples
fm-mini-reboot: resource triggered but idle window not confirmed, staying put:
fm-mini-reboot:   confirmed=no
fm-mini-reboot:   reason=not-both-idle(now=busy: no homes registered for idle-check (config/mini-reboot-homes) prev=none)
fm-mini-reboot:   verdict=busy: no homes registered for idle-check (config/mini-reboot-homes)
--- check at t=700 ---
fm-mini-reboot: resource condition triggered:
fm-mini-reboot:   reason=maintenance-threshold-crossed-in-3-samples
fm-mini-reboot: resource triggered but idle window not confirmed, staying put:
fm-mini-reboot:   confirmed=no
fm-mini-reboot:   reason=not-both-idle(now=busy: no homes registered for idle-check (config/mini-reboot-homes) prev=busy: no homes registered for idle-check (config/mini-reboot-homes))
fm-mini-reboot:   verdict=busy: no homes registered for idle-check (config/mini-reboot-homes)
--- observable helper side effect ---
HELPER_NOT_INVOKED
```
</details>

## Pipeline

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>
<summary>🔧 **Review** - 7 issues found → auto-fixed (10) ✅</summary>

- 🚨 `bin/fm-mini-reboot-lib.sh:329` - `latest_pressure != normal` treats `unknown` as confirmed memory pressure. Three samples with an unreadable pressure sysctl and rising swapouts therefore authorize a reboot without evidence that pressure is non-normal. Restrict this trigger to explicit `warn`/`critical` values and fail closed on `unknown`.
- 🚨 `bin/fm-mini-reboot-lib.sh:363` - The idle check skips all state-based blockers when `&lt;home&gt;/state` is absent, so a registered home with only an empty backlog can report idle. Enumeration errors are also hidden by `find | wc/head`, producing zero/empty results. Require a readable state directory and preserve each enumeration command&#39;s failure status before declaring the home idle.
- 🚨 `bin/fm-mini-reboot-lib.sh:448` - `tasks-axi public-followup ready` does not represent every reply still owed: a freshly bound commitment can remain in `pending-work`, as documented and exercised by the existing public-followup lifecycle. With no in-flight/held backlog item, such a promise can be missed and the home declared idle. Query the owned public-followup list/state interface and block on every unresolved owed state. The new test also permits fixture registration failure to skip this required case.
- 🚨 `bin/fm-mini-reboot-lib.sh:657` - The lock is not crash-safe as claimed: `mkdir` publishes the lock before lines 673-675 write its owner. If the process dies in that interval, later checks see an ownerless lock, never reclaim it, and skip forever. Publish an initialized owner atomically at the shared lock boundary, or safely reclaim abandoned ownerless claims; also bind ownership beyond a reusable PID.
- 🚨 `bin/fm-mini-reboot-lib.sh:362` - Intent requires idleness to reuse “fm-crew-state.sh&#39;s current-state read,” but the changed implementation never invokes that owner and instead treats metadata presence alone as busy; docs nevertheless claim the current-state reader is reused. Confirm the intended metadata/current-state semantics and implement the required read or revise the authoritative requirement.
- 🚨 `bin/fm-mini-reboot-lib.sh:561` - The mini-only criterion says `config/host-role` must contain exactly `mini`, but deleting all whitespace accepts malformed values such as `m i n i`. Since this is the reboot authorization gate, ask whether only an exact line with optional surrounding whitespace is intended, then enforce that representation.
- 🚨 `tests/fm-mini-reboot-guard.test.sh:85` - The newly added drain/admission and spawn/promote/handoff tests only grep implementation source for strings. This violates the test-quality rule and cannot prove behavior because matching text may be dead or behavior may exist under different names. Remove these assertions or replace them with executable-interface behavioral checks.

🔧 Fix: Harden reboot pressure, idleness, and locking checks
5 errors still open:

- 🚨 `bin/fm-mini-reboot-lib.sh:369` - Intent requires idleness to reuse “fm-crew-state.sh&#39;s current-state read,” but the implementation only counts `state/*.meta` files and never invokes that current-state owner; documentation nevertheless claims it is reused. Confirm the intended current-state semantics and either implement the required read or revise the authoritative requirement.
- 🚨 `bin/fm-mini-reboot-lib.sh:573` - The mini-only criterion says `config/host-role` must contain exactly `mini`, but deleting all whitespace also accepts malformed values such as `m i n i`. Confirm whether only an exact line with optional surrounding whitespace is intended, then enforce that representation.
- 🚨 `bin/fm-mini-reboot-lib.sh:107` - Metric collection converts an unreadable or malformed `zprint` result to zone_bytes=0. With one older padding sample, that failed zero sample followed by one normal 10 GiB reading appears as a steep recent slope and can authorize a reboot without a genuine leak trend. Reject invalid metric samples at the collection boundary, and make `fm_mrb_check_cycle` stop when appending the current sample fails rather than evaluating stale history.
- 🚨 `bin/fm-mini-reboot-lib.sh:609` - The reboot marker records an empty boot-session ID when `kern.bootsessionuuid` is temporarily unreadable. On the next successful read, empty→current is misclassified as proof of a reboot, so the marker clears and permits another helper invocation in the same boot session. Require a valid boot-session ID before publishing the marker or otherwise distinguish an unknown recorded ID from a changed live ID.
- 🚨 `bin/fm-mini-reboot-lib.sh:699` - Stale-lock reclamation is still racy: after a contender reads stale owner A, A can remove its lock and process B can acquire the same path before this unconditional `rm -rf`; the contender then deletes B&#39;s live lock and enters concurrently. Reclaim through an atomic ownership-preserving boundary so the exact stale claim validated is the one removed.

🔧 Fix: Harden sampling, reboot markers, and lock reclamation

- 🚨 `bin/fm-mini-reboot-lib.sh:386` - Intent requires idleness to reuse “fm-crew-state.sh&#39;s current-state read,” but this path only counts `state/*.meta` and never invokes that owner; `docs/mini-reboot-guard.md` nevertheless claims it is reused. Confirm the intended current-state semantics and either implement the required read or revise the requirement.
- 🚨 `bin/fm-mini-reboot-lib.sh:590` - Intent requires `config/host-role=mini` exactly, but deleting all whitespace also authorizes malformed content such as `m i n i`. Confirm whether only an exact line with optional surrounding whitespace is intended, then enforce that representation.
- ⚠️ `bin/fm-mini-reboot-lib.sh:650` - Deleting all whitespace from `config/mini-reboot-helper` corrupts valid executable paths containing spaces, causing a correctly provisioned helper to be reported missing. Preserve the configured path while removing only the line terminator, unless whitespace is intentionally forbidden.
- 🚨 `bin/fm-mini-reboot-lib.sh:478` - The public-followup parser treats a successful but malformed response such as `{}` as an empty obligation list because of `(.public_followups // [])`. With all other signals clear, the home incorrectly reports idle instead of failing closed. Require `public_followups` to exist and be an array before counting unresolved entries.

🔧 Fix: Fail closed on malformed public followup responses
7 issues (4 errors, 3 warnings) still open:

- 🚨 `bin/fm-mini-reboot-lib.sh:386` - The required idle check must reuse “fm-crew-state.sh&#39;s current-state read,” but this implementation only counts `state/*.meta` files and never invokes that owner; `docs/mini-reboot-guard.md` nevertheless claims it is reused. Confirm whether terminal metadata may be idle, then implement the required current-state read or revise the authoritative requirement.
- 🚨 `bin/fm-mini-reboot-lib.sh:592` - The mini-only criterion requires `config/host-role=mini` exactly, but deleting all whitespace also authorizes malformed content such as `m i n i`. Confirm whether only an exact line with optional surrounding whitespace is intended, then enforce that representation.
- ⚠️ `bin/fm-mini-reboot-lib.sh:652` - Deleting all whitespace from `config/mini-reboot-helper` corrupts valid executable paths containing spaces, causing a correctly provisioned helper to be reported missing. Preserve the configured path while removing only its line terminator unless whitespace is intentionally forbidden.
- 🚨 `bin/fm-mini-reboot-lib.sh:287` - The consecutive-threshold loop ignores the wired value already read from each row and looks it up again by epoch. With three samples sharing an epoch—first wired above threshold and the next two below—every lookup returns the first row, incorrectly reporting three consecutive crossings. Use the current row&#39;s `_w` field directly.
- 🚨 `bin/fm-mini-reboot-lib.sh:322` - The required triggers are three consecutive threshold crossings, a positive slope forecast, or pressure plus rising swapouts, but this branch adds an unconditional fourth trigger whenever the latest zone value is at the hard ceiling. For example, samples at 0, 13 GiB, then 12 GiB trigger despite only two over-threshold samples and a falling recent slope. Confirm whether this additional trigger is intended or remove it.
- ⚠️ `bin/fm-mini-reboot-lib.sh:456` - The pending-reply owner deliberately retains records after setting `phase=resolved`, but this check treats every retained file as an outstanding reply. After any normal resolved reply, the home can therefore remain permanently busy. Parse the owned record state, block only unresolved phases, and fail closed on malformed records.
- ⚠️ `bin/fm-mini-reboot-lib.sh:498` - The process-event owner defines registration as the presence of `state/procevent/*.source`, while this recursive enumeration treats any artifact as a live source. An interrupted publication can leave `.source.*` behind and permanently suppress reboot despite no registered source. Reuse the owner&#39;s registration predicate while preserving fail-closed read handling.

🔧 Fix: Honor reconciled idleness and exact reboot configuration
1 error still open:

- 🚨 `bin/fm-mini-reboot-lib.sh:312` - The required fix says the reboot “must never fire on a falling/improving memory trend,” but the slope uses only the window’s oldest and latest samples. For samples 0 GiB, 11.5 GiB, then 11 GiB, the latest trend is falling while oldest-to-latest `dz` remains positive and forecasts the ceiling, incorrectly triggering. Enforce the non-falling invariant in `fm_mrb_evaluate` using the latest segment/recent trend, and strengthen the regression with a below-ceiling falling endpoint.

🔧 Fix: Block slope triggers during falling recent trends
1 error still open:

- 🚨 `bin/fm-mini-reboot-lib.sh:336` - The pressure trigger sums persisted swapout deltas, including the first selected row&#39;s delta from outside the selected window. For counters 0, 100, 50, 50 with warning pressure, the last three rows are falling then flat, but their deltas sum to +50 and incorrectly trigger a reboot. Determine the trend from the selected rows&#39; raw swapout values, excluding the pre-window delta.

🔧 Fix: Compute swapout trends within the selected window
1 error still open:

- 🚨 `bin/fm-mini-reboot-lib.sh:755` - The required invariant says the single-instance lock is “crash-safe” and a dead owner is reclaimed, but a crash after line 763 removes `check.lock` and before line 768 removes `check.lock.reap` leaves an orphan reap inode. A later process can acquire and release a new lock normally; when that process subsequently dies, `fm_mrb_reclaim_lock` sees the old reap path, skips creating a hard link to the current lock, fails the `-ef` check, and never reclaims the dead owner. Handle a reap artifact that does not identify the current lock by safely retiring it and claiming the current lock before evaluating ownership.

🔧 Fix: Simplify reboot locking to mkdir and pid ownership
2 errors still open:

- 🚨 `bin/fm-mini-reboot-lib.sh:719` - The required lock is crash-safe, but acquisition creates the lock directory before publishing its owner at line 740. A crash in that interval leaves an ownerless directory; later invocations retry three times and permanently skip because empty owners are never reclaimed. Reconcile ownerless locks after the bounded publication grace or use an acquisition boundary that publishes ownership atomically.
- 🚨 `bin/fm-mini-reboot-lib.sh:726` - Dead-owner reclamation breaks single-instance execution under concurrent reclaimers. Two invocations can both observe the dead PID; after the first removes and reacquires the directory, the second can execute its previously authorized `rm -rf`, delete the first invocation&#39;s live lock, reacquire it, and run concurrently. Reclamation must conditionally retire only the exact stale lock generation observed, which conflicts with the requested plain unguarded removal and requires user review.

🔧 Fix: Reclaim ownerless locks and document concurrency limits

- 🚨 `bin/fm-mini-reboot-lib.sh:639` - A detected boot-session change clears only the reboot marker, leaving samples and idle snapshots from the previous boot. After a reboot, one new over-threshold sample can combine with two pre-reboot samples, while one new idle read can combine with the prior boot&#39;s idle snapshot, causing another reboot despite the requirement that a single sample never trigger. Track the boot generation for persisted detection state and reset sample and idle-confirmation history whenever it changes, including after manual reboots.
- ⚠️ `docs/mini-reboot-guard.md:40` - The operator documentation says overlapping invocations “never race,” but the implemented lock explicitly permits concurrent execution during dead-owner reclamation as an accepted limitation. Describe the lock as best-effort redundancy prevention and document that reboot safety comes from the resource-plus-idle gate.

🔧 Fix: Reset reboot evidence across boot sessions
2 errors still open:

- 🚨 `bin/fm-mini-reboot-lib.sh:313` - The consecutive-threshold check ignores each row&#39;s parsed wired value and instead looks it up by epoch. Duplicate timestamps therefore reuse the first row&#39;s wired value: rows `(100, wired=high)`, `(100, wired=low)`, `(200, wired=high)` incorrectly satisfy three consecutive over-threshold samples. Use the `_w` field already read from each row so every sample is evaluated independently.
- 🚨 `tests/fm-mini-reboot-guard.test.sh:333` - The required “full end-to-end proof ... never when only one [condition] does” lacks an idle-only case. This test registers no idle home and its zprint fixture contains no recognized zone, so sampling fails and neither condition holds; it does not prove confirmed idleness alone cannot reboot. Provide valid below-threshold samples plus two confirmed-idle reads, then assert the helper remains uninvoked.

🔧 Fix: Evaluate wired thresholds per persisted sample
</details>

<details>

✅ No issues found.
- <code>`git diff --stat 8fa0505e48a155da78a9aeeb50911719dc558710..d7afce6ad8077f731230731171ceff2bca0083ac` and targeted diff inspection</code>
- `bash tests/fm-mini-reboot-sampler.test.sh`
- `bash tests/fm-mini-reboot-idle.test.sh`
- `bash tests/fm-mini-reboot-guard.test.sh`
- <code>Manual end-to-end execution of `bin/fm-mini-reboot-guard.sh check` across four scheduled samples, once with a registered idle fleet and once with no idle registry; verified the configured helper side effect occurred only in the resource-plus-confirmed-idle scenario</code>
- `git diff --name-only 8fa0505e48a155da78a9aeeb50911719dc558710..HEAD -- bin/fm-spawn.sh bin/fm-promote.sh bin/fm-backlog-handoff.sh`
</details>

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -78,6 +78,9 @@ config/cmux-socket-password  optional cmux control-socket password; LOCAL, gitig
 config/wedge-alarm  optional away-mode wedge-alarm active-alert directives; LOCAL, gitignored; absent means auto (macOS Notification Center when available); see docs/wedge-alarm.md
 config/watched-tools.json  optional list of the tools this home depends on, read by the update check armed with bin/fm-tool-update-check.sh; LOCAL, gitignored, firstmate-maintained but human-editable, and NOT inherited by secondmate homes; see docs/configuration.md "Watched tool updates"
 config/x-mode.env    generated Relay watcher cadence; LOCAL, gitignored; source before arming watcher when present
+config/host-role  must equal exactly "mini" to arm bin/fm-mini-reboot-guard.sh's reboot execution on this host; LOCAL, gitignored, and not inherited; see docs/mini-reboot-guard.md
+config/mini-reboot-homes  explicit registry of FM_HOME paths bin/fm-mini-reboot-guard.sh idle-checks before a reboot attempt; LOCAL, gitignored, and not inherited; absent or empty means the guard never reboots; see docs/mini-reboot-guard.md
+config/mini-reboot-helper  path to the root-authorized reboot helper bin/fm-mini-reboot-guard.sh invokes when both conditions hold; LOCAL, gitignored, and not inherited; absent blocks the reboot attempt rather than faking it; see docs/mini-reboot-guard.md
 data/                personal fleet records; LOCAL, gitignored as a whole
   backlog.md         task queue, dependencies, history
   captain.md         this home's domain-local captain preferences and working style; LOCAL, gitignored, canonical even if harness memory mirrors it, and updated with inspect-then-update
@@ -116,6 +119,7 @@ state/               runtime records and signals; gitignored
   procevent-inbox/   private captured results and their durable handled-acknowledgement markers; source output lives here and never in an event line
   decision-bindings/ private records marking a captured-answer source as feeding the keyed-answer intake, with a legacy origin on pre-collapse records; written only by bin/fm-captain-hold.sh bind, dropped by unbind and by source retirement (section 13; docs/captain-hold-lifecycle.md)
   when/              private condition->action watch specs, their trust bindings, and single-fire markers; written only by bin/fm-procevent-when.sh (section 13's process-event-sources trigger)
+  mini-reboot/       host-level (not per-task) samples, idle snapshots, the single-instance check lock, and the reboot-in-progress marker owned by bin/fm-mini-reboot-guard.sh; see docs/mini-reboot-guard.md
   inbox/             captain notes captured out of band by bin/fm-inbox.sh, including the voice handover's queued requests; each note appends one `check` wake and stays pending until acknowledged with `bin/fm-inbox.sh drain --ack <id>`, which moves it to inbox/handled/ (docs/voice-relay.md)
   x-inbox/           generated Relay pending mention payloads; fmx-respond drains it (section 14)
   x-context/         generated Relay durable per-request reply context and one-wake offer markers, keyed by request_id; survives inbox cleanup and expires within seven days (section 14; bin/fm-x-lib.sh)
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -78,7 +78,7 @@ config/cmux-socket-password  optional cmux control-socket password; LOCAL, gitig
 config/wedge-alarm  optional away-mode wedge-alarm active-alert directives; LOCAL, gitignored; absent means auto (macOS Notification Center when available); see docs/wedge-alarm.md
 config/watched-tools.json  optional list of the tools this home depends on, read by the update check armed with bin/fm-tool-update-check.sh; LOCAL, gitignored, firstmate-maintained but human-editable, and NOT inherited by secondmate homes; see docs/configuration.md "Watched tool updates"
 config/x-mode.env    generated Relay watcher cadence; LOCAL, gitignored; source before arming watcher when present
-config/host-role  must equal exactly "mini" to arm bin/fm-mini-reboot-guard.sh's reboot execution on this host; LOCAL, gitignored, and not inherited; see docs/mini-reboot-guard.md
+config/host-role  must equal exactly "mini" to arm bin/fm-mini-reboot-guard.sh's captain-authorized, narrowly scoped automatic reboot exception on this host (no per-reboot approval); LOCAL, gitignored, and not inherited; see docs/mini-reboot-guard.md
 config/mini-reboot-homes  explicit registry of FM_HOME paths bin/fm-mini-reboot-guard.sh idle-checks before a reboot attempt; LOCAL, gitignored, and not inherited; absent or empty means the guard never reboots; see docs/mini-reboot-guard.md
 config/mini-reboot-helper  path to the root-authorized reboot helper bin/fm-mini-reboot-guard.sh invokes when both conditions hold; LOCAL, gitignored, and not inherited; absent blocks the reboot attempt rather than faking it; see docs/mini-reboot-guard.md
 data/                personal fleet records; LOCAL, gitignored as a whole
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -573,7 +573,7 @@ A current, explicit, concrete captain instruction overrides any conflicting stan
 The instruction must be specific and recent: it must identify the concrete action, object, or bounded set it governs.
 Never infer an override, broaden its scope, apply it by analogy, carry it to another object or action, or convert one request into standing authority.
 Ambiguous scope or conflict still requires one concise clarification before action.
-Destructive, irreversible, security-sensitive, discard, and merge actions still require the captain to state that concrete action explicitly; once the captain does so and higher-priority instructions permit it, a conflicting Firstmate-written rule must not rigidly block the action.
+Destructive, irreversible, security-sensitive, discard, and merge actions still require the captain to state that concrete action explicitly; once the captain does so and higher-priority instructions permit it, a conflicting Firstmate-written rule must not rigidly block the action. The sole standing exception is a deterministic host-maintenance capability that the captain explicitly authorized and separately provisioned under `VISION.md`; it may perform only its documented action when every fail-closed gate holds.
 Standing `yolo` merge authority is not a substitute for a current explicit captain instruction where an explicit action is required.
 
 ## Maintaining this file
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -119,7 +119,7 @@ state/               runtime records and signals; gitignored
   procevent-inbox/   private captured results and their durable handled-acknowledgement markers; source output lives here and never in an event line
   decision-bindings/ private records marking a captured-answer source as feeding the keyed-answer intake, with a legacy origin on pre-collapse records; written only by bin/fm-captain-hold.sh bind, dropped by unbind and by source retirement (section 13; docs/captain-hold-lifecycle.md)
   when/              private condition->action watch specs, their trust bindings, and single-fire markers; written only by bin/fm-procevent-when.sh (section 13's process-event-sources trigger)
-  mini-reboot/       host-level (not per-task) samples, idle snapshots, the single-instance check lock, and the reboot-in-progress marker owned by bin/fm-mini-reboot-guard.sh; see docs/mini-reboot-guard.md
+  mini-reboot/       host-level (not per-task) samples, idle snapshots, the best-effort overlap check lock, and the reboot-in-progress marker owned by bin/fm-mini-reboot-guard.sh; see docs/mini-reboot-guard.md
   inbox/             captain notes captured out of band by bin/fm-inbox.sh, including the voice handover's queued requests; each note appends one `check` wake and stays pending until acknowledged with `bin/fm-inbox.sh drain --ack <id>`, which moves it to inbox/handled/ (docs/voice-relay.md)
   x-inbox/           generated Relay pending mention payloads; fmx-respond drains it (section 14)
   x-context/         generated Relay durable per-request reply context and one-wake offer markers, keyed by request_id; survives inbox cleanup and expires within seven days (section 14; bin/fm-x-lib.sh)
--- a/VISION.md
+++ b/VISION.md
@@ -23,8 +23,8 @@ Presentation and convenience features that serve that experience are welcome whe
 The captain is the default authority for every gate; autonomy exists only as an explicit grant, never as a default, and new capability ships as an option to enable, never as behavior that assumes consent.
 The first mate reads projects but does not change them; project changes belong to workers in isolated copies, delivered through each project's selected path.
 The first mate stays free to command by never doing the work itself: even the smallest change is a worker's job, because trivial is a guess and command attention does not scale.
-Merging, discarding work, and anything destructive, irreversible, or security-sensitive require the captain's explicit word.
-Standing autonomy is scoped consent granted per project, exercised only within the captain's original request, and it never quietly widens.
+Merging and discarding work always require the captain's explicit word for the concrete action. Destructive, irreversible, or security-sensitive actions do too, except for a deterministic host-maintenance capability the captain has explicitly granted standing autonomy and separately provisioned on that host; such a capability may perform only its documented action when all of its fail-closed gates hold.
+Standing autonomy is scoped consent granted per project or for one narrowly defined host-maintenance capability, exercised only within the captain's original request, and it never quietly widens.
 Evidence is never authorization: a diagnosis, a report, or a recommendation authorizes nothing by itself.
 Initiative beyond a stated request is legitimate only where the captain has committed a vision precise enough to adjudicate it, and even then only as an explicit opt-in.
 A current, explicit captain instruction outranks any standing rule the first mate wrote for itself, exactly as stated and no further.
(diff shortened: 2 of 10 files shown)


kunchenguid:
Closing: this is personal Mac mini host maintenance, not Firstmate shared template. Rebuilding as a standalone facility in private machine config. Do not merge.

PR #2963, opened 2026-08-24
fix(watch): hold the wedge clock open on evidence of real work

huynhtandat223:
## The defect

`bin/fm-watch.sh` escalated a stale pane as a possible wedge once `FM_STALE_ESCALATE_SECS` (240s) elapsed, even when the crew was provably working. That fixed timer is shorter than the work: a frontend gate measured at **311 seconds** crosses the threshold before it can finish, so the same worker false-alarmed on every cycle - 15 consecutive escalations over roughly 90 minutes in the observed case, each one a full supervision turn.

The predicate was never the problem. `crew_absorb_class` answers "is this crew working" correctly; the 240s cap overrode a TRUE answer.

## The change

The stale path now consults a second, structural source before firing that timer. `crew_work_evidence` (`bin/fm-classify-lib.sh`) reads two kernel facts about the task's recorded `worktree=`:

- a live process whose working directory is at or under it (`/proc/<pid>/cwd`);
- the newest mtime among the files git reports as changed there (`git --no-optional-locks status`, so the worker's index is never written).

A positive answer restarts the escalation clock instead of escalating. Nothing reads rendered output, a spinner, a keybind glyph, or a vendor process name.

## Bounds

- **Only positive evidence buys an extension.** No `worktree=` recorded, an unreadable path, a `git` that cannot answer: today's escalation stands. An unanswerable probe can never suppress an escalation.
- **One stale episode may spend at most `FM_WORK_EVIDENCE_MAX_SECS`** on extensions, measured from the first one and defaulting to `FM_BUSY_TURN_MAX_SECS` (3600s). A hung process that never exits therefore escalates within `FM_WORK_EVIDENCE_MAX_SECS + FM_STALE_ESCALATE_SECS` (~64 minutes) of the pane going quiet, and then keeps escalating on the ordinary 240s cadence with its escalation count and `demand-deep-inspection` marker intact.
- The spent window is bound to the stale episode's own hash, so a leftover record cannot silently disable the extension for a later episode.

## Cost

The probe runs only when an escalation is otherwise about to fire - at most once per `FM_STALE_ESCALATE_SECS` per stale window - and adds nothing to the per-wake path. Measured on Linux, 177 processes, a 526-file worktree: **2ms** for the process scan, **9ms** when it falls through to the changed-file read.

## Not covered

- **macOS has no procfs**, so the live-process source cannot answer there and only the changed-file signal applies; the fallback is today's behaviour, not a weaker one. No `lsof` sweep is attempted, because a blocking `lsof` inside the watcher's poll loop would trade a false alarm for a stall.
- **The away-mode daemon's own stale recheck is unchanged** (out of scope). It still escalates on the plain timer while `state/.afk` is active; the shared library now carries the probe if that path adopts it later.
- Evidence is presence, not progress: a process that is alive but doing nothing looks the same as one doing work. That is exactly what the outer bound is for.

## Verification

- `tests/fm-stale-work-evidence.test.sh` (new, `watcher-wake-lock` family): real processes, real mtimes, no harness. Covers the long gate not escalating, no-evidence still escalating, evidence going stale, the outer bound catching a hung worker and continuing to count, the unanswerable probe leaving behaviour unchanged, symlinked worktree paths, and episode binding. The two sources are driven apart deliberately - each positive case asserts what the verdict becomes when that one source is removed, so no case can pass vacuously on the other.
- `bin/fm-test-run.sh --family watcher-wake-lock`: 16/16 pass, including the existing wedge and daemon suites.
- `bin/fm-lint.sh` clean (shellcheck 0.11.0; actionlint absent locally, CI covers it).
- `bin/fm-doc-audience-check.sh` clean; `--check-coverage` clean.
- `tests/fm-test-run.test.sh` fails on this host for want of `ruby`, identically on a clean tree - pre-existing, unrelated.

The contract is patched at its existing owners: `bin/fm-watch.sh`'s header and `bin/fm-classify-lib.sh`'s header, with cross-references in `docs/architecture.md`, the new knobs in `docs/configuration.md`, and the new state file added to `AGENTS.md`'s watcher-internals listing.


huynhtandat223:
Closing - opened here by mistake. This targets a downstream fork's own default branch, not upstream; it was never intended as a contribution to this repository. Apologies for the noise.

PR #2942, opened 2026-08-24
fix(bin): bound remote job worker supervisor restarts

stanzhang:
## Intent

Fix upstream issue 2921 in Firstmate's bin/fm-remote-job-worker.sh. Add a per-account Linux supervisor singleton lock distinct from the serving-worker lock, reusing the established directory/PID/start/command staleness, quarantine, and symlink-refusal semantics. Supervisor startup must have three observable outcomes: started one supervisor, an already-running healthy supervisor is serving, or could not start; ensure/start callers may treat only the first two as success. Never release or reclaim a live healthy same-root incumbent. For a healthy same-root incumbent, wait a bounded interval for its ready identity; if it exits, start a replacement; if it remains live without becoming ready or releasing within the bound, fail observably. Release a live incumbent only when it cannot serve the requested root. Preserve existing stale/crashed supervisor recovery policy. Fix restart accounting so children dying just after the healthy threshold cannot restart without bound, without changing tunable defaults. Preserve abandoned-root checking, orphan grace, quarantine handling, TERM/HUP/INT traps, and exit statuses including 0 and 75. Do not redesign, add configuration, refactor unrelated code, or fix the unrelated fm-public-followup.sh failure. Add colocated tests proving duplicate supervisor refusal, stale/crashed recovery, bounded barely-healthy restart exhaustion, shutdown/start serving, and same-root concurrent-start incumbent survival. Keep existing tests passing and bin shellcheck-clean. Hard boundary: if review finds another start/ensure/singleton correctness failure on this corrected approach, stop and escalate rather than applying another correction. Deliver only through no-mistakes: push branch to https://github.com/stanzhang/firstmate.git and open a PR against kunchenguid/firstmate; never push origin and never merge. PR description must link issue 2921, explain the missing supervisor singleton and barely-healthy restart-accounting defects, and record two outliving findings: the pinned upstream lock cannot regenerate from its own manifest under strict npm, and Vite 8 runs on rolldown rather than Rollup.

## What Changed

- `bin/fm-remote-job-worker.sh`: the Linux supervisor now counts every failed child against `FM_REMOTE_JOB_SUPERVISOR_MAX_RESTARTS` as a lifetime total and checks it before the healthy-window branch, so a child that dies just past `FM_REMOTE_JOB_SUPERVISOR_HEALTHY_SECONDS` no longer resets the only guard and restarts without bound. Staying up for the healthy window still clears the consecutive-failure backoff, and no tunable defaults changed.
[...]

PR #2980, opened 2026-08-24
fix(composer): treat an NBSP-padded composer as empty, not pending

wmbryce:
## The bug

`fm_composer_classify_content` classifies a composer holding a prompt glyph followed by U+00A0 (non-breaking space) as `pending`, identically to real typed text.

```
$ source bin/fm-composer-lib.sh
$ NBSP=$(printf '\xc2\xa0')
$ fm_composer_classify_content 1 '❯'            -> empty
$ fm_composer_classify_content 1 "❯${NBSP}"     -> pending      # the bug
$ fm_composer_classify_content 1 '❯ '           -> empty
$ fm_composer_classify_content 1 '❯ hello'      -> pending
```

The glyph-stripping `case` matches `'❯ '*` (glyph plus ASCII space), so an NBSP-padded row falls through to the single-character `'❯'*` branch and leaves `\xC2\xA0` behind.
The following trim uses `[![:space:]]`, and bash's `[:space:]` does not match U+00A0 in the C/POSIX locale, so the NBSP survives, the row reads non-empty, and the verdict is `pending`.

## Why it matters

The away-mode daemon injects only on an affirmatively `empty` composer.
A Claude Code pane on the herdr backend renders its idle composer as `❯` followed by U+00A0, verified from a raw ANSI capture:

```
\e[38;2;153;153;153m❯\xC2\xA0
```

That pane can therefore never be injected into, so away-mode supervision is silently dead for it rather than degraded.

Observed on 2026-08-24: an overnight autonomous run stalled with **63,681 seconds (about 17.7 hours) of undelivered escalations**.
Three workers completed and nothing after them started, because the wake path into the supervised session could not be used.
`state/.supervise-daemon.log` repeated:

```
inject deferred: supervisor composer not confirmed-empty (state=pending: pending input, dead-shell prompt, or unreadable pane)
```

The wedge alarm fired correctly and was the only signal that anything was wrong.

## The fix

Treat Unicode whitespace as whitespace when judging composer emptiness, while keeping `pending` protective of real input.

A row of only invisible characters is empty.
Invisible characters around visible content remain `pending`, so a half-typed line is still protected:

```
fm_composer_classify_content 1 "❯${NBSP}"        -> empty
fm_composer_classify_content 1 "❯${NBSP}hello"   -> pending
fm_composer_classify_content 1 '❯ hello'         -> pending
fm_composer_classify_content 1 '❯'               -> empty
```

## Also fixed: the deferral log conflated two verdicts

`state=pending: pending input, dead-shell prompt, or unreadable pane` named three conditions in one string.
Per the classifier's own contract, `pending` means real unsubmitted text, while a dead shell or an unreadable pane is `unknown`.
The conflated message sent the first reader looking for a dead shell that did not exist.
Each verdict now prints only what it actually means.

## Tests

Extended `tests/fm-composer-lib.test.sh` rather than adding a runner.
Cases cover the bare glyph, glyph plus NBSP, glyph plus ASCII space, glyph plus NBSP plus real text, and a row of only invisible characters.
Tests exercise the public interface and assert no implementation bytes.

`bin/fm-lint.sh` passes.

### DIFF
--- a/bin/fm-composer-lib.sh
+++ b/bin/fm-composer-lib.sh
@@ -127,11 +127,21 @@ fm_composer_strip_ansi() {
 #   U+202F NARROW NO-BREAK SPACE      U+205F MEDIUM MATHEMATICAL SPACE
 #   U+3000 IDEOGRAPHIC SPACE
 # ASCII whitespace is absent because POSIX `[[:space:]]` already covers it.
-# U+200B ZERO WIDTH SPACE is deliberately absent: Unicode gives it
-# White_Space=No (a format character), so listing it would substitute this
-# owner's own guess for the property it claims to follow. The live harness
-# guard (bin/fm-test-run.sh, live-harness-optin) is what catches a harness
-# that starts drawing its composer with a character outside this property.
+# The zero-width format characters - U+200B ZERO WIDTH SPACE, U+2060 WORD
+# JOINER, U+FEFF ZERO WIDTH NO-BREAK SPACE - are all deliberately absent:
+# Unicode gives every one of them White_Space=No, so listing them would
+# substitute this owner's own guess for the property it claims to follow. Two
+# further reasons hold the line exactly there. Mapping them onto a SPACE would
+# be wrong in substance: this table substitutes rather than deletes precisely
+# because its members occupy width the harness drew, while a zero-width
+# character occupies none and a space would invent a separation that was never
+# rendered. And the failure directions are not symmetric: a row of zero-width
+# characters classifies `pending`, which only DEFERS an injection - the safe
+# direction, and one the max-defer alarm surfaces - whereas admitting an
+# unproven character widens `empty`, the direction that types an escalation
+# over a captain's half-written line. The live harness guard
+# (bin/fm-test-run.sh, live-harness-optin) is what catches a harness that
+# starts drawing its composer with a character outside this property.
 FM_COMPOSER_UNICODE_SPACES=()
 for _fm_composer_space_octal in \
   '\0302\0205' '\0302\0240' '\0341\0232\0200' \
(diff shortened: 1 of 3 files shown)


kunchenguid:
Speaking as Kun's firstmate: first look on current `main` `038d0f7ec6ba` (#2942). Thank you, wmbryce.

**class=corrective leftover.** Issue #1988 was closed already-fixed on 8/23 via #2102 (`fm_composer_normalize_trim_var` + matrix test). Verified on THIS main, not from metadata:

`fm_composer_classify_content 1` of glyph plus U+00A0 returns empty (ambient and LC_ALL=C). Glyph plus NBSP plus hello returns pending.

Main already maps U+00A0 (and the rest of Unicode White_Space outside ASCII) onto ASCII space before trim. `tests/fm-composer-lib.test.sh` `test_matrix_claude_bare_nbsp_row` already pins that idle Claude row as empty.

**This DIFF vs main:** `bin/fm-composer-lib.sh` is comments-only (zero-width White_Space=No rationale). Classifier logic is unchanged. Added tests re-pin existing behavior (FIGURE SPACE / NARROW NBSP / zero-width defer). `bin/fm-supervise-daemon.sh` splits the deferral log by verdict - a logging nit, not a remaining padding/path gap. Not a competing fix; closing rather than opening another PR.

**VISION.md:** closing **aligns**. Away-mode already injects on a genuinely empty NBSP-padded composer; authority/scripts/restart/spine/vendor/scope are n/a for a leftover close.

**This HEAD:** `6bef4f25e300b586119ce48fda4fb5943cdffc35`. MERGEABLE / UNSTABLE, ahead 1 / behind 0. No `no-mistakes-pipeline-attestation:v1`. CI action_required 32780664193 / 32780664023 - not approved (closing). First-time fork; no security issue in the diff.

Land-eligible rec: **NO** (already on main via #2102). Captain-flag NOW: **no**.

Closing as already fixed. This is not waiting-on-author and not a captain-decision.

kunchenguid:
Speaking as Kun's firstmate: closing as already fixed on main by #2102.

PR #2982, opened 2026-08-24
Add omni url to allowlist

CooperCode7:
Need the Omni URL in the allowlist so the Omni CLI can be used by the Omni plugins when working with dashboards through Firstmate.


CooperCode7:
Closing out. Merging into wrong repo

Issue #2883, opened 2026-08-23
Session lock is published under the shared Claude worker-pool daemon pid, so a sibling session in the same pool passes ownership


karotkriss:
## Summary

When a firstmate session acquires the session lock through Claude Code's reparented per-user worker pool, `bin/fm-lock.sh` publishes the **shared pool daemon's** pid rather than the acquiring session's own pid. Because ownership is decided by ancestry membership, any *other* Claude session served by that same per-user pool then passes the ownership test for that home. That breaks the exclusivity the lock exists to provide.

This is pre-existing behaviour on unmodified `main`. Filing it separately because it was surfaced during review of #2839, which is deliberately scoped to a different mechanism and is inert on this path.

## Reproduction path

1. `state/.lock` names a prior session that is now dead.
2. A new session starts with `CLAUDE_PID=700`, but its tool call is served from the pool, so its ancestry reaches only `320 -> 310 -> 300 -> 1` and never reaches 700.
3. `me=$(fm_harness_ancestry_pid)` therefore resolves to the shared pool daemon pid `300`.
4. The stale lock is replaced and publication records `300`.
5. A different Claude session served by that same per-user pool also has `300` in its ancestry, so `fm_session_lock_owned_by_self` returns true for it against the same home.

## Evidence that this is pre-existing

On unmodified `main`:

- `bin/fm-lock.sh` line 36 already derives the published identity purely from ancestry:
  ```sh
  me=$(fm_harness_ancestry_pid) || { echo "error: cannot locate harness process in ancestry" >&2; exit 1; }
  ```
  Under a pool reparented to init, that is the daemon pid, so `main` publishes `300` in this scenario.
- `bin/fm-session-lock-lib.sh` `fm_session_lock_owned_by_self` already decides ownership by ancestry membership of the recorded pid, so a sibling session under the same pool already satisfies it.

Both halves of the defect are therefore present without any change applied.

## Impact

Two live sessions sharing one per-user worker pool can each believe they own the same home's session lock. Lock-gated operations - spawn, steer, merge, wake drain, supervision repair - are then no longer serialized between them.

## Suggested direction

The identity that gets *written* needs resolving at the shared acquisition boundary, after the existing lock has been proven free or stale, rather than being taken from an ancestry walk that cannot see past the pool. Any fix should preserve the existing rule that a published session id alone must never let a caller claim a lock a different live session currently holds.

[...]

PR #2839, opened 2026-08-23
fix(bin): recognize Claude session locks across worker pools

karotkriss:
## Intent

Fix issue #2314: the session-lock self-ownership check fails under Claude Code's daemon-based Bash execution, producing a false "another session" refusal that forces the real primary session into permanent read-only mode (no spawn, steer, merge, wake-drain, or supervision repair).

Mechanism: fm_session_lock_owned_by_self() in bin/fm-session-lock-lib.sh decided "is this lock mine?" by walking the process ancestry of the current Bash tool invocation and requiring the lock file's recorded PID to appear in that chain. Newer Claude Code builds serve tool and hook commands from a persistent per-user worker pool that is reparented to init, so the ancestry terminates at PID 1 inside the pool and never reaches the interactive session PID that SessionStart correctly wrote into state/.lock. The same ancestry assumption reaches bin/fm-lock.sh, where the recorded live session PID is then reported as "another live firstmate session holds the lock (pid <same pid>)".

Approach and the reasoning behind it, so the diff is not read as accidental:

- Claude Code exports CLAUDE_PID (the interactive session's PID) into the environment of every tool and hook command it runs, including those served through the reparented pool. It is the identity that survives the PID-1 gap ancestry cannot cross. Claude Code is the only verified harness publishing such a variable today, so every other harness keeps the existing ancestry-only behaviour bit-for-bit unchanged.
- A new fm_harness_session_pid() helper returns that published PID and validates it twice: the PID must still be a live process, and that process must still be a Claude harness. A value inherited from an exited session whose PID has been recycled onto something else, or onto a different harness, is rejected rather than trusted.
- The evidence is used ONLY to WIDEN ownership, never to replace the ancestry test. This is a deliberate trust-boundary decision documented in the code: an environment variable is inherited by any child process, so on its own it proves "a Claude session named this PID", never "I am that session". The only conclusion drawn is that a lock ALREADY recording exactly that PID belongs to a live session rather than a competing one. The change therefore can never take a lock away from another session, can never turn an unheld lock into a held one, and can never introduce a NEW refusal - it can only remove a false one.
- Both shared ownership entry points route through the helper, because fixing only the function named in the issue title would not have fixed the reported reproduction.
[...]

PR #3023, opened 2026-08-25
fix(lock): acquire the session lock on Windows/Git Bash

seanofabob:
## Problem

On Windows / Git Bash (MSYS), the per-home session lock never acquires, so every session is forced into permanent read-only mode. Three layered platform defects combine to cause it:

1. **Harness-ancestry walk uses unsupported `ps` flags.** `fm_harness_ancestry_pids` reads the process tree with `ps -o comm= -p` and `ps -o ppid= -p`. The MSYS `ps` bundled with Git Bash rejects `-o` ("unknown option -- o"), so the walk fails on its first call and `fm_harness_ancestry_pid` never resolves the harness.

2. **Spawned scripts are orphaned from the harness.** MSYS emulates `fork()` through a helper that then exits, so a script launched as a child (session-start spawns `fm-lock.sh`) has a Win32 parent pid pointing at an already-dead launcher. No parent-chain walk can climb from it to `claude.exe`, even with a native process query.

3. **The symlink-based lock primitive silently copies.** `fm_lock_try_create` builds every lock with `ln -s`. MSYS's default `ln -s` copies the target instead of creating a symlink, so `fm_lock_points_to_owner` never validates and `fm_lock_acquire_wait` spins forever.

## Fix

- **Native Windows ancestry walk** (`fm-session-lock-lib.sh`): read the process tree from a single `Win32_Process` query and climb Windows PIDs, handing the same `(comm, args)` evidence to the existing `fm_harness_process_matches` contract.
- **`CLAUDE_PID` anchor**: Claude Code exports its own pid to every descendant; it survives the orphaning. Anchor the walk there when it names a live harness, falling back to the honest parent-chain walk when absent or stale.
- **Native symlinks** (`fm-wake-lib.sh`): request `winsymlinks:nativestrict` on Windows so `ln -s` creates a real NTFS symlink (needs Developer Mode / `SeCreateSymbolicLink`) instead of copying.

All Windows behavior is gated on `OSTYPE` / `FM_FORCE_WINDOWS`; the POSIX paths are byte-for-byte unchanged.

## Tests

- `tests/fm-session-lock-ancestry.test.sh`: new cases inject a deterministic Windows process table (`FM_WIN_PROCTABLE_CMD` / `FM_WIN_SELF_WINPID` / `FM_FORCE_WINDOWS`) and assert anchor resolution, orphan rescue, parent-walk resolution, fail-closed on an invalid/stale anchor, and that an unrelated harness never owns the lock. Existing POSIX cases are pinned to the POSIX path so they stay host-independent; the POSIX-`ps` e2e fixtures skip on hosts without `ps -o`.
- `tests/fm-wake-queue.test.sh`: asserts a Windows source requests native symlinks, preserves an operator's explicit policy, and is a no-op off Windows.

Verified on Windows 10 / Git Bash: the lock now acquires (`lock acquired: harness pid <n>`), holds idempotently, and reports correct status; the new tests pass, and shellcheck (0.11.0, LF-normalized) is clean.

### DIFF
--- a/bin/fm-session-lock-lib.sh
+++ b/bin/fm-session-lock-lib.sh
@@ -88,6 +88,128 @@ fm_harness_process_matches() {  # <comm> <args>
   return 1
 }
 
+# --- Windows (Git Bash/MSYS) native ancestry --------------------------------
+#
+# The POSIX walk below reads the process tree with `ps -o comm= -p` and climbs by
+# `ps -o ppid= -p`. Neither works under Git Bash/MSYS: its Cygwin `ps` rejects the
+# `-o` fields outright, and even where a process listing is available its parent
+# chain dead-ends at the MSYS boundary (the outermost shell reports ppid 1), so it
+# can never reach the native harness process such as claude.exe that actually owns
+# the session. The result was that fm_harness_ancestry_pid always failed and every
+# Windows session refused the fleet lock as read-only.
+#
+# On Windows these helpers walk the REAL Windows process tree instead, keyed on
+# Windows PIDs throughout, and hand the same (comm, args) evidence to
+# fm_harness_process_matches so the identity contract is exactly the one the POSIX
+# path uses. The current shell's Windows PID is read from /proc/<pid>/winpid, which
+# Git Bash exposes. The process table comes from a single Win32_Process query (a
+# stable WMI/kernel fact, not a release-note-fragile string), captured once per
+# invocation. FM_WIN_PROCTABLE_CMD and FM_WIN_SELF_WINPID override the two sources
+# so the walk is testable against a deterministic tree on any host, and
+# FM_FORCE_WINDOWS forces this path for that test.
+fm_is_windows() {
+  case "${FM_FORCE_WINDOWS:-${OSTYPE:-}}" in
+    1|msys*|mingw*|cygwin*) return 0 ;;
+  esac
+  return 1
+}
+
+# Print the current shell's Windows PID - the pid whose ancestry is walked.
+fm_win_self_winpid() {
+  local wp
+  if [ -n "${FM_WIN_SELF_WINPID:-}" ]; then
+    printf '%s\n' "$FM_WIN_SELF_WINPID"
+    return 0
+  fi
+  wp=$(cat "/proc/$$/winpid" 2>/dev/null) || return 1
+  case "$wp" in
+    ''|*[!0-9]*) return 1 ;;
+  esac
+  printf '%s\n' "$wp"
+}
+
+# Query the native process table once and emit it as
+# "<pid><TAB><ppid><TAB><name><TAB><commandline>" lines. Command-line tabs and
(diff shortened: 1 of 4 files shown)


kunchenguid:
Speaking as Kun's firstmate: first look on current `main` `85d6c7226c93` (#2953). Never messaged the captain.

**class=corrective.** Windows/Git Bash session lock never acquired: MSYS `ps` rejects `-o`, fork-helper orphaning breaks parent walk, default `ln -s` copies. Native Win32_Process walk + `CLAUDE_PID` anchor + `winsymlinks:nativestrict` when no MSYS policy is set. POSIX path unchanged. Standing lock hold: do not auto-land a third lock PR.

**VISION.md (inspected `bin/fm-session-lock-lib.sh` `fm_is_windows` / `fm_win_harness_ancestry_pids` / `fm_win_harness_anchor_pid`, `bin/fm-wake-lib.sh` MSYS symlink policy, `tests/fm-session-lock-ancestry.test.sh` / `tests/fm-wake-queue.test.sh`):**
- One captain, one interface: **aligns** (permanent read-only on Git Bash is machinery, not a captain-facing event).
- Authority is explicit: **aligns** (does not take an unheld lock; fail-closed on stale/non-harness `CLAUDE_PID`).
- Scripts own the mechanics: **aligns** (process table + identity contract).
- A restart is a non-event: **aligns** (lock can actually publish on Windows).
- Delegation with a spine: **aligns** (honest refusal still).
- The fleet outlives any vendor: **mixed** (`CLAUDE_PID` is Claude-published; other harnesses use the parent walk). Same family as #2839 / issue #2883.
- Scope: **aligns** for the Git Bash acquire gap.

**This HEAD:** `1a021442901f51c28257fc1e024810699207804e`. MERGEABLE / UNSTABLE, ahead 1 / behind 1.
Attestation **missing**. Greptile FAILURE (not a gate): existing non-strict `winsymlinks` policies are preserved, so `ln -s` can still copy and `fm_lock_acquire_wait` can retry without bound. Review note only; land hold is the lock family.

First-time fork (seanofabob has no other firstmate PRs; Behavior/Lint/CI had not run — Greptile-only does not count). Approved after full diff review. Workflows this pass: CI 32813622208, Require no-mistakes 32813622260. Not green at comment time.

Security: none that blocks CI approval. Local WMI process listing; test-only `FM_WIN_PROCTABLE_CMD` eval hook; no secrets/webhooks. `CLAUDE_PID` as lock identity is why the lock family stays a maintainer hold.

Overlap / standing hold: lock **#2437** (OPEN) / **#2839** (OPEN) / issue **#2883** (OPEN). Do not squash a third lock PR. Not spawn-freshen / herdr.

Land-eligible rec: **NO**. Captain-flag NOW: **no**.

This is a **captain-decision** hold on the lock family (which mechanism, if any), not waiting-on-author for that product call. Fork CI was approved this pass and is not yet green. Not a merge I will recommend.

PR #3025, opened 2026-08-25
feat(bin): capture Telegram messages into Mr Beanz

bingb0t5:
## Intent

Captain ruled 2026-08-24: Telegram is both capture and interrupt. The in-flight firstmate Telegram channel is interrupt/wake only; its brief forbids capture-into-the-brain. This task is the capture half only.

Messages from the captain, and later group discussion if that is turned on, are recorded in Mr Beanz. Beanz owns memory (people, decisions, promises, ideas, captured communication). Firstmate owns the capture path that writes into that brain. Do not take over Beanz product engineering.

Do not expand the current interrupt/wake adapter run. Do not hijack, rebase onto, or pile commits onto whatever Telegram interrupt PR is already in flight in another copy. Land capture as its own follow-up that composes with interrupt/wake, not as a rewrite of that work.

Delivery: no-mistakes, yolo off, captain fork bingb0t5/firstmate. Do not merge. Never weaken a validation gate to obtain a pass. Never add an agent name as a commit co-author. Destructive, irreversible, and security-sensitive actions still require the captain; Telegram posts are not automatic authority for those. Keep flag and wire mechanics in the owning script header. Follow one-owner and knowledge-placement rules.

Implemented as bin/fm-telegram-brain-capture.sh plus its engine: consume canonical Telegram message JSON from the interrupt adapter (or the same JSON lines), POST to Mr Beanz /v1/capture, receipt by update_id, captain DMs on by default, group discussion off until config/telegram-brain-capture-group is on. Never poll Telegram or advance getUpdates. A failed brain write must stop before Telegram ack. Captured text is recorded memory, not interrupt authority.

## What Changed

- Adds `bin/fm-telegram-brain-capture.sh` and its `bin/fm_telegram_brain_capture.py` engine: `capture` reads canonical Telegram message JSON lines (stdin or file) and POSTs each eligible message to Mr Beanz `POST /v1/capture`, `from-result` classifies one interrupt result via `fm-procevent-telegram.sh` and captures only a `message` result (the named `blocked`/`credential`/`protocol`/`transport-budget` kinds are zero-exit no-ops, an unrecognized kind is refused), and `doctor` reports non-secret readiness field by field. The path never polls Telegram or advances a getUpdates offset, and a failed brain write exits non-zero so the Telegram ack is withheld.
- Capture behavior: captain DMs are captured by default as source `firstmate-telegram`; negative chat ids are captured as `firstmate-telegram-group` only when local `config/telegram-brain-capture-group` (or `FM_TELEGRAM_BRAIN_CAPTURE_GROUP`) is `on`. Success is receipted per `update_id` under mode-0700 `state/telegram-brain-capture/`, so retries skip already-written messages. Credentials come solely from a mode-0600 `~/.config/beanz/mcp.env` (ambient `BEANZ_MCP_URL` ignored, token passed to curl via stdin config, no redirects); an entirely unconfigured home reports `capture-unconfigured` and exits zero. Unsupported lines and per-`update_id` refusals (disagreeing receipt, payload-scoped 4xx) fail that line and keep walking, while systemic failures (transport, 3xx/5xx, 401/403/404/405/429, bad credentials or receipt store) stop the batch and report `unattempted <count>`.
- Documents the path and its config in `docs/configuration.md`, `docs/scripts.md`, `AGENTS.md` (new config flag and state directory), and the `process-event-sources` skill, which now requires a zero-exit capture before acting on or acknowledging a Telegram result and states captured text is memory, not interrupt authority. Adds `tests/fm-telegram-brain-capture.test.sh` (~1.1k lines) covering these cases with a fake curl and stub adapter.

## Risk Assessment

✅ Low: The change is a self-contained, fail-closed new script plus engine with broad behavioral test coverage; the latest fix commit is narrow, matches the user's explicit round-12 instruction, and I verified its claimed behavior against the shipped CLI, leaving only an operator-facing count inaccuracy in a rare config-failure path and one stale documentation sentence.

## Testing

I ran the change's own behavior suite (50 cases) and the two repo-consistency suites its AGENTS.md and docs edits touch - all green - then, because passing unit tests do not show an operator that memory reaches the brain, built an end-to-end demonstration: a stand-in Mr Beanz serving POST /v1/capture over TLS, driven by the shipped CLI through the real curl binary, with only the brain and the not-yet-landed interrupt adapter substituted. The resulting transcript shows the whole capture story an end user would live through - readiness reporting, a captain DM captured with its text byte-exact (including non-ASCII and an embedded U+2028 that did not split the record), group discussion skipped until the config flag is on and then captured under its own source name, a non-captain private chat never captured, receipt-backed replay costing zero brain writes, a brain outage exiting non-zero with an unattempted count and no receipts so the Telegram ack stays withheld, the retry then capturing the whole batch, and a wedged receipt store or bad timeout refusing the batch before any POST. The final wire log confirms every outbound request went to the configured brain origin only, with no api.telegram.org traffic, no getUpdates or offset advance, and no bearer token in curl argv. This change has no UI, HTML, or rendered surface - it is a CLI plus an HTTP write path - so the reviewer-visible evidence is a CLI transcript and the brain-side request log rather than a screenshot. Overall result: everything passed and I found nothing actionable.

<details>
<summary>Evidence: End-to-end CLI transcript: Telegram payloads recorded in Mr Beanz over real TLS</summary>

Source: [End-to-end CLI transcript: Telegram payloads recorded in Mr Beanz over real TLS](https://github.com/bingb0t5/firstmate/blob/8913aee752db4a13be974575242ab5a6acbdfd16/.no-mistakes/evidence/fm/telegram-brain-capture/e2e-capture-transcript.txt)

```text

===== 1. doctor on a home that has never configured capture =====

$ fm-telegram-brain-capture.sh doctor
brain-env: missing
brain-url: unknown
captain-chat: configured
group-capture: off
receipts: absent

===== 2. the captain drops in the brain credential file, then re-checks readiness =====

$ fm-telegram-brain-capture.sh doctor
brain-env: present
brain-url: https://localhost:46811
captain-chat: configured
group-capture: off
receipts: absent

===== 3. capture a batch of canonical Telegram JSON lines from the interrupt adapter =====

$ cat batch.jsonl
{"chat_id": 4242, "date": 1756100000, "from_id": 909, "text": "Remember: the homestay partner signs Thursday - caf\u00e9 \u2615 deal is on", "update_id": 5001}
{"chat_id": -1001999, "date": 1756100000, "from_id": 909, "text": "group chatter about the roadmap", "update_id": 5002}
{"chat_id": 777001, "date": 1756100000, "from_id": 909, "text": "hello from a stranger DM", "update_id": 5003}
{"chat_id": 4242, "date": 1756100000, "from_id": 909, "text": "two paragraphs, one Telegram message:\u2028second paragraph", "update_id": 5004}

$ fm-telegram-brain-capture.sh capture batch.jsonl
captured 5001 beanz-0001
skipped:group 5002
skipped:private 5003
captured 5004 beanz-0002
exit=0

===== 4. what Mr Beanz actually received on the wire =====

$ cat beanz-received.jsonl
{
  "accepted": true,
  "authorization_ok": true,
  "body": {
    "source": "firstmate-telegram",
    "text": "Remember: the homestay partner signs Thursday - café ☕ deal is on"
  },
  "content_type": "application/json",
  "path": "/v1/capture"
}
{
  "accepted": true,
  "authorization_ok": true,
  "body": {
    "source": "firstmate-telegram",
    "text": "two paragraphs, one Telegram message: second paragraph"
  },
  "content_type": "application/json",
  "path": "/v1/capture"
}

===== 5. the captured text is byte-exact against the payload the adapter emitted =====
update 5001  bytes match: b'Remember: the homestay partner signs Thursday - caf\xc3\xa9 \xe2\x98\x95 deal is on'
update 5004  bytes match: b'two paragraphs, one Telegram message:\xe2\x80\xa8second paragraph'

===== 6. the durable receipt this home now holds =====

$ ls -la state/telegram-brain-capture && cat state/telegram-brain-capture/5001
total 16
drwx------ 2 rich rich 4096 Aug 25 13:15 .
drwx------ 3 rich rich 4096 Aug 25 13:15 ..
-rw------- 1 rich rich  183 Aug 25 13:15 5001
-rw------- 1 rich rich  183 Aug 25 13:15 5004

{"capture_id":"beanz-0001","captured_at":1787638520,"payload_sha256":"3d1c65225e3a5c792eb240e1816b4a9fcf06abd10c7b3cc2ed3a4a56514fafa9","source":"firstmate-telegram","update_id":5001}

===== 7. replay of the identical batch (the Telegram ack was lost) =====

$ fm-telegram-brain-capture.sh capture batch.jsonl
already-captured 5001 beanz-0001
skipped:group 5002
skipped:private 5003
already-captured 5004 beanz-0002
exit=0
brain writes during the replay: 0

===== 8. the captain turns group discussion on =====

$ printf 'on\n' > config/telegram-brain-capture-group && fm-telegram-brain-capture.sh doctor
brain-env: present
brain-url: https://localhost:46811
captain-chat: configured
group-capture: on
receipts: present

$ fm-telegram-brain-capture.sh capture batch.jsonl
already-captured 5001 beanz-0001
captured 5002 beanz-0003
skipped:private 5003
already-captured 5004 beanz-0002
exit=0

$ tail -1 beanz-received.jsonl   # group text lands under its own source name
{"accepted": true, "authorization_ok": true, "body": {"source": "firstmate-telegram-group", "text": "group chatter about the roadmap"}, "content_type": "application/json", "path": "/v1/capture"}

===== 9. the brain goes down mid-conversation: no ack, nothing lost =====

$ fm-telegram-brain-capture.sh capture outage.jsonl   # brain returns 500
error: brain capture failed with HTTP 500
unattempted 2
exit=1

$ ls state/telegram-brain-capture   # no receipt for any 6xxx payload
5001
5002
5004

===== 10. the retry that the withheld Telegram ack guarantees =====

$ fm-telegram-brain-capture.sh capture outage.jsonl
captured 6001 beanz-0004
captured 6002 beanz-0005
captured 6003 beanz-0006
exit=0

$ the memories Mr Beanz now holds, in the order it accepted them
firstmate-telegram         Remember: the homestay partner signs Thursday - café ☕ deal is on
firstmate-telegram         two paragraphs, one Telegram message:<U+2028>second paragraph
firstmate-telegram-group   group chatter about the roadmap
firstmate-telegram         decision: we ship the capture half first
firstmate-telegram         promise: I will call the accountant Friday
firstmate-telegram         idea: fold the digest into the morning brief

===== 11. an unusable receipt store refuses the whole batch before any brain write =====

$ chmod 0755 state/telegram-brain-capture && fm-telegram-brain-capture.sh doctor
brain-env: present
brain-url: https://localhost:43763
captain-chat: configured
group-capture: on
receipts: unreadable

$ fm-telegram-brain-capture.sh capture wedged.jsonl
error: receipt directory mode is not 700
unattempted 3
exit=1
brain writes attempted during the refusal: 0

===== 12. a bad capture timeout is refused in the same pre-flight =====

$ FM_BEANZ_CAPTURE_TIMEOUT=0 fm-telegram-brain-capture.sh capture wedged.jsonl
error: FM_BEANZ_CAPTURE_TIMEOUT must be a positive integer
unattempted 3
exit=1

===== 13. composing with the interrupt adapter: from-result =====

$ PATH=adapterbin:$PATH fm-telegram-brain-capture.sh from-result result.json
captured 8001 beanz-0007
exit=0

$ tail -1 beanz-received.jsonl
{"accepted": true, "authorization_ok": true, "body": {"source": "firstmate-telegram", "text": "from the interrupt result: book the flights"}, "content_type": "application/json", "path": "/v1/capture"}

$ every call this path made to the interrupt adapter
classify <work>/result.json
messages <work>/result.json

===== 14. every outbound request this path made, start to finish =====

$ cat wire.log   # only the configured brain origin
request-url: "https://localhost:<port>/v1/capture"
argv: -q -s -o <work>/home/state/telegram-brain-capture/.beanz-resp.ggc5i00l -w %{http_code} --max-time 30 -K -
request-url: "https://localhost:<port>/v1/capture"
argv: -q -s -o <work>/home/state/telegram-brain-capture/.beanz-resp.cb6oedyg -w %{http_code} --max-time 30 -K -
request-url: "https://localhost:<port>/v1/capture"
argv: -q -s -o <work>/home/state/telegram-brain-capture/.beanz-resp.wbpb6j46 -w %{http_code} --max-time 30 -K -
request-url: "https://localhost:<port>/v1/capture"
argv: -q -s -o <work>/home/state/telegram-brain-capture/.beanz-resp.0bml61dp -w %{http_code} --max-time 30 -K -
request-url: "https://localhost:<port>/v1/capture"
argv: -q -s -o <work>/home/state/telegram-brain-capture/.beanz-resp.q94ut57t -w %{http_code} --max-time 30 -K -
request-url: "https://localhost:<port>/v1/capture"
argv: -q -s -o <work>/home/state/telegram-brain-capture/.beanz-resp.pws8_v7p -w %{http_code} --max-time 30 -K -
request-url: "https://localhost:<port>/v1/capture"
argv: -q -s -o <work>/home/state/telegram-brain-capture/.beanz-resp.okwhwgq3 -w %{http_code} --max-time 30 -K -
request-url: "https://localhost:<port>/v1/capture"
argv: -q -s -o <work>/home/state/telegram-brain-capture/.beanz-resp._vf_fj_d -w %{http_code} --max-time 30 -K -

requests to api.telegram.org: 0
getUpdates / offset advances: 0
brain token anywhere in the logged curl argv: 0
```
</details>
<details>
<summary>Evidence: E2E driver used to produce the transcript</summary>

Source: [E2E driver used to produce the transcript](https://github.com/bingb0t5/firstmate/blob/8913aee752db4a13be974575242ab5a6acbdfd16/.no-mistakes/evidence/fm/telegram-brain-capture/e2e_capture_demo.sh)

```text
#!/usr/bin/env bash
# End-to-end demonstration of bin/fm-telegram-brain-capture.sh against a real
# HTTPS "Mr Beanz" that speaks POST /v1/capture. Real curl, real TLS, real
# sockets, real receipts on disk. Nothing about the capture path is mocked:
# only the brain itself and (in step 12) the interrupt adapter that is not in
# this branch are stand-ins.
#
#   usage: e2e_capture_demo.sh <repo-root> <evidence-dir>
set -uo pipefail
exec 2>&1                     # interleave the path's own error lines in place

REPO=${1:?repo root}
EV=${2:?evidence dir}
WORK=$(mktemp -d /tmp/fm-capture-e2e.XXXXXX)
SERVER_PID=""
stop_brain() { [ -n "$SERVER_PID" ] && kill "$SERVER_PID" 2>/dev/null; SERVER_PID=""; }
trap 'stop_brain; rm -rf "$WORK"' EXIT

CAPTURE="$REPO/bin/fm-telegram-brain-capture.sh"
TOKEN='tok_captain_brain_write_secret'
CAPTAIN_CHAT=4242
GROUP_CHAT=-1001999
STRANGER_CHAT=777001

hr() { printf '\n===== %s =====\n' "$1"; }
run() { printf '\n$ %s\n' "$*"; }

# --- a self-signed brain, so the https-only credential rule still holds ------
openssl req -x509 -newkey rsa:2048 -nodes -keyout "$WORK/key.pem" -out "$WORK/cert.pem" \
  -days 1 -subj '/CN=localhost' -addext 'subjectAltName=DNS:localhost' >/dev/null 2>&1

export BEANZ_LOG="$WORK/beanz-received.jsonl"
export BEANZ_EXPECT_TOKEN="$TOKEN"
export BEANZ_CERT="$WORK/cert.pem" BEANZ_KEY="$WORK/key.pem"
export BEANZ_PORT_FILE="$WORK/port"
: > "$BEANZ_LOG"

start_brain() {
  stop_brain
  rm -f "$BEANZ_PORT_FILE"
  BEANZ_MODE="$1" python3 "$EV/fake_beanz_brain.py" >"$WORK/brain.err" 2>&1 &
  SERVER_PID=$!
  for _ in $(seq 1 100); do [ -s "$BEANZ_PORT_FILE" ] && break; sleep 0.05; done
  PORT=$(cat "$BEANZ_PORT_FILE")
  [ -n "${FM_BEANZ_ENV_FILE:-}" ] && [ -f "${FM_BEANZ_ENV_FILE:-}" ] \
    && sed -i "s#BEANZ_MCP_URL=.*#BEANZ_MCP_URL=https://localhost:$PORT#" "$FM_BEANZ_ENV_FILE"
  return 0
}

# --- a curl shim that logs every outbound request, then execs the real curl --
mkdir -p "$WORK/shim"
cat > "$WORK/shim/curl" <<'SH'
#!/usr/bin/env bash
cfg=$(mktemp); cat > "$cfg"
sed -n 's/^url = /request-url: /p' "$cfg" >> "$WIRE_LOG"
printf 'argv: %s\n' "$*" >> "$WIRE_LOG"
exec /usr/bin/curl "$@" < "$cfg"
SH
chmod +x "$WORK/shim/curl"
export WIRE_LOG="$WORK/wire.log"
: > "$WIRE_LOG"
export PATH="$WORK/shim:$PATH"
export CURL_CA_BUNDLE="$WORK/cert.pem"

# --- a firstmate home that has never been bootstrapped ----------------------
HOME_DIR="$WORK/home"
mkdir -p "$HOME_DIR/config"        # note: no state/ - capture must bootstrap it
export FM_HOME="$HOME_DIR"
export FM_BEANZ_ENV_FILE="$WORK/mcp.env"
export FM_TELEGRAM_CAPTAIN_CHAT_ID="$CAPTAIN_CHAT"

start_brain ok

payload() { # update_id text chat_id
  python3 - "$@" <<'PY'
import json, sys
print(json.dumps({"update_id": int(sys.argv[1]), "text": sys.argv[2],
                  "chat_id": int(sys.argv[3]), "date": 1756100000,
                  "from_id": 909}, sort_keys=True))
PY
}

hr "1. doctor on a home that has never configured capture"
run "fm-telegram-brain-capture.sh doctor"
"$CAPTURE" doctor

hr "2. the captain drops in the brain credential file, then re-checks readiness"
umask 077
cat > "$FM_BEANZ_ENV_FILE" <<EOF
BEANZ_MCP_TOKEN=$TOKEN
BEANZ_MCP_URL=https://localhost:$PORT
EOF
chmod 600 "$FM_BEANZ_ENV_FILE"
run "fm-telegram-brain-capture.sh doctor"
"$CAPTURE" doctor

hr "3. capture a batch of canonical Telegram JSON lines from the interrupt adapter"
BATCH="$WORK/batch.jsonl"
{
  payload 5001 'Remember: the homestay partner signs Thursday - café ☕ deal is on' "$CAPTAIN_CHAT"
  payload 5002 'group chatter about the roadmap' "$GROUP_CHAT"
  payload 5003 'hello from a stranger DM' "$STRANGER_CHAT"
  payload 5004 "$(printf 'two paragraphs, one Telegram message: second paragraph')" "$CAPTAIN_CHAT"
} > "$BATCH"
run "cat batch.jsonl"
cat "$BATCH"
run "fm-telegram-brain-capture.sh capture batch.jsonl"
"$CAPTURE" capture "$BATCH"; printf 'exit=%d\n' $?

hr "4. what Mr Beanz actually received on the wire"
run "cat beanz-received.jsonl"
python3 - "$BEANZ_LOG" <<'PY'
import json, sys
for line in open(sys.argv[1], encoding="utf-8"):
    print(json.dumps(json.loads(line), ensure_ascii=False, indent=2, sort_keys=True))
PY

hr "5. the captured text is byte-exact against the payload the adapter emitted"
python3 - "$BATCH" "$BEANZ_LOG" <<'PY'
import json, sys
sent = {json.loads(l)["update_id"]: json.loads(l)["text"] for l in open(sys.argv[1], encoding="utf-8")}
got = [json.loads(l)["body"]["text"] for l in open(sys.argv[2], encoding="utf-8")]
for update_id, text in sorted(sent.items()):
    if text in got:
        print("update %d  bytes match: %r" % (update_id, text.encode("utf-8")))
PY

hr "6. the durable receipt this home now holds"
run "ls -la state/telegram-brain-capture && cat state/telegram-brain-capture/5001"
ls -la "$HOME_DIR/state/telegram-brain-capture"
echo
cat "$HOME_DIR/state/telegram-brain-capture/5001"; echo

hr "7. replay of the identical batch (the Telegram ack was lost)"
before=$(wc -l < "$BEANZ_LOG")
run "fm-telegram-brain-capture.sh capture batch.jsonl"
"$CAPTURE" capture "$BATCH"; printf 'exit=%d\n' $?
printf 'brain writes during the replay: %d\n' "$(( $(wc -l < "$BEANZ_LOG") - before ))"

hr "8. the captain turns group discussion on"
run "printf 'on\\n' > config/telegram-brain-capture-group && fm-telegram-brain-capture.sh doctor"
printf 'on\n' > "$HOME_DIR/config/telegram-brain-capture-group"
"$CAPTURE" doctor
run "fm-telegram-brain-capture.sh capture batch.jsonl"
"$CAPTURE" capture "$BATCH"; printf 'exit=%d\n' $?
run "tail -1 beanz-received.jsonl   # group text lands under its own source name"
tail -1 "$BEANZ_LOG"

hr "9. the brain goes down mid-conversation: no ack, nothing lost"
start_brain 500
OUTAGE="$WORK/outage.jsonl"
{
  payload 6001 'decision: we ship the capture half first' "$CAPTAIN_CHAT"
  payload 6002 'promise: I will call the accountant Friday' "$CAPTAIN_CHAT"
  payload 6003 'idea: fold the digest into the morning brief' "$CAPTAIN_CHAT"
} > "$OUTAGE"
run "fm-telegram-brain-capture.sh capture outage.jsonl   # brain returns 500"
"$CAPTURE" capture "$OUTAGE"; printf 'exit=%d\n' $?
run "ls state/telegram-brain-capture   # no receipt for any 6xxx payload"
ls "$HOME_DIR/state/telegram-brain-capture"

hr "10. the retry that the withheld Telegram ack guarantees"
start_brain ok
run "fm-telegram-brain-capture.sh capture outage.jsonl"
"$CAPTURE" capture "$OUTAGE"; printf 'exit=%d\n' $?
run "the memories Mr Beanz now holds, in the order it accepted them"
python3 - "$BEANZ_LOG" <<'PY'
import json, sys
for line in open(sys.argv[1], encoding="utf-8"):
    rec = json.loads(line)
    if rec["accepted"]:
        print("%-26s %s" % (rec["body"]["source"], rec["body"]["text"].replace("\u2028", "<U+2028>")))
PY

hr "11. an unusable receipt store refuses the whole batch before any brain write"
chmod 0755 "$HOME_DIR/state/telegram-brain-capture"
run "chmod 0755 state/telegram-brain-capture && fm-telegram-brain-capture.sh doctor"
"$CAPTURE" doctor
WEDGED="$WORK/wedged.jsonl"
{ payload 7001 'one' "$CAPTAIN_CHAT"; payload 7002 'two' "$CAPTAIN_CHAT"; payload 7003 'three' "$CAPTAIN_CHAT"; } > "$WEDGED"
before=$(wc -l < "$BEANZ_LOG")
run "fm-telegram-brain-capture.sh capture wedged.jsonl"
"$CAPTURE" capture "$WEDGED"; printf 'exit=%d\n' $?
printf 'brain writes attempted during the refusal: %d\n' "$(( $(wc -l < "$BEANZ_LOG") - before ))"
chmod 0700 "$HOME_DIR/state/telegram-brain-capture"

hr "12. a bad capture timeout is refused in the same pre-flight"
run "FM_BEANZ_CAPTURE_TIMEOUT=0 fm-telegram-brain-capture.sh capture wedged.jsonl"
FM_BEANZ_CAPTURE_TIMEOUT=0 "$CAPTURE" capture "$WEDGED"; printf 'exit=%d\n' $?

hr "13. composing with the interrupt adapter: from-result"
mkdir -p "$WORK/adapterbin"
cat > "$WORK/adapterbin/fm-procevent-telegram.sh" <<ADP
#!/usr/bin/env bash
# Stand-in for the in-flight interrupt/wake adapter, which is not in this
# branch. It answers the same two questions capture asks of it.
printf '%s %s\n' "\$1" "\$2" >> "$WORK/adapter.calls"
case "\$1" in
  classify) echo message ;;
  messages) cat "$WORK/result.messages.jsonl" ;;
  *) exit 2 ;;
esac
ADP
chmod +x "$WORK/adapterbin/fm-procevent-telegram.sh"
payload 8001 'from the interrupt result: book the flights' "$CAPTAIN_CHAT" > "$WORK/result.messages.jsonl"
echo '{"kind":"message"}' > "$WORK/result.json"
run "PATH=adapterbin:\$PATH fm-telegram-brain-capture.sh from-result result.json"
PATH="$WORK/adapterbin:$PATH" "$CAPTURE" from-result "$WORK/result.json"; printf 'exit=%d\n' $?
run "tail -1 beanz-received.jsonl"
tail -1 "$BEANZ_LOG"
run "every call this path made to the interrupt adapter"
sed "s#$WORK#<work>#g" "$WORK/adapter.calls"

hr "14. every outbound request this path made, start to finish"
run "cat wire.log   # only the configured brain origin"
sed -e "s#localhost:[0-9]*#localhost:<port>#g" -e "s#$WORK#<work>#g" "$WIRE_LOG"
printf '\nrequests to api.telegram.org: %d\n' "$(grep -c 'api\.telegram\.org' "$WIRE_LOG")"
printf 'getUpdates / offset advances: %d\n' "$(grep -c -i 'getupdates\|offset' "$WIRE_LOG")"
printf 'brain token anywhere in the logged curl argv: %d\n' "$(grep -c -- "$TOKEN" "$WIRE_LOG")"
```
</details>
<details>
<summary>Evidence: Stand-in Mr Beanz that speaks the real /v1/capture wire protocol over TLS</summary>

Source: [Stand-in Mr Beanz that speaks the real /v1/capture wire protocol over TLS](https://github.com/bingb0t5/firstmate/blob/8913aee752db4a13be974575242ab5a6acbdfd16/.no-mistakes/evidence/fm/telegram-brain-capture/fake_beanz_brain.py)

```text
#!/usr/bin/env python3
"""A stand-in Mr Beanz that speaks the real /v1/capture wire protocol over TLS.

Used only to produce end-to-end evidence: the capture path talks to this with
the real curl binary, so every request recorded here crossed a real socket.

Every capture attempt is appended to $BEANZ_LOG as one JSON line recording
whether it was accepted, the path, whether the bearer token matched, and the
decoded request body.
"""
import http.server, json, os, ssl

LOG = os.environ["BEANZ_LOG"]
MODE = os.environ.get("BEANZ_MODE", "ok")   # ok | 500
TOKEN = os.environ["BEANZ_EXPECT_TOKEN"]


def accepted_so_far():
    if not os.path.exists(LOG):
        return 0
    with open(LOG, encoding="utf-8") as fh:
        return sum(1 for line in fh if line.strip() and json.loads(line)["accepted"])


counter = {"n": accepted_so_far()}


class Handler(http.server.BaseHTTPRequestHandler):
    protocol_version = "HTTP/1.1"

    def log_message(self, *a):
        pass

    def do_POST(self):
        length = int(self.headers.get("Content-Length", "0"))
        raw = self.rfile.read(length)
        auth_ok = self.headers.get("Authorization", "") == "Bearer " + TOKEN
        routed = auth_ok and self.path == "/v1/capture"
        accepted = routed and MODE == "ok"
        record = {
            "accepted": accepted,
            "authorization_ok": auth_ok,
            "body": json.loads(raw.decode("utf-8")),
            "content_type": self.headers.get("Content-Type", ""),
            "path": self.path,
        }
        with open(LOG, "a", encoding="utf-8") as fh:
            fh.write(json.dumps(record, ensure_ascii=False, sort_keys=True) + "\n")
        if not routed:
            self.send_error(401)
            return
        if accepted:
            counter["n"] += 1
            payload = json.dumps(
                {"capture_id": "beanz-%04d" % counter["n"], "status": "captured"}
            ).encode("utf-8")
            self.send_response(200)
        else:
            payload = b'{"error":"brain is down"}'
            self.send_response(500)
        self.send_header("Content-Type", "application/json")
        self.send_header("Content-Length", str(len(payload)))
        self.end_headers()
        self.wfile.write(payload)


httpd = http.server.ThreadingHTTPServer(("127.0.0.1", 0), Handler)
ctx = ssl.SSLContext(ssl.PROTOCOL_TLS_SERVER)
ctx.load_cert_chain(os.environ["BEANZ_CERT"], os.environ["BEANZ_KEY"])
httpd.socket = ctx.wrap_socket(httpd.socket, server_side=True)
with open(os.environ["BEANZ_PORT_FILE"], "w") as fh:
    fh.write(str(httpd.server_address[1]))
httpd.serve_forever()
```
</details>
<details>
<summary>Evidence: Key excerpt: capture, receipts, outage, retry, and the outbound wire log</summary>

```text
$ fm-telegram-brain-capture.sh capture batch.jsonl
captured 5001 beanz-0001
skipped:group 5002
skipped:private 5003
captured 5004 beanz-0002
exit=0

--- what Mr Beanz received ---
{"accepted": true, "authorization_ok": true, "body": {"source": "firstmate-telegram", "text": "Remember: the homestay partner signs Thursday - café ☕ deal is on"}, "content_type": "application/json", "path": "/v1/capture"}

--- receipts on disk ---
drwx------ 2 rich rich 4096 state/telegram-brain-capture
-rw------- 1 rich rich 183 state/telegram-brain-capture/5001
{"capture_id":"beanz-0001","captured_at":1787638520,"payload_sha256":"3d1c6522...","source":"firstmate-telegram","update_id":5001}

--- replay after a lost Telegram ack ---
already-captured 5001 beanz-0001
already-captured 5004 beanz-0002
exit=0
brain writes during the replay: 0

--- brain outage: ack withheld, nothing lost ---
$ fm-telegram-brain-capture.sh capture outage.jsonl # brain returns 500
error: brain capture failed with HTTP 500
unattempted 2
exit=1

--- the retry the withheld ack guarantees ---
captured 6001 beanz-0004
captured 6002 beanz-0005
captured 6003 beanz-0006
exit=0

--- unusable receipt store: refuses the whole batch, posts nothing ---
receipts: unreadable
error: receipt directory mode is not 700
unattempted 3
exit=1
brain writes attempted during the refusal: 0

--- every outbound request, start to finish ---
request-url: "https://localhost:<port>/v1/capture" (x8)
requests to api.telegram.org: 0
getUpdates / offset advances: 0
brain token anywhere in the logged curl argv: 0
```
</details>

## Pipeline

_... (13 earlier update rounds omitted to keep the PR body within GitHub's 65536-char limit; full history is in the run log.)_

<details>

🔧 Fix: split payload-scoped 4xx from systemic, make from_id optional

- ⚠️ `bin/fm_telegram_brain_capture.py:338` - `http_failure_error` routes every 4xx except 401/403/429 to `CaptureError`, i.e. &#34;this one message was rejected, keep walking&#34;. 404 and 405 are not message-scoped: they say the endpoint does not exist at the configured base, which is systemic and identical for every payload. Concrete reachable path: the credential key is named `BEANZ_MCP_URL`, and the Beanz MCP endpoint actually registered on this machine is `https://brain.mrbea.nz/mcp` (verified in ~/.claude.json), so an operator putting that value in `~/.config/beanz/mcp.env` - the natural reading of the key name - makes `post_capture` build `https://brain.mrbea.nz/mcp/v1/capture`. Batch [9910, 9911, 9912]: line 420 raises `CaptureError` three times, `command_capture`&#39;s line 537 branch prints three anonymous `error: brain capture failed with HTTP 404` lines, `failed` is set, no `unattempted` is printed, exit 1. The operator sees what the header (bin/fm-telegram-brain-capture.sh:33-35) defines as &#34;a brain that rejects this one message&#34;, when in fact nothing is reachable at all; every payload is POSTed pointlessly on every retry, and the header&#39;s promise at lines 39-43 that a stop names the unattempted remainder never fires. Add 404 and 405 (and arguably 408) to `SYSTEMIC_HTTP_STATUSES` at line 38 so an unreachable or wrong endpoint stops the batch once and is diagnosed as systemic, matching the rule the header and docs/configuration.md already state.
- ⚠️ `bin/fm_telegram_brain_capture.py:537` - Now that a `CaptureError` fails only its own line and the batch keeps walking, stderr is the only record of which payload failed - and it does not identify it. Six of the seven `CaptureError` messages carry no update_id (lines 267, 281, 283, 285, 289, 291) and neither does the HTTP path at line 420; only the receipt-disagreement at line 469 does. The `except CaptureError` branch at 537-539 prints `error: &lt;message&gt;` to stderr and writes nothing to stdout, so the failed update_id appears in no output stream and no receipt. Concrete sequence: a backup restore leaves `state/telegram-brain-capture/9920` at mode 0644. Batch [9920, 9921, 9922] runs; `read_receipt` line 285 raises, stdout gets `captured 9921 cap-x` and `captured 9922 cap-y`, stderr gets a bare `error: receipt mode is not 600`, exit 1. The run will never exit 0 - so Telegram is never acked and the same result is re-announced forever - until an operator performs the recovery the header documents at bin/fm-telegram-brain-capture.sh:103-105, &#34;removes state/telegram-brain-capture/&lt;update_id&gt;&#34;, using an update_id the program never printed. With more than one failing line the messages are also indistinguishable from each other. Include the update_id in every per-line refusal (pass it into the receipt helpers, or wrap the raise site in `capture_line` which already has it) so each failed line is attributable, e.g. `error: 9920 receipt mode is not 600`.
- ⚠️ `bin/fm_telegram_brain_capture.py:419` - `post_capture` accepts exactly status 200; `if status != 200` sends everything else through `http_failure_error`, and any 2xx or 3xx that is not 200 falls into the `UserError` (systemic) bucket. A create endpoint that answers `201 Created` or `202 Accepted` with a perfectly good `{&#34;capture_id&#34;: ...}` body is therefore read as a systemic brain failure: line 420 raises `UserError`, line 540 breaks the batch, `unattempted &lt;n&gt;` is printed, exit 1, no receipt is written, and because the exit is non-zero the Telegram result is never acked - so every subsequent run repeats the same POST and the same stop. The memory is written into Mr Beanz on every attempt while firstmate records it as failed, and the interrupt is wedged indefinitely with no operator recovery documented. This also contradicts the header&#39;s own enumeration at bin/fm-telegram-brain-capture.sh:36-38, which lists the systemic set as transport failure, 5xx, 429, 401/403, credentials, and the receipt store - a 201 is none of those. Nothing in this checkout pins Mr Beanz&#39;s response code for `POST /v1/capture`, and every test asserts against a fake curl that returns 200, so the contract is assumed rather than verified. Flagged ask-user because widening the accepted status set is a decision about an external service&#39;s wire contract; the low-risk form is to accept any 2xx whose body yields a valid capture_id and keep every non-2xx classified as it is now.
- ℹ️ `bin/fm_telegram_brain_capture.py:440` - `classify_chat` returns `CAPTAIN_SOURCE` on `chat_id == captain_chat` before it ever consults `group_on`, and `captain_chat_id` deliberately accepts a negative value (`re.fullmatch(r&#34;-?[1-9][0-9]*&#34;, ...)` at lines 197 and 202). If `TELEGRAM_CAPTAIN_CHAT_ID` is set to a group or supergroup id - the shared `~/.config/beanz/telegram.env` is also the interrupt adapter&#39;s file, and a bot addressed in a group is a normal setup - then every message in that group is captured while `config/telegram-brain-capture-group` is off, and each is written to the brain under source `firstmate-telegram` rather than `firstmate-telegram-group`. That is a wrong label and a wrong set, produced without any error: it contradicts the header at bin/fm-telegram-brain-capture.sh:62-68 (&#34;Group discussion means a Telegram group, supergroup, or channel, which is exactly a negative chat_id&#34; / &#34;off until config/telegram-brain-capture-group contains the bare word `on`&#34;) and the same claim in docs/configuration.md. Not live on this machine - the configured id is a positive 10-digit user id - so this is a latent misconfiguration hazard rather than a current defect. Either reject a negative captain chat id at line 197/202 as not being a direct message, or make `classify_chat` apply the group gate and the group source whenever `chat_id &lt; 0` regardless of whether it matches the captain id. Flagged ask-user because which of those two is correct depends on whether a group is ever meant to serve as the captain channel.

🔧 Fix: accept 2xx, treat 404/405 systemic, attribute per-line refusals
1 warning still open:

- ⚠️ `bin/fm_telegram_brain_capture.py:535` - `command_capture` reads the whole payload batch with `raw = sys.stdin.read()`, which decodes using `sys.stdin.encoding` (the ambient locale, or `PYTHONIOENCODING`). Every other read in this module pins the codec explicitly - `read_env_file` at line 118 (`b&#34;&#34;.join(chunks).decode(&#34;utf-8&#34;)`), `group_capture_enabled` at line 221, `read_receipt` at line 287 - and the script header at bin/fm-telegram-brain-capture.sh:56-57 makes byte-exact text fidelity an invariant, tested at tests/fm-telegram-brain-capture.test.sh:494-513 for U+2028/U+2029/U+0085. This one unpinned decode breaks it silently. Concrete trace, verified on this machine: with `PYTHONIOENCODING=latin-1` (or any non-UTF-8 locale) in the wake handler&#39;s environment, a line `{&#34;update_id&#34;:9401,&#34;chat_id&#34;:4242,&#34;text&#34;:&#34;caf\xc3\xa9 wh\xc4\x81nau&#34;}` decodes as latin-1 into `café whÄ\x81nau`. That string is still valid JSON, `parse_payload` accepts it (nonempty text, integer chat_id), `classify_chat` returns CAPTAIN_SOURCE, `payload_hash` digests the corrupted text, `post_capture` writes it to Mr Beanz where raw captures are immutable, `write_receipt` records the corrupted digest, stdout prints `captured 9401 cap-x`, and the run exits 0 - so Telegram is acked and the receipt guarantees the message is never re-captured. Nothing errors at any step; the memory is permanently wrong. The same line also raises a bare `UnicodeDecodeError` (Python traceback, not a `UserError`) for input the ambient codec rejects outright, which the suite already treats as a defect elsewhere (tests/fm-telegram-brain-capture.test.sh:395 asserts no traceback for the unreadable group flag). Fix at the same boundary: read `sys.stdin.buffer.read()` and decode with an explicit `.decode(&#34;utf-8&#34;)`, raising `UserError(&#34;payload input is not UTF-8: ...&#34;)` on failure so the batch is fail-closed rather than corrupted or traced back.

🔧 Fix: read payload batch as UTF-8, not ambient encoding
4 infos still open:

- ℹ️ `bin/fm_telegram_brain_capture.py:597` - `command_doctor` calls `group_capture_enabled(config)` inline at line 597, and that helper raises `UserError` for a flag file that is a symlink, a non-regular file, unreadable, or not UTF-8 (lines 218-223). The exception escapes to `main`, which prints `error: ...` and returns 1 *before* line 603 prints anything. So the one command whose documented job is &#34;doctor reports non-secret readiness&#34; (bin/fm-telegram-brain-capture.sh:50) reports nothing at all when a single unrelated flag file is broken. Concrete: `chmod 000 config/telegram-brain-capture-group`, then `fm-telegram-brain-capture.sh doctor` prints only `error: cannot read telegram-brain-capture-group: ...` on stderr and no `brain-env:`, `brain-url:`, `captain-chat:`, or `receipts:` line - this is exactly what tests/fm-telegram-brain-capture.test.sh:398 exercises, and it only asserts the stderr line, so the missing readiness report is not caught. The asymmetry is local to this function: `brain_credentials()` and `captain_chat_id()` are both wrapped in `try/except UserError` at lines 582-595 and degrade to `unreadable`, while the group flag is the only probe allowed to kill the whole report. Give it the same treatment - wrap line 597 and emit `group-capture: unreadable` - so a broken flag costs one field instead of the entire diagnosis. Flagged ask-user because it adds a new value to doctor&#39;s documented output vocabulary.
- ℹ️ `bin/fm_telegram_brain_capture.py:581` - `command_doctor` initialises `brain_url = DEFAULT_BRAIN_URL` at line 581 and only overwrites it inside the `try` at line 583. When `brain_credentials()` raises - which line 586 explicitly treats as the `unreadable` case, i.e. the file exists but has the wrong mode, no token, an invalid key, or an unsafe `BEANZ_MCP_URL` - `brain_url` keeps the hardcoded default and line 604 prints it as fact. Concrete: `~/.config/beanz/mcp.env` at mode 0644 containing `BEANZ_MCP_TOKEN=...` and `BEANZ_MCP_URL=https://staging.brain.test`. Doctor prints `brain-env: unreadable` followed by `brain-url: https://brain.mrbea.nz`, and exits 0. The operator diagnosing why capture fails reads a destination that is neither the one configured nor the one that would be used (no capture happens at all in this state), with nothing erroring to signal that the value was never sourced. This also contradicts docs/configuration.md&#39;s &#34;That mode-0600 file alone owns the brain destination&#34;: doctor is asserting a destination the file did not supply. Print `brain-url: unknown` (or suppress the line) whenever `brain_state` is `unreadable`, and keep the default only for `missing`, where the default genuinely is what a fixed home would use. Flagged ask-user because it changes doctor&#39;s user-visible output contract.
- ℹ️ `bin/fm_telegram_brain_capture.py:551` - The script header states the batch rule as a pair: &#34;Only a systemic failure stops the batch: a transport failure, any 3xx or 5xx, HTTP 429, 401, 403, 404 or 405, an unusable credential or config, or a receipt store this home cannot use&#34; and then &#34;A stop prints `unattempted &lt;count&gt;`&#34; (bin/fm-telegram-brain-capture.sh:38-44); docs/configuration.md:620 repeats it. Only the in-loop `except UserError` at lines 565-573 honours the second half. `brain_credentials()` (551), `captain_chat_id()` (552), `group_capture_enabled()` (553), and `decode_payload_input()` (550) all run before the loop, so their `UserError` propagates past `command_capture` to `main`&#39;s handler at line 625, which prints `error: ...` and returns 1 with no `unattempted` line. Concrete: a batch of three captain payloads with `~/.config/beanz/mcp.env` at mode 0644 prints `error: /home/.../mcp.env mode is not 600` and exits 1 - stdout is empty, so an operator reading the documented contract sees no pending count and can reasonably read the absence as &#34;nothing was left behind&#34;, when in fact all three payloads are unwritten. tests/fm-telegram-brain-capture.test.sh:241-254 covers this case but only asserts the stderr line, so the gap is untested either way. Either count the whole batch in the pre-flight path or narrow the header and docs sentence to a mid-batch stop; the doc-only fix is the low-risk one. Flagged ask-user because which side moves is an authorial call about the output contract.
- ℹ️ `bin/fm_telegram_brain_capture.py:258` - `ensure_receipt_dir` creates the receipt directory with `os.mkdir(str(receipts), 0o700)` at line 258 and returns immediately, but `mkdir(2)` masks its mode argument with the process umask, so the directory&#39;s real mode is `0o700 &amp; ~umask`. On the very next run line 254 re-checks `stat.S_IMODE(info.st_mode) != 0o700` and raises `UserError(&#34;receipt directory mode is not 700&#34;)` - a systemic failure that stops every batch and, because the exit is non-zero, means the Telegram result is never acked and the same failure repeats forever with no recovery documented anywhere in the header&#39;s RECEIPTS section. Concrete: a wake handler running under `umask 0200` creates the directory at mode 0500; the first capture in that run succeeds and receipts, and every later run refuses the whole batch. Every other mode-sensitive write in this module is already immune because it sets the mode explicitly after creation - `write_receipt` uses `os.fchmod(fd, 0o600)` at line 301 and `post_capture` at lines 353 and 356 - so this is the one place that trusts the umask. Add `os.chmod(str(receipts), 0o700)` after the mkdir (or create then chmod) so the directory the code just made always satisfies the check the code applies to it. Reachability is low - a umask with owner bits set is unusual - but the failure is permanent and the fix is one line consistent with the rest of the module.

🔧 Fix: degrade doctor fields, count preflight stops, fix receipt dir mode
2 infos still open:

- ℹ️ `bin/fm_telegram_brain_capture.py:615` - `command_doctor` labels the receipt store with `receipts.is_dir() and not receipts.is_symlink()` (line 615), a test that disagrees with the one `ensure_receipt_dir` actually enforces, so doctor reports readiness for a store capture will refuse. Concrete: `ln -s /tmp/elsewhere state/telegram-brain-capture` makes `is_dir()` True (it follows the link) and `is_symlink()` True, so doctor prints `receipts: absent` and exits 0; the next capture dies with `receipt directory is unsafe` from line 253. Symmetrically, `chmod 750 state/telegram-brain-capture` makes doctor print `receipts: present` while every capture stops with `receipt directory mode is not 700` (line 255). This is exactly the misleading-readiness class this fix round just eliminated for `brain-url` (line 594) and `group-capture` (lines 610-613); the receipts field is the one probe still answering a different question than capture asks. Compounding it, the header&#39;s RECEIPTS section documents recovery for a mismatched receipt (bin/fm-telegram-brain-capture.sh:126-128, `rm state/telegram-brain-capture/&lt;update_id&gt;`) but names no recovery for a wedged directory, so the operator gets a green field plus a permanent systemic stop with no printed way out. Reusing `ensure_receipt_dir`&#39;s own lstat+mode check here, adding an `unusable` value the way the other fields got `unreadable`, would make doctor&#39;s answer the same answer capture gives, and one header line would name the `chmod 700` recovery. Flagged ask-user because it adds a value to doctor&#39;s documented output vocabulary at bin/fm-telegram-brain-capture.sh:59.
- ℹ️ `bin/fm_telegram_brain_capture.py:525` - When the new preflight stop fires before `captain_chat_id()` resolves, `pending_post_count` counts every parseable payload without a receipt, including ones the code would never have posted. Concrete: `chmod 000 config/telegram-brain-capture-group`, then feed a batch of one third-party DM (`chat_id: 777001`). `group_capture_enabled` raises first (line 565), so `captain_chat` is still None; the guard at line 525 skips classification entirely and line 532 counts the payload. Output is `unattempted 1`, but a private chat that is not the captain&#39;s is never captured on or off flag (bin/fm-telegram-brain-capture.sh:84) - on retry that same line prints `skipped:private`, so the true pending count is 0. This contradicts the header&#39;s own wording at bin/fm-telegram-brain-capture.sh:43-44 (&#34;counting the payloads behind it that would have been posted&#34;) and the contract the `unattempted-count` case at tests/fm-telegram-brain-capture.test.sh:771-795 asserts for the mid-batch path; the new preflight test at line 863 does not reach it because its stop is `brain_credentials`, by which point `captain_chat` is known. The error is conservative (over-count, never under-count), and the `if captain_chat is not None:` guard reads as a deliberate choice to count everything when classification is impossible, so this is recorded as an accepted tradeoff rather than a change to make: resolving `captain_chat_id()` before `group_capture_enabled()` would let the flag-independent private-chat exclusion apply, but group payloads would then need a tri-state `group_on` to avoid the worse opposite error of under-counting.

🔧 Fix: judge receipt store with one shared predicate

- ⚠️ `bin/fm_telegram_brain_capture.py:393` - `timeout_seconds()` is validated lazily inside `post_capture` (line 393) instead of in the preflight block the fix round added at lines 578-581, so an invalid `FM_BEANZ_CAPTURE_TIMEOUT` is a config failure that surfaces mid-loop and under-counts `unattempted`. Verified by running the shipped CLI: a batch of three captain payloads with `FM_BEANZ_CAPTURE_TIMEOUT=0` printed `error: FM_BEANZ_CAPTURE_TIMEOUT must be a positive integer` followed by `unattempted 2` and exited 1, although zero payloads reached the brain. The `except UserError` handler at line 596 reports `lines[index + 1:]`, which is correct for a failure that happened after the POST but wrong here because the current line was never attempted either. This contradicts the header&#39;s own contract at bin/fm-telegram-brain-capture.sh:46 (&#34;A credential or config failure stops before the first write and counts the whole batch the same way&#34;) and docs/configuration.md&#39;s &#34;a stop reports `unattempted &lt;count&gt;` for the payloads behind it whether it happened mid-batch or before the first write&#34;, and it is the under-count direction the round-10 note called the worse error. The same class reaches `ensure_receipt_dir` when the store is wedged, but the timeout is the reachable-by-env case. Earliest shared boundary: call `timeout_seconds()` inside the existing preflight `try` at line 578 alongside `brain_credentials()`, so every config read is validated before the loop and the whole batch is counted. Note this makes an invalid timeout exit non-zero even for an all-skipped batch, matching how `brain_credentials()` already behaves.
- ℹ️ `bin/fm_telegram_brain_capture.py:247` - `inspect_receipt_dir` maps any `lstat` failure on the state directory to `(&#34;unreadable&#34;, &#34;state directory is unreadable: %s&#34;)`, including plain ENOENT. `state/` is gitignored (.gitignore line 2), so a checkout that has not been through `bin/fm-bootstrap.sh` (which mkdirs it at line 1043) has no `state/`. Verified by running the shipped CLI against such a home: `doctor` printed `brain-env: present / captain-chat: configured / group-capture: off / receipts: unreadable`, and `capture` exited 1 with `error: state directory is unreadable: [Errno 2] No such file or directory: &#39;.../state&#39;`. Two consequences. First, this is a label regression from commit a2ba4d4: `command_doctor` previously printed `absent` for this home, so the readiness command doctor exists to serve now reports the one field that needs no operator repair as if the store were wedged, next to `missing`/`off` on fields that are genuinely just unset. Second, capture never self-heals: no ack means retry, and every retry hits the same refusal, while the header&#39;s new recovery line (bin/fm-telegram-brain-capture.sh:128-129) only names `chmod 700` for a real-but-wrong directory and the error text never tells the operator to mkdir. Every sibling script that owns state does `mkdir -p &#34;$STATE&#34;` first (bin/fm-lease.sh:47, bin/fm-afk-return.sh:78, bin/fm-procevent-remote-reply.sh:358); this one alone requires the parent to pre-exist. Either create the missing parent in `ensure_receipt_dir` (keeping the symlink and non-directory refusals by lstat-ing first) so the store bootstraps itself, or classify ENOENT on the state directory as `absent` so doctor stops calling a fresh home broken. Flagged ask-user because it moves both doctor&#39;s documented output vocabulary and whether capture may create `$FM_HOME/state`. Note tests/fm-telegram-brain-capture.test.sh&#39;s `make_home` always mkdirs `state`, so no case reaches this path in either direction.
- ℹ️ `bin/fm-telegram-brain-capture.sh:35` - A 4xx outside SYSTEMIC_HTTP_STATUSES fails one line and sets `failed = True` (bin/fm_telegram_brain_capture.py:593-595), so the run exits non-zero and, per the skill guidance added in .agents/skills/process-event-sources/SKILL.md, the Telegram result must not be acknowledged. For a transient 4xx that is right. For a permanent one it wedges: if Mr Beanz rejects a captain message with, say, 422 because the text exceeds a server-side limit, `capture_payload` writes no receipt (line 507 raises before line 509), so every replay re-posts the same text, gets the same 422, and exits non-zero again. The other payloads in the batch return `already-captured`, so nothing else is lost, but the interrupt result stays permanently unackable. The header documents an operator lever for the only other permanent per-line refusal - a disagreeing receipt, recovered by `rm state/telegram-brain-capture/&lt;update_id&gt;` (lines 130-132) - and names none for this one, and there is no way to mark a payload as deliberately not capturable. Either name a recovery in the RECEIPTS section (for example writing a receipt that records the refusal so replay stops re-posting) or state explicitly that a permanently-rejected payload requires a brain-side fix. Flagged ask-user because the fail-closed exit is exactly what the intent requires (&#34;A failed brain write must stop before Telegram ack&#34;), so only the author can decide what the escape hatch should be.

🔧 Fix: preflight timeout, bootstrap state dir, document 4xx recovery
1 warning still open:

- ⚠️ `bin/fm_telegram_brain_capture.py:505` - A receipt-store refusal still under-counts `unattempted` by one, because `ensure_receipt_dir(state)` is called per-payload at line 505 rather than as a batch-scoped precondition. When it raises, the loop&#39;s `except UserError` handler at line 614 reports `lines[index + 1:]`, which assumes the current line was attempted - true for a POST that failed, false here, since the store check happens before any POST for that line. Reproduced with the shipped CLI: a batch of three captain payloads against a `state/telegram-brain-capture` at mode 0755 printed `error: receipt directory mode is not 700`, then `unattempted 2`, and exited 1, although the fake curl was never invoked and all three payloads were unattempted. The same path is reachable through a symlinked or non-directory store, a `state/` this home cannot create or write, and `post_capture`&#39;s pre-POST `mkstemp` staging failure. This is the exact class the round-11 finding flagged (&#34;The same class reaches `ensure_receipt_dir` when the store is wedged&#34;); commit 719ca64 closed only the `timeout_seconds()` half, so the under-count direction round 10 called the worse error survives. It contradicts the owning header at bin/fm-telegram-brain-capture.sh:38-45, which lists &#34;a receipt store this home cannot use&#34; among the systemic stops and then promises &#34;A credential or config failure stops before the first write and counts the whole batch the same way&#34;, and docs/configuration.md&#39;s &#34;a stop reports `unattempted &lt;count&gt;` for the payloads behind it whether it happened mid-batch or before the first write&#34;. Impact is a misleading operator count only: the ack is still correctly withheld by the non-zero exit and the retry re-walks the batch, so no memory is lost. Existing coverage cannot catch it - the mode-750 and symlinked-store cases at tests/fm-telegram-brain-capture.test.sh:949-971 assert the error text and never the count. Earliest supported boundary that keeps current semantics: distinguish a failure raised before this line&#39;s POST from one raised after it (for example an `attempted` flag on the raised error, or a dedicated pre-POST error type) and report `lines[index:]` for the pre-POST case. Do not simply hoist `ensure_receipt_dir` into the preflight block at lines 588-592: that would make a batch containing only skipped payloads exit non-zero on a wedged store, where it correctly exits 0 today because `classify_chat` returns before the store is ever touched.

🔧 Fix: resolve receipt store once in capture preflight
3 infos still open:

- ℹ️ `bin/fm_telegram_brain_capture.py:588` - `group_capture_enabled(config)` is the first call in the preflight `try`, so when it raises, `captain_chat` is still `None` at line 595 and `pending_post_count` skips classification entirely, counting every parseable payload including ones that could never be posted under any flag value. Verified with the shipped CLI: with `config/telegram-brain-capture-group` at mode 000 and a batch of one captain message (chat 4242) plus two third-party private DMs (chat 777001, 777002), the run printed `error: cannot read telegram-brain-capture-group: [Errno 13] Permission denied` followed by `unattempted 3` and exited 1, although only one payload was ever capturable. `classify_chat` skips a non-captain chat_id &gt;= 0 as `private` unconditionally, independent of `group_on`, so that portion of the count is knowable even when the flag is unreadable. This contradicts the owning header at bin/fm-telegram-brain-capture.sh:43-44 (&#34;counting the payloads behind it that would have been posted, not the input lines it stopped reading&#34;). Impact is an operator-facing count only, in the conservative direction: the non-zero exit still withholds the Telegram ack and the retry re-walks the batch, so no memory is lost. Earliest shared boundary: resolve `captain_chat_id()` before `group_capture_enabled(config)` in the preflight so the unconditional private-chat skip is honored, and let `pending_post_count` treat an unresolved group flag as &#34;count negative chat_ids too&#34; (for example a tri-state `group_on=None`) so fixing the over-count does not introduce an under-count for genuine group payloads when the flag was meant to be `on`.
- ℹ️ `docs/configuration.md:623` - &#34;A home that has never bootstrapped needs no hand setup: capture creates `state/` and its receipt store at mode 0700 on the first write.&#34; Commit eb195ba moved `ensure_receipt_dir(state)` into the preflight block (bin/fm_telegram_brain_capture.py:592), so the store is now created on every `capture` run before any payload is classified or posted, not on the first write. Verified with the shipped CLI: an empty batch on a home with no `state/` created `state` and `state/telegram-brain-capture` both at mode 700 and exited 0 without invoking curl. The same commit updated the owning script header (bin/fm-telegram-brain-capture.sh:48-51) to describe the new pre-flight resolution but left this sentence behind, so the operator doc now states a trigger the code no longer has, and a later run of `doctor` reports `receipts: present` on a home that has never captured anything.
- ℹ️ `.agents/skills/process-event-sources/SKILL.md:88` - The skill now instructs agents to run `bin/fm-telegram-brain-capture.sh from-result &lt;result-file&gt;` and require a zero exit before acknowledging a Telegram result, but `bin/fm-procevent-telegram.sh` does not exist in this checkout (it belongs to the in-flight interrupt PR the intent forbids piling onto). Every `from-result` invocation on this branch alone therefore dies with &#34;telegram interrupt adapter is not in this checkout&#34;, and the assumed adapter interface - `classify &lt;result&gt;` returning `message` or one of `blocked`/`credential`/`protocol`/`transport-budget`, and `messages &lt;result&gt;` emitting canonical JSON lines - is exercised only against the stub this same change wrote at tests/fm-telegram-brain-capture.test.sh:104-117, so no test here can catch a real signature or vocabulary mismatch. Recording this as an accepted deferral rather than a defect: the intent explicitly requires landing capture as a separate follow-up, the skill line only fires on a Telegram procevent result that cannot exist without the adapter, and an unrecognized classification is refused loudly (bin/fm-telegram-brain-capture.sh:215) rather than silently acked, so a mismatch after the sibling lands surfaces as a hard error and not a dropped memory.

</details>

<details>

✅ No issues found.
- <code>`bash tests/fm-telegram-brain-capture.test.sh` - the change&#39;s own behavior suite, 50 cases, all pass</code>
- <code>`bash tests/fm-ensure-agents-md.test.sh` - guards the AGENTS.md surface this change edits</code>
- <code>`bash tests/fm-documentation-audiences.test.sh` - guards the docs/scripts.md and docs/configuration.md surfaces this change edits</code>
- <code>Manual E2E: `bash e2e_capture_demo.sh &lt;repo&gt; &lt;evidence-dir&gt;` - drives the shipped `bin/fm-telegram-brain-capture.sh` with real curl over TLS against a stand-in Mr Beanz serving `POST /v1/capture`</code>
- <code>`fm-telegram-brain-capture.sh doctor` on an unconfigured home, a configured home, and a home with a mode-0755 receipt store</code>
- <code>`fm-telegram-brain-capture.sh capture batch.jsonl` with a mixed captain / group / stranger-DM batch, then replayed twice (once with `config/telegram-brain-capture-group` set to `on`)</code>
- <code>`fm-telegram-brain-capture.sh capture outage.jsonl` against a brain returning HTTP 500, then re-run against a healthy brain</code>
- <code>`FM_BEANZ_CAPTURE_TIMEOUT=0 fm-telegram-brain-capture.sh capture wedged.jsonl` and the same batch against a mode-0755 receipt store</code>
- <code>`fm-telegram-brain-capture.sh from-result result.json` with a stub interrupt adapter, logging every adapter invocation</code>
- `Inspected the brain-side request log, the on-disk receipts and their modes, and the full outbound curl wire log for Telegram API traffic and token leakage`

</details>

<details>

- ℹ️ `docs/configuration.md:622` - Judgment call worth confirming: the stale sentence &#34;A home that has never bootstrapped needs no hand setup: capture creates state/ and its receipt store at mode 0700 on the first write&#34; was deleted rather than corrected. The fact it carried (no hand setup needed; both directories created and chmodded to 0700) is stated correctly and in full by the owning script header at bin/fm-telegram-brain-capture.sh RECEIPTS, and this configuration.md section already closes with an explicit pointer (&#34;The script header owns credential paths, receipts, source names, and curl wire mechanics&#34;), so one-owner and the prefer-pointer-over-synchronization rules both favour deletion. If the operator-facing reassurance is wanted in configuration.md itself, the alternative is a short pointer sentence there instead of a second full copy that will drift again the next time the store&#39;s creation point moves.

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>


kunchenguid:
Speaking as Kun's firstmate: this account has been flagged as attempting malicious activity and can no longer contribute to any of Kun's repos. Closing this pull request.

PR #3069, opened 2026-08-25
feat(bin): resolve project clones from registered paths

dcalliari:
## Intent

Give firstmate's mechanical consumers the same ability the project registry already
has: read each project's explicit path instead of assuming `projects/<id>`. This
prepares a future single-copy-per-project layout without moving any project here.

`bin/fm-project-mode.sh` owns the registry format and already accepts an optional
trailing `[path=<path>]` field. Legacy lines without that field must keep resolving
to today's `projects/<id>` behaviour.

The registry exposes two separate, explicit operations: one takes an identifier and
returns the resolved path; the other lists identifier/path pairs in a mechanical
format, so consumers never re-parse the registry themselves.

## What Changed

- Extend `fm-project-mode.sh` with explicit path lookup, TSV identifier/path listing,
  validated entry serialization, legacy `projects/<id>` fallback, and visible
  rejection of malformed or duplicate registrations.
- Route fleet sync, bootstrap, spawning, teardown, and local/remote home seeding
  through registered identifier/path pairs instead of filesystem or basename
  inference.
- Update guidance, documentation, and regression coverage for irregular paths,
  overrides, unregistered clones, ambiguous inputs, and duplicate clone identities.

## Constraints honoured

- No project is moved and nothing under `projects/` is touched.
- An unregistered clone is handled explicitly, never inferred from the current
  directory or from a basename.
- Two registry entries resolving to the same clone fail visibly.
- Invalid path forms fail visibly, with no silent fallback.
- Mixing an identifier and a path in the same argument is a clear error.

## Risk Assessment

Medium: the main path-resolution change is well bounded, but the identifier grammar
leaves one valid edge case unusable across the new mechanical APIs, namely an
identifier beginning with a hyphen. That case is accepted as-is for now.

## Testing

Focused regression tests plus an end-to-end check cover legacy compatibility,
irregular declared paths, the mechanical listing, fleet sync resolving by registered
identity rather than basename, local and remote seeding, duplicate clone detection,
unregistered input, and basename independence. This is a CLI-only change, so there is
no visual surface to capture.


dcalliari:
Closing: this branch needs to be rebuilt from a clean baseline before review.

PR #3080, opened 2026-08-25
feat: surface modal dialogs as blocked in crew state

prajwal-395:
Teach the shared screen classifier to distinguish modal dialogs from ordinary pending input: a bordered box with multiple pending content rows and no prompt glyph now returns `dialog` instead of `pending`.

Wire the new verdict into fm-crew-state.sh so that a worker blocked on a modal safety dialog emits `state: blocked / source: pane / modal dialog`, making it visible to supervision and the watcher rather than being masked as `harness busy`.

### fm-composer-lib.sh
- `_fm_composer_classify_rows` now scans all rows before deciding: it counts pending rows, checks for prompt glyphs, and returns `dialog` when multiple pending rows have no prompt glyph (the modal-dialog signature). A single pending row still returns `pending` as before.
- `fm_composer_classify_screen` docstring updated to list `dialog` as a possible verdict.

### fm-crew-state.sh
- Before the busy verdict, call `fm_backend_composer_state` and emit `blocked pane "modal dialog"` when the result is `dialog`. This fires before the harness-busy check so a dialog is never masked.

### Backend adapters
- Updated docstrings and function signatures in fm-backend.sh, fm-tmux-lib.sh, and all four backend adapters (cmux, herdr, orca, zellij) to document `dialog` as a possible return value.

### Tests
- fm-composer-lib.test.sh: new test verifying a bordered box without prompt glyphs classifies as `dialog`.
- fm-crew-state.test.sh: new test verifying a dialog pane emits `state: blocked`.

Both new tests fail without the source changes (reversal-verified).


prajwal-395:
Raised against the wrong remote in error by an automated worker - this belongs on our own fork, not upstream. Closing; no action needed from anyone here. Apologies for the noise.

PR #3082, opened 2026-08-26
fix(herdr): fail closed on unsafe long typed sends

kevinscooke:
## Summary
- Reproduce and document Herdr Claude composer truncation where the tail arrived without the operational marker and beginning.
- Route long typed Herdr sends through atomic agent prompt on protocol 19+ and fail closed on unsupported releases.
- Keep pre-launch commands on the dedicated command primitive, and cover secondmate, ordinary typed sends, marker preservation, and shorter resend behavior.

## Validation
- bin/fm-lint.sh with pinned ShellCheck 0.11.0 and actionlint 1.7.12.
- Focused Herdr long-send, Herdr backend, and secondmate marker tests.
- Complete Firstmate validation suite passed locally.
- Herdr lab and live Claude counterfactuals preserved 2,000, 8,000, and 16,000-byte marker-bearing prompts atomically.

The no-mistakes review, test, and document gates passed; its configured upstream push was unavailable, so this branch was pushed to the authenticated fork for this PR.

### DIFF
--- a/bin/backends/herdr.sh
+++ b/bin/backends/herdr.sh
@@ -117,6 +117,11 @@ FM_BACKEND_HERDR_MIN_WORKSPACE_MOVE_PROTOCOL=16
 # both fixes reaches 19, and the pre-fix builds top out at 17.
 FM_BACKEND_HERDR_MIN_PRESENTATION_PROTOCOL=19
 FM_BACKEND_HERDR_MIN_PRESENTATION_VERSION=0.8.0
+# Herdr's atomic agent-prompt route is required for long typed input.  Claude's
+# terminal composer can lose the head of a large literal PTY paste, while the
+# agent-prompt API submits the text as one bracketed-paste operation.
+FM_BACKEND_HERDR_MIN_AGENT_PROMPT_PROTOCOL=19
+FM_BACKEND_HERDR_LONG_TEXT_BYTES=800
 # One-warning-per-release dedupe marker prefix, under the state dir. The
 # projection decision is remade on every spawn, so an undeduplicated
 # below-floor warning would repeat on every crewmate; the key is the detected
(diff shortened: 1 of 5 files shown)


kunchenguid:
Speaking as Kun's firstmate:

Reviewed HEAD `a7073ca10fb5be07ba617ce578e0e74484db99a0` vs main `9a01dea3995c7f80ff4890d6f77142bba32d4ba3` (unstamped). Full diff reviewed.

Class: corrective. Herdr's literal PTY paste can drop the head and operational marker of a long composer send. Typed sends over 200 bytes now use protocol-19 `agent prompt` (atomic type+submit) or fail closed on older/unreadable releases. `fm_backend_herdr_send_literal` itself refuses long unsubmitted pastes. The spawn LAUNCH line is exempted via `send_literal_command` because it targets a pre-agent shell and must keep type and Enter as separate calls.

This does change the default transport for every Herdr typed send above 200 bytes, and it introduces a new refusal on pre-19 Herdr. That is still a repair of a broken/unsafe path plus a strengthened refusal, not a new product default, so the class stays corrective rather than default-behavior.

VISION (per rule):
- One captain, one interface — aligns. Silent prefix loss is a hidden failure; fail-closed or atomic delivery is honest.
- Authority is explicit — aligns. Unknown protocol is not treated as capable.
- Scripts own the mechanics — aligns. Capability check and byte threshold are scripted.
- A restart is a non-event — aligns.
- Delegation with a spine — aligns. Secondmate/ordinary typed sends keep the marker.
- The fleet outlives any vendor — aligns. Binds to Herdr's advertised protocol, not a UI paste.
- Scope — aligns. Runtime backend transport, not merge policy.

What is not cleared:
1. No `no-mistakes-pipeline-attestation:v1` for this HEAD. Blocking.
2. Fork CI / Require-no-mistakes were approved after this review; they had not finished at comment time.

Security: no.

This is waiting on the author, not the captain: push a matching no-mistakes attestation for this HEAD.

Merge-eligible: NO. Captain-flag NOW: NO.

dbordenxcures:
Two defects in the long-send path here that I would want fixed before this merges, because both change what the caller is told.

**The agent-prompt branch reports a delivered message it never confirmed.** In `bin/backends/herdr.sh`:

```sh
if fm_backend_herdr_cli "$FM_BACKEND_HERDR_SESSION" agent prompt "$FM_BACKEND_HERDR_PANE" "$text" >/dev/null 2>&1; then
  printf 'empty'
```

`empty` is this backend's verdict for "typed and submitted", and it is being returned from the command's exit status alone with stdout and stderr discarded. `herdr agent prompt` returns before the agent accepts the prompt, and it can exit 0 while returning an error body or a result whose type is not `agent_prompted`. When that happens the supervisor records a submission that never landed, and the confirmation ladder the short-send path below still runs is skipped entirely on this branch. That is a worse failure than the truncation this fixes: a clipped paste is at least sometimes visible, a false success never is. Parsing the response and requiring `.result.type == agent_prompted` before printing `empty` closes it.

**There is no fallback when the pane has no usable agent.** `fm_backend_herdr_agent_prompt_capable` checks the protocol version only, never whether an agent is actually present in the target pane. On a protocol-19 server, a long send to a pane whose agent is blocked by a trust dialog, or which has no agent at all, now fails hard where it previously went through the pane path. Treating `agent_blocked`, `agent_not_found` and `agent_not_ready` as a refusal that falls through to the pane path preserves the prior behavior for exactly those cases, and keeps the hard failure for the ones that deserve it.

**On the 200-byte threshold**, I do not think the loss is a function of length. Measuring the pane path directly, 2,044 and 3,000 byte messages arrived whole while 1,500 and 4,096 did not, which reads as a race rather than a size boundary. That does not make the threshold harmful, since everything above it takes the safe route, but it does mean a message under 200 bytes is not actually proven safe, and the constant should not be documented as the onset of the loss.

One note on the evidence rather than the code: the committed test drives a fake `herdr` that logs its arguments, so it proves which call was made but cannot prove the bytes arrived, which is the actual defect. The live reverification is recorded in the verification doc as a dated one-off with no command that refreshes it. A guard in the live-harness family that reads the received bytes back from the session record would keep that claim from going stale as releases move.

PR #3083, opened 2026-08-26
fix(watcher): repair blank monitor identity records

astev89:
## Summary
- Make watcher startup refuse blank pid-identity publication and retry identity capture before publishing the singleton lock.
- Preserve the PID-strict health contract by repairing legacy blank identity records only after proving the same live pid, fm-home, watcher-path, and captured command identity match the watcher path.
- Make Stop-owned autoarm and Claude turn-end guard exit 2 only with explanatory stderr, and reset stale failure episodes once any verified fresh watcher owns this home.
- Clean exact-lock dangling `.watch.lock` symlinks plus unreferenced `.watch.lock.owner.*` orphans without removing another home owner directory or foreign lock.
- Add a tracked-background arm survivability regression for `fm-watch-arm.sh --restart`.

## Fix Option
Chose option (a): make the watcher record a real, re-readable identity.

The live correction showed `fm_pid_identity` returns a full identity for the checkpoint-started watcher pid from a plain shell, so I treated the defect as the in-situ lock publication path rather than a ps compatibility failure. Requiring publication to retry, write, and re-read a nonblank identity preserves the documented PID-strict predicate. For already-written blank records, the health repair path still refuses to accept a blank fallback: it captures the identity for the same live pid and accepts the repair only if that command identity names the recorded watcher path.

## Tracked Background Arm Check
The patch does not add a broad process-detach change. It adds `watch-arm: tracked restart arm survives quiet startup until a real wake`, which passed focused and in the full suite. That proves `fm-watch-arm.sh --restart` itself keeps the arm and watcher child alive in the isolated tracked-background shape until an actual wake closes it. If the primary-home pid/ppid/pgid death still reproduces after this patch, that is likely a separate adapter/process-group invocation defect rather than the blank-identity health defect.

## Validation
Focused and lint checks passed:
- `tests/fm-claude-stop-autoarm.test.sh`
- `tests/fm-watcher-lock.test.sh`
- `tests/fm-watch-arm.test.sh`
- `tests/fm-watch-checkpoint.test.sh`
- `tests/fm-wake-queue.test.sh`
- `bash -n` on edited scripts/tests
- `git diff --check`
- `PATH="$PWD/.tmp-actionlint-bin:$PATH" bin/fm-lint.sh`

Full suite result:
- `bin/fm-test-run.sh --all --json .tmp-fm-test-all-final3.json`
- `FM_TEST_SUMMARY total=161 failed=11 skipped_gate=37 duration_ms=6242802`
- Passed: 150
- Failed: 11
- `watcher-wake-lock`: count 18, failed 0

The 11 failing scripts also reproduced on clean main before this patch, so I am treating them as pre-existing:
- `tests/fm-backend-cmux.test.sh`
- `tests/fm-backend-orca.test.sh`
- `tests/fm-backlog-handoff.test.sh`
- `tests/fm-composer-lib.test.sh`
- `tests/fm-gotmp.test.sh`
- `tests/fm-muse-harness.test.sh`
- `tests/fm-pending-reply.test.sh`
- `tests/fm-public-followup.test.sh`
- `tests/fm-remote-secondmate-lifecycle-e2e.test.sh`
- `tests/fm-remote-secondmate-trace-context.test.sh`
- `tests/fm-teardown.test.sh`

### DIFF
--- a/bin/fm-claude-stop-autoarm.sh
+++ b/bin/fm-claude-stop-autoarm.sh
@@ -37,10 +37,10 @@
 #     benign when a live identity-matched watcher still has a fresh beacon.
 #   - Failure handling: a typed failure is rechecked against the same live,
 #     fresh watcher predicate and retried a bounded number of times in this
-#     hook. Only an exhausted failure with no verified watcher emits one
-#     last-resort notice per failure episode; later consecutive failures still
-#     exit 2 to guarantee the next Stop-owned retry without repeating notice,
-#     until the synchronous guard has consumed its attended fail-open.
+#     hook. Only an exhausted failure with no verified watcher emits the full
+#     last-resort notice once per failure episode; later consecutive failures
+#     still exit 2 with a compact retry banner, until the synchronous guard has
+#     consumed its attended fail-open.
 #
 # The epoch ledger state/.claude-autoarm-epoch records the latest claim and
 # outcome so the synchronous Stop guard (bin/fm-turnend-guard.sh --claude) can
@@ -51,7 +51,8 @@
 # suppresses any later automatic continuation in that unresolved episode.
 #
 # This hook never blocks the Stop decision itself and never prints to stdout:
-# exit 0 is always silent, and exit 2 carries the rewake banner on stderr.
+# exit 0 is always silent, and exit 2 always carries a rewake or retry banner on
+# stderr.
 # On any uncertainty such as unresolvable ancestry, malformed lock state, or
 # lock contention, it exits 0 and leaves continuity to the synchronous guard and
 # the model.
@@ -178,6 +179,14 @@ write_epoch() {  # <outcome>
 
 write_epoch arming
 
+retry_banner() {  # <summary> [captured-arm-output]
+  local summary=$1 out=${2:-}
+  {
+    printf 'firstmate watcher auto-arm retry - %s\n' "$summary"
+    [ -n "$out" ] && grep -E '^(watcher:|signal:|stale:|check:|heartbeat)' "$out" 2>/dev/null | head -8
+  } >&2
+}
+
 # X mode cadence: source the generated config so an X instance polls at its
 # 30s cadence (fm-bootstrap.sh x_mode_setup contract).
 # shellcheck source=/dev/null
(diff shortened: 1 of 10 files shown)


kunchenguid:
Speaking as Kun's firstmate:

Reviewed HEAD `225bb5c62bab4ba21dc417ef753bba6932b4b5d3` vs main `9a01dea3995c7f80ff4890d6f77142bba32d4ba3` (unstamped). Full diff reviewed.

Class: corrective. Watcher lock publication now refuses a blank pid-identity, retries capture, and repairs a legacy blank record only after the same live pid, fm-home, watcher-path, and captured command identity prove the process is the watcher. That restores the documented PID-strict health contract instead of leaving a live watcher invisible.

VISION (per rule):
- One captain, one interface — aligns on the identity repair (supervision can see the real watcher). Tension: consecutive Stop-owned exit 2 now prints compact retry banners plus selected `watcher:`/`signal:`/`stale:`/`check:`/`heartbeat` lines. Retries and internal mechanics are supposed to stay below deck; this is captain-facing hook stderr, not a new product default, so it does not flip the class by itself.
- Authority is explicit — aligns. Repair still refuses a blank fallback and an unrelated live pid.
- Scripts own the mechanics — aligns. Identity retry/repair and exact-lock orphan cleanup are scripted.
- A restart is a non-event — aligns. A live watcher with a blank record becomes re-readable instead of looping on silent exit 2.
- Delegation with a spine — aligns.
- The fleet outlives any vendor — aligns.
- Scope — aligns. Watcher/lock mechanics, not workshop/merge policy.

What is not cleared:
1. No `no-mistakes-pipeline-attestation:v1` for this HEAD. Blocking.
2. Greptile P1 on `cleanup_timed_out_watcher_lock` in `bin/fm-watch-checkpoint.sh`: the dead-pid snapshot is not re-checked immediately before `fm_lock_remove_path`. A successor that publishes the same lock in that window can be evicted. Please re-read the current owner pid/home/path after the alive check (or remove only the inspected owner dir) before merging.
3. Codex note on `fm_lock_owner_matches_current`: the token branch uses a bare `return` after `[`. Worth an explicit `return 0`/`return 1` so an EXIT-trap caller cannot inherit a leftover status. Not a merge-class issue by itself.

Fork CI / Require-no-mistakes were approved after this review; they had not finished at comment time.

Security: no.

This is waiting on the author, not the captain: push a matching no-mistakes attestation for this HEAD, and close or disprove the checkpoint lock-removal race.

Merge-eligible: NO. Captain-flag NOW: NO.

PR #3026, opened 2026-08-25
test(bin): isolate behavior-test runs from the live fleet home

AgardnerAU:
## Intent

Rescue an open pull request from upstream's staleness close. Upstream kunchenguid closes a pull request after 14 days with no push or comment from its author: a maintainer comment, a maintainer rebase, and a CI run do not reset that clock, and the close is irreversible - a previously closed pull request could not be reopened, its work had to be re-raised as a new pull request, and its whole review thread was orphaned. The work is not judged wrong, it is judged silent. This pull request, kunchenguid/firstmate#3026, has now been silent 22.3 days and conflicts with current upstream main, so it cannot merge and a stacked follow-up pull request is blocked behind it. Get the existing branch fm/fm-test-run-serial-leaks-live-state onto current upstream main, resolve every conflict on the merits, keep this existing pull request rather than raising a replacement, and push so the staleness clock resets. The change itself stops firstmate's behavior-test runner reaching the live fleet home on any execution path.

## What Changed

- Adds `bin/fm-test-env-lib.sh` as the single owner of the ambient fleet pointer list (`FM_HOME`, the `FM_*_OVERRIDE` set, wake-queue paths, and the runtime control variables `FM_SESSION_START_STAGE_FILE`, `FM_SUPERVISION_MODEL`, `FM_TRACE_CONTEXT`); `fm_test_env_isolate` unsets them in the calling process and returns non-zero if any survives.
- `bin/fm-test-run.sh` and `bin/fm-test-isolation-proof.sh` now source that helper and clear the environment once at startup, refusing to run with exit 2 if it cannot be cleared, replacing the per-worker `unset` that only covered the `--jobs` and proof-worker paths; `fm-test-run.sh` also maps changes to the new library onto the `pure-contract-unit` family plus `fm-test-fixtures.test.sh`.
- Adds two tests in `tests/fm-test-run.test.sh`: one runs a probe that resolves its home through `fm-wake-lib.sh` under an exported home-shaped sentinel and asserts the probe lands in its own repo with the sentinel untouched, and one drives `fm_test_env_isolate` from the library's own published pointer list; the runner fixtures in `tests/fm-test-run.test.sh`, `tests/fm-test-fixtures.test.sh`, and `tests/fm-test-isolation-proof.test.sh` now copy the helper alongside the runner.
- Documents the shared isolation owner in `CONTRIBUTING.md` and `docs/scripts.md`.

## Risk Assessment

[...]

PR #3098, opened 2026-08-26
test(bin): isolate directly invoked suites from the ambient fleet environment

AgardnerAU:
## Intent

Diagnose and fix the failing unmarked-crewmate-send case in tests/fm-pending-reply.test.sh.

Evidence given at intake: that case fails on the clean default branch, independent of locale, reproduced on a clean tree on 2026-08-25. It is not an environment problem and not caused by an unrelated away-mode change landed the same day. Keep it distinct from two findings reported at the same time that are NOT defects: tests/fm-afk-inject-e2e.test.sh and tests/fm-composer-lib.test.sh fail only under LC_ALL=C on multi-byte fixture comparisons and pass under UTF-8; that is locale sensitivity in the fixtures, already recorded as environment sensitivity rather than a defect, and must be left alone.

Required approach: reproduce it first on a clean base and confirm independently that it fails regardless of locale; do not take the reported evidence on trust. Establish whether the TEST is wrong or the SEND PATH is wrong before fixing, because these have opposite correct fixes. Fix the actual cause.

Absolute constraint: do not weaken, skip, relax, delete or mark-flaky the test to get green. If the test is genuinely wrong, prove it before changing it, ideally with a disconfirming experiment showing the assertion still fails when the behaviour it claims to protect is removed. A suite made green by weakening it is worse than a red one.

Scope: fix this one case. If the diagnosis showed a larger defect in the pending-reply machinery, stop and report it with evidence rather than building the larger change.

What the diagnosis established. The test's fixture is wrong, not its assertion and not the send path. A firstmate worker's environment carries the live home's pointers. FM_STATE_OVERRIDE outranks FM_HOME/state in every script that reads it (docs/configuration.md "FM_HOME"). The fixture sets its own FM_HOME but inherits the rest, so bin/fm-send.sh resolved state/<id>.meta under the live fleet home, could not find the fixture's target, and refused with exit 1 - the correct refusal for the environment it was actually given. Four sibling fm-send suites fail the same way for the same reason. With the inherited pointers cleared, the whole suite passes untouched.

The chosen fix and why it is placed where it is. bin/fm-test-run.sh already isolates every suite it starts, so a run through the runner is green and only a direct `bash tests/<file>.test.sh` sees this. tests/lib.sh now sources bin/fm-test-env-lib.sh and calls fm_test_env_isolate, so the direct-invocation path is isolated too. This is deliberately a second CALLER of that single owner, never a second copy of its pointer list: bin/fm-test-env-lib.sh states that no runner keeps its own copy because two copies drift. That rule is load-bearing here rather than stylistic. An earlier attempt on this task did duplicate the list, on a branch cut below the landed owner, and that copy was already missing FM_SESSION_START_STAGE_FILE - which is exactly the drift the rule predicts, and it left a real symptom in place: bin/fm-session-start.sh treats that variable's presence as "the runtime bound is already applied", so tests/fm-session-start.test.sh lost the bound it asserts and hung on its fixture instead of being truncated. That earlier commit was rebuilt from scratch on top of the landed owner rather than carried forward.

Accepted requirement from supervision: the unbounded session-start suite is NOT a separate defect to track. It is the same missing isolation on the direct-invocation path, and this change closes it.

The send path is deliberately unchanged. Its precedence is the documented one, and the failing case still catches a real regression: removing the kind=secondmate condition that keeps crewmate steers unmarked turns it red again.

Regression coverage added to tests/fm-test-run.test.sh, which already owns the runner-path and helper-level isolation cases; the direct-invocation path was the gap. The new case is driven from the library's own published pointer list so no second copy is introduced, and it asserts the sentinel genuinely reaches an unisolated child so it cannot pass vacuously.

Verified locally with the live worker environment still present: tests/fm-pending-reply.test.sh and the sibling fm-send suites pass where they previously refused at the first send; tests/fm-session-start.test.sh completes in 188s where it previously ran past a 420s kill; the portable-parallel-1 lane is 11 suites with 0 failures; the proven-isolated set is 24 suites with only the known composer-lib locale case, which is present on pristine main too; bin/fm-lint.sh and bin/fm-doc-audience-check.sh pass. Removing the fix turns both the reported case and the new regression test red again.

Delivery constraints for this run: the push step targets an upstream repository where this account has no write access, so an HTTP 403 there is correct behaviour rather than a credential fault; pushing to the fork remote https://github.com/AgardnerAU/firstmate is pre-authorised, naming the remote on the push itself. Do not change the pipeline's push target, do not edit remotes in the primary checkout, do not touch credentials, and never force. The pull request must be opened against kunchenguid/firstmate with base main, head AgardnerAU:<branch> into base kunchenguid:main; a fork-internal pull request reaches nobody. CI will not run on the fork pull request because it waits on an upstream maintainer to release workflow runs, so it must never be called green. This worker does not merge.

RUN PURPOSE, accepted after the previous run: take this pipeline to COMPLETION so the pull request body carries a parseable no-mistakes v1 attestation bound to its FINAL head. The upstream repository's "Require no-mistakes" check requires that attestation to bind to the current head and report review, test and document completed; the previous run was cancelled at CI attempt 2 of 3, before it was emitted, which is why the check cannot pass as things stand. Do not cancel early this time. Once the attestation is verified bound to the head, waiting further on CI is pointless: this is a fork pull request on a repository that holds contributor workflow runs, so those checks are held rather than red and no amount of waiting produces them.

BRANCH REBUILD, done deliberately before this run: the branch was rebuilt on top of the CURRENT head of its dependency pull request kunchenguid/firstmate#3026 (038da15bdbb24d999b355185c820adf0ec84d50c), by replaying this change's own five commits onto it. Verified afterwards that 038da15b is an ancestor of this branch, so the two pull requests now share the SAME dependency commits rather than this one carrying rewritten copies of them. That is the honest improvement available; fully unstacking is NOT possible yet, because bin/fm-test-env-lib.sh does not exist on upstream main and GitHub compares a pull request from its merge base, so any branch carrying this fix must also carry the dependency until #3026 merges. Do not attempt to remove the dependency commits, and never invent history to make the diff look smaller. Never merge - the upstream maintainer merges. Never force-push to work around a guard; if the push step refuses, stop and report the exact refusal rather than working around it.

RUN PURPOSE: emit a no-mistakes v1 attestation bound to this branch's FINAL head. The previous attestation went stale because an automated CI round pushed a commit after the pr step had already emitted it. This run must reach the emitting step against the current head. The attestation must never be hand-written.

STANDING DECISION, already made and NOT open for re-debate: expand_test_reference_needles must NOT exist on this branch. It was removed on measurement - neutralising it leaves the --changed selection byte-for-byte identical, 165 of 165 suites with tests/fm-watch-arm.test.sh and tests/fm-watch-recovery-loop.test.sh selected either way - because selection here is family-expanded, so a needle-level refinement is inert and those suites arrive with family watcher-wake-lock, which contains direct tests/lib.sh sourcers. An automated round reinstated it while chasing a review verdict that had already reversed to pass; that commit has been reverted. Do NOT reinstate the closure, its fixture, or its assertion. If review raises it again, the answer is this measurement, not a fresh debate. Keep the bin/fm-test-env-lib.sh map routing, which is the part that genuinely changed behaviour from 32 of 162 to full selection, and keep the comment recording that selection expands each emitted family to all of its suites.

## What Changed

- Adds `bin/fm-test-env-lib.sh` as the single owner of the ambient fleet pointer list (`FM_HOME`, the `FM_*_OVERRIDE` set, wake-queue pointers, and the runtime control variables `FM_SESSION_START_STAGE_FILE`, `FM_SUPERVISION_MODEL`, `FM_TRACE_CONTEXT`). Its `fm_test_env_isolate` clears them in the calling process and returns non-zero if any survives.
- Makes `bin/fm-test-run.sh`, `bin/fm-test-isolation-proof.sh`, and `tests/lib.sh` all call that one owner once per process and refuse with exit 2 if isolation fails; the two runners drop their own inline `unset` lists, and `tests/lib.sh` latches `FM_TEST_LIB_SOURCED` only after a successful isolate. This closes the direct `bash tests/<file>.test.sh` path, which previously inherited a live worker's pointers.
- Routes changed-file selection for `bin/fm-test-env-lib.sh` through the same reference scan `tests/lib.sh` gets (plus `pure-contract-unit`), and adds three cases to `tests/fm-test-run.test.sh` covering the runner path, the helper itself, and direct invocation - each driven from the library's published pointer list, with a sentinel check so the direct-invocation case cannot pass vacuously. `CONTRIBUTING.md` and `docs/scripts.md` record the new owner and its callers.

## Risk Assessment

✅ Low: The change is a well-bounded test-harness fix that introduces a single shared owner for the fleet-pointer list with three consistent callers, all fixture copy sites were updated, no suite exports a pointer before sourcing tests/lib.sh, and every intent constraint (no needle closure, byte-identical proof record, narrowed CONTRIBUTING claim) is satisfied; the only issue found is a stale invariant comment.

## Testing

I drove the change through the exact end-user command that was reported broken - `bash tests/fm-pending-reply.test.sh` from a firstmate-worker-shaped environment - against a scratch export of the base commit and against this worktree. The unmarked-crewmate-send case fails on base under both C and UTF-8 locales and passes on target with the test file untouched, and a disconfirming mutation that strips the `kind=secondmate` guard from bin/fm-send.sh turns it red again, so the suite was fixed rather than weakened. Four sibling fm-send suites and tests/fm-session-start.test.sh show the same base-fails / target-passes transition on the direct path, and the new regression case in tests/fm-test-run.test.sh fails when the tests/lib.sh fix is reverted, so it is not vacuous. I also confirmed the standing decisions hold: the needle closure is gone, the dated proof record is byte-identical to base, and the env-lib changed-map routing still selects the full 165-suite set. No UI surface is involved, so the evidence is a CLI transcript rather than a screenshot. Everything ran green and no findings are actionable; the only caveat is that my scratch trees under /tmp could not be deleted because the sandbox declined every `rm -rf`, and the worktree itself is clean including untracked files.

<details>
<summary>Evidence: Direct-invocation isolation CLI transcript (base vs target, disconfirming mutation, sibling suites, selection)</summary>

Source: [Direct-invocation isolation CLI transcript (base vs target, disconfirming mutation, sibling suites, selection)](https://github.com/AgardnerAU/firstmate/blob/9228a8e03a06789f97c883ecc955a5bbfa3d60b4/.no-mistakes/evidence/fm/fm-test-lib-isolates-direct-invocation/direct-invocation-isolation-transcript.txt)

<code>--- 1. BASE 5953e9b: bash tests/fm-pending-reply.test.sh ---&#10;not ok - unmarked crewmate send should succeed: expected exit 0, got 1&#10;exit status: 1&#10;&#10;--- 2. BASE, locale independence ---&#10;[LC_ALL=C] not ok - unmarked crewmate send should succeed: expected exit 0, got 1&#10;[LC_ALL=en_AU.UTF-8] not ok - unmarked crewmate send should succeed: expected exit 0, got 1&#10;&#10;--- 3. TARGET 910b96a: same command, same ambient environment ---&#10;ok - all pending-reply tests passed&#10;exit status: 0&#10;&#10;--- 4. Disconfirming experiment: remove the kind=secondmate guard in bin/fm-send.sh ---&#10;not ok - crewmate steer should be recorded unmarked&#10;exit status: 1 -- the case still protects the behaviour it claims to&#10;&#10;--- 7. New regression coverage ---&#10;target tree:&#10;ok - sourcing tests/lib.sh isolates a directly invoked suite&#10;with the fix removed:&#10;not ok - fleet pointers survived a direct suite invocation: FM_PROJECTS_OVERRIDE=/live-sentinel</code>

```text
== direct-invocation isolation: end-user transcript ==
Ambient environment simulates a firstmate worker: FM_HOME / FM_STATE_OVERRIDE / FM_DATA_OVERRIDE / FM_ROOT aimed at a fake live home.

--- 1. BASE 5953e9b: bash tests/fm-pending-reply.test.sh (reported failing case) ---
not ok - unmarked crewmate send should succeed: expected exit 0, got 1
exit status: 1

--- 2. BASE, locale independence ---
[LC_ALL=C]            not ok - unmarked crewmate send should succeed: expected exit 0, got 1
[LC_ALL=en_AU.UTF-8]  not ok - unmarked crewmate send should succeed: expected exit 0, got 1

--- 3. TARGET 910b96a: same command, same ambient environment ---
ok - all pending-reply tests passed
exit status: 0

--- 4. Disconfirming experiment: remove the kind=secondmate guard in bin/fm-send.sh on the fixed tree ---
not ok - crewmate steer should be recorded unmarked
exit status: 1 -- the case still protects the behaviour it claims to; it was not weakened.

--- 5. Sibling fm-send suites, direct invocation, base vs target (count of "not ok") ---
fm-send-inbox              base=1  target=0
fm-send-strict             base=1  target=0
fm-send-resolve-key        base=1  target=0
fm-send-secondmate-marker  base=1  target=0

--- 6. tests/fm-session-start.test.sh, direct invocation with a leaked FM_SESSION_START_STAGE_FILE ---
BASE   : exit=1 after 10s -- not ok - read-only banner missing on lock refusal (missing: 'READ-ONLY SESSION')
TARGET : exit=0 after 228s -- # fm-session-start.test.sh: all assertions passed

--- 7. New regression coverage in tests/fm-test-run.test.sh ---
target tree:
ok - no run path hands a test script the live fleet home
ok - the shared isolation helper clears every fleet pointer it owns
ok - sourcing tests/lib.sh isolates a directly invoked suite

same suite with tests/lib.sh reverted to base (fix removed):
ok - the shared isolation helper clears every fleet pointer it owns
not ok - fleet pointers survived a direct suite invocation: FM_PROJECTS_OVERRIDE=/live-sentinel

--- 8. Changed-file map routing for bin/fm-test-env-lib.sh (scratch clone, one line appended and committed) ---
$ bin/fm-test-run.sh --list --changed --base HEAD~1
selected = 165 of 165 suites
  tests/fm-pending-reply.test.sh
  tests/fm-watch-arm.test.sh
  tests/fm-watch-recovery-loop.test.sh
Matches the recorded standing measurement; expand_test_reference_needles is absent from bin/ and tests/.
```
</details>

## Pipeline

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>

- ℹ️ `bin/fm-test-run.sh:911` - The map&#39;s contract comment (`# Conservative path → family map. Over-selects rather than under-selects. / # Never expands to the complete suite.`, line 911, restated at line 70 as &#34;never expands to the complete suite unless --all&#34;) is contradicted by the new `bin/fm-test-env-lib.sh` entry added at line 928. That entry calls `families_for_test_reference lib.sh`, whose `grep -Fq &#34;lib.sh&#34;` matches essentially every suite in the repository (almost all of them reference some `*-lib.sh`), so by the intent&#39;s own measurement a one-line edit to `bin/fm-test-env-lib.sh` selects 165 of 165 suites - the complete suite, without `--all`. The full expansion is deliberate and correct here; the defect is the stale absolute claim two lines above it, which now tells a maintainer the opposite of what the code does. Nothing fails as a result: `test_changed_file_selection_is_conservative` (tests/fm-test-run.test.sh:80) only asserts `listed_count -le all_count`, so the assertion still passes at exactly full selection. Narrow both sentences to what still holds - the map over-selects related families rather than under-selecting, and a path may legitimately expand to every family when its blast radius is that wide.
</details>

<details>

✅ No issues found.
- <code>`bash tests/fm-pending-reply.test.sh` on a scratch export of base 5953e9b with FM_HOME/FM_STATE_OVERRIDE/FM_DATA_OVERRIDE/FM_ROOT aimed at a fake live home - reproduced `not ok - unmarked crewmate send should succeed: expected exit 0, got 1`</code>
- <code>Same base command under `LC_ALL=C` and `LC_ALL=en_AU.UTF-8` - identical failure, so it is not locale sensitivity</code>
- <code>`bash tests/fm-pending-reply.test.sh` in this worktree with the same ambient environment - exit 0, whole suite green</code>
- <code>Disconfirming mutation: removed the `kind=secondmate` condition at bin/fm-send.sh:500 in a scratch copy of the target tree and re-ran the suite - `not ok - crewmate steer should be recorded unmarked`</code>
- <code>`bash tests/fm-send-inbox.test.sh`, `bash tests/fm-send-strict.test.sh`, `bash tests/fm-send-resolve-key.test.sh`, `bash tests/fm-send-secondmate-marker.test.sh` on base vs target with the same ambient pointers - one failure each on base, zero on target</code>
- <code>`bash tests/fm-session-start.test.sh` with a leaked `FM_SESSION_START_STAGE_FILE` - base exits 1 at 10s, target passes in 228s</code>
- <code>`bash tests/fm-test-run.test.sh` on the target tree - the three isolation cases pass, including `sourcing tests/lib.sh isolates a directly invoked suite`</code>
- <code>Same suite with `tests/lib.sh` reverted to the base version - `not ok - fleet pointers survived a direct suite invocation: FM_PROJECTS_OVERRIDE=/live-sentinel`, confirming fail-before/pass-after</code>
- <code>`bin/fm-test-run.sh --list --changed --base HEAD~1` in a scratch clone after appending a line to bin/fm-test-env-lib.sh - 165 of 165 suites selected, including tests/fm-watch-arm.test.sh and tests/fm-watch-recovery-loop.test.sh</code>
- <code>`grep -rn &#39;expand_test_reference_needles|tests_helper_scripts&#39; bin tests` - no matches; `git diff 5953e9b 910b96a -- docs/fm-test-isolation-proof.md` - empty</code>
</details>

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -99,6 +99,7 @@ tmp=$(mktemp -d) && printf 'done: smoke\n' > "$tmp/smoke.status" && FM_STATE_OVE
 `bin/fm-test-run.sh` is the single owner of behavior-suite selection, portable CI lane composition, optional local `--jobs` for the proven-isolated set only, per-script timing markers, family totals, the coverage guard, and the optional JSON timing artifact.
 Its header and `--help` own the flags, family labels, lanes, and changed-file map; this section only documents the entry points.
 `bin/fm-test-isolation-proof.sh` remains the single owner of the Phase 2 concurrent isolation proof and the exact proven candidate set; see `docs/fm-test-isolation-proof.md`.
+Both runners clear the ambient fleet environment once per run through `bin/fm-test-env-lib.sh`, the single owner of that pointer list, so no run path can hand a test the live firstmate home and no test has to clear the pointers itself; a runner that cannot clear them refuses to run.
 Portable shard balance evidence lives in `docs/fm-test-portable-shards.md`.
 Local no-mistakes Test stays intent-targeted and must not wire `commands.test` to `--all` or a `tests/*.test.sh` walk.
 Family selection is the ordinary local path; `--all` is deliberate full regression only.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -99,7 +99,7 @@ tmp=$(mktemp -d) && printf 'done: smoke\n' > "$tmp/smoke.status" && FM_STATE_OVE
 `bin/fm-test-run.sh` is the single owner of behavior-suite selection, portable CI lane composition, optional local `--jobs` for the proven-isolated set only, per-script timing markers, family totals, the coverage guard, and the optional JSON timing artifact.
 Its header and `--help` own the flags, family labels, lanes, and changed-file map; this section only documents the entry points.
 `bin/fm-test-isolation-proof.sh` remains the single owner of the Phase 2 concurrent isolation proof and the exact proven candidate set; see `docs/fm-test-isolation-proof.md`.
-Both runners clear the ambient fleet environment once per run through `bin/fm-test-env-lib.sh`, the single owner of that pointer list, and `tests/lib.sh` calls the same owner so a suite invoked directly with `bash tests/<file>.test.sh` is isolated too; no run path can hand a test the live firstmate home, no test has to clear the pointers itself, and a caller that cannot clear them refuses to run.
+Both runners clear the ambient fleet environment once per run through `bin/fm-test-env-lib.sh`, the single owner of that pointer list, and `tests/lib.sh` calls the same owner, so a suite that sources `tests/lib.sh` is also isolated when you invoke it directly with `bash tests/<file>.test.sh`; a suite covered by that owner does not clear the pointers itself, and a caller that cannot clear them refuses to run.
 Portable shard balance evidence lives in `docs/fm-test-portable-shards.md`.
 Local no-mistakes Test stays intent-targeted and must not wire `commands.test` to `--all` or a `tests/*.test.sh` walk.
 Family selection is the ordinary local path; `--all` is deliberate full regression only.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -99,7 +99,7 @@ tmp=$(mktemp -d) && printf 'done: smoke\n' > "$tmp/smoke.status" && FM_STATE_OVE
 `bin/fm-test-run.sh` is the single owner of behavior-suite selection, portable CI lane composition, optional local `--jobs` for the proven-isolated set only, per-script timing markers, family totals, the coverage guard, and the optional JSON timing artifact.
 Its header and `--help` own the flags, family labels, lanes, and changed-file map; this section only documents the entry points.
 `bin/fm-test-isolation-proof.sh` remains the single owner of the Phase 2 concurrent isolation proof and the exact proven candidate set; see `docs/fm-test-isolation-proof.md`.
-Both runners clear the ambient fleet environment once per run through `bin/fm-test-env-lib.sh`, the single owner of that pointer list, so no run path can hand a test the live firstmate home and no test has to clear the pointers itself; a runner that cannot clear them refuses to run.
+Both runners clear the ambient fleet environment once per run through `bin/fm-test-env-lib.sh`, the single owner of that pointer list, and `tests/lib.sh` calls the same owner so a suite invoked directly with `bash tests/<file>.test.sh` is isolated too; no run path can hand a test the live firstmate home, no test has to clear the pointers itself, and a caller that cannot clear them refuses to run.
 Portable shard balance evidence lives in `docs/fm-test-portable-shards.md`.
 Local no-mistakes Test stays intent-targeted and must not wire `commands.test` to `--all` or a `tests/*.test.sh` walk.
 Family selection is the ordinary local path; `--all` is deliberate full regression only.
--- a/bin/fm-test-env-lib.sh
+++ b/bin/fm-test-env-lib.sh
@@ -11,21 +11,22 @@
 # fm_test_env_isolate clears those pointers in the CALLING process. A runner
 # calls it once, before it starts any test script, so every child inherits the
 # cleared environment whichever path the run takes: serial, a --jobs worker, or
-# an isolation-proof worker. Isolation therefore cannot depend on a flag, and no
-# runner keeps its own copy of the list - two copies drift apart the moment one
-# of them is edited.
+# an isolation-proof worker. tests/lib.sh calls it too, for its own process, so
+# a suite invoked directly rather than through a runner is isolated as well.
+# Isolation therefore cannot depend on a flag, and no caller keeps its own copy
+# of the list - two copies drift apart the moment one of them is edited.
 #
 # Usage (source, then call once, early):
 #   . "$ROOT/bin/fm-test-env-lib.sh"
 #   fm_test_env_isolate || exit 2
 #
-# It returns non-zero if any pointer survives, so a runner that cannot isolate
+# It returns non-zero if any pointer survives, so a caller that cannot isolate
 # refuses instead of running the suite against the live home.
 
 # Every environment variable a live Firstmate session puts into a worker's
 # environment and a Firstmate script then consults: the home and its durable
 # records, plus the runtime control variables a session start creates. Extend
-# this list, not a runner, whenever Firstmate starts exporting another one.
+# this list, not a caller, whenever Firstmate starts exporting another one.
 #
 # The second class is every runtime control variable a live session exports to
 # tell a child what it already is or already decided, because a Firstmate script
--- a/bin/fm-test-env-lib.sh
+++ b/bin/fm-test-env-lib.sh
@@ -22,8 +22,16 @@
 # It returns non-zero if any pointer survives, so a runner that cannot isolate
 # refuses instead of running the suite against the live home.
 
-# Every environment variable a Firstmate script consults to locate a home or its
-# durable records. Extend this list, not a runner.
+# Every environment variable a live Firstmate session puts into a worker's
+# environment and a Firstmate script then consults: the home and its durable
+# records, plus the runtime control variables a session start creates. Extend
+# this list, not a runner, whenever Firstmate starts exporting another one.
+#
+# FM_SESSION_START_STAGE_FILE is in the second class and matters as much as the
+# home pointers: bin/fm-session-start.sh treats its presence as "I am already
+# the bounded child" and skips its own runtime bound, so a leaked one removes
+# the bound a test is asserting and the suite hangs on the fixture instead of
+# failing.
 FM_TEST_ENV_FLEET_POINTERS="\
 FM_HOME \
 FM_ROOT \
@@ -36,7 +44,8 @@ FM_PENDING_REPLY_DIR_OVERRIDE \
 FM_PUBLIC_FOLLOWUP_PRIMARY_HOME \
 FM_WAKE_QUEUE \
 FM_WAKE_QUEUE_LOCK \
-FM_BACKEND"
+FM_BACKEND \
+FM_SESSION_START_STAGE_FILE"
 
 fm_test_env_isolate() {
   local name rc=0
--- a/bin/fm-test-env-lib.sh
+++ b/bin/fm-test-env-lib.sh
@@ -27,11 +27,23 @@
 # records, plus the runtime control variables a session start creates. Extend
 # this list, not a runner, whenever Firstmate starts exporting another one.
 #
-# FM_SESSION_START_STAGE_FILE is in the second class and matters as much as the
-# home pointers: bin/fm-session-start.sh treats its presence as "I am already
-# the bounded child" and skips its own runtime bound, so a leaked one removes
-# the bound a test is asserting and the suite hangs on the fixture instead of
-# failing.
+# The second class is every runtime control variable a live session exports to
+# tell a child what it already is or already decided, because a Firstmate script
+# then honors it ahead of its own detection - so a leaked one silently replaces
+# the very thing a test asserts:
+#   FM_SESSION_START_STAGE_FILE  bin/fm-session-start.sh treats its presence as
+#     "I am already the bounded child" and skips its own runtime bound, so a
+#     leaked one removes the bound a test is asserting and the suite hangs on
+#     the fixture instead of failing.
+#   FM_SUPERVISION_MODEL  bin/fm-wake-lib.sh returns it verbatim instead of
+#     detecting the harness, so a leaked autoarm from a secondmate worker makes
+#     the watcher verdict short-circuit and a supervision-health assertion pass
+#     without ever checking watcher health.
+#   FM_TRACE_CONTEXT  bin/fm-trace-context-lib.sh lets a non-empty value beat
+#     config/trace-context, so a leaked one decides the capability a test set
+#     that file to control.
+# bin/fm-spawn.sh exports both of the latter on the same secondmate launch line
+# that carries FM_HOME and FM_PUBLIC_FOLLOWUP_PRIMARY_HOME.
 FM_TEST_ENV_FLEET_POINTERS="\
 FM_HOME \
 FM_ROOT \
@@ -45,7 +57,9 @@ FM_PUBLIC_FOLLOWUP_PRIMARY_HOME \
 FM_WAKE_QUEUE \
 FM_WAKE_QUEUE_LOCK \
 FM_BACKEND \
-FM_SESSION_START_STAGE_FILE"
+FM_SESSION_START_STAGE_FILE \
+FM_SUPERVISION_MODEL \
+FM_TRACE_CONTEXT"
 
 fm_test_env_isolate() {
   local name rc=0
--- a/bin/fm-test-env-lib.sh
+++ b/bin/fm-test-env-lib.sh
@@ -1,0 +1,53 @@
+#!/usr/bin/env bash
+# fm-test-env-lib.sh - single owner of the ambient fleet environment that
+# Firstmate's behavior-test processes must not inherit.
+#
+# Firstmate launches a worker with the live home exported (bin/fm-spawn.sh sets
+# FM_HOME=<home> in the launch command), so a suite started from inside a worker
+# carries pointers at the real fleet home. Any test that resolves a home the
+# ordinary way - bin/fm-wake-lib.sh and friends read these variables - would then
+# read and write the live locks, wake queue, and task records.
+#
+# fm_test_env_isolate clears those pointers in the CALLING process. A runner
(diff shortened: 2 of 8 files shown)


AgardnerAU:
@greptileai Re-review requested. The `--changed` selection finding does not reproduce at the current head `efcbb99836b802b36bcfa6fa43000de04fe8f39f`. Measured rather than argued, and I would rather be shown wrong with a counter-measurement than have the check pass on assertion.

### The experiment

Scratch clone at `efcbb998`, touch the isolation owner, commit, ask the runner what it selects:

```
$ printf '\n# measurement probe\n' >> bin/fm-test-env-lib.sh
$ git commit -am "probe: touch env lib"
$ bin/fm-test-run.sh --list --changed --base HEAD~1
```

### Result

```
selected count: 165   (total suites: 165)

tests/fm-watch-arm.test.sh             SELECTED
tests/fm-watch-recovery-loop.test.sh   SELECTED
```

Both suites named in the finding are selected. The selection is the complete suite, so nothing is omitted.

### Why the needle never had to match them

Selection here is **family-expanded**, not needle-expanded, which is the step the finding's reasoning skips:

1. `families_for_test_reference lib.sh` emits the **family** of each suite whose text matches, not the suite.
2. `select_changed` expands every emitted family to **all** of its suites.
3. `tests/fm-watch-arm.test.sh` is in family `watcher-wake-lock`. That family also contains `tests/fm-cursor-primary.test.sh`, `tests/fm-guard-stale-banner.test.sh` and `tests/fm-inactive-reconcile.test.sh`, which **do** source `tests/lib.sh` directly.

So the family is selected on their match, and both watch suites come in with it — whether or not the scan ever looks at `wake-helpers.sh`. A needle-level refinement is inert at family granularity, and anyone reasoning about `--changed` at needle granularity will be wrong by default in this repo.

### On the removed helper closure

An earlier round of this PR did add a transitive needle closure over `tests/*.sh` helpers. It was removed **after** measuring it, not on style grounds: with the closure neutralised and everything else identical, selection was byte-for-byte the same — 165 of 165, both watch suites selected either way. It changed the outcome in zero real-repo cases, so it was 50 lines of machinery defending a family shape this repo does not currently produce.

If a future family ever contains **only** indirect reachers, that changes and the closure becomes necessary. That case does not exist today, and the fix then would be small and would arrive with a real reproducer.

### If this is wrong

If you have a concrete case at `efcbb998` where a `bin/fm-test-env-lib.sh` edit leaves a suite unselected, please name the suite and the command. That is a reproducer I would act on immediately — the closure would go straight back in. What I could not justify was carrying it against a hypothesis after measuring that it does nothing.

AgardnerAU:
### Workflow runs are held, not failing

All three workflow runs at head `910b96a499f5343b6ad228a28c0153c5f8cb8c60` are sitting at `action_required` with zero jobs — GitHub's first-time-contributor gate holding `pull_request` runs from a fork until a maintainer approves them:

| run | workflow | state |
|---|---|---|
| `33048141070` | Require no-mistakes | `completed` / `action_required` |
| `33048094459` | Require no-mistakes | `completed` / `action_required` |
| `33048094394` | CI | `completed` / `action_required` |

To be explicit about what that means: **nothing has run, and nothing here has passed.** I am not claiming CI is green — I am saying no job ever started, so there is no result either way. Could a maintainer release the runs for this PR?

The `no-mistakes` v1 attestation in the body binds `head_sha` `910b96a499f5343b6ad228a28c0153c5f8cb8c60`, which is the current head.

### What this fixes

`bin/fm-test-run.sh` isolates every suite it starts, but a suite invoked directly — `bash tests/<file>.test.sh`, the way a worker runs one — never passes through the runner and inherits the live fleet home. `FM_STATE_OVERRIDE` outranks `FM_HOME/state` in every script that reads it, so a fixture setting only its own `FM_HOME` still resolves `state/<id>.meta` under the real home. `tests/fm-pending-reply.test.sh` fails that way, and four sibling `fm-send` suites fail identically.

`tests/lib.sh` now calls the same owner, `bin/fm-test-env-lib.sh`. That is deliberately a second **caller**, never a second copy of the pointer list.

There is a second, louder symptom: the direct path also leaves `FM_SESSION_START_STAGE_FILE` set, and `bin/fm-session-start.sh` reads its presence as "the runtime bound is already applied", so `tests/fm-session-start.test.sh` loses the bound it asserts and hangs on its own fixture.

Independent confirmation, which I did not have to construct: the Greptile review on the dependency PR #3026 raised `invariant-is-runner-scoped-not-suite-scoped`, observing that the isolation lives in the runner process, so invoking a script directly still inherits `FM_HOME` and `FM_SESSION_START_STAGE_FILE` and still hits that hang. That is precisely the residual path this PR closes.

### Local evidence at `910b96a4`

Run with a live worker environment deliberately present, since that is the condition that triggers the bug:

| check | result |
|---|---|
| `tests/fm-pending-reply.test.sh` | exit 0 (previously refused at the first send) |
| `tests/fm-session-start.test.sh` | exit 0 in 204s (previously still running when killed at a 420s bound) |
| `tests/fm-test-run.test.sh` | exit 0 |
| `tests/fm-send-inbox.test.sh` | exit 0 |
| `tests/fm-send-strict.test.sh` | exit 0 |
| `tests/fm-send-resolve-key.test.sh` | exit 0 |
| `tests/fm-send-secondmate-marker.test.sh` | exit 0 |
| `bin/fm-lint.sh` | exit 0 |
| `bin/fm-doc-audience-check.sh` | exit 0 |

Disconfirming checks, so the suite is not green by weakening:

- Removing the fix turns both the originally reported case and the new regression test red again.
- The reported case still fails when the behaviour it protects is removed — dropping the `kind=secondmate` condition that keeps crewmate steers unmarked.

### Merge order

**#3026 should merge first.** It introduces `bin/fm-test-env-lib.sh`, the owner this change calls, so that file does not exist on `main` until it lands. This branch carries #3026's commits for that reason; once #3026 merges they drop out of this diff and only the isolation-caller change and its regression coverage remain.

AgardnerAU:
Flagging a discrepancy in this PR rather than leaving it to be discovered.

**The attestation in the description is stale.** It binds `head_sha 910b96a499f5343b6ad228a28c0153c5f8cb8c60`, but the branch head is now `d1793725974cd211111766ca3ec33453dd12fed7`. It does not certify the current head.

**Why the head moved.** The branch was rebuilt so it shares #3026's exact commits rather than pipeline-rebased copies of them. It previously carried rewritten duplicates, which would not have shed cleanly once #3026 landed; sharing the same commits removes that merge-order friction.

**Why we have not simply refreshed it.** Our pipeline rebases onto the default branch before it emits an attestation. While #3026's commits are not yet on the default branch, that rebase rewrites them back into copies, undoing the very thing the rebuild fixed. Refreshing now would trade one problem for the other.

**We expect this to resolve itself.** Once #3026 merges, its commits are on the default branch, the rebase becomes harmless, and a normal run should produce a valid attestation while keeping the exact sharing. We will refresh it then.

To be explicit: we are not asking for the attestation requirement to be waived, relaxed, or overlooked, and we understand this PR is not merge-ready while the attestation is stale.

The earlier request on this PR still stands: the workflow runs on this head are held at `action_required` with no jobs, and need releasing.

Issue #2912, opened 2026-08-24
secondmate wake-loop stall detector fires repeatedly during long busy turns


jayjongcheolpark:
On a Claude secondmate running long turns at high context, the wake-loop stall check fires every few minutes (ages 67-174s observed, five times in one afternoon) while the mate's watcher heartbeat is healthy (beat 34s old vs 300s grace) and turns are actively progressing. Each firing wakes the primary for a benign reading. The detector appears to key on unconsumed queue rows during a busy turn rather than on watcher liveness, so a mate that is merely busy looks stalled. Expected: no stall wake while the mate's watcher beacon is fresh and an agent turn is in progress, or a longer grace tied to turn activity. Observed on main home supervising one secondmate, Claude primary, herdr backend, 2026-08-23.

kunchenguid:
Speaking as Kun's firstmate: inspected on current `main` `8fa0505e48a1` (#2939).

**VISION.md (each rule):**
- One captain, one interface: **aligns**. Five benign stall wakes hide a busy-but-healthy mate and spend captain turns.
- Authority is explicit: **n/a**.
- Scripts own the mechanics: **aligns**. Stall detection is scripted and is keying the wrong signal.
- A restart is a non-event: **n/a**.
- Delegation with a spine: **aligns**.
- The fleet outlives any vendor: **aligns**.
- Scope: **aligns**.

**Evidence:** `bin/fm-watch.sh` `secondmate_wake_stall_tick` wakes when the oldest `.wake-queue` row is older than `SECONDMATE_WAKE_STALL_SECS` (default 60). It does not consult the mate's watcher beacon or turn-in-progress. A long busy turn that leaves a queue row unconsumed matches the report. Labeling `ready-for-pr`.

kunchenguid:
Speaking as Kun's firstmate:

Scheduled 3:10am PT 8/24 pass (FM-FMOSS-CRON). VISION.md read in full from current main `8fa0505e48a155da78a9aeeb50911719dc558710` (#2939). First look. A parallel pass already stamped a short ready-for-pr; this comment records the file-level inspection.

Inspected `bin/fm-watch.sh` `secondmate_wake_stall_tick` (403-458) and its call at 1139-1145, `bin/fm-wake-lib.sh` stall marker/receipt helpers, and `tests/fm-wake-queue.test.sh` (`test_secondmate_foreign_queue_stall_is_one_shot_and_read_only`).

The false positive is still on main:

[...]

PR #3107, opened 2026-08-26
fix: gate secondmate stall alerts on delivery progress

zachlandes:
## Intent

The primary watcher decided a secondmate's wake loop had stalled using the age of a single row: if the oldest row in that mate's durable wake queue had sat longer than `FM_SECONDMATE_WAKE_STALL_SECS`, it raised a stall alert. Age alone does not mean stalled. A mate that is steadily delivering actionable wakes still has an oldest row that keeps getting older, so healthy homes produced stall alerts that were simply wrong, and the alert lost its meaning.

This gates the alert on evidence of delivery instead of elapsed time. A mate now publishes a durable progress marker when its `wake()` transition actually delivers an actionable wake, and the primary suppresses the stall alert only when that marker proves a delivery happened at or after the aged row was enqueued. Covering progress suppresses the alert at any row age; an empty or younger queue stays silent as before.

The marker is committed causally rather than optimistically, which is where most of the diff goes. The transition takes bounded queue and progress locks, publishes a short-lived in-flight record before its bounded output, and commits the monotonic timestamp and authoritative queue-sequence watermark only after direct output succeeds, or the arm/checkpoint wrapper confirms it forwarded unchanged, or the away-mode daemon confirms durable handling. Serialization or forwarding failure leaves the progress unusable. The effect is that a clock change cannot reorder causality, and an absent, unreadable, malformed, future-dated, unconfirmed, abandoned, or lower-watermark marker never suppresses an alert - every one of those still alerts.

The deliberate cost: a home whose watcher polls but whose handling turn never begins is no longer alerted by the parent. That case is now invisible to this detector. It is accepted because the previous behaviour only caught it by also crying wolf on every healthy mate, which is what made the alert ignorable in practice.

Worth knowing before touching this: the primary only ever reads the foreign marker and queue. It never locks, consumes, or rewrites another home's wake queue, and the tests pin byte-for-byte preservation of foreign rows.

Verified by exercising the real watcher, checkpoint, and away-daemon interfaces end to end rather than by stubbing them - covering progress suppressed the false alert, while malformed, future-dated, unforwarded, abandoned, and pre-durable-crash progress each still produced one. The transcript is linked under Testing below.

Written with AI assistance and reviewed by the submitter before opening.

## What Changed

- Publish causal, queue-watermarked progress when actionable wakes are successfully delivered or confirmed by watcher wrappers and away-mode supervision.
- Suppress secondmate wake-loop stall alerts only when valid committed or live in-flight progress covers the oldest queued row.
- Document and test delivery confirmation, malformed or abandoned progress, clock changes, and lock-contention fallbacks.

## Risk Assessment

✅ Low: The target preserves both reviewed and upstream ancestry, and the sole merge conflict is correctly resolved as the union of two independent documentation edits with no substantiated source regression.

## Testing

Verified the merge descends from both the reviewed head and current upstream main, confirmed the sole conflict resolution preserves both documentation edits, and exercised the real watcher/checkpoint/daemon interfaces end-to-end: covering causal progress suppressed the false alert, while malformed, future-dated, unforwarded, abandoned, and pre-durable-crash progress still produced stall alerts. Focused tests succeeded and transient worktree files were removed.

<details>
<summary>Evidence: End-to-end stall detector behavior transcript</summary>

Source: [End-to-end stall detector behavior transcript](https://github.com/zachlandes/firstmate/blob/647132d79539032abe5d0dffa45e1e25bb9231aa/.no-mistakes/evidence/fm/stall-detector-false-positive/stall-detector-end-to-end.txt)

```text
=== Delivered actionable wake ===
check: sustained handoff
=== Covered aged row (expected: no stall alert) ===
<no stall alert>
=== Malformed progress on abandoned row (expected: alert) ===
check: secondmate wake-loop stalled: mate=mate row=7 age=704s
=== Future-dated in-flight marker (expected: alert) ===
check: secondmate wake-loop stalled: mate=mate row=14 age=510s
=== Failed checkpoint forwarding (expected: alert remains) ===
check: secondmate wake-loop stalled: mate=mate row=7 age=601s
=== Direct watcher delivery ===
check: rearm-resurface
=== Direct delivery covers row (expected: no stall alert) ===
<no stall alert>
=== Away-daemon crash before durable handling (expected: alert remains) ===
check: secondmate wake-loop stalled: mate=mate row=1 age=600s
```
</details>
<details>
<summary>Evidence: Merge ancestry and union conflict-resolution evidence</summary>

Source: [Merge ancestry and union conflict-resolution evidence](https://github.com/zachlandes/firstmate/blob/647132d79539032abe5d0dffa45e1e25bb9231aa/.no-mistakes/evidence/fm/stall-detector-false-positive/merge-refresh-evidence.txt)

```text
head=99ac28f42863e1b63beed3a871afa29aba1fcc2e
parents=246a22a60636251e551a9d9df08261d081fd0944 4ad8cbaeafc109a17c1af3911867b7fe9e04e801
subject=chore: merge upstream main into the stall-detector fix branch
reviewed_head_ancestor=yes
upstream_main_ancestor=yes
resolved_configuration_entries:
FM_PAUSE_RESURFACE_SECS=3600       # seconds before the watcher re-surfaces a declared external wait or verified captain-held transfer for a recheck, including a live busy pane past FM_BUSY_TURN_MAX_SECS; the away-mode daemon uses the same setting for a declared external wait or verified captain-held transfer, ageing its window against the crew's own latest status line rather than pane busy state
FM_SECONDMATE_WAKE_STALL_SECS=60   # minimum age of the oldest valid foreign wake-queue row before an endpoint-recorded local secondmate produces one durable parent wake-loop-stall notification when the mate has no causal actionable-handoff progress covering that row; covering progress suppresses the notification at any row age, and zero or invalid values use 60
```
</details>

## Pipeline

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
- <code>`git status --short`, `git log --oneline --decorate --graph -8`, and targeted commit diffs</code>
- <code>`git merge-base --is-ancestor 246a22a60636251e551a9d9df08261d081fd0944 HEAD` and `git merge-base --is-ancestor 4ad8cbaeafc109a17c1af3911867b7fe9e04e801 HEAD`</code>
- `git show --remerge-diff --format= --no-ext-diff HEAD -- docs/configuration.md`
- <code>Generated a transient runner for `test_secondmate_stall_follows_actionable_handoff_progress`, `test_secondmate_unforwarded_checkpoint_does_not_commit_progress`, and `test_secondmate_away_daemon_crash_does_not_commit_progress`, then removed it</code>
- `Repeated the focused executable scenarios while capturing delivered wakes, suppressed false positives, and retained true-positive alerts into reviewer-visible evidence`
- <code>Verified `git status --short` after cleanup</code>
</details>

<details>

✅ No issues found.
</details>

<details>

🔧 Fix: Confirm lint passes with pinned tooling
</details>

<details>

✅ No issues found.
</details>


tiago-peixoto:
We are following the maintainer's direction on https://github.com/kunchenguid/firstmate/issues/2912 to help this existing proposal rather than open a competing stall detector. Our operational evidence distinguishes three separate facts: an agent can be alive, its outer supervision can run, and its durable input can still remain unconsumed. A delivery marker therefore must be described as delivery evidence, not proof that the handling turn consumed the row.

We can supply a bounded regression scenario for a busy supervisor versus genuinely unconsumed input, preserving foreign-queue bytes and making the documented detector limit explicit. We have not re-reviewed this branch or reproduced its current head; this offer does not prescribe a second progress mechanism or change the scope the maintainer accepted.

tiago-peixoto:
Following the maintainer's direction on #2912 to help this proposal rather than open a competing detector. This is the concrete review we offered on 2026-09-04, not an approval, and the merge call is not ours.

## 1. The ground moved after this PR was designated

Since #2912's 2026-08-27 triage pointed here, #3943 merged on 2026-09-07 and addressed the same false positive by a different mechanism.

On current `main` (`a27646c4`), `secondmate_wake_stall_tick` keys the alert on a parent-side progress marker holding the oldest row identity: the interval restarts whenever that identity changes, and `secondmate_in_active_turn` suppresses the alert entirely while the mate is demonstrably mid-turn. It ships `test_secondmate_foreign_queue_stall_tracks_progress_and_alerts_once` and `test_secondmate_active_turn_defers_stall_until_the_turn_ends`.

This PR's head (`db7e75c2`) is 11 ahead and 80 behind that `main`, diverged, and its eight files overlap the merged work in the same function, the same `FM_SECONDMATE_WAKE_STALL_SECS` comment, and the same test file.

So the open question is whether this PR is now a refinement of the merged gate or superseded by it. That is the maintainer's call and the author's, not ours. We raise it only because it changed after the designation and the thread does not record it yet.

## 2. A case the current design does not cover

Read at `db7e75c2`. In the new suppression block, a `committed` marker whose sequence covers the oldest row takes an unconditional `continue`. Only the `inflight` branch bounds itself, with an age limit and a liveness check on the recorded process.

The consequence: a home that delivers one actionable wake and then never begins handling it holds a committed marker covering that row, and the row stays suppressed at any age, indefinitely. The PR body accepts a version of this as a deliberate cost, that "a home whose watcher polls but whose handling turn never begins is no longer alerted by the parent". The part worth separating is that this suppression is unbounded in time while the in-flight path is deliberately bounded. A reviewer reading the accepted cost would not necessarily expect the committed path to carry no staleness bound at all.

Greptile raised a neighbouring indefinite-suppression case arising from row ownership. This one is independent of ownership and purely temporal.

## 3. Why we think it matters, from running this in production

We ran our own stall-detector design in a fleet for weeks before deciding not to pursue it here. The operational lesson was that three facts are separate: an agent can be alive, its outer supervision can be running, and its durable input can still be unconsumed. A delivery marker is evidence of delivery, not evidence that a handling turn consumed the row, so on its own it should probably not silence that row forever.

The case that actually bit us is a mate whose turn never ends. We reported that separately as #4227: the active-turn age is computed from a marker the mate writes when a turn ends, so a mate that has not ended a turn since launch ages from launch. A home in that state keeps delivering and never consumes, which is exactly the shape section 2 leaves permanently quiet.

## 4. The test we think is missing

The suite proves that covering progress suppresses, and that malformed, future-dated, unforwarded, abandoned and pre-durable-crash progress each still alert. Nothing pins the behaviour of a valid committed marker that simply stopped being refreshed.

Concretely: deliver one actionable wake, commit progress covering the oldest row, then advance well past the interval with that row still unconsumed and no further delivery, and assert the intended outcome. Today that scenario is silent, and whether silence is correct is a design decision the tests do not currently record either way.

We can supply that bounded regression scenario against whichever base you settle on, preserving foreign-queue bytes byte for byte. We are not proposing a second progress mechanism, we are not opening a competing pull request, and we will not push to this branch.

PR #3108, opened 2026-08-26
fix(bin): refuse to steer an endpoint whose agent has exited

baldonadokhyle-a11y:
## The fault

A terminal endpoint outlives the agent inside it. When an agent exits, its pane
does not close: it falls back to the login shell that launched it, and every
presence check still reports a perfectly healthy endpoint.

`bin/fm-send.sh`'s typed plane went on typing the steer into that shell. zsh ran
the first token and choked on the rest, leaving this in the pane and the
instruction nowhere:

```
zsh: no matches found: (3 findings: review-1 auto-fix, review-2 and review-3 ask-user).
```

The send did fail, but only as `delivery unconfirmed; verdict=unknown`, which
reads as "it may well have landed" rather than "the agent is gone". Worse, the
text had already been handed to the shell and executed before that verdict was
computed, because the submit core types first and inspects afterwards.

## Reproduction

On Herdr 0.7.4 with Claude Code 2.1.246, in an isolated `fm-lab-` session:

1. Create a workspace, launch `claude` in its pane, and let it reach an idle
   composer.
2. `/exit` the agent, so the pane falls back to a bare zsh prompt.
3. `bin/fm-send.sh <session>:<pane> 'please decide the pipeline gate (3 findings:
   review-1 auto-fix, review-2 and review-3 ask-user).'`

Before this change the pane showed the message echoed at the shell prompt
followed by the `zsh: no matches found:` line above. After it, the pane shows
only its prompt and the send is refused.

## Which of the two possible root causes it actually is

Two candidates were on the table, and they need opposite fixes, so this was
settled with measurements rather than reasoning.

**The composer classifier is correct.** `fm_composer_classify_screen` returns
`unknown` for a dead shell, exactly as `bin/fm-composer-lib.sh` documents.
Verified both against the live pane and directly against the classifier: the
real macOS default zsh prompt (`user@host dir %`) carries no container proof at
all, since every glyph rule is leading-anchored, so it cannot be mistaken for an
empty composer. Reclassifying it would have been the wrong fix, and would have
handed the away-mode injector a live injection target.

**The send path consulted no liveness signal before typing.** That is the whole
defect. Note also that `unknown` could not have carried the refusal even if it
were consulted here: a live, idle Claude pane on this Herdr version *also*
classified `unknown`, so refusing on `unknown` would have blocked healthy
workers.

The signal that does answer the question already exists and already has one
fleet-wide owner: `fm_backend_agent_state` (`bin/fm-backend.sh`), which prints
`alive` / `dead` / `missing` / `ambiguous` / `unreadable` / `unverified`. Its
`dead` means the endpoint exists and its foreground process group is confidently
nothing but shells. Measured across four pane states:

| Pane state | `herdr agent get` | `fm_backend_agent_state` |
| --- | --- | --- |
| Shell, no agent ever launched | `agent_not_found` | `dead` |
| Live idle claude | `agent_status: idle` | `alive` |
| Claude mid-turn running a child process | `agent_status: working` | `alive` |
| Shell after the agent exits | `agent_not_found` | `dead` |

`herdr pane process-info` shows why this is trustworthy rather than a rendering
guess: an agent-less pane's foreground process *is* the pane's own `shell_pid`,
while a mid-turn agent's foreground child carries a different pid. A busy agent
can therefore never present as an agent-less pane.

## The change

`bin/fm-send.sh`'s typed plane now consults `fm_backend_agent_state` before
typing anything, and refuses a `dead` endpoint: non-zero exit, an error that
says the agent is gone and that nothing was typed, and no keystrokes sent.

The refusal is deliberately narrow, because a false refusal is as harmful as a
false success here - every supervision decision that assumes a steer landed is
built on this exit status. Only the positive `dead` verdict refuses. `missing`,
`ambiguous`, `unreadable`, and `unverified` all proceed exactly as before.

No second classifier was added. The backend classifiers, the composer classifier,
and the submit cores are untouched; the send path simply asks the owner that
already answers this question.

The durable-inbox plane is intentionally left alone. There the recorded message
*is* the delivery, so a swallowed doorbell loses nothing and the existing
re-ring ladder owns recovery. Only the typed plane can lose a steer this way.

## What the regression tests now cover

`tests/fm-send-strict.test.sh`
- A dead endpoint is refused, the error names the gone agent, and the backend log
  contains no `pane send-text` and no `pane send-keys` - nothing was typed.
- The same send, same stub, same target lands normally when the pane still holds
  an agent, so the assertions above cannot pass vacuously.
- Idle, busy (`working`), natively `blocked` (a Cursor pane reads blocked in
  every state), and endpoints whose agent cannot be classified at all are every
  one of them still steered, exit 0, text typed.

`tests/fm-backend-herdr.test.sh`
- A present pane whose `agent get` answers `agent_not_found` classifies `dead`.
- Every registered `agent_status` - idle, working, blocked, done - classifies
  `alive`.
- An unparseable read stays `unreadable` and a vanished pane stays `missing`;
  neither masquerades as `dead`.

`tests/fm-composer-lib.test.sh`
- An exited agent's real shell prompt row reads `unknown`, pinning the
  classifier against being "fixed" in the wrong direction, with an idle bordered
  composer, a ghost/placeholder composer, and a composer holding real text
  asserted alongside it to keep their existing verdicts.

`docs/verification/runtime-backends.md` records the dated live measurement and
points at both portable regressions as the commands that refresh it.

## Validation

- `tests/fm-send-strict.test.sh`, `tests/fm-send-inbox.test.sh`,
  `tests/fm-send-settle.test.sh`, `tests/fm-send-popup-settle.test.sh`,
  `tests/fm-send-resolve-key.test.sh`, `tests/fm-send-secondmate-marker.test.sh`,
  `tests/fm-send-remote-delivery.test.sh`, `tests/fm-backend-herdr.test.sh` (184
  cases), `tests/fm-composer-ghost.test.sh`, `tests/fm-control.test.sh`,
  `tests/fm-backend-tmux-smoke.test.sh`, and
  `tests/fm-documentation-audiences.test.sh` all pass.
- `tests/fm-composer-lib.test.sh` passes apart from one pre-existing case,
  `test_matrix_herdr_halfblock_rule_bounds_bare_wrap`, which fails identically on
  an unmodified checkout here because it uses `$'▀'` and this machine's
  `bash` is 3.2 (the escape needs 4.2). `tests/fm-backend.test.sh` likewise has
  one pre-existing unrelated failure on an unmodified checkout. Both were
  confirmed against a clean tree; neither is touched by this change.

### DIFF
--- a/bin/fm-send.sh
+++ b/bin/fm-send.sh
@@ -57,7 +57,13 @@
 # type the literal
 # text through the target backend's verified submit core: typed ONCE, then
 # Enter retried (never retyped) until the backend confirms a submit or reports
-# an inconclusive send. Typed-plane exit contract: 0 = submit confirmed;
+# an inconclusive send. Before ANY of that the endpoint must not be a proven
+# dead one: a task's endpoint outlives the agent inside it, and typing into the
+# login shell an exited agent falls back to would hand the steer to that shell
+# instead of losing it visibly. fm_backend_agent_state (bin/fm-backend.sh) owns
+# that verdict fleet-wide and only its positive `dead` refuses, so an idle,
+# ghost-text, or busy composer is never blocked. Typed-plane exit contract:
+# 0 = submit confirmed;
 # 3 = the text was typed into the live endpoint and
 # Enter was sent, but the submit read-back stayed unconfirmed (verify the pane
 # before any resend, and never re-type blindly; a marked request's
@@ -862,6 +868,27 @@ else
       2) echo "fm-send: doorbell did not reach $T; the steer is durably recorded at $INBOX_RECORD and the watcher will re-ring" >&2 ;;
     esac
     exit 0
+  fi
+  # Refuse a dead endpoint BEFORE typing anything into it.
+  # The typed plane hands its text to whatever owns the terminal, and a task's
+  # endpoint outlives the agent inside it: once the agent exits, the pane falls
+  # back to the login shell that launched it. Typing there hands the steer to
+  # that SHELL, which runs the first token and discards the rest, so the
+  # instruction is gone while the endpoint still looks perfectly alive.
+  # The question "does this endpoint still hold an agent" already has one
+  # fleet-wide owner - fm_backend_agent_state (bin/fm-backend.sh) - so this
+  # consults it instead of adding a second classifier. Only its positive `dead`
+  # verdict refuses: `dead` means the endpoint exists and its foreground process
+  # group is confidently nothing but shells, which no live agent can produce
+  # (an agent running a child process reads `ambiguous`, never `dead`).
+  # `missing`, `ambiguous`, `unreadable`, and `unverified` all proceed exactly as
+  # before, because a false refusal here is as harmful as a false success: every
+  # supervision decision that assumes a steer landed is built on this exit.
+  if [ "$(fm_backend_agent_state "$TARGET_BACKEND" "$T")" = dead ]; then
+    fm_send_known_undelivered_cleanup || \
+      echo "error: known-undelivered pending-reply state could not be reset for $TARGET_TASK_ID" >&2
+    echo "error: text not sent to $T: the agent there is gone and the endpoint is now a bare shell, so nothing was typed (tried $RESOLUTION_TRIED). Relaunch the agent before re-sending; typing into the shell would run the message as a command and lose it." >&2
+    exit 1
   fi
   # Slash commands open a completion popup in some TUIs (verified on codex);
   # submitting too fast selects nothing, so give the popup time to settle before
--- a/docs/verification/runtime-backends.md
+++ b/docs/verification/runtime-backends.md
@@ -623,6 +623,22 @@ ok - real herdr: an agent that does not stop fails closed instead of being repor
 The registry read through `herdr pane report-agent` is the same source `fm_backend_herdr_agent_state` classifies, so registering and not registering an agent on a plain shell pane exercises exactly the gate every lifecycle verb depends on, with no real agent launched.
 That command is the guard that refreshes this record; run it after every Herdr upgrade rather than trusting the version above.
 
+### Steering a dead endpoint
+
+`bin/fm-send.sh`'s typed plane refuses an endpoint `fm_backend_agent_state` classifies `dead`, so that verdict is a delivery guarantee and not only a recovery one.
+Measured 2026-08-26 on Herdr 0.7.4 with Claude Code 2.1.246, in an isolated `fm-lab-` session, by launching claude in a pane, exiting it with `/exit`, and steering the pane in each state.
+
+| Pane state | `herdr agent get <pane>` | `fm_backend_agent_state` | `fm-send.sh <target> "<text>"` |
+| --- | --- | --- | --- |
+| Shell, no agent ever launched | `{"error":{"code":"agent_not_found",...}}` | `dead` | refused, exit 1, nothing typed |
+| Live idle claude | `{"agent":"claude","agent_status":"idle"}` | `alive` | delivered, exit 0 |
+| Claude mid-turn running a child process | `{"agent":"claude","agent_status":"working"}` | `alive` | delivered, exit 0 |
+| Shell after the agent exits | `{"error":{"code":"agent_not_found",...}}` | `dead` | refused, exit 1, nothing typed |
+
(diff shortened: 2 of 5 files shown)


kunchenguid:
Speaking as Kun's firstmate:

Reviewed HEAD `0f060896fa7db886c152f5c70e432efe62aa12d3` vs main `9ce69acf95b901cb6dc38dd5fa55f37a8038e16c`. Whole thread read (Greptile noted a check-to-send race). Full diff reviewed: `bin/fm-send.sh`, `docs/verification/runtime-backends.md`, `tests/fm-backend-herdr.test.sh`, `tests/fm-composer-lib.test.sh`, `tests/fm-send-strict.test.sh`. No `.github/workflows/*`. Not disguised security.

Class: **corrective**. Typed-plane send now refuses before typing when `fm_backend_agent_state` returns positive `dead`, so a steer is not handed to the login shell after the agent exits. Only `dead` refuses; missing/ambiguous/unreadable/unverified still proceed. Strengthens a real refusal path.

- One captain, one interface — aligns. Honest failure instead of silent shell execution / delivery-unknown.
- Authority is explicit — aligns. Narrow refuse on proven-dead only; no new consent assumption.
- Scripts own the mechanics — aligns. Consults existing backend liveness owner; no second classifier.
- A restart is a non-event — aligns.
- Delegation with a spine — aligns. Strengthens a refusal path; supervised delivery.
- The fleet outlives any vendor — aligns. Backend-state owned; Herdr measurement documented.
- Scope — aligns. Send-path refuse; not workshop growth.

Greptile's remaining TOCTOU (agent exits between check and type) is a known residual race, not a reason to reject the corrective refuse; not the blocker today.

**First-time fork CI:** after that full-diff review I approved workflow runs `32959702015` (CI) and `32959702094` (Require no-mistakes) for this HEAD. Forge behavior suites are green; Require no-mistakes **failed** — body was not raised through no-mistakes (missing signature + structured `no-mistakes-pipeline-attestation:v1` for this HEAD).

**Not merging.** Waiting on the author for a proper no-mistakes pipeline body (signature + matching attestation). Not with the captain.

Merge-eligible: NO.

kunchenguid:
Speaking as Kun's firstmate:

Reviewed HEAD `0f060896fa7db886c152f5c70e432efe62aa12d3` vs main `9ce69acf95b901cb6dc38dd5fa55f37a8038e16c`. Whole thread read. Full diff reviewed: `bin/fm-send.sh`, `docs/verification/runtime-backends.md`, `tests/fm-send-strict.test.sh`, `tests/fm-backend-herdr.test.sh`, `tests/fm-composer-lib.test.sh`. No `.github/workflows/*`. Not disguised security. Author is not blocked.

Class: **corrective**. A task endpoint outlives the agent; after exit the pane is a login shell and the typed plane was handing the steer to that shell (`zsh: no matches found: ...`) while reporting only unconfirmed delivery. `fm-send.sh` now consults `fm_backend_agent_state` before typing and refuses only a positive `dead`. `missing` / `ambiguous` / `unreadable` / `unverified` still proceed. Durable-inbox plane unchanged. Tests pin dead-refuses-and-types-nothing, live idle/working/blocked/unclassifiable still steered, and the composer classifier still reading a real zsh prompt as `unknown` (so this is not "fixed" in the wrong layer).

- One captain, one interface — aligns. A lost steer is no longer reported as "may have landed".
- Authority is explicit — aligns. Narrow dead-only refusal; no new capability.
- Scripts own the mechanics — aligns. Consults the existing backend owner; no second classifier.
- A restart is a non-event — aligns. No durable-state contract change.
- Delegation with a spine — aligns. Strengthens a refusal path before keystrokes go to a shell.
- The fleet outlives any vendor — aligns. Uses fleet-wide `fm_backend_agent_state`, not a vendor pixel.
- Scope — aligns. Send-path repair, not a new workshop layer.

On the remaining check-to-send window: I read the new guard immediately before `fm_backend_send_text_submit`. Neither submit core revalidates liveness, so an agent that exits in that interval can still be typed into. That is a residual TOCTOU on a check-then-act path, not the defect this PR set out to close (already-dead panes). I am not holding the PR for that residual. Closing it would mean changing the submit cores, which this change correctly left alone.

Forge CI on this HEAD is green. Require-no-mistakes failed: this PR was not raised through no-mistakes (no signature line, no `no-mistakes-pipeline-attestation:v1`). First-time fork CI was already approved on the prior pass; no new HEAD is waiting.

**Waiting on the author**, not the captain: re-push via `git push no-mistakes` so a matching `no-mistakes-pipeline-attestation:v1` for HEAD `0f060896fa7db886c152f5c70e432efe62aa12d3` lands and the required check can pass.

Merge-eligible: NO.

PR #3013, opened 2026-08-25
fix(bin): match harness command names case-insensitively in the session lock

ansellchiu:
## Intent

Make firstmate's harness-identity check recognize a capitalized harness process name, so a session running under the macOS Claude desktop app can take the home's session lock instead of being permanently read-only.

The bug: fm_harness_process_matches in bin/fm-session-lock-lib.sh matched the process basename against FM_HARNESS_RE case-sensitively. The macOS Claude desktop app reports its executable as 'Claude' (capital C), so the match failed, fm_harness_ancestry_pid found no harness, and bin/fm-lock.sh exited with 'error: cannot locate harness process in ancestry'. Confirmed on the captain's machine: the live ancestry contained a pid whose comm and args were both 'Claude', and printf 'Claude' | grep -qE "$FM_HARNESS_RE" does not match. Two observed consequences: every Claude-desktop session in the home was refused the lock and ran read-only (cannot spawn, steer, merge, or repair supervision); and bin/fm-claude-stop-autoarm.sh, which gates on fm_session_lock_owned_by_self and then on fm-lock.sh succeeding, silently exited on every firing, so supervision could never restart automatically and bin/fm-turnend-guard.sh alarmed on every turn end with no path to recovery.

Required change, exactly as specified: in fm_harness_process_matches, fold the basename to lower case before the FM_HARNESS_RE test and before the *claude* test that sets FM_HARNESS_IS_CLAUDE. Keep FM_HARNESS_RE itself unchanged so the ^pi$ and ^pi-signed$ anchors keep their exact current meaning. Deliberately do NOT widen any other evidence path: fm_harness_path_name, the bare-interpreter (node/python) branch, and the Cursor branch stay exactly as they are. This bug is only about the reported command name's case, so the narrow scope is an accepted decision, not an oversight.

Accepted constraints: this is firstmate shared tracked material under AGENTS.md section 1, so the firstmate-coding-guidelines skill was loaded and followed. The change must stay minimal and contained because this is a safety-critical identity check - do not refactor the surrounding functions. A short comment explaining why the fold exists (a macOS app bundle reports a capitalized executable name) is required so a later reader does not 'simplify' it away. Do not change the session-lock write semantics, the ancestry walk, or the auto-arm gating.

[...]

PR #2929, opened 2026-08-24
fix(bin): decide session-lock ownership by session cohort and make the turn-end fail-open reachable

adriantoczydlowski:
## Intent

GOAL: Fix the session-lock identity gap in the firstmate repo, and the unreachable turn-end fail-open it exposes. Two defects with one root cause, both measured on a live fleet on 2026-08-17.

DEFECT 1 - ownership is decided by the process tree, and the tree is not the session. fm_harness_ancestry_pids in bin/fm-session-lock-lib.sh walked the current shell's ancestry and fm_session_lock_owned_by_self asked whether state/.lock names any pid in that walk. Two live failures. (1a) Work running beside the session window is refused: a Claude background session launched from the same window runs under "claude bg-pty-host -> claude.exe", a tree that terminates at /init and never includes the window's own claude pid that holds the lock, so bin/fm-lock.sh reported "another live firstmate session holds the lock" and the whole session start degraded to read-only while exactly one session existed. Both processes carried the same HERDR_PANE_ID. (1b) A SUSPENDED holder never releases: after a session resume the window's previous "claude --resume" sat in state Tl (stopped by Ctrl-Z) on the same tty, and fm_harness_pid_alive used kill -0, which succeeds on a stopped process, so the lock stayed pointed at something that would never let go and every subsequent session was refused.

DEFECT 2 - the turn-end safety valve cannot fire when the auto-arm never claimed. bin/fm-claude-stop-autoarm.sh used the same ownership test, so under defect 1 it never claimed the home and supervision never started. bin/fm-turnend-guard.sh then blocked every turn end while work was in flight, and its bounded attended fail-open was unreachable, because docs/turnend-guard.md made that path require an auto-arm that recorded an exhausted failure - and here the auto-arm never ran at all, so there was no failure episode to advance. The block therefore repeated until Claude's own 8-block override. The captain treats this as the MORE URGENT of the two: a correctness gap in identity detection must not be able to turn into an unbounded block, so the valve has to be reachable whenever supervision cannot start for a reason the guard itself cannot repair, however that reason arose, and it must stand on its own so that even an imperfect fix to defect 1 cannot wedge a session again.

[...]

Issue #1933, opened 2026-08-08
Fix Codex Desktop fleet-lock recognition


butlersrepos:
Reproduction: in a managed Codex Desktop session, run bin/fm-session-start.sh.

Expected: the session identifies itself as Codex, safely reclaims a dead lock owner, and begins normal supervision.

Observed: bin/fm-harness.sh reports unknown; bin/fm-lock.sh refuses because it cannot locate a harness process in ancestry. The recorded lock PID is dead, while this environment exposes CODEX_THREAD_ID and CODEX_CI=1.

Scope: add a verified, tightly scoped Codex Desktop identity marker to harness and lock ownership detection. Preserve fail-closed behavior for unmanaged shells and live competing lock owners. Add regression coverage.

aoline:
## Local fix details

A local fix has been implemented and verified in an isolated clone.

### Patch scope

- Recognize managed Codex Desktop from the exact `CODEX_CI=1` plus UUID-shaped `CODEX_THREAD_ID` markers.
- Preserve the numeric `state/.lock` PID format.
- Add a Codex thread-identity sidecar so ownership survives short-lived tool subprocesses.
- Reclaim dead owners while refusing competing live or different Codex thread identities.
- Add harness and session-lock regression coverage.

### Verification

- `bash -n` passed for all changed shell scripts.
- Isolated temporary-home session startup acquired the lock and rendered Codex supervision.
- Full suite remains environment-blocked because sandboxed `ps` is denied.
- ShellCheck 0.11.0 is unavailable in the current environment.

The patch is committed locally as `7e61351` on `codex/codex-desktop-lock`.

Opening a PR is currently blocked because the authenticated `aoline` account has no push permission to `kunchenguid/firstmate`, and no `aoline/firstmate` fork exists.

kunchenguid:
Speaking as Kun's firstmate:


Verified against main `f7a387f44ed4fa942914e8944be4981a42e1e102`.

**VISION:** aligns — The fleet outlives any vendor; contracts bind to semantics a vendor actually exposes, not to today's pixels. Codex Desktop publishes `CODEX_CI=1` and a UUID-shaped `CODEX_THREAD_ID`. A restart is a non-event only if a dead lock owner can be reclaimed and a live session can identify itself. Fail-closed for unmanaged shells and live competing owners must stay.

[...]

PR #3122, opened 2026-08-26
fix(bin): record the harness-declared session pid as the session lock identity

tomglenn:
## The failure

A background agent could write a session lock naming a process that was not its own, then permanently fail to recognise that lock as its own once that process exited. From then on it behaved as though another session held the home.

The cause is that the lock writer identified "this session" by walking to the outermost process in the contiguous harness ancestry. Two arrangements produce ancestries the process table cannot tell apart:

- a session launched directly, where the outermost harness process *is* the session, and
- a session hosted by a shared harness daemon, where the outermost harness process is the daemon and the session is below it.

The walk answers "outermost" in both cases. In the first that is right. In the second it records the shared daemon, which outlives no single session and eventually exits, stranding the agent below it.

## The change

`bin/fm-session-lock-lib.sh` now prefers the harness's own declaration of which process is its session (`CLAUDE_PID`) over the ancestry walk, because the harness knows what the process table cannot infer.

That declaration is trusted only when both hold:

- the declared pid is alive, and
- it is a member of *this* session's contiguous harness ancestry.

An inherited value from an unrelated session, a stale pid, a live process outside this ancestry, and a malformed value are all inert: each falls through to the previous behaviour. `tests/fm-session-lock-ancestry.test.sh` covers each of those rejection paths with the accepted and rejected answers driven to deliberately different pids, so an assertion cannot pass for both branches.

**The ancestry fallback is unchanged for every harness that declares nothing.** For those, the walk stops at the first match, so outermost and innermost are the same process and the overreach never existed.

## Measurement

Claude Code 2.1.241, a print-mode session and a real background agent, each launched with a deliberately wrong `CLAUDE_PID`:

- both reported their own process, not the planted value;
- without the declaration, the background agent would have recorded the shared process above it.

Recorded in `docs/verification/supervision.md`, refreshed by the opt-in guard `tests/fm-session-lock-declaration-live-e2e.test.sh`. That guard needs a real harness, so it is env-gated and self-skipping rather than part of the default suite.

## What to look at

- `bin/fm-session-lock-lib.sh` â the two guards that make the declaration safe to trust.
- `tests/fm-session-lock-ancestry.test.sh` â the rejection-path coverage, and the three-level launcher/daemon/session shape that reproduces the original failure.
- `bin/fm-test-run.sh` â the changed-path to family map keeps its over-selecting contract for the lock library.


kunchenguid:
Speaking as Kun's firstmate:

Reviewed HEAD `7223eb894360849ae4850fee235ed929bcb9a152` vs main `9ce69acf95b901cb6dc38dd5fa55f37a8038e16c`. Whole thread read (Greptile success). Full diff reviewed: `bin/fm-session-lock-lib.sh`, `bin/fm-lock.sh`, `bin/fm-test-run.sh`, docs/verification, and the ancestry + live-declaration tests. No `.github/workflows/*`. Not disguised security. tomglenn is not blocked.

Class: **corrective**. Under a session-hosting daemon the ancestry walk's outermost pid can record the launching session; this prefers harness-declared `CLAUDE_PID` only when numeric, live, and a member of this contiguous harness ancestry, otherwise the previous outermost fallback. Self-ownership predicate is untouched.

- One captain, one interface — aligns. Stops a background session permanently misreading its own lock as a competitor (body + three-level fixture).
- Authority is explicit — aligns. Declaration is inert unless live+in-ancestry; no new captain-facing capability.
- Scripts own the mechanics — aligns. Deterministic identity resolution in `fm_harness_ancestry_pid`.
- A restart is a non-event — aligns. Lock identity that outlives the wrong ancestor is the durability fix.
- Delegation with a spine — aligns. Restores Stop auto-arm self-ownership so supervision can claim.
- The fleet outlives any vendor — aligns. Uses Claude's declared session pid; harnesses that declare nothing keep the walk.
- Scope — aligns. Session-lock identity library, not workshop.

**Lock-identity class (classified, not a standing pair-hold):** told apart from open #2929 (session-cohort / HERDR_PANE_ID ownership + suspended holder + turn-end fail-open), open #2964 (omp Claude identity), closed #3013 (case-insensitive basename), and ready-for-pr issue #1933 (Codex Desktop). This PR is the narrow declared-pid writer fix. Not landing only because attestation/CI gates are not met — not because of an untellable overlap hold.

**First-time fork CI:** after that full-diff review I approved workflow runs `32990820159` (CI) and `32990820110` (Require no-mistakes) for this HEAD.

**Not merging.** Waiting on the author: body has no `no-mistakes-pipeline-attestation:v1` matching this HEAD (Require-no-mistakes failed). Re-run `git push no-mistakes` so the structured stamp lands. Not with the captain.

Merge-eligible: NO.

PR #3144, opened 2026-08-27
fix: atomically pair backlog and task record transitions

kunchenguid:
## Intent

Make Firstmate's backlog<->meta records atomically consistent by construction, so the three record-disagreement classes an earlier investigation catalogued (orphan_in_flight, unowned_current, terminal_in_flight) cannot arise from the ordinary dispatch/completion path, and add durable terminal-close recovery for the one class that can still occur across a crash.

This implements the recommendation of a completed scout investigation (data/fm-secondmate-record-consistency-scout-s1/report.md, private to the operator's home). The design decision is already made and is NOT open for re-litigation: do not propose a distributed transaction, two-phase commit, or a new journal subsystem - the report explains why none fits a single-writer/single-host/single-filesystem-per-task system.

THE INVARIANT ENFORCED: the script that performs the mechanical physical-state change (create or remove state/<id>.meta) must be the SOLE owner of the paired backlog transition, executed atomically in the same process under the same per-task lock it already holds, BEFORE it reports success - never a second, independently-timed step left to a future agent turn, an LLM's memory, or a printed reminder. Before this change the backlog half was prose in AGENTS.md telling a later agent turn to run `tasks-axi start`/`tasks-axi done`, with zero durable tracking if that second command never happened.

REQUIRED CHANGES, all implemented:
1. Dispatch (bin/fm-spawn.sh): after it durably writes state/<id>.meta, and under the per-task meta lock, fold the backlog In-flight transition into the spawn itself as the last step before declaring success. A failed transition fails the spawn loudly with an explicit error rather than leaving a silently orphaned meta file. The relaunch path re-verifies (does not assume) backlog state rather than blindly re-running the transition.
2. Completion (bin/fm-teardown.sh): in the same guarded section that removes state/<id>.meta, fold the backlog completion transition (tasks-axi done with the right --pr/--report/--note from $PR_URL, $KIND, $MODE, all already resolved locally) into teardown, under the meta lock, before it reports success. The existing backlog-refresh reminder becomes a confirmation echo of what teardown already durably did, not a load-bearing instruction. Teardown's existing unlanded-work / landed-work safety gates are NOT weakened: the transition is folded in only on the paths that already proceed to remove meta.
3. Same-home self-heal on restart (bin/fm-bootstrap.sh): a same-home, self-triggered reconciliation added to the existing bootstrap MUTATING sweeps (already run under the session lock) so a home that crashed mid-transition heals its OWN books on its own restart instead of waiting for a parent's cross-home nudge cooldown. Same-home only: it never reaches into another home, and it neither duplicates nor replaces the existing cross-home reconcile-nudge (bin/fm-secondmate-reconcile.sh) or the fleet-snapshot classifier (bin/fm-fleet-snapshot.sh) - those deliberately stay as defense-in-depth backstops.
4. Respect the configured backlog backend: when config/backlog-backend selects `manual`, routine backlog transitions are hand-edited, so the folded transition must not hard-fail a spawn/teardown on a manual-backend home. It follows the existing backend-selection contract (bin/fm-tasks-axi-lib.sh / docs/configuration.md) rather than calling tasks-axi unconditionally.

CONCURRENCY AND LOCKING: reuse the per-id locks the codebase already has (fm_meta_lock_path, and the reconcile_lock/control_lock/meta_lock pattern). No new lock namespaces are to be invented. The fused tasks-axi call runs inside the lock the mechanical step already holds, so a steer racing a teardown on the same id stays serialized exactly as before.

SCOPE DISCIPLINE: orphan_in_flight and unowned_current must be prevented by construction on the ordinary path. terminal_in_flight shrinks to a millisecond crash window inside one script; the self-heal sweep plus the existing detect/nudge backstop cover the residual. The markerless-remote-mate gap is explicitly OUT OF SCOPE - it is owned by a separate authorized ship (fm-secondmate-reconcile-missing-spawn-gen-r1) and must not be folded in.

DESIGN DECISIONS MADE WHILE IMPLEMENTING (deliberate, not accidents):
- A new shared library bin/fm-backlog-transition-lib.sh is the single owner of the invariant and of the gate that decides when a transition applies. It deliberately does not source bin/fm-tasks-axi-lib.sh, matching this repo's flat "libs never source libs" convention; callers source both.
- The transition is skipped, never failed, for: kind=secondmate (persistent agents are never backlog items, AGENTS.md section 10 - this matches the pre-existing scope of the teardown reminder exactly), a config/backlog-backend=manual home, a home without compatible tasks-axi, and a home that keeps no data/backlog.md at all. The last of those is what keeps ad-hoc and fixture homes working: a home with no backlog file has no books to keep consistent.
- Dispatch REFUSES up front - before any endpoint, worktree, or record exists - when the home's backlog has no item for the id, or already records it as finished. This is deliberate strictness: it is what makes unowned_current structurally impossible rather than merely unlikely, and refusing at the cheapest possible point means there is nothing to unwind. The error names the exact `tasks-axi add` / `tasks-axi reopen` command to run. Refusing a `done` row rather than silently reopening it also avoids a tasks-axi round-trip quirk that pollutes the item title. The row is re-read under the lock, so an item closed while the endpoint was being stood up is refused rather than resurrected.
- Teardown records the intended close in state/<id>.backlog-close BEFORE removing the record, because the completion links (PR, report path, local-main note) live only in the record being removed - a process killed between the halves would otherwise leave nothing to reconstruct the close from. The marker is stamped with the record's own incarnation so replay can tell an interrupted cleanup from a later redispatch. A landed close removes that record. This is one small marker file, deliberately not a journal subsystem. Replay is idempotent because `tasks-axi done` on an already-closed task backfills links without moving the close date.
- The bootstrap sweep heals only unambiguous cases: it finishes an interrupted cleanup of the very incarnation it recorded, drops a close whose id has since been dispatched again, and moves In flight any queued, non-held item this home already owns a record for. It deliberately never touches a captain-held item (that is the captain's to move) or a closed one (it must not resurrect an item), never blocks on a record a live action owns, never reaches into another home, and costs nothing on a home that owns no records.
- tasks-axi is always addressed with an explicit `--file <data>/backlog.md` and run from that data directory's PARENT, so the mutation lands in the home that owns the task regardless of the caller's working directory, and that home's own .tasks.toml supplies done_keep and the archive path. The parent of the data directory is the addressing root rather than FM_HOME so a relocated data directory keeps its backlog and archive together - and so a test fixture cannot mutate the developer's real backlog. The scout report link is derived from the configured data directory relative to that addressing root, not hardcoded as data/<id>/report.md.
- On a failed dispatch transition the just-published record AND the busy generation armed alongside it are both removed, so the unwind is complete rather than half-done. A relaunch instead preserves its pre-existing record and fails loudly, because that record and its work predate the run.

TESTS: new tests/fm-backlog-atomicity.test.sh drives the real fm-spawn.sh, fm-teardown.sh, and fm-bootstrap.sh against a real backlog and the real tasks-axi, asserting observable record state - never reminder wording. It covers: dispatch moving the item In flight in the same run; dispatch refusing an id with no item and an already-closed item; a failed transition failing loudly and leaving no record or busy state; dispatch not resurrecting a row closed after preflight; completion closing local-only and scout items with their landing note and report; completion refusing to report success while its item is open and recording the close it owes; the recovery cases (replay a recorded close for the matching incarnation, drop a stale one, preserve a pending close across a transient backlog read failure, preserve both records when meta removal fails, mark an owned record In flight); leaving a captain-held item alone; a manual-backlog home dispatching and completing without a hard failure; a secondmate home keeping its OWN books through dispatch and completion (the invariant is single-host, so a parent<->mate freshness delay is deliberately NOT asserted as a correctness violation); and a persistent secondmate needing no backlog item at all. Relaunch cases in tests/fm-control-relaunch.test.sh prove the relaunch path re-reads the row instead of re-running the transition, and heals a drifted one.
Two pre-existing assertions in tests/fm-teardown.test.sh that only checked the printed reminder TEXT were deliberately replaced with assertions on the resulting backlog state - that weak coverage is precisely what let the two records drift. That fixture was also pinned to its own data directory so a test can never mutate the operator's real backlog.

Documentation is updated at its owners: the three script headers, AGENTS.md (state layout, dispatch/completion no longer a separate agent step, the mutating sweeps list), docs/configuration.md's Backlog backend section, docs/scripts.md, and the bootstrap-diagnostics skill for the new BACKLOG_RECONCILE diagnostic line. bin/fm-secondmate-reconcile.sh's header now states it is a backstop rather than the primary mechanism.

Constraints: changed bash must be stock macOS bash 3.2 compatible; bin/*.sh must pass bin/fm-lint.sh and shellcheck with the source=/dev/null boundary; repo style is one sentence per line in tracked Markdown, plain dash never an em dash, and no agent co-author on commits. Delivery is no-mistakes with merge authority held by the captain: do not merge.

## What Changed

- Move backlog dispatch and completion transitions into the locked spawn and teardown record lifecycle, with strict dispatch validation and failed-transition cleanup.
- Persist pending completion markers and reconcile interrupted closes or drifted owned records during same-home bootstrap.
- Respect manual, secondmate, incompatible-tool, and backlog-free exemptions, with expanded lifecycle regression coverage and documentation.

## Risk Assessment

✅ Low: The lifecycle changes consistently preserve ownership on uncertain rollback, pair record transitions under existing per-task locks, and provide validated durable close recovery without a substantiated new defect.

## Testing

Inspected the base-to-target change, then exercised real tasks-axi-backed dispatch, completion, crash recovery, relaunch, manual/secondmate exemptions, failure rollback, and teardown safety behavior; all targeted end-to-end checks succeeded, reviewer-visible CLI transcripts were captured, and no transient worktree files remained.

<details>
<summary>Evidence: Backlog atomicity end-to-end transcript</summary>

Source: [Backlog atomicity end-to-end transcript](https://github.com/kunchenguid/firstmate/blob/3d2292e40e186d92afdd6a776ebdfb7ab95fffca/.no-mistakes/evidence/fm/fm-secondmate-record-atomic-transitions-r1/fm-backlog-atomicity-transcript.txt)

```text
FM_TEST_BEGIN 2026-08-29T22:22:16Z tests/fm-backlog-atomicity.test.sh family=session-bootstrap expected_gate_skip=none
ok - dispatch publishes the record and moves the backlog item In flight in one run
ok - dispatch reads backlog rows from the backlog addressing root
ok - recovery uses the parent of a trailing-slash data record
ok - completion targets nested relative data from the caller directory
ok - an immediate-child absolute data path keeps one paired backlog
ok - bare relative data addresses one backlog through dispatch and completion
ok - dispatch refuses an unresolvable backlog data directory
ok - completion refuses before mutation when backlog data is unresolvable
ok - dispatch refuses, before creating anything, when the home has no item for the id
ok - dispatch distinguishes backlog read failures from missing items
ok - dispatch refuses a closed item instead of silently reopening it
ok - dispatch cannot commit without a verified task-record publication
ok - a failed backlog transition fails the dispatch loudly and leaves no record
ok - dispatch reports when failed-transition rollback cannot remove its record
ok - dispatch verifies both task and busy records during rollback
ok - dispatch commits neither record nor backlog state before launch delivery succeeds
ok - interrupted Kimi delivery preserves durable ownership for every backlog backend
ok - failed Kimi delivery stops workers or retains their ownership
ok - unqueryable failed dispatch retains durable ownership
ok - interrupted launch submission retains unconfirmed ownership
ok - dispatch retries interrupted transitions before honoring termination
ok - dispatch does not resurrect a row closed after preflight
ok - dispatch fails when its backlog row vanishes after preflight
ok - completion closes a local-only ship, with its landing note, before reporting success
ok - completion closes a scout item against its report
ok - completion leaves legacy records open when no incarnation can be recorded
ok - completion records relocated scout reports relative to the backlog root
ok - space-containing scout report paths round-trip through recovery
ok - non-ASCII close markers replay across process locales
ok - control-byte data paths fail closed before paired transitions
ok - unreplayable data paths are refused before destructive cleanup
ok - completion preserves recovery state when task-record removal fails
ok - completion refuses to report success while its item is still open, and records what it owes
ok - completion reports failure until its durable close marker is removed
ok - session start retries a close whose marker could not be removed
ok - session start reports owned backlog rows it cannot read
ok - Orca cleanup recovery is excluded from backlog lifecycle transitions
ok - session start marks an item In flight when this home already owns a worker for it
ok - session start finishes a close an interrupted cleanup recorded but never landed
ok - session start preserves a pending close across a transient backlog read failure
ok - session start finishes a close for the matching meta incarnation
ok - recovery preserves both records when meta removal fails
ok - recovery binds close-marker identity to its locked filename
ok - recovery rejects close markers targeting another home's data
ok - recovery validates an unterminated final marker field
ok - recovery rejects lexical traversal before resolving marker data
ok - recovery rejects marker control bytes before parsing
ok - recovery rejects malformed PR URL values
ok - a retained pending close is never started by reconciliation
ok - recovery rejects a bare-parent report path
ok - recovery rejects close-marker arguments outside its protocol
ok - recovery reports and rejects dangling symlink close markers
ok - session start drops a close recorded for an older meta incarnation
ok - session start rejects an unversioned close marker
ok - bootstrap stops when backlog data disappears before reconciliation
ok - bootstrap preserves secondmate, manual, and absent-backlog exemptions
ok - session start leaves a captain-held item where the captain put it
ok - a home with no backlog remains exempt from lifecycle transitions
ok - interrupted backlog-exempt spawns retain durable ownership after launch submission
ok - a manual-backlog home dispatches and completes without a hard failure
ok - a secondmate home keeps its own books paired through dispatch and completion
ok - dispatching a persistent secondmate needs no backlog item
FM_TEST_END 2026-08-29T22:25:09Z tests/fm-backlog-atomicity.test.sh exit=0 duration_ms=172950 gate_skip=false
FM_TEST_SUMMARY total=1 failed=0 skipped_gate=0 duration_ms=173016
FM_TEST_SUMMARY_FAMILY family=session-bootstrap count=1 duration_ms=172950 failed=0
FM_TEST_SLOWEST rank=1 script=tests/fm-backlog-atomicity.test.sh duration_ms=172950
fm-test-run: wrote timing artifact: ~/.no-mistakes/evidence/01M15B2X4TSJ4HEJ732WBPZRNP/fm-backlog-atomicity-results.json
```
</details>
<details>
<summary>Evidence: Relaunch backlog-state verification transcript</summary>

Source: [Relaunch backlog-state verification transcript](https://github.com/kunchenguid/firstmate/blob/3d2292e40e186d92afdd6a776ebdfb7ab95fffca/.no-mistakes/evidence/fm/fm-secondmate-record-atomic-transitions-r1/fm-control-relaunch-transcript.txt)

```text
FM_TEST_BEGIN 2026-08-29T22:25:13Z tests/fm-control-relaunch.test.sh family=backend-dispatch expected_gate_skip=none
ok - fm-control relaunch: a same-harness relaunch replaces the agent in the same endpoint and worktree
ok - fm-control relaunch: durable task metadata survives replacement launch publication
ok - fm-control relaunch: delivery and concurrent task metadata publication serialize
ok - fm-control relaunch: disabling tracing clears metadata and pane context
ok - fm-control relaunch: the progress note lands in the instructions the replacement reads
ok - fm-control relaunch: a ship task refuses without the progress note its replacement needs
ok - fm-control relaunch: switching harness is one ordinary relaunch, and the old wiring goes with the old agent
ok - fm-control relaunch: a harness switch resets model and effort unless they are named too
ok - fm-control relaunch: a prefixed recorded harness can switch adapters transactionally
ok - fm-control relaunch: a prefixed command requires an explicit replacement harness
ok - fm-control relaunch: a same-harness relaunch keeps the profile axes it was running with
ok - fm-control relaunch: explicit model and effort win over the recorded ones
ok - fm-control relaunch: refuses to relaunch onto an adapter with no verified mechanics
ok - fm-control relaunch: the retired incarnation's global turn-end token is revoked
ok - fm-control relaunch: wiring cleanup failure refuses replacement arming
ok - fm-control-lib: one owner resolves each harness's turn-end registry entry, and refuses a malformed token
ok - fm-control relaunch: a secondmate relaunch re-resolves its durable configured harness pin
ok - fm-control relaunch: invalid configured effort is ignored before stop
ok - fm-control relaunch: an adapter unverified for this task kind refuses before the agent is stopped
ok - fm-control relaunch: explicit secondmate harness resets unnamed profile axes
ok - fm-control relaunch: a ship task keeps its recorded harness instead of re-reading crew config
ok - fm-spawn --relaunch: with no explicit harness it reuses the task's recorded one, never the crew default
ok - fm-spawn --relaunch: wiring armed under a prefixed harness name is still retired
ok - fm-spawn --relaunch: switching away from muse retires its session binding
ok - fm-spawn --relaunch: switching away from cursor retires its session binding
ok - fm-control relaunch: an unaccountable local copy refuses before the agent is touched
ok - fm-control relaunch: a worker with nothing to work from is never launched
ok - fm-control relaunch: a refusal before the agent is stopped leaves the durable record untouched
ok - fm-control relaunch: checkpoint inspection failures refuse before stopping
ok - fm-control relaunch: a launch failure after the stop keeps the prior record and reports the real state
ok - fm-control relaunch: unpublished rollback keeps concurrent durable metadata
ok - fm-control relaunch: post-publication failure keeps the new durable record
ok - fm-control relaunch: partial stop reconciles actual agent state
ok - fm-control relaunch: failed journal replacement preserves durable phase
ok - fm-spawn relaunch: prepublication abort removes replacement state
ok - fm-control relaunch: the checkpoint records the exact unlanded work it preserved
ok - fm-control relaunch: a secondmate's child work is accounted for and its charter is left alone
ok - fm-control relaunch: a secondmate home that is not this secondmate's is refused
ok - fm-control relaunch: unreadable and untraversable child state fails checkpoint
ok - fm-control relaunch: two control actions on one task serialize instead of interleaving
ok - fm-spawn relaunch: direct entry participates in lifecycle serialization
ok - fm-promote: promotion participates in lifecycle serialization
ok - fm-spawn --relaunch: refuses to launch a second agent into a live endpoint
ok - fm-spawn --relaunch: every identity axis comes from the record, and a contradicting flag refuses
ok - fm-spawn --relaunch: an unrecorded task is refused
ok - fm-spawn --relaunch: refuses to start a replacement outside the copy holding the work
ok - relaunch re-reads the backlog item instead of blindly re-running the transition
ok - relaunch heals an item that drifted out of In flight while the task stayed live
FM_TEST_END 2026-08-29T22:26:28Z tests/fm-control-relaunch.test.sh exit=0 duration_ms=75013 gate_skip=false
FM_TEST_SUMMARY total=1 failed=0 skipped_gate=0 duration_ms=75074
FM_TEST_SUMMARY_FAMILY family=backend-dispatch count=1 duration_ms=75013 failed=0
FM_TEST_SLOWEST rank=1 script=tests/fm-control-relaunch.test.sh duration_ms=75013
fm-test-run: wrote timing artifact: ~/.no-mistakes/evidence/01M15B2X4TSJ4HEJ732WBPZRNP/fm-control-relaunch-results.json
```
</details>
<details>
<summary>Evidence: Completion and teardown safety transcript</summary>

Source: [Completion and teardown safety transcript](https://github.com/kunchenguid/firstmate/blob/3d2292e40e186d92afdd6a776ebdfb7ab95fffca/.no-mistakes/evidence/fm/fm-secondmate-record-atomic-transitions-r1/fm-teardown-transcript.txt)

```text
FM_TEST_BEGIN 2026-08-29T22:26:34Z tests/fm-teardown.test.sh family=pr-forge expected_gate_skip=none
ok - local-only worktree with HEAD on a fork remote is torn down and the home summary is refreshed
●  1 task(s) in flight, but no live watcher process holds this home lock (last beat: 1s ago).
●  repair a missing or failed watcher cycle with the Pi tool fm_watch_arm_pi, or restart Pi with -e ~/.no-mistakes/worktrees/52b07e9083e7/01M15B2X4TSJ4HEJ732WBPZRNP/.pi/extensions/fm-primary-turnend-guard.ts -e ~/.no-mistakes/worktrees/52b07e9083e7/01M15B2X4TSJ4HEJ732WBPZRNP/.pi/extensions/fm-primary-pi-watch.ts if the extensions are not loaded.
ok - teardown closes its own backlog item before reporting success
●  1 task(s) in flight, but no live watcher process holds this home lock (last beat: 0s ago).
●  repair a missing or failed watcher cycle with the Pi tool fm_watch_arm_pi, or restart Pi with -e ~/.no-mistakes/worktrees/52b07e9083e7/01M15B2X4TSJ4HEJ732WBPZRNP/.pi/extensions/fm-primary-turnend-guard.ts -e ~/.no-mistakes/worktrees/52b07e9083e7/01M15B2X4TSJ4HEJ732WBPZRNP/.pi/extensions/fm-primary-pi-watch.ts if the extensions are not loaded.
ok - teardown honors config/backlog-backend=manual and still finishes cleanly
ok - local-only worktree with truly unpushed work is refused (safety preserved)
ok - local-only worktree with work merged into local main is torn down (no regression)
ok - no-mistakes worktree with HEAD on origin is torn down (no regression)
ok - no-mistakes worktree with genuinely unlanded work is refused (safety preserved)
ok - local-only worktree with unpushed work is torn down under --force (escape hatch)
ok - teardown completes when an exact busy-state sidecar is already absent
ok - herdr teardown removes pane-owned escalation dedupe state
ok - herdr flat teardown refuses before returning the isolated copy under lock contention and the retry completes cleanly
ok - herdr flat teardown never erases records when pane presence is unparseable
ok - herdr flat teardown preflight refuses before every destructive change
ok - forced secondmate teardown preflights every Herdr child before cleanup mutation
ok - forced secondmate teardown holds every descendant lifecycle and metadata lock
ok - forced secondmate teardown retains Herdr child identity until exact pane disappearance
ok - forced teardown retains a nested secondmate home and its grandchild's Herdr identity when the grandchild close is unconfirmed
ok - herdr projection teardown retires its journal only after confirming the exact recorded pane is gone
ok - herdr projection teardown retains every record when post-close presence is unknown
ok - herdr projection teardown surfaces failed focus restoration without turning confirmed cleanup into a hard failure
ok - squash-merged + deleted-branch worktree (PR merged) is torn down (the fix)
ok - squash-merged PR accepts a local HEAD that is an ancestor of the final PR head
ok - teardown discovers a merged PR by branch name and tears down when no pr= was ever recorded
ok - squash-merged PR accepts replayed unpushed local patches contained in the PR head
ok - merged PR does not allow teardown after a later local commit
●  1 task(s) in flight, but no live watcher process holds this home lock (last beat: 0s ago).
●  repair a missing or failed watcher cycle with the Pi tool fm_watch_arm_pi, or restart Pi with -e ~/.no-mistakes/worktrees/52b07e9083e7/01M15B2X4TSJ4HEJ732WBPZRNP/.pi/extensions/fm-primary-turnend-guard.ts -e ~/.no-mistakes/worktrees/52b07e9083e7/01M15B2X4TSJ4HEJ732WBPZRNP/.pi/extensions/fm-primary-pi-watch.ts if the extensions are not loaded.
WARNING: watcher still down (same stale episode; last beat: 1s ago, grace 300s) - full banner already printed this episode.
ok - fm-pr-check does not refresh PR head after HEAD moves
●  1 task(s) in flight, but no live watcher process holds this home lock (last beat: 1s ago).
●  repair a missing or failed watcher cycle with the Pi tool fm_watch_arm_pi, or restart Pi with -e ~/.no-mistakes/worktrees/52b07e9083e7/01M15B2X4TSJ4HEJ732WBPZRNP/.pi/extensions/fm-primary-turnend-guard.ts -e ~/.no-mistakes/worktrees/52b07e9083e7/01M15B2X4TSJ4HEJ732WBPZRNP/.pi/extensions/fm-primary-pi-watch.ts if the extensions are not loaded.
ok - fm-pr-check records the remote PR head when the local worktree lags
ok - worktree whose content already landed in the default branch is torn down (content fallback)
ok - content fallback refreshes origin default before comparing trees
ok - dirty worktree is refused even when its committed work has landed (dirty always wins)
ok - gh lookup error with content not in default refuses (fail-safe)
ok - provably-stale worktree index.lock (old, no live holder) is cleared and teardown succeeds
ok - live-held worktree index.lock is never removed and teardown refuses
ok - lsof errors leave worktree index.lock in place and refuse teardown
ok - stale lock cleanup rechecks and refuses dirty worktree before return
ok - normal repo index.lock is resolved from the worktree and cleared when stale
ok - lock mtime read failures leave worktree index.lock in place and refuse teardown
ok - transient index.lock cleared after first failed return is retried successfully without force-remove
ok - persistent index.lock exhausts retries and refuses without force-removing the lock
ok - empty retry wait overrides use the default without aborting teardown
ok - fractional legacy retry wait remains supported without arithmetic
ok - a task's own parked no-mistakes run is aborted, not orphaned, before the worker is removed
ok - teardown refuses before reap or removal when a task-owned run remains parked
ok - a different run cannot confirm the targeted abort
ok - empty post-abort status is not accepted as confirmation
ok - the CLI's exact run-not-found signal confirms completion
ok - a parked run on another branch is never aborted by this task's teardown (ownership is precise)
ok - a task-owned autonomous running step is left alone rather than aborted
ok - a leaked descendant process rooted under the task's worktree is reaped by teardown, not left surviving
ok - a leaked descendant process rooted under the task's per-task tasktmp is reaped by teardown too
ok - missing lsof falls back to reaping the tmux pane process group
ok - an erroring lsof scan refuses teardown and preserves the task
ok - a reused pid with a different start time is never force-killed
ok - an exec change preserves birth identity and the process is reaped
ok - a process spawned during grace is reaped on a later pass
ok - persistent leaked processes refuse teardown after bounded retries
ok - a process exiting during identity lookup does not block teardown
ok - the run abort and the leaked-process reap both complete before the destructive worktree return
FM_TEST_END 2026-08-29T22:29:09Z tests/fm-teardown.test.sh exit=0 duration_ms=154732 gate_skip=false
FM_TEST_SUMMARY total=1 failed=0 skipped_gate=0 duration_ms=154792
FM_TEST_SUMMARY_FAMILY family=pr-forge count=1 duration_ms=154732 failed=0
FM_TEST_SLOWEST rank=1 script=tests/fm-teardown.test.sh duration_ms=154732
fm-test-run: wrote timing artifact: ~/.no-mistakes/evidence/01M15B2X4TSJ4HEJ732WBPZRNP/fm-teardown-results.json
```
</details>

## Pipeline

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>
<summary>🔧 **Review** - 2 issues found → auto-fixed (25) ✅</summary>

- 🚨 `bin/fm-backlog-transition-lib.sh:186` - The required design says “a landed close removes that record,” but every marker deletion suppresses failure and teardown reports success. If marker removal fails after `tasks-axi done`, reopening the same ID leaves a queued row with an old marker and no meta; the next bootstrap replays that old close and incorrectly closes the reopened task. Make marker removal verify absence and propagate failure through teardown/replay instead of returning `closed` or `stale`.
- 🚨 `bin/fm-spawn.sh:797` - The required failed-dispatch invariant says the just-published record and busy generation are both removed, but abort cleanup clears `SPAWN_FRESH_COMMIT_PENDING` before attempting an unchecked `rm`. The commit-failure path at line 3040 likewise cannot recover from persistent unlink failure. With a failed `tasks-axi start` plus an unwritable/unremovable meta, spawn exits with the row queued while `state/&lt;id&gt;.meta` remains, recreating `unowned_current`. Centralize rollback and verify record removal before claiming cleanup succeeded.

🔧 Fix: Verify rollback and close-marker removal
2 warnings still open:

- ⚠️ `bin/fm-spawn.sh:1967` - The preflight collapses every `tasks-axi show` failure into an empty state and then claims the item is missing. For an unreadable or transiently locked backlog containing the item, spawn gives an incorrect `tasks-axi add` remedy that cannot succeed. Branch on `FM_BACKLOG_ROW_RESULT` as the commit-time check already does, and surface `FM_BACKLOG_ROW_ERROR` for genuine read failures.
- ⚠️ `bin/fm-spawn.sh:2972` - Ignoring HUP/INT/TERM here is inherited by every directly executed delivery command and remains active through Kimi&#39;s readiness and delivery polling, potentially about 50 seconds, plus the unbounded backlog mutation. A hung backend or tasks-axi child therefore cannot be terminated normally and keeps the per-task meta lock indefinitely. Preserve the atomic commit behavior without passing ignored signal dispositions to fallible child processes, and avoid this region entirely when no backlog transition applies.

🔧 Fix: Handle backlog read failures and interruptible delivery
2 errors still open:

- 🚨 `bin/fm-teardown.sh:2855` - The required invariant says completion pairs record removal with backlog closure, but this unchecked `rm -f` can fail and teardown still runs `tasks-axi done`, clears the recovery marker, and reports success. A selective unlink failure leaves `state/&lt;id&gt;.meta` permanently paired with a Done row and no marker. Verify the meta is absent before mutating the backlog; on failure, preserve the marker and fail loudly.
- 🚨 `bin/fm-spawn.sh:2825` - The required dispatch sequence begins “after it durably writes state/&lt;id&gt;.meta,” but publication failure is unchecked. If this redirection fails, spawn continues through worker delivery, moves the queued row In flight, clears `SPAWN_FRESH_COMMIT_PENDING`, and reports success without a meta record, producing `orphan_in_flight`. Check publication success and record existence before delivery or backlog commit, and unwind through the existing failure path.

🔧 Fix: Centralize verified backlog record transitions

- 🚨 `bin/fm-spawn.sh:798` - During a fresh Orca spawn, interruption after SPAWN_FRESH_COMMIT_PENDING is set can make abort cleanup fail to remove the Orca worktree and publish this recovery meta, but line 808 immediately runs rollback and removes that newly published record. The leaked Orca resource is then left without its durable cleanup identity. Perform provisional rollback before publishing the Orca recovery record, or otherwise ensure rollback cannot delete recovery metadata.
- 🚨 `bin/fm-teardown.sh:2873` - An early failed Orca spawn whose worktree cleanup fails leaves the recovery-only meta written by fm-spawn.sh, but teardown treats it as a completed worker and closes its queued backlog item. Bootstrap can likewise move that item In flight at bin/fm-bootstrap.sh:1232 despite no worker having launched. Explicitly distinguish Orca cleanup-recovery records and exclude them from normal backlog start/done transitions.
- 🚨 `bin/fm-backlog-transition-lib.sh:105` - The intent requires tasks-axi to be “always addressed with an explicit --file &lt;data&gt;/backlog.md and run from that data directory&#39;s PARENT,” but row probes invoke `tasks-axi show` directly from the caller&#39;s working directory. Consequently the target home&#39;s `.tasks.toml` is not selected for dispatch and reconciliation reads. Run the probe from `fm_backlog_root &#34;$data&#34;`, matching mutations.
- ⚠️ `bin/fm-bootstrap.sh:1232` - A transient or malformed `tasks-axi show` failure is converted to an empty row and produces no BACKLOG_RECONCILE diagnostic, so an owned queued record remains unreconciled without explanation. Branch on FM_BACKLOG_ROW_RESULT and report FM_BACKLOG_ROW_ERROR for genuine read failures while continuing the sweep.

🔧 Fix: Isolate Orca cleanup recovery from backlog transitions
1 error still open:

- 🚨 `bin/fm-backlog-transition-lib.sh:105` - The required criterion says tasks-axi is “always addressed with an explicit --file &lt;data&gt;/backlog.md and run from that data directory&#39;s PARENT,” but row probes run `tasks-axi show` in the caller&#39;s current directory. With relocated data such as `/srv/fm-records`, dispatch and reconciliation therefore resolve configuration from the invocation directory instead of `/srv`. Run the probe from `fm_backlog_root &#34;$data&#34;`, matching the mutation boundary.

🔧 Fix: Address backlog reads from configured root
1 error still open:

- 🚨 `bin/fm-backlog-transition-lib.sh:63` - The intent requires every tasks-axi call to run from the data directory&#39;s parent, but `fm_backlog_root` returns the data directory itself for a trailing-slash override (`/srv/fm-records/` -&gt; `/srv/fm-records`) and returns `.` for a root-level directory (`/data` -&gt; `.`). Reads and mutations then resolve the wrong `.tasks.toml`, done_keep, and archive path. Normalize trailing slashes and preserve `/` as the parent at this shared addressing boundary.

🔧 Fix: Normalize backlog addressing root paths

- 🚨 `bin/fm-teardown.sh:2854` - Teardown accepts existing task records without `spawn_gen` and writes a new marker with an empty incarnation. If killed after this write but before meta removal, replay treats the marker as stale at bin/fm-backlog-transition-lib.sh:318, deletes it, and leaves the meta plus In-flight row indefinitely. This contradicts the required durable terminal-close recovery. Before destructive teardown, either give a legacy record a durable incarnation or refuse automatic closure when no incarnation can be recorded.
- ⚠️ `bin/fm-backlog-transition-lib.sh:80` - The intent requires scout report links to be relative to the backlog addressing root, but for data directory `/data`, the root is `/` and the pattern `&#34;$root&#34;/*` becomes `//*`, so this returns `/data`; teardown records the absolute `/data/&lt;id&gt;/report.md` instead of `data/&lt;id&gt;/report.md`. Handle `/` explicitly when stripping the addressing-root prefix.

🔧 Fix: Refuse unversioned closes and normalize report paths
2 errors still open:

- 🚨 `bin/fm-backlog-transition-lib.sh:120` - The required addressing invariant says every tasks-axi call targets `&lt;data&gt;/backlog.md` from the data directory&#39;s parent, but a nested relative data path is prefixed twice after `cd`: from the caller directory, `FM_DATA_OVERRIDE=relocated/data` passes the applies gate, then runs from `relocated` with `--file relocated/data/backlog.md`, targeting `relocated/relocated/data/backlog.md`. Resolve the backlog path for the post-`cd` working directory at this shared read/mutation boundary.
- 🚨 `tests/fm-public-followup.test.sh:1148` - Three changed teardown fixtures create backlog-tracked metadata without `spawn_gen` (also lines 1977 and 2136). The new mandatory legacy-record gate therefore refuses before the public-followup behavior under test; two success cases fail outright and the refusal case cannot reach its asserted reason. Give these synthetic current-generation records explicit spawn generations.

🔧 Fix: Add incarnations to teardown test fixtures
1 error still open:

- 🚨 `bin/fm-backlog-transition-lib.sh:120` - The required addressing invariant is still violated for nested relative data paths. With `FM_DATA_OVERRIDE=relocated/data`, the applies gate finds `relocated/data/backlog.md`, but the probe and mutation helpers then `cd relocated` while still passing `--file relocated/data/backlog.md`, targeting `relocated/relocated/data/backlog.md`. Teardown can consequently remove meta yet leave an unreplayable close marker. Normalize the backlog path for the post-`cd` root at this shared read/mutation boundary.

🔧 Fix: Normalize relative backlog data paths
1 error still open:

- 🚨 `bin/fm-backlog-transition-lib.sh:115` - The new path resolver returns 2 when the data directory cannot be resolved, but every caller treats all nonzero results as an authorized skip. During teardown, if the configured data directory is missing or inaccessible, both checks fall through, destructive cleanup proceeds, and line 2870 classifies the error as a skip before removing meta and reporting success without closing the backlog; it also bypasses the required missing-spawn_gen refusal. Preserve the legitimate return-1 exemptions, but make spawn, teardown, and bootstrap propagate return 2 before mutation; evaluate secondmate/manual exemptions before path resolution so those required skips remain skips.

🔧 Fix: Propagate fatal backlog addressing errors
2 errors still open:

- 🚨 `bin/fm-backlog-transition-lib.sh:104` - The required scout-link contract says the report path is relative to the backlog addressing root, but with FM_DATA_OVERRIDE=/ this helper returns an empty string, so bin/fm-teardown.sh:1183 records /&lt;id&gt;/report.md as an absolute path. Handle the data-directory-equals-root case explicitly and also avoid producing //backlog.md from fm_backlog_file.
- 🚨 `bin/fm-bootstrap.sh:1309` - backlog_record_reconcile explicitly returns 2 for fatal addressing failure, but this call ignores that status because bootstrap uses set -u rather than set -e. If DATA becomes unavailable after the initial gate but before reconciliation, bootstrap prints an error and then continues its remaining sweeps and can exit successfully, contradicting the required fatal propagation. Check the return and stop bootstrap on status 2.

🔧 Fix: Normalize root addressing and propagate reconciliation failures
1 error still open:

- 🚨 `tests/fm-backlog-atomicity.test.sh:371` - The required behavioral matrix is incomplete. The added cases cover a bare relative path and repeated trailing slashes, but do not drive real dispatch/teardown for data directories `/`, `/data`, and `/srv/fm-records`, nor verify the secondmate bootstrap exemption with an unresolvable data directory. The prior accepted requirement explicitly requires all listed addressing inputs and the secondmate/manual/no-backlog exemption paths.

🔧 Fix: Complete safe addressing and bootstrap exemption coverage

- 🚨 `bin/fm-teardown.sh:1189` - The required completion behavior is “tasks-axi done with the right --pr/--report/--note,” but a PR-based task whose meta lacks `pr=` is closed without `--pr`. Teardown explicitly supports this case by resolving a merged PR from the branch in `pr_is_merged`, yet that resolution never updates `PR_URL`, so `BACKLOG_DONE_ARGS` remains empty and the durable completion link is lost. Preserve the resolved PR URL and include it in the close marker and `tasks-axi done`.
- 🚨 `bin/fm-backlog-transition-lib.sh:354` - Replay trusts `id=` from the marker without verifying it matches the marker filename whose meta lock bootstrap acquired. A copied or malformed `foo.backlog-close` containing `id=bar` runs under foo&#39;s lock but can remove `bar.meta` and close bar&#39;s backlog row; traversal-shaped ids can also escape the state directory. At replay&#39;s parsing boundary, reject symlinks, require exactly one path-safe id matching the marker basename, and validate the permitted close arguments before any mutation.
- ⚠️ `bin/fm-teardown.sh:1204` - With a relocated `FM_DATA_OVERRIDE`, teardown correctly closes `&lt;configured-data&gt;/backlog.md` but then confirms that the item was closed in `data/backlog.md`. This makes the required confirmation echo factually wrong and also misdirects manual-backend users at line 1206. Render the actual configured backlog path, preferably relative to the same addressing root used for mutation.

🔧 Fix: Secure close replay and preserve completion links
1 error still open:

- 🚨 `bin/fm-backlog-transition-lib.sh:365` - Intent requires same-home reconciliation that “never reaches into another home,” but replay accepts `data=` directly from the marker and later probes/mutates that backlog. A copied or malformed `&lt;id&gt;.backlog-close` with a valid matching id and `data=/another/home/data` can close another home&#39;s row while holding only this home&#39;s meta lock. Validate the recorded data against an authorized same-home backlog location at the replay boundary before removing meta or accessing the backlog.

🔧 Fix: Harden close-marker replay validation
1 error still open:

- 🚨 `bin/fm-backlog-transition-lib.sh:363` - The hardened parsing boundary still accepts malformed markers that it was required to reject. Bash&#39;s `while read` skips an unterminated final line, so a trailing duplicate or unrecognized field is ignored, and the traversal check at line 394 accepts paths such as `$DATA/sub/..` when they resolve back to the authorized directory. Process the final unterminated line and reject any lexical `..` component before resolution so every marker field is validated before replay mutates meta or backlog state.

🔧 Fix: Complete close-marker validation hardening
2 errors still open:

- 🚨 `bin/fm-bootstrap.sh:1240` - A failed close replay that preserves both marker and meta is immediately undermined by the second reconciliation loop. For example, if meta removal fails while the row is `queued no`, replay correctly returns without backlog mutation, but this loop then sees the retained meta and calls `tasks-axi start`, changing the row to In flight despite the authoritative pending close. Skip record reconciliation whenever the corresponding `.backlog-close` still exists, including malformed or temporarily unreplayable markers.
- 🚨 `bin/fm-backlog-transition-lib.sh:420` - The hardened marker boundary still accepts malformed PR values such as `https://` because the wildcard may be empty. That contradicts the selected requirement to accept only permitted close-argument values and can pass a bogus completion link to `tasks-axi done`. Require a nonempty valid authority/path and reject whitespace or other URL-invalid characters before replay mutation.

🔧 Fix: Preserve pending closes and validate completion values
1 error still open:

- 🚨 `bin/fm-backlog-transition-lib.sh:369` - The required exhaustive marker validation remains bypassable. Shell variables cannot preserve NUL bytes, so `read` can normalize a NUL-bearing field before the printable/value checks; for example, a NUL appended to the authorized `data=` path can be accepted and allow meta removal and backlog closure. The PR parser also accepts malformed URLs such as `https://github.com:abc/pull/1` and invalid percent escapes because it permits arbitrary colons and `%`. Reject NUL/control bytes from the raw marker before parsing, and validate URL authority/escapes before any mutation.

🔧 Fix: Reject malformed marker bytes and completion URLs
1 error still open:

- 🚨 `bin/fm-backlog-transition-lib.sh:457` - The host check accepts malformed labels such as `https://foo.-bar.com/pull/1` or `https://foo-.bar.com/pull/1`, so a crafted marker can still reach `tasks-axi done` despite the required well-formed-host boundary. Validate each dot-separated label and reject labels beginning or ending with `-` before any replay mutation.

🔧 Fix: Validate PR host labels before replay

✅ No issues found.

- 🚨 `bin/fm-backlog-transition-lib.sh:333` - A scout using a configured data directory containing whitespace generates a marker such as `arg=crew data/&lt;id&gt;/report.md`, but replay rejects all report values containing whitespace at line 502. If teardown is killed after removing meta but before `tasks-axi done`, the claimed durable recovery can never close the row. Ensure marker writing accepts only replayable values before destructive cleanup, or define a safe reversible encoding for generated report paths.
- ⚠️ `bin/fm-bootstrap.sh:1214` - The close-marker sweep checks only `-e`, so a dangling `.backlog-close` symlink never reaches replay&#39;s explicit symlink rejection or emits `BACKLOG_RECONCILE`. The later record sweep still sees `-L` and permanently suppresses reconciliation for that id. Include symlinks in marker discovery so malformed pending closes remain fail-closed but actionable.

🔧 Fix: Replay spaced reports and surface dangling markers
1 error still open:

- 🚨 `bin/fm-backlog-transition-lib.sh:327` - The writer rejects only newline in arguments and does not validate control bytes in `data`, while replay rejects every control byte except newline. For example, a scout whose configured data path contains a tab writes a pending-close marker successfully; if teardown then crashes after meta removal, bootstrap permanently rejects that marker as an invalid control byte. Validate every serialized field and argument against replay&#39;s control-byte rules before publishing the marker or beginning destructive cleanup.

🔧 Fix: Unify close-marker writer and replay validation
1 error still open:

- 🚨 `bin/fm-backlog-transition-lib.sh:60` - `fm_backlog_data_absolute` captures `pwd -P` with command substitution, which strips trailing newlines from a configured data-directory name. The gate then probes the wrong stripped path and treats the home as having no backlog: dispatch can publish meta without starting the real row, while teardown can remove meta without writing a marker or closing the real row. Reject control bytes in the input at this shared boundary before canonicalization so this case fails closed.

🔧 Fix: Reject control-byte backlog paths before canonicalization
1 warning still open:

- ⚠️ `bin/fm-backlog-transition-lib.sh:460` - The report-path traversal check rejects `../x` and paths ending in `/..`, but not the exact value `..`. A pending-close marker with `arg=--report` and `arg=..` therefore passes validation and executes `tasks-axi done --report ..`, recording an escaping completion link without error. Reject exact `..` alongside the existing traversal forms.

🔧 Fix: Reject bare report traversal paths

✅ No issues found.

✅ No issues found.

- 🚨 `bin/fm-backlog-transition-lib.sh:383` - Pending-close validation is locale-dependent. A teardown running under a UTF-8 locale accepts and publishes a marker whose valid data path contains non-ASCII characters (for example `/tmp/café`), but bootstrap under `LC_ALL=C` rejects the same marker via `[![:print:]]`, permanently blocking durable close recovery. Make the shared validator enforce locale-independent byte rules so any marker accepted by the writer remains replayable across process locales.

🔧 Fix: Stabilize marker replay and failed-worker rollback
1 error still open:

- 🚨 `bin/fm-spawn.sh:714` - Failed-dispatch cleanup treats any false result from `fm_backend_target_exists` as proof the launched endpoint was stopped, but that helper explicitly returns false for both absence and an unqueryable backend. For example, if cmux becomes unreachable, `fm_backend_cmux_kill` skips the close because readiness cannot be verified, the subsequent existence check also fails, and rollback deletes the task and busy records while the workspace and worker may remain live. Retain ownership unless the backend positively confirms the endpoint is gone, using a recovery-grade confirmed-absence boundary rather than this liveness boolean.

🔧 Fix: Require confirmed endpoint absence before rollback
1 error still open:

- 🚨 `bin/fm-spawn.sh:3078` - The live-worker mark is set only after the fallible Enter submission. If a deferred signal interrupts this command after the backend accepted Enter but before it returns successfully, `set -e` reaches EXIT cleanup with `SPAWN_WORKER_LIVE=0`; rollback then deletes meta and busy records without stopping or confirming the launched endpoint. Conservatively mark the endpoint live before issuing Enter so every ambiguous submission failure uses the confirmed-stop rollback path.

🔧 Fix: Mark workers live before launch submission
</details>

<details>

✅ No issues found.
- <code>Inspected `git diff 420721401c4080d1a4f6982b0ef6769e2a749b23..2f0d3fae51924971b63cde933cd1b0c83763aa1c` and repository status.</code>
- `bin/fm-test-run.sh tests/fm-backlog-atomicity.test.sh`
- <code>`bin/fm-test-run.sh tests/fm-control-relaunch.test.sh tests/fm-teardown.test.sh` initially exposed an unpinned fixture data path in the teardown test.</code>
- <code>Added `FM_DATA_OVERRIDE=&#34;$case_dir/data&#34;` to the affected Herdr teardown fixture.</code>
- `bin/fm-test-run.sh tests/fm-teardown.test.sh`
- `bin/fm-test-run.sh tests/fm-control-relaunch.test.sh`

✅ No issues found.
- `git status --short && git diff --stat 4eb587d670ff1a34afaba32fa94d02e74fed0399..HEAD && git diff --name-only 4eb587d670ff1a34afaba32fa94d02e74fed0399..HEAD`
- `./bin/fm-test-run.sh tests/fm-backlog-atomicity.test.sh`
- `./bin/fm-test-run.sh tests/fm-control-relaunch.test.sh`
- <code>Verified the worktree remained clean after testing with `git status --short`.</code>

✅ No issues found.
- <code>`git status --short` and `git diff --stat/--name-only 4eb587d670ff1a34afaba32fa94d02e74fed0399..9c82c2e33efd34fb9e479db11e54ff00aa21304c`</code>
- `bin/fm-test-run.sh tests/fm-backlog-atomicity.test.sh tests/fm-control-relaunch.test.sh`
- <code>Manual CLI fixture using real `tasks-axi add/show`, `bin/fm-spawn.sh`, and `bin/fm-teardown.sh`; verified queued/absent → in_flight/present → done/absent with no residual close marker</code>
- <code>`git diff --check` and final `git status --short` transient-artifact check</code>

✅ No issues found.
- `./bin/fm-test-run.sh tests/fm-backlog-atomicity.test.sh`
- `./bin/fm-test-run.sh tests/fm-control-relaunch.test.sh`
- <code>`git status --short` confirmed testing left no worktree changes</code>

✅ No issues found.
- <code>Inspected `git status --short` and the commit diff from `4eb587d670ff1a34afaba32fa94d02e74fed0399` to `0331595e38b10ae280454a6e3761de04609daf3e`.</code>
- `bin/fm-test-run.sh tests/fm-backlog-atomicity.test.sh`
- `bin/fm-test-run.sh tests/fm-control-relaunch.test.sh tests/fm-teardown.test.sh`
- <code>Verified `git status --short --untracked-files=all` remained clean after testing.</code>

✅ No issues found.
- `git status --short && git diff --stat 0ace60ab1a6eb8afbb3a3e7b104731e4e5b7baaa..HEAD && git diff --name-only 0ace60ab1a6eb8afbb3a3e7b104731e4e5b7baaa..HEAD`
- `bin/fm-test-run.sh tests/fm-backlog-atomicity.test.sh`
- `bin/fm-test-run.sh tests/fm-control-relaunch.test.sh`
- <code>`git status --short` after testing to confirm no transient worktree changes remained</code>

✅ No issues found.
- `git status --short && git diff --stat 0ace60ab1a6eb8afbb3a3e7b104731e4e5b7baaa..HEAD && git diff --name-only 0ace60ab1a6eb8afbb3a3e7b104731e4e5b7baaa..HEAD`
- `bin/fm-test-run.sh tests/fm-backlog-atomicity.test.sh --json ~/.no-mistakes/evidence/01M15B2X4TSJ4HEJ732WBPZRNP/fm-backlog-atomicity-results.json`
- `bin/fm-test-run.sh tests/fm-control-relaunch.test.sh --json ~/.no-mistakes/evidence/01M15B2X4TSJ4HEJ732WBPZRNP/fm-control-relaunch-results.json`
- `bin/fm-test-run.sh tests/fm-teardown.test.sh --json ~/.no-mistakes/evidence/01M15B2X4TSJ4HEJ732WBPZRNP/fm-teardown-results.json`
- <code>`git status --short` after testing confirmed the worktree remained clean.</code>
</details>

<details>
<summary>⚠️ **Document** - 0 issues</summary>

✅ No issues found.

✅ No issues found.

✅ No issues found.

✅ No issues found.

- ⚠️ `bin/fm-spawn.sh:3088` - The failure diagnostic says no worker remains after rollback, then instructs the operator to close the already-launched endpoint and local copy. Correcting this user-facing executable output is outside the documentation-only phase.

🔧 Fix: Verify backlog lifecycle documentation
1 warning still open:

- ⚠️ `bin/fm-spawn.sh:3088` - The failure diagnostic still claims no worker remains while instructing the operator to close the already-launched endpoint and local copy. Fixing executable output is outside this documentation-only phase.

✅ No issues found.

✅ No issues found.
</details>

<details>
<summary>🔧 **Lint** - 1 issue found → auto-fixed (2) ✅</summary>

🔧 Fix: Quote done literals in backlog atomicity assertions

✅ No issues found.

✅ No issues found.

🔧 Fix: Quote done literal in backlog atomicity test

✅ No issues found.

✅ No issues found.

✅ No issues found.
</details>

<details>

✅ No issues found.

✅ No issues found.

✅ No issues found.

✅ No issues found.

✅ No issues found.

✅ No issues found.

✅ No issues found.
</details>

### DIFF
--- a/.agents/skills/bootstrap-diagnostics/SKILL.md
+++ b/.agents/skills/bootstrap-diagnostics/SKILL.md
@@ -55,10 +55,11 @@ When any diagnostic needs captain attention, report the plain consequence and re
   Read the named record for the recorded reasons, then reproduce with a direct `bin/fm-home-summary-refresh.sh` (no `--best-effort`, which is what keeps the failure quiet) so the refresh error reaches you.
   A recorded deadline means the complete refresh did not finish inside `FM_HOME_SUMMARY_TIMEOUT`, so inspect lock acquisition and producer completion before validation or publication, and fix the blocked phase rather than raising this load-bearing bound.
 
-- `BACKLOG_RECONCILE: <id>: recorded backlog close could not be replayed: <reason>` - an interrupted cleanup recorded exactly which backlog item it still owed a close, and this session start could not land it.
-  The endpoint and local copy are already gone; the task record is normally gone too, but remains paired with the marker when recovery could not remove it safely.
-  Read the named reason, fix the record or backlog-file problem, and rerun session start so the same recorded close replays.
-  Never hand-close the item by deleting `state/<id>.backlog-close` - that discards the completion link the cleanup captured.
+- `BACKLOG_RECONCILE: <id>: recorded backlog close could not be replayed: <reason>` - this session start found a pending-close record but could not land it.
+  For a valid teardown record, the endpoint and local copy are already gone; the task record is normally gone too, but remains paired with the marker when recovery could not remove it safely.
+  A validation error means the record cannot be trusted, so do not assume cleanup completed or follow any path or argument stored in it.
+  Read the named reason, inspect the marker as inert data when validation failed, fix the record or backlog-file problem, and rerun session start so a valid recorded close replays.
+  Never hand-close the item by deleting `state/<id>.backlog-close` - that can discard a completion link the cleanup captured, and the surviving marker prevents the record sweep from starting the item meanwhile.
 - `BACKLOG_RECONCILE: <id>: worker record exists but its backlog item could not be read: <reason>` - this home could not determine whether the item matches its worker record.
   Resolve the named backlog read problem and rerun session start; never guess by starting or closing an unreadable item.
 - `BACKLOG_RECONCILE: <id>: worker record exists but its backlog item could not be moved to In flight: <reason>` - this home owns a worker whose backlog item is still queued, and the reconciliation could not correct it.
--- a/.agents/skills/bootstrap-diagnostics/SKILL.md
+++ b/.agents/skills/bootstrap-diagnostics/SKILL.md
@@ -2,7 +2,7 @@
 name: bootstrap-diagnostics
 description: >-
   Agent-only handling playbook for session-start bootstrap diagnostics.
-  Use whenever the session-start digest's bootstrap or network-checks section prints an actionable diagnostic line - MISSING, MISSING_MANUAL, BACKEND_INVALID, NEEDS_GH_AUTH, TANGLE, STARTUP_MEMORY_BUDGET, CREW_DISPATCH invalid, FLEET_SYNC, NETWORK_CHECKS, PR_CHECK_MIGRATION, HOME_SUMMARY, SECONDMATE_SYNC, SECONDMATE_LIVENESS, SECONDMATE_HANDOFF, NUDGE_SECONDMATES, or FMX - or when a standalone bin/fm-bootstrap.sh or bin/fm-startup-network.sh run prints one of those lines.
+  Use whenever the session-start digest's bootstrap or network-checks section prints an actionable diagnostic line - MISSING, MISSING_MANUAL, BACKEND_INVALID, NEEDS_GH_AUTH, TANGLE, STARTUP_MEMORY_BUDGET, CREW_DISPATCH invalid, FLEET_SYNC, NETWORK_CHECKS, PR_CHECK_MIGRATION, HOME_SUMMARY, BACKLOG_RECONCILE, SECONDMATE_SYNC, SECONDMATE_LIVENESS, SECONDMATE_HANDOFF, NUDGE_SECONDMATES, or FMX - or when a standalone bin/fm-bootstrap.sh or bin/fm-startup-network.sh run prints one of those lines.
   A silent bootstrap section, or a BOOTSTRAP_INFO fact, means no skill load.
 user-invocable: false
 metadata:
@@ -55,6 +55,11 @@ When any diagnostic needs captain attention, report the plain consequence and re
   Read the named record for the recorded reasons, then reproduce with a direct `bin/fm-home-summary-refresh.sh` (no `--best-effort`, which is what keeps the failure quiet) so the refresh error reaches you.
   A recorded deadline means the complete refresh did not finish inside `FM_HOME_SUMMARY_TIMEOUT`, so inspect lock acquisition and producer completion before validation or publication, and fix the blocked phase rather than raising this load-bearing bound.
 
+- `BACKLOG_RECONCILE: <id>: recorded backlog close could not be replayed: <reason>` - an interrupted cleanup recorded exactly which backlog item it still owed a close, and this session start could not land it.
+  The task's own record is already gone, so the item is showing as running when the work is finished; read the named reason, fix what is blocking the backlog file, and rerun session start so the same recorded close replays.
+  Never hand-close the item by deleting `state/<id>.backlog-close` - that discards the completion link the cleanup captured.
+- `BACKLOG_RECONCILE: <id>: worker record exists but its backlog item could not be moved to In flight: <reason>` - this home owns a worker whose backlog item is still queued, and the reconciliation could not correct it.
+  Until it is corrected, the fleet view reads that worker as work no backlog item owns; resolve the named backlog problem and rerun session start.
 - `SECONDMATE_SYNC: secondmate <id>: skipped: <reason>` - secondmate convergence left a live home on its existing checkout because the home was dirty, diverged, unsafe, on the wrong branch, missing its placement-specific target commit, unreachable, or otherwise not fast-forwardable, or because inherited local-material propagation failed; bootstrap continued, but inspect the reason because the secondmate's tracked instructions, inherited settings, or shared captain preferences may be stale after a primary update.
 - `SECONDMATE_LIVENESS: secondmate <id>: skipped: <reason>|respawn failed after <cause>: <reason>` - the session-start liveness sweep could not guarantee that the registered secondmate is running a real agent process.
   Investigate the reason because that secondmate is not guaranteed live.
--- a/.agents/skills/bootstrap-diagnostics/SKILL.md
+++ b/.agents/skills/bootstrap-diagnostics/SKILL.md
@@ -59,6 +59,8 @@ When any diagnostic needs captain attention, report the plain consequence and re
   The endpoint and local copy are already gone; the task record is normally gone too, but remains paired with the marker when recovery could not remove it safely.
   Read the named reason, fix the record or backlog-file problem, and rerun session start so the same recorded close replays.
   Never hand-close the item by deleting `state/<id>.backlog-close` - that discards the completion link the cleanup captured.
+- `BACKLOG_RECONCILE: <id>: worker record exists but its backlog item could not be read: <reason>` - this home could not determine whether the item matches its worker record.
+  Resolve the named backlog read problem and rerun session start; never guess by starting or closing an unreadable item.
 - `BACKLOG_RECONCILE: <id>: worker record exists but its backlog item could not be moved to In flight: <reason>` - this home owns a worker whose backlog item is still queued, and the reconciliation could not correct it.
   Until it is corrected, the fleet view reads that worker as work no backlog item owns; resolve the named backlog problem and rerun session start.
 - `SECONDMATE_SYNC: secondmate <id>: skipped: <reason>` - secondmate convergence left a live home on its existing checkout because the home was dirty, diverged, unsafe, on the wrong branch, missing its placement-specific target commit, unreachable, or otherwise not fast-forwardable, or because inherited local-material propagation failed; bootstrap continued, but inspect the reason because the secondmate's tracked instructions, inherited settings, or shared captain preferences may be stale after a primary update.
--- a/.agents/skills/bootstrap-diagnostics/SKILL.md
+++ b/.agents/skills/bootstrap-diagnostics/SKILL.md
@@ -56,7 +56,8 @@ When any diagnostic needs captain attention, report the plain consequence and re
   A recorded deadline means the complete refresh did not finish inside `FM_HOME_SUMMARY_TIMEOUT`, so inspect lock acquisition and producer completion before validation or publication, and fix the blocked phase rather than raising this load-bearing bound.
 
 - `BACKLOG_RECONCILE: <id>: recorded backlog close could not be replayed: <reason>` - an interrupted cleanup recorded exactly which backlog item it still owed a close, and this session start could not land it.
-  The task's own record is already gone, so the item is showing as running when the work is finished; read the named reason, fix what is blocking the backlog file, and rerun session start so the same recorded close replays.
+  The endpoint and local copy are already gone; the task record is normally gone too, but remains paired with the marker when recovery could not remove it safely.
+  Read the named reason, fix the record or backlog-file problem, and rerun session start so the same recorded close replays.
   Never hand-close the item by deleting `state/<id>.backlog-close` - that discards the completion link the cleanup captured.
 - `BACKLOG_RECONCILE: <id>: worker record exists but its backlog item could not be moved to In flight: <reason>` - this home owns a worker whose backlog item is still queued, and the reconciliation could not correct it.
   Until it is corrected, the fleet view reads that worker as work no backlog item owns; resolve the named backlog problem and rerun session start.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -307,7 +307,7 @@ Write the task-specific brief under section 11 before spawning.
 
 Spawn only through `bin/fm-spawn.sh` after the profile and backend checks in section 4.
 The spawn must resolve a genuine isolated task worktree distinct from the primary checkout; a failed isolation assertion stops the task.
-The spawn itself moves the work item to In flight and refuses rather than dispatching work this home has no item for, so recording the dispatch is never a separate step to remember.
+When the configured tasks-axi backlog gate applies, the spawn itself moves the work item to In flight and refuses rather than dispatching work this home has no item for, so recording the dispatch is never a separate step to remember; a manual-backend home retains the hand-editing contract in `docs/configuration.md`.
 After spawning, confirm the worker is processing the brief and handle any trust dialog through `harness-adapters`.
 A persistent secondmate is recorded in the secondmate registry and runtime state, never as a backlog work item.
 
@@ -499,7 +499,7 @@ Work routed to a secondmate is recorded in that secondmate home's own backlog, n
 A decision is simply a task held for the captain: `tasks-axi hold <id> --reason "<reason>" --kind captain`, with `--until <date>` when the captain defers it.
 When a main-side thread such as a pending captain decision or relay reminder is worth durable tracking, file it as its own work item and hold it the same way.
 Captain calls discovered by investigations or visual reviews follow `captain-hold-lifecycle`, which owns their completion gate and recorded-answer rules.
-Dispatch and completion move the item themselves - `bin/fm-spawn.sh` and `bin/fm-teardown.sh` own those transitions and refuse rather than report success without them - so what remains yours is filing the item before dispatch, recording decisions, and keeping notes current.
+Under the automatic tasks-axi backend, dispatch and completion move the item themselves - `bin/fm-spawn.sh` and `bin/fm-teardown.sh` own those transitions and refuse rather than report success without them - so what remains yours is filing the item before dispatch, recording decisions, and keeping notes current; `docs/configuration.md` owns the manual-backend exception.
 Re-evaluate queued work after every teardown and heartbeat, dispatching items only when dependencies and time gates have cleared.
 
 `.tasks.toml`, `docs/configuration.md`, and current `tasks-axi --help` own the backlog schema, compatibility, retention, and routine command syntax.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -97,6 +97,7 @@ state/               runtime records and signals; gitignored
   <id>.muse-session  muse busy-source binding (sessions root plus task worktree) written by fm-spawn; removed by teardown
   <id>.cursor-session  cursor busy-source binding (projects root, task worktree, prior conversations) written by fm-spawn; removed by teardown
   <id>.reconcile-nudged  epoch second of the last inventory-reconcile nudge sent to this secondmate; bin/fm-secondmate-reconcile.sh owns its per-home cooldown window
+  <id>.backlog-close  the exact backlog close a teardown recorded before removing the task's record, so an interrupted cleanup can still be finished at the next session start; bin/fm-backlog-transition-lib.sh owns its format and replay, and a landed close removes it
   <id>.inbox/          durable steering inbox: sequenced firstmate instruction records the worker acknowledges by moving them into its handled/ subdirectory; written by fm-send, with ordinary records re-rung and escalated by the watcher while explicit fire-and-forget records are excluded from that ladder, and removed by teardown (bin/fm-task-inbox-lib.sh)
   <id>.meta          task metadata; each producer script's header owns its exact fields and mutation contract, with docs/configuration.md routing operator-facing backend and trace-context details
   <id>.herdr-presentation  quarantinable attempt and restart-binding journal for Herdr's optional visual projection; never task or endpoint authority; see docs/herdr-backend.md "Presentation spaces"
@@ -168,7 +169,7 @@ When that section reports its checks still in progress it names exactly what is 
 1. **Lock** - acquires the per-home session lock first, before anything mutates shared state, then starts the deferred network stage above.
 2. **Bootstrap** - detect-only checks (tool/version problems, the worktree-tangle check, harness override, dispatch-profile validation, backlog-backend status) always run, but routine confirmations stay silent by default.
    When the lock could not be acquired, the worktree-tangle check uses read-only advisory wording without a checkout repair command.
-   Home-local stale Herdr projection cleanup and the six bootstrap MUTATING sweeps - non-executing legacy PR-check migration, fleet sync, secondmate convergence, secondmate liveness, pending remote handoff retry, and Relay artifact writes - run only when this session actually holds the lock from step 1; the four network ones among them run in the deferred stage rather than in this section.
+   Home-local stale Herdr projection cleanup and the seven bootstrap MUTATING sweeps - non-executing legacy PR-check migration, same-home backlog reconciliation, fleet sync, secondmate convergence, secondmate liveness, pending remote handoff retry, and Relay artifact writes - run only when this session actually holds the lock from step 1; the four network ones among them run in the deferred stage rather than in this section.
    The secondmate liveness sweep deterministically accounts for every registered secondmate: it relaunches only from the recovery-grade `dead` or `missing` states, preserves ambiguous, unreadable, or unreachable remote targets, and reports skipped or failed guarantees as `SECONDMATE_LIVENESS:` lines (`bin/fm-bootstrap.sh`; `bin/fm-backend.sh`'s `fm_backend_agent_state`; `docs/remote-secondmates.md`).
 3. **Wake queue** - when locked, presents the durable wake queue and prints the raw records prominently as this turn's first work queue; a clearly labeled status-event annotation may follow a valid `signal` record and includes every status line still unread at the presentation cursor, but never replaces the raw record or current-state reconciliation, and a lapsed watcher chain still surfaces here via the same guard alarm.
    Presented records remain durable until the handling turn runs the generation-bound acknowledgement printed by the drain.
@@ -306,7 +307,8 @@ Write the task-specific brief under section 11 before spawning.
 
 Spawn only through `bin/fm-spawn.sh` after the profile and backend checks in section 4.
 The spawn must resolve a genuine isolated task worktree distinct from the primary checkout; a failed isolation assertion stops the task.
-After spawning, confirm the worker is processing the brief, handle any trust dialog through `harness-adapters`, and record ship or scout work as under way.
+The spawn itself moves the work item to In flight and refuses rather than dispatching work this home has no item for, so recording the dispatch is never a separate step to remember.
+After spawning, confirm the worker is processing the brief and handle any trust dialog through `harness-adapters`.
 A persistent secondmate is recorded in the secondmate registry and runtime state, never as a backlog work item.
 
 Steer a worker with ordinary text through fail-closed `fm-send`: the message becomes a durable record in the task's steering inbox (multi-line text is legal, local and remote alike) and the worker's terminal receives only a constant doorbell line, with the watcher re-ringing an unacknowledged local message and escalating a stuck one (`bin/fm-task-inbox-lib.sh`; `bin/fm-send.sh` owns the typed-plane carve-outs).
@@ -497,7 +499,7 @@ Work routed to a secondmate is recorded in that secondmate home's own backlog, n
 A decision is simply a task held for the captain: `tasks-axi hold <id> --reason "<reason>" --kind captain`, with `--until <date>` when the captain defers it.
 When a main-side thread such as a pending captain decision or relay reminder is worth durable tracking, file it as its own work item and hold it the same way.
 Captain calls discovered by investigations or visual reviews follow `captain-hold-lifecycle`, which owns their completion gate and recorded-answer rules.
-Update the backlog on every dispatch, completion, and decision for a work item.
+Dispatch and completion move the item themselves - `bin/fm-spawn.sh` and `bin/fm-teardown.sh` own those transitions and refuse rather than report success without them - so what remains yours is filing the item before dispatch, recording decisions, and keeping notes current.
 Re-evaluate queued work after every teardown and heartbeat, dispatching items only when dependencies and time gates have cleared.
 
 `.tasks.toml`, `docs/configuration.md`, and current `tasks-axi --help` own the backlog schema, compatibility, retention, and routine command syntax.
@@ -535,7 +537,7 @@ It performs guarded fast-forward updates of firstmate and registered secondmate 
 
 These skills are not captain-invocable; load them only at their precise triggers.
 
-- `bootstrap-diagnostics` - load whenever the session-start digest's bootstrap or network-checks section prints an actionable diagnostic line (`MISSING:`, `MISSING_MANUAL:`, `BACKEND_INVALID:`, `NEEDS_GH_AUTH`, `TANGLE:`, `STARTUP_MEMORY_BUDGET:`, `CREW_DISPATCH: invalid`, `FLEET_SYNC:`, `NETWORK_CHECKS:`, `PR_CHECK_MIGRATION:`, `HOME_SUMMARY:`, `SECONDMATE_SYNC:`, `SECONDMATE_LIVENESS:`, `SECONDMATE_HANDOFF:`, `NUDGE_SECONDMATES:`, or `FMX:`); silence and `BOOTSTRAP_INFO:` need no load.
+- `bootstrap-diagnostics` - load whenever the session-start digest's bootstrap or network-checks section prints an actionable diagnostic line (`MISSING:`, `MISSING_MANUAL:`, `BACKEND_INVALID:`, `NEEDS_GH_AUTH`, `TANGLE:`, `STARTUP_MEMORY_BUDGET:`, `CREW_DISPATCH: invalid`, `FLEET_SYNC:`, `NETWORK_CHECKS:`, `PR_CHECK_MIGRATION:`, `HOME_SUMMARY:`, `BACKLOG_RECONCILE:`, `SECONDMATE_SYNC:`, `SECONDMATE_LIVENESS:`, `SECONDMATE_HANDOFF:`, `NUDGE_SECONDMATES:`, or `FMX:`); silence and `BOOTSTRAP_INFO:` need no load.
 - `diagnostic-reasoning` - load before scoping a reported bug and before acting on a diagnostic report.
 - `ask-user-authority` - load before deciding any ask-user finding.
 - `quota-array-dispatch` - load before choosing among a matched crew-dispatch profile array from current quota-axi default TOON.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -499,7 +499,7 @@ Work routed to a secondmate is recorded in that secondmate home's own backlog, n
 A decision is simply a task held for the captain: `tasks-axi hold <id> --reason "<reason>" --kind captain`, with `--until <date>` when the captain defers it.
 When a main-side thread such as a pending captain decision or relay reminder is worth durable tracking, file it as its own work item and hold it the same way.
 Captain calls discovered by investigations or visual reviews follow `captain-hold-lifecycle`, which owns their completion gate and recorded-answer rules.
-Under the automatic tasks-axi backend, dispatch and completion move the item themselves - `bin/fm-spawn.sh` and `bin/fm-teardown.sh` own those transitions and refuse rather than report success without them - so what remains yours is filing the item before dispatch, recording decisions, and keeping notes current; `docs/configuration.md` owns the manual-backend exception.
+When the automatic transition gate applies, dispatch and completion move the item themselves - `bin/fm-spawn.sh` and `bin/fm-teardown.sh` own those transitions and refuse rather than report success without them - so what remains yours is filing the item before dispatch, recording decisions, and keeping notes current; `docs/configuration.md` owns gate applicability and the manual-backend exception.
 Re-evaluate queued work after every teardown and heartbeat, dispatching items only when dependencies and time gates have cleared.
 
 `.tasks.toml`, `docs/configuration.md`, and current `tasks-axi --help` own the backlog schema, compatibility, retention, and routine command syntax.
--- a/bin/fm-backlog-transition-lib.sh
+++ b/bin/fm-backlog-transition-lib.sh
@@ -44,6 +44,9 @@
 FM_BACKLOG_TRANSITION_SKIP=
 # Set by the mutating helpers when they return non-zero.
 FM_BACKLOG_TRANSITION_ERROR=
+FM_BACKLOG_ROW_RESULT=
+FM_BACKLOG_ROW_STATE=
+FM_BACKLOG_ROW_ERROR=
 # Set by fm_backlog_close_marker_replay: closed | stale | noop.
 # shellcheck disable=SC2034 # Output global, read by the sourcing caller.
 FM_BACKLOG_CLOSE_REPLAY_RESULT=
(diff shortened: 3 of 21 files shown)


kunchenguid:
Superseded by #3322 (https://github.com/kunchenguid/firstmate/pull/3322), which was rebuilt cleanly from origin/main and merged. Closing this stale PR in favor of that one.

PR #3172, opened 2026-08-27
fix(bin): bound self-scheduling loops and crewmate task scope

nbost130:
## Problem

`bin/fm-subagent-pretool-check.sh` classifies a harness tool name by shape and denies delegation-shaped names, but it goes inert in a crewmate/scout task worktree. The stated rationale is that "a crewmate using delegation tools inside its own task worktree is legitimate."

That rationale holds for **ordinary delegation**, which creates a *separate* unit of work. It does not hold for **self-scheduling**, which creates no separate worker at all: it re-invokes the caller's **own** session on a timer.

The two carry different hazards:

- Ordinary delegation's hazard is work firstmate does not know about. That only matters where a fleet is operated, so the primary-only scope is right.
- Self-scheduling's hazard is an unbounded loop over one growing session, and it is worst **exactly where the scope test was inert**.

Each timer wake replays the caller's entire accumulated context, so a one-line "nothing changed yet" turn is billed at the whole session's size, and every additional cycle costs more than the last because the session keeps growing. Nothing bounds it: no durable fleet record, no turn ceiling, no wall-clock ceiling. Supervision can neither see the loop nor stop it.

It is also redundant. Firstmate already owns a strictly cheaper wait: `bin/fm-watch.sh` runs `state/<id>.check.sh` polls as plain shell in the watcher process and wakes a session **only when a check actually prints something**, so an empty poll costs no model context at all. The generated brief already tells workers to use a `paused:` status line for a bounded external wait — the guard simply never stopped a worker from reaching for a timer instead.

The asymmetry before this change:

| Tool | Primary home | Task worktree |
|---|---|---|
| `ScheduleWakeup` | deny | **allow** |
| `CronCreate` | deny | **allow** |
| `Agent` | deny | allow |

The same name firstmate refuses in a primary was freely available in the worktree where long-running workers actually run.

## Change

Split the classification so each shape gets the scope its hazard needs.

- New `SELF_SCHEDULING_STEMS` (`schedul`, `cron`), moved out of `DELEGATION_STEMS` so each stem has exactly one owner.
- Self-scheduling is denied in a task worktree **as well as** a primary or secondmate home. Ordinary delegation keeps today's behavior exactly and stays allowed in a worktree — a crewmate spawning a sub-agent for a slice of its own task is still legitimate.
- Neither shape fires outside firstmate territory, so an unrelated repository on the machine is untouched and a scheduling-shaped name there is never classified.
- The self-scheduling deny names the supported wait path (a `paused:`/`blocked:` status line, with firstmate arming a custom condition via `bin/fm-check-register.sh`) rather than the brief-then-spawn route, which is wrong advice for a worker that only needs to wait.
- `crondelete` joins `OBSERVE_ONLY_TOOLS`. Cancelling a schedule stops work and creates none, and without it a worker that already owns a schedule would have no way to cancel one once denial reaches worktrees. This relaxes the primary too, which that list's own stated rationale ("never the reason a runaway task cannot be stopped") intends.

Classification stays **shape-based**, not a fixed name list, so a future scheduling tool nobody has seen is still caught.

Resulting scope:

| Shape | Primary / secondmate | Task worktree | Non-firstmate repo |
|---|---|---|---|
| Ordinary delegation | denied | allowed | inert |
| Self-scheduling | denied | **denied** | inert |

### Second, smaller fix

The generated crewmate brief bounded delivery and escalation but said nothing about work **discovered mid-task**, so a mission that keeps finding new items of the same kind had no terminal condition — it could absorb newly appearing work into the current task indefinitely.

`AGENTS.md` already carries that rule ("prefer routing new requirements to follow-up work rather than expanding the current task"), but a crewmate runs in a worktree of the *target project* and never loads firstmate's `AGENTS.md`, so the contract never reached the worker who needed it. The secondmate charter has the idea in stronger form; crewmates had no equivalent.

Added one rule to the ship and scout scaffolds: the briefed task is the whole scope, discovered adjacent work is reported rather than absorbed, and firstmate decides whether it becomes a separate task. It carries an explicit carve-out so corrections genuinely required to make the briefed work correct stay in scope — without that, the rule would tell workers to escalate ordinary bug-fixing.

## Preserved invariants

Every existing safety property is intact and still covered by the original tests: `mcp__*` is never classified; both exclusion lists stay exact-name; `FM_ALLOW_SUBAGENT=1` remains the single escape hatch and releases the new denial too; malformed/empty stdin, a missing tool name, and a missing `jq` all still fail open; a broken environment fails open rather than denying; deny output shape is unchanged (exit 2, empty stdout under `--claude`, Claude object on stderr, Grok object on stdout otherwise); the message still starts with `[subagent-dispatch]` and contains `blocked tool: <name>`; a marked secondmate home stays guarded. The primary-home guard is not weakened anywhere.

## Tests

Extended the two already-registered suites rather than adding files, so the coverage manifest and per-file timing budgets are unaffected.

New coverage: self-scheduling denied in a task worktree (including invented future names `ScheduleJob`, `CronSchedule` to pin the shape match); ordinary delegation still allowed there (`Agent`, `Task`, `SpawnWorker`, `EnterWorktree`) to pin the fix's narrowness; ordinary tools still allowed; the escape hatch releasing the new denial; the deny message being the wait-path text and *not* the spawn text; `CronDelete`/`CronList` allowed in both scopes; inertness in a non-firstmate repo; and the brief scope rule in both scaffolds.

Both new tests were confirmed red on the pre-change code and green after:

```
not ok - self-scheduling tool ScheduleWakeup must be denied in task worktree, got exit 0
not ok - ship brief missing the bounded scope rule
```

Results:

```
tests/fm-subagent-pretool-check.test.sh    PASS (15 checks)
tests/fm-brief.test.sh                     PASS (23 checks)
```

Suites that consume generated briefs, checked for ripple, all passing:

```
fm-ask-user-authority   fm-secondmate-safety (77)   fm-tangle-guard (6)
fm-classify-corr-token (12)                         fm-voice-relay (40)
```

Gates:

```
bin/fm-lint.sh                  exit 0  (ShellCheck 0.11.0 extended, actionlint 1.7.12)
bin/fm-test-run.sh --check-coverage
  FM_TEST_COVERAGE ok total=166 parallel=24 serial=130 serial_shards=4 herdr=12
```

Also verified against a real on-disk task worktree in the `fm-spawn` environment shape (`FM_HOME` at the primary home, overrides cleared): `ScheduleWakeup`, `CronCreate`, `ScheduleJob` deny; `Agent`, `Task`, `SpawnWorker`, `Bash`, `CronDelete`, `CronList` allow.

### DIFF
--- a/bin/fm-brief.sh
+++ b/bin/fm-brief.sh
@@ -356,6 +356,11 @@ The report is the only thing that survives, so anything worth keeping must be in
 7. Never stop, restart, or update the shared \`no-mistakes\` daemon - it is one instance serving
    every lane/home, so restarting it kills other lanes' in-flight pipeline runs. On ANY no-mistakes
    daemon error, append \`blocked: {the daemon error}\` and stop; only firstmate manages the daemon.
+8. The task as briefed is the whole scope. Work you discover mid-task that is not required to finish
+   the briefed task - newly appearing items of the same kind, adjacent cleanups, follow-on opportunities -
+   is reported to firstmate, not absorbed into this task. Report it with a status line and keep going
+   on the briefed scope; firstmate decides whether it becomes a separate task. Corrections genuinely
+   required to make the briefed work correct are part of this task, not new scope.
 
 $INBOX_SECTION
 
@@ -475,6 +480,11 @@ $RULE1
 7. Never stop, restart, or update the shared \`no-mistakes\` daemon - it is one instance serving
    every lane/home, so restarting it kills other lanes' in-flight pipeline runs. On ANY no-mistakes
    daemon error, append \`blocked: {the daemon error}\` and stop; only firstmate manages the daemon.
+8. The task as briefed is the whole scope. Work you discover mid-task that is not required to finish
+   the briefed task - newly appearing items of the same kind, adjacent cleanups, follow-on opportunities -
+   is reported to firstmate, not absorbed into this task. Report it with a status line and keep going
+   on the briefed scope; firstmate decides whether it becomes a separate task. Corrections genuinely
+   required to make the briefed work correct are part of this task, not new scope.
 
 $INBOX_SECTION
 
--- a/bin/fm-subagent-pretool-check.sh
+++ b/bin/fm-subagent-pretool-check.sh
@@ -55,15 +55,22 @@ set -u
 # Lowercase substrings that mark a tool name as delegation-shaped: it creates
 # work, an agent, a schedule, or an isolated workspace that firstmate would not
 # know about. This list is the single owner of the shipped classification.
-DELEGATION_STEMS='agent subagent task workflow cron schedul worktree delegate spawn dispatch handoff remote sendmessage monitor'
+DELEGATION_STEMS='agent subagent task workflow worktree delegate spawn dispatch handoff remote sendmessage monitor'
+
+# Lowercase substrings that mark a tool name as self-scheduling: it re-invokes
+# the caller's own session on a timer. Denied in both a primary and a task
+# worktree because the re-invocation loops compound context costs and bypass
+# supervision.
+SELF_SCHEDULING_STEMS='schedul cron'
 
 # Exact lowercase tool names that match a stem above but only OBSERVE or STOP
-# work that already exists. Reading or ending unaccounted work is not creating
-# it, and denying these would strand already-running work with no way to inspect
-# or end it. A local Claude deny list may still remove these from the
-# schema; this shipped guard deliberately stays narrower so it can never be the
-# reason a runaway task cannot be stopped.
-OBSERVE_ONLY_TOOLS='taskoutput taskstop taskget tasklist cronlist bashoutput killshell'
+# work that already exists (e.g. crondelete stops an existing schedule). Reading
+# or ending unaccounted work is not creating it, and denying these would strand
+# already-running work with no way to inspect or end it. A local Claude deny
+# list may still remove these from the schema; this shipped guard deliberately
+# stays narrower so it can never be the reason a runaway task cannot be stopped.
+# This relaxes the primary too, which is correct and intended.
+OBSERVE_ONLY_TOOLS='taskoutput taskstop taskget tasklist cronlist crondelete bashoutput killshell'
 
 # Exact lowercase tool names that match a stem above but create no RUNNABLE
 # work. These write only the harness's session-local todo list, which has no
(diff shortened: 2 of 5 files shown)


kunchenguid:
Speaking as Kun's firstmate:

Reviewed HEAD `75bb4e9582f63106e5a14fe03d37315f01a02ed3` vs main `10b93b2cc6f4241e87fccaee2e357c33a7347a53`. Full diff reviewed (bin/fm-subagent-pretool-check.sh, bin/fm-brief.sh, docs/subagent-guard.md, tests). No `.github/workflows/*` changes. Not security/fishy.

Class: **default-behavior**. Self-scheduling denial now covers every crewmate/scout task worktree (fleet-wide default capability change; `FM_ALLOW_SUBAGENT=1` remains the escape hatch), and brief rule 8 is a new standing instruction on every ship/scout brief. That is not corrective-only, so this will not auto-merge even when green.

- One captain, one interface — aligns. Unbounded timer loops burn tokens and stay invisible to the captain-facing surface; the deny text points at `paused:`/`blocked:` + supervision-owned polls.
- Authority is explicit — aligns. Denial is the default; deliberate exception is an explicit launch-time env grant. Adjacent work is reported up, not absorbed.
- Scripts own the mechanics — aligns. Stem classification and scope predicates stay in the script; tests cover worktree deny vs ordinary delegation allow.
- A restart is a non-event — aligns weakly. No durable-record change; the point is preventing unsupervised loops that leave nothing durable to reconcile.
- Delegation with a spine — aligns. Strengthens a refusal path and names the existing wait contract instead of inventing a new task shape.
- The fleet outlives any vendor — aligns. Stem shapes, not a vendor UI binding; non-firstmate repos stay inert.
- Scope — aligns. Guard/docs/tests in the command layer; no forge/CI/merge-authority expansion.

What is not cleared:
1. No `no-mistakes-pipeline-attestation:v1` on this HEAD. Blocking (Require-no-mistakes already failed).
2. Greptile P1 (agreed on inspection): `fm_any_firstmate_scope_matches` requires a worktree-local `state/`. When `FM_HOME`/`FM_STATE_OVERRIDE` are unset, `STATE` resolves there; a missing directory fails open and self-scheduling stays allowed in the worktree the change claims to cover.
3. Greptile P1 (agreed): brief rule 8 says "report it with a status line" without naming a form supervision will actually surface (a bare `working:` stays non-actionable).
4. CI is still running after fork approval.

Fork CI approved after this review: CI run `33080006390`, Require no-mistakes run `33080006398`.

This is waiting on the author, not the captain: push a matching attestation, fix the fail-open scope predicate and the brief's actionable status form, and let CI finish green. Even then this stays default-behavior (no auto-merge).

Merge-eligible: NO.

PR #3175, opened 2026-08-27
fix(bin): make session-lock harness identity work on Windows Git Bash

learning2type:
## Problem

On Windows under Git Bash / MSYS, every firstmate session comes up read-only. The session lock identifies the owning session by walking the process ancestry for the harness process (`fm_harness_ancestry_pids`), and it did this exclusively through `ps -o comm=/args=/ppid= -p <pid>`.

The MSYS `ps` bundled with Git Bash does not support that syntax (it rejects `-o`) and cannot see native Windows processes anyway. The Claude harness is a real ancestor two hops up the *native* Windows process tree:

```
bash.exe (tool)  ->  bash.exe (wrapper)  ->  claude.exe  ->  cmd.exe  ->  terminal
```

So the walk terminates immediately, finds no harness, and `bin/fm-lock.sh` reports `cannot locate harness process in ancestry`. The whole home is then held read-only: no spawn, steer, merge, drain, or supervision repair.

## Fix

Introduce a platform-aware process-inspection layer under the two low-level primitives, leaving the harness-identity **policy** (and therefore all Linux and macOS behavior) unchanged:

- `_fm_ancestry_rows` emits the ancestry innermost-first as `pid<TAB>comm<TAB>args`. On Unix it uses the existing `ps -o` walk; on Windows it reads the true tree from `Win32_Process` via a new `bin/fm-win-ancestry.ps1`.
- `fm_harness_pid_alive` reads liveness+identity from `ps -W` (which lists native processes with their Windows pids) instead of `kill -0` + `ps -o`, since the stored pid is a WINPID on Windows.
- Pids are reported in the platform's own space and stored in the lock consistently, so no consumer mixes spaces.
- `FM_LOCK_PLATFORM` overrides detection so the Windows path is exercisable from a Unix test host.

Every existing consumer - lock acquire/inspect, the Claude Stop auto-arm, the turn-end guards, and session start - routes through these two primitives, so all are fixed by this one change.

## Tests

- The existing unix-shaped unit fixtures are pinned to `FM_LOCK_PLATFORM=unix` (they model the `ps -o` world), so they stay valid even on a Windows test host.
- A new Windows-path unit group shadows the three native seams and covers: harness discovery above the Git Bash hops (with `~/.claude` command lines correctly *not* treated as harness evidence), native-table liveness including a two-token `STIME` column, and ownership stopping at the first non-harness above the session.

Verified end-to-end on real Windows 11 Git Bash (harness resolved, liveness and ownership correct) and linted clean with the repo's pinned ShellCheck 0.11.0.

## Scope

This fixes the session lock only. Other scripts (`fm-harness.sh`, `fm-backend.sh`, `fm-teardown.sh`, `fm-watch.sh`, and more) use the same Unix-process pattern and will need equivalent treatment for full Windows operation; those are follow-ups.

### DIFF
--- a/bin/backends/tmux.sh
+++ b/bin/backends/tmux.sh
@@ -58,15 +58,37 @@ fm_backend_tmux_send_text_submit() {  # <target> <text> <retries> <enter-sleep> 
   fm_tmux_submit_core "$@"
 }
 
+# fm_backend_tmux_pane_shell: on Windows, the native path of the shell firstmate's
+# panes must run. psmux (the Windows tmux) defaults panes to PowerShell, but
+# firstmate drives its panes with bash commands (treehouse get, the harness
+# launch, status appends), so every session/window it creates must launch Git
+# Bash instead. Resolve the running MSYS bash to a forward-slash Windows path
+# psmux can spawn. Prints nothing on non-Windows, where the platform default is
+# already a POSIX shell - keeping firstmate self-contained without a per-machine
+# ~/.tmux.conf.
+fm_backend_tmux_pane_shell() {
+  local bash_path
+  case "$(uname -s 2>/dev/null)" in
+    MINGW*|MSYS*|CYGWIN*) ;;
+    *) return 0 ;;
+  esac
+  bash_path=$(command -v bash 2>/dev/null) || return 0
+  cygpath -m "$bash_path" 2>/dev/null || printf '%s\n' "$bash_path"
+}
+
 # fm_backend_tmux_container_ensure: reuse the current tmux session when
 # firstmate itself runs inside tmux, else ensure a dedicated detached
 # "firstmate" session exists. Mirrors fm-spawn.sh's container-ensure block;
 # prints the resolved session name.
 fm_backend_tmux_container_ensure() {
+  local shell
+  local -a shell_arg=()
+  shell=$(fm_backend_tmux_pane_shell)
+  [ -n "$shell" ] && shell_arg=("$shell")
   if [ -n "${TMUX:-}" ]; then
     tmux display-message -p '#S'
   else
-    tmux has-session -t firstmate 2>/dev/null || tmux new-session -d -s firstmate
+    tmux has-session -t firstmate 2>/dev/null || tmux new-session -d -s firstmate "${shell_arg[@]}"
     printf 'firstmate'
   fi
 }
@@ -87,12 +109,15 @@ fm_backend_tmux_container_ensure() {
 # The returned window id lets callers target the window even if its name is ever
 # lost, so worktree discovery cannot fall back to the active client's window.
 fm_backend_tmux_create_task() {  # <session> <window-name> <proj-abs> -> prints window id
-  local ses=$1 wname=$2 proj_abs=$3 wid
+  local ses=$1 wname=$2 proj_abs=$3 wid shell
+  local -a shell_arg=()
   if tmux list-windows -t "$ses" -F '#{window_name}' | grep -qx "$wname"; then
     echo "error: window $ses:$wname already exists" >&2
     return 1
   fi
-  wid=$(tmux new-window -dP -F '#{window_id}' -t "$ses:" -n "$wname" -c "$proj_abs") || return 1
+  shell=$(fm_backend_tmux_pane_shell)
+  [ -n "$shell" ] && shell_arg=("$shell")
+  wid=$(tmux new-window -dP -F '#{window_id}' -t "$ses:" -n "$wname" -c "$proj_abs" "${shell_arg[@]}") || return 1
   tmux set-window-option -t "$wid" automatic-rename off 2>/dev/null || true
   tmux set-window-option -t "$wid" allow-rename off 2>/dev/null || true
   printf '%s\n' "$wid"
--- a/bin/backends/tmux.sh
+++ b/bin/backends/tmux.sh
@@ -124,8 +124,14 @@ fm_backend_tmux_send_literal() {  # <target> <text>
 # fm_backend_tmux_kill: remove one explicitly named task window, best-effort.
 # Empty, omitted, and malformed targets return nonzero before invoking tmux so
 # tmux can never interpret an empty target as the caller's current window.
+#
+# The `=name` exact-match target prefix that keeps this from fuzzy-matching a
+# different window is a real-tmux feature psmux (Windows) does not implement -
+# `kill-window -t "=s:=w"` fails there with "can't find window: =w" and the
+# window leaks. On Windows resolve the window id by an EXACT name match instead,
+# which is equally unambiguous, then kill by that id.
 fm_backend_tmux_kill() {  # <target>
-  local target=${1:-} session window
+  local target=${1:-} session window wid
   case "$target" in
     *:*)
       session=${target%%:*}
@@ -136,7 +142,16 @@ fm_backend_tmux_kill() {  # <target>
   case "$session:$window" in
     :*|*:|*:*:*) return 1 ;;
   esac
-  tmux kill-window -t "=$session:=$window" 2>/dev/null || true
+  case "$(uname -s 2>/dev/null)" in
+    MINGW*|MSYS*|CYGWIN*)
+      wid=$(tmux list-windows -t "$session" -F '#{window_id} #{window_name}' 2>/dev/null \
+        | awk -v n="$window" '$2 == n { print $1; exit }')
+      [ -n "$wid" ] && tmux kill-window -t "$wid" 2>/dev/null || true
+      ;;
+    *)
+      tmux kill-window -t "=$session:=$window" 2>/dev/null || true
+      ;;
+  esac
 }
 
 # fm_backend_tmux_current_command: <target>'s live foreground process name -
--- a/bin/fm-session-lock-lib.sh
+++ b/bin/fm-session-lock-lib.sh
@@ -88,6 +88,106 @@ fm_harness_process_matches() {  # <comm> <args>
   return 1
 }
 
+# --- portable process inspection ---------------------------------------------
+#
+# The ancestry walk and the liveness predicate both need three facts about a
+# process - its command name, full argument string, and parent pid. On Linux and
+# macOS `ps -o comm=/args=/ppid= -p <pid>` supplies them. On Windows under Git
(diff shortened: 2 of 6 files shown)


kunchenguid:
Speaking as Kun's firstmate:

Reviewed HEAD `e976feb30a0aa5ca438dd352db42cb47ed27a4c4` vs main. Whole thread read (Greptile success). Full diff reviewed: `bin/fm-session-lock-lib.sh`, `bin/fm-win-ancestry.ps1`, `bin/backends/tmux.sh`, `bin/fm-spawn.sh`, `bin/fm-teardown.sh`, `tests/fm-session-lock-ancestry.test.sh`. No `.github/workflows/*`. learning2type is not blocked. Not disguised security.

The PowerShell helper only walks `Win32_Process` ancestry and prints `pid<TAB>name<TAB>commandline`. Unix `ps -o` path is unchanged unless `FM_LOCK_PLATFORM` is set. Spawn/teardown/tmux changes are Windows-gated (`MINGW*|MSYS*|CYGWIN*`).

Class: **corrective**. On Windows Git Bash the session lock could not see the native harness ancestor, so every home came up read-only. The later commits fix the same Windows lifecycle (subprocess ancestry, psmux worktree/cwd, teardown kill/retry, Git Bash pane shell).

- One captain, one interface — aligns. A read-only lock that blocks spawn/steer/merge is a hidden failure (`fm-session-lock-lib.sh` Windows walk).
- Authority is explicit — aligns. No new autonomy; lock policy unchanged.
- Scripts own the mechanics — aligns. Platform inspection under the existing two primitives.
- A restart is a non-event — aligns. Lock identity is process-tree, not chat.
- Delegation with a spine — aligns. Existing lock/spawn/teardown, not a new task shape.
- The fleet outlives any vendor — aligns. Windows Git Bash / psmux is another session-manager host, Unix default unchanged.
- Scope — aligns. Command-layer lock and Windows pane lifecycle.

**Attestation:** none on the PR body. no-mistakes is blocking.

**First-time fork CI:** after that full-diff review I approved workflow runs `33135147402` (CI) and `33135147449` (Require no-mistakes).

**Not merging.** Waiting on the author for a HEAD-matching `no-mistakes-pipeline-attestation:v1` and green CI. Not with the captain.

SandroHub013:
Coming here from #3396 at @kunchenguid's direction, to help this PR rather than open a competing one.

I ran this branch's code on a second Windows machine — Windows 11 Home 10.0.26200, **Cygwin ps 3.6.7** (the issue reporter's is 3.4.10), Claude Code primary under WezTerm. Everything below is measured on that box against your actual functions, copied verbatim, not reasoned from the diff. The design is right and `_fm_win_top_msys_winpid` is a cleverer answer than the one I had; the findings are about two parsers and a cost.

## 1. Blocking: `_fm_win_proc_info` reports the live harness as dead here

`fm_harness_pid_alive` returns false for the running harness on this machine, which is the one predicate the lock depends on.

The cause is that `ps -W` does **not** render every process as a Windows path. A process launched from inside the Git Bash shell — which is exactly how the harness starts — is a Cygwin process, and `ps -W` prints its POSIX command:

```
    PID    PPID    PGID     WINPID   TTY         UID    STIME COMMAND
    307     297     307      49996  cons15    197610   Aug 31 /c/Users/39349/.local/bin/claude
4212376       0       0      18072  ?              0   Aug 31 C:\Windows\explorer.exe
```

(Incidentally that second row is also why not walking `ps -W`'s PPID column was the right call: a purely native process gets a synthesized pid — `0x400000 + winpid` — with PPID and PGID reported as `0`.)

The parser anchors COMMAND on the first token matching `^[A-Za-z]:[\\/]` or `^\\\\`. The `claude` row has no such token, so `start == 0`, the row is skipped, and `END { if (!found) exit 1 }` fires. Tested against your function with the real winpids of all three live harness processes:

```
winpid 49996  ->  returns 1   (live harness reported not alive)
winpid 58376  ->  returns 1
winpid 72560  ->  returns 1
winpid 18072  ->  OK  'C:\Windows\explorer.exe'      (purely native: your rule works)
```

The code comment says "a native process whose COMMAND is a bare name (System, Registry) has no such token and is never a harness, so it is correctly reported as not found." That holds for bare names; the class it misses is Cygwin-launched processes, and the harness is in it. I suspect your box launches `claude.exe` natively (from cmd/PowerShell), which is why it renders as a drive-letter path there and this never showed up.

The fix does not need the path shape at all. Your header comment already identifies the only variable-width column: STIME is one token (`HH:MM:SS`) for recent processes and two (`MMM DD`) for older ones. So the COMMAND offset is decidable from STIME alone:

```awk
start = ($7 ~ /^[A-Za-z][a-z][a-z]$/) ? 9 : 8
if (start > NF) next
```

I validated that against every row of `ps -W` on this machine:

| | rows | COMMAND found | discarded |
|---|---|---|---|
| current rule | 512 | 427 | **85** |
| STIME-width rule | 512 | **512** | 0 |

**Zero disagreements** on the resulting column where both rules find one, so it is a strict superset rather than a different reading. The 85 rescued rows are the Cygwin-launched processes (`/usr/bin/bash`, `/c/.../claude`) plus the bare-name natives (`System`, `Registry`, `MemCompression`, `*** unknown ***`).

## 2. Where the time actually goes

Your helper costs **1444–2698 ms** per call here (3 runs, 13-hop walk). Splitting it:

| | ms |
|---|---|
| `powershell.exe -NoProfile` startup, empty command | **1468** |
| `Get-CimInstance Win32_Process` inside an already-running process | 716 |
| the same query with `-Property ProcessId,ParentProcessId,Name,CommandLine` | ~330–600 |

So the dominant cost is **starting powershell.exe**, not the query. Narrowing `-Property` is worth ~150–200 ms and is free to add, but the structural lever is the number of invocations: one call per lock operation, with the table reused for the whole walk, rather than one per consumer. You already fetch the table once per invocation, which is the right shape — it is the invocation itself that is expensive.

For calibration: a bare `bash -c 'echo hi'` costs **279 ms** here (20-run average), so a `powershell.exe` start is worth roughly five bash forks. This helper is expensive, but it is not an outlier on the platform — it is one more process creation on a host where those are the scarce resource.

## 3. Platform detection vs. capability probe

`_fm_lock_platform` branches on `uname -s`. I would probe the capability instead — run `ps -o comm= -p $$` once, cache the answer for the run — for a reason your own PR body names:

> The existing unix-shaped unit fixtures are pinned to `FM_LOCK_PLATFORM=unix` (they model the `ps -o` world), so they stay valid even on a Windows test host.

That pinning is the workaround. **Nine** test files in this suite inject a fake `ps` implementing the `-o` contract — `fm-backend`, `fm-grok-harness`, `fm-kimi-harness`, `fm-remote-job-orphan-reap`, `fm-secondmate-harness`, `fm-session-lock-ancestry`, `fm-session-start`, `fm-startup-network`, `fm-watcher-lock`. Under a platform test the Windows branch is taken on a Windows host and the injected fake is never consulted, so the fixture silently stops testing what it names. Under a capability probe the fake owns the answer wherever it is installed, and `FM_LOCK_PLATFORM` becomes a debugging override rather than a correctness requirement. Cygwin `ps` rejects the unknown option with status 1, so exit status alone separates the two shapes.

One detail that cost me time if you go this way: **probe with `comm`, not `ppid`.** Two fakes here answer a `ppid` request with a non-zero exit, which makes a `ppid`-based probe conclude "no `-o` support" and fall through to the wrong branch:

- `tests/fm-grok-harness.test.sh` implements `comm=` and `args=` only — it never matches `ppid=` at all and ends in `exit 1`.
- `tests/fm-session-start.test.sh` is explicit about it: `*"ppid="*) [ -n "${FM_FAKE_HARNESS_PID:-}" ] || exit 1`.

## 4. A creation-time guard on the Win32 walk is nearly free

`fm-win-ancestry.ps1` follows `ParentProcessId` with no check that the parent predates the child. Windows reuses pids aggressively and does not clear a dead parent's id, so a stale link can point at an unrelated live process and the walk climbs into a foreign branch. Because this walk decides who owns the lock, a false match grants write authority to the wrong session rather than merely returning junk.

Measured here right now: of **462** processes whose parent id resolves to a live process, **2** have a parent created *after* the child. Rare, not theoretical.

`CreationDate` comes back in the same query you already run, so the guard is one comparison:

```powershell
if ($parentObj.CreationDate -gt $p.CreationDate) { break }
```

Walking from a live process here, the chain was monotone across all 11 hops and terminated cleanly when `explorer.exe`'s parent had exited and was absent from the table — so the guard costs nothing on the healthy path.

## 5. One premise that did not reproduce here

The comment on `_fm_win_ancestry_start_winpid` says an MSYS subprocess is fork/exec-orphaned, so "a Win32 walk from here stops at the first hop and never reaches the harness". On this machine it does not stop — walking from this shell's own winpid reaches the harness at hop 5 and keeps going:

```
hop 1 bash.exe   hop 2 bash.exe   hop 3 bash.exe   hop 4 powershell.exe
hop 5 claude.exe   hop 6-8 bash.exe   hop 9-10 herdr.exe   hop 11 powershell.exe
hop 12 wezterm-gui.exe   hop 13 explorer.exe
```

Relatedly, `_fm_win_top_msys_winpid` returns this shell's own winpid here: the MSYS `ps` ppid for this shell is `1`, pid 1 is not in the table, so the climb breaks immediately and your fallback carries it. Net effect on this machine is correct — but by the fallback rather than by the pre-positioning, so the mechanism is untested here. Worth knowing that both environments exist; keeping the climb *and* the fallback, as you do, covers both.

## 6. Correction to something I wrote in #3396

I claimed there that Cygwin `ps` reading native command lines yields bytes iconv rejects, and recommended parsing under `LC_ALL=C`. I could not reproduce that today on `ps -W`: 496 of 496 rows survived a UTF-8 round trip. `LC_ALL=C` is still the honest locale for a byte table, but treat my mojibake claim as unconfirmed until I can pin down what differed in that earlier sample. Flagging it because it is on the public record next to findings that did hold up.

---

Happy to turn items 1 and 4 into commits against `fix/windows-session-lock-ancestry` if that helps — say the word and I will open a PR into your branch, or you can lift the two hunks directly, whichever is less friction. Item 1 is the one I would not ship without, since it makes the Windows liveness predicate fail on a Cygwin-launched harness.

Issue #3206, opened 2026-08-27
Wedge detector escalates on workers holding a declared paused: wait; Grok panes worst-hit because busy detection misses the live footer


josh-padnick:
## In plain English

Firstmate watches its workers and raises an alarm when one goes quiet for too long. The problem: a worker that is *supposed* to be quiet - because it wrote "paused: waiting for the captain's review" and is deliberately holding - still gets treated like a stuck worker. The supervisor then gets woken over and over to check on a worker that is doing exactly what it should, and on Grok panes the alarms escalate hard because the monitor also misreads Grok's busy indicator. Every false alarm costs a supervision turn, and worse, it teaches operators to ignore the alarm that matters when a worker really is stuck. The fix: when a worker has declared a wait, check on it occasionally instead of escalating, and read Grok's actual on-screen footer for busy state. A working patch exists (josh-padnick/firstmate#10); we'd rather it live upstream than as a fork customization.

## The defect

The watcher's stale/wedge triage escalates on workers that are deliberately, legitimately waiting - a task whose latest status line declares `paused: <bounded external wait>` still climbs the escalation ladder (up to demand-deep-inspection) as if it were wedged.

Two compounding parts:

1. **The stale classifier does not consistently honor a declared pause across harnesses.**
   A Claude pane with a declared pause mostly gets the long-cadence recheck, but an idle Grok pane with the same declared pause is escalated repeatedly.
   Observed live: 4 escalations in ~20 minutes on one Grok worker holding a declared pause, with durable state fully verified on every recheck (2026-08-19).
   A lighter form also reproduces on Claude panes: a scout that wrote `paused: advisor open, awaiting the captain in this window` still produced a steady stream of `stale:` wakes on the ordinary cadence (2026-08-27), each costing a supervisor turn to verify and absorb.

2. **Grok busy detection misses the live footer, so an idle-but-healthy Grok pane reads wrong.**
   Grok's only rendered busy signal is the `Ctrl+c:cancel` token in its keybind bar (idle bar shows `Shift+Tab:mode | Ctrl+.:shortcuts`). The current matcher does not track the footer Grok actually renders in current builds, so pause/busy adjudication for Grok panes rests on weaker signals and the wedge path wins.

[...]

PR #3213, opened 2026-08-28
fix(bin): supervise quiet panes on evidence, not on repetition

shadohead:
Refs #3206.

This is the watcher-triage half of #3206: the ordinary `bin/fm-watch.sh` stale path escalating panes that are deliberately, legitimately quiet. It does **not** touch the Grok busy-footer matching that issue also asks for, and it is independent of the alternative patch referenced there (josh-padnick/firstmate#10) - happy to defer or rework if you prefer that direction.

## Violated invariant

Pane-idle supervision must escalate on evidence the supervisor has not already seen.
The crew's own latest declaration sets the recheck cadence, an escalation firstmate has already handled does not repeat on the detection cadence, and a lost endpoint is covered by neither.

The watcher broke that three ways. Each was reproduced against a real watcher poll loop over a scratch state root with a real tmux endpoint **before** any change.

## Root cause

**1. The declared-wait allowance was keyed to the pane hash, not the declaration.**

A live declared wait earns one prompt surface, so an external-decision gate is never hidden behind the long pause cadence. That allowance was keyed to the pane hash. A quiet pane that redraws - a progress tick, a token counter, a resized footer - produces a new hash every few seconds, and each new hash re-earned the same immediate wake. A crew that declared a 40-60 minute wait was surfaced 19 times in 6 minutes on the fixture and never once reached the long cadence its declaration is supposed to buy.

The allowance is now keyed to the declaration's own status-log signature. A redraw under a wait firstmate already knows about is absorbed; only a fresh append re-arms the prompt surface. Measured on the same fixture: 19 stale wakes down to 1.

**2. `wedge_timer_check` deleted its own timer when it escalated.**

The next poll therefore restarted the same short `STALE_ESCALATE_SECS` countdown against an unchanged pane while the escalation counter compounded. Nothing in the bookkeeping consumed the one fact that mattered - that firstmate had already drained the previous escalation, inspected the pane, found it healthy, and acknowledged it. A live crew was alarmed every ~250s for ~35 minutes, reaching `demand-deep-inspection`, with every single wake already handled.

The interval is now evidence-driven. An escalation still queued unacknowledged keeps the detection cadence, so a ladder nobody has read still climbs to `demand-deep-inspection` unchanged. An **acknowledged** repeat on an unchanged pane doubles the required quiet interval, up to `FM_WEDGE_ACKED_BACKOFF_MAX_SECS` (default 1800). The durable wake queue is the acknowledgement authority, read only inside the narrow band where it can change the answer - never once per poll per pane.

**3. A recorded window absent from a successful backend inventory produced no wake at all.**

The stale loop skipped such a window on the pane-capture failure, so a crew that died mid-wait was silent indefinitely, with its own `paused:` line explaining the silence. `paused:` declares that an *external dependency* is pending; it never declares that the worker itself may disappear.

Such an endpoint now surfaces on its own evidence, before and independently of any declared wait, and re-arms if the endpoint returns. Only a `missing` verdict qualifies - every other unreadable answer keeps the previous silence, so a backend hiccup cannot manufacture a wake.

### The two-consecutive-polls decision

The lost-endpoint alarm initially fired on a single `missing` verdict, which is also the normal shape of an ordinary cleanup: `bin/fm-teardown.sh` closes the runtime endpoint *before* it removes the task record, and the watcher enumerates windows from that record without taking the metadata lock. A poll landing in that gap sees a window that is genuinely absent and genuinely still recorded, and alarmed "inspect and recover" for a task completing normally - the same class of false wake this branch exists to remove.

The alarm now requires the missing verdict on two consecutive polls of the same window (`ENDPOINT_GONE_CONFIRM_POLLS`). A worker that is not coming back pays one poll interval of detection latency, which is irrelevant against a process that is already gone; a completing task pays nothing, because its record disappears well before the next poll. Only an unbroken run counts - any other verdict, including the transient `unreadable` a backend hiccup produces, drops the pending confirmation, so two hiccups a minute apart can never add up to an alarm. The pending record also expires after `ENDPOINT_GONE_CONFIRM_POLLS` intervals, so a marker left behind by a retired task cannot let a single missing verdict alarm instantly the next time that window name is recorded.

## Scope

`blocked:` and `needs-decision:` are unaffected - only the `paused:` and captain-held verbs are declared waits. The escalation ladder still reaches `demand-deep-inspection` for an undeclared quiet pane with no prior handled acknowledgement.

## Test evidence

Regression coverage drives the real `bin/fm-watch.sh` poll path against scratch-home fixtures in `tests/fm-watch-triage.test.sh`, not a reimplementation:

- `test_declared_pause_survives_a_redrawing_pane` - the declared-wait cadence across redraws, plus its `blocked:` exclusion.
- `test_acknowledged_wedge_escalation_backs_off` - an acknowledged repeat backs off; an unacknowledged one does not.
- `test_endpoint_gone_alarms_despite_a_declared_pause` - a dead endpoint still alarms under a standing `paused:` line.
- `test_teardown_race_does_not_alarm_as_a_lost_endpoint` - a first missing poll records only a pending confirmation; the task record is then removed as cleanup finishes and no alarm is queued.

The two existing ladder fixtures now pin the backoff cap so they measure the ladder rather than its spacing.

`tests/wake-helpers.sh` gains an opt-in missing-window flag file so a fixture can take an endpoint away and give it back mid-run, driving both the inventory and the capture - the pair real tmux actually presents for an absent target. Real-tmux evidence for that capture-failure behaviour (tmux 3.6a, macOS arm64) is recorded in `docs/verification/runtime-backends.md`. While adding it, `watch_bg`'s documented extra-env arguments were fixed: they went through as expanded words, which bash does not treat as assignments, so they became the command name.

Validated through the repo's `no-mistakes` pipeline (review, tests, document, lint all green) before rebasing onto current `main`.


shadohead:
Closing in favour of a replacement PR raised through the no-mistakes pipeline, so the body carries a genuine pipeline attestation and satisfies the Require no-mistakes gate. This PR was opened by hand after the pipeline's push step failed on fork permissions; the code is unchanged and the replacement will reference #3206 with the same evidence. Apologies for the noise.

PR #3253, opened 2026-08-28
fix: preference files are captain-authorized only, drop no-action chatter

altave-guilherme-monteiro:
## Summary
- `data/captain.md` and `data/captain-shared.md` are now captain-authorized only: the stow skill's tier defaults, per-file action list, and reporting contract are updated consistently so it can read and propose but never rewrite/prune/archive/consolidate either file; the same fix applies to secondmate-provisioning's captain.md trim step.
- AGENTS.md section 6 (and the layout description) now states the explicit-instruction requirement instead of inspect-then-update.
- A routine wake with nothing to act on now ends the turn with no captain-facing text, replacing the fixed "Captain, shipshape." acknowledgement. Section 9's escalation triggers are untouched; the nautical-seasoning mention of "shipshape" is left as-is.

## Test plan
- [ ] `bin/fm-doc-audience-check.sh` passes (ran locally: ok)
- [ ] Manual read-through of stow SKILL.md for internal consistency (tier defaults / action verbs / reporting contract agree)


altave-guilherme-monteiro:
Closing: opened against the wrong repository by mistake. Retargeted to the fork where this change belongs.

Issue #2502, opened 2026-08-16
codex harness: perpetual procevent sources go ownerless across idle turns (no cross-turn continuity)


tiago-peixoto:
## Summary

A Codex-harness primary or secondmate cannot keep a perpetual process-event source alive across idle turns. `bin/fm-watch-checkpoint.sh` reconciles and restarts sources at each checkpoint, but only while a turn is active; when the last actionable turn ends, nothing owns the source until the next captain-triggered turn. A perpetual source that goes single-shot after each result therefore sits ownerless indefinitely.

## Observed behavior, measured

Secondmate on `harness=codex`, owning a perpetual GitHub PR-monitoring source (`bin/fm-procevent.sh`, single-shot after each actionable result by design, relying on ordinary watcher reconciliation to re-arm):

- Last successful inventory `2026-08-15T11:51:57Z`; it emitted sequence 14, the source exited, the result stayed pending.
- At diagnosis ~40 hours later: `bin/fm-procevent.sh list` reported `OWNER none, PENDING 1`; the session-start guard reported the last watcher beat ~144,900 s old.
- Consequence: a captain-authored PR opened during the interval was never inventoried, and a base squash-merge that turned another PR conflict-dirty was never observed. Both had correct predicates; there was simply no observation interval.

Diagnosis and evidence: the secondmate's own report confirmed "Codex checkpoints reconcile and restart sources at each checkpoint, but only while a turn is active, so there is no cross-turn continuity."

## Why this is a harness-protocol defect, not a source defect

The `procevent` contract is explicit: the handler acknowledges the captured result and ordinary reconciliation keeps the perpetual source armed. On Claude the Stop-owned auto-arm (`bin/fm-claude-stop-autoarm.sh`) supplies that continuity across idle turns with no model tokens. On Pi/opencode the extension model does. Codex's protocol is bounded foreground checkpoints (`docs/supervision-protocols/codex.md`) precisely because Codex cannot reason while a foreground call runs - which is correct for wakes but leaves no idle-turn owner for source reconciliation. The same-shaped hazard exists for any Codex home relying on `procevent`.

## What we would find useful

- A Codex-side idle-turn continuity mechanism for procevent reconciliation - the smallest shared correction, not a per-source sidecar or scheduler (which the secondmate correctly refused to build).
[...]

PR #3264, opened 2026-08-29
feat(bin): persistent wake-doorbell owner for Codex-primary supervision

### LINKED ISSUE #3223: Codex-primary supervision cannot wake an idle Codex turn (no extension/service-owned watcher bridge)
## Summary

A Codex-primary firstmate home has no mechanism that can wake an idle Codex turn. Supervision runs as a bounded foreground `bin/fm-watch-checkpoint.sh` loop inside one Codex tool call (`docs/supervision-protocols/codex.md`). While that checkpoint runs, the watcher classifies `signal`/`stale`/`check`/`heartbeat` events and queues actionable ones durably in `state/.wake-queue`. When the model's turn ends, no watcher runs, and nothing can open a new Codex turn: Codex has no extension API, no background-task completion notification, and no Stop-rewake (its Stop hook can only block once per turn with a banner). Durable wake rows therefore wait until the captain happens to type or the model happens to run another fleet command.

## Observed impact

- An idle Codex primary sits blind while a crewmate finishes, reports a decision, or wedges; the wake is durable but undelivered until a human or the model happens to act.
- Perpetual process-event sources go ownerless across idle turns on Codex homes (#2502 is the procevent-side twin of this same root cause).
- The only working injection path into a Codex pane today is the away-mode daemon's, and it is presence-gated to `state/.afk` (it exists only while the captain is away).

## Why the foreground checkpoint cannot carry this

Codex cannot reason while a foreground tool call is running, and a foreground checkpoint dies with its tool call. A checkpoint observes events only while the model is actively spending a turn on it; between turns there is no owner. Shell `&`, `nohup`, `disown`, and untracked detached terminals are explicitly forbidden supervision mechanics (they are reaped by the harness or leak unowned processes).

## What would fix it

A supported, extension-/service-owned path (Codex has no extension API, so this must be a service, not a Pi-style extension) that:

1. Persists independently of one Codex foreground tool call and wakes an idle Codex turn by injecting the canonical operational prompt into the exact active Codex primary pane, e.g.
   `FIRSTMATE_OP: v1 watcher: FIRSTMATE WATCHER WAKE: stale: default:wH:p2 (idle 244s, possible wedge, escalation 1)`
   followed by the canonical instruction to run `bin/fm-wake-drain.sh` first and handle the queued wake.
2. Uses the existing verified backend submit path with positive submit acknowledgement (`bin/fm-backend.sh`'s `fm_backend_send_text_submit`) and the canonical operational-input grammar (`bin/fm-operational-input.sh`, kind `watcher`).
3. Treats injection as a doorbell only: `.wake-queue` durability and the generation-bound `fm-wake-drain`/ack contract stay the source of truth; injection deduplicates against what was already doorbelled.
4. Binds ownership to one exact canonical Firstmate home, session lock, watcher generation, backend session, workspace/tab/pane, and Codex primary identity - never inferring authority from labels, never sweeping ambiguous panes, never allowing duplicate owners, and refusing cross-home injection.
5. Detects a busy or non-empty composer and defers safely with bounded, identity-revalidated retries; never overwrites captain input.
6. Rearms continuously across handled wakes and ordinary Codex turn completion, and fails closed (standing down, never injecting) on session start/shutdown, crash recovery, stale owners, lock-refused read-only mode, session replacement, and watcher-generation replacement.
7. Works for Herdr (the current captain backend) and tmux, and leaves away-mode (`.afk` / `fm-supervise-daemon.sh`) semantics untouched.

## Constraints that must hold

- Codex CLI remains the primary harness; no Pi-primary requirement is introduced.
- The durable wake queue and the `WAKE_ACK_REQUIRED` generation-bound acknowledgement remain the only source of truth for handling.
- No shell `&` / `nohup` / `disown` / untracked detached terminal as the wake mechanism.

Related: #2502 (procevent continuity across idle turns - same root cause, per-source surface).

SilentMathematician:
Closes #3223. Related to #2502.

## Problem

Codex CLI is the captain-approved PRIMARY Firstmate, but Codex cannot reason while a foreground tool call runs and offers no extension API, background-task completion notification, or Stop-rewake. External `signal`, `stale`, `check`, and `heartbeat` wakes therefore sat durable in `.wake-queue` until a human or the model happened to open a new turn; supervision required a foreground `fm-watch-checkpoint` loop held by the model itself.

## Approach

Add `bin/fm-watch-codex-bridge.sh`: one persistent owner, launched by the model's own foreground arm call in a tracked non-visible backend terminal, that keeps a live watcher cycle running across ordinary Codex turn boundaries and injects the canonical `FIRSTMATE_OP: v1 watcher: <wake>` doorbell into the exact recorded primary pane through the backend's verified submit path with positive acknowledgement.

- Codex CLI remains PRIMARY; no Pi-primary launcher, preference, or requirement is added or retained.
- Ownership binds one exact canonical Firstmate home, session-lock pid and identity, watcher generation, backend session, workspace/tab/pane ids, and the Codex primary identity (the pane's shell must positively contain the session pid in its descendant process tree); every injection revalidates all of it and stands down instead of injecting on any mismatch.
- The launch binding rides a 0600 state record rather than the tracked terminal's keystroke transport, which truncates long command lines issued from inside a pane.
- Delivery is doorbell-only: the durable `.wake-queue` and the generation-bound `fm-wake-drain` acknowledgement stay the source of truth; duplicate suppression is exactly nothing-newer-than-the-last-confirmed-injection.
- Busy panes and pending composers defer behind bounded identity-revalidated retries; captain input is never overwritten.
- Away mode stands the bridge down; `fm-supervise-daemon.sh` semantics are untouched.
- The bounded foreground checkpoint remains the documented fallback; the renderer, arm-command policy, supervision protocol, and verification records move to the bridge as the owner.
- Herdr support uses the backend's guarded non-default lab contract for live tests; the captain's `default` session is never touched, and tmux compatibility is reviewed without weakening either backend.

## Tests

- `tests/fm-watch-codex-bridge.test.sh`: portable regression covering ownership binding, exact prompt encoding, queue durability, duplicate suppression, busy-composer deferral, verified delivery, rearming, stale/replaced panes, cross-home/session refusal, cleanup, and failure recovery.
- `tests/fm-watch-codex-bridge-herdr-e2e.test.sh`: opt-in guarded real-Herdr-lab e2e (arm, inject, dedupe, busy and pending-composer deferral, replaced-session stand-down) run green against the live guarded lab.
- `bin/fm-lint.sh` green (shellcheck full extended analysis + actionlint); `fm-test-run.sh --changed` green on this head, with the one remaining `fm-calm-pi-extension` live-e2e failure demonstrated pre-existing on the base commit (fails with a different case label per run, independent of this diff).


SilentMathematician:
Closing per the repository owner's direction: contributions should not be submitted to this public upstream. The work is moving to a private clone on the author's own GitHub; this PR will not be re-opened here.

SilentMathematician:
closed

PR #3295, opened 2026-08-29
feat(bin): detect a worktree claimed by two task records

cisrd:
## Problem

A pooled task worktree can end up recorded by two `state/*.meta` records at once: one task's teardown
partially fails and leaves its record behind, the pool re-leases the slot to a new task, and nothing
notices that two records now name the same checkout. Acting on the stale record then reaches into a
copy a live task is still using.

## Change

Adds a detect-only check that reports a worktree claimed by more than one record, and says which
kind of collision it is.

The design that survived review is a two-axis one, and the reason matters:

- **`kind` reflects process concurrency.** Two live agents contending is a different fact from work
  that might be lost, and fusing them is what produced the original misattribution.
- **Unlanded work is always surfaced as a path-level caveat**, never as a kind trigger, so the
  do-not-discard warning cannot be swallowed by a label change.

The check refuses to claim more than its probes established. Every probe that cannot answer -
a path git cannot inspect, an unreadable working-tree state, a reachability test that cannot run,
an empty path argument - resolves to one shared `unverifiable:<cause>` verdict that names its cause
and keeps the do-not-discard caveat. Only a probe that actually succeeded may report `landed`.

Grouping compares resolved physical paths rather than the recorded strings, so two records spelling
one checkout differently are still detected; the printed line shows the resolved key alongside each
claimant's own recorded spelling.

The dirty-work filter is byte-for-byte the one `bin/fm-teardown.sh` already uses, so the detector and
the component that actually discards cannot disagree about what "dirty" means.

## Documentation

`.agents/skills/bootstrap-diagnostics/SKILL.md`'s `stale` bullet previously routed the reader into
ordinary teardown, which reads the worktree path from the stale record's own metadata and returns
the shared checkout. It now states that a stale claimant's bookkeeping is reconciled by removing
only that claimant's own state records, never through teardown's physical worktree-return path.

The two-axis contract has one owner (the library); `bin/fm-bootstrap.sh` and the skill point at it
rather than restating it, after three copies had already drifted apart once.

## Validation

Ran through the no-mistakes pipeline: review, test, document and lint completed. The push step could
not publish - the account has pull-only access to this repository - so the branch is delivered from
the `cisrd` fork. No validation step was skipped.

## Known remainder, in order of seriousness

1. `group-key-recheck-can-silence-a-real-collision` - a real collision can be silenced by a
   time-of-check/time-of-use race between the two grouping passes.
2. `printed-path-documented-as-always-resolved` - three places document the printed path as always
   resolved, while the code falls back to the recorded spelling when resolution fails.
3. `enotdir-ancestor-blamed-on-permissions` - an `ENOTDIR` ancestor is reported as "not consultable"
   when the real cause is different.

A follow-up is separately tracked to harden `bin/fm-teardown.sh` itself to refuse when the target's
worktree is claimed by another record whose process reads alive; that protects every caller rather
than only the reader of the diagnostics skill, and is a deliberate scope decision rather than an
oversight.


cisrd:
Superseded by #3329, which carries the identical reviewed content on a fresh branch for a new no-mistakes pipeline run.

PR #3293, opened 2026-08-29
feat(backends): add thurbox as an experimental session-provider backend

LeTuR:
## Intent

Rebind the no-mistakes pipeline attestation in PR #3293 to the exact current PR head. The thurbox backend content was fully validated earlier in this session (review, test, document and lint all completed clean, and the CI phase's own docs/config correction was locally validated); this pass rebases onto current main, re-reviews, pushes and re-stamps so the PR description binds the head that CI will run against.

## What Changed

- Add the thurbox backend adapter (`bin/backends/thurbox.sh`), driven entirely through `thurbox-cli` on the 2.10 surface, and wire it into backend detection/known-list, spawn, busy-state, Cursor composer capture, teardown, control, bootstrap, config inheritance, hometag, and daemon supervision (`bin/fm-backend.sh`, `bin/fm-spawn.sh`, `bin/fm-busy-lib.sh`, `bin/fm-composer-lib.sh`, `bin/fm-supervisor-target-lib.sh`, `bin/fm-teardown.sh`, `bin/fm-tmux-lib.sh`, `bin/fm-control.sh`, `bin/fm-control-lib.sh`, `bin/fm-cursor-lib.sh`, `bin/fm-bootstrap.sh`, `bin/fm-config-inherit-lib.sh`, `bin/fm-backend-hometag-lib.sh`, `bin/fm-afk-launch.sh`, `bin/fm-supervise-daemon.sh`, `bin/fm-test-run.sh`).
- Document the new backend (`docs/thurbox-backend.md`) and sync thurbox into existing backend enumerations across `README.md`, `AGENTS.md`, `CONTRIBUTING.md`, `docs/architecture.md`, `docs/configuration.md`, `docs/scripts.md`, `docs/agent-control.md`, `docs/trace-context.md`, `docs/codex-app-backend.md`, `docs/herdr-backend.md`, `docs/tmux-backend.md`, `docs/verification/runtime-backends.md`, `docs/documentation-audiences.json`, and the affected skill references.
- Add thurbox-specific test coverage (`tests/fm-backend-thurbox.test.sh`, `tests/thurbox-test-safety.sh`) and extend existing backend, daemon, control, bootstrap, busy-state, and test-run suites to exercise thurbox handling alongside the other session-provider backends.

## Risk Assessment

[...]

PR #3287, opened 2026-08-29
fix(bin): widen Grok busy signature past its approval dialog

cisrd:
## Intent

Finish PR 3287 by authenticating the already-reviewed Grok idle-misclassification fix through a complete no-mistakes pipeline run. Preserve the rebased branch and its two accepted documentation corrections, publish the resulting head to the configured cisrd fork, and let the pipeline write the PR attestation. Do not skip any pipeline step and do not hand-compose or hand-edit the attestation.

## What Changed

- Widened `FM_DELIVERY_GROK_BUSY_REGEX_DEFAULT` (and the defensive literal fallback in `fm_busy_grok_tail_busy`) from `Ctrl\+c:cancel` to `Esc:cancel|Ctrl\+c:cancel`, so a Grok pane that is thinking, responding, writing a command, or running an approved tool now classifies `busy grok-regex` instead of idle — `Ctrl+c:cancel` only ever renders in Grok's separate tool-approval dialog.
- Deliberately left `Esc:cancel` out of the harness-less delivery union `FM_DELIVERY_BUSY_REGEX_DEFAULT`, keeping a mid-turn Grok steer at the safe unconfirmed `pending` verdict rather than converting it to `empty`, since Grok's mid-turn Enter is not verified to queue the way opencode's does; the exclusion and its rationale are documented inline in `bin/fm-composer-lib.sh`.
- Recorded the grok 1.0.5 live captures in `docs/verification/runtime-backends.md` and the harness reference, including the re-verified `Ctrl+C` interrupt behavior and an explicit open question about whether Escape still only focuses scrollback (`fm_control_interrupt_key` stays on `Ctrl+C` until that is re-verified); added regression tests in `tests/fm-busy-state.test.sh` and `tests/fm-tmux-submit-busy.test.sh` covering all three footer shapes, the fallback literal without the canonical owner sourced, and the union-scope boundary.

## Risk Assessment

✅ Low: The change widens exactly one per-harness busy signature plus an in-sync defensive literal, every affected call site moves in the conservative (defer rather than act) direction, the delivery union is deliberately and correctly left untouched, and the new tests exercise the real classifier against live-captured footer shapes rather than source text.

## Testing

[...]

PR #3303, opened 2026-08-29
fix(spawn,teardown): keep briefs out of agent argv, and confirm endpoint retirement before releasing a worktree

matthewstrud:
## The defect

Crewmate sessions were dying mid-task: agents exiting cleanly while still working, each after roughly an hour of API spend, each leaving unfinished work in its worktree.

Neither of the two hypotheses under investigation was the cause.

**Root cause.** `bin/fm-spawn.sh` expanded each worker's entire brief into the launch command. A command substitution's result becomes one argv element, so every crewmate carried ~12.5 KB of its own brief in `/proc/<pid>/cmdline` and in every `ps` listing on the host. That made the agent process itself **matchable by any word of its own brief**. A worker clearing stray project processes with the ordinary `ps -ef | grep <script> | awk '{print $2}' | xargs -r kill` idiom therefore selected four sibling `claude` processes and terminated them.

Evidence: the killing command is in one worker's own transcript at `10:47:59.500Z`; four agents died between `10:47:59.994` and `10:48:00.322`, the killer last. All four dead briefs contained the matched word; the one live worker whose brief did not, survived.

**Reproduced end to end.** Pre-fix, the pattern matched only the agent process; `SIGTERM` produced exactly the observed clean shutdown records and left a `claude --resume` line over a bash prompt. Post-fix, the same pattern matches nothing and the agent keeps working.

## Fixes

1. **argv** — the launch prompt carries a short pointer; the brief stays on disk. This generalises the pattern kimi has always used, so the pointer sentence has one owner shared with kimi's typed delivery. An unwritable pointer refuses the spawn rather than launching on an empty prompt. A raw launch command that expands the brief as prompt text is refused.
2. **Doorbell** — the steering doorbell is submitted as keystrokes, and once an agent exits the endpoint's owner is the worker's shell at a prompt, so the line executed as a shell command (observed as `bash: Firstmate: command not found`). The ring now refuses on a proven agent-free endpoint. The gate is narrow: only recovery-grade `dead`/`missing` verdicts refuse, so ambiguous or unverified endpoints ring exactly as before and no live worker is starved.
3. **One worktree, one live task** — teardown released the worktree before retiring the endpoint, so a refused close left the pool calling it free while the task's record still claimed it. Release now runs after confirmed endpoint retirement *and* after all fallible task-state retirement. Retirement is confirmed per backend: tmux, Herdr, Zellij and cmux require exact structural absence, Orca requires a typed native close acknowledgment, and any other backend refuses by name. Every refusal retains both the record and the provider reservation.
4. **Process rule** — generated briefs and `AGENTS.md` ban pattern kills and give the explicit-PID remedy.

**The spawn-side guard and the teardown ordering are two deliberate protections, not one with a duplicate.** Refusing a worktree another live task's record names holds regardless of backend and regardless of whether any close can be confirmed; the teardown ordering holds even when no second task ever tries to claim the worktree. Neither should later be removed as redundant with the other.

Regressions accompany each fix and were each verified to fail against the pre-fix code.

## Two things a reader should know

**This branch closes a test failure that predates it.** `tests/fm-secondmate-harness.test.sh` was failing at the branch base with "first config push did not reach pointer delivery". The fixture work here fixed it and the suite now passes clean, so that suite going from failing to green is this branch's doing rather than unrelated noise.

**A separate defect is filed, deliberately not fixed here.** Teardown's leaked-process reap depends on `lsof`; where `lsof` is absent it degrades silently, so teardowns can fail to reap processes they believe they have reaped. It is the same defect class as the confirmation work above — a cleanup step that cannot confirm its own success and proceeds anyway — and is tracked separately.

## What validating this cost

The lint step was killed by the kernel out-of-memory killer six times. Each kill took the shared validation daemon down with it and cancelled the run in flight, presenting as an unexplained shutdown followed by a clean restart seconds later. It was identified only from the systemd journal reporting `Failed with result oom-kill` at an 8.4 GB peak, with the victim named as `ghc_ticker` — a Haskell runtime thread, ShellCheck being a Haskell program. Every one of those lint logs contained only its two startup lines, because extended analysis ballooned past 8 GB and was killed before emitting a single finding.

What was hiding behind all six kills was **two informational notes about deliberate single quotes** in a test file. Nothing was wrong with the code.

The interim in place on that host is a host-side setting of one analyser worker instead of two. Measured peak 6.54 GB against 8.4 GB, both from systemd cgroup accounting on the same basis — a 1.9 GB saving, not the halving a naive reading suggests, because one worker lints the whole file set rather than a shard. It is **not a substitute for swap** on a 15 GB host that has none: the remaining margin is roughly 2 GB and it was measured with the machine otherwise idle. It is deliberately host-side rather than pinned into tracked configuration, which would halve lint throughput for every contributor on every machine to work around one undersized host.


kunchenguid:
Speaking as Kun's firstmate:

Reviewed HEAD `8417ba5ed99ab6009047cdfd399ed6fcccb4f8e5` vs main. Whole thread read (body, Greptile 5/5, Copilot/Codex nits, no human reviews). 45 files; spawn/teardown/inbox/send/brief/backend/tmux + docs/tests/AGENTS. **No** `.github/workflows/*`. matthewstrud is not blocked. Not disguised security.

Contract-class: **restore**. Isolated copies, one live task per worktree, and send-to-agent-not-shell were already specified and broken. Pointer-not-brief keeps brief text out of argv (kimi already launched from a pointer; raw `__BRIEF__` refused). Doorbell refuses only recovery-grade `dead`/`missing` (`fm_backend_agent_alive` maps both to `dead`); ambiguous endpoints still ring; the steer stays durable. Spawn refuses a worktree another live meta still names; teardown confirms endpoint retirement (tmux/Herdr/Zellij/cmux structural absence, Orca typed close-ack) before releasing the provider reservation. That uniqueness was already assumed by teardown's leaked-process reap. Always-on confirmation generalizes herdr's already-specified fail-closed retention; it is not a new captain-facing default. `tmux_window_id=` is the retirement identity for new tmux metas; legacy records without it still fall back.

VISION (per rule):
- One captain, one interface — aligns (sibling-kill and doorbell-as-shell were silent fleet failures).
- Authority is explicit — aligns (no new autonomy; unconfirmed teardown retains).
- Scripts own the mechanics — aligns (pointer, claim check, live-agent gate, backend confirmation + tests).
- A restart is a non-event — aligns (durable inbox; retain record+reservation for retry).
- Delegation with a spine — aligns (isolated copies; unlanded work not torn down).
- The fleet outlives any vendor — aligns (harness-agnostic pointer; per-backend retirement semantics).
- Scope — aligns (command-layer spawn/send/teardown).

No `<!-- no-mistakes-pipeline-attestation:v1 ... -->` in the PR body. no-mistakes is blocking. Fork CI approved this pass: `33277214581` (CI), `33277214532` (Require no-mistakes). Waiting on the author to add a no-mistakes pipeline attestation bound to this HEAD. Also reconcile `tests/fm-teardown.test.sh` persistent-index-lock (expects meta gone on failed return) with teardown.sh / worktree-ownership (keep meta on failed release), and the ShellCheck SC2086 fixtures Copilot flagged. Overlap with open #3287 (docs only) and #3293 (spawn/teardown/backend) — not a captain-hold; I am not merging over it. Not with the captain (not otherwise ready). No auto-merge. I am not opening a competing PR.

matthewstrud:

PR #3326, opened 2026-08-30
feat(bin): deliver allowlisted composer commands as keystrokes into this home's own session

tomastoth4:
## Summary

When the captain writes a command such as `/compact` in the firstmate UI composer, it was only ever filed as an ordinary note (`bin/fm-inbox.sh note`) and read as conversation text by firstmate - never actually executed, because a slash command is intercepted locally by the CLI's own input loop, not sent to the model. This adds a default-off path that delivers a recognized note as real keystrokes into **this home's own** primary tmux session, reusing the existing tmux submit primitives (`bin/fm-tmux-lib.sh`) the away-mode daemon and `bin/fm-send.sh` already rely on.

- **Recognition rule**: a note body is a command invocation only when, after trimming leading/trailing whitespace, it is byte-identical to one entry in a fixed allowlist. `/compact now`, `please /compact`, wrong case, extra lines, or any argument all fail to match and the note behaves exactly as it does today - queued as text, nothing typed anywhere.
- **Allowlist**: `/compact` only, hardcoded in `bin/fm-composer-command-lib.sh` (not configurable - the config file can only ever turn delivery of this fixed, reviewed set on or off, never grow it). It's local and fully reversible with no side effect outside the current session's own context, which is why it's the only entry that clears the "never destructive, irreversible, or security-sensitive" bar.
- **Session resolution**: firstmate can only identify its own pane with certainty from *within* its own live turn (`$TMUX_PANE` is genuinely inherited there). Since note delivery is triggered by `bin/fm-inbox.sh` running as a separate process with no such inheritance, resolution is split in two: `fm_composer_command_session_start` captures `discover_supervisor_target`/`discover_supervisor_backend` (the same primitives `FM_SUPERVISOR_TARGET`/`FM_SUPERVISOR_BACKEND` use) once, at session start, and binds the record to that session's lock pid; `fm_composer_command_session_effective` only ever trusts that record while the bound lock still matches the *current* lock, so a restart into a different session invalidates the prior record instead of resolving it as current. Nothing is ever scanned for by pane name and nothing is guessed - an unresolvable session refuses loudly instead of guessing a target.
- **Busy-session decision: refuse and report, never silently retry.** The delivery guard reuses `bin/fm-tmux-lib.sh`'s existing delivery-busy primitives verbatim (`fm_pane_is_busy`, `fm_tmux_composer_state`) - the same ones `bin/fm-supervise-daemon.sh`'s `inject_msg` already uses to decide whether it's safe to type into a live pane. A busy pane or a composer that isn't confirmed empty defers delivery with a clear, reported reason rather than queueing an automatic retry: firstmate cannot safely poll for the pane going idle without either creating a second supervision cycle (forbidden) or hooking the safety-critical turn-end/auto-arm chain (out of scope and too risky for this change). The note itself stays queued exactly as it does today, so nothing is ever silently dropped - the captain and firstmate both see plainly that it wasn't auto-executed and can resend once idle.
- **No double delivery, including across a restart**: delivery is keyed by the note's own durable id. A confirmed submit (the tmux composer actually cleared) writes a durable, empty marker file under `state/.composer-command-delivered/<key>` *before* returning success, and every call checks that marker first. A short-lived, mkdir-based per-key claim additionally rejects a second concurrent attempt at the same key.
- **Config knob**: `config/composer-commands` - local, gitignored, presence-only, default off. Absent means byte-for-byte today's behavior. Documented in `docs/configuration.md` ("Composer command delivery") and `AGENTS.md`'s config inventory, matching the existing entries' shape.

## Files

- `bin/fm-composer-command-lib.sh` (new): the single owner of recognition, session capture/resolution, the delivery guard, and idempotency. Full contract and rationale in its header.
- `bin/fm-inbox.sh`: `queue_note` attempts delivery for a recognized note after queueing it as before; every other note is untouched.
- `bin/fm-session-start.sh`: one gated call (`fm_composer_command_session_start`) that captures the durable session record, mirroring how `config/trace-context` wires in.
- `docs/configuration.md`, `AGENTS.md`: document the new config file.
- `tests/fm-composer-command-lib.test.sh`, `tests/fm-composer-command-inbox.test.sh`: unit + integration coverage (see Test plan).

## Acceptance - real end-to-end demonstration

Demonstrated against a **throwaway scratch session**, never the captain's live firstmate session:

1. Created a private tmux socket (`fm-composer-demo-<pid>`), fully isolated from any real fleet session.
2. Launched a fresh `claude --dangerously-skip-permissions` process in that socket's pane, cwd a scratch empty directory with no repo config - a brand-new session with zero relation to any real firstmate home.
3. Confirmed via the real classifier that its idle composer reads `empty` (`fm_tmux_composer_state`).
4. Ran the *real* production code (`fm_composer_command_session_start` with `TMUX_PANE` pointed at that scratch pane, then `fm_composer_command_deliver "/compact" ...`) against a scratch `FM_HOME`.
5. Verified `/compact` was genuinely typed and submitted: the pane's scrollback shows `❯ /compact` followed by Claude Code's own real response, `⎿  Not enough messages to compact.` - proof the CLI's local slash-command interceptor actually ran it, not just displayed the text.
6. Verified idempotency: a second `fm_composer_command_deliver` call for the same key returned `already delivered`, with no further tmux interaction.
7. Killed the private tmux server and removed every scratch file; the real firstmate session, its watcher, and `/home/tom/firstmate/{state,data,config}` were never touched.

## Test plan

- [x] `bin/fm-lint.sh` (full repo, ShellCheck 0.11.0 + actionlint 1.7.12) - clean.
- [x] `bash tests/fm-composer-command-lib.test.sh` - recognition/allowlist, enablement, session capture bound to the lock (including restart invalidation), the full delivery pipeline (busy/composer defer, unconfirmed submit, gone endpoint, unresolved session, success, idempotent replay after a simulated restart, concurrent in-flight claim) - all pass.
- [x] `bash tests/fm-composer-command-inbox.test.sh` - ordinary notes byte-for-byte unchanged when the feature is off or the note isn't a recognized command; a recognized command with no durable session record visibly attempts delivery and reports why; a command-shaped note outside the allowlist is refused before any session resolution - all pass.
- [x] `bash tests/fm-composer-lib.test.sh`, `bash tests/fm-voice-relay.test.sh` (adjacent suites touching the composer classifier and `bin/fm-inbox.sh`) - unaffected, all pass.
- [x] `bash tests/fm-session-start.test.sh` - 47/49 pass; the remaining 2 (`test_output_ordering_diagnostics_lead`, `test_composition_invokes_real_scripts`) fail identically against an **unmodified checkout of this branch's base commit**, caused by this sandbox having a system-wide `/usr/bin/node` the fixture's fake-toolchain PATH doesn't account for - a pre-existing environmental gap, not a regression from this change (verified by reverting `bin/fm-session-start.sh` alone and re-running).
- [x] `bin/fm-doc-audience-check.sh` - clean.


tomastoth4:
Closing: this work is moving to the captain's own repository.

PR #3367, opened 2026-08-31
fix(bin): absorb orphaned status, turn-end, and check markers

sctru:
## Intent

Determine whether firstmate's watcher still has the 'orphaned marker' problem against CURRENT code (the watcher, the shared classifier library, and the supervise-daemon) - a marker file (a task's status log, its turn-end ping, or its custom check script) that survives with no matching task-metadata record getting surfaced as a wake instead of being absorbed - and if it does, re-implement the guard against current code. A July 2026 local fix for this (commits 8b9222e, 62bf0f4, f43a968, 5bf45c6) was dropped during a 2026-08-30 upstream reconciliation that substantially rewrote the watcher, and it was unknown whether upstream's rewrite (a presentation-cursor dedup, a generation-bound wake acknowledgement mechanism, a watcher-downtime marker) already covered the same ground.

Diagnosis phase (scout, no code changes) reproduced the problem in an isolated scratch home and confirmed it still occurs, in two forms: (1) a repeatedly-touched orphan status/turn-end log (the classic leaked-hook-still-firing-after-cleanup case) still re-surfaces forever, though a one-time static leftover now self-heals thanks to upstream's cursor dedup; (2) an orphan custom check produces an undeduped "rejected unauthenticated state checks" wake every check interval with no self-healing at all, which is a new, worse-than-July failure mode introduced by upstream's registration-based check authorization model. Findings and the disconfirming-evidence check are recorded in data/fm-orphan-guard-x2/report.md.

Implementation phase (this branch) was then explicitly authorized to re-implement the guard exactly as the report recommended, with these accepted constraints: do not cherry-pick the four old commits as-is, since their target functions were renamed and restructured; re-implement equivalent logic at current call sites; the two generated fixed-name poll shims must stay exempt, since both are keyed by a constant id rather than a spawned task id and so neither ever has a matching metadata record by design, and a genuine authorization failure for either must keep surfacing rather than vanish silently; anything belonging to a live task (a metadata record present) must behave identically before and after, since this is the fleet's supervision path and a mistake here silently breaks monitoring for every project; reinstate the July orphan tests retargeted at current function names, especially the repeated-touch regression test, the highest-value one since it is the case the static-dedup mechanism alone does not cover; load the coding-guidelines skill before touching this shared tracked material; never touch the default branch or the backup refs.

During implementation, verification against the real test suite surfaced one more genuine call site beyond the report's original three: the watcher's own always-on twin of the daemon's catch-all backstop, documented in its own header comment that way, previously uncounted. It got the identical guard. Verification also caught a real design bug in the first custom-check implementation: an upfront metadata-presence gate would have wrongly deleted a validly-registered custom check that has no task-shaped id, confirmed via an existing security test's own control-check fixture, and by extension any future non-task custom check registered through the operator registration entrypoint. The corrected design applies the metadata check only within the "neither poll nor trust registration validated" rejection path, so a check with valid trust registration is never affected regardless of whether it has a task id.

Fixing the guard required adding metadata fixtures to several pre-existing tests across five test files that never needed one before this guard existed, including a genuine architectural case: a remote secondmate's parent-side status file also needs a metadata record recording it as a secondmate, matching how a real registered secondmate is recorded, which one affected test fixture was missing. The full watcher/daemon-dependent test surface (13 files) was run to completion; two failures were confirmed via a from-scratch pristine-checkout comparison (identical failure with zero of this branch's changes applied) to be pre-existing environment issues unrelated to this change, and are not regressions from this work.

## What Changed

- `bin/fm-watch.sh`: `scan_signals` and `heartbeat_scan_finds_actionable` now skip any `.status`/`.turn-ended` file that has no matching `state/<id>.meta`, treating it as an orphan from a torn-down (or never-recorded) task rather than a wake candidate; `scan_signals` also best-effort removes the orphaned file so a leaked writer can't leave it permanently "fresh". The custom-check rejection path applies the same `.meta` check before adding a check to `rejected_checks`, deleting the orphaned `.check.sh`/`.check-trust` pair instead of reporting it — except for the `tool-updates` fixed-name poll shim (alongside the existing `x-watch.check.sh` carve-out), which is exempt because it is keyed by a constant id and never has a matching `.meta` by design, so a genuine authorization failure there still surfaces.
- `bin/fm-supervise-daemon.sh`: `housekeeping`'s catch-all `.status` scan applies the identical `state/<id>.meta` guard (read-only skip; the watcher's own scan owns removal).
- `CONTRIBUTING.md`: updates the watcher re-arm smoke test snippet to also create a matching `.meta` fixture alongside the `.status` fixture, since the new guard would otherwise absorb it as an orphan.
- Test suite: reinstates orphan-guard coverage across `tests/fm-watch-triage.test.sh` (including a repeated-touch regression case and unregistered tool-updates bypass coverage), `tests/fm-daemon.test.sh`, `tests/fm-watch-arm.test.sh`, `tests/fm-afk-inject-e2e.test.sh`, and `tests/fm-wake-daemon-lifecycle-e2e.test.sh`/`tests/fm-wake-queue.test.sh`, and adds `.meta` fixtures to several pre-existing tests (including a remote secondmate's parent-side status file) that previously didn't need one.

## Risk Assessment

✅ Low: All four call sites (fm-watch.sh's scan_signals, heartbeat_scan_finds_actionable, the check-loop rejected-checks path, and fm-supervise-daemon.sh's housekeeping catch-all) correctly gate on state/<id>.meta presence before absorbing a marker, the tool-updates/x-watch fixed-name exemptions are scoped precisely to the rejection branch (never bypassing a valid trust registration), and both prior review-round fixes (CONTRIBUTING.md .meta pairing, the missing negative test for the tool-updates exemption) verify correct against current code with no regressions; tracing concrete orphan/live/exempt/broken-registration scenarios through the code produced no wrong-result case, and the change conforms to every constraint in the stated intent.

## Testing

All targeted watcher/daemon orphan-guard tests pass (fm-watch-triage.test.sh's 8 new orphan-guard cases plus the heartbeat-backstop case, fm-daemon.test.sh's catch-all orphan case, fm-watch-arm.test.sh's secondmate-metadata fixture case, and the untouched-but-dependent afk/wake-queue/lifecycle e2e suites), and manual CLI verification of the CONTRIBUTING.md smoke-test fix shows the documented command now correctly surfaces the actionable signal with the required .meta fixture while reproducing the exact silent-absorption false negative when that fixture is omitted. The one observed failure (a process-event fixture timeout in fm-watch-triage.test.sh) was verified via a pristine base-commit worktree to be a pre-existing environment issue unrelated to this change, matching what the author's intent already documented.

<details>
<summary>Evidence: CONTRIBUTING.md watcher re-arm smoke test — with the fixed .meta fixture (post-fix behavior)</summary>

```text
$ tmp=$(mktemp -d) && printf 'done: smoke\n' > "$tmp/smoke.status" && printf 'window=fake\n' > "$tmp/smoke.meta" && FM_STATE_OVERRIDE="$tmp" FM_SIGNAL_GRACE=1 FM_POLL=1 FM_HEARTBEAT=999999 bin/fm-watch-arm.sh
watcher: started pid=738440 (beacon fresh)
signal: /tmp/tmp.jLjqROffdz/smoke.status
EXIT=0
```
</details>
<details>
<summary>Evidence: Same smoke test WITHOUT .meta — reproduces the pre-fix false negative (orphan guard silently absorbs and deletes smoke.status)</summary>

```text
$ tmp=$(mktemp -d) && printf 'done: smoke\n' > "$tmp/smoke.status" && FM_STATE_OVERRIDE="$tmp" FM_SIGNAL_GRACE=1 FM_POLL=1 FM_HEARTBEAT=999999 timeout 10 bin/fm-watch-arm.sh
watcher: started pid=740019 (beacon fresh)
EXIT=124 # timed out, no 'signal:' line printed, smoke.status deleted out from under the caller
```
</details>

## Pipeline

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>
<summary>🔧 **Review** - 2 issues found → auto-fixed (2) ✅</summary>

- ⚠️ `CONTRIBUTING.md:99` - CONTRIBUTING.md&#39;s documented watcher re-arm smoke test creates only a `.status` file (`$tmp/smoke.status`) with no matching `.meta`. After this change, scan_signals()&#39;s new orphan guard in bin/fm-watch.sh will silently absorb and delete that file instead of surfacing &#39;an actionable signal&#39; as the comment promises, so a contributor who copy-pastes this exact command to smoke-test a watcher change gets a false negative (silence instead of a signal, and the fixture file is deleted out from under them). The July 2026 predecessor fix (commit 5d0e962) hit the identical problem on this same line and fixed it by adding a matching `$tmp/smoke.meta` write; that doc update was not carried over in this re-implementation.
- ℹ️ `docs/architecture.md:49` - The July 2026 fix documented the &#39;a .status/.turn-ended/.check.sh with no matching state/&lt;id&gt;.meta is absorbed, even under afk (except state/x-watch.check.sh)&#39; behavior contract in AGENTS.md, docs/architecture.md, and docs/scripts.md. All three doc updates were dropped along with the code during the 2026-08-30 upstream rewrite, and this re-implementation (correctly) restores the code and tests but does not restore the doc prose describing the now-real contract again, so the documented behavior contract and the actual code are back in sync only by coincidence, not by explicit documentation.

🔧 Fix: docs: pair CONTRIBUTING.md smoke test .status with .meta
1 warning still open:

- ⚠️ `tests/fm-watch-triage.test.sh:1672` - test_tool_updates_check_sh_survives_orphan_guard_and_runs registers tool-updates.check.sh through the real fm-check-register.sh entrypoint, giving it valid check-trust. With valid trust, fm_custom_check_snapshot_prepare() (bin/fm-check-lib.sh:48) succeeds and the check runs via the `elif` branch in bin/fm-watch.sh&#39;s check loop, never reaching the `else` fallback where the new `[ &#34;$id&#34; = tool-updates ] || [ -e &#34;$STATE/$id.meta&#34; ]` bypass at bin/fm-watch.sh:1630 lives. This test would pass identically with or without that bypass condition, so it provides no coverage for the specific behavior the intent calls out: &#39;a broken tool-updates registration must keep surfacing, exactly like x-watch&#39;s own name-based carve-out.&#39; A negative case (tool-updates.check.sh present but unregistered/broken, no state/tool-updates.meta) analogous to test_live_task_unregistered_check_sh_still_surfaces is needed to actually exercise that line.

🔧 Fix: test: cover unregistered tool-updates orphan-guard bypass
</details>

<details>

✅ No issues found.
- `bin/fm-test-run.sh tests/fm-watch-triage.test.sh (core orphan-guard regressions: test_orphan_turn_ended_never_surfaces, test_orphan_turn_ended_recreated_every_poll_never_surfaces, test_orphan_status_with_captain_relevant_verb_never_surfaces, test_orphan_check_sh_never_runs, test_live_task_unregistered_check_sh_still_surfaces, test_x_watch_check_sh_survives_orphan_guard_and_runs, test_tool_updates_check_sh_survives_orphan_guard_and_runs, test_tool_updates_check_sh_unregistered_still_surfaces, test_heartbeat_backstop_never_surfaces_an_orphan_status — all ok)`
- `bin/fm-test-run.sh tests/fm-daemon.test.sh (covers housekeeping()'s catch-all orphan guard incl. test_heartbeat_scan_orphan_status_not_escalated — all ok)`
- `bin/fm-test-run.sh tests/fm-watch-arm.test.sh (covers the remote secondmate .meta fixture in test_rearm_resurfaces_durable_queue_and_remote_open_decision — all ok)`
- `bin/fm-test-run.sh tests/fm-afk-inject-e2e.test.sh tests/fm-wake-daemon-lifecycle-e2e.test.sh tests/fm-wake-queue.test.sh (all ok)`
- `Manual comparison: re-ran tests/fm-watch-triage.test.sh against a disposable git worktree checked out at the pristine base commit a56a78a with none of this branch's changes applied — the single failing case (test_procevent_captured_result_surfaces_proactively / 'the fixture captured no process-event result') reproduced identically, confirming it is a pre-existing environment issue, not a regression`
- `Manual CLI verification of the CONTRIBUTING.md smoke test: ran the documented command with the new smoke.meta fixture (signal correctly surfaces, exit 0) and re-ran it without smoke.meta to reproduce the pre-fix false negative (orphan guard silently absorbs and deletes smoke.status, watcher times out with no signal line)`
</details>

<details>

✅ No issues found.
</details>

<details>

🔧 Fix: Install pinned ShellCheck 0.11.0 and actionlint 1.7.12 in lint environment
1 warning still open:

</details>

<details>

✅ No issues found.
</details>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -96,7 +96,7 @@ bin/fm-test-isolation-proof.sh --pool watcher-wake-lock --jobs 4   # re-run an a
 @AGENTS.md
 EOF
 [ "$(readlink .claude/skills)" = "../.agents/skills" ]
-tmp=$(mktemp -d) && printf 'done: smoke\n' > "$tmp/smoke.status" && FM_STATE_OVERRIDE="$tmp" FM_SIGNAL_GRACE=1 FM_POLL=1 FM_HEARTBEAT=999999 bin/fm-watch-arm.sh  # watcher re-arm smoke test (prints arm status, then an actionable signal)
+tmp=$(mktemp -d) && printf 'done: smoke\n' > "$tmp/smoke.status" && printf 'window=fake\n' > "$tmp/smoke.meta" && FM_STATE_OVERRIDE="$tmp" FM_SIGNAL_GRACE=1 FM_POLL=1 FM_HEARTBEAT=999999 bin/fm-watch-arm.sh  # watcher re-arm smoke test (prints arm status, then an actionable signal); the .meta file is required, else the orphan guard absorbs the status as a torn-down task's leftover
 ```
 
 `bin/fm-test-run.sh` is the single owner of behavior-suite selection, portable CI lane composition, bounded concurrency admission, per-script timing markers, family totals, the coverage guard, and the optional JSON timing artifact.
--- a/bin/fm-supervise-daemon.sh
+++ b/bin/fm-supervise-daemon.sh
@@ -1142,6 +1142,13 @@ housekeeping() {  # <state>
     for f in "$state"/*.status; do
       [ -e "$f" ] || [ -L "$f" ] || continue
       task=$(basename "$f"); task="${task%.status}"
+      # Same orphan guard as bin/fm-watch.sh's scan_signals: a .status
+      # surviving with no matching state/<id>.meta belongs to a torn-down (or
+      # never-recorded) task and has no owner for firstmate to act on, so it
+      # must never wake firstmate via this catch-all either. Skipped here, not
+      # removed - the watcher's own per-poll scan already best-effort removes
+      # the file, so this stays a pure read.
+      [ -e "$state/$task.meta" ] || continue
       record=$(status_span_first_actionable_record "$f" \
         "$(status_seen_offset "$state" "$task")")
       rc=$?
--- a/bin/fm-watch.sh
+++ b/bin/fm-watch.sh
@@ -15,7 +15,12 @@
 # on every wake. Printed reason lines:
 #   signal: <file>...      status/turn-end signals, surfaced when a listed status
 #                          span has a captain-relevant event OR a no-verb signal lacks
-#                          positive execution evidence, unless afk is active
+#                          positive execution evidence, unless afk is active. A
+#                          .status/.turn-ended with no matching state/<id>.meta is
+#                          an orphan from a torn-down task (or a leaked writer
+#                          still touching its old path) and is always absorbed
+#                          and removed, even under afk - never classified, never
+#                          actionable, no matter how many times it is recreated.
 #   stale: <window>        a provably-working stale is ALWAYS absorbed (with a wedge
 #                          timer) regardless of what the status log says - an active
 #                          run-step or busy pane outranks even a captain-relevant log
@@ -70,7 +75,11 @@
 #                          captured generation, never again while that record
 #                          stays queued and never once it is acknowledged
 #   check: rejected unauthenticated state checks: <paths>
-#                          unsafe state checks were refused without execution
+#                          unsafe state checks were refused without execution.
+#                          A check.sh with no matching state/<id>.meta is instead
+#                          an orphan (unless its fixed name is an exempt shim like
+#                          x-watch.check.sh or tool-updates.check.sh) and is
+#                          absorbed and removed here rather than reported
 #   check: rejected unauthenticated PR poll retirement receipts: <paths>
 #                          invalid pending retirements were preserved without
 #                          running a check or removing poll artifacts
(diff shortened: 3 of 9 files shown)


sctru:
Closing this one - we're carrying this change locally rather than upstream. No action needed on your side.

PR #3368, opened 2026-08-31
fix(bin): accept Orca's own worktree-id shape in endpoint validation

mlim1972:
## Summary
- `fm_backend_validate_task_endpoint()` validated `orca_worktree_id` with the generic tmux/Herdr-style atom check, which bans `:` and `/`.
- Orca's real worktree ids are `<repo-id>::<absolute-path>` (see `orca worktree show`), so every real Orca-backed task failed this check.
- Effect: `bin/fm-control.sh <id> exit` and `bin/fm-teardown.sh <id>` permanently refused with "Orca endpoint metadata ... is malformed or inconsistent" for every Orca task, including fully merged/landed ones.
- Added `fm_backend_orca_worktree_id_valid()`, a dedicated check for just this field that also allows `:` and `/`. The value is always consumed as a quoted `--worktree` CLI argument (`bin/backends/orca.sh`), never shell-interpolated, so this doesn't reopen any injection risk.
- Added a regression test using a realistic `<repo-id>::<path>` id; existing simplified test fixtures (e.g. `wt-teardown`) still pass unchanged.

## Test plan
- [x] `bin/fm-lint.sh bin/fm-backend.sh` - clean (ShellCheck 0.11.0)
- [x] `bin/fm-lint.sh tests/fm-backend-orca.test.sh` - clean
- [x] `bash tests/fm-backend-orca.test.sh` - 52/52 pass
- [x] `bash tests/fm-backend.test.sh` - 28/28 pass (no regression on other backends)

### DIFF
--- a/bin/fm-backend.sh
+++ b/bin/fm-backend.sh
@@ -383,6 +383,17 @@ fm_backend_endpoint_atom_valid() {  # <value>
   esac
 }
 
+fm_backend_orca_worktree_id_valid() {  # <orca-worktree-id>
+  # Orca's own id shape is "<repo-id>::<absolute-path>" (see `orca worktree
+  # show`), so ':' and '/' are legitimate here unlike other backends' atoms.
+  # Always consumed as a quoted --worktree "id:$worktree_id" CLI argument
+  # (bin/backends/orca.sh), never shell-interpolated, so widening the charset
+  # does not reopen injection risk.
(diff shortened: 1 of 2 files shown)


mlim1972:
Opened by mistake against the wrong repo (gh defaulted to the fork parent) - closing and re-opening against mlim1972/firstmate.

PR #3384, opened 2026-08-31
fix(bin): launch crewmates under their harness's own approval and sandbox controls

npayette84:
## Intent

Harden Firstmate's crewmate launch defaults so a spawned worker runs under its harness's own approval, review, and sandbox controls instead of with them switched off, and solve unattended network, write, and prompt behavior for Codex, Cursor, and Claude in this one reviewed change. This is the first of two deliberately split PRs; the narrow Codex writable-roots grant is a separate stacked PR and must NOT be pulled forward into this one.

Launch posture: Claude uses --permission-mode auto with CLAUDE_CODE_DISABLE_FAST_MODE=1; Codex uses '-s workspace-write -a never' plus an explicit sandbox_workspace_write.network_access grant; Cursor uses '--trust --auto-review --sandbox enabled'.

Captain rulings, all explicit and all in scope - do not flag them as undeclared scope and do not revert them:

1. Codex network. workspace-write defaults network_access to false and Firstmate never set it, so on a default machine a codex worker could not push, use gh, or install dependencies. The grant is now explicit. Write confinement stays intact; the full-bypass flag was NOT restored.

2. Cursor. --auto-review enforces the sandbox but prompts, and --force never prompts but was MEASURED to defeat --sandbox enabled while being a documented alias of the --yolo it replaced. Shipping --force was refused as false hardening. Cursor therefore keeps real filesystem confinement and is BARRED from every ordinary unattended worker kind: ship, scout, and secondmate. The bar holds however cursor was selected, EXPLICIT OR INHERITED from runtime auto-detection. An unconfigured cursor-hosted home is stopped with an exact instruction to set config/crew-harness or add an eligible crew-dispatch profile; Firstmate never silently substitutes another tool. This is a deliberate refusal, not a regression.

3. Cursor exemptions. The only opt-in is a per-invocation --cursor-exemption naming 'attended' (a person is in the pane) or 'envelope:<name>' (a named, mechanically proven outer isolation envelope such as the approved routing benchmark). It is never ambient and never inherited from the environment. Across relaunch, an envelope grant IS inherited because it describes a durable property that still governs the replacement agent, while an attended grant is NEVER inherited because the person who attested may be gone and firstmate's own stuck-worker recovery relaunches unattended. A cursor exemption is refused on a non-cursor harness rather than recorded, so no stale grant can later authorize a relaunch onto cursor. Remote secondmate launches forward only an envelope grant, through self-describing arguments whose compatibility degradation is stated in code and fails closed in both directions; attended is refused for a remote launch because it cannot describe a worker on another host.

4. Claude. Only CLAUDE_CODE_DISABLE_FAST_MODE=1 was verifiable, and unverified levers were deliberately rejected rather than cargo-culted. The captain KNOWINGLY ACCEPTS the residual risk that server-controlled triggers (plan, model, settings, classifier transcript) can still degrade auto mode and strand an unattended worker; that residual risk is documented in operator guidance rather than coded away. Documented residual risk is the accepted outcome, not a defect.

5. Operator guidance was refreshed across the adapter references, AGENTS.md, docs/configuration.md, and docs/agent-control.md, and bootstrap now diagnoses an ineligible cursor crew-dispatch rule instead of validating clean and failing at spawn.

An earlier fix round that would have made auto-detected cursor silently fall back to another adapter was explicitly invalidated by the captain, aborted through the supported abort command, custody recovered through the structured next action, and never committed. Do not reintroduce any silent-fallback or adapter-substitution behavior.

Repo conventions: one full sentence per line in tracked Markdown, plain dashes not em dashes, shellcheck-clean bin scripts, no agent co-author, and tests must exercise behavior through executable interfaces rather than asserting implementation source bytes.

## What Changed

- Crewmate launch templates now run each worker under its own harness's approval and sandbox controls instead of switching them off: `claude` launches with `--permission-mode auto` plus `CLAUDE_CODE_DISABLE_FAST_MODE=1` (replacing `--dangerously-skip-permissions`), `codex` with `-s workspace-write -a never` and an explicit `-c sandbox_workspace_write.network_access=true` grant (replacing `--dangerously-bypass-approvals-and-sandbox`, since workspace-write also confines egress and would otherwise break `git push`, `gh`, and dependency installs), and `cursor` with `--trust --auto-review --sandbox enabled` (replacing `--yolo`). `grok` is deliberately untouched.
- `bin/fm-control-lib.sh` gains a shared admissibility layer (`fm_control_launch_refusal` composed from `fm_control_harness_supports_kind`, `fm_control_cursor_exemption_valid`, `fm_control_cursor_exemption_inherited`, and friends) that now owns the former muse-secondmate refusal and bars cursor from ship, scout, and secondmate spawns however cursor was selected. The only way past it is a per-invocation `--cursor-exemption attended|envelope:<name>`, recorded in `state/<id>.meta` and echoed on the success line. `bin/fm-spawn.sh`, `bin/fm-control.sh` relaunch (asked before the old agent is stopped, so a refusal cannot strand a task), and the remote secondmate path all ask that one function. An envelope grant is inherited across relaunch and secondmate liveness recovery; `attended` never is, and a grant on a non-cursor harness is refused rather than recorded.
- Remote secondmate launch forwards only envelope grants, over self-describing `traceparent:`/`exemption:` trailing arguments that fail closed on both sides of a version skew; `bin/fm-bootstrap.sh` now reports a `CREW_DISPATCH` advisory when a crew-dispatch rule names cursor, instead of validating clean and failing later at spawn. Operator guidance was refreshed across the harness adapter references, `AGENTS.md`, `docs/configuration.md`, and `docs/agent-control.md`, including the accepted residual risk that server-side triggers can still drop claude's auto mode back to prompting. Tests cover the launch flags, the exemption grammar, inheritance and non-inheritance, the wire contract, and the bootstrap advisory; a pre-existing bash 3.2 empty-array crash in the remote home seed is fixed along the way.

## Risk Assessment

⚠️ Medium: The cursor policy, grant inheritance, pre-stop composite, wire contract, and launch postures all now match the intent and are covered by behavior-driving tests; the two residual issues are an operator-output contract gap in bootstrap and a misconditioned remedy clause, both safe to address as follow-ups.

## Testing

I reproduced the round-1 defect exactly as reported against the pre-fix file and confirmed the fixed file gets past it: before, a --no-projects seed died at the empty-array expansion and exited 0 because bash 3.2's EXIT trap status masks the failure, so it looked like a silent success that provisioned nothing; after, the loop is skipped and the seed reaches the first real remote step and fails honestly with exit 1. Because the fix is in an otherwise untouched production script, I checked it is not just satisfying the new test - the pre-existing tests/fm-remote-secondmate-lifecycle-e2e.test.sh, which died at the same call, now runs green, and the branch's new tests/fm-remote-secondmate-trace-context.test.sh executes on stock macOS bash for the first time, so the remote cursor wire boundary that round 1 could only verify by hand is now machine-verified in both compatibility directions. I then re-ran the six suites covering the intent itself (launch posture for claude/codex/cursor, the cursor unattended bar across ship/scout/secondmate, exemption grant validity and relaunch inheritance, bootstrap's ineligible-cursor-rule diagnosis) and all passed with no gate skips. Evidence is a before/after CLI transcript plus the named wire assertions the fix unblocked; the end-to-end launch transcript and vendor flag corroboration captured last round at 8e6b1f9 remain accurate, since the only commit since touches the seed script and no launch path. No UI surface is involved in this change, so no visual artifact applies.

<details>
<summary>Evidence: bash 3.2 remote seed crash: before/after CLI transcript, plus the remote wire coverage the fix unblocked</summary>

Source: [bash 3.2 remote seed crash: before/after CLI transcript, plus the remote wire coverage the fix unblocked](https://github.com/npayette84/firstmate/blob/0ead3a0fa7256ec85ec562dc20f504b49129cf4f/.no-mistakes/evidence/fm/harden-crew-launch-sandbox/bash32-remote-seed-before-after.txt)

```text
Pre-existing bash 3.2 empty-array crash in the remote seed - before/after
branch fm/harden-crew-launch-sandbox, target commit 5d8bcab
Captured 2026-08-31T22:50:31Z on Darwin 27.0.0 (arm64), GNU bash, version 3.2.57(1)-release (arm64-apple-darwin26)

bin/fm-remote-home-seed.sh expanded "${PROJECT_NAMES[@]}" on an empty array
under `set -u`. Stock macOS bash 3.2 rejects that as an unbound variable, so
every --no-projects seed died at that line before doing any work. The defect is
independent of this branch's launch hardening; it is fixed here only because the
branch's new remote secondmate coverage cannot run on macOS without it.

Both runs below drive the real bin/fm-remote-home-seed.sh with an identical
stub ssh that answers the readiness probe and refuses the dispatch that follows,
so the only difference between them is the one-line array expansion. Paths are
scrubbed to <repo> and <tmp>.


########## BEFORE (bin/fm-remote-home-seed.sh at 8e6b1f9, the fix's parent) ##########

$ bash --version -> GNU bash, version 3.2.57(1)-release (arm64-apple-darwin26)
$ fm-remote-home-seed.sh probe remote-mac /srv/fm-root /srv/fm-home --no-projects
<repo>/bin/fm-remote-home-seed.sh: line 190: PROJECT_NAMES[@]: unbound variable
  exit=0

  The seed never reached the remote host, and worse, it reported exit 0: in
  bash 3.2 the EXIT trap's own final status replaces the failing status, so the
  crash surfaced to the operator as a silent success that provisioned nothing.
  Measured separately with the status captured directly:

    $ FM_HOME=<tmp>/home ... fm-remote-home-seed.sh probe remote-mac /srv/fm-root /srv/fm-home --no-projects
    <repo>/bin/fm-remote-home-seed.sh: line 190: PROJECT_NAMES[@]: unbound variable
    BEFORE_EXIT=0


########## AFTER (bin/fm-remote-home-seed.sh at 5d8bcab) ##########

$ bash --version -> GNU bash, version 3.2.57(1)-release (arm64-apple-darwin26)
$ fm-remote-home-seed.sh probe remote-mac /srv/fm-root /srv/fm-home --no-projects
stub-ssh: reached the remote dispatch step
error: remote runtime preflight failed; nothing was provisioned. Close the gaps listed above, or update the remote code root if it predates the current fm-remote-doctor.sh
  exit=1

  The project loop is now skipped for a project-less seed, the run continues to
  the first real remote step, and the failure it hits there is reported honestly
  with exit 1 instead of being masked.


########## What the fix unblocked: the remote coverage now runs on stock macOS bash ##########

Both remote secondmate suites reach a --no-projects seed on their first step, so
before the fix neither could execute at all on this machine. Both now run green
under the same stock bash 3.2, including the wire behavior round 1 could only
check by hand:

$ bin/fm-test-run.sh tests/fm-remote-secondmate-trace-context.test.sh tests/fm-remote-secondmate-lifecycle-e2e.test.sh

  ok - remote route: an unattended cursor secondmate is refused before the remote host is reached
  ok - remote route: a cursor exemption is refused on a non-cursor harness rather than recorded
  ok - remote route: an envelope grant and a carrier both cross the wire and reach the remote spawn
  ok - recovery: an unflagged respawn inherits the recorded envelope grant and brings the secondmate back
  ok - recovery: a recorded attended grant is never inherited, so it cannot resurrect a cursor secondmate
  ok - wire: an older parent's bare positional carrier is still accepted as the traceparent
  ok - wire: a bare argument that is not a carrier is still refused
  ok - wire: the receiver forwards only an envelope grant, refusing attended and malformed grants
  ...
  ok - remote seeding proceeds once the repair closes every gap        <- the pre-existing
                                                                          test that died at
                                                                          the same call
  FM_TEST_SUMMARY total=2 failed=0 skipped_gate=0 duration_ms=459328
```
</details>
<details>
<summary>Evidence: Hardened crewmate launch defaults: end-to-end CLI transcript (captured at 8e6b1f9; launch paths unchanged since)</summary>

Source: [Hardened crewmate launch defaults: end-to-end CLI transcript (captured at 8e6b1f9; launch paths unchanged since)](https://github.com/npayette84/firstmate/blob/0ead3a0fa7256ec85ec562dc20f504b49129cf4f/.no-mistakes/evidence/fm/harden-crew-launch-sandbox/launch-hardening-transcript.txt)

```text
Firstmate hardened crewmate launch defaults - end-to-end CLI transcript
branch fm/harden-crew-launch-sandbox, target commit 8e6b1f9
Captured 2026-08-31T22:32:37Z on Darwin 27.0.0 (arm64), GNU bash 3.2.57

Every block below is real output from bin/fm-spawn.sh, bin/fm-control.sh,
bin/fm-remote-secondmate-control.sh and bin/fm-bootstrap.sh driven against an
isolated firstmate home and a fake tmux endpoint that records the literal
command firstmate types into a crewmate pane. Absolute paths are scrubbed to
<repo> and <tmp>.


########## 1. Claude crewmate runs under claude's own approval classifier ##########

$ fm-spawn.sh ev-claude-1 <project> --mode no-mistakes --yolo off      # crew-harness=claude
spawned ev-claude-1 harness=claude kind=ship mode=no-mistakes yolo=off window=firstmate:fm-ev-claude-1 worktree=<tmp>/claude/wt

  [command typed into the crewmate pane]
  env -u CURSOR_AGENT -u CURSOR_INVOKED_AS CLAUDE_CODE_ENABLE_PROMPT_SUGGESTION=false CLAUDE_CODE_DISABLE_FAST_MODE=1 claude --permission-mode auto "$('<repo>/bin/fm-operational-input.sh' encode launch-brief < '<tmp>/claude/home/data/ev-claude-1/brief.md')"

  BEFORE this change the same launch was: claude --dangerously-skip-permissions ...


########## 2. Codex crewmate: workspace sandbox, no prompts, explicit network egress ##########

$ fm-spawn.sh ev-codex-1 <project> --mode no-mistakes --yolo off       # crew-harness=codex
spawned ev-codex-1 harness=codex kind=ship mode=no-mistakes yolo=off window=firstmate:fm-ev-codex-1 worktree=<tmp>/codex/wt

  [command typed into the crewmate pane]
  env -u CURSOR_AGENT -u CURSOR_INVOKED_AS codex -s workspace-write -a never -c sandbox_workspace_write.network_access=true -c "notify=[\"bash\",\"-c\",\"touch '/private/var/folders/70/p814fs691f103nxddhx58mhh0000gn/T/fm-evidence-demo.WVQnz7/codex/home/state/ev-codex-1.turn-ended'\"]" "$('<repo>/bin/fm-operational-input.sh' encode launch-brief < '<tmp>/codex/home/data/ev-codex-1/brief.md')"

  BEFORE: codex --dangerously-bypass-approvals-and-sandbox ...
  NOW:    -s workspace-write -a never, plus the network grant a worker needs to push, use gh, and install deps.


########## 3. Cursor is barred from every ordinary unattended kind ##########

$ fm-spawn.sh ev-cursor-ship <project> --mode no-mistakes --yolo off   # crew-harness=cursor
error: cursor is a verified adapter but is refused for an unattended ship launch: its --auto-review classifier prompts for calls it does not deem safe, the pane has no approver, and the parked pane keeps reading as busy. The bar applies however cursor was selected, INCLUDING when firstmate inherited it by detecting its own runtime, because silently substituting another tool would change which adapter runs the captain's work without saying so. If this home is running inside cursor and resolved it that way, set config/crew-harness to a verified adapter such as codex or claude, or add a crew-dispatch profile eligible for this kind; firstmate will not choose one for you. If a person is in the pane or a proven outer isolation envelope governs this worker, pass it on the invocation itself with --cursor-exemption attended or --cursor-exemption envelope:<name>, which both bin/fm-spawn.sh and bin/fm-control.sh's relaunch verb accept.
  exit=1
  [crewmate pane: nothing was ever launched]

$ fm-spawn.sh ev-cursor-scout <project> --scout
error: cursor is a verified adapter but is refused for an unattended scout launch: its --auto-review classifier prompts for calls it does not deem safe, the pane has no approver, and the parked pane keeps reading as busy. The bar applies however cursor was selected, INCLUDING when firstmate inherited it by detecting its own runtime, because silently substituting another tool would change which adapter runs the captain's work without saying so. If this home is running inside cursor and resolved it that way, set config/crew-harness to a verified adapter such as codex or claude, or add a crew-dispatch profile eligible for this kind; firstmate will not choose one for you. If a person is in the pane or a proven outer isolation envelope governs this worker, pass it on the invocation itself with --cursor-exemption attended or --cursor-exemption envelope:<name>, which both bin/fm-spawn.sh and bin/fm-control.sh's relaunch verb accept.
  exit=1
  [crewmate pane: nothing was ever launched]


########## 3b. The bar holds when cursor was INHERITED from runtime auto-detection ##########

$ CURSOR_INVOKED_AS=cursor-agent fm-spawn.sh ev-cursor-detect <project> --mode no-mistakes --yolo off   # no crew-harness configured
error: cursor is a verified adapter but is refused for an unattended ship launch: its --auto-review classifier prompts for calls it does not deem safe, the pane has no approver, and the parked pane keeps reading as busy. The bar applies however cursor was selected, INCLUDING when firstmate inherited it by detecting its own runtime, because silently substituting another tool would change which adapter runs the captain's work without saying so. If this home is running inside cursor and resolved it that way, set config/crew-harness to a verified adapter such as codex or claude, or add a crew-dispatch profile eligible for this kind; firstmate will not choose one for you. If a person is in the pane or a proven outer isolation envelope governs this worker, pass it on the invocation itself with --cursor-exemption attended or --cursor-exemption envelope:<name>, which both bin/fm-spawn.sh and bin/fm-control.sh's relaunch verb accept.
  exit=1  -- an exact instruction to configure, never a silent substitution onto another adapter


########## 3c. A cursor SECONDMATE is refused too, with the knob that governs ITS resolution ##########

$ fm-spawn.sh sm-ev <secondmate-home> --secondmate    # config/secondmate-harness=cursor
error: cursor is a verified adapter but is refused for an unattended secondmate launch: its --auto-review classifier prompts for calls it does not deem safe, the pane has no approver, and the parked pane keeps reading as busy. The bar applies however cursor was selected, INCLUDING when firstmate inherited it by detecting its own runtime, because silently substituting another tool would change which adapter runs the captain's work without saying so. If this home is running inside cursor and resolved it that way, set config/secondmate-harness to a verified adapter such as codex or claude, which always governs a secondmate, or set config/crew-harness while config/secondmate-harness is unset or "default", since a secondmate resolves the former first and falls back to the latter; crew-dispatch profiles are not consulted for a secondmate; firstmate will not choose one for you. If a person is in the pane or a proven outer isolation envelope governs this worker, pass it on the invocation itself with --cursor-exemption attended or --cursor-exemption envelope:<name>, which both bin/fm-spawn.sh and bin/fm-control.sh's relaunch verb accept.
  exit=1  -- a whole firstmate instance would stall invisibly; note the remedy names config/secondmate-harness
           and does NOT offer a crew-dispatch profile, which a secondmate never consults.


########## 4. The only opt-in: a per-invocation --cursor-exemption ##########

$ fm-spawn.sh ev-cursor-env <project> --mode no-mistakes --yolo off --cursor-exemption envelope:routing-benchmark
spawned ev-cursor-env harness=cursor kind=ship mode=no-mistakes yolo=off cursor_exemption=envelope:routing-benchmark window=firstmate:fm-ev-cursor-env worktree=<tmp>/cursorexempt/wt

  [command typed into the crewmate pane]
  env -u CLAUDECODE -u PI_CODING_AGENT -u GROK_AGENT -u FM_PI_HARNESS -u CURSOR_INVOKED_AS '<tmp>/cursorexempt/fake/fakebin/cursor-agent' --trust --auto-review --sandbox enabled --workspace '<tmp>/cursorexempt/wt' "$('<repo>/bin/fm-operational-input.sh' encode launch-brief < '<tmp>/cursorexempt/home/data/ev-cursor-env/brief.md')"

  [audit trail recorded in state/ev-cursor-env.meta]
  harness=cursor
  kind=ship
  mode=no-mistakes
  yolo=off
  cursor_exemption=envelope:routing-benchmark

  BEFORE: cursor --trust --yolo ("Run Everything"). NOW: real filesystem confinement, --force/--yolo deliberately 

... [264 bytes truncated] ...

gits, '.', '_', or '-' (the named outer isolation envelope that governs this worker); 'envelope:routing benchmark' names neither, and a grant that is unnamed, free-form, or carrying a line break could neither be audited later nor written into the task record without displacing another recorded field


########## 5. A grant is never ambient and never leaks to the next spawn ##########

$ fm-spawn.sh ev-leak-a <project> ... --cursor-exemption attended
spawned ev-leak-a harness=cursor kind=ship mode=no-mistakes yolo=off cursor_exemption=attended window=firstmate:fm-ev-leak-a worktree=<tmp>/cursorleak/wt

$ FM_CURSOR_EXEMPTION=attended CURSOR_EXEMPTION=attended fm-spawn.sh ev-leak-b <project> ...   # no flag on THIS invocation
error: cursor is a verified adapter but is refused for an unattended ship launch: its --auto-review classifier prompts for calls it does not deem safe, the pane has no approver, and the parked pane keeps reading as busy. The bar applies however cursor was selected, INCLUDING when firstmate inherited it by detecting its own runtime, because silently substituting another tool would change which adapter runs the captain's work without saying so. If this home is running inside cursor and resolved it that way, set config/crew-harness to a verified adapter such as codex or claude, or add a crew-dispatch profile eligible for this kind; firstmate will not choose one for you. If a person is in the pane or a proven outer isolation envelope governs this worker, pass it on the invocation itself with --cursor-exemption attended or --cursor-exemption envelope:<name>, which both bin/fm-spawn.sh and bin/fm-control.sh's relaunch verb accept.
  exit=1  -- neither the previous spawn nor an environment variable opens the bar


########## 6. A cursor grant on a non-cursor harness is refused, not recorded ##########

$ fm-spawn.sh ev-codex-grant <project> ... --cursor-exemption attended   # crew-harness=codex
error: --cursor-exemption applies only to a cursor launch, but this spawn resolved harness=codex; drop the flag rather than recording a cursor grant that would outlive it
  exit=1   task record written: none - no stale grant can later authorize a relaunch onto cursor


########## 7. Remote secondmate wire forwards an envelope grant only ##########

$ fm-remote-secondmate-control.sh launch sm cursor - - herdr exemption:attended
error: a remote secondmate launch cannot carry an 'attended' cursor exemption: it asserts a person at the sending pane and says nothing about a worker on this host; send exemption:envelope:<name> instead

$ fm-remote-secondmate-control.sh launch sm cursor - - herdr 'exemption:envelope:bad name'
error: unrecognized remote cursor exemption 'envelope:bad name'; a remote secondmate launch forwards only exemption:envelope:<name>, where <name> starts with a letter or digit and continues with letters, digits, '.', '_', or '-'

$ fm-remote-secondmate-control.sh launch sm cursor - - herdr exemption:envelope:routing-benchmark
error: remote home is not a seeded secondmate home
  -- the envelope grant passed the wire gate and the launch went on to ordinary home validation

$ fm-remote-secondmate-control.sh launch sm cursor - - herdr 00-0af7651916cd43dd8448eb211c80319c-b7ad6b7169203331-01   # OLD parent, bare W3C carrier
error: remote home is not a seeded secondmate home
  -- a bare well-formed carrier from a not-yet-upgraded parent is still accepted as the traceparent it always was

$ fm-remote-secondmate-control.sh launch sm cursor - - herdr future-token   # unknown trailing argument
error: unrecognized remote launch argument 'future-token'; expected traceparent:<value> or exemption:<grant>
  -- self-describing wire arguments that fail closed in both directions


########## 8. Relaunch: an attended grant is never inherited, and the refusal lands BEFORE the agent is stopped ##########
  state/rl-att.meta records cursor_exemption=attended; a claude-shaped agent is live in the pane.

$ fm-control.sh rl-att relaunch --note 'stalled overnight'
error: cursor is a verified adapter but is refused for an unattended ship launch: its --auto-review classifier prompts for calls it does not deem safe, the pane has no approver, and the parked pane keeps reading as busy. The bar applies however cursor was selected, INCLUDING when firstmate inherited it by detecting its own runtime, because silently substituting another tool would change which adapter runs the captain's work without saying so. If this home is running inside cursor and resolved it that way, set config/crew-harness to a verified adapter such as codex or claude, or add a crew-dispatch profile eligible for this kind; firstmate will not choose one for you. If a person is in the pane or a proven outer isolation envelope governs this worker, pass it on the invocation itself with --cursor-exemption attended or --cursor-exemption envelope:<name>, which both bin/fm-spawn.sh and bin/fm-control.sh's relaunch verb accept. Relaunching rl-att onto it would stop the running agent for a launch that must be refused.

  running agent in the pane afterwards: claude (never stopped)
  bytes typed into the pane by this relaunch: none
  durable record left untouched: cursor_exemption=attended


########## 9. Relaunch: a named isolation envelope IS inherited, so stuck-worker recovery still works ##########

$ fm-control.sh rl-env relaunch --note 'stalled overnight'
relaunched rl-env harness=cursor from=cursor model=default effort=default backend=tmux endpoint=fmses:fm-rl-env worktree=<tmp>/ctl-envelope/wt

  [replacement command typed into the pane]
  unset TRACEPARENT; env -u CLAUDECODE -u PI_CODING_AGENT -u GROK_AGENT -u FM_PI_HARNESS -u CURSOR_INVOKED_AS '<tmp>/ctl-envelope/fakebin/cursor-agent' --trust --auto-review --sandbox enabled --workspace '<tmp>/ctl-envelope/wt' "$('<repo>/bin/fm-operational-input.sh' encode launch-brief < '<tmp>/ctl-envelope/home/data/rl-env/brief.md')"
  recorded grant after relaunch: cursor_exemption=envelope:routing-benchmark


########## 10. Relaunch: a fresh --cursor-exemption on the verb is accepted ##########

$ fm-control.sh rl-fresh relaunch --note 'captain is watching' --cursor-exemption attended
relaunched rl-fresh harness=cursor from=cursor model=default effort=default backend=tmux endpoint=fmses:fm-rl-fresh worktree=<tmp>/ctl-fresh/wt

  [replacement command typed into the pane]
  unset TRACEPARENT; env -u CLAUDECODE -u PI_CODING_AGENT -u GROK_AGENT -u FM_PI_HARNESS -u CURSOR_INVOKED_AS '<tmp>/ctl-fresh/fakebin/cursor-agent' --trust --auto-review --sandbox enabled --workspace '<tmp>/ctl-fresh/wt' "$('<repo>/bin/fm-operational-input.sh' encode launch-brief < '<tmp>/ctl-fresh/home/data/rl-fresh/brief.md')"
  recorded grant after relaunch: cursor_exemption=attended


########## 11. Relaunch onto another harness drops the cursor grant instead of carrying it ##########

$ fm-control.sh rl-switch relaunch --harness codex --note 'moving off cursor'
relaunched rl-switch harness=codex from=cursor model=default effort=default backend=tmux endpoint=fmses:fm-rl-switch worktree=<tmp>/ctl-switch/wt
  recorded harness: harness=codex   recorded grant: (none - dropped with the harness switch)


########## 12. Bootstrap diagnoses an ineligible cursor crew-dispatch rule at session start ##########

$ fm-bootstrap.sh   # config/crew-dispatch.json routes a rule to cursor
  CREW_DISPATCH: config/crew-dispatch.json - cursor is ineligible for ordinary dispatch here: these rules 
  resolve the harness for crewmate and scout spawns, both of which are unattended, and cursor is refused for 
  every ordinary unattended kind. A rule naming cursor is satisfied only when the spawn itself also passes 
  --cursor-exemption attended or --cursor-exemption envelope:<name>; without one it validates clean here and is 
  refused at spawn time. Name a harness verified for unattended work, such as codex or claude, unless every 
  spawn routed by this rule will carry a grant.

  Reported as its own advisory, not on the "invalid config/crew-dispatch.json" channel: the rule still works
  for a spawn that carries a grant. Before this change the same file validated clean and only failed at spawn time.
```
</details>
<details>
<summary>Evidence: Vendor CLI flag corroboration for the codex/cursor/claude launch flags</summary>

Source: [Vendor CLI flag corroboration for the codex/cursor/claude launch flags](https://github.com/npayette84/firstmate/blob/0ead3a0fa7256ec85ec562dc20f504b49129cf4f/.no-mistakes/evidence/fm/harden-crew-launch-sandbox/vendor-cli-flag-corroboration.txt)

```text
Vendor CLI corroboration for the flags firstmate now launches with
Captured 2026-08-31T22:32:50Z on Darwin 27.0.0 (arm64)

Installed versions
  codex        codex-cli 0.150.1
  cursor-agent 2026.08.25-3e8eec8
  claude       2.1.252 (Claude Code)

== codex: the exact flag set firstmate now passes is accepted ==
$ codex -s workspace-write -a never -c sandbox_workspace_write.network_access=true --help
  exit=0  (flags parsed)
$ codex -s workspace-write -a never --not-a-real-flag --help      # negative control
  exit=2
  error: unexpected argument '--not-a-real-flag' found
  

codex --help, the three options that matter here:
    -s, --sandbox <SANDBOX_MODE>
        --dangerously-bypass-approvals-and-sandbox
    -a, --ask-for-approval <APPROVAL_POLICY>
  (the full-bypass flag still exists in the CLI; firstmate deliberately stopped using it)

== cursor-agent --help: what --auto-review, --sandbox and --force actually are ==
  -f, --force                  Force allow commands unless explicitly denied
  --yolo                       Alias for --force (Run Everything) (default:
  --auto-review                Use Auto-review (Smart Auto): a server classifier
  --sandbox <mode>             Explicitly enable or disable sandbox mode
  --trust                      Trust the current workspace without prompting

== claude: 'auto' is a real permission mode, distinct from bypassPermissions ==
$ CLAUDE_CODE_DISABLE_FAST_MODE=1 claude --permission-mode auto --help
  exit=0  (accepted)
$ claude --permission-mode not-a-mode --help                      # negative control
  exit=1
  error: option '--permission-mode <mode>' argument 'not-a-mode' is invalid. Allowed choices are acceptEdits, auto, bypassPermissions, manual, dontAsk, plan.

claude --help:
  --permission-mode <mode>              Permission mode to use for the session
  (choices: "acceptEdits", "auto",
```
</details>
<details>
<summary>Evidence: Targeted suite summaries</summary>

```text
tests/fm-remote-secondmate-trace-context.test.sh + tests/fm-remote-secondmate-lifecycle-e2e.test.sh
FM_TEST_SUMMARY total=2 failed=0 skipped_gate=0 duration_ms=459328

tests/fm-spawn-dispatch-profile.test.sh, fm-control-relaunch, fm-control, fm-bootstrap, fm-secondmate-harness, fm-backend-orca
FM_TEST_SUMMARY total=6 failed=0 skipped_gate=0 duration_ms=556810
```
</details>
- Outcome: 🔧 1 issue found → auto-fixed ✅ across 2 runs (40m39s)

## Pipeline

_... (8 earlier update rounds omitted to keep the PR body within GitHub's 65536-char limit; full history is in the run log.)_

<details>

🔧 Fix: unify pre-stop and launch refusal behind one predicate
- ⚠️ `bin/fm-control-lib.sh:261` - The new composite predicate&#39;s contract is stated as absolute - &#34;Composing every launch-admissibility rule here means a rule added for the launch path is automatically enforced pre-stop and the two cannot drift&#34; (bin/fm-control-lib.sh:250-252), echoed at bin/fm-control.sh:684-687 (&#34;fm_control_launch_refusal composes EVERY launch-admissibility rule, so this cannot ask a narrower question than the launch will&#34;) and at bin/fm-spawn.sh:1504-1509. It does not compose every launch-admissibility rule. The launch owner still refuses AFTER the pre-stop point, at bin/fm-spawn.sh:1517 (`resolve_pi_executable` -&gt; &#34;pi executable not found on PATH&#34;), bin/fm-spawn.sh:1533 (`fm_cursor_resolve_binary` -&gt; exit 1), and bin/fm-spawn.sh:1537 (&#34;Cursor model &#39;&lt;m&gt;&#39; is not available from &#39;--list-models&#39;&#34;). All three sit below the `fm_control_launch_refusal` call at :1510, and `resolve_relaunch_profile` resolves TARGET_MODEL (bin/fm-control.sh:700-710) without consulting any of them. Concrete failing sequence, verified by reading the flow: `bin/fm-control.sh t1 relaunch --harness pi --note n` on a live task on a host with no `pi` on PATH -&gt; `fm_control_harness_supported pi` passes (:664), `fm_control_launch_refusal pi ship &#39;&#39;` returns 0, `do_exit` stops the agent (:869), `fm-spawn.sh --relaunch` exits 1 at :1517, `do_relaunch` dies with &#34;the replacement agent for t1 could not be launched on pi&#34; and the task is left with no agent. The same shape is now reachable on cursor too, because the relaunch verb accepts the grant: `relaunch --harness cursor --cursor-exemption attended --model &lt;id-not-in-catalog&gt; --note n` strands the task at :1537. These three refusals predate this branch, so this is not a regression in behavior - it is the newly added claim that is wrong, and that claim is exactly what would stop the next reader from looking. Minimum honest fix: narrow the statement at all three sites to say it composes every HARNESS/KIND/GRANT admissibility rule and name the executable-resolution and model-catalog refusals as still living past the pre-stop boundary. Closing it for real is cheap too, since `fm_cursor_resolve_binary`, `resolve_pi_executable`, and the catalog probe are all side-effect-free pre-checks that could be asked from the same shared question.
- ℹ️ `bin/fm-spawn.sh:542` - `spawn_remote_secondmate` still asks the two constituent predicates in sequence - `fm_control_cursor_exemption_applies` at bin/fm-spawn.sh:542 and `fm_control_harness_supports_kind` at :548 - rather than the `fm_control_launch_refusal` composite the same commit (af57eb8) introduced and declared the single owner of &#34;would a launch with this profile be refused&#34;. The rules and their order are identical today, so there is no live hole; the point is that the composite was added precisely so a route cannot ask a narrower question, and this route is already outside it. A third rule added to `fm_control_launch_refusal` would be enforced on the local spawn and on the pre-stop relaunch check and silently skipped on the remote route - the same class of miss that produced the round-1 remote bypass and the round-3 flag-keyed guard. The refactor is mechanical: capture the message with `if ! remote_refusal=$(fm_control_launch_refusal &#34;$harness&#34; secondmate &#34;$CURSOR_EXEMPTION&#34;)`, release the two locks, print `error: $remote_refusal`, return 1, and keep the remote-only `attended` refusal at :557 after it (ordering is unchanged, since the composite reports grant-then-kind exactly as these two do).
- ℹ️ `tests/fm-control-relaunch.test.sh:888` - The round-6 and af57eb8 insertions left three unrelated rationale blocks concatenated with no separation immediately above `assert_relaunch_refused_without_stopping` (tests/fm-control-relaunch.test.sh:888-905), while the two tests those blocks actually describe now sit further down with no rationale: the &#34;An envelope grant is inherited across relaunch, but it exempts a worker from CURSOR&#39;s unattended bar...&#34; block belongs to `test_relaunch_onto_another_harness_drops_the_cursor_grant` and the &#34;The refusal an attended cursor task hits names --cursor-exemption as the way forward...&#34; block belongs to `test_relaunch_accepts_a_fresh_attended_grant_on_the_verb`. Separately, the comment above `test_control_relaunch_refuses_an_attended_cursor_grant_before_stopping` (:848-854) still asserts &#34;`relaunch` forwards no --cursor-exemption&#34;, which stopped being true when the verb gained the flag (bin/fm-control.sh:233 parses it, :870 forwards it). The test body is still correct - it passes no flag - but its stated premise is not, and in a file whose comments are the reviewed record of why each door is closed, a stale premise is the thing a future reader trusts. Move each block back above the test it describes and reword that premise to &#34;this invocation passes no --cursor-exemption, so the launch inherits only what survives an unattended relaunch&#34;.

🔧 Fix: bound pre-stop invariant claim, route remote through composite
- ⚠️ `bin/fm-control-lib.sh:233` - The cursor refusal is parameterized by kind but its remedies are crewmate-only, so for kind=secondmate - the case this change calls the worst one - it names two knobs that cannot fix the refusal. The message says &#34;set config/crew-harness to a verified adapter such as codex or claude, or add a crew-dispatch profile eligible for this kind&#34;. For a secondmate, bin/fm-spawn.sh:1441 resolves the harness through `fm-harness.sh secondmate`, whose chain is config/secondmate-harness -&gt; config/crew-harness -&gt; own (bin/fm-harness.sh:159-168), so a home holding `config/secondmate-harness=cursor` keeps resolving cursor no matter what crew-harness says; and the crew-dispatch backstop at bin/fm-spawn.sh:1447 is on the non-secondmate branch only, so a dispatch profile is never consulted for a secondmate at all. Concrete path: `printf &#39;cursor\n&#39; &gt; config/secondmate-harness` then `bin/fm-spawn.sh sm &lt;home&gt; --secondmate` (exactly the fixture in tests/fm-secondmate-harness.test.sh:625 and, for the remote route, tests/fm-remote-secondmate-trace-context.test.sh:315) prints this message; an operator who follows it edits config/crew-harness and gets the identical refusal again. docs/configuration.md:299 already states the rule correctly (&#34;Set config/crew-harness or config/secondmate-harness&#34;), so the code message is narrower than its own guidance. Since the function already takes &lt;kind&gt;, name config/secondmate-harness for the secondmate case and drop the dispatch-profile clause there. Neither of this change&#39;s two new secondmate refusal tests asserts the remedy text, so nothing catches it.
- ℹ️ `bin/fm-control-lib.sh:260` - The newly written bound on fm_control_launch_refusal is right about the CLASS but its parenthetical reads as an exhaustive list and is not. bin/fm-control-lib.sh:259-261 says bin/fm-spawn.sh &#34;still refuses after this point when the harness executable is absent from PATH (pi, cursor) and when a requested cursor model is missing from the live `--list-models` catalog&#34;, and bin/fm-control.sh:691 narrows it further to &#34;a missing pi or cursor executable&#34;. Two more environmental refusals live below the composite call at bin/fm-spawn.sh:1512 and can strand a relaunch the same way: `resolve_muse_binary` (bin/fm-spawn.sh:1716 -&gt; &#34;muse executable not found on PATH&#34;) and `resolve_kimi_binary` (bin/fm-spawn.sh:1736 -&gt; &#34;kimi executable not found&#34;), plus `muse_credential_present` (bin/fm-spawn.sh:1720), which refuses a muse launch on a host with no worker-reachable credential. So `bin/fm-control.sh &lt;id&gt; relaunch --harness muse --note n` on a secondmate-incapable... on a crewmate task on a host without muse installed stops the agent and is then refused, which the stated bound does not cover. docs/agent-control.md:95 got this right by staying generic (&#34;a harness whose executable is missing from this machine&#34;); the two code sites should either match that generality or name muse and kimi too. Flagging because the previous round&#39;s whole point was that an inexact statement is what stops the next reader from looking.
- ℹ️ `bin/fm-spawn.sh:544` - On the remote secondmate route the shared refusal offers `--cursor-exemption attended`, which that same route refuses twelve lines later. bin/fm-spawn.sh:544 asks the composite, which for an unexempted cursor secondmate prints the shared text ending &#34;pass it on the invocation itself with --cursor-exemption attended or --cursor-exemption envelope:&lt;name&gt;&#34; (bin/fm-control-lib.sh:233); an operator who takes the first option hits bin/fm-spawn.sh:556, &#34;an &#39;attended&#39; cursor exemption asserts a person at this pane and cannot authorize a worker on a remote host&#34;. The second message is clear and points at the right fix, so this is a two-step rather than a dead end - milder than the round-5 case where the named flag did not exist on the verb at all. Worth noting because the remote route is the only place where half the advertised opt-in is invalid, and the route already knows it is remote before it prints: refusing `attended` before asking the composite, or having the remote route add one clause naming envelope-only, would keep the first message true where it is read.

🔧 Fix: name per-kind cursor remedy, generalize environmental bound
- ⚠️ `bin/fm-spawn.sh:552` - The remote-route qualifier added this round is appended to EVERY refusal the composite can return, not only the cursor one it was written for. `spawn_remote_secondmate` prints `&#34;error: $remote_launch_refusal On this REMOTE secondmate route only --cursor-exemption envelope:&lt;name&gt; is accepted, because &#39;attended&#39; asserts a person at this pane...&#34;` unconditionally, but `fm_control_launch_refusal` reaches that branch for two non-cursor reasons as well, because the route now admits every harness `fm_control_harness_supported` accepts (bin/fm-spawn.sh:518, widened from the former literal list that excluded muse). Reproduced against the installed lib: (1) `config/secondmate-harness=muse` on a remote secondmate prints &#34;muse is a verified crewmate/scout adapter only and cannot run a secondmate; it has no primary supervision protocol. Select a harness verified for secondmates. On this REMOTE secondmate route only --cursor-exemption envelope:&lt;name&gt; is accepted...&#34; - cursor-grant guidance stapled to a refusal that has nothing to do with cursor; (2) a grant on a non-cursor harness prints &#34;--cursor-exemption applies only to a cursor launch... drop the flag&#34; immediately followed by &#34;only --cursor-exemption envelope:&lt;name&gt; is accepted&#34;, which tells the operator to drop the flag and to pass a flag in the same breath. Gate the qualifier on the refusal actually being cursor&#39;s (for example append it only when `fm_control_cursor_exemption_applies &#34;$harness&#34;`), so the shared text is qualified where the qualification is true and left alone otherwise.
- ⚠️ `bin/fm-control-lib.sh:241` - The per-kind secondmate remedy added this round states as an absolute something that is false in exactly the case its own preamble introduces. The message reads &#34;If this home is running inside cursor and resolved it that way, set config/secondmate-harness to a verified adapter such as codex or claude; it resolves ahead of config/crew-harness, so changing crew-harness alone will not clear this&#34;. &#34;Resolved it that way&#34; is `detect_own`, which is reached only when config/secondmate-harness is absent or `default` AND config/crew-harness is absent or `default` (bin/fm-harness.sh:158-168). In that state config/crew-harness DOES clear the refusal, because resolve_secondmate defers to resolve_crew. Verified in a scratch home: with secondmate-harness unset and crew-harness=codex, `fm-harness.sh secondmate` prints codex; only once secondmate-harness itself holds cursor does crew-harness stop mattering. So the operator most likely to read this line - the unconfigured cursor-hosted home - is told the fleet-wide knob is useless when it is the simpler correct fix, and the claim contradicts docs/configuration.md:299 (&#34;Set config/crew-harness or config/secondmate-harness&#34;), AGENTS.md, and the intent&#39;s own wording that an unconfigured cursor-hosted home is stopped with an instruction to set config/crew-harness. The new test at tests/fm-secondmate-harness.test.sh:648 cannot catch it because its fixture writes config/secondmate-harness=cursor, the one case where the claim is true. Make the clause conditional: name config/secondmate-harness as the knob that always works, and say crew-harness clears it only while secondmate-harness is unset.
- ℹ️ `bin/fm-remote-secondmate-control.sh:176` - The wire receiver forwards whatever grant arrives to the remote host&#39;s own `fm-spawn.sh --cursor-exemption`, including `attended`, which that host will honor as a person-in-the-pane attestation for a worker nobody is watching. `cmd_launch`&#39;s loop accepts any `exemption:&lt;grant&gt;` and passes it through at :221; the &#34;only an envelope grant is forwardable&#34; rule is enforced solely on the sending side (bin/fm-spawn.sh:556 and the `envelope:?*` case at :674). No shipped caller sends `exemption:attended` today - the parent refuses it before composing launch_args - so this is not currently reachable, which is why it is info. It is worth closing anyway for the same reason this branch moved every other guard behind one owner: the intent states &#34;Remote secondmate launches forward only an envelope grant&#34;, and the boundary that consumes the grant is the one place that rule is not checked. One `case &#34;$exemption&#34; in attended) die ...` beside the existing prefix parse would make the receiver refuse what the sender already refuses to send.

🔧 Fix: condition remote qualifier and secondmate remedy, validate wire grant
- ⚠️ `tests/fm-control-relaunch.test.sh:874` - Two new tests require a REAL installed Cursor CLI and go red on any host without one, unlike every sibling suite that exercises a successful cursor launch. `test_relaunch_inherits_a_named_isolation_envelope_grant` (tests/fm-control-relaunch.test.sh:874) asserts `assert_contains &#34;$out&#34; &#34;spawned rl41 harness=cursor&#34;`, and `test_relaunch_accepts_a_fresh_attended_grant_on_the_verb` (:945) asserts `expect_code 0` plus `relaunched rl44 harness=cursor`. Both drive a real `fm-spawn.sh --relaunch --harness cursor`, which reaches `CURSOR_BIN=$(fm_cursor_resolve_binary) || exit 1` at bin/fm-spawn.sh:1533 - an ENVIRONMENTAL check that lives past the composite and is not stubbed here. The suite&#39;s only fakebin is `make_tmux_stub` (tests/fm-control-relaunch.test.sh:52), which creates `tmux` and `sleep` and no `cursor-agent`, and `run_spawn`/`run_control` prepend it to the inherited `$PATH` (:181, :166), so resolution falls through to the developer&#39;s own install. Verified directly: `env -i HOME=/tmp/nonexistent PATH=/usr/bin:/bin bash -c &#39;. bin/fm-cursor-lib.sh; fm_cursor_resolve_binary&#39;` returns non-zero, while on this machine it resolves `~/.local/bin/cursor-agent` (a real 2026.08.25 install), which is why the pair passes here. The repo already has both fixes for exactly this: tests/fm-spawn-dispatch-profile.test.sh:41-46 writes a fake `cursor-agent` into its fakebin (structural resolution accepts it on basename alone), and tests/fm-secondmate-harness.test.sh:606 prints `skip: cursor executable not resolvable in this environment` instead of failing. The other five new cursor relaunch cases are unaffected because they all refuse at the composite (bin/fm-spawn.sh:1526) before binary resolution. Add the same `cursor-agent` stub to this suite&#39;s fakebin so the two positive cases prove the grant path rather than the developer&#39;s install.
- ℹ️ `bin/fm-bootstrap.sh:1158` - The new crew-dispatch diagnostic states as an absolute something this same change makes false. It reports the whole file as `invalid config/crew-dispatch.json` with &#34;A rule naming cursor can never be satisfied, so it would validate clean here and fail at spawn time.&#34; A dispatch-resolved cursor rule CAN be satisfied: with `config/crew-dispatch.json` present, bin/fm-spawn.sh:1025 requires an explicit `--harness`, and an explicit `--harness cursor` plus `--cursor-exemption attended` (or `envelope:&lt;name&gt;`) launches cleanly - that is exactly the path tests/fm-spawn-dispatch-profile.test.sh&#39;s `test_cursor_exemption_permits_an_attended_or_enveloped_spawn` proves, and .agents/skills/harness-adapters/references/harness/cursor.md:70 and AGENTS.md:198-199 both describe cursor as remaining available for precisely those two cases. So a captain who registers a cursor rule for attended or enveloped crewmate work is told at every bootstrap that their configuration is invalid and can never work. The diagnostic itself is authorized by the intent (&#34;bootstrap now diagnoses an ineligible cursor crew-dispatch rule&#34;); only the absolute clause is wrong. This is the same overstated-claim shape the last four rounds have been correcting elsewhere - condition it on the case it is printed for (a rule naming cursor cannot serve an ORDINARY unattended dispatch and is satisfiable only when the spawn also carries --cursor-exemption) rather than asserting a universal the code does not enforce.
- ℹ️ `bin/fm-remote-secondmate-control.sh:197` - The receiver&#39;s new grant check is a third, looser copy of a rule this branch has otherwise consolidated behind one owner, and its comment claims more than it does. bin/fm-remote-secondmate-control.sh:193-199 matches `envelope:?*` by hand while `fm_control_cursor_exemption_valid` (bin/fm-control-lib.sh:148) is declared &#34;the ONE owner of what a cursor exemption grant may say&#34; and additionally bounds the envelope NAME to a locale-pinned `[A-Za-z0-9][A-Za-z0-9._-]*`. Verified against the installed lib: `envelope:bad name`, `envelope:-x`, and a newline-bearing `envelope:x\nyolo=on` are all rejected by the owner but accepted by `envelope:?*`, so they pass this boundary and are forwarded to the remote `fm-spawn.sh --cursor-exemption`, which then refuses them. Behavior is fail-closed and no shipped sender can compose those, so this is latent drift rather than a hole - but the comment directly above says this boundary &#34;validates what it accepts rather than trusting the sender&#34;, which is only true of the attended/envelope split and not of the charset the owner enforces. bin/fm-control-lib.sh is explicitly a pure sourceable contract with no side effects (its header, lines 15-17), and this script already sources four libs at :45-52, so `. &#34;$SCRIPT_DIR/fm-control-lib.sh&#34;` and calling the owner would make the receiver&#39;s check identical to the rule it names, per the standing preference for one shared check over parallel ones.

🔧 Fix: stub cursor binary, condition dispatch advisory, share grant owner
2 infos still open:
- ℹ️ `bin/fm-bootstrap.sh:1168` - The new cursor advisory returns early, so a crew-dispatch config the same change now treats as valid-but-conditional loses the dispatch facts every other valid config emits. `crew_dispatch_validate` prints `CREW_DISPATCH: config/crew-dispatch.json - &lt;advisory&gt;` and then `return 0` (bin/fm-bootstrap.sh:1166-1170), which sits above the `FM_BOOTSTRAP_VERBOSE_FACTS` block at :1176-1194 - the only producer of `BOOTSTRAP_INFO: crew dispatch active`, `... crew dispatch rule: ...`, and `... crew dispatch default: ...`. Concrete path: a home with `FM_BOOTSTRAP_VERBOSE_FACTS=1` and `{&#34;rules&#34;:[{&#34;when&#34;:&#34;attended cursor work&#34;,&#34;use&#34;:{&#34;harness&#34;:&#34;cursor&#34;}}]}` used to get its rule inventory printed (before this branch a cursor rule validated clean and fell through); it now gets the advisory and nothing else, even though the advisory&#39;s own text says the rule is satisfiable with a grant. The schema-failure branch below it returns early for the right reason - that file cannot work at all - but this one is precisely the case the round-10 change split out as usable. Print the advisory and fall through to the facts block rather than returning. Separately, the file&#39;s own output contract at bin/fm-bootstrap.sh:12 still lists only `&#34;CREW_DISPATCH: invalid config/crew-dispatch.json - &lt;reason&gt;&#34;`, so the non-invalid shape this change emits is undocumented in the header that enumerates every line bootstrap can print.
- ℹ️ `bin/fm-control-lib.sh:244` - The cursor refusal&#39;s remedy is gated behind an antecedent that is false in the reachable case the repo&#39;s own tests use. The message reads &#34;... The bar applies however cursor was selected, INCLUDING when firstmate inherited it by detecting its own runtime ... If this home is running inside cursor and resolved it that way, &lt;remedy&gt;; firstmate will not choose one for you.&#34; `resolved it that way` is `detect_own`, reached only when both config knobs are absent or `default` (bin/fm-harness.sh:158-168). But the ordinary way to hit this refusal is an EXPLICIT config value: `printf &#39;cursor\n&#39; &gt; config/secondmate-harness` then `bin/fm-spawn.sh sm &lt;home&gt; --secondmate` is exactly the fixture at tests/fm-secondmate-harness.test.sh:625, and `config/crew-harness=cursor` is the ship/scout equivalent. For those operators the sentence&#39;s condition is plainly false, yet the remedy that follows it is the only one that clears the refusal. This is the mirror of the shape the last several rounds corrected: instead of a universal asserted where the code branches, it is a universally-applicable remedy conditioned on one branch. Neither new secondmate test catches it, because both assert only that the remedy substring is present, not that it is reachable guidance for the case that produced it. Ungate the remedy - state it unconditionally, and keep the detection case as the separate note it already is in the preceding sentence.

</details>

<details>

- ⚠️ `bin/fm-remote-home-seed.sh:160` - tests/fm-remote-secondmate-trace-context.test.sh (new in this change) fails on this macOS machine, but not because of this change: bin/fm-remote-home-seed.sh line 160 expands &#34;${PROJECT_NAMES[@]}&#34; on an empty array under `set -u`, which stock macOS bash 3.2 rejects as unbound, so every `--no-projects` seed dies before doing anything. That file is untouched by this branch, and the pre-existing tests/fm-remote-secondmate-lifecycle-e2e.test.sh fails the same way at the same call. Reproduced standalone: `bash bin/fm-remote-home-seed.sh probe remote-mac /tmp/remote-root &lt;home&gt; --no-projects` -&gt; &#34;line 190: PROJECT_NAMES[@]: unbound variable&#34;. Linux CI (bash 5) is unaffected and the repo&#39;s macos-stock-bash lane only parses files plus two snapshot suites, so CI will stay green. Consequence for this phase: the new remote traceparent/exemption wire coverage cannot execute on stock macOS bash, so I verified that boundary manually instead (attended refused, malformed envelope refused, valid envelope accepted past the grant gate, bare W3C carrier still accepted, unknown token refused). Fixing it means a one-line change to an untouched production script (`&#34;${PROJECT_NAMES[@]:-}&#34;` plus the same guard at the count), which is outside this change&#39;s declared scope and the split-PR boundary - the captain&#39;s call whether to take it here or separately.
- <code>`bin/fm-test-run.sh tests/fm-spawn-dispatch-profile.test.sh tests/fm-control-relaunch.test.sh tests/fm-control.test.sh tests/fm-remote-secondmate-trace-context.test.sh tests/fm-secondmate-harness.test.sh tests/fm-bootstrap.test.sh tests/fm-cursor-harness.test.sh tests/fm-harness-adapter-references.test.sh tests/fm-backend-orca.test.sh` - 8 of 9 suites pass; only the new remote trace-context suite fails, on the pre-existing macOS bash 3.2 seed defect</code>
- <code>Manual end-to-end transcript driving the real `bin/fm-spawn.sh` for claude, codex and cursor ship/scout/secondmate spawns against an isolated FM_HOME and a fake tmux endpoint, capturing the literal launch command, the `spawned ...` line, and `state/&lt;id&gt;.meta`</code>
- <code>Manual `bin/fm-control.sh &lt;id&gt; relaunch` runs covering: attended record refused before the agent is stopped (pane command unchanged, nothing sent, record untouched), envelope record inherited and relaunched, fresh `--cursor-exemption attended` on the verb accepted, `--harness codex` switch dropping the grant</code>
- <code>Manual `bin/fm-remote-secondmate-control.sh launch sm cursor - - herdr &lt;arg&gt;` for `exemption:attended`, `exemption:envelope:bad name`, `exemption:envelope:routing-benchmark`, a bare W3C traceparent, and an unknown `future-token`</code>
- <code>`bin/fm-bootstrap.sh` on a home whose config/crew-dispatch.json routes a rule to cursor - emits the CREW_DISPATCH advisory, not the invalid-config channel</code>
- <code>Vendor CLI corroboration with negative controls: `codex -s workspace-write -a never -c sandbox_workspace_write.network_access=true --help`, `claude --permission-mode auto --help` vs `--permission-mode not-a-mode`, and `cursor-agent --help` flag semantics</code>
- <code>`codex sandbox -c sandbox_mode=workspace-write [-c sandbox_workspace_write.network_access=true] -- ...` egress and write probes (network comparison inconclusive - this environment has no DNS at all, host shell included)</code>
- <code>`bash bin/fm-remote-home-seed.sh probe remote-mac /tmp/remote-root &lt;home&gt; --no-projects` - standalone reproduction isolating the bash 3.2 unbound-array failure behind the one failing suite</code>
- <code>`bash tests/fm-remote-secondmate-lifecycle-e2e.test.sh` - pre-existing suite confirming the same seed failure exists independently of this change</code>

🔧 Fix: fix pre-existing bash 3.2 empty-array crash in remote seed
- <code>`bash /tmp/repro-seed.sh &lt;pre-fix-copy&gt;` - standalone `fm-remote-home-seed.sh probe remote-mac /srv/fm-root /srv/fm-home --no-projects` against bin/fm-remote-home-seed.sh at 8e6b1f9: reproduces `line 190: PROJECT_NAMES[@]: unbound variable`, exit 0 (masked by the EXIT trap)</code>
- <code>`bash /tmp/repro-seed.sh &lt;worktree&gt;` - same invocation at 5d8bcab: loop skipped, run reaches the first real remote step, honest exit 1</code>
- <code>`bin/fm-test-run.sh tests/fm-remote-secondmate-trace-context.test.sh tests/fm-remote-secondmate-lifecycle-e2e.test.sh` - 2 scripts, failed=0, skipped_gate=0 (both were unrunnable on this machine before the fix)</code>
- <code>`bin/fm-test-run.sh tests/fm-spawn-dispatch-profile.test.sh tests/fm-control-relaunch.test.sh tests/fm-control.test.sh tests/fm-bootstrap.test.sh tests/fm-secondmate-harness.test.sh tests/fm-backend-orca.test.sh` - 6 scripts, failed=0, skipped_gate=0</code>
- <code>Manual: `FM_HOME=&lt;tmp&gt; FM_ROOT_OVERRIDE=&lt;tmp&gt; bin/fm-bootstrap.sh` on a home whose config/crew-dispatch.json names cursor - emits the CREW_DISPATCH ineligibility diagnosis naming --cursor-exemption at session start rather than validating clean</code>
- <code>`git status --porcelain` - worktree clean, no transient artifacts left behind</code>

</details>

<details>

- ℹ️ `bin/fm-remote-secondmate-control.sh:141` - Judgment call, resolved as no-op: the mixed-version wire degradation for remote secondmate launches (a new parent sending `traceparent:&lt;value&gt;` to a not-yet-upgraded host breaks every carrier-bearing remote launch, including ordinary codex and claude ones) is operator-visible fleet-upgrade guidance, but it is documented only in cmd_launch&#39;s code comment. The user intent states that degradation is deliberately stated in code, and docs/remote-secondmates.md has no version-skew section to extend, so documenting it in prose would mean creating a new surface and a second copy of a code-owned contract. Left as is.
- ℹ️ `.agents/skills/harness-adapters/SKILL.md:35` - Out-of-scope consolidation worth a follow-up: the router&#39;s &#34;Non-negotiable safety&#34; section states muse&#39;s kind restriction inline (&#34;Muse is verified only for crewmate and scout work, never a secondmate or primary&#34;) but not the parallel, now-broader cursor unattended bar, which lives in references/harness/cursor.md and AGENTS.md. The two comparable kind rules are placed inconsistently. Not fixed here because adding cursor&#39;s rule would duplicate AGENTS.md and cursor.md, and removing muse&#39;s line is unrelated restructuring; a follow-up should pick one placement for both.

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>


kunchenguid:
Speaking as Kun's firstmate: triage on HEAD `07f98469664e4f66b2757e1883e51114b752118d`.

**Contract-class: new-default.** Unconfigured spawn on main today launches Claude with `--dangerously-skip-permissions`, Codex with `--dangerously-bypass-approvals-and-sandbox`, and Cursor with `--yolo` (documented in the Cursor adapter table). This PR replaces those with `--permission-mode auto`, `-s workspace-write -a never`, and `--auto-review --sandbox enabled`. That is a safer default posture, but it changes the specified unconfigured launch path, so it is **not** restore and will **not** auto-merge.

**Attestation: MISSING** (no bound no-mistakes raise / Pipeline section on this HEAD).
**workflow-zero: yes** (no `.github/workflows/*` vs main `0866a770`).
**CI:** first-time fork runs approved this pass — CI `33396826905` (in progress after approve), Require no-mistakes `33396826877` (failed: not raised via no-mistakes). Greptile FAILURE noted (non-blocking here; coverage nit on real harness acceptance). mergeable MERGEABLE/UNSTABLE.

**VISION (short):** aligns with Authority (stops assuming full bypass consent) and fleet-outlives-vendor (uses each harness's own controls); Scope/Delegation stay scripts-owned launch composition. Resist-angle: new-default autonomy flip needs an explicit captain decision once otherwise ready — do not infer consent from "safer".

**Action for author:** please re-raise this HEAD through `git push no-mistakes` so attestation MATCHES `07f98469664e4f66b2757e1883e51114b752118d`. Also keep #3385 stacked; landing this alone (as you noted) puts Codex crewmates in the denied state.

**Waiting on:** author (attestation) + CI. **No Firstmate flag yet** — new-default only escalates when green CI, safe review, and attestation MATCH leave only the captain decision. **No merge.**

kunchenguid:
Speaking as Kun's firstmate: re-triage on newer activity. HEAD is now `5f4af7d9d61cac56c3119152e25aa442abad0c61` (previous stamp was `07f98469664e`; many no-mistakes review rounds since).

**Contract-class: new-default.** Unconfigured spawn on main `6c1d2db194cb20e08232ba2fa2c414592f724b44` still launches Claude with `--dangerously-skip-permissions`, Codex with `--dangerously-bypass-approvals-and-sandbox`, and Cursor with `--yolo`. This PR replaces those with `--permission-mode auto` + `CLAUDE_CODE_DISABLE_FAST_MODE=1`, `-s workspace-write -a never` + explicit `sandbox_workspace_write.network_access=true`, and `--auto-review --sandbox enabled`, and bars Cursor from every ordinary unattended kind unless `--cursor-exemption attended|envelope:<name>` is passed. Safer, but it changes the specified unconfigured launch path (and newly refuses unattended Cursor, including runtime-detected). **Not restore. Will not auto-merge.**

**Attestation: MATCH** — body `no-mistakes-pipeline-attestation:v1` `head_sha` `5f4af7d9d61cac56c3119152e25aa442abad0c61` equals this HEAD.
**workflow-zero: yes** (no `.github/workflows/*` vs main).
**CI this pass:** first-time fork runs approved for this HEAD — CI `33449411973`, Require no-mistakes `33449411970` (synchronize) and `33449480972` (edited). All three were `action_required` and are now queued. Prior-HEAD CI `33396826905` SUCCESS is not this HEAD. mergeable MERGEABLE/UNSTABLE.

**VISION (short):** aligns with Authority (stops assuming full-bypass consent; Cursor exemption is per-invocation, charset-bounded, not env-inherited; `attended` never inherits across relaunch) and fleet-outlives-vendor (uses each harness's own controls). Scripts own the launch/admissibility table (`fm_control_launch_refusal`). Resist-angle: new-default autonomy/sandbox flip and a new unattended-Cursor refusal need an explicit captain decision once otherwise ready — do not infer consent from "safer" or from the PR body's claimed captain rulings.

**Waiting on:** CI on this HEAD. **No merge.** **No Firstmate flag yet** — new-default only escalates when green CI, green no-mistakes, MATCH attestation, and safe review leave only the captain decision. After CI is green this remains a captain-decision hold, not an auto-merge. Keep #3385 stacked; landing this alone puts Codex crewmates in the denied-roots state that #3385 restores.

viperfx:
Superseded: these docs are moving onto the LVX-4883 working branch instead of shipping separately.

viperfx:
Superseded - these documentation changes are moving onto the LVX-4883 working branch instead of shipping separately. Please close.

PR #3392, opened 2026-08-31
feat(bin): add fleet usage ledger harvester and report reader

npayette84:
## Intent

Fix the two defects exposed by the first real Harvester rows in firstmate's fleet usage ledger (bin/fm-usage-harvest.sh).

DEFECT 1 - Pi-backed rows reported null token usage. Add a Pi session-log parser for the CURRENT ~/.pi/agent/sessions format so Pi-backed rows report real token usage instead of null.

DEFECT 2 - Stop deriving task start time from mutable metadata modification time. Use the epoch embedded in the meta's spawn_gen=s<epoch>.<pid>.<random> token as the PRIMARY durable source, with a filesystem birth-time fallback only where necessary, so later metadata updates such as PR registration cannot collapse wall_secs to zero.

Accepted constraints and decisions from the task brief and the supervising firstmate:
- Load the firstmate-coding-guidelines skill before editing any shared tracked material (done). Its rules govern this change: one-owner rule for contracts, knowledge-placement decision tree, one full sentence per line in tracked Markdown, plain dash never em dash, no agent co-author, bin/*.sh shellcheck-clean, tests colocated in tests/ as <subject>.test.sh extending the existing runner.
- Preserve behavior for every supported harness and runtime; inspect their integration surfaces rather than assuming. Deliberately verified: claude and codex parser branches unchanged; opencode, grok, kimi, cursor, muse and remote-secondmate tasks continue to fall through to source=unavailable with null token fields; runtime backends are untouched because the harvester reads only task metadata plus session logs.
- Add BEHAVIORAL tests through public script interfaces. Explicitly do NOT assert implementation-source bytes.
- Update the SINGLE authoritative documentation or verification owner ONLY where current behavior changes. Deliberate decision: bin/fm-usage-harvest.sh's own header comment is that owner (it already declares itself sole owner of the ledger line schema and of every per-harness usage source), so the header carries the schema enum, the new pi source, the new wall-clock rule and the new FM_USAGE_PI_DIR seam. docs/scripts.md's one-line row stays accurate and was deliberately left unchanged; no docs/verification/ record was added because the colocated regression tests own the guarantee. fm-doc-audience-check.sh passes.
- Run shellcheck and bin/fm-lint.sh (both clean, ShellCheck 0.11.0 and actionlint 1.7.12 pinned) and ship through the repository's full validation path.

STANDING COMPLETION-EVIDENCE RULE added by the supervisor mid-task: completion may not be reported without the strongest available command or artifact evidence, with exact reproducible output, showing (a) a REAL Pi session produces non-null token usage and (b) a metadata rewrite no longer collapses wall time. Visual evidence is explicitly optional because this surface is non-visual. This evidence must be re-established against the FINAL validated head.

Deliberate implementation choices a reviewer reading only the diff would not know:
- The Pi parser binds a log to a task by the cwd recorded in the log's own "session" record, matching the existing codex branch's cwd-matching precedent, rather than by reconstructing Pi's directory-name encoding. This is deliberate: it is immune to encoding drift, whereas the claude branch must use path encoding because claude's logs carry no cwd.
- pi and pi-signed deliberately share ONE scan of ~/.pi/agent/sessions. Verified fact: pi-signed is a signed wrapper around the same Pi application over the same ~/.pi/agent state tree (the harness-adapters skill records that the plain pi command also execs that signed launcher and that trust decisions persist in the shared ~/.pi/agent/trust.json).
- Pi's model is reported provider-qualified as "<provider>/<model>" (e.g. zai/glm-5.3-flash). This is deliberate so a log-derived model matches the spelling fm-spawn already records in task metadata, keeping a row's model field stable whether or not logs were found.
- Pi usage fields are summed as disjoint counts (input + output + cacheRead verified to equal Pi's own totalTokens on real logs), so cached_input_tokens folds cacheRead + cacheWrite exactly as the claude branch folds cache_read + cache_creation.
- Records are deduped on the record's own .id only WHERE PRESENT, and records without an id are counted unconditionally. This is deliberate: unlike claude, Pi writes one record per message rather than one per content block, so the dedupe is defensive against a replayed line and must not collapse unkeyed records into one.
- A start later than the end is pinned to the end rather than inverting the window. Deliberate: the pre-existing '[ $WALL -ge 0 ] || WALL=0' guard hid a negative duration but still left the log-matching window inverted, which silently dropped every session log.

THIRD DEFECT found while verifying, fixed in the same change because it is the same root cause and the same observed symptom set: bin/fm-teardown.sh called the harvest AFTER status_retire_presentation_task, which deletes state/<id>.status. The harvest therefore lost both the task window and the turn count on every real teardown, which is why real rows showed turns:0 alongside wall_secs:0 and source:unavailable. The harvest call was moved ahead of that retirement at both call sites, which is what the call's own pre-existing comment already claimed. This is a minimal reordering of two independent, backend-agnostic steps; it was verified not to regress tests/fm-teardown.test.sh or tests/fm-teardown-endpoint-safety.test.sh.

ALSO deliberately fixed in the same pass, in the test file being edited: tests/fm-usage-harvest.test.sh used BSD-only 'date -r <epoch>' to pin fixture timestamps, which would not pin them on the Linux CI runners. Every call now routes through portable touch_stamp/iso_utc helpers that try the BSD form then the GNU form, mirroring the idiom the harvester script itself already uses.

Verification already performed on this head: all 15 tests in tests/fm-usage-harvest.test.sh pass; each NEW case was confirmed to FAIL against HEAD with the exact real-world symptoms (wall_secs:0, turns:0, source:unavailable) and pass after, so none is vacuous; a smoke run against a REAL unmodified ~/.pi/agent/sessions log produced input 24664, cached 25088, output 1084, reasoning 879, matching an independent jq recomputation over the same file exactly; tests/fm-teardown-endpoint-safety.test.sh, tests/fm-classify-decision-key.test.sh and tests/fm-wake-drain-unread-status.test.sh are all green; bin/fm-test-run.sh --check-coverage passes.

KNOWN PRE-EXISTING, NOT CAUSED BY THIS CHANGE: tests/fm-teardown.test.sh fails one case, 'herdr-preflight-missing-adapter', on this machine. It was confirmed to fail identically at HEAD with the same 12 preceding cases passing.

## What Changed

- Added `bin/fm-usage-harvest.sh`, which appends one JSON row per finished task to the gitignored `data/usage-ledger.jsonl` and owns that line schema. It sums per-request token usage from claude, codex, and pi/pi-signed session logs (every other harness records `source=unavailable` with null token fields), binds pi and codex logs to a task by the cwd in the log's own session record, and derives the task window from the status file's birth epoch and the epoch in the meta's `spawn_gen` token rather than mutable file mtimes, so a later meta write such as PR registration cannot collapse `wall_secs` to zero. `bin/fm-usage-report.sh` reads the ledger back as aligned per-model and per-task totals. `tests/fm-usage-harvest.test.sh` covers the harvest through the script interface with fixture session logs and portable timestamp helpers.
- `bin/fm-teardown.sh` now runs the best-effort harvest before `status_retire_presentation_task` at both call sites. That retirement deletes `state/<id>.status`, so the previous ordering lost the task window and the turn count on every real teardown; a harvest failure still only warns and never blocks teardown.
- `bin/fm-spawn.sh` replaces blanket permission bypasses in the crewmate launch templates with sandboxed autonomy flags: claude uses `--permission-mode auto`, codex uses `-s workspace-write -a never`, and cursor uses `--trust --auto-review --sandbox enabled`. `AGENTS.md` and `docs/scripts.md` gained the ledger file and the two new script rows.

## Risk Assessment

✅ Low: The change is well-bounded, shellcheck-clean, covered by 21 behavioral cases that exercise public interfaces only, and its two core parsers were verified here against real Pi and Claude session logs; the three pipeline fix rounds correctly resolved the TSV field-collapse, the touch/UTC skew, the GNU stat-shim concatenation and the ledger temp-file leak, leaving only one low-realism parser edge.

## Testing

I ran the colocated harvester suite (22 cases, all green) plus the two teardown suites the reordering touches, then went past unit tests to demonstrate the intent the way an operator experiences it: a real harness=pi task harvested against this machine's actual, unmodified ~/.pi/agent/sessions tree, with the meta mtime deliberately rewritten to the end of the task to reproduce the PR-registration write. The pre-change harvester, extracted verbatim from 12f6613 and run on identical fixtures, produced exactly the reported real-world symptoms (wall_secs 0, turns 0, all token fields null, source unavailable); this change produces wall_secs 36000, turns 3, input 764567, cached 11715456, output 109463, reasoning 71737 and source pi-sessions. An independent jq recomputation over the same two real logs matched all four token fields exactly, and bin/fm-usage-report.sh renders the row correctly, so the ledger and the reader agree. A spot check confirmed opencode, grok, kimi, cursor and muse still fall through to null/unavailable with the same real log tree in reach. No visual artifact applies: the surface is a JSONL ledger line and a text report, both captured as CLI transcripts. The only failing test, tests/fm-teardown.test.sh 'herdr-preflight-missing-adapter', reproduces identically in a clean checkout of the pre-change commit and is therefore not a regression from this change. The worktree is clean and all scratch copies were removed.

<details>
<summary>Evidence: Real Pi harvest: before/after transcript, independent jq cross-check, and rendered usage report</summary>

Source: [Real Pi harvest: before/after transcript, independent jq cross-check, and rendered usage report](https://github.com/npayette84/firstmate/blob/a1f13d5d6b828740c67aadb60193003695b5a2ba/.no-mistakes/evidence/fm/fm-harvester-v2/real-pi-harvest-evidence.txt)

<code>### BEFORE (harvester at 12f6613, teardown had already removed the status log)&#10;{&#34;task&#34;:&#34;before-a&#34;,&#34;harness&#34;:&#34;pi&#34;,&#34;model&#34;:&#34;zai/glm-4.6&#34;,&#34;effort&#34;:null,&#34;spawned_at&#34;:&#34;2026-08-31T14:20:00Z&#34;,&#34;completed_at&#34;:&#34;2026-08-31T14:20:00Z&#34;,&#34;wall_secs&#34;:0,&#34;turns&#34;:0,&#34;input_tokens&#34;:null,&#34;cached_input_tokens&#34;:null,&#34;output_tokens&#34;:null,&#34;reasoning_tokens&#34;:null,&#34;source&#34;:&#34;unavailable&#34;}&#10;&#10;### BEFORE (harvester at 12f6613, status log kept)&#10;{&#34;task&#34;:&#34;before-b&#34;,...,&#34;wall_secs&#34;:0,&#34;turns&#34;:3,&#34;input_tokens&#34;:null,...,&#34;source&#34;:&#34;unavailable&#34;}&#10;&#10;### AFTER (this change, status log kept)&#10;{&#34;task&#34;:&#34;after-a&#34;,&#34;harness&#34;:&#34;pi&#34;,&#34;model&#34;:&#34;zai/glm-5.3-flash&#34;,&#34;effort&#34;:null,&#34;spawned_at&#34;:&#34;2026-08-31T04:20:00Z&#34;,&#34;completed_at&#34;:&#34;2026-08-31T14:20:00Z&#34;,&#34;wall_secs&#34;:36000,&#34;turns&#34;:3,&#34;input_tokens&#34;:764567,&#34;cached_input_tokens&#34;:11715456,&#34;output_tokens&#34;:109463,&#34;reasoning_tokens&#34;:71737,&#34;source&#34;:&#34;pi-sessions&#34;}&#10;&#10;### AFTER (this change, status log removed - wall + tokens still survive on spawn_gen)&#10;{&#34;task&#34;:&#34;after-b&#34;,...,&#34;wall_secs&#34;:36000,&#34;turns&#34;:0,&#34;input_tokens&#34;:764567,&#34;cached_input_tokens&#34;:11715456,&#34;output_tokens&#34;:109463,&#34;reasoning_tokens&#34;:71737,&#34;source&#34;:&#34;pi-sessions&#34;}&#10;&#10;Independent jq recomputation over the same real logs in window whose session cwd matches:&#10;/Users/npayette/.treehouse/firstmate-e5afef/3/firstmate input=677858 cached=11216256 output=85861 reasoning=59337 records=108&#10;/Users/npayette/.treehouse/firstmate-e5afef/3/firstmate input=86709 cached=499200 output=23602 reasoning=12400 records=10&#10;&#10;Arithmetic check: harvested row == sum of the independent jq rows&#10;input_tokens harvested= 764567 independent= 764567 MATCH&#10;cached_input_tokens harvested= 11715456 independent= 11715456 MATCH&#10;output_tokens harvested= 109463 independent= 109463 MATCH&#10;reasoning_tokens harvested= 71737 independent= 71737 MATCH&#10;&#10;End-user surface: bin/fm-usage-report.sh over the harvested ledger&#10;per-model totals (source-available rows):&#10;model tasks input cached output reasoning wall_secs&#10;zai/glm-5.3-flash 1 764567 11715456 109463 71737 36000&#10;&#10;per-task rows:&#10;task harness model effort input cached output reasoning wall_secs source&#10;after-a pi zai/glm-5.3-flash - 764567 11715456 109463 71737 36000 pi-sessions</code>

```text
 Fixture: one finished harness=pi task
   worktree      : /Users/npayette/.treehouse/firstmate-e5afef/3/firstmate
   spawn_gen     : s1788150000.41234.a1b2c3   (2026-08-31T04:20:00Z)
   last status   : 1788186000                   (2026-08-31T14:20:00Z)
   true wall     : 36000s over 3 'working:' turns
   Pi log tree   : /Users/npayette/.pi/agent/sessions   (real, unmodified, read-only)
   meta mtime was rewritten to the END of the task (PR registration)

### BEFORE (harvester at 12f6613, teardown had already removed the status log)
{"task":"before-a","harness":"pi","model":"zai/glm-4.6","effort":null,"spawned_at":"2026-08-31T14:20:00Z","completed_at":"2026-08-31T14:20:00Z","wall_secs":0,"turns":0,"input_tokens":null,"cached_input_tokens":null,"output_tokens":null,"reasoning_tokens":null,"source":"unavailable"}

### BEFORE (harvester at 12f6613, status log kept)
{"task":"before-b","harness":"pi","model":"zai/glm-4.6","effort":null,"spawned_at":"2026-08-31T14:20:00Z","completed_at":"2026-08-31T14:20:00Z","wall_secs":0,"turns":3,"input_tokens":null,"cached_input_tokens":null,"output_tokens":null,"reasoning_tokens":null,"source":"unavailable"}

### AFTER (this change, status log kept - the teardown order it now runs in)
{"task":"after-a","harness":"pi","model":"zai/glm-5.3-flash","effort":null,"spawned_at":"2026-08-31T04:20:00Z","completed_at":"2026-08-31T14:20:00Z","wall_secs":36000,"turns":3,"input_tokens":764567,"cached_input_tokens":11715456,"output_tokens":109463,"reasoning_tokens":71737,"source":"pi-sessions"}

### AFTER (this change, status log removed - wall + tokens still survive on spawn_gen)
{"task":"after-b","harness":"pi","model":"zai/glm-5.3-flash","effort":null,"spawned_at":"2026-08-31T04:20:00Z","completed_at":"2026-08-31T14:20:00Z","wall_secs":36000,"turns":0,"input_tokens":764567,"cached_input_tokens":11715456,"output_tokens":109463,"reasoning_tokens":71737,"source":"pi-sessions"}

 Independent recomputation, straight jq over the same real Pi logs
 in [1788150000,1788186000] whose session record's cwd is the task worktree
/Users/npayette/.treehouse/firstmate-e5afef/3/firstmate	input=677858 cached=11216256 output=85861 reasoning=59337 records=108
/Users/npayette/.treehouse/firstmate-e5afef/3/firstmate	input=86709 cached=499200 output=23602 reasoning=12400 records=10

 Arithmetic check: harvested row == sum of the independent jq rows
input_tokens           harvested=    764567  independent=    764567  MATCH
cached_input_tokens    harvested=  11715456  independent=  11715456  MATCH
output_tokens          harvested=    109463  independent=    109463  MATCH
reasoning_tokens       harvested=     71737  independent=     71737  MATCH

 End-user surface: bin/fm-usage-report.sh over the harvested ledger
usage ledger: /var/folders/70/p814fs691f103nxddhx58mhh0000gn/T//fm-usage-evidence.5DEROd/home-after-a/data/usage-ledger.jsonl (1 rows)

per-model totals (source-available rows):
model                     tasks        input       cached       output    reasoning  wall_secs
zai/glm-5.3-flash             1       764567     11715456       109463        71737      36000

per-task rows:
task                     harness  model                effort          input       cached       output    reasoning  wall_secs source          
after-a                  pi       zai/glm-5.3-flash    -              764567     11715456       109463        71737      36000 pi-sessions     
```
</details>
<details>
<summary>Evidence: Reproducible evidence script (real Pi tree, before/after, cross-check, report)</summary>

Source: [Reproducible evidence script (real Pi tree, before/after, cross-check, report)](https://github.com/npayette84/firstmate/blob/a1f13d5d6b828740c67aadb60193003695b5a2ba/.no-mistakes/evidence/fm/fm-harvester-v2/real-pi-harvest-evidence.sh)

```text
#!/usr/bin/env bash
# End-to-end evidence for the two harvester defects, run against the REAL,
# unmodified ~/.pi/agent/sessions tree on this machine. Nothing under ~/.pi is
# written; it is only read.
set -eu
WT_REPO=${WT_REPO:?set WT_REPO to the worktree checkout}
NEW_BIN="$WT_REPO/bin"
PI_DIR="$HOME/.pi/agent/sessions"
TASK_CWD=/Users/npayette/.treehouse/firstmate-e5afef/3/firstmate
START=1788150000        # task spawn epoch, as recorded in meta spawn_gen
END=1788186000          # last status append

SCRATCH=$(mktemp -d "${TMPDIR:-/tmp}/fm-usage-evidence.XXXXXX")
trap 'rm -rf "$SCRATCH"' EXIT

# The pre-fix harvester, taken verbatim from the commit before this change.
OLD_BIN="$SCRATCH/oldbin"
mkdir -p "$OLD_BIN"
cp "$NEW_BIN"/*.sh "$OLD_BIN"/
git -C "$WT_REPO" show 12f6613:bin/fm-usage-harvest.sh > "$OLD_BIN/fm-usage-harvest.sh"
chmod +x "$OLD_BIN/fm-usage-harvest.sh"

stamp() { date -r "$1" +%Y%m%d%H%M.%S 2>/dev/null || date -d "@$1" +%Y%m%d%H%M.%S; }

# make_home <id> <keep-status> : a firstmate home holding one finished pi task
# whose worktree is a real Pi session cwd. The meta's mtime is pushed to the
# END of the task, exactly as a late PR registration write does.
make_home() {
  local id=$1
  local keep_status=$2
  local home="$SCRATCH/home-$id"
  mkdir -p "$home/state" "$home/data"
  cat > "$home/state/$id.meta" <<META
window=firstmate:fm-$id
endpoint_task_id=$id
worktree=$TASK_CWD
harness=pi
kind=ship
mode=no-mistakes
model=zai/glm-4.6
effort=default
spawn_gen=s$START.41234.a1b2c3
META
  printf 'working: started\nworking: halfway\nworking: wrapping up\ndone: finished\n' \
    > "$home/state/$id.status"
  touch -t "$(stamp "$END")" "$home/state/$id.status"
  [ "$keep_status" = keep ] || rm -f "$home/state/$id.status"
  # The late metadata write (PR registration) lands at the end of the task.
  touch -t "$(stamp "$END")" "$home/state/$id.meta"
  printf '%s' "$home"
}

run_harvest() {  # <bin> <home> <id>
  FM_STATE_OVERRIDE="$2/state" FM_DATA_OVERRIDE="$2/data" \
  FM_USAGE_PI_DIR="$PI_DIR" \
    "$1/fm-usage-harvest.sh" "$3" || echo "(harvest exited $?)"
  cat "$2/data/usage-ledger.jsonl" 2>/dev/null || echo "(no ledger row)"
}

echo "=================================================================="
echo " Fixture: one finished harness=pi task"
echo "   worktree      : $TASK_CWD"
echo "   spawn_gen     : s$START.41234.a1b2c3   ($(date -r $START -u +%FT%TZ))"
echo "   last status   : $END                   ($(date -r $END -u +%FT%TZ))"
echo "   true wall     : $((END - START))s over 3 'working:' turns"
echo "   Pi log tree   : $PI_DIR   (real, unmodified, read-only)"
echo "   meta mtime was rewritten to the END of the task (PR registration)"
echo "=================================================================="
echo
echo "### BEFORE (harvester at 12f6613, teardown had already removed the status log)"
run_harvest "$OLD_BIN" "$(make_home before-a nostatus)" before-a
echo
echo "### BEFORE (harvester at 12f6613, status log kept)"
run_harvest "$OLD_BIN" "$(make_home before-b keep)" before-b
echo
echo "### AFTER (this change, status log kept - the teardown order it now runs in)"
run_harvest "$NEW_BIN" "$(make_home after-a keep)" after-a
echo
echo "### AFTER (this change, status log removed - wall + tokens still survive on spawn_gen)"
run_harvest "$NEW_BIN" "$(make_home after-b nostatus)" after-b
echo
echo "=================================================================="
echo " Independent recomputation, straight jq over the same real Pi logs"
echo " in [$START,$END] whose session record's cwd is the task worktree"
echo "=================================================================="
for f in $(find "$PI_DIR" -type f -name '*.jsonl'); do
  m=$(stat -f %m "$f" 2>/dev/null || stat -c %Y "$f")
  [ "$m" -ge "$START" ] && [ "$m" -le "$END" ] || continue
  jq -rs --arg wt "$TASK_CWD" '
    if (map(select(.type=="session"))|.[0].cwd) == $wt then
      [.[] | select(.type=="message" and .message.role=="assistant" and .message.usage != null)]
      | "\($wt)\tinput=\(map(.message.usage.input//0)|add) cached=\(map((.message.usage.cacheRead//0)+(.message.usage.cacheWrite//0))|add) output=\(map(.message.usage.output//0)|add) reasoning=\(map(.message.usage.reasoning//0)|add) records=\(length)"
    else empty end' "$f"
done

echo
echo "=================================================================="
echo " Arithmetic check: harvested row == sum of the independent jq rows"
echo "=================================================================="
python3 - "$SCRATCH/home-after-a/data/usage-ledger.jsonl" <<'PYCHK'
import json, sys
row = json.loads(open(sys.argv[1]).read().strip())
expect = {"input_tokens": 677858 + 86709, "cached_input_tokens": 11216256 + 499200,
          "output_tokens": 85861 + 23602, "reasoning_tokens": 59337 + 12400}
for k, v in expect.items():
    print(f"{k:22} harvested={row[k]:>10}  independent={v:>10}  {'MATCH' if row[k]==v else 'MISMATCH'}")
PYCHK

echo
echo "=================================================================="
echo " End-user surface: bin/fm-usage-report.sh over the harvested ledger"
echo "=================================================================="
FM_DATA_OVERRIDE="$SCRATCH/home-after-a/data" "$NEW_BIN/fm-usage-report.sh"
```
</details>
<details>
<summary>Evidence: Untouched harnesses still report unavailable with the real Pi tree in reach</summary>

Source: [Untouched harnesses still report unavailable with the real Pi tree in reach](https://github.com/npayette84/firstmate/blob/a1f13d5d6b828740c67aadb60193003695b5a2ba/.no-mistakes/evidence/fm/fm-harvester-v2/other-harnesses-unchanged.txt)

<code>{&#34;harness&#34;:&#34;opencode&#34;,&#34;input_tokens&#34;:null,&#34;cached_input_tokens&#34;:null,&#34;output_tokens&#34;:null,&#34;reasoning_tokens&#34;:null,&#34;source&#34;:&#34;unavailable&#34;}&#10;{&#34;harness&#34;:&#34;grok&#34;,&#34;input_tokens&#34;:null,&#34;cached_input_tokens&#34;:null,&#34;output_tokens&#34;:null,&#34;reasoning_tokens&#34;:null,&#34;source&#34;:&#34;unavailable&#34;}&#10;{&#34;harness&#34;:&#34;kimi&#34;,&#34;input_tokens&#34;:null,&#34;cached_input_tokens&#34;:null,&#34;output_tokens&#34;:null,&#34;reasoning_tokens&#34;:null,&#34;source&#34;:&#34;unavailable&#34;}&#10;{&#34;harness&#34;:&#34;cursor&#34;,&#34;input_tokens&#34;:null,&#34;cached_input_tokens&#34;:null,&#34;output_tokens&#34;:null,&#34;reasoning_tokens&#34;:null,&#34;source&#34;:&#34;unavailable&#34;}&#10;{&#34;harness&#34;:&#34;muse&#34;,&#34;input_tokens&#34;:null,&#34;cached_input_tokens&#34;:null,&#34;output_tokens&#34;:null,&#34;reasoning_tokens&#34;:null,&#34;source&#34;:&#34;unavailable&#34;}&#10;{&#34;harness&#34;:&#34;pi&#34;,&#34;input_tokens&#34;:764567,&#34;cached_input_tokens&#34;:11715456,&#34;output_tokens&#34;:109463,&#34;reasoning_tokens&#34;:71737,&#34;source&#34;:&#34;pi-sessions&#34;}</code>

```text
{"harness":"opencode","input_tokens":null,"cached_input_tokens":null,"output_tokens":null,"reasoning_tokens":null,"source":"unavailable"}
{"harness":"grok","input_tokens":null,"cached_input_tokens":null,"output_tokens":null,"reasoning_tokens":null,"source":"unavailable"}
{"harness":"kimi","input_tokens":null,"cached_input_tokens":null,"output_tokens":null,"reasoning_tokens":null,"source":"unavailable"}
{"harness":"cursor","input_tokens":null,"cached_input_tokens":null,"output_tokens":null,"reasoning_tokens":null,"source":"unavailable"}
{"harness":"muse","input_tokens":null,"cached_input_tokens":null,"output_tokens":null,"reasoning_tokens":null,"source":"unavailable"}
{"harness":"pi","input_tokens":764567,"cached_input_tokens":11715456,"output_tokens":109463,"reasoning_tokens":71737,"source":"pi-sessions"}
```
</details>
<details>
<summary>Evidence: Untouched-harness spot-check script</summary>

Source: [Untouched-harness spot-check script](https://github.com/npayette84/firstmate/blob/a1f13d5d6b828740c67aadb60193003695b5a2ba/.no-mistakes/evidence/fm/fm-harvester-v2/other-harnesses-unchanged.sh)

```text
#!/usr/bin/env bash
# Spot-check that the harnesses this change did NOT touch still fall through to
# source=unavailable with null token fields, even with the real Pi log tree in
# reach and a worktree that a real Pi session recorded.
set -eu
WT_REPO=${WT_REPO:?}
S=$(mktemp -d "${TMPDIR:-/tmp}/fm-other-harness.XXXXXX"); trap 'rm -rf "$S"' EXIT
CWD=/Users/npayette/.treehouse/firstmate-e5afef/3/firstmate
stamp() { date -r "$1" +%Y%m%d%H%M.%S 2>/dev/null || date -d "@$1" +%Y%m%d%H%M.%S; }
for h in opencode grok kimi cursor muse pi; do
  mkdir -p "$S/$h/state" "$S/$h/data"
  printf 'worktree=%s\nharness=%s\nmodel=default\neffort=default\nspawn_gen=s1788150000.1.a\n' \
    "$CWD" "$h" > "$S/$h/state/t-$h.meta"
  printf 'working: a\ndone: b\n' > "$S/$h/state/t-$h.status"
  touch -t "$(stamp 1788186000)" "$S/$h/state/t-$h.status" "$S/$h/state/t-$h.meta"
  FM_STATE_OVERRIDE="$S/$h/state" FM_DATA_OVERRIDE="$S/$h/data" \
    FM_USAGE_PI_DIR="$HOME/.pi/agent/sessions" \
    "$WT_REPO/bin/fm-usage-harvest.sh" "t-$h"
  jq -c '{harness,input_tokens,cached_input_tokens,output_tokens,reasoning_tokens,source}' \
    "$S/$h/data/usage-ledger.jsonl"
done
```
</details>
<details>
<summary>Evidence: tests/fm-usage-harvest.test.sh transcript (22 cases, all pass)</summary>

Source: [tests/fm-usage-harvest.test.sh transcript (22 cases, all pass)](https://github.com/npayette84/firstmate/blob/a1f13d5d6b828740c67aadb60193003695b5a2ba/.no-mistakes/evidence/fm/fm-harvester-v2/fm-usage-harvest-tests.txt)

```text
ok - claude harvest: per-request dedupe, cache folding, encoding, window, idempotency
ok - claude harvest: window survives a host without usable birth time
ok - codex harvest: cwd/window matching, delta sums, model capture
ok - pi harvest: cwd/window matching, per-request sums, replay dedupe
ok - pi-signed harvest: cwd/window matching, per-request sums, replay dedupe
ok - usage harvest: spawn_gen is the durable start, immune to later meta writes
ok - usage harvest: a malformed spawn_gen falls back to file timestamps
ok - usage harvest: a start later than the end is pinned, not inverted
ok - usage harvest: a relaunch never narrows the row below the turn count's span
ok - usage harvest: the birthless relaunch narrowing matches the stated contract
ok - usage harvest: a corrupt log line costs that line, not the file
ok - usage harvest: a valid non-object log line is skipped, not fatal
ok - usage harvest: an absent log model never shifts the token fields
ok - usage harvest: a source is named only after a log yields assistant usage
ok - cursor harvest: unavailable source renders null token fields
ok - remote secondmate harvest: local logs are never misattributed
ok - usage harvest: concurrent harvests append exactly one row
ok - usage harvest: a held ledger lock bounds the acquire, no hang or dup
ok - usage report: per-model totals, per-task rows, missing-ledger tolerance
ok - usage report: an absent harness renders as a placeholder without shifting columns
ok - teardown integration: the harvest runs while the status log still exists
ok - teardown integration: harvest failure is non-fatal and leaves no work file
```
</details>
<details>
<summary>Evidence: tests/fm-teardown.test.sh at this head</summary>

Source: [tests/fm-teardown.test.sh at this head](https://github.com/npayette84/firstmate/blob/a1f13d5d6b828740c67aadb60193003695b5a2ba/.no-mistakes/evidence/fm/fm-harvester-v2/fm-teardown-tests-after.txt)

<code>ok - herdr flat teardown never erases records when pane presence is unparseable&#10;not ok - herdr-preflight-missing-adapter: teardown continued without its required preflight</code>

```text
ok - local-only worktree with HEAD on a fork remote is torn down (fix holds)
ok - teardown prompts tasks-axi backlog refresh when compatible
ok - teardown honors config/backlog-backend=manual even when tasks-axi is compatible
ok - local-only worktree with truly unpushed work is refused (safety preserved)
ok - local-only worktree with work merged into local main is torn down (no regression)
ok - no-mistakes worktree with HEAD on origin is torn down (no regression)
ok - no-mistakes worktree with genuinely unlanded work is refused (safety preserved)
ok - local-only worktree with unpushed work is torn down under --force (escape hatch)
ok - teardown completes when an exact busy-state sidecar is already absent
ok - herdr teardown removes pane-owned escalation dedupe state
ok - herdr flat teardown refuses before returning the isolated copy under lock contention and the retry completes cleanly
ok - herdr flat teardown never erases records when pane presence is unparseable
not ok - herdr-preflight-missing-adapter: teardown continued without its required preflight

[exited with code 0]
```
</details>
<details>
<summary>Evidence: tests/fm-teardown.test.sh at pre-change commit 12f6613 (same case fails)</summary>

Source: [tests/fm-teardown.test.sh at pre-change commit 12f6613 (same case fails)](https://github.com/npayette84/firstmate/blob/a1f13d5d6b828740c67aadb60193003695b5a2ba/.no-mistakes/evidence/fm/fm-harvester-v2/fm-teardown-tests-before-12f6613.txt)

<code>ok - herdr flat teardown never erases records when pane presence is unparseable&#10;not ok - herdr-preflight-missing-adapter: teardown continued without its required preflight</code>

```text
ok - local-only worktree with unpushed work is torn down under --force (escape hatch)
ok - teardown completes when an exact busy-state sidecar is already absent
ok - herdr teardown removes pane-owned escalation dedupe state
ok - herdr flat teardown refuses before returning the isolated copy under lock contention and the retry completes cleanly
ok - herdr flat teardown never erases records when pane presence is unparseable
not ok - herdr-preflight-missing-adapter: teardown continued without its required preflight
SCRATCH=/var/folders/70/p814fs691f103nxddhx58mhh0000gn/T//fm-prechange.0T17nl

[exited with code 0]
```
</details>
- Outcome: ⚠️ 1 warning across 1 run (6m44s)

## Pipeline

<details>

✅ No issues found.

</details>

<details>

- ⚠️ `.agents/skills/afk/SKILL.md` - branch carries 10 commit(s) that exist on your local main branch but were never pushed to origin/main; rebasing would bundle this unrelated work (266 file(s)) into the PR:
- 12f6613 Merge pull request #1 from npayette84/fm/fm-harvester-ship
- 3ad41e8 no-mistakes: apply CI fixes
- ad467de no-mistakes: apply CI fixes
- b61545d no-mistakes: apply CI fixes
- 6b74583 no-mistakes(document): document usage harvester scripts; refresh stale spawn launch comments
- 1a16a6d no-mistakes(review): fix claude dot-encoding and birthless harvest window collapse
- 6f3e8ec fix(bin): create the ledger data dir before the harvest write, not after
- 9a2def5 add the fleet usage harvester, report reader, and teardown harvest hook
- 96b4c98 harden the Cursor launch default to sandboxed review
- 210839c harden crewmate launch defaults to sandboxed approval

</details>

<details>

- ⚠️ `bin/fm-usage-harvest.sh:185` - epoch_to_touch&#39;s GNU fallback renders the stamp in UTC (`date -u -d &#34;@$1&#34; +%Y%m%d%H%M.%S`) but `touch -t` interprets its argument in LOCAL time, so on GNU/Linux hosts in a non-UTC timezone both window ref files are shifted by the UTC offset. Verified on this machine (TZ=EDT): `touch -t &#34;$(date -u -r 1756600000 +%Y%m%d%H%M.%S)&#34;` produced mtime 1756614400 instead of 1756600000, a 14400s skew. Because matched_files() gates candidates on `find -newer &#34;$REFDIR/start&#34; ! -newer &#34;$REFDIR/end&#34;`, a west-of-UTC host shifts both refs later and drops every session log whose mtime falls in [START, START+offset]; an east-of-UTC host shifts them earlier and drops the tail of the window. The per-file epoch filter at line 205 cannot recover a file `find` never printed, so the row silently degrades to source=unavailable with null tokens - the exact symptom this change was written to eliminate. The BSD branch (`date -r &lt;epoch&gt;`) is correct because it renders local time; only the `-u` on the fallback is wrong. Fix: drop `-u` from the fallback (or use GNU `touch -d &#34;@$epoch&#34;` directly). The `-u` in iso_from_epoch (line 147) is correct and must stay. The same defect was copied into the new test helper tests/fm-usage-harvest.test.sh:41 (touch_stamp), where it will make the new wall_secs/spawned_at assertions fail on any non-UTC Linux runner; fix both together.
- ⚠️ `bin/fm-usage-harvest.sh:324` - `IFS=$&#39;\t&#39; read -r cwd m it ct ot rt &lt;&lt;&lt;&#34;$row&#34;` cannot safely parse the jq @tsv row, because tab is an IFS *whitespace* character in bash: adjacent and leading empty fields are collapsed rather than preserved, shifting every subsequent field left. Verified: `row=$&#39;/tmp/wt\t\t100\t200\t300\t0&#39;; IFS=$&#39;\t&#39; read -r cwd m it ct ot rt` yields cwd=/tmp/wt, m=100, it=200, ct=300, ot=0, rt=(empty). jq renders a null as an empty TSV field (confirmed: `jq -rn &#39;[null,100,200,300,0]|@tsv&#39;` emits a leading tab). Concrete failure: for the pi branch, a session log whose cwd matches the worktree but whose assistant records carry usage without `.message.model` still passes the `[ &#34;$cwd&#34; = &#34;$WORKTREE&#34; ]` gate, then records MODEL_LOG as the input-token count and sums cacheRead into input_tokens, output into cached_input_tokens, reasoning into output_tokens, and 0 into reasoning_tokens. The identical defect is in the claude branch (line 243, 5 fields with .m first: an in-window *.jsonl in the encoded dir with no assistant `model` produces row `\t0\t0\t0\t0` and writes `&#34;model&#34;:&#34;0&#34;` to the ledger) and the codex branch (line 278, a rollout with session_meta but no turn_context). The three accumulate loops are byte-identical apart from the jq program and the cwd gate, so fix this once at that shared boundary rather than three times: either have jq substitute a non-empty sentinel for the nullable string fields and test for it, or read the fields without IFS-whitespace collapsing (e.g. `readarray -t -d $&#39;\t&#39;`).
- ⚠️ `bin/fm-usage-harvest.sh:69` - The header states &#34;A corrupt log line is skipped best-effort by the parser&#34;, but all three parsers use `jq -rn &#39;reduce inputs as $l (...)&#39; &#34;$f&#34; 2&gt;/dev/null || true`, which only emits after the whole file is consumed. A single malformed JSONL line aborts jq with no stdout at all, so `row` is empty and `[ -n &#34;$row&#34; ] || continue` discards the entire file&#39;s usage. Verified: `jq -rn &#39;reduce inputs as $l (0; .+($l.a // 0))&#39;` over a 3-line file with one corrupt middle line exits 5 and prints nothing, losing the two valid records. For the codex and pi branches the loss is worse than an undercount: the cwd binding lives in the same reduce, so a task whose only matching log has one truncated tail line (a realistic state for an append-only session log) leaves found=0 and the whole row degrades to source=unavailable with null tokens rather than reporting partial usage. Either make the parse per-line tolerant (e.g. `jq -Rrn &#39;reduce (inputs | try fromjson catch empty) as $l (...)&#39;`) or correct the header claim, since this file is the declared sole owner of that contract.
- ⚠️ `bin/fm-usage-harvest.sh:164` - Making spawn_gen the primary start silently rescopes a relaunched task&#39;s row to its final incarnation only, which is a behavior change the intent does not call out. bin/fm-spawn.sh:2624 recomputes SPAWN_GEN unconditionally on every spawn AND every --relaunch, and preserve_relaunch_meta (fm-spawn.sh:2636) lists spawn_gen among the keys it replaces rather than carries forward, while state/&lt;id&gt;.status is only ever appended to (no truncation on the relaunch path). Consequence for a task relaunched once: START_EPOCH is the relaunch epoch, so (a) wall_secs measures only the final incarnation, (b) matched_files() filters out the first incarnation&#39;s session logs because their mtime predates the relaunch, undercounting tokens, yet (c) TURNS still counts `^working:` lines from every incarnation. The row therefore reports turns from the whole task against tokens and wall time from one slice of it. The prior status-birth-epoch start spanned all incarnations, so this is a narrowing, not a pre-existing gap. Both readings are defensible - spawn_gen is genuinely an incarnation token - so please confirm whether the ledger row is meant to describe the task or the last incarnation, and if the former, whether preserve_relaunch_meta should carry the original spawn_gen forward (or a separate first-spawn epoch be recorded).
- ℹ️ `bin/fm-usage-harvest.sh:220` - The claude branch sets `SRC=claude-projects` on file presence alone, before any parse, while the codex (line 290) and pi (line 336) branches both gate on `[ &#34;$found&#34; -eq 1 ] || SRC=unavailable`. An in-window *.jsonl in the encoded dir that yields no assistant usage therefore produces a row claiming source=claude-projects with input_tokens 0, where the same shape under pi or codex would honestly report source=unavailable with nulls. Since this file is the sole owner of the source enum&#39;s meaning, the three harnesses disagreeing on what &#34;available&#34; asserts is worth a deliberate decision. This is also the path that lets the field-collapse defect above surface as a literal `&#34;model&#34;:&#34;0&#34;` in the ledger.
- ℹ️ `bin/fm-teardown.sh:2550` - Moving the harvest ahead of `status_retire_presentation_task &#34;$STATE&#34; &#34;$ID&#34; || exit 1` is correct and necessary (that call deletes state/&lt;id&gt;.status, which supplies both the window and the turn count), but it also means the ledger row is now committed before teardown&#39;s last abort point instead of after it. If status retirement fails and teardown exits 1, the row persists; a later successful teardown hits the idempotency guard at line 377 and keeps the first attempt&#39;s window and turns rather than re-measuring. The failure path is narrow and the two measurements would be seconds apart, so this is a noted tradeoff rather than something to change - flagging it only because the ordering constraint is now load-bearing in both directions and any future reshuffle of these steps needs to know that.

🔧 Fix: fix(usage): task-scoped window, tolerant parse, safe TSV read
- ⚠️ `bin/fm-usage-harvest.sh:22` - The new header states the row scope unconditionally: &#34;a row describes the WHOLE task, spanning every relaunch, and wall_secs, the session-log window and turns all report that one span.&#34; The code only delivers that when the filesystem reports a birth time. START is min(START_GEN, START_BIRTH) at lines 185-193; on a host where file_birth_epoch fails, START_BIRTH is empty and START collapses to START_GEN, which fm-spawn.sh rewrites to the relaunch epoch on every --relaunch. Concrete reachable path: birthless filesystem (the script itself carries a dedicated fallback for it at line 195, and the suite simulates it with nobirth_stat_bin at tests/fm-usage-harvest.test.sh:375) plus one relaunch -&gt; wall_secs and the find window cover only the final incarnation, while TURNS at line 200 still counts every &#39;^working:&#39; line in the append-only status log. That is exactly the three-way disagreement the header now says cannot happen. Second, concrete consequence: tests/fm-usage-harvest.test.sh:644 relaunch_scope_case hard-asserts wall_secs:600, spawned_at=base, source=claude-projects and input_tokens:21 with no birth-time shim, so on any runner whose tmp dir lacks statx btime (tmpfs, overlayfs, some container mounts) four assertions fail. Every other timestamp case in the suite is birth-agnostic: spawn_gen_case and teardown_status_case put spawn_gen BEFORE the birth so min() picks the token either way, and claude_case computes its expectation with `file_birth_epoch || file_mtime_epoch`. Only the relaunch case depends on the host. Recommended fix, keeping the owner&#39;s scope decision intact and fm-spawn out of scope: (a) qualify the Row scope paragraph to say the whole-task span holds wherever the status file&#39;s birth time is readable, and that on a birthless host a relaunched task&#39;s window and token sum narrow to the last incarnation while turns still span the task; (b) make relaunch_scope_case independent of the host, either by deriving the expectation from file_birth_epoch the way claude_case does, or by pinning birth availability. Note also that relaunch_scope_case takes `base` from the status file&#39;s MTIME and then asserts against the harvester&#39;s BIRTH-derived start, so even on a birth-capable host it flakes by one second whenever creation and the fixture write straddle a second boundary.
- ℹ️ `bin/fm-usage-harvest.sh:297` - `reduce (inputs | try fromjson catch empty) as $l` catches a fromjson failure but not an indexing failure on a line that parses to a valid non-object. Verified: a three-line file whose middle line is `&#34;just a string&#34;` makes jq exit 5 with no stdout (`jq: error: Cannot index string with string &#34;type&#34;`), so `row` is empty, the file is skipped entirely, and for codex/pi the cwd binding goes with it and the row degrades to source=unavailable with null tokens. Blank lines and truncated JSON are correctly tolerated (both verified to emit the surviving records), so the realistic append-only truncation case is covered; this is the residual. The header at line 86 makes the claim without qualification (&#34;A corrupt log line is skipped best-effort by the parser, which reads each line on its own and keeps the rest of that file&#39;s usage&#34;), and this file is the sole owner of that contract. One-token fix in all three programs: `try (fromjson | objects) catch empty`, which drops a non-object line instead of aborting.
- ℹ️ `bin/fm-usage-report.sh:55` - The per-task loop still reads the jq @tsv stream with `IFS=$&#39;\t&#39; read -r task harness model effort it ct ot rt wall source`, the exact pattern just removed from the harvester because tab is an IFS whitespace character and empty fields collapse. Most columns are protected by a `// &#34;-&#34;` default at lines 50-54, but `.harness` is emitted bare, and the harvester writes `&#34;harness&#34;:$harness` from an unguarded `meta_get harness` (bin/fm-usage-harvest.sh:136, no `&#34;&#34; -&gt; null` mapping unlike model and effort), so a task record missing a `harness=` line produces an empty second field and shifts model, effort and every token column one slot left in the rendered report. Impact is display-only and the input is malformed metadata, but the fix is one character: emit `(.harness // &#34;-&#34;)` with an empty-string guard, or apply the same `tr &#39;\t&#39; &#39;\037&#39;` separator the harvester now uses.

🔧 Fix: fix(usage): document scope limit, skip non-objects, guard report columns
2 issues (1 error, 1 info) still open:
- 🚨 `tests/fm-usage-harvest.test.sh:390` - Both `stat` test shims resolve the `%Y` format with `/usr/bin/stat -f %m -- &#34;$file&#34; 2&gt;/dev/null || /usr/bin/stat -c %Y -- &#34;$file&#34;`, two commands sharing one stdout. On GNU/Linux `stat -f %m -- &lt;file&gt;` treats `%m` as an operand, fails on it, but still prints the full file-system info block for &lt;file&gt; to stdout and exits 1. Verified in debian:stable-slim: the captured value is `[  File: &#34;/tmp/f&#34;\n    ID: ... Namelen: 255 Type: overlayfs\nBlock size: 4096 ...\nInodes: Total: ... Free: ...\n1788185391]`. The `||` then appends the real epoch to that same stream, so the harvester&#39;s `file_mtime_epoch` (bin/fm-usage-harvest.sh:157) rejects it as non-numeric and returns 1 for every file, including the status and meta files. `END_EPOCH=$(file_mtime_epoch &#34;$STATUS&#34; 2&gt;/dev/null || file_mtime_epoch &#34;$META&#34;)` at line 190 therefore fails under `set -eu` and fm-usage-harvest.sh exits 1 without writing a ledger row. Three cases assert `expect_code 0` and then parse the row, so all three fail on ubuntu-latest (every job in .github/workflows/ci.yml): claude_nobirth_case and relaunch_birthless_case via nobirth_stat_bin (line 390), and relaunch_scope_case via fixedbirth_stat_bin (line 412), which this fix round newly introduced by copying the same pattern. They pass on macOS only because the BSD probe succeeds there, which is why local verification did not catch it. Fix both shims the way the harvester itself already does it, with two separate assignments so the second read replaces the first&#39;s output rather than appending to it: `t=$(/usr/bin/stat -f %m -- &#34;$file&#34; 2&gt;/dev/null) || t=$(/usr/bin/stat -c %Y -- &#34;$file&#34;); printf &#39;%s\n&#39; &#34;$t&#34;`. Production code is not affected. Note this is the same class of Linux-portability defect the change already fixed deliberately for `date -r` in touch_stamp/iso_utc.
- ℹ️ `bin/fm-usage-harvest.sh:434` - The ledger append writes jq&#39;s output to `$DATA/usage-ledger.jsonl.tmp.$$` and then `cat`s that file into the ledger with `&amp;&amp; rm -f`. The whole check-and-append already runs inside the `$DATA/.usage-ledger.lock` critical section acquired at lines 406-413, so the temp file adds no atomicity, only a leak path. When `cat ... &gt;&gt; &#34;$LEDGER&#34;` fails (for instance with the ledger path unwritable, which is exactly the state tests/fm-usage-harvest.test.sh:671 constructs by making `data/usage-ledger.jsonl` a directory), the `&amp;&amp;` suppresses the `rm -f`, `set -e` exits the script, and `harvest_cleanup` at line 217 removes only REFDIR and the lock. Every failed harvest therefore leaves another `usage-ledger.jsonl.tmp.&lt;pid&gt;` in the data dir, and teardown&#39;s `|| echo warning` path means those failures are silent-but-continuing rather than loud. The same leak occurs if jq itself fails, since the exit happens before the cat. Redirecting jq straight to `&gt;&gt; &#34;$LEDGER&#34;` is behaviorally equivalent under the lock, one line shorter, and has no leak path.

🔧 Fix: fix(usage): portable stat shim, drop ledger temp file
1 info still open:
- ℹ️ `bin/fm-usage-harvest.sh:352` - The per-line tolerance guard `try (fromjson | objects) catch empty` covers parse failures and non-object lines, but not a type error raised inside the reduce update expression, so the header&#39;s claim at line 95 (&#34;A corrupt log line is skipped best-effort by the parser, which reads each line on its own and keeps the rest of that file&#39;s usage&#34;) is still not true for one shape. Verified end-to-end through the public script interface: a pi fixture whose middle line is `{&#34;type&#34;:&#34;message&#34;,&#34;id&#34;:&#34;m2&#34;,&#34;message&#34;:&#34;truncated-but-valid-json&#34;}` makes `$l.message.role` raise `Cannot index string with string &#34;role&#34;`, jq exits 5 with no stdout, the file is skipped whole, the cwd binding is lost with it, and the harvester writes `&#34;input_tokens&#34;:null … &#34;source&#34;:&#34;unavailable&#34;` instead of the surrounding 105 input tokens. The same class hits the claude branch via `.seen[$id]` when `.message.id` is a number (`Cannot index object with number`, jq exit 5) and the codex branch via `$l.payload.type` when `.payload` is a scalar. Realism is low - real Pi, Claude and Codex logs never emit these shapes, and the realistic append-only truncation case produces invalid JSON that the existing guard already tolerates correctly (confirmed) - so this is a residual edge rather than a live defect. If it is worth closing, the smallest fix is to wrap the reduce update expression in `try (&lt;body&gt;) catch .` in all three programs, which drops the offending line and keeps the accumulator; otherwise the header sentence should name the limit.

</details>

<details>
<summary>⚠️ **Test** - 1 warning</summary>

- ⚠️ `tests/fm-teardown.test.sh:1524` - tests/fm-teardown.test.sh fails the &#39;herdr-preflight-missing-adapter&#39; case on this machine. Confirmed PRE-EXISTING and not caused by this change: the same case fails identically in a clean scratch checkout of the pre-change commit 12f6613, with the same preceding cases passing. Root cause looks like a fixture issue rather than a product one - the case deletes bin/backends/herdr.sh from a copied test-root but still passes FM_ROOT_OVERRIDE=&#34;$ROOT&#34;, so fm-teardown.sh resolves the adapter from the intact repo root and never hits the missing-adapter preflight. Flagged for the author to decide, since it is outside this change&#39;s scope.
- <code>`bash tests/fm-usage-harvest.test.sh` - all 22 cases pass (claude/codex/pi/pi-signed parsing, spawn_gen window, relaunch scope, corrupt-line tolerance, cursor and remote unavailable rows, ledger lock idempotency, report rendering, teardown integration)</code>
- <code>`bash tests/fm-teardown-endpoint-safety.test.sh` - all cases pass (one skip: tmux not installed)</code>
- <code>`bash tests/fm-teardown.test.sh` - 12 pass, 1 pre-existing failure `herdr-preflight-missing-adapter`</code>
- <code>Pre-existence check: `git archive 12f6613 | tar -x -C $SCRATCH &amp;&amp; bash tests/fm-teardown.test.sh` in the scratch copy - same single case fails identically at the pre-change commit</code>
- <code>Manual end-to-end harvest against the REAL unmodified `~/.pi/agent/sessions` tree, comparing the pre-change harvester (`git show 12f6613:bin/fm-usage-harvest.sh`) with this change over identical fixtures: `real-pi-harvest-evidence.sh`</code>
- <code>Independent recomputation of the same real Pi logs with a standalone `jq -rs` program, cross-checked field by field against the harvested ledger row</code>
- <code>`bin/fm-usage-report.sh` rendered over the harvested ledger to capture the end-user reporting surface</code>
- <code>Untouched-harness spot check for opencode, grok, kimi, cursor, muse and pi against the same real Pi tree and worktree: `other-harnesses-unchanged.sh`</code>

</details>

<details>

- ℹ️ `docs/verification/runtime-backends.md:776` - docs/verification/runtime-backends.md still records the cursor probe&#39;s child argv as `--trust --yolo` and an `Autonomy | --yolo` row (lines 776, 853-854). I deliberately left these unchanged: they sit under a dated `VERIFIED ... 2026-08-11` heading and are maintainer-verification evidence of what was observed at that time, not a statement of the current launch contract, which now lives in bin/fm-spawn.sh. If the project would rather have that record re-probed against the new `--auto-review --sandbox enabled` posture, that is a follow-up verification task rather than a prose edit, since rewriting the observed output would falsify the evidence.

🔧 Fix: drop out-of-scope launch hunks, rewrap harvester header

</details>

<details>

</details>

<details>

✅ No issues found.

</details>

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -87,6 +87,7 @@ data/                personal fleet records; LOCAL, gitignored as a whole
   learnings.md       fleet-local operational facts and gotchas; LOCAL, gitignored; dated, evidence-backed, curated, and updated with inspect-then-update - rewrite and prune rather than append forever, the same contract as captain.md; created lazily, absent until this home has a learning to store
   projects.md        thin fleet navigation registry recording each project's standing delivery posture; firstmate-private, parsed for mechanical sync and seeding by fm-project-mode.sh (section 6)
   secondmates.md      local and remote secondmate routing table; firstmate-private, maintained by the secondmate seed helpers (section 6)
+  usage-ledger.jsonl append-only fleet cost ledger, one JSON line per harvested task; written only by bin/fm-usage-harvest.sh, whose header owns the line schema, alongside its .usage-ledger.lock append lock, and read by bin/fm-usage-report.sh
   <id>/brief.md      per-task crewmate brief, or per-secondmate charter brief when kind=secondmate
   <id>/report.md     scout task deliverable, written by the crewmate; survives teardown
 projects/            cloned repos; gitignored; read-only except under hard rule 1's concrete captain-approved project operation exception
--- a/bin/fm-spawn.sh
+++ b/bin/fm-spawn.sh
@@ -1251,8 +1251,7 @@ launch_template() {
     # grok (Grok Build TUI): a positional prompt starts the supervised interactive
     # session. --always-approve auto-approves every tool execution (verified: the
     # crewmate runs fully autonomously, no permission gate), which an unattended
-    # crewmate needs; it is the targeted equivalent of claude's
-    # --permission-mode auto. grok's turn-end signal does NOT ride the
+    # crewmate needs. grok's turn-end signal does NOT ride the
     # launch command - it is a Stop-event hook installed below (global hook +
     # per-task pointer), so the template is identical for ship/scout/secondmate.
     grok) printf '%s' 'grok --always-approve __MODELFLAG____EFFORTFLAG__"$(__OPINPUT__ encode launch-brief < __BRIEF__)"' ;;
--- a/bin/fm-spawn.sh
+++ b/bin/fm-spawn.sh
@@ -1231,12 +1231,12 @@ launch_template() {
     # does NOT suppress the interactive ghost text (verified empirically), so the env
     # var is the correct control. The dim-aware composer reader in fm-tmux-lib.sh is
     # the defense-in-depth backstop for any pane this flag cannot reach.
-    claude) printf '%s' 'CLAUDE_CODE_ENABLE_PROMPT_SUGGESTION=false claude --dangerously-skip-permissions __MODELFLAG____EFFORTFLAG__"$(__OPINPUT__ encode launch-brief < __BRIEF__)"' ;;
+    claude) printf '%s' 'CLAUDE_CODE_ENABLE_PROMPT_SUGGESTION=false claude --permission-mode auto __MODELFLAG____EFFORTFLAG__"$(__OPINPUT__ encode launch-brief < __BRIEF__)"' ;;
     codex)
       if [ "$kind" = secondmate ]; then
-        printf '%s' 'codex __MODELFLAG____EFFORTFLAG__--dangerously-bypass-approvals-and-sandbox "$(__OPINPUT__ encode launch-brief < __BRIEF__)"'
+        printf '%s' 'codex __MODELFLAG____EFFORTFLAG__-s workspace-write -a never "$(__OPINPUT__ encode launch-brief < __BRIEF__)"'
       else
-        printf '%s' 'codex __MODELFLAG____EFFORTFLAG__--dangerously-bypass-approvals-and-sandbox -c "notify=[\"bash\",\"-c\",\"touch __TURNEND__\"]" "$(__OPINPUT__ encode launch-brief < __BRIEF__)"'
+        printf '%s' 'codex __MODELFLAG____EFFORTFLAG__-s workspace-write -a never -c "notify=[\"bash\",\"-c\",\"touch __TURNEND__\"]" "$(__OPINPUT__ encode launch-brief < __BRIEF__)"'
       fi
       ;;
     opencode) printf '%s' 'OPENCODE_CONFIG_CONTENT='\''{"permission":{"*":"allow"}}'\'' opencode __MODELFLAG__--prompt "$(__OPINPUT__ encode launch-brief < __BRIEF__)"' ;;
--- a/bin/fm-spawn.sh
+++ b/bin/fm-spawn.sh
@@ -1268,7 +1268,7 @@ launch_template() {
     # inherited CLAUDECODE cannot outrank cursor's own marker in a process that
     # only reads the environment. Cursor exposes no effort flag, so the shared
     # effort axis is deliberately omitted and stays in task metadata only.
-    cursor) printf '%s' 'env -u CLAUDECODE -u PI_CODING_AGENT -u GROK_AGENT -u FM_PI_HARNESS -u CURSOR_INVOKED_AS __CURSORBIN__ --trust --yolo __MODELFLAG__--workspace __WORKTREE__ "$(__OPINPUT__ encode launch-brief < __BRIEF__)"' ;;
+    cursor) printf '%s' 'env -u CLAUDECODE -u PI_CODING_AGENT -u GROK_AGENT -u FM_PI_HARNESS -u CURSOR_INVOKED_AS __CURSORBIN__ --trust --auto-review --sandbox enabled __MODELFLAG__--workspace __WORKTREE__ "$(__OPINPUT__ encode launch-brief < __BRIEF__)"' ;;
     # Kimi Code rejects a positional prompt, so it launches bare and receives
     # only an absolute brief pointer after the TUI readiness gate below.
     # Its turn-end signal is a globally configured Stop hook plus a guarded
--- a/bin/fm-spawn.sh
+++ b/bin/fm-spawn.sh
@@ -1251,7 +1251,8 @@ launch_template() {
     # grok (Grok Build TUI): a positional prompt starts the supervised interactive
     # session. --always-approve auto-approves every tool execution (verified: the
     # crewmate runs fully autonomously, no permission gate), which an unattended
-    # crewmate needs. grok's turn-end signal does NOT ride the
+    # crewmate needs; it is the targeted equivalent of claude's
+    # --permission-mode auto. grok's turn-end signal does NOT ride the
     # launch command - it is a Stop-event hook installed below (global hook +
     # per-task pointer), so the template is identical for ship/scout/secondmate.
     grok) printf '%s' 'grok --always-approve __MODELFLAG____EFFORTFLAG__"$(__OPINPUT__ encode launch-brief < __BRIEF__)"' ;;
--- a/bin/fm-spawn.sh
+++ b/bin/fm-spawn.sh
@@ -1252,14 +1252,15 @@ launch_template() {
     # session. --always-approve auto-approves every tool execution (verified: the
     # crewmate runs fully autonomously, no permission gate), which an unattended
     # crewmate needs; it is the targeted equivalent of claude's
-    # --dangerously-skip-permissions. grok's turn-end signal does NOT ride the
+    # --permission-mode auto. grok's turn-end signal does NOT ride the
     # launch command - it is a Stop-event hook installed below (global hook +
     # per-task pointer), so the template is identical for ship/scout/secondmate.
     grok) printf '%s' 'grok --always-approve __MODELFLAG____EFFORTFLAG__"$(__OPINPUT__ encode launch-brief < __BRIEF__)"' ;;
-    # Cursor Agent CLI. --trust suppresses the workspace-trust prompt, which
-    # --yolo does NOT cover and which would otherwise block every spawn, since
-    # each task gets a fresh worktree path cursor has never seen. --yolo is the
-    # --force alias whose TUI label is "Run Everything". --workspace pins the
+    # Cursor Agent CLI. --trust suppresses the workspace-trust prompt that
+    # would otherwise block every spawn, since each task gets a fresh worktree
+    # path cursor has never seen, and it is also what loads the project hooks.
+    # --auto-review --sandbox enabled replaces the former blanket --yolo bypass
+    # with a sandboxed, auto-reviewing autonomy posture. --workspace pins the
     # exact worktree. -w/--worktree is deliberately never passed: it allocates a
     # SECOND worktree under ~/.cursor/worktrees and would break firstmate's
     # isolation contract. The binary is resolved rather than named because
--- a/bin/fm-teardown.sh
+++ b/bin/fm-teardown.sh
@@ -683,6 +683,10 @@ remote_secondmate_teardown() {
   grep -vE "^- $ID( |$)" "$SECONDMATE_REG" > "$tmp" || true
   mv -f -- "$tmp" "$SECONDMATE_REG"
   status_retire_presentation_task "$STATE" "$ID" || return 1
+# Best-effort fleet usage harvest runs while the task's state files still
+# exist; a harvest failure must never block teardown.
+"$FM_ROOT/bin/fm-usage-harvest.sh" "$ID" >/dev/null \
+  || echo "warning: usage harvest for $ID failed; continuing teardown" >&2
   fm_backlog_atomic_transition remove "$STATE/$ID.meta" "task record" "$STATE" || return 1
   rm -f -- "$STATE/$ID.turn-ended"
   printf 'teardown %s complete (remote %s:%s)\n' "$ID" "$remote_host" "$remote_home"
@@ -2872,6 +2876,10 @@ fm_backend_clear_transition "$BACKEND" "$STATE" "$T" || true
 remove_pr_poll_artifacts "$STATE" "$ID" || exit 1
 retire_busy_state "$STATE" "$ID" "$BUSY_GEN" || exit 1
 status_retire_presentation_task "$STATE" "$ID" || exit 1
+# Best-effort fleet usage harvest runs while the task's state files still
+# exist; a harvest failure must never block teardown.
+"$FM_ROOT/bin/fm-usage-harvest.sh" "$ID" >/dev/null \
+  || echo "warning: usage harvest for $ID failed; continuing teardown" >&2
 rm -f "$STATE/$ID.turn-ended" \
   "$STATE/$ID.pi-ext.ts" "$STATE/$ID.grok-turnend-token" \
   "$STATE/$ID.kimi-turnend-token" "$STATE/$ID.muse-session" \
--- a/bin/fm-teardown.sh
+++ b/bin/fm-teardown.sh
@@ -682,11 +682,13 @@ remote_secondmate_teardown() {
   tmp="$SECONDMATE_REG.tmp.$$"
   grep -vE "^- $ID( |$)" "$SECONDMATE_REG" > "$tmp" || true
   mv -f -- "$tmp" "$SECONDMATE_REG"
+  # Best-effort fleet usage harvest runs while the task's state files still
+  # exist. It must precede the status retirement below, which deletes
+  # state/<id>.status: without that log the harvest has no task window and no
+  # turn count. A harvest failure must never block teardown.
+  "$FM_ROOT/bin/fm-usage-harvest.sh" "$ID" >/dev/null \
+    || echo "warning: usage harvest for $ID failed; continuing teardown" >&2
   status_retire_presentation_task "$STATE" "$ID" || return 1
-# Best-effort fleet usage harvest runs while the task's state files still
-# exist; a harvest failure must never block teardown.
-"$FM_ROOT/bin/fm-usage-harvest.sh" "$ID" >/dev/null \
-  || echo "warning: usage harvest for $ID failed; continuing teardown" >&2
   fm_backlog_atomic_transition remove "$STATE/$ID.meta" "task record" "$STATE" || return 1
   rm -f -- "$STATE/$ID.turn-ended"
   printf 'teardown %s complete (remote %s:%s)\n' "$ID" "$remote_host" "$remote_home"
@@ -2875,11 +2877,13 @@ fm_backend_clear_transition "$BACKEND" "$STATE" "$T" || true
 [ -n "$TASK_TMP" ] && rm -rf "$TASK_TMP"
 remove_pr_poll_artifacts "$STATE" "$ID" || exit 1
 retire_busy_state "$STATE" "$ID" "$BUSY_GEN" || exit 1
-status_retire_presentation_task "$STATE" "$ID" || exit 1
 # Best-effort fleet usage harvest runs while the task's state files still
-# exist; a harvest failure must never block teardown.
+# exist. It must precede the status retirement below, which deletes
+# state/<id>.status: without that log the harvest has no task window and no
+# turn count. A harvest failure must never block teardown.
 "$FM_ROOT/bin/fm-usage-harvest.sh" "$ID" >/dev/null \
   || echo "warning: usage harvest for $ID failed; continuing teardown" >&2
+status_retire_presentation_task "$STATE" "$ID" || exit 1
 rm -f "$STATE/$ID.turn-ended" \
   "$STATE/$ID.pi-ext.ts" "$STATE/$ID.grok-turnend-token" \
   "$STATE/$ID.kimi-turnend-token" "$STATE/$ID.muse-session" \
--- a/bin/fm-teardown.sh
+++ b/bin/fm-teardown.sh
@@ -1,6 +1,9 @@
 #!/usr/bin/env bash
 # Tear down a finished task: return the treehouse worktree, release the Orca
 # worktree, or retire a secondmate home; kill the recorded runtime endpoint,
+# harvest the task's fleet usage-ledger row while its state files still exist
+# (best effort: bin/fm-usage-harvest.sh owns that ledger, and a harvest failure
+# only warns on stderr rather than blocking teardown),
 # clear volatile state, and CLOSE this home's backlog item for ship and scout
 # tasks before reporting success (a secondmate teardown closes none, since
 # secondmates are not backlog items), then refresh/prune the project's clone for
--- a/bin/fm-usage-harvest.sh
+++ b/bin/fm-usage-harvest.sh
@@ -244,6 +244,7 @@ EFFORT=$EFFORT_META
 SPAWNED=$(iso_from_epoch "$START_EPOCH" || true)
 COMPLETED=$(iso_from_epoch "$END_EPOCH" || true)
 
+mkdir -p -- "$DATA"
 jq -cn \
   --arg task "$ID" --arg harness "$HARNESS" \
   --arg model "$MODEL" --arg effort "$EFFORT" \
@@ -259,5 +260,4 @@ jq -cn \
     wall_secs:$wall, turns:$turns,
     input_tokens:$it, cached_input_tokens:$ct, output_tokens:$ot,
     reasoning_tokens:$rt, source:$source}' >> "$DATA/usage-ledger.jsonl.tmp.$$"
-mkdir -p -- "$DATA"
 cat "$DATA/usage-ledger.jsonl.tmp.$$" >> "$LEDGER" && rm -f -- "$DATA/usage-ledger.jsonl.tmp.$$"
--- a/bin/fm-usage-harvest.sh
+++ b/bin/fm-usage-harvest.sh
@@ -46,12 +46,13 @@
 # relaunch and never rewrites it afterwards, so it carries the start on a host
 # whose filesystem reports no usable birth time; because a relaunch replaces
 # it with the relaunch epoch, the status birth is what holds the window open
-# over the whole task, which is the narrowing condition stated above. The meta file's own mtime is NOT a start source in its
-# own right: later meta writes (PR registration, busy-state updates) move it
-# forward to near the end of the task and collapse the window. Only when
-# neither durable source is available (a task spawned before the token
-# existed, on a birthless filesystem) does the start fall back to the meta
-# mtime, then the status mtime.
+# over the whole task, which is the narrowing condition stated above.
+# The meta file's own mtime is NOT a start source in its own right: later
+# meta writes (PR registration, busy-state updates) move it forward to near
+# the end of the task and collapse the window.
+# Only when neither durable source is available (a task spawned before the
+# token existed, on a birthless filesystem) does the start fall back to the
+# meta mtime, then the status mtime.
 # A start later than the end (a spawn with no status append after it) is
 # pinned to the end so spawned_at never postdates completed_at.
 # Turn estimate: count of "^working:" lines in the status file.
@@ -63,10 +64,10 @@
 #     request's usage at .message.usage (input_tokens,
 #     cache_read_input_tokens, cache_creation_input_tokens, output_tokens) and
 #     Claude logs one entry per content block, so requests are deduped by
-#     .message.id before summing. cached_input_tokens folds cache_read + cache_creation (both
-#     billed on top of input_tokens); reasoning_tokens captures
-#     output_tokens_details.thinking_tokens when present (a subset of
-#     output_tokens).
+#     .message.id before summing. cached_input_tokens folds cache_read +
+#     cache_creation (both billed on top of input_tokens); reasoning_tokens
+#     captures output_tokens_details.thinking_tokens when present (a subset
+#     of output_tokens).
 #   harness=codex: <codex-sessions>/**/*.jsonl in the task window whose
 #     session_meta cwd equals the meta worktree. Per-turn token_count events
 #     carry one request's delta at .payload.info.last_token_usage
--- a/bin/fm-usage-harvest.sh
+++ b/bin/fm-usage-harvest.sh
@@ -19,27 +19,42 @@
 #    "output_tokens":<int|null>,"reasoning_tokens":<int|null>,
 #    "source":<claude-projects|codex-sessions|pi-sessions|unavailable>}
 #
+# Row scope: a row describes the WHOLE task, spanning every relaunch, and
+# wall_secs, the session-log window and turns all report that one span. The
+# turn count dictates the scope because it can only be read from
+# state/<task-id>.status, which is appended to across relaunches and carries
+# no incarnation delimiter; per-incarnation turns are therefore not derivable
+# from durable data, and scoping only the window to an incarnation would
+# report whole-task turns against a one-incarnation window and token sum.
+#
 # Wall clock: task start epoch -> status-file mtime epoch; the meta file's
 # mtime is the fallback end when the status file is absent.
-# The start comes from the epoch embedded in the meta's
-# spawn_gen=s<epoch>.<pid>.<random> incarnation token, which fm-spawn writes
-# once per spawn or relaunch and never rewrites afterwards. The meta file's
-# own mtime is NOT a start source in its own right: later meta writes (PR
-# registration, busy-state updates) move it forward to near the end of the
-# task and collapse the window. Only when spawn_gen is absent or malformed
-# (a task spawned before that token existed) does the start fall back to the
-# status-file birth epoch, then the meta mtime, then the status mtime.
-# A start later than the end (a relaunch with no status append after it) is
+# The start is the EARLIEST durable evidence of the task's first spawn: the
+# status file's birth epoch (created on the first spawn, only appended to
+# afterwards, and removed only when teardown retires the task) and the epoch
+# embedded in the meta's spawn_gen=s<epoch>.<pid>.<random> incarnation token,
+# whichever of the two is earlier. fm-spawn writes spawn_gen once per spawn or
+# relaunch and never rewrites it afterwards, so it carries the start on a host
+# whose filesystem reports no usable birth time; because a relaunch replaces
+# it with the relaunch epoch, the status birth is what holds the window open
+# over the whole task. The meta file's own mtime is NOT a start source in its
+# own right: later meta writes (PR registration, busy-state updates) move it
+# forward to near the end of the task and collapse the window. Only when
+# neither durable source is available (a task spawned before the token
+# existed, on a birthless filesystem) does the start fall back to the meta
+# mtime, then the status mtime.
+# A start later than the end (a spawn with no status append after it) is
 # pinned to the end so spawned_at never postdates completed_at.
 # Turn estimate: count of "^working:" lines in the status file.
 #
 # Per-request usage sources:
 #   harness=claude: <claude-projects>/<worktree with '/' and '.' -> '-'>/*.jsonl in
-#     the task window. Each assistant message carries one API request's usage
-#     at .message.usage (input_tokens, cache_read_input_tokens,
-#     cache_creation_input_tokens, output_tokens) and Claude logs one entry
-#     per content block, so requests are deduped by .message.id before
-#     summing. cached_input_tokens folds cache_read + cache_creation (both
+#     the task window. Claude's logs record no cwd, so that path encoding is
+#     what binds a log to this task. Each assistant message carries one API
+#     request's usage at .message.usage (input_tokens,
+#     cache_read_input_tokens, cache_creation_input_tokens, output_tokens) and
+#     Claude logs one entry per content block, so requests are deduped by
+#     .message.id before summing. cached_input_tokens folds cache_read + cache_creation (both
 #     billed on top of input_tokens); reasoning_tokens captures
 #     output_tokens_details.thinking_tokens when present (a subset of
 #     output_tokens).
@@ -64,9 +79,12 @@
 #     meta, and bare when the record carries no provider.
 #   harness=cursor, a task with a recorded remote_host (its worker ran on
 #     another machine, so its logs are not on this filesystem), an absent log
-#     tree, or no in-window match: token fields are null with source
-#     "unavailable".
-# A corrupt log line is skipped best-effort by the parser.
+#     tree, or no in-window log that yields this task's assistant usage: token
+#     fields are null with source "unavailable". Every harness applies that
+#     one rule, so a source name always asserts a parsed match and never the
+#     mere presence of an in-window file.
+# A corrupt log line is skipped best-effort by the parser, which reads each
+# line on its own and keeps the rest of that file's usage.
 #
 # Idempotent: if the ledger already contains a line whose "task" is
 # <task-id>, the command exits 0 without appending.
@@ -161,7 +179,20 @@ spawn_gen_epoch() {  # <spawn_gen token>
 }
 
 END_EPOCH=$(file_mtime_epoch "$STATUS" 2>/dev/null || file_mtime_epoch "$META")
-START_EPOCH=$(spawn_gen_epoch "$SPAWN_GEN" 2>/dev/null || file_birth_epoch "$STATUS" 2>/dev/null || file_mtime_epoch "$META" 2>/dev/null || file_mtime_epoch "$STATUS" 2>/dev/null || printf '%s' "$END_EPOCH")
+# The two durable first-spawn witnesses are combined by taking the earlier of
+# them, so a relaunch (which rewrites spawn_gen but only appends to the status
+# log) cannot narrow the window below the span the turn count already covers.
+START_GEN=$(spawn_gen_epoch "$SPAWN_GEN" 2>/dev/null || true)
+START_BIRTH=$(file_birth_epoch "$STATUS" 2>/dev/null || true)
+START_EPOCH=
+for start_candidate in "$START_GEN" "$START_BIRTH"; do
+  [ -n "$start_candidate" ] || continue
+  if [ -z "$START_EPOCH" ] || [ "$start_candidate" -lt "$START_EPOCH" ]; then
+    START_EPOCH=$start_candidate
+  fi
+done
+[ -n "$START_EPOCH" ] \
+  || START_EPOCH=$(file_mtime_epoch "$META" 2>/dev/null || file_mtime_epoch "$STATUS" 2>/dev/null || printf '%s' "$END_EPOCH")
 # Pin an impossible start to the end rather than inverting the window, which
 # would both report a negative duration and drop every session log.
 [ "$START_EPOCH" -le "$END_EPOCH" ] 2>/dev/null || START_EPOCH=$END_EPOCH
@@ -182,7 +213,9 @@ harvest_cleanup() {
 }
 trap harvest_cleanup EXIT
 epoch_to_touch() {  # <epoch>
-  date -r "$1" +%Y%m%d%H%M.%S 2>/dev/null || date -u -d "@$1" +%Y%m%d%H%M.%S
+  # Both renderings are LOCAL time because that is what touch -t reads; a UTC
+  # stamp would shift both window refs by the host's offset and drop real logs.
+  date -r "$1" +%Y%m%d%H%M.%S 2>/dev/null || date -d "@$1" +%Y%m%d%H%M.%S
 }
 # find -newer compares sub-second mtimes, so the refs only narrow to
 # [START-1, END+1]; the per-file epoch filter below then applies the true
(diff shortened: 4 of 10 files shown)


npayette84:
Closing without merging: this PR was built on the wrong base. It targets upstream main, which does not yet contain the fleet usage harvester, so the branch structurally bundled the separate crewmate launch-hardening work (bin/fm-spawn.sh permission flags and the three test files tracking those strings) alongside the harvester change this PR was meant to deliver. Its two required workflows (CI and Require no-mistakes) also never executed - both sat at action_required awaiting approval - so the only check that ran was the third-party review bot.

A replacement PR will be opened from current upstream main carrying only the accepted harvester code, tests, and documentation.

The branch fm/fm-harvester-v2 and its exact head 1b34b954ff82415ac2f4db141b3a20445e47c6eb are deliberately preserved for audit and have not been force-pushed or deleted.

PR #3404, opened 2026-08-31
fix(herdr): make presentation-lock namespace per-user

jknair:
## Problem

`fm_backend_herdr_presentation_lock_namespace()` returns the fixed path `/tmp/firstmate-herdr-presentation`.
On a multi-user host, the first user's firstmate creates it at mode 700 and owns it; every other user's firstmate then fails `fm_backend_herdr_presentation_lock_namespace_valid` (owner != current uid) and cannot create it either.
The result is that every herdr task teardown (and every spawn with presentation spaces enabled) fails with `herdr session presentation lock could not be resolved`, with no recovery available to the locked-out user.

Observed live on a shared host with two firstmate homes under different Unix users.

## Fix

Resolve the lock namespace to a deterministic per-user path: `/tmp/firstmate-herdr-presentation-<uid>`, preserving every existing validation guarantee (mode 700, owned by the current uid, not a symlink).

An earlier revision preferred `$XDG_RUNTIME_DIR`; review correctly flagged that two processes of the same user can see different `XDG_RUNTIME_DIR` states and would then resolve different lock namespaces, splitting the lock identity. The environment preference is dropped: every process of a user resolves the same path unconditionally.

A foreign-owned directory at the legacy shared path is simply ignored - never chowned, chmodded, deleted, or migrated.
The locks are ephemeral, so no migration is needed.

## Tests

Extended `tests/fm-backend-herdr.test.sh`:

- deterministic per-user path resolution, identical across environments of the same user,
- a foreign-owned/wrong-mode directory at the legacy path is ignored and resolution succeeds at the per-user path,
- validation still refuses symlinked and wrong-mode namespaces.

`bin/fm-lint.sh` and shellcheck pass on the changed script; the backend test file passes end to end.

### DIFF
--- a/bin/backends/herdr.sh
+++ b/bin/backends/herdr.sh
@@ -658,18 +658,17 @@ fm_backend_herdr_projection_workspace_label() {  # <task-id> <projection-id>
 # fm_backend_herdr_presentation_lock_namespace: per-user namespace, never a
 # fixed shared path. On a multi-user host a fixed /tmp path is created and
 # owned by whichever user gets there first, permanently locking every other
-# user out (mode 700, foreign uid). Prefers $XDG_RUNTIME_DIR (already
-# per-user, mode 700 by the XDG Base Directory spec) when it is set and
-# usable, else falls back to a uid-suffixed /tmp path. A foreign-owned legacy
-# /tmp/firstmate-herdr-presentation (no uid suffix) is never read, chowned,
-# chmodded, or migrated - it is simply not this function's return value
-# anymore, and the locks themselves are ephemeral.
+# user out (mode 700, foreign uid). Always resolves the deterministic
+# uid-suffixed /tmp path - never $XDG_RUNTIME_DIR, whose presence and value
+# can differ between two processes of the SAME user (a systemd user session
+# vs. a cron/sudo/detached invocation, or two different session ids), which
+# would split one user's own lock identity across two different directories
+# and silently defeat the mutual exclusion the lock exists for. A
+# foreign-owned legacy /tmp/firstmate-herdr-presentation (no uid suffix) is
+# never read, chowned, chmodded, or migrated - it is simply not this
+# function's return value anymore, and the locks themselves are ephemeral.
 fm_backend_herdr_presentation_lock_namespace() {
   local uid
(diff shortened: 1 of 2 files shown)


kunchenguid:
Speaking as Kun's firstmate:

Re-review of HEAD `c71a1784370fca58f812e983f9f0221f93c39976` on main `4ad8cbaeafc109a17c1af3911867b7fe9e04e801`.

**Attestation MISSING**: PR body has no `no-mistakes-pipeline-attestation:v1` binding this HEAD. Require no-mistakes is failing on that (run 33424170663). Re-raise via no-mistakes so the published attestation MATCHES this HEAD.

**Class: restore** (corrective). The fixed `/tmp/firstmate-herdr-presentation` path was mode-700 owned by whichever uid created it first, permanently locking every other user out of herdr presentation spawn/teardown. HEAD always resolves the deterministic `/tmp/firstmate-herdr-presentation-<uid>` path (XDG_RUNTIME_DIR preference dropped after Codex/Greptile P1 on `f999c51` — same-uid env split would defeat mutual exclusion). Locks are ephemeral; foreign legacy path is ignored, never chowned/chmodded/migrated. Files: `bin/backends/herdr.sh` + tests only; no `.github/workflows/*`, no AGENTS.md.

VISION (per rule):
- One captain, one interface — aligns. Multi-user hosts stop failing silently on presentation locks.
- Authority is explicit — aligns. No new autonomy or consent assumption.
- Scripts own the mechanics — aligns. Deterministic uid path; validation (mode 700, own uid, not symlink) unchanged.
- A restart is a non-event — aligns. Same uid always resolves the same namespace.
- Delegation with a spine — aligns. Shared session lock contract restored across homes of the same user.
- The fleet outlives any vendor — n/a / aligns.
- Scope — aligns. Backend adapter correctness, not merge policy.

Security: none (no workflow RCE / pin games; path stays under `/tmp` with existing ownership/mode checks).

Fork workflows approved after diff review: CI 33424170908 (in_progress), Require no-mistakes 33424170663 (FAILURE — attestation). Older HEAD runs also approved: 33422681295 / 33422681328.

This is waiting on the author (matching no-mistakes attestation for this HEAD), not the captain. Portable CI still in progress. Merge-eligible: NO until attestation MATCH + green CI. Captain-flag NOW: NO.

jknair:
Superseded by #3409, which raises the same change through the required no-mistakes pipeline with a matching attestation. Closing to keep a single canonical PR.

PR #3428, opened 2026-09-01
fix(bin): stop away-mode daemon from reading its own pane as permanently busy

cm-maple7:
## Intent

Fix the away-mode supervision daemon's injection wedge (data/fm-afk-inject-wedge/report.md, scout-investigated, root cause proven live). The daemon runs inside the captain's own supervisor pane by design (the /afk skill's no-separate-terminal exception for harnesses with a native in-pane background tool). On Herdr, a live tracked background shell pins the pane's native agent_status to 'working' for as long as the shell is alive, even after the turn has ended and the pane sits idle. pane_is_busy trusted that native busy verdict alone and short-circuited before ever checking the rendered footer, so the daemon read its own supervisor pane as permanently busy and could never deliver an away-mode escalation for its entire lifetime - proven against 7+ hours of production logs (3717 busy deferrals, 0 successful max-defer recoveries in the whole log). Implemented the report's three recommended, priority-ordered fixes: (1) root cause - pane_is_busy now requires the harness-scoped rendered busy footer (the same fm_busy_lines_match signature already used elsewhere to confirm a queued-while-busy submit) to corroborate a native busy verdict before treating the pane as unavailable; a native idle/unknown verdict was never trusted alone either, so this only changes the busy case, and falls back to trusting native only when the pane cannot be captured at all (fail toward not injecting into an unreadable pane). (2) backstop - the max-defer escape must actually escape rather than retrying the identical blocked call forever: inject_msg and escalate_flush gained a forced mode (force=1) that drops only the busy guard while keeping the composer-empty guard (the guard that actually protects against merging with a human's half-typed line), bounding worst-case silent non-delivery to FM_MAX_DEFER_SECS instead of being unbounded; also improved the wedge alarm's active-alert banner to carry the buffered item count and first line instead of only an age and a marker path, since 47 identical contentless banners in one production episode read as noise rather than an alert. (3) independent SIGTERM budget fix - fm_afk_launch_stop's wait for the daemon to exit after SIGTERM was a fixed 10s, but the daemon's cleanup trap can legitimately need up to FM_POLL seconds (15s default) plus flush time to reap its watcher child, which traps TERM but can be sitting in a foreground sleep; the wait now scales as (FM_POLL + 10s margin) so a watcher caught mid-sleep no longer produces a false 'did not exit after SIGTERM' that leaves lifecycle state needlessly preserved for retry. Explicitly out of scope / deferred to the captain as a design decision, not implemented here: the report's 'also worth doing' suggestion to reconsider hosting the away-mode daemon outside the supervisor pane entirely (removing the /afk skill's no-separate-terminal exception for Claude) - that is a bigger design tradeoff the fix above already resolves the concrete bug for. Added new regression tests for all three fixes in tests/fm-daemon.test.sh and tests/fm-afk-launch.test.sh, and updated one existing test that pinned the old (buggy) short-circuit behavior. Two pre-existing test failures were found during verification (tests/fm-wake-queue.test.sh 'a subshell reclaimed its parent's live hold', tests/fm-watch-triage.test.sh 'the fixture captured no process-event result') and confirmed unrelated: neither test file references anything touched by this change, and both failures reproduce identically with these changes fully reverted. They are filed as their own separate task and are explicitly out of scope here.

## What Changed
- `pane_is_busy` in `bin/fm-supervise-daemon.sh` now requires the harness-scoped rendered busy footer (`fm_busy_lines_match`, the same signature `fm_backend_herdr_send_text_submit` already uses) to corroborate a native `busy` verdict before treating the supervisor pane as unavailable, falling back to trusting the native verdict only when the pane cannot be captured at all; a native idle/unknown verdict was already not trusted alone.
- `inject_msg` and `escalate_flush` gained a `force` parameter (used only by the max-defer escape in `housekeeping()`) that drops the busy guard while keeping the composer-empty guard, bounding worst-case silent non-delivery to `FM_MAX_DEFER_SECS`; `inject_wedge_alarm`'s active-alert banner now carries the buffered item count and first line instead of only an age and marker path.
- `fm_afk_launch_stop` in `bin/fm-afk-launch.sh` now scales its post-SIGTERM wait with `FM_POLL` (`(FM_POLL + 10) * 4` quarter-second iterations) instead of a fixed 10s budget, so a watcher child caught in a foreground sleep no longer triggers a false "did not exit after SIGTERM".
- Added regression tests for all three fixes in `tests/fm-daemon.test.sh` and `tests/fm-afk-launch.test.sh`, updated one existing test that pinned the old busy-verdict short-circuit, and updated `.agents/skills/afk/SKILL.md` to document the corroboration requirement and forced max-defer escape.

## Risk Assessment

✅ Low: The three fixes are small, well-scoped, and each directly traces to the proven root cause: pane_is_busy now requires the rendered busy footer to corroborate a native busy verdict (falling back to native only when capture fails), the max-defer escape correctly bypasses only the busy guard via a new force parameter while the composer-empty guard (the one that actually prevents corrupting a human's half-typed line) still applies unconditionally, and the SIGTERM wait now scales with FM_POLL to match the watcher's real shutdown budget; all call sites and new regression tests were traced and are consistent with the implementation, and the two pre-existing unrelated test failures were confirmed to share no code path with this change.

## Testing

Targeted tests for both changed source files pass in full (180 assertions total, 0 failures), including all new regression coverage for the three priority-ordered fixes described in the intent; a direct manual exercise of pane_is_busy reproduces the original wedge condition (native 'busy' pinned by a live background shell against an actually-idle rendered footer) and confirms the fix now allows delivery in that case while preserving safe behavior for genuinely busy or unreadable panes. The two pre-existing unrelated failures (fm-wake-queue.test.sh, fm-watch-triage.test.sh) called out in the intent were spot-checked and reproduce as described; they are out of scope for this phase per the task instructions and are not reported as findings here.

## Pipeline

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
- `./tests/fm-daemon.test.sh (127 assertions, exit 0) - includes new tests test_pane_is_busy_herdr_native_busy_uncorroborated_by_rendered_idle, test_pane_is_busy_herdr_falls_back_to_native_when_capture_fails, test_max_defer_forced_escape_bypasses_busy_guard, test_batch_flush_does_not_force_bypass_busy_guard, and the updated test_pane_is_busy_herdr_native_busy_corroborated_by_rendered`
- `./tests/fm-afk-launch.test.sh (53 assertions, exit 0) - includes new unit_stop_wait_exceeds_configured_poll`
- `Manual direct invocation of pane_is_busy() (sourced from bin/fm-supervise-daemon.sh) against 3 scenarios: native-busy+idle-footer (reproduces the reported wedge, now resolves as not-busy), native-busy+busy-footer (still correctly reports busy), and native-busy+capture-failure (falls back to native verdict, fails safe)`
- `git status --porcelain confirmed no transient artifacts left in the worktree after testing`
</details>

<details>

- ℹ️ `.agents/skills/afk/SKILL.md:106` - .agents/skills/afk/SKILL.md documents the max-defer escape contract twice (Busy-guard section and Injection hardening section) and pane_is_busy behavior likewise appears near duplicated phrasing across the file; both copies were updated to stay accurate for this change, but a follow-up could consolidate into one owner passage with a pointer.
</details>

<details>

🔧 Fix: No code changes needed; lint passes after actionlint install
</details>

<details>

✅ No issues found.
</details>

### DIFF
--- a/.agents/skills/afk/SKILL.md
+++ b/.agents/skills/afk/SKILL.md
@@ -92,7 +92,8 @@ The daemon never injects into an in-use pane. Two checks run before every
 injection, dispatched through `bin/fm-backend.sh` for the supervisor's own
 backend (tmux or herdr; see "Auto-discovered supervisor pane" below):
 
-- **Primary-pane busy guard** - `pane_is_busy` trusts Herdr native `busy` when available, otherwise matches rendered output against only the detected primary harness's signature.
+- **Primary-pane busy guard** - `pane_is_busy` requires the rendered output to match the detected primary harness's busy signature; a native `busy` verdict (e.g. Herdr's `agent_status`) is corroborating evidence, not sufficient alone, because a live tracked background shell (such as this daemon's own launch) can pin native state to `working` long after the pane is actually idle.
+  When the pane cannot be captured at all, `pane_is_busy` falls back to the native verdict rather than guessing idle.
   This narrow delivery guard never classifies a recorded worker task and never uses a global union of vendor patterns.
 - **Composer-state guard** - `inject_msg` reads the full `empty`/`pending`/`pending-unproven`/`unknown` verdict from `fm_backend_composer_state` and injects only when it is affirmatively `empty`.
   Every other or future verdict defers, including an unreadable pane, ambiguous geometry, a blank unidentified row, and a bare shell prompt left after the agent exits.
@@ -105,12 +106,17 @@ In afk mode the composer guard is belt-and-suspenders (no human is typing), but 
 
 **Max-defer escape (the daemon must never silently wedge).**
 If anything stays buffered past `FM_MAX_DEFER_SECS` (default 300), the daemon
-attempts one normal flush, which still requires an idle pane and an affirmatively empty composer.
+attempts one FORCED flush that drops only the busy guard; the composer-empty guard
+still applies unconditionally, since that is what actually protects against merging
+with a human's half-typed line.
+This bounds worst-case silent non-delivery to `FM_MAX_DEFER_SECS` instead of retrying
+the identical blocked call forever.
 The alarm is defense in depth rather than a substitute for keeping every genuinely idle supported composer injectable.
-If that submit cannot be confirmed, it raises a loud, rate-limited wedge alarm:
+If that forced submit still cannot be confirmed, it raises a loud, rate-limited wedge alarm:
 an ERROR in the daemon log, a durable
 `state/.subsuper-inject-wedged` marker (surface it on the "while you were out"
-catch-up if present), a tmux status-line flash when applicable, and a configurable backend-independent active alert.
+catch-up if present), a tmux status-line flash when applicable, and a configurable backend-independent active alert
+carrying the buffered item count and first line so the banner is actionable rather than a bare age and marker path.
 `docs/wedge-alarm.md` owns the alert channel setup, and `docs/verification/supervision.md` "Wedge-alarm channels" owns active evidence.
 So a guard false-positive becomes a visible stall, never an unbounded silent no-op.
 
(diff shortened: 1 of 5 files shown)


cm-maple7:
Withdrawn - opened against upstream by mistake; this change is being landed in our own fork. Apologies for the noise.

PR #3429, opened 2026-09-01
refactor(agents): move nameable-situation contracts out of AGENTS.md

lmktechnology:
## What and why

`AGENTS.md` is loaded into every firstmate session's context on every turn, so content that only applies to a nameable situation pays always-loaded cost for occasional value. This applies the knowledge-placement decision tree from `firstmate-coding-guidelines` to every section and moves that content to its real owner.

**No behavior or safety boundary changes.** Every hard rule, escalation boundary, and lifecycle contract is still enforced - either inline because it applies every turn, or through a skill trigger that fires on the same condition that made it visible before.

`AGENTS.md`: **72,768 -> 50,720 bytes (30.3% smaller)**, 585 -> 483 lines. All 16 sections and all 5 hard rules intact.

## What moved where

| Removed from AGENTS.md | New owner | Why |
|---|---|---|
| Section 2's exhaustive `state/` + `config/` tree | `docs/configuration.md` | Pre-existing one-owner violation: that doc already declares itself single owner of the home layout and explicitly declines to hold an exhaustive state tree |
| Section 3's digest stage enumeration | `bin/fm-session-start.sh` header | Was a drifting second copy of the script's own stage list |
| Section 4's quota selection procedure | `quota-array-dispatch` | Skill already owned it |
| Section 4's effort fallback detail | `harness-adapters` | Skill already owned it, including the never-`max`-without-captain-preference rule |
| Section 7's validation-run supersession sequence | new `validation-run-changes` skill | Nameable situation, not every-turn content |
| Section 7's decision-delivery contract | `ask-user-authority` | Skill already owned the decision itself; delivery was the missing half |
| Custom-check authoring contract | `bin/fm-check-register.sh` header | Exact mechanics belong with the owning script |
| Promoted-worker procedure | `bin/fm-promote.sh` header | Same |

What was deliberately **kept** inline: section 9's captain-facing translation glossary. Three skills (`bootstrap-diagnostics`, `bearings`, `ahoy`) cite "AGENTS.md section 9's captain-facing translation contract" as its owner, and it applies to every captain-facing message. That is an intentional non-removal.

Every removed fact was audited for an owner elsewhere first. Exactly one lived only in `AGENTS.md` - that a registered process-to-event source keeps a live supervision cycle required even with no fleet work - and it was relocated into section 8 rather than dropped.

## Cross-references swept

Rather than leaving pointers aimed at removed text:

- `bin/fm-spawn.sh`'s muse reasoning-effort comment cited section 4 for the max-effort rule that now lives in `harness-adapters`; repointed.
- `project-management` no longer restates the per-task surface classification verbatim; it points at section 7.
- `quota-array-dispatch`'s ownership sentence now names what section 4 keeps versus what the skill owns.
- The status-line-is-a-wake-event rule was placed in section 8 rather than section 2, because `bin/fm-session-start.sh` and `bin/fm-line-cap-lib.sh` both cite section 8 as its owner.
- `bin/fm-classify-lib.sh` now documents its own open-decisions cursor path, its removal on teardown, and why the file is safe to delete by hand.

## Tests

No tests added, deliberately. The three touched shell files (`bin/fm-check-register.sh`, `bin/fm-classify-lib.sh`, `bin/fm-spawn.sh`) changed in **comments only** - there is no executable behavior change in this diff. A test asserting `AGENTS.md` contains or omits particular strings would be the source-grep anti-pattern, since instruction prose is deliberately rewritten over time. The executable enforcement is `bin/fm-doc-audience-check.sh`, which passes at 90 classified surfaces (the new skill is registered in `docs/documentation-audiences.json`).

## Validation

Local no-mistakes pipeline: review **0 findings**, test **0 findings**, document **0 findings**. The document step contributed one fix, included here - it collapsed a partial duplicate in the new skill into a pointer at section 7, which is the same one-owner rule this PR is about.

One caveat for reviewers: the pipeline's lint step did not execute. `.no-mistakes.yaml` sets `lint: 'bin/fm-lint.sh'` as a bare path, which cmd.exe cannot run on Windows (`'bin' is not recognized as an internal or external command`). The linter failed to launch rather than reporting problems. That config is pre-existing and untouched here; it silently disables the lint gate for every Windows run and is worth a separate fix. Lint was verified manually with pinned ShellCheck 0.11.0 and full extended analysis on all changed scripts against an LF-normalized copy, and CI on Linux runs it for real on this PR.

### DIFF
--- a/.agents/skills/ask-user-authority/SKILL.md
+++ b/.agents/skills/ask-user-authority/SKILL.md
@@ -47,6 +47,16 @@ State all five of these elements in one concise, evidence-first escalation:
 
 Do not relay reviewer labels or gate output as if they settled the decision.
 
+## Deliver the decision
+
+This skill also owns how a decided finding reaches the worker, because the delivery step is what makes the decision stick.
+
+Send the same worker one exact decision naming the decision key, the step, the action, the affected finding IDs, instructions where they are needed, and the exact response command to run.
+Pass `fm-send`'s `--resolve-key` so the worker's open decision record closes at answer time rather than staying open behind a later `working:` or `done:` line.
+Require the matching `resolved` event before treating the decision as landed.
+Forbid `--yes`: it would silently bypass firstmate's authority check and any required captain escalation.
+Require the worker to process every synchronous return until completion or a genuinely new escalation, rather than stopping at the first gate response.
+
 ## Classification examples
 
 - Fixing a concrete defect that violates an original acceptance criterion is firstmate's to decide, regardless of implementation difficulty.
--- a/.agents/skills/project-management/SKILL.md
+++ b/.agents/skills/project-management/SKILL.md
@@ -41,7 +41,7 @@ Choose that posture when adding or creating the project:
 - `no-mistakes` runs the full validation pipeline before a PR.
 - `direct-PR` pushes and opens a PR without the no-mistakes pipeline.
 - `local-only` has no required remote or PR and lands only through the approved local fast-forward path.
-- `no-mistakes-prod-only` is a conditional policy rather than one flat mode: genuinely internal-only tooling, automation, contributor or operator process, and release or submission work ships `direct-PR`, while product-facing, mixed, and uncertain work ships `no-mistakes`.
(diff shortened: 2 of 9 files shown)


lmktechnology:
Superseded: this change already landed directly on main via cherry-pick (commit 0801240). The ps-ancestry fix that was blocking this fork PR's CI is now up separately as #3551.

PR #3460, opened 2026-09-01
feat(bin): publish the fleet snapshot on an opt-in cadence

milesibastos:
## Intent

Make a fleet home publish its snapshot on a cadence, so a surface that watches the file rather than running the command stays current on its own.

The finding: bin/fm-fleet-snapshot.sh is invoked on demand only (by fm-fleet-view.sh, fm-bearings-snapshot.sh, fm-home-summary-refresh.sh). Verified 2026-09-01: no timer, no watcher hook, no registered source writes it periodically. That is fine for a command but not for a consumer that watches a FILE, which is what the quarterdeck panel does - it reads bytes and watches directories for change, and contains no way to run a command by design (child_process is banned repository-wide there, and that ban is the whole safety argument for its write path). So the panel renders whatever the last manual run left and honestly badges it "43M AGO - OLD".

The captain answered on 2026-09-01: republish on a cadence from the fleet side. They explicitly REJECTED tying it only to the supervision poll, on the ground that a stopped supervision cycle would then silently freeze the surface - which is the exact state one home is in right now.

What to build: a guarded way for a home to publish its own current snapshot to a stable path on a cadence, off any conversational turn. Design decisions were mine, under these constraints, each a real one rather than a preference:
- Cadence is configurable and opt-in per home. A home with no surface watching a file must not pay for this. Absent configuration means absent behaviour, and the mechanism must say plainly which state it is in.
- A publish is atomic. A consumer must never observe a half-written snapshot; it either sees the previous whole one or the next whole one.
- A failed read never destroys the last good snapshot. Degrading to stale is correct; degrading to absent or truncated is not, because the consumer's own honesty depends on being able to say how old the picture is.
- The snapshot's own age must remain readable from the published artifact.
- A snapshot read costs real work on a large fleet. Say what the cadence costs and make a defensible default, rather than choosing a small number because it feels responsive.
- It must not run in a conversational turn, and must not require an agent to be alive. Look at how this repository already supervises recurring background work before inventing a new mechanism - if an existing one fits, use it and say why; if it does not, say precisely why not.

EXPLICITLY OUT OF SCOPE: do not change bin/fm-fleet-snapshot.sh's output, schema, or schema identifier. A consumer pins fm-fleet-snapshot.v1 and refuses to render on a mismatch, which is correct behaviour that must not be tripped. This adds a publisher; it does not edit the thing published.

Definition of done beyond the standard contract:
- Enabled on a real home and observed to republish on its own, with the artifact's age visibly advancing - proved, not asserted.
- Colocated shell tests in this repo's style covering at minimum: the disabled default, an enabled publish, a failed underlying read leaving the previous snapshot intact, and the atomicity of a publish.
- docs/configuration.md carries the configuration, since that file is the single owner of the operational-home configuration schema.
Required reading before editing: the firstmate-coding-guidelines skill (this is firstmate's shared tracked material), then bin/fm-fleet-snapshot.sh's header and docs/configuration.md.

DECISIONS AND TRADEOFFS MADE WHILE DOING THE WORK (a reviewer reading only the diff would not know these were deliberate):

1. No existing mechanism was reused, and the script header states precisely why, because the brief demanded that justification. Watcher check shims (the pattern bin/fm-tool-update-check.sh arms) and process-event sources are both polled by a watcher that an AGENT arms and that ends with the agent's cycle - that is exactly the supervision coupling the captain rejected. The away-mode sub-supervisor is gated on state/.afk by contract. There is no launchd, cron, or systemd unit anywhere in bin/, and adding a host-level timer would put a unit outside the operational home that no fm- command owns. What WAS reused: bin/fm-startup-network.sh's proven detach shape (nohup, own process group, stdio to /dev/null) for work that must outlive its launching shell, and bin/fm-home-summary-refresh.sh's publication contract (serialize under a lock, validate, rename).

2. Deliberately NOT touching bin/fm-fleet-snapshot.sh at all. The published artifact is its exact --json bytes. In particular the artifact carries no generated_epoch and no added freshness field, because adding one would change the pinned schema; age is read from the document's own ISO-8601 `generated` string, which was already there.

3. Cadence default and the 30s floor are argued from measurement, not feel. One read measured 2.9s wall / 0.4s CPU / 49KB JSON on a primary home with 6 task records and no secondmates, dated 2026-09-01 in the header, docs, and commit message. 300s is recommended (~1% of a core). 60s was explicitly rejected: 5% of a core continuously to shave four minutes off a staleness badge. Below the 30s floor a configured value is REFUSED rather than accepted and degraded; FM_FLEET_PUBLISH_MIN_CADENCE is the documented seam that lets tests use a short cadence.

4. Malformed configuration is refused as malformed, never silently treated as a default. This is the same strictness contract config/startup-memory-budget uses (regular, single-linked file, one positive integer, exactly one newline, non-symlinked config/). Rationale: a home that believes it is publishing and is not is the precise failure this whole mechanism exists to remove. read_cadence therefore returns three distinct states (0 configured, 1 absent/disabled, 2 present-but-unusable) and reports through globals rather than stdout specifically so a caller cannot lose the reason to a subshell - an earlier draft did exactly that and printed "misconfigured ()".

5. The daemon re-reads its configuration every tick, not only at startup. Without that, "absent configuration means absent behaviour" would only be true at start time. Consequence: removing the file stops a running publisher within one tick and leaves the last published snapshot in place, and a changed cadence takes effect without a restart.

6. Liveness is two independent facts - the recorded pid is alive AND a beacon touched every tick is fresh - not the pid alone. A reused pid would report a stopped publisher as healthy, and a surface that silently stops updating is the exact failure being removed. The same evidence gates `stop`, so a recorded pid that a reboot handed to an unrelated process is refused rather than signalled.

7. Temporary files are dot-prefixed in the state directory. This is a real constraint, not hygiene: the consumer watches the DIRECTORY, so a visible temporary name would be a file-appearance event it could try to read. A test asserts no non-dot temporary ever appears beside the artifact.

8. Session-start arming was added to bin/fm-bootstrap.sh as a seventh local MUTATING sweep, because something has to bring the publisher back after a reboot or crash and a locked session boundary is where a home reliably notices. It is idempotent, silent on success, prints an actionable FLEET_PUBLISH: line on failure, and never runs under FM_BOOTSTRAP_DETECT_ONLY. This required updating the "six MUTATING sweeps" counts in that script's header and AGENTS.md section 3, and adding a FLEET_PUBLISH: handler to the bootstrap-diagnostics skill whose trigger list AGENTS.md section 13 owns.

9. Knowledge placement followed the firstmate-coding-guidelines decision tree: exact mechanics live in the script header, the operator-facing schema in docs/configuration.md (the single owner), one-line entries in AGENTS.md sections 2/3/13, one row in docs/scripts.md, and the diagnostic handling in the agent-only skill. No contract is stated twice.

PROOF OBTAINED (the DoD asked for proved, not asserted):
- Republishing observed through the REAL producer on a seeded home, `generated` advancing 15:18:07 -> 15:18:12 -> 15:18:47 -> 15:18:52 -> 15:18:57Z across shells that had already exited.
- Detachment proven: PPID=1 and its own process group.
- Crash recovery proven: SIGKILL the publisher, then both `start` and session-start arming steal the dead holder's singleton lock and resume.
- 16 colocated tests in tests/fm-fleet-publish.test.sh, registered in the snapshot-bearings family with a weight entry.
- Production enablement on the captain's live home was deliberately NOT done: worktree isolation rules forbid writing outside this copy, so it is left as a one-line config write.

KNOWN ENVIRONMENT GAP, not caused by this change: actionlint 1.7.12 is absent from this host's PATH, so bin/fm-lint.sh exits 1 on the workflow-lint half. With actionlint installed into a scratch directory on PATH, bin/fm-lint.sh passes fully (ShellCheck 0.11.0 clean, 3 workflow files valid). No workflow files were changed by this work.

## What Changed

- Added `bin/fm-fleet-publish.sh`, a detached publisher daemon (`start`/`stop`/`status`) that republishes `bin/fm-fleet-snapshot.sh --json` byte-identical to `state/fleet-snapshot.json` on a configurable, opt-in cadence — writing via a dot-prefixed temp file and atomic rename, leaving the previous artifact untouched on a failed read, and tracking liveness through a pid record plus a per-tick beacon guarded by a singleton lock.
- Wired the publisher into session start: `bin/fm-bootstrap.sh` gained a seventh local MUTATING sweep (`fleet_publish_arm`) that idempotently (re)starts the publisher when `config/fleet-snapshot-cadence` is present and prints a `FLEET_PUBLISH:` diagnostic on failure; updated the "six sweeps" references to seven across `bin/fm-bootstrap.sh`, `bin/fm-session-start.sh`, and `AGENTS.md`.
- Documented the new `config/fleet-snapshot-cadence` schema (format, floor via `FM_FLEET_PUBLISH_MIN_CADENCE`, atomicity, and cost guidance) in `docs/configuration.md`, added a `docs/scripts.md` row for the new script, added a `FLEET_PUBLISH:` handler to the `bootstrap-diagnostics` skill's trigger list, and registered `tests/fm-fleet-publish.test.sh` in the `snapshot-bearings` family with a weight entry in `bin/fm-test-run.sh`.

## Risk Assessment

✅ Low: The round-1 finding (oversized cadence values silently bypassing the floor check via a bash `[ -lt ]` overflow-to-false) is durably fixed: a safe string-length check (`${#value} -gt 10`) now runs before the numeric comparison, refusing anything that could overflow while leaving ample headroom (up to ~10 billion seconds) for legitimate cadences; the fix is accompanied by a behavioral regression test that fails before and passes after, driven through the real CLI. Header docs and docs/configuration.md were updated consistently. A full pass over the rest of the branch (daemon lifecycle, bootstrap session-start arming, docs, test registration) found no additional defects, and all touched files are consistent with the stated design decisions.

## Testing

The reviewed fix correctly closes the overflow gap (pre-fix script silently accepts and starts a publisher on a 20-digit cadence; post-fix script refuses it as misconfigured at status/start/run), and the full 16-test colocated suite plus manual end-to-end runs against the real snapshot producer confirm the broader feature's DoD claims (unattended republish with advancing artifact age, atomic publish, failed-read leaves snapshot intact, cadence removal stops cleanly, session-start arming). The sandbox's Bash tool lacks a controlling terminal, which breaks macOS `nohup` (used identically elsewhere in the repo by fm-startup-network.sh) independent of any product code; this was worked around with a PATH-shimmed `nohup` that only replaces the broken console-detach step, not any behavior under test, to get full suite coverage rather than reporting it as a blocker.

<details>
<summary>Evidence: Full test suite + regression proof + manual republish/stop verification transcript</summary>

Source: [Full test suite + regression proof + manual republish/stop verification transcript](https://github.com/milesibastos/firstmate/blob/8870b88e37d8816ed5f964ce090186d71e44ea8c/.no-mistakes/evidence/fm/fm-snapshot-cadence-e2/manual-republish-verification.log)

```text
Verification of bin/fm-fleet-publish.sh on branch fm/fm-snapshot-cadence-e2
target commit 101578ebc5c659f096882a617c8c55e9f11dc64b, run 2026-09-01.

=== Environment note: sandbox nohup gap, worked around ===

`cmd_start` in bin/fm-fleet-publish.sh detaches via `nohup ... &`. In this
sandboxed test-execution shell, bare `nohup` fails before exec'ing anything:

  $ nohup echo test >/dev/null 2>&1 </dev/null
  nohup: can't detach from console: Inappropriate ioctl for device
  $ [ -t 0 ] && echo tty || echo not-a-tty
  not-a-tty

This is a seatbelt/no-controlling-terminal artifact of this Bash tool's
sandbox, reproduces identically with zero product code involved, and would hit
the repo's only other user of this exact idiom (bin/fm-startup-network.sh)
the same way. Worked around for this run with a minimal PATH-shimmed `nohup`
(`trap '' HUP; exec "$@"`) that preserves every other real behavior the tests
check (own process group via `set -m`, stdio->/dev/null, outliving the caller)
and only substitutes the console-detach step that this sandbox cannot perform.

=== Full suite, target commit, with the nohup shim on PATH ===

`PATH=/tmp/nohup-shim:$PATH ./tests/fm-fleet-publish.test.sh` (jq present, bash 5.3.15), exit 0:

  ok - an unconfigured home reports itself disabled and publishes nothing
  ok - a malformed or too-small cadence is refused rather than silently defaulted
  ok - a cadence too large to compare safely is refused rather than accepted
  ok - an enabled home publishes a schema-valid snapshot carrying its own age
  ok - a detached publisher republishes the snapshot on its own cadence
  ok - the publisher is detached from the shell and process group that started it
  ok - starting an already-running publisher does not start a second one
  ok - a crashed publisher is recovered by the next start rather than blocking it
  ok - stop refuses a recorded pid that is not beating instead of signalling it
  ok - a failed snapshot read leaves the previous published snapshot intact
  ok - publication resumes on its own once the producer recovers
  ok - removing the cadence stops the publisher and leaves the snapshot in place
  ok - a publish is atomic: a reader sees the previous whole snapshot or the next one
  ok - session start leaves a home that did not opt in alone
  ok - session start reports an unusable cadence instead of publishing nothing quietly
  ok - a read-only session start arms no publisher
  ok - session start arms the publisher on an opted-in home

All 16 behavioral assertions pass, including detachment/own-process-group,
crash recovery, atomicity, failed-read-leaves-snapshot-intact, and session-start
arming/opt-out.

=== Regression check: does the new test actually catch the pre-fix bug? ===

Checked out bin/fm-fleet-publish.sh from 4b0063b (parent commit, before the
review-round fix in 101578e) in place, ran it against a 20-digit cadence:

  $ printf '99999999999999999999\n' > config/fleet-snapshot-cadence
  $ FM_HOME=<tmp home> ./bin/fm-fleet-publish.sh status
  ./bin/fm-fleet-publish.sh: line 236: [: 99999999999999999999: integer expected
  publisher: enabled cadence=99999999999999999999s daemon=stopped (run: bin/fm-fleet-publish.sh start)

  $ FM_FLEET_PUBLISH_START_WAIT=3 FM_HOME=<tmp home> ./bin/fm-fleet-publish.sh start
  ./bin/fm-fleet-publish.sh: line 236: [: 99999999999999999999: integer expected
  publisher: started pid=38502 cadence=99999999999999999999s
  start exit=0
  PRE-FIX: publisher IS running with oversized/overflowed cadence

Confirms the exact fail-open described in the review finding: the pre-fix
floor check (`[ "$value" -lt "$MIN_CADENCE" ]`) errors on overflow and `if`
reads that as false, so the absurd value is silently accepted as "enabled"
and a publisher actually starts with it. Restored bin/fm-fleet-publish.sh to
the target commit (101578e) afterward; `git status` confirmed a byte-identical,
clean working tree. Post-fix, the same scenario is refused: `status` reports
"misconfigured (...20 digits, more than the 10-digit maximum...)" and both
`start` and `run` exit non-zero (see "ok - a cadence too large to compare
safely is refused rather than accepted" above).

=== Manual product-level proof: artifact age advances unattended ===

Independently of the daemon-detach test, ran the tick loop directly
(`fm-fleet-publish.sh run`, backgrounded with plain `&`) against the REAL
`bin/fm-fleet-snapshot.sh` producer on a freshly seeded temp home, cadence=1s
via the documented FM_FLEET_PUBLISH_MIN_CADENCE test seam:

  publisher pid=62086
  t+1x2s generated=2026-09-01T18:26:01Z
  t+2x2s generated=2026-09-01T18:26:04Z
  t+3x2s generated=2026-09-01T18:26:06Z
  t+4x2s generated=2026-09-01T18:26:07Z
  t+5x2s generated=2026-09-01T18:26:10Z
  publisher still alive
  status: snapshot: .../fleet-snapshot.json generated=2026-09-01T18:26:10Z published=0s ago

Then removing the cadence file stopped the publisher within one tick and left
the last published snapshot exactly in place:

  before removal, generated=2026-09-01T18:26:27Z, publisher alive=yes
  after cadence removed, publisher alive=no
  snapshot after removal generated=2026-09-01T18:26:27Z (unchanged, last-good preserved)
  log: [2026-09-01T18:26:29Z] cadence configuration was removed; the publisher stopped and left the last published snapshot in place

Worktree left clean; no stray fleet-publish processes or temp files remain.
```
</details>

## Pipeline

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
- <code>`./tests/fm-fleet-publish.test.sh` (target commit, real PATH): 4/16 assertions run before hitting a sandbox-only nohup/tty limitation (no controlling terminal) unrelated to this diff; all 4 passed, including the new oversized-cadence regression test</code>
- <code>`PATH=&lt;nohup-shim&gt;:$PATH ./tests/fm-fleet-publish.test.sh` (target commit, with a minimal PATH shim that only substitutes nohup&#39;s broken console-detach ioctl step, preserving own-process-group/detach/stdio behavior): all 16/16 assertions pass, including detachment, crash recovery, atomicity, failed-read-leaves-snapshot-intact, and session-start arming/opt-out</code>
- <code>Regression proof: checked out bin/fm-fleet-publish.sh from parent commit 4b0063b in place, confirmed a 20-digit cadence overflows `[ -lt ]`, is silently accepted, and `start` actually launches a publisher with it (`publisher: started pid=... cadence=99999999999999999999s`); restored the target commit&#39;s script and confirmed `git status` clean</code>
- <code>Manual verification: `fm-fleet-publish.sh run` backgrounded against the real bin/fm-fleet-snapshot.sh producer on a seeded temp home, observed `generated` advance across 5 unattended polls, then observed cadence-file removal stop the publisher within one tick while leaving the last snapshot byte-identical, with the reason logged</code>
</details>

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>


milesibastos:
Withdrawing this: our fleet has adopted a fork-only policy for its customisations, so this work belongs in our own fork rather than here. Nothing about the change was wrong and no review time was wasted — thank you for the automated pass, whose P1 on beacon identity binding was a real defect and is being fixed on our side. If this capability would be wanted upstream, we're glad to re-open it as a deliberate contribution rather than a default route.

PR #10, opened 2026-06-12
fix: keep idle watcher supervision silent

kunchenguid:
## Intent

The captain asked me to update firstmate's own instructions so future supervision through bin/fm-watch.sh does not waste user-facing turns with idle status messages. The change should clarify that waiting on the watcher is intentionally silent, empty polls and elapsed waiting time are internal tool bookkeeping, unchanged heartbeat reviews should not be reported, and watcher mechanics such as restart/poll/no-change confirmations do not reach the captain unless they reveal a captain-relevant issue.

## What Changed
- Clarifies firstmate supervision instructions so waiting on `bin/fm-watch.sh`, empty polls, elapsed waiting time, and no-change heartbeat reviews remain internal instead of user-facing updates.
- Updates the README and watcher script wording to describe `fm-watch.sh` as waking when supervision work is due, including heartbeat reviews, not only when a crewmate needs attention.
- Documents that routine watcher mechanics such as polling, restarting, and confirming no status changed should not reach the captain unless they surface a relevant issue.

## Risk Assessment

✅ Low: The change only clarifies documentation for watcher communication behavior and does not alter executable code or operational mechanics.

## Testing

I verified the documentation-only behavior change by inspecting the changed AGENTS.md supervision flow and producing an evidence transcript showing every intended watcher-silence rule present; no source tests were applicable, and the working tree remained clean.

<details>
<summary>Evidence: Watcher-silence instruction evidence</summary>

<code>309:Waiting on the watcher is intentionally silent.&#10;310:After starting or restarting it, do not send idle progress updates to the captain; wait until it returns `signal`, `stale`, `check`, or `heartbeat`, unless the captain asks for status.&#10;311:Empty polls, elapsed waiting time, and &#34;still no change&#34; are tool bookkeeping, not conversational progress.&#10;324: A heartbeat with no captain-relevant change is internal; do not report that the fleet is unchanged.&#10;344:Silence is the correct state while a healthy background watcher is waiting.&#10;366:Routine watcher mechanics include restarting the watcher, polling a waiting watcher, and confirming that no status changed.</code>

```text
Changed watcher-silence instructions in AGENTS.md:

309:Waiting on the watcher is intentionally silent.
310:After starting or restarting it, do not send idle progress updates to the captain; wait until it returns `signal`, `stale`, `check`, or `heartbeat`, unless the captain asks for status.
311:Empty polls, elapsed waiting time, and "still no change" are tool bookkeeping, not conversational progress.
[...]

PR #3506, opened 2026-09-02
fix(wardroom): stop the preflight gate asserting a property of the machine

stoneevenson-biz:
`main` is red in CI on `tests/fm-wardroom-t1-preflight-rules.test.sh`. It came in with #10 and was merged without a CI check, so this is repairing a known-bad merge.

## What was actually wrong

**A fixture, not the preflight.** `tests/fixtures/preflight/primary-checkout.md` hardcoded `/Users/<captain>/firstmate/data/old-task-q4/report.md` as one of its three offenders. That path is under a checkout root only where `$HOME` is `/Users/<captain>`. Everywhere else it is under no root at all, so the preflight correctly did not report it, and the assertion naming it failed. The gate was green on exactly one laptop on earth.

That is our own standard broken: **a test must never assert a property of the machine it runs on.** CI's `/home/runner/firstmate` was never the problem; the fixture pinning one machine was.

**On the reported "collapsed glob": there is no collapse.** The glob in the refusal is the text the brief itself wrote — *"fix the 68 stale briefs under `data/*/brief.md`"* — quoted back with its line number, which is exactly what naming the offender means. All three offenders were always emitted as three separate TSV findings; in CI only two of them existed. The two-line output in the CI log was a dropped offender, not a merged one.

Verified directly:

```
primary-checkout  <root>/firstmate/data/*/brief.md
primary-checkout  <root>/firstmate/AGENTS.md
primary-checkout  <root>/firstmate/data/old-task-q4/report.md
```

The gate now **counts the records**, so "three findings" can no longer be confused with "one finding that happens to contain three substrings" — which is the assertion that would have caught the original defect, and the one a substring check structurally cannot make.

## The fix

- Fixtures write `__PRIMARY__` wherever an absolute checkout root is meant; the suite substitutes a root it chose itself.
- `$HOME` is synthetic for every preflight run, so the `~/firstmate` form the other fixtures use no longer reaches for the machine's real layout either.
- Both primary-checkout cases — the refusal and the lookalike — run under **two roots** differing in depth and in name, asserting each offender against the root that run used.
- The lookalike's sharpest near-miss is now `<root>-old` rendered from the root under test. As a literal it was a real boundary test on one machine and vacuously true everywhere else; it now catches a boundary regression anywhere.

`bin/fm-preflight-lib.sh` is unchanged — the preflight was doing the right thing.

## Gated

**Red first, twice.** With `render()` pinned back to one machine's root the suite fails naming the dropped offender. With `under()`'s boundary check reduced to a bare prefix test, the lookalike case fails under both roots. Green under three different `$HOME`s.

- `bash tests/run-all.sh` — **71 ran, 2 skipped, 0 failed**
- `bash gates/verify.sh` — `green:48 frozen:0 red:2 unproven:0 wip:2`; the two reds are the declared ones, `gate-l2-loop-audit-level` and `m1-hook-registered`, and nothing else. The nine gates the sweep demoted are re-frozen, so the ledger diff against `origin/main` is timestamps plus one observable edit.
- `shellcheck bin/*.sh tests/*.sh` — clean.
- Rebased onto `origin/main` (93162ce4).

## Expected to stay red, and not from this branch

CI's **`PR must be raised via no-mistakes`** check fails on every branch now, because firstmate moved this project to `direct-PR` mode and raises PRs itself. That is a policy gap between the delivery mode and the workflow, not something this change touches or could fix.

So: the suite and the gate ledger pass; the no-mistakes provenance check remains red for every PR until that workflow is reconciled with `direct-PR` mode.

## Note

`CONTRIBUTING.md` gains the durable rule next to the other test-hygiene rules. It went there rather than `AGENTS.md` because that is where gate- and test-authoring rules already live.

https://claude.ai/code/session_01PzTiy6wHvj4nUJFnwXiuu8


stoneevenson-biz:
Opened against the wrong repository by mistake — this change belongs on the fork it was branched from, not here. Closing; apologies for the noise.

PR #3521, opened 2026-09-02
docs(agents): trim section 2 state/config detail into pointers

alvinleyble:
## Summary
- AGENTS.md section 2 inlined config defaults/fallback chains and the full `<id>.meta` field schema that docs/configuration.md and each producing script's header already own
- Replaced those entries with name + purpose + pointer, and collapsed three separate "never touch" internal-glob lines in state/ into one
- No behavior change; reduces the file's always-loaded token cost by roughly 10%

## Test plan
- [ ] Confirm no doc/skill elsewhere in the repo pointed at the removed inline detail (spot-checked: none found)
- [ ] Read through the diff to confirm every dropped clause is genuinely restated in docs/configuration.md or the named script's header

https://claude.ai/code/session_017HS1989PWucaLZWWWhyeDh


alvinleyble:
Opened against the wrong repo by mistake (should target alvinleyble/firstmate, not this upstream). Closing and resending correctly.

PR #3544, opened 2026-09-02
feat(bin): add agy (Antigravity CLI) as a verified crewmate/scout adapter

tpeixinho:
## Intent

Add `agy` (Google's Antigravity CLI) as a VERIFIED firstmate harness adapter, scoped to crewmate and scout ONLY - the same limited scope muse already has. It must NOT become a primary or a second mate.

A harness value stays undispatchable until the executable owner, the portable regression, the applicable live guard, and the verification record land together, so all eight owners land in one change: bin/fm-harness.sh (marker + ancestry), bin/fm-spawn.sh (launch shape, autonomy, guarded global Stop-hook install) with removal in bin/fm-teardown.sh, bin/fm-composer-lib.sh (busy signature), bin/fm-busy-lib.sh (the agy arm), bin/fm-control-lib.sh (interrupt/exit/resume), bin/backends/tmux.sh (liveness by process name), the harness-adapters reference plus its harness-adapter-routing-v1 map entry, and a dated record under docs/verification/ plus docs/configuration.md and AGENTS.md section 4.

MEASURED FACTS THE ADAPTER HONOURS (agy 1.1.24, macOS arm64, Herdr 0.8.2, isolated lab, 2026-09-02):
1. Launch: agy REJECTS a positional prompt ("Error: unexpected argument") and reads a prompt only from -p/--print, -i/--prompt-interactive, or stdin. Supervised form is `agy --dangerously-skip-permissions -i "<brief>"`. Autonomy is --dangerously-skip-permissions (footer shows accept-edits).
2. Detection: ANTIGRAVITY_AGENT=1 on child/tool processes; sets neither CLAUDECODE nor GROK_AGENT. The marker is checked ahead of CLAUDECODE because agy is not known to clear an inherited one. Process name is exactly `agy`, matched ANCHORED, never globbed, because ordinary words such as `magyar` and `stagy` contain the substring.
3. Turn end - the most important rule in this change. agy pushes long work into the background after about ten seconds, ENDS THE TURN saying it will wait, then wakes itself and ends a SECOND turn once the work lands. ONLY a Stop payload with fullyIdle:true counts as a turn end. Measured on a real 90-second command: a fullyIdle:false Stop arrived with workspacePaths ALREADY POPULATED, 81 seconds before the true end (reproduced three times). A grok-shaped hook keyed on the workspace alone would have reported a validation run finished while it was still running, closing a ship task mid-pipeline. protojson omits default values, so a false fullyIdle can arrive as an ABSENT key; the gate tests `select(.fullyIdle == true)` and never jq's `//`, which treats false as null. Covered by a portable regression carrying the real captured payloads; removing the gate makes that test fail by name.
4. Hook installation differs from grok's for one measured reason: agy's global hooks live in ONE shared hooks.json keyed by hook NAME, which is the operator's own file. Firstmate installs a PLUGIN instead (plugins/fm-turn-end/), a firstmate-owned directory agy discovers and enables itself; verified live that agy's config.json is left byte-identical. Everything else is grok's guarded pattern: a .fm-agy-turnend worktree pointer, a private registry token, a hook that is a silent no-op for every non-firstmate agy session, and teardown that retires the per-task entries. It never edits project configuration nor agy's trust store. jq is a hard requirement of the installed hook and the spawn refuses naming it, the same contract as the Kimi installer.
5. Effort: --effort low|medium|high; xhigh and max are omitted rather than passed. agy validates that value inside its TUI rather than at flag-parse time. Model: --model <id>; discovery via `agy models`, which answers without opening a session.
6. Lifecycle: interrupt is a single Escape and leaves the composer EMPTY, so no clear key follows; exit is `/exit` plus Enter, which prints `Resume with -c (or command below): agy --conversation=<id>` and returns the pane to the shell; resume is `agy --conversation=<id>` or -c.
7. Workspace trust: outside trustedWorkspaces agy asks "Do you trust the contents of this project?" with Yes preselected and one Enter resolves it. --dangerously-skip-permissions does NOT suppress it, and a session blocked on that dialog still fires Stop with fullyIdle:false.

TWO DELIBERATE DEVIATIONS FROM THE ORIGINAL TASK, both forced by measurement and both intentional, not oversights:
A. The task asked for a grok-shaped rendered busy arm using `esc to cancel` in fm_busy_classify. Measurement contradicted that premise: while agy waits on backgrounded work it renders the SAME idle footer (`? for shortcuts`) as when the turn is genuinely over, and `esc to cancel` appears only while tokens stream. A grok-shaped arm would report a working agent as idle for minutes, which is a false idle - the one verdict the busy contract never permits. agy therefore ships with NO worker-state source and classifies `unknown agy-unverified`. Its Stop hook cannot be a semantic writer either, because the fullyIdle:false events that must be ignored are exactly what would otherwise settle an interrupted turn's record, and PreInvocation fires once per model invocation (measured ten across two turns), not once per turn. `esc to cancel` IS registered, but only as a DELIVERY acknowledgement in fm-composer-lib.sh, a positive-only signal whose worst case is one extra submit retry. On Herdr a genuinely streaming turn is still proven busy by the native arm, which runs first, so the oscillating native `done` can never produce a false idle.
B. The task said bin/backends/tmux.sh mattered only for a second mate. It does not: the same classifier backs crewmate liveness and fm-control's exit and relaunch postcondition proofs, so the anchored agy entry is required and was implemented rather than declared out of scope.

DELIBERATE BOUNDARIES, declared as boundaries and not omissions:
(a) Everything was measured on Herdr. tmux is UNVERIFIED because tmux is not installed on this machine; only the kernel-level process name was confirmed, directly via `ps -o comm=`, which is backend-independent. No tmux composer, footer, or capture behaviour was checked.
(b) quota-axi marks agy's windows unresolved_windows, so agy has NO spendPriority and cannot enter a quota-balanced profile array; for now it is selectable only as an explicit per-task choice. Resolving that belongs to quota-axi and is deliberately not attempted here.
(c) Scope is crewmate and scout ONLY, never primary and never second mate. fm-spawn refuses a --secondmate agy launch and the control plane refuses that kind before stopping anything. agy is deliberately absent from bin/fm-session-lock-lib.sh, which is the primary-session identity surface; adding it there would imply a primary supervision protocol that does not exist.
Also untested and declared: husk recovery, relaunch, and durable steering-inbox delivery; one conversation and one model (Gemini 3.7 Flash), so the Claude and GPT-OSS models inside agy were not exercised.

ALSO INCLUDED, deliberately: a pre-existing macOS failure in tests/fm-muse-harness.test.sh is fixed. It faked a named process by copying bash, which macOS SIGKILLs because the copy still claims Identifier=com.apple.bash but no longer validates. Both suites now share tests/lib.sh's fm_test_named_interpreter, which PROVES a construction works on the host rather than assuming one, because procps on Linux reports the kernel exec name while macOS reports argv[0]. A follow-up commit resolves that construction once per host and caches it in a file, because callers capture the path with command substitution so a shell variable would never reach them, and on macOS every doomed copy attempt pays a full code-signature evaluation.

PR BODY REQUIREMENT: the PR body must explain, for a reader who saw none of this work, why the fullyIdle rule exists and why agy ends up with no rendered busy classifier despite Herdr's native state being unreliable for it. It must also name two PRE-EXISTING test failures that are NOT introduced by this change and are NOT fixed here: tests/fm-composer-lib.test.sh "a half-block rule row must count as a structural edge" and tests/fm-wake-queue.test.sh "a subshell reclaimed its parent's live hold". Both reproduce on a clean clone of main and both come from macOS shipping bash 3.2, which lacks \uXXXX escapes (added 4.2) and BASHPID (added 4.0); CI runs bash 5 on Linux where they pass. They are left out of this delivery because they are unrelated subsystems and the fix pattern spans six test files.

RESUMPTION NOTE: this is a resume after the review step hit its 30-minute per-agent timeout (agent killed with signal 143 while still producing output), not a fresh start. FOUR review fix rounds are ALREADY APPLIED and are present in the commits being validated: (1) fix agy secondmate parsing and atomic hook publication, (2) answer the agy trust dialog and accept Linux comm truncation, (3) require positive agy readiness and disclose the plugin install, (4) fold the agy readiness read and clear the agy marker in sibling detection suites. Those correspond to review findings already decided and applied: agy-secondmate-positional-bypass, agy-global-hook-install-race, agy-workspace-trust-dialog-unanswered, named-interpreter-proof-fails-on-linux-comm-truncation, agy-ready-gate-proves-absence-not-readiness, agy-global-plugin-write-not-in-operator-config-doc, agy-live-guard-proves-refusal-by-attempting-it, antigravity-marker-not-cleared-in-sibling-detection-suites, agy-ready-gate-unreachable-footer-branch-on-a-scrollback-capture, and stale-harness-enumerations-in-operator-docs. Do not re-derive them from scratch. The branch was also rebased onto a much newer main during that run. TWO findings were consciously ACCEPTED rather than fixed and must be named as such in the PR body: named-interpreter-cache-in-shared-tmp (test-only, predictable cache path in shared tmp, requires a local attacker to guess the PID) and crew-dispatch-validation-calls-agy-unverified (crew_dispatch_validate deliberately omits agy, which is consistent with the declared boundary that agy has no spendPriority and is selectable only as an explicit per-task choice).

ACCEPTED DECISION STILL TO LAND: the agy turn-end registry root must be FIXED with no environment override. An earlier iteration added FM_AGY_CONFIG_HOME as a supported override; that was reversed on purpose. Removing it makes spawn and teardown agree on the default root BY CONSTRUCTION, so the documented retirement guarantee cannot be broken by a teardown process whose environment differs from the spawn's. The smaller alternative that satisfies the accepted contract without expanding it was preferred over recording new durable per-task state. This deliberately leaves agy different from grok, which does support GROK_HOME; that difference is intentional and the PR body should say so, noting that a maintainer who later wants the override must add it together with the durable record it requires. Adjust or remove any test that exercises that override.

SIZE NOTE for the PR body: this change is about 36 files and roughly 1979 insertions against current origin/main. Any larger figure includes upstream work that arrived through the rebase and is not part of this proposal.

## What Changed

- Adds `agy` across every owner a dispatchable harness needs under `bin/`: the detection marker `ANTIGRAVITY_AGENT=1` ordered ahead of `CLAUDECODE` plus an anchored `agy` process-name match in `fm-harness.sh`; the launch shape in `fm-spawn.sh` (`agy --dangerously-skip-permissions -i "<brief>"`, since agy rejects a positional prompt, `--effort` limited to low/medium/high, `jq` refused early as a hard requirement, the workspace-trust dialog answered with one Enter, and a `--secondmate` launch refused); a firstmate-owned global plugin installed at `~/.gemini/config/plugins/fm-turn-end/` with a per-task `.fm-agy-turnend` pointer; teardown folding the grok, kimi and agy registry retirements into one `remove_turnend_auth`; the delivery signature `esc to cancel` in `fm-composer-lib.sh`; an explicit `unknown agy-unverified` arm in `fm-busy-lib.sh`; interrupt, exit and resume mechanics in `fm-control-lib.sh`; and an anchored liveness entry in `bin/backends/tmux.sh`.
- Documents the adapter in the harness-adapters skill reference and its routing map, a dated verification record at `docs/verification/agy.md`, `docs/configuration.md`, `AGENTS.md` section 4, and the documentation-audiences map.
- Adds a portable regression suite carrying the real captured Stop payloads plus a live surface guard, and extracts a shared `fm_test_named_interpreter` helper into `tests/lib.sh` that proves the naming construction on the host and caches the winning method per host. That helper also fixes a pre-existing macOS failure in `tests/fm-muse-harness.test.sh`, which faked a named process by copying bash and was SIGKILLed because the copy still claimed Apple's identifier. Sibling detection suites now clear `ANTIGRAVITY_AGENT`.

## Why fullyIdle:true is the only turn end

agy pushes long work into the background after about ten seconds, ends the turn saying it will wait, then wakes itself and ends a second turn once the work lands. Only a Stop payload carrying `fullyIdle:true` is a real turn end. Measured on a real 90-second command, a `fullyIdle:false` Stop arrived with its workspace paths already populated, 81 seconds before the true end, and that reproduced three times. A grok-shaped hook keyed on the workspace alone would have reported a validation run finished while it was still running, closing a ship task mid-pipeline. Because protojson omits default values, a false `fullyIdle` can arrive as an absent key, so the gate tests `select(.fullyIdle == true)` and never uses jq's `//`, which treats false as null. Removing the gate makes the regression fail by name.

## Why agy has no rendered busy classifier

Herdr's native state oscillates for agy and cannot be trusted for idle, but it runs first in the classifier and still positively proves a streaming turn busy. Nothing else can. agy renders the same idle footer (`? for shortcuts`) while waiting on backgrounded work as it does when the turn is genuinely over; the busy footer appears only while tokens stream. A grok-shaped regex arm would therefore report a working agent as idle for minutes, which is a false idle, the one verdict the busy contract never permits. The Stop hook cannot be a state writer either, because the `fullyIdle:false` events it must ignore are exactly what would otherwise settle an interrupted turn's record, and `PreInvocation` fires once per model invocation rather than once per turn (ten measured across two turns). agy therefore ships with no worker-state source and classifies `unknown agy-unverified`. The `esc to cancel` match is registered only as a delivery acknowledgement, a positive-only signal whose worst case is one extra submit retry.

## Deliberate differences and boundaries

- The turn-end registry root is fixed with no environment override, deliberately unlike grok's `GROK_HOME`. That makes spawn and teardown agree on the root by construction, so the documented retirement guarantee cannot be broken by a teardown whose environment differs from the spawn's. A maintainer who later wants the override must add it together with the durable per-task record it requires.
- The `tmux.sh` entry was implemented rather than declared out of scope. The same classifier backs crewmate liveness and fm-control's exit and relaunch postconditions, so it is not secondmate-only.
- Scope is crewmate and scout only, never primary and never second mate, and agy is deliberately absent from the primary-session identity surface. Verification was done on Herdr; tmux is unverified beyond the kernel-level process name. agy has no `spendPriority`, so it is selectable only as an explicit per-task choice, and that restriction is enforced rather than advisory.
- Two review findings were consciously accepted rather than fixed. The named-interpreter cache lives at a predictable path in shared tmp, which is test-only and requires a local attacker to guess the PID. `crew_dispatch_validate` deliberately omits agy, which is consistent with the declared boundary that agy carries no spendPriority.

## Pre-existing failures not addressed here

Two test failures reproduce on a clean clone of `main` and are neither introduced nor fixed by this branch: `tests/fm-composer-lib.test.sh` "a half-block rule row must count as a structural edge" and `tests/fm-wake-queue.test.sh` "a subshell reclaimed its parent's live hold". Both come from macOS shipping bash 3.2, which lacks `\uXXXX` escapes and `BASHPID`; CI runs bash 5 on Linux where they pass. They are left out because they touch unrelated subsystems and the fix pattern spans six test files.

Size: 36 files, 1,965 insertions against the base commit.

## Risk Assessment

✅ Low: The adapter is additive and fenced behind explicit per-task selection, every enumeration owner and doc owner landed together, the two findings from the previous round were fixed minimally with no scope creep, and the only remaining item is a doc-only inconsistency in one sentence.

## Testing

Ran the targeted agy suite, the two touched sibling suites, and the muse/fixtures suites on macOS arm64 with bash 3.2, all green, then produced product-level transcripts for the claims that unit passes alone cannot show. The live surface guard ran against the installed agy 1.1.24, the exact version the intent measured, and confirmed the positional-prompt refusal, the -i flag, the low|medium|high effort vocabulary, session-free model discovery, and the customization root; its FM_AGY_TURNEND_LIVE_E2E sub-check was skipped because it requires a freshly captured Stop payload from a real turn. No live agy session was spawned at all, deliberately: that would spend the operator's quota, need a Herdr session, and risk the blind Enter this change exists to avoid. Instead the fullyIdle rule was proven against the hook the spawn actually installs, showing the turn-end signal appear only for fullyIdle=true and not for a false, an absent key, or a trust-blocked Stop, plus the hook returning at 5 seconds with the writer still holding stdin open and teardown retiring the registry entry. There is no screenshot because the only rendered surface here is agy's own TUI trust dialog and reaching it means a live session; the pane text the launch poll reads and the resulting Enter decision are captured instead, showing zero trust Enters on an already-trusted workspace and on a row shape the anchor does not match, and exactly one when the dialog is on screen. The tmux backend stays unverified per the declared boundary since tmux is not installed here, and the two pre-existing failures the intent names were not run. The included macOS muse fix was demonstrated at root cause: a copied /bin/bash exits 137 under code-signing while the symlink construction reports the requested process name. The worktree is clean and all evidence lives in the run's evidence directory.

<details>
<summary>Evidence: fullyIdle turn-end rule driven through the installed agy Stop hook</summary>

Source: [fullyIdle turn-end rule driven through the installed agy Stop hook](https://github.com/tpeixinho/firstmate/blob/3e97d9d4f69720284757715063d3f14c71b1ca30/.no-mistakes/evidence/fm/fm-agy-adapter/agy-turnend-fullyidle.txt)

<code>== what the agy spawn installed under agy&#39;s global customization root ==&#10;fm-turn-end.d/fm.fongoEvGG2t8&#10;plugins/fm-turn-end/fm-turn-end.sh&#10;plugins/fm-turn-end/hooks.json&#10;plugins/fm-turn-end/plugin.json&#10;&#10;== feeding the installed hook the four captured Stop payloads ==&#10;fullyIdle=false, workspace POPULATED -&gt; no turn end (state/&lt;id&gt;.turn-ended absent)&#10;fullyIdle key ABSENT (protojson default) -&gt; no turn end (state/&lt;id&gt;.turn-ended absent)&#10;blocked on trust dialog, no workspace -&gt; no turn end (state/&lt;id&gt;.turn-ended absent)&#10;fullyIdle=true &lt;- the real turn end -&gt; TURN END SIGNALLED (state/&lt;id&gt;.turn-ended exists)&#10;&#10;== the same hook is a silent no-op for a NON-firstmate agy session ==&#10;foreign workspace, fullyIdle=true -&gt; exit 0, turn-ended not written&#10;&#10;== bounded read: writer holds stdin open and sends nothing ==&#10;hook returned after 5s with the writer still alive: yes (agy kills its hooks at 10s)&#10;&#10;== teardown retires the per-task wiring ==&#10;before: registry entry fm.fongoEvGG2t8, state token 1&#10;registry entries left for this task: 0</code>

```text
== what the agy spawn installed under agy's global customization root ==
  fm-turn-end.d/fm.fongoEvGG2t8
  plugins/fm-turn-end/fm-turn-end.sh
  plugins/fm-turn-end/hooks.json
  plugins/fm-turn-end/plugin.json

== the firstmate worktree pointer the hook reads ==
  /private/var/folders/3r/6cq6fg_x461_vvfymrdp2npw0000gn/T/fm-agy-harness.hCIiaW/turnend-evidence/wt/.fm-agy-turnend -> token=fm.fongoEvGG2t8

== feeding the installed hook the four captured Stop payloads ==
  fullyIdle=false, workspace POPULATED           -> no turn end            (state/agy-turnend-evidence-x1.turn-ended absent)
  fullyIdle key ABSENT (protojson default)       -> no turn end            (state/agy-turnend-evidence-x1.turn-ended absent)
  blocked on trust dialog, no workspace          -> no turn end            (state/agy-turnend-evidence-x1.turn-ended absent)
  fullyIdle=true  <- the real turn end           -> TURN END SIGNALLED  (state/agy-turnend-evidence-x1.turn-ended exists)

== the same hook is a silent no-op for a NON-firstmate agy session ==
  foreign workspace, fullyIdle=true          -> exit 0, turn-ended not written

== bounded read: writer holds stdin open and sends nothing ==
  hook returned after 5s with the writer still alive: yes (agy kills its hooks at 10s)

== teardown retires the per-task wiring ==
  before: registry entry fm.fongoEvGG2t8, state token 1
  registry entries left for this task: 0
```
</details>
<details>
<summary>Evidence: agy launch command line and the crewmate/scout scope refusals</summary>

Source: [agy launch command line and the crewmate/scout scope refusals](https://github.com/tpeixinho/firstmate/blob/3e97d9d4f69720284757715063d3f14c71b1ca30/.no-mistakes/evidence/fm/fm-agy-adapter/agy-scope-and-launch-shape.txt)

<code>== the command line an agy crewmate is actually launched with ==&#10;env -u CURSOR_AGENT -u CURSOR_INVOKED_AS env -u CLAUDECODE -u PI_CODING_AGENT -u GROK_AGENT -u FM_PI_HARNESS agy --dangerously-skip-permissions&#10;--effort &#39;high&#39; -i &#34;$(&#39;.../bin/fm-operational-input.sh&#39; encode launch-brief &lt; &#39;.../brief.md&#39;)&#34;&#10;&#10;== --effort xhigh is OMITTED, never passed (agy accepts only low|medium|high) ==&#10;no --effort in the launch line&#10;&#10;== scope: agy refuses a secondmate spawn ==&#10;error: agy is a verified crewmate/scout adapter only and cannot run a secondmate; it has no primary supervision protocol. Select a harness verified for secondmates.&#10;&#10;== scope: agy is not implicitly selectable from config/crew-harness ==&#10;error: agy is selectable only as an explicit per-task choice; pass --harness agy for this crewmate or scout&#10;global hook installed by that refused spawn: 0&#10;&#10;== scope: an implicitly resolved agy secondmate names the secondmate boundary ==&#10;error: agy is a verified crewmate/scout adapter only and cannot run a secondmate; it has no primary supervision protocol. Select a harness verified for secondmates.</code>

```text
== the command line an agy crewmate is actually launched with ==
  env -u CURSOR_AGENT -u CURSOR_INVOKED_AS env -u CLAUDECODE -u PI_CODING_AGENT -u GROK_AGENT -u FM_PI_HARNESS agy --dangerously-skip-permissions 
  --effort 'high' -i "$('~/.no-mistakes/worktrees/bae792236fae/01M1H7RWFXBK724G6XKAHM75BM/bin/fm-operational-input.sh' encode 
  launch-brief < 
  '/private/var/folders/3r/6cq6fg_x461_vvfymrdp2npw0000gn/T/fm-agy-harness.LicqGF/shape-evidence/home/data/agy-shape-evidence-x1/brief.md')"

== --effort xhigh is OMITTED, never passed (agy accepts only low|medium|high) ==
  no --effort in the launch line

== scope: agy refuses a secondmate spawn ==
  error: agy is a verified crewmate/scout adapter only and cannot run a secondmate; it has no primary supervision protocol. Select a harness verified for secondmates.

== scope: agy is not implicitly selectable from config/crew-harness ==
  error: agy is selectable only as an explicit per-task choice; pass --harness agy for this crewmate or scout
  global hook installed by that refused spawn: 0

== scope: an implicitly resolved agy secondmate names the secondmate boundary ==
  error: agy is a verified crewmate/scout adapter only and cannot run a secondmate; it has no primary supervision protocol. Select a harness verified for secondmates.
```
</details>
<details>
<summary>Evidence: Live vendor surface guard against the installed agy 1.1.24</summary>

Source: [Live vendor surface guard against the installed agy 1.1.24](https://github.com/tpeixinho/firstmate/blob/3e97d9d4f69720284757715063d3f14c71b1ca30/.no-mistakes/evidence/fm/fm-agy-adapter/agy-live-surface-guard.txt)

<code># agy 1.1.24 (~/.local/bin/agy)&#10;ok - agy 1.1.24 rejects a positional prompt and still offers -i&#10;ok - agy 1.1.24 still advertises effort low|medium|high and a model flag&#10;ok - agy 1.1.24 lists models without opening a session&#10;ok - agy 1.1.24&#39;s global customization root is where the turn-end installer writes&#10;# skip: set FM_AGY_TURNEND_LIVE_E2E=1 with FM_AGY_TURNEND_PAYLOAD to check a captured Stop payload&#39;s shape&#10;all fm-agy-surface-live-e2e checks passed against agy 1.1.24</code>

```text
# agy 1.1.24 (~/.local/bin/agy)
ok - agy 1.1.24 rejects a positional prompt and still offers -i
ok - agy 1.1.24 still advertises effort low|medium|high and a model flag
ok - agy 1.1.24 lists models without opening a session
ok - agy 1.1.24's global customization root is where the turn-end installer writes
# skip: set FM_AGY_TURNEND_LIVE_E2E=1 with FM_AGY_TURNEND_PAYLOAD to check a captured Stop payload's shape
all fm-agy-surface-live-e2e checks passed against agy 1.1.24
```
</details>
<details>
<summary>Evidence: Trust-dialog pane renders and the Enter the launch poll delivers for each</summary>

Source: [Trust-dialog pane renders and the Enter the launch poll delivers for each](https://github.com/tpeixinho/firstmate/blob/3e97d9d4f69720284757715063d3f14c71b1ca30/.no-mistakes/evidence/fm/fm-agy-adapter/agy-trust-dialog-pane.txt)

```text
== workspace agy ALREADY TRUSTS - no dialog, baseline ==
  what the pane renders after the launch line is submitted:
    |   >
    | 
    | ? for shortcuts
  Enters seen after the launch line was typed: 1  (the first is the launch submit itself)
  => trust Enters delivered: 0
  spawn outcome: completed with no failure recorded

== workspace NOT trusted - dialog on screen ==
  what the pane renders after the launch line is submitted:
    | agy 1.1.24
    | 
    | Do you trust the contents of this project?
    | > Yes, I trust this folder
    |   No, browse in restricted mode
  Enters seen after the launch line was typed: 2  (the first is the launch submit itself)
  => trust Enters delivered: 1
  spawn outcome: completed with no failure recorded

== dialog above a bottom-anchored footer in a tall frame ==
  what the pane renders after the launch line is submitted:
    | agy 1.1.24
    | 
    | Do you trust the contents of this project?
    | > Yes, I trust this folder
    |   No, browse in restricted mode
    | 
    |   >
    | ? for shortcuts
  Enters seen after the launch line was typed: 2  (the first is the launch submit itself)
  => trust Enters delivered: 1
  spawn outcome: completed with no failure recorded

== a row shape the anchor does not match - left unanswered on purpose ==
  what the pane renders after the launch line is submitted:
    | agy 1.1.24
    | 
    | Do you trust the contents of this project?
    | > Yes, I trust this folder   (recommended)   enter to confirm
    |   No, browse in restricted mode   esc to go back
  Enters seen after the launch line was typed: 1  (the first is the launch submit itself)
  => trust Enters delivered: 0
  spawn outcome: completed with no failure recorded

== pane renders nothing at all - spawn still completes ==
  what the pane renders after the launch line is submitted:
  Enters seen after the launch line was typed: 1  (the first is the launch submit itself)
  => trust Enters delivered: 0
  spawn outcome: completed with no failure recorded
```
</details>
<details>
<summary>Evidence: Trust anchor checked against the counter-examples the verification record cites</summary>

Source: [Trust anchor checked against the counter-examples the verification record cites](https://github.com/tpeixinho/firstmate/blob/3e97d9d4f69720284757715063d3f14c71b1ca30/.no-mistakes/evidence/fm/fm-agy-adapter/agy-trust-anchor-counterexamples.txt)

<code># anchor under test (bin/fm-spawn.sh:2488): ^[^A-Za-z]*Yes, I trust this folder[^A-Za-z]*$&#10;&#10;## the row observed live in agy 1.1.24&#10;[&gt; Yes, I trust this folder] MATCHES (would draw the trust Enter)&#10;&#10;## the three counter-examples docs/verification/agy.md records verbatim as MATCHING&#10;[- Yes, I trust this folder] MATCHES (would draw the trust Enter)&#10;[1. Yes, I trust this folder] MATCHES (would draw the trust Enter)&#10;[- &#34;Yes, I trust this folder&#34;] MATCHES (would draw the trust Enter)&#10;&#10;## the only form the anchor really excludes: letters around the label&#10;[ the row reads Yes, I trust this folder here] no match (dialog left unanswered)&#10;&#10;## a row shape with a trailing hint the anchor does not match&#10;[&gt; Yes, I trust this folder (enter)x] no match (dialog left unanswered)</code>

```text
# anchor under test (bin/fm-spawn.sh:2488): ^[^A-Za-z]*Yes, I trust this folder[^A-Za-z]*$
# each row below is checked with the same predicate the launch poll uses

## the row observed live in agy 1.1.24
  [> Yes, I trust this folder]                   MATCHES (would draw the trust Enter)

## the three counter-examples docs/verification/agy.md records verbatim as MATCHING
  [- Yes, I trust this folder]                   MATCHES (would draw the trust Enter)
  [1. Yes, I trust this folder]                  MATCHES (would draw the trust Enter)
  [- "Yes, I trust this folder"]                 MATCHES (would draw the trust Enter)

## the only form the anchor really excludes: letters around the label
  [  the row reads Yes, I trust this folder here] no match (dialog left unanswered)

## a row shape with a trailing hint the anchor does not match
  [> Yes, I trust this folder   (enter)x]        no match (dialog left unanswered)
```
</details>
<details>
<summary>Evidence: macOS root cause of the included muse fix: copied bash is SIGKILLed, symlink works</summary>

Source: [macOS root cause of the included muse fix: copied bash is SIGKILLed, symlink works](https://github.com/tpeixinho/firstmate/blob/3e97d9d4f69720284757715063d3f14c71b1ca30/.no-mistakes/evidence/fm/fm-agy-adapter/named-interpreter-macos-construction.txt)

<code># host: Darwin arm64 / GNU bash, version 3.2.57(1)-release (arm64-apple-darwin26)&#10;&#10;## OLD construction (the one tests/fm-muse-harness.test.sh used): copy /bin/bash&#10;bash: line 9: 68111 Killed: 9 &#34;$D/muse-bin-0.9.9&#34; -c ...&#10;exit status: 137 (128 + 9, SIGKILL: the copy still claims Identifier=com.apple.bash and no longer validates)&#10;&#10;## NEW construction, resolved by tests/lib.sh fm_test_named_interpreter&#10;built: &lt;tmp&gt;/muse-bin-0.9.9 -&gt; /bin/bash&#10;ps -o comm= reports: &lt;tmp&gt;/muse-bin-0.9.9&#10;cached method for this host: symlink</code>

```text
# host: Darwin arm64 / GNU bash, version 3.2.57(1)-release (arm64-apple-darwin26)
# real bash: /bin/bash

## OLD construction (the one tests/fm-muse-harness.test.sh used): copy /bin/bash
bash: line 9: 68111 Killed: 9               "$D/muse-bin-0.9.9" -c "p=\$(ps -o comm= -p \$\$); printf %s\\\\n \"\$p\""
exit status: 137  (137 = 128 + 9, SIGKILL: the copy still claims Identifier=com.apple.bash and no longer validates)

## NEW construction, resolved by tests/lib.sh fm_test_named_interpreter
built:   <tmp>/muse-bin-0.9.9
lrwxr-xr-x@ 1 tpeixinho  staff  9 Sep  2 18:56 <tmp>/muse-bin-0.9.9 -> /bin/bash
ps -o comm= reports: /var/folders/3r/6cq6fg_x461_vvfymrdp2npw0000gn/T//fm-namedinterp.9FsBQR/muse-bin-0.9.9
cached method for this host: symlink
```
</details>
<details>
<summary>Evidence: muse and fixtures suites green on macOS bash 3.2</summary>

Source: [muse and fixtures suites green on macOS bash 3.2](https://github.com/tpeixinho/firstmate/blob/3e97d9d4f69720284757715063d3f14c71b1ca30/.no-mistakes/evidence/fm/fm-agy-adapter/muse-named-interpreter-macos.txt)

```text
\### host: Darwin arm64, GNU bash, version 3.2.57(1)-release (arm64-apple-darwin26)

\### tests/fm-test-fixtures.test.sh
ok - fake no-mistakes --version is the shared constant and overridable
ok - init/doctor no-mistakes stub touches markers and refuses other verbs
ok - fake gh authenticates and fake gh-axi reports the shared version
ok - spawn fakebin answers pane path, logs -l payloads, and installs extra tools
ok - send stubs log typed text and fake ssh records argv with a controllable exit
ok - spawn-home layout writes harness pin, beat, and brief
ok - the named-interpreter proof accepts a platform truncation but not a short report

\### tests/fm-muse-harness.test.sh
ok - muse is detected through any versioned muse-bin ancestor
ok - muse detection does not claim unrelated muse-containing commands
ok - muse launch clears foreign harness markers before ancestry detection
ok - muse spawn launches with autonomy, privacy control, and a positional brief
ok - muse maps the shared effort vocabulary and reaches ultra only via explicit max
ok - muse spawn refuses when no credential can reach the provider
ok - muse spawn refuses a META_API_KEY that cannot reach the worker
ok - muse spawn accepts a stored credential without META_API_KEY
ok - muse resolves relative XDG roots before preflight and launch
ok - muse is refused as a secondmate harness
ok - muse spawn writes a session binding that teardown removes
ok - every accepted muse Escape alias clears the restored composer
ok - the composer clear is scoped to muse and does not touch other adapters
ok - a failed muse composer clear fails loudly instead of leaving stale input
ok - the run fold tracks open, settled, interrupted, reopened, and run-free logs
ok - a nested terminal record never settles an in-flight run
ok - the session binding folds only the log matching this task's worktree
ok - workspace bindings compare decoded paths literally
ok - spawn-time exclusions select the current session across equal mtimes
ok - the Muse session cache avoids rescans and refreshes safely across incarnations
ok - a changed Muse namespace revalidates cached session uniqueness
ok - sub-agent session logs are excluded from the parent's busy fold
ok - every unproven muse binding classifies unknown rather than idle
ok - a settled session log reads idle for both completed and interrupted turns
ok - muse trusts no busy record source
```
</details>
<details>
<summary>Evidence: agy portable regression suite output</summary>

Source: [agy portable regression suite output](https://github.com/tpeixinho/firstmate/blob/3e97d9d4f69720284757715063d3f14c71b1ca30/.no-mistakes/evidence/fm/fm-agy-adapter/agy-portable-regression.txt)

```text
ok - agy is detected by its marker, ahead of an inherited claude marker
ok - agy is detected through its process ancestry
ok - agy detection does not claim unrelated agy-containing commands
ok - the tmux classifier recognizes agy and stays anchored
ok - agy launches with -i and skip-permissions, never a positional prompt
ok - agy can only be selected explicitly for one crewmate or scout
ok - an implicitly resolved agy secondmate is refused for being a secondmate
ok - agy effort is clamped to the levels agy accepts
ok - agy refuses a secondmate spawn and says why
ok - positional agy refuses a secondmate spawn and says why
ok - agy's workspace-trust dialog is answered with exactly one Enter, and only when it is on screen
ok - a trust dialog above the footer is answered rather than read past
ok - a row shape the anchor does not match is left unanswered rather than answered blind
ok - the answer label inside a sentence never draws the trust Enter
ok - a pane that never renders completes the spawn without an unwarranted Enter
ok - only a fullyIdle=true Stop counts as an agy turn end
ok - the bounded hook preserves a complete multi-line payload and returns on a wall clock
ok - the agy global hook requires a firstmate registry token
ok - the agy hook always lets the stop proceed, on every payload shape
ok - an agy spawn refuses without jq before it drives a pane, and installs nothing
ok - teardown retires the agy per-task turn-end wiring
ok - agy's busy signature is registered and collides with no other adapter
ok - agy worker state is unknown with a reason, never idle
ok - agy's lifecycle table matches its verified mechanics
ok - agy's wiring and registry paths are complete and refuse a bad token
all fm-agy-harness tests passed
```
</details>
<details>
<summary>Evidence: Touched sibling suites output</summary>

Source: [Touched sibling suites output](https://github.com/tpeixinho/firstmate/blob/3e97d9d4f69720284757715063d3f14c71b1ca30/.no-mistakes/evidence/fm/fm-agy-adapter/sibling-suites.txt)

```text
\### tests/fm-spawn-dispatch-profile.test.sh
ok - no --model/--effort records defaults and types the claude launch instructions
ok - non-cursor launches clear inherited Cursor identity markers
ok - a non-agy launch clears an inherited ANTIGRAVITY_AGENT so the child keeps its own identity
ok - relative home overrides ignore CDPATH and become absolute before spawn launch construction
ok - FM_HOME defaults resolve relative paths and preserve absolute spellings
ok - absolute override spellings are preserved in spawn launch paths
ok - unresolvable relative spawn overrides fail with named diagnostics
ok - active crew-dispatch profile requires an explicit harness for ship spawns
ok - active crew-dispatch profile requires an explicit harness for scout spawns
ok - active crew-dispatch profile allows an explicit resolved harness
ok - active crew-dispatch profile allows the legacy positional harness form
ok - active crew-dispatch profile allows the raw launch-command escape hatch
ok - claude receives --model and --effort profile flags
ok - codex receives --model and model_reasoning_effort profile flags
ok - codex omits unsupported max effort instead of passing a bad config value
ok - grok receives --model and --reasoning-effort profile flags
ok - grok omits unsupported max reasoning effort
ok - grok omits unsupported xhigh reasoning effort
ok - cursor receives its model-qualified reasoning class and exact task workspace
ok - cursor refuses model ids absent from its resolved binary's live catalog
ok - cursor preserves the requested model when its live catalog is unreachable
ok - opencode receives --model and omits the unsupported effort axis
ok - pi receives --model and --thinking max profile flags
ok - Pi launch probing omits --tui-mode on older Pi and preserves it on supporting Pi
ok - pi-signed shares Pi launch semantics while preserving its configured and recorded identity
ok - pi-signed refuses safely and actionably when the selected executable is unavailable
ok - pi-signed is a distinct persistent secondmate runtime with shared Pi supervision semantics
ok - batch dispatch forwards shared --harness, --model, and --effort to every pair
ok - claude forwards firstmate's CLAUDE_CONFIG_DIR so the crewmate uses the same credential store
ok - claude omits the config-dir prefix when firstmate runs with the single-store default
ok - non-claude harnesses do not receive the claude CLAUDE_CONFIG_DIR prefix
ok - active crew-dispatch profile does not block secondmate launches
# all fm-spawn-dispatch-profile tests passed

\### tests/fm-secondmate-harness.test.sh
ok - A1 fm-harness.sh secondmate resolves the fallback chain; crew mode unchanged
ok - fm-harness detects only Cursor Agent CLI's exact invocation marker
ok - C1 fm-harness.sh secondmate-model/secondmate-effort resolve the optional tokens; bare harness stays empty (backward-compat)
ok - pi-signed identity: authoritative launch selection distinguishes shared wrapper ancestry
ok - harness identity: dash-leading ps command names are basename operands, not options
ok - B1 propagate_inheritable_config: copy, idempotence, convergence, absence-mirror, exclusion, no-op, skip diagnostics
ok - B2 spawn: secondmate runs the secondmate harness; its home inherits declared config
ok - B3 spawn: an absent secondmate-harness falls back to the crew harness (backward-compat)
ok - B4 spawn: no config at all -> own harness and no propagation side effects
ok - B5 spawn: an explicit per-spawn harness arg overrides config/secondmate-harness
ok - B6 spawn: an unverified resolved secondmate harness is refused (guard intact)
skip: cursor executable not resolvable in this environment, so the launch could not be built
ok - B5b spawn: FM_BACKEND wins over inherited config/backend
ok - B5c spawn: explicit --backend wins over FM_BACKEND and inherited config/backend
ok - C2 spawn: a bare harness-only secondmate-harness file launches with no model/effort flag (backward-compat)
ok - C3 spawn: config/secondmate-harness's model token threads --model into the launch and meta
ok - C4 spawn: config/secondmate-harness's model+effort tokens thread into the launch and meta
ok - C5 spawn: an explicit --model overrides config/secondmate-harness's model token; the file's effort token still applies
ok - C6 spawn: an explicit --effort overrides config/secondmate-harness's effort token; the file's model token still applies
ok - C7 spawn: an explicit --harness starts with clean model/effort defaults
ok - C8 spawn: an explicit --harness still honors explicit model/effort flags
ok - C9 spawn: secondmate launch pins supervision to its own harness
ok - C9 spawn: the harness fallback chain still resolves with no tokens; crew/scout launches are unaffected by this feature
ok - B7 bootstrap sweep pushes, re-converges, and mirrors absence; never inherits secondmate-harness
ok - B8 bootstrap sweep propagates config even when the home's tracked files are already current
ok - B9 bootstrap sweep defers new inherited config until the home ignores it
ok - B10 bootstrap sweep materializes and inherits the startup-memory default while fast-forwarding
ok - B12b backend inheritance: present values and primary absence converge exactly
ok - B12c presentation inheritance: the primary default converges on, and only an explicit opt-out propagates off
ok - B11 bootstrap sweep surfaces config propagation failures
ok - B11 bootstrap rereads completed config writes after partial propagation
ok - B12 config-push propagates via shared live discovery, reports items, rereads on change only, and does not fast-forward
ok - B13 config-push reports dirty, non-allowing, and invalid homes without failing warnings-only runs
ok - B14 config-push exits nonzero on real propagation errors
ok - B14 config-push rereads completed config writes after partial propagation
ok - B15 config reread is per-home, exact-byte, ordered, and pointer-only
ok - B16 config reread isolation, ABSENT, generation safety, send failure, and retry
ok - B20 config reread publication failures retain exact generations for retry
ok - B21 config reread instruction-write failures retain exact retry generations
ok - B21 config reread preserves exact bytes when temporary adoption also fails
ok - B21 config reread serializes concurrent propagation and delivery
ok - B22 full config reread retry queues drain before new publication
ok - B23 mixed config reread delivery failures still bound sent history
ok - B26 config reread delivery stops after the oldest failed generation
ok - B17 config reread skips unchanged homes and reads destination post-write bytes
ok - B18 bootstrap config reread path works; spawn flexibility remains defaults-only
ok - B19 bootstrap respawns before inherited-config reread
ok - B25 spawn quarantines stale rereads without blocking relaunch
ok - B24 bootstrap detect-only mode remains filesystem read-only
# all fm-secondmate-harness tests passed
```
</details>
<details>
<summary>Evidence: Evidence driver: turn-end demonstration</summary>

Source: [Evidence driver: turn-end demonstration](https://github.com/tpeixinho/firstmate/blob/3e97d9d4f69720284757715063d3f14c71b1ca30/.no-mistakes/evidence/fm/fm-agy-adapter/agy-turnend-demo.sh)

```text
#!/usr/bin/env bash
# Evidence driver: drive the REAL agy Stop hook that bin/fm-spawn.sh installs,
# with the captured payloads from docs/verification/agy.md, and narrate what the
# task's turn-end signal does after each one.
set -u
ROOT_REPO=$1
cd "$ROOT_REPO"
. tests/lib.sh
. tests/fixtures.sh
SHIM=$(mktemp -d "${TMPDIR:-/tmp}/fm-agy-demo.XXXXXX")
ln -s "$ROOT_REPO/tests/lib.sh" "$SHIM/lib.sh"
ln -s "$ROOT_REPO/tests/fixtures.sh" "$SHIM/fixtures.sh"
sed -n '1,899p' tests/fm-agy-harness.test.sh > "$SHIM/agy-helpers.sh"
. "$SHIM/agy-helpers.sh"

rec=$(make_spawn_case turnend-evidence)
IFS='|' read -r case_dir home proj wt fakebin agy_home id <<EOF
$rec
EOF
out=$(run_agy_spawn "$home" "$proj" "$wt" "$fakebin" "$agy_home" "$id" --mode no-mistakes --yolo off) \
  || { echo "spawn failed: $out"; exit 1; }
hook="$agy_home/plugins/fm-turn-end/fm-turn-end.sh"
turnend="$home/state/$id.turn-ended"

echo "== what the agy spawn installed under agy's global customization root =="
(cd "$agy_home" && find plugins fm-turn-end.d -type f | sort | sed 's/^/  /')
echo
echo "== the firstmate worktree pointer the hook reads =="
echo "  $wt/.fm-agy-turnend -> $(cat "$wt/.fm-agy-turnend")"
echo
rm -f "$turnend"
fire() { # <label> <payload>
  printf '%s' "$2" | bash "$hook" >/dev/null
  if [ -f "$turnend" ]; then s="TURN END SIGNALLED  (state/$id.turn-ended exists)"; else s="no turn end            (state/$id.turn-ended absent)"; fi
  printf '  %-46s -> %s\n' "$1" "$s"
}
echo "== feeding the installed hook the four captured Stop payloads =="
fire 'fullyIdle=false, workspace POPULATED'  "$(agy_payload "$AGY_STOP_WAITING" "$wt")"
fire 'fullyIdle key ABSENT (protojson default)' "$(agy_payload "$AGY_STOP_ABSENT" "$wt")"
fire 'blocked on trust dialog, no workspace'  "$AGY_STOP_TRUST"
fire 'fullyIdle=true  <- the real turn end'   "$(agy_payload "$AGY_STOP_DONE" "$wt")"
echo
echo "== the same hook is a silent no-op for a NON-firstmate agy session =="
rm -f "$turnend"
other=$(mktemp -d "${TMPDIR:-/tmp}/fm-agy-other.XXXXXX")
printf '%s' "$(agy_payload "$AGY_STOP_DONE" "$other")" | bash "$hook" >/dev/null; rc=$?
printf '  foreign workspace, fullyIdle=true          -> exit %s, turn-ended %s\n' "$rc" \
  "$([ -f "$turnend" ] && echo WRITTEN || echo 'not written')"
echo
echo "== bounded read: writer holds stdin open and sends nothing =="
silent="$case_dir/silent"; mkfifo "$silent"; sleep 30 > "$silent" & w=$!
t0=$SECONDS; bash "$hook" < "$silent" >/dev/null; el=$((SECONDS - t0))
printf '  hook returned after %ss with the writer still alive: %s (agy kills its hooks at 10s)\n' \
  "$el" "$(kill -0 $w 2>/dev/null && echo yes || echo no)"
kill $w 2>/dev/null; wait $w 2>/dev/null
echo
echo "== teardown retires the per-task wiring =="
token=$(sed -n 's/^token=//p' "$wt/.fm-agy-turnend")
echo "  before: registry entry $token, state token $(ls "$home/state/$id.agy-turnend-token" 2>/dev/null | wc -l | tr -d ' ')"
HOME="$home" FM_HOME="$home" FM_STATE_OVERRIDE="$home/state" FM_DATA_OVERRIDE="$home/data" \
  FM_PROJECTS_OVERRIDE="$home/projects" FM_CONFIG_OVERRIDE="$home/config" PATH="$fakebin:$PATH" \
  "$ROOT/bin/fm-teardown.sh" "$id" --force >/dev/null 2>&1 || echo "  teardown returned non-zero"
(cd "$agy_home" && find fm-turn-end.d -type f 2>/dev/null | sed 's/^/  leftover: /')
echo "  registry entries left for this task: $(find "$agy_home/fm-turn-end.d" -type f 2>/dev/null | wc -l | tr -d ' ')"
rm -rf "$SHIM" "$other"
```
</details>
<details>
<summary>Evidence: Evidence driver: launch shape and scope refusals</summary>

Source: [Evidence driver: launch shape and scope refusals](https://github.com/tpeixinho/firstmate/blob/3e97d9d4f69720284757715063d3f14c71b1ca30/.no-mistakes/evidence/fm/fm-agy-adapter/agy-scope-demo.sh)

```text
#!/usr/bin/env bash
# Evidence driver: the operator-visible surface of the agy adapter - the exact
# command line agy is launched with, the effort clamp, and the two refusals that
# keep agy inside its declared crewmate/scout scope.
set -u
ROOT_REPO=$1
cd "$ROOT_REPO"
. tests/lib.sh
. tests/fixtures.sh
SHIM=$(mktemp -d "${TMPDIR:-/tmp}/fm-agy-scope.XXXXXX")
ln -s "$ROOT_REPO/tests/lib.sh" "$SHIM/lib.sh"
ln -s "$ROOT_REPO/tests/fixtures.sh" "$SHIM/fixtures.sh"
sed -n '1,899p' tests/fm-agy-harness.test.sh > "$SHIM/agy-helpers.sh"
. "$SHIM/agy-helpers.sh"

newcase() {
  rec=$(make_spawn_case "$1")
  IFS='|' read -r case_dir home proj wt fakebin agy_home id <<EOF
$rec
EOF
}

echo "== the command line an agy crewmate is actually launched with =="
newcase shape-evidence
out=$(FM_FAKE_LAUNCH_LOG="$case_dir/launch.log" \
  run_agy_spawn "$home" "$proj" "$wt" "$fakebin" "$agy_home" "$id" \
  --mode no-mistakes --yolo off --effort high) || { echo "spawn failed: $out"; exit 1; }
sed "s#$wt#<worktree>#g" "$case_dir/launch.log" | fold -s -w 150 | sed 's/^/  /'
echo
echo "== --effort xhigh is OMITTED, never passed (agy accepts only low|medium|high) =="
newcase shape-xhigh
out=$(FM_FAKE_LAUNCH_LOG="$case_dir/launch.log" \
  run_agy_spawn "$home" "$proj" "$wt" "$fakebin" "$agy_home" "$id" \
  --mode no-mistakes --yolo off --effort xhigh) || { echo "spawn failed: $out"; exit 1; }
if grep -q -- --effort "$case_dir/launch.log"; then echo "  --effort WAS passed"; else echo "  no --effort in the launch line"; fi
echo
echo "== scope: agy refuses a secondmate spawn =="
newcase scope-secondmate
run_agy_spawn "$home" "$proj" "$wt" "$fakebin" "$agy_home" "$id" --secondmate 2>&1 \
  | tail -3 | sed 's/^/  /'
echo
echo "== scope: agy is not implicitly selectable from config/crew-harness =="
newcase scope-implicit
printf '%s\n' agy > "$home/config/crew-harness"
HOME="$home" fm_test_run_spawn "$home" "$wt" "$fakebin" "$id" "$proj" \
  --mode no-mistakes --yolo off 2>&1 | tail -3 | sed 's/^/  /'
echo "  global hook installed by that refused spawn: $(ls "$agy_home/plugins/fm-turn-end/fm-turn-end.sh" 2>/dev/null | wc -l | tr -d ' ')"
echo
echo "== scope: an implicitly resolved agy secondmate names the secondmate boundary =="
newcase scope-implicit-sm
printf '%s\n' agy > "$home/config/crew-harness"
HOME="$home" fm_test_run_spawn "$home" "$wt" "$fakebin" "$id" --secondmate 2>&1 \
  | tail -3 | sed 's/^/  /'
rm -rf "$SHIM"
```
</details>
<details>
<summary>Evidence: Evidence driver: trust-dialog pane renders</summary>

Source: [Evidence driver: trust-dialog pane renders](https://github.com/tpeixinho/firstmate/blob/3e97d9d4f69720284757715063d3f14c71b1ca30/.no-mistakes/evidence/fm/fm-agy-adapter/agy-trust-demo.sh)

```text
#!/usr/bin/env bash
# Evidence driver: what the agy launch poll sees on the pane and what it sends.
set -u
ROOT_REPO=$1
cd "$ROOT_REPO"
. tests/lib.sh
. tests/fixtures.sh
SHIM=$(mktemp -d "${TMPDIR:-/tmp}/fm-agy-trust.XXXXXX")
ln -s "$ROOT_REPO/tests/lib.sh" "$SHIM/lib.sh"
ln -s "$ROOT_REPO/tests/fixtures.sh" "$SHIM/fixtures.sh"
sed -n '1,899p' tests/fm-agy-harness.test.sh > "$SHIM/agy-helpers.sh"
. "$SHIM/agy-helpers.sh"

run_screen() { # <label> <screen>
  rec=$(make_spawn_case "trustdemo-$2")
  IFS='|' read -r case_dir home proj wt fakebin agy_home id <<EOF
$rec
EOF
  echo "== $1 =="
  echo "  what the pane renders after the launch line is submitted:"
  FM_FAKE_AGY_SCREEN=$2 bash "$fakebin/tmux" capture-pane -p 2>/dev/null | sed 's/^/    | /'
  out=$(FM_FAKE_AGY_SCREEN=$2 FM_FAKE_AGY_PHASE="$case_dir/phase" \
    FM_FAKE_AGY_EVENT_LOG="$case_dir/events.log" \
    FM_AGY_POLL_INTERVAL=0 FM_AGY_TRUST_POLLS=8 \
    run_agy_spawn "$home" "$proj" "$wt" "$fakebin" "$agy_home" "$id" \
    --mode no-mistakes --yolo off 2>&1) || { echo "  spawn FAILED: $out"; return 1; }
  n=$(agy_enters_after_launch "$case_dir/events.log")
  echo "  Enters seen after the launch line was typed: $n  (the first is the launch submit itself)"
  echo "  => trust Enters delivered: $((n - 1))"
  st=$(cat "$home/state/$id.status" 2>/dev/null || true)
  echo "  spawn outcome: ${st:-completed with no failure recorded}"
  echo
}
run_screen "workspace agy ALREADY TRUSTS - no dialog, baseline" ready
run_screen "workspace NOT trusted - dialog on screen" trust
run_screen "dialog above a bottom-anchored footer in a tall frame" trust-tall
run_screen "a row shape the anchor does not match - left unanswered on purpose" trust-hinted
run_screen "pane renders nothing at all - spawn still completes" blank
rm -rf "$SHIM"
```
</details>

## Pipeline

_... (11 earlier update rounds omitted to keep the PR body within GitHub's 65536-char limit; full history is in the run log.)_

<details>

🔧 Fix: Restore agy trust anchor; time-bound the Stop hook read
2 warnings still open:

- ⚠️ `bin/fm-spawn.sh:2481` - The pinned anchor `^[^A-Za-z]*Yes, I trust this folder[^A-Za-z]*$` is exactly the form you demanded and should NOT be changed. The defect is the claim built around it. The comment at bin/fm-spawn.sh:2470-2478, docs/verification/agy.md:151-153, and the test name `test_a_brief_quoting_the_dialog_draws_no_enter` all assert that a brief quoting the label cannot draw an Enter. I verified against grep that it can: `- Yes, I trust this folder`, `1. Yes, I trust this folder`, and `- &#34;Yes, I trust this folder&#34;` all MATCH, because the leading bullet, number, or quote is non-letter text the class absorbs. Only the mid-sentence form is rejected, and that is the only form the new fixture (tests/fm-agy-harness.test.sh:506, the `quoted-dialog` screen) exercises, so the case passes for a reason narrower than its name states. Reachability is the documented steady state rather than an edge case: on a recycled pool worktree agy is already trusted, draws no dialog, renders the brief immediately, and the poll then reads that brief for its full ~60s window. A brief for agy-adapter work (this repository&#39;s own) plausibly lists the dialog&#39;s options as bullets while also quoting the question, which satisfies both regexes and sends one blind Enter into a session that has already started its turn - the exact harm you named, an extra turn on the operator&#39;s quota. The fixer did satisfy the literal instruction (a label inside a rendered brief LINE); what is wrong is that the prose and the test name promise more than the anchor delivers. Two remedies, both leaving the regex as pinned: narrow the comment, the verification prose, and the test name to the mid-sentence case and record the bullet form as a known accepted match, or additionally require the selection marker that docs/verification/agy.md:148 records as observed (`&gt;` before the label). The second changes matching behaviour, so it is your call rather than mine.
- ⚠️ `tests/fm-agy-surface-live-e2e.test.sh:148` - Fix round 4 (commit d73c5ff, &#34;Refuse agy without jq early; match trust dialog structure&#34;) also rewrote the live payload-shape guard from `has(&#34;fullyIdle&#34;)` to `(has(&#34;fullyIdle&#34;) | not) or (.fullyIdle | type == &#34;boolean&#34;)`, and rewrote docs/verification/agy.md:34-35 to match. No finding in that round asked for it; the round&#39;s only recorded finding was the trust-row anchor. The guard&#39;s stated purpose (file header, lines 22-28) is to detect the vendor dropping a key the installed hook parses. After the change it cannot: a capture with no `fullyIdle` at all now passes, which is precisely the regression that would silently stop every agy turn-end signal from arriving. The justification given (protojson omits defaults, so an absent key is a legitimate false) does not describe any observed payload: all four measured Stops at docs/verification/agy.md:51-61 carry `fullyIdle` explicitly, including the interrupted turn, and the `AGY_STOP_ABSENT` fixture is a synthesized hypothetical rather than a capture. So the weakening solved a problem that has never occurred and gave up the one vendor regression the guard exists to catch. Because this is fix-round code that exceeded what any finding required, the remedy is to revert that hunk and the paired prose at docs/verification/agy.md:34-35 rather than to patch further, restoring the presence assertion for the operator-supplied capture.

🔧 Fix: Correct agy trust-anchor claim; restore fullyIdle presence guard
1 info still open:

- ℹ️ `docs/verification/agy.md:182` - The same event, a stray Enter delivered into an agy session that has already started its turn, is given two incompatible costs inside one record. Line 164 and the code comment at bin/fm-spawn.sh:2485 both state that a surplus Enter opens another turn and spends the operator&#39;s quota, which is the observed fact the anchor&#39;s whole shape is justified by. Line 182, describing the untested relaunch path where a predecessor&#39;s dialog text is still in the capture, calls the worst case &#39;one empty submit&#39; instead. A reader who reaches line 182 first concludes the relaunch hazard is free. Align line 182 with the observed cost already stated at line 164.

</details>

<details>

✅ No issues found.
- <code>`FM_AGY_SURFACE_LIVE_E2E=1 bash tests/fm-agy-surface-live-e2e.test.sh` - live guard against the installed agy 1.1.24; the FM_AGY_TURNEND_LIVE_E2E sub-check reported itself skipped because it needs a fresh operator-captured Stop payload</code>
- <code>`bash tests/fm-agy-harness.test.sh` - the portable agy regression, 25 cases green</code>
- <code>`bash tests/fm-spawn-dispatch-profile.test.sh` and `bash tests/fm-secondmate-harness.test.sh` - the two sibling suites touched for the agy marker scrub and the secondmate refusal</code>
- <code>`bash tests/fm-muse-harness.test.sh` and `bash tests/fm-test-fixtures.test.sh` - the included pre-existing macOS failure, now green on this exact platform</code>
- `Manual evidence driver: installed the agy Stop hook through the real spawn path, then fed it AGY_STOP_WAITING (fullyIdle=false, workspace populated), AGY_STOP_ABSENT (key omitted), AGY_STOP_TRUST (no workspace) and AGY_STOP_DONE (fullyIdle=true), checking state/<id>.turn-ended after each`
- <code>Manual evidence driver: fired the same installed hook with a foreign workspace to show the global hook is a silent no-op for non-firstmate agy sessions, timed its bounded read against a writer that holds stdin open and sends nothing, and ran `bin/fm-teardown.sh &lt;id&gt; --force` to show the registry entry retired</code>
- <code>Manual evidence driver: captured the exact agy launch command line, the --effort xhigh omission, and the verbatim refusal text for `--secondmate`, for an implicitly resolved `config/crew-harness=agy`, and for an implicitly resolved agy secondmate</code>
- `Manual evidence driver: rendered each trust-dialog pane fixture the launch poll reads and recorded how many trust Enters the poll delivered for each`
- <code>Manual check of the trust anchor `^[^A-Za-z]*Yes, I trust this folder[^A-Za-z]*$` against the three counter-examples docs/verification/agy.md cites verbatim plus the mid-sentence form it does exclude</code>
- <code>Manual check: copied /bin/bash exits 137 (SIGKILL) on this host while `fm_test_named_interpreter` resolves the symlink construction and the process table reports the requested name</code>
- <code>`grep -rn FM_AGY_CONFIG_HOME bin tests docs .agents` - returns nothing, confirming the accepted decision that the turn-end registry root is fixed with no environment override</code>
- <code>`grep -n &#39;has(&#34;fullyIdle&#34;)&#39; tests/fm-agy-surface-live-e2e.test.sh` - the presence assertion the last round restored is in place</code>
- <code>`grep -n agy bin/fm-session-lock-lib.sh` - agy is absent from the primary-session identity surface, as the declared scope boundary requires</code>

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>


tpeixinho:
On "Worker can forge turn completion" (`bin/fm-spawn.sh`, the `.fm-agy-turnend` pointer)

The finding is correct, and I want to be clear that I am not disputing it. What I am reporting is that it is not introduced by this change: the agy adapter copies the pointer-plus-token design that grok and Kimi already use, rather than inventing it.

Same file, current branch:

| Adapter | Pointer written into the worktree | Excluded | Hook reads it from the workspace |
|---|---|---|---|
| grok | line 2830 | line 2831 | line 2815 |
| Kimi | line 3009 | line 3010 | installed hook, `bin/fm-kimi-turnend-hook.sh` |
| agy (this change) | line 2994 | line 2995 | line 2973 |

In all three, the global hook is a guarded no-op that acts only when the workspace holds a pointer naming a token in firstmate's private registry, and the worktree pointer is what the worker itself can reach.

So the property the finding describes is shared by the established pattern. Addressing it properly means changing how all three adapters bind a turn-end signal to a task, which is a wider design decision than adding one adapter, and I did not want to make that call unilaterally inside this pull request.

I am leaving the decision with you. If you would prefer this adapter not ship until the pattern is changed, or would like it tracked as follow-up work across all three, say which and I will follow that.

Issue #3451, opened 2026-09-01
Session lock wedges a home permanently: liveness anchor selected by harness command-name match can bind a long-lived unrelated process (e.g. ChatGPT desktop's bundled codex app-server)


3264studios:
## Summary

The session lock's liveness/ownership check identifies the lock-holding process by matching command names against the known-harness pattern (`FM_HARNESS_RE` in `bin/fm-session-lock-lib.sh`). A long-lived unrelated process whose command line looks like a harness satisfies that match, so the lock reads as live forever and the home wedges permanently until the operator intervenes by hand.

## Observed failure (2026-08-31, production)

`state/.lock` recorded pid 55359 as its liveness anchor. That pid belonged to the **ChatGPT desktop app's bundled `codex` app-server** — a long-lived process that is not a firstmate session at all, but whose command matched the harness pattern. Every subsequent session start refused the lock ("another active session"), with nothing to time out, because the anchor process genuinely stays alive indefinitely.

The failure shape needs only: any always-running app that embeds/bundles a harness-named binary (the ChatGPT desktop app's app-server is one concrete instance) plus one lock publication that lands on that pid.

## Root cause, stated precisely

Two boundaries disagree about the same principle:

1. **Publisher**: the ancestry walk (`fm_harness_ancestry_pids` and its first-pid consumer) selects the innermost ancestor **whose command matches a harness name** and publishes that pid as the lock's liveness anchor.
2. The library's own comment near the classifier says the opposite: *"A command line is deliberately absent from this decision. A harness-looking command can belong to a long-lived unrelated app."*

So a command match that is deliberately excluded from classification still decides what pid gets published, and in the shell → app-server ancestry shape it publishes the unrelated app-server.

## Findings from an abandoned local fix attempt (offered as design input)

We ran six review rounds on a local fix and ultimately abandoned it rather than land it; two hard-won constraints from that effort may save the next attempt:

- **Publication needs positive session-lifetime evidence**, not a command match: resolve the published pid at a boundary that can prove it dies with the session (covering the app-server-hosted shape and Claude's documented bg-spare/transient-worker chain, where a command-matched pid can be a short-lived worker — the opposite failure: another session reclaims while the real one is live).
[...]

Issue #3396, opened 2026-08-31
Windows/Git Bash: session lock always refuses - harness ancestry cannot be resolved


amitjethva:
## Symptom

Every session start on Windows + Git Bash refuses the fleet lock and drops to read-only:

```
LOCK
error: cannot locate harness process in ancestry
●  READ-ONLY SESSION - FLEET LOCK OWNERSHIP WAS NOT VERIFIED
```

The session can never spawn, steer, merge, drain the wake queue, or repair supervision. It is not a stale lock — `state/.lock` is absent and no other session is running.

## Cause 1 — `ps -o` is not supported by Cygwin/MSYS ps

`fm_harness_ancestry_pids` (`bin/fm-session-lock-lib.sh:109`) walks the ancestry with:

```sh
comm=$(ps -o comm= -p "$pid" 2>/dev/null) || break
args=$(ps -o args= -p "$pid" 2>/dev/null)
...
pid=$(ps -o ppid= -p "$pid" 2>/dev/null | tr -d ' ')
```

Git Bash ships Cygwin `ps`, which has no `-o`:

```
$ ps --version
ps (cygwin) 3.4.10

$ ps -o comm= -p $$
ps: unknown option -- o
Try `ps --help' for more information.
$ echo $?
1
```

The first `ps` fails, `|| break` fires on hop one, `printed` stays 0, and the function returns non-zero. Every caller then fails closed.

## Cause 2 — the harness is not visible in the Cygwin process tree

This one is not fixed by a portable rewrite. Shells the Windows harness spawns are reparented in Cygwin's view:

```
$ ps -p $$
      PID    PPID    PGID     WINPID   TTY         UID    STIME COMMAND
    28461       1   28461       8796  ?         197609 21:29:52 /usr/bin/bash

$ awk '{print $4}' /proc/$$/stat      # /proc agrees
1
```

PPID is 1, not the `claude` process. The Windows-side parent exists (`ps -W` lists it) but the parent link does not cross the Cygwin boundary, so no amount of walking `ppid` reaches the harness. `/proc/<pid>/ppid` reports 1 as well.

So the contiguous-harness-run model in the header comment cannot be satisfied on this platform by ancestry alone.

## Environment

- Windows 11 Pro for Workstations 10.0.26200
- Git Bash / Cygwin ps 3.4.10
- Claude Code, primary harness `claude`
- firstmate @ c7fdef9

## Impact

[...]

Issue #3279, opened 2026-08-29
fm_lock_acquire_wait spins forever when a reused PID masks a dead lock owner (watcher wedge)


ohigs:
## Summary

`fm_lock_acquire_wait` (in `bin/fm-wake-lib.sh`) can spin forever waiting on a lock whose owner is already dead. The stale-lock recovery checks liveness with a bare `kill -0` on the recorded PID. When the OS reuses that PID for an unrelated live process, the dead owner looks alive forever, so every watcher cycle waits out its full lock-wait instead of recovering the stale lock.

## Observed impact

With a stale `state/.home-summary-refresh.lock` in this shape, the main watcher (`bin/fm-watch.sh`) wedged repeatedly inside its poll loop:

- The loop-top liveness beacon (`state/.last-watcher-beat`) went stale for minutes at a time (repeatedly observed 2-3 min; one recorded ~33 min stretch).
- A macOS `sample` of the wedged process showed a deep recursive bash frame chain ending in `read` - the lock-wait retry path recursing instead of failing fast.
- Cycle-time stalls cascaded into liveness alarms, watcher re-arms, and duplicate watcher processes across homes.
- `state/home-summary.json` stopped refreshing entirely until the stale lock was removed by hand.

## Reproduction

1. Create a lock dir with an owner dir recording a PID that no longer exists.
2. Let the OS reuse that PID for an unrelated live process (or forge the owner dir to name any live PID that is not the real owner).
3. Run the lock-wait acquire path: it spins until its timeout instead of detecting that the recorded PID no longer points at the process that created the lock.

The PID-reuse condition is rare per-lock but near-certain over weeks of watcher uptime across multiple homes, which matches the sporadic "beat stale, watcher alive" reports.

## Suggested fix

Make stale-lock liveness identity-bound rather than PID-only: verify the recorded PID still belongs to the process that created the lock (e.g. compare process start time / launch identity captured at acquire, stored beside the PID in the owner dir) before treating the lock as live. Recover (or fail fast) when identity does not match.

## Verification

Reproduced end to end with a reused-PID owner dir, fixed locally, and covered by a regression test (`tests/fm-watcher-lock.test.sh`, case: a reused PID masking a dead owner no longer wedges the acquire path and the lock is recovered).

kunchenguid:
Speaking as Kun's firstmate: this is real on latest main (`f66be0f`) and distinct from #1508 / #3267.

[...]

Issue #3267, opened 2026-08-29
Windows: fm_lock_try_create can never succeed without symlink privilege (ln -s silently degrades to a copy)


trngthnh369:
### Summary

On Windows without the symlink privilege, `fm_lock_try_create()` can **never** succeed, so every `fm-spawn.sh` invocation fails immediately with:

```
error: this home's task set is locked by another operation (a forced teardown is
enumerating or removing its tasks); refusing to create task <id> rather than racing it
```

This happens with a **completely clean lock state** — no stale claim, no contention, no prior run. It is distinct from #1508, which is about stale claims not being *stealable* (`fm_pid_alive` / `kill -0` false positives). That issue assumes the lock can be created in the first place; this one is upstream of it — the lock is never creatable at all.

### Root cause

`fm_lock_try_create()` (`bin/fm-wake-lib.sh:406`) uses a **symlink** as the lock primitive:

```bash
if ln -s "$ownerdir" "$lockdir" 2>/dev/null && fm_lock_points_to_owner "$lockdir" "$ownerdir"; then
```

and identity is resolved with `readlink`:

```bash
fm_lock_points_to_owner() {
  local lockdir=$1 ownerdir=$2 actual
  actual=$(readlink "$lockdir" 2>/dev/null) || return 1
  [ "$actual" = "$ownerdir" ]
}
```

On MSYS/Git Bash, when the user has neither Developer Mode nor `SeCreateSymbolicLinkPrivilege`, `ln -s` **silently degrades to a copy**: it exits 0, the path appears to exist, but it is a plain directory. `readlink` then returns nothing, `fm_lock_points_to_owner` returns 1, and `fm_lock_try_acquire` reports the lock as held by someone else.

The misleading part is that `ln -s` reports success, so nothing in the code can tell that the primitive silently changed semantics.

### Reproduction

Windows 11, Git Bash (MINGW64), non-elevated, Developer Mode off.

```console
$ MSYS= ln -s "$PWD/target" link ; echo "exit=$?" ; [ -L link ] && echo "symlink" || echo "NOT a symlink"
exit=0
NOT a symlink

$ MSYS=winsymlinks:nativestrict ln -s "$PWD/target" link
ln: failed to create symbolic link 'link': Operation not permitted
```

Then, from a fresh clone with no prior run:

```console
$ bash bin/fm-spawn.sh test-1 /path/to/some/repo --scout
error: this home's task set is locked by another operation ...
```

[...]

PR #3553, opened 2026-09-02
fix(session-lock): identify a harness session on Windows/Git Bash

lmktechnology:
## What and why

Fixes the read-only wedge in #3396. On Cygwin (Git for Windows) every session start refused the fleet lock with `error: cannot locate harness process in ancestry`, so spawning, steering, merging, the wake-queue drain, and supervision repair were skipped on every start.

There are two independent causes on the identity path, and fixing only the first still leaves the session read-only.

**Cause 1 - `ps -o` does not exist on Cygwin.** Cygwin's `ps` accepts only `[-aefls] [-u UID] [-p PID]` and fails the whole invocation with `unknown option -- o`, so the ancestry walk aborted on hop one. `comm`, `args`, and `ppid` now go through accessors that fall back to Cygwin's fixed columns; the procps/BSD path is unchanged.

**Cause 2 - the parent link does not cross the Cygwin boundary.** The harness is a native Windows process, and Cygwin reports the PPID of a shell it spawns as `1`, so no walk can reach it. Identity now comes from the session pid the harness publishes, **confirmed against the Windows process table before use**: the pid must still be live *and* its executable must independently identify a verified harness by the same rules every other platform uses. An absent, stale, or non-harness value is discarded, not bound.

## A Windows parent-chain walk was implemented, then deliberately removed

The obvious fallback - walk the real Windows parent chain via `Win32_Process` - was built and tested, and it is **not safe on this platform**:

- MSYS emulates `exec` by spawning a fresh Windows process and exiting the old one, so intermediate shells vanish constantly and a child's recorded parent is routinely a pid that no longer exists. Measured directly: a parent id appeared in neither `ps -W` nor a full `Win32_Process` snapshot taken moments earlier. The chain simply breaks.
- Windows never reparents an orphan, so that dangling parent id stays on the child and Windows is free to reissue it. Following it can land on an unrelated live process and bind a home's session lock to it - the same wrong-process-binding class as #3451 and #3279.

A harness that publishes nothing is left unresolved instead, which keeps the session read-only exactly as before. That is the safe direction, and `bin/fm-lock.sh` now says so specifically rather than blaming an ancestry that could never contain the answer.

## Windows pids are tagged, not stored bare

They are a different namespace. `kill -0` reports a live Windows process as dead, and the number can collide with an unrelated live Cygwin pid. Storing a bare number would be indistinguishable from a local pid, so the value is tagged `win:<pid>`. That makes it non-numeric, so any consumer treating it as a local pid - including a future `kill` - refuses it instead of acting on the wrong process:

```
$ kill -0 win:7204
bash: kill: `win:7204': not a pid or valid job spec
```

Nothing signals the session lock today; the two `kill -TERM` sites read the *watcher* lock, a different file.

## Every gate that reads the lock accepts the tagged identity

Introducing a second identity shape into `state/.lock` was the risky part of this change, and the first revision got it wrong. Six gates read that field, and each answered "is this value usable at all" with its own inline numeric test, so every one of them read a valid Windows holder as malformed. Thanks to the review bot for catching the first two; auditing the field turned up four more.

The visible cost was concentrated in startup completion: the record was never written, and the clear/compact check that consumes it could never match, so **every clear or compact on Windows repeated the full startup sequence**. The deferred network sweeps reported ownership as changed when it had not, and the Stop auto-arm treated a dead Windows session as an unreadable lock rather than a recoverable one.

`bin/fm-lease.sh` was the outlier and the one worth calling out. Rather than refusing a value it could not use, `tr -cd '0-9'` reduced `win:7204` to `7204` - a number naming an unrelated process in the *local* table. It happened to fail closed downstream because the equality check against the untouched lock value could never match, but deriving a wrong-namespace pid is exactly what the tag exists to prevent, so it now takes the value whole and requires a local pid, falling through to the shell pid when the lock holds anything else.

`fm_session_pid_valid` is now the single owner of that question and all six gates delegate to it, so a third identity shape cannot silently split them again.

## `bin/fm-sessionstart-nudge.sh`

Its private ancestry walk had the same `ps -o` problem, so it now reads ppid through the portable accessor, and defers to `fm_session_lock_owned_by_self` for a Windows-tagged holder, which is not in the local process table at all.

Its own question is otherwise left alone. An earlier revision of this branch replaced the whole walk with `fm_session_lock_owned_by_self`; that was wrong. The nudge asks whether a process in this ancestry took the lock, while the ownership predicate additionally requires a *verified harness* in that ancestry, so a lock written by a plain shell started being nudged to run session start again. `tests/fm-sessionstart-nudge.test.sh` pins the looser contract deliberately and caught it.

## Relationship to #3551

#3551 fixes cause 1 only. I simulated its exact walk on the affected machine: `ps -f -p $$` returns parent `1`, so it stops on hop one and the session stays read-only. The accessors here are deliberately close to that PR's shape so whichever lands first, the other is a small merge. Credit to that author for the `-o` diagnosis.

## Verification

Windows 11 26200, Git Bash, Cygwin `ps` 3.4.10, harness `claude`:

| Case | Result |
| --- | --- |
| acquire | `lock acquired: harness pid win:7204` |
| stored value | `win:7204` (confirmed as the real `claude.exe`) |
| `status` | `lock: held by live harness pid win:7204` |
| re-acquire | idempotent |
| ownership recognized | yes |
| published pid that is dead | refused |
| published pid naming `explorer.exe` | refused |
| published pid naming `C:\tools\claude-notes\helper.exe` | refused (path-component safety survives backslashes) |
| no published pid | refused with a platform-specific diagnostic |
| nudge with our lock / another's lock / no lock | silent / nudges / nudges |

Each identity gate was then exercised against its **shipped bytes** with a tagged lock in place, comparing this branch to the same function on `main`:

| Gate | `main` | this branch |
| --- | --- | --- |
| `session_start_completed` (clear/compact) | reruns the full startup | re-emits only |
| startup completion record | not recorded | records `win:7204` |
| `lock_unchanged` (network sweeps) | sweeps downgraded | sweeps run |
| `network_mutation_authorized` | false "ownership changed" | authorized |
| lease holder pid from a tagged lock | `7204` (wrong namespace) | empty, falls back to this shell |

Five regressions added to `tests/fm-session-lock-ancestry.test.sh`. Both platform departures are reproduced behind a fake process table - the `-o` rejection and the severed parent link - so they run on Linux and macOS CI rather than only on Windows. They also pin the namespace separation (a tagged pid whose number *is* a live harness in the Cygwin table must still be refused) and that a published identity is accepted by the gates that read it back.

Run on Linux (WSL Ubuntu), since neither the pinned ShellCheck nor this suite's e2e layer can run on the Windows host. Note `tests/fm-session-start.test.sh` pins a minimal `FM_TEST_BASE_PATH`, so `jq` must be reachable from it or the home-summary assertions fail for environmental reasons on `main` too:

| Suite | Result |
| --- | --- |
| `fm-session-lock-ancestry` | 12/12, including the three e2e cases |
| `fm-session-start` | 49/49 (`main`: 47/47) |
| `fm-bootstrap` | 28/28 |
| `fm-bootstrap-network-parallel` | 2/2 |
| `fm-startup-network` | 18/18 |
| `fm-claude-stop-autoarm` | 39/39 |
| `fm-sessionstart-nudge` | 7/8; the failure is the pre-existing `OpenCode exact nudge delivery: expected exit 0, got 127`, identical on `main` (no `opencode` binary in that environment) |

`bin/fm-lint.sh` clean on the full tree with ShellCheck 0.11.0 (pinned 0.11.0), full extended analysis. Its only remark is that `actionlint` is absent in that environment; no workflow files are touched by this branch.

On the Windows host the e2e layer of the ancestry suite fails identically on `main` and on this branch - it cannot exec its symlinked fixture (the `ln -s` limitation of #3267) - so it is unchanged, not regressed.

https://claude.ai/code/session_01F9T29YDqYSkQeDTC2Nfi7h

### DIFF
--- a/.agents/skills/ask-user-authority/SKILL.md
+++ b/.agents/skills/ask-user-authority/SKILL.md
@@ -47,6 +47,16 @@ State all five of these elements in one concise, evidence-first escalation:
 
 Do not relay reviewer labels or gate output as if they settled the decision.
 
+## Deliver the decision
+
+This skill also owns how a decided finding reaches the worker, because the delivery step is what makes the decision stick.
+
+Send the same worker one exact decision naming the decision key, the step, the action, the affected finding IDs, instructions where they are needed, and the exact response command to run.
+Pass `fm-send`'s `--resolve-key` so the worker's open decision record closes at answer time rather than staying open behind a later `working:` or `done:` line.
+Require the matching `resolved` event before treating the decision as landed.
+Forbid `--yes`: it would silently bypass firstmate's authority check and any required captain escalation.
+Require the worker to process every synchronous return until completion or a genuinely new escalation, rather than stopping at the first gate response.
+
 ## Classification examples
 
 - Fixing a concrete defect that violates an original acceptance criterion is firstmate's to decide, regardless of implementation difficulty.
--- a/.agents/skills/project-management/SKILL.md
+++ b/.agents/skills/project-management/SKILL.md
@@ -41,7 +41,7 @@ Choose that posture when adding or creating the project:
 - `no-mistakes` runs the full validation pipeline before a PR.
 - `direct-PR` pushes and opens a PR without the no-mistakes pipeline.
 - `local-only` has no required remote or PR and lands only through the approved local fast-forward path.
-- `no-mistakes-prod-only` is a conditional policy rather than one flat mode: genuinely internal-only tooling, automation, contributor or operator process, and release or submission work ships `direct-PR`, while product-facing, mixed, and uncertain work ships `no-mistakes`.
+- `no-mistakes-prod-only` is a conditional policy rather than one flat mode: some of its work ships `no-mistakes` and some ships `direct-PR`, and `AGENTS.md` section 7 owns which surface takes which.
 
 `no-mistakes-prod-only` is the default for a newly added or created remote-backed project when the captain specifies nothing, and a project with no remote defaults to `local-only`.
 State that resolved default while confirming the source, local name, and posture instead of asking the captain to choose from scratch, and record a flat mode instead whenever they ask for one.
--- a/.agents/skills/quota-array-dispatch/SKILL.md
+++ b/.agents/skills/quota-array-dispatch/SKILL.md
@@ -13,7 +13,7 @@ metadata:
 # quota-array-dispatch
 
 This skill is the single owner of the completion-aware profile-array selection procedure.
-`AGENTS.md` section 4 owns the always-loaded intake boundary, load trigger, malformed-config refusal, every-candidate accounting, and strongest-reasoning/tie safety rules.
+`AGENTS.md` section 4 owns the always-loaded intake boundary, this skill's load trigger, and the malformed-config refusal that fires before any array is resolved; the every-candidate accounting, strongest-reasoning-class, and tie rules are owned here.
 `harness-adapters` owns harness verification, model/provider discovery, and effort fallback.
 `quota-axi` remains data-only: it publishes `spendPriority` as a comparable scalar and never recommends, selects, ranks, or infers a route.
 Do not add a daemon, opaque composite score, routing wrapper, hard-coded model-specific policy, or producer-side route recommendation.
--- a/.agents/skills/validation-run-changes/SKILL.md
+++ b/.agents/skills/validation-run-changes/SKILL.md
@@ -1,0 +1,50 @@
+---
+name: validation-run-changes
+description: >-
+  Agent-only contract for changing a task while its no-mistakes validation run is active.
+  Load before expanding a task under an active run, and before acting on a captain instruction that invalidates work being validated.
+  Owns the in-scope boundary during validation and the abort-settle-replace-revalidate sequence that is the only supported way to supersede a run.
+user-invocable: false
+metadata:
+  internal: true
+---
+
+# validation-run-changes
+
+Load this once a no-mistakes validation run is active and something wants to change the work under it.
+`AGENTS.md` section 7 owns the always-loaded facts: the worker that starts a run drives it, firstmate never calls `no-mistakes axi respond` for a crew-owned run, and nothing outside this skill authorizes hand-editing, committing, aborting, or restarting while a run owns the branch.
+`ask-user-authority` owns gate findings; this skill covers requirement changes arriving from outside the pipeline.
+
+## In-scope boundary during validation
+
+Once validation starts, prefer routing new requirements to follow-up work rather than expanding the current task, unless a new requirement completely invalidates the work being validated.
+The following stay within the current task even when they touch files not named at intake:
+
+- The smallest downstream changes needed to keep already accepted product or engineering behavior correct.
+- Behavioral tests added where an executable contract exists.
+- Documentation kept accurate for the accepted behavior.
+
+Corrections required to satisfy already accepted intent are not new requirements.
+
+## Superseding a run
+
+Only a current, explicit captain instruction that completely invalidates the work being validated keeps the task with the same worker instead of routing it to follow-up work or handing it to a replacement.
+When that happens, the worker follows this sequence exactly and in order.
+
+1. **Abort.**
+   Cancel the active run through no-mistakes axi's supported abort command, and confirm through axi status that the run has stopped before changing any code.
+   That single supported abort is the only permitted intervention: do not hand-edit, commit, restart, or start a second validation run while the obsolete run still owns the branch.
+2. **Settle ownership.**
+   Follow `branch_sync.next_action` from structured axi status.
+   Use axi sync's supported guarded recovery only when its code is `recover_custody`.
+   Otherwise proceed only when structured status confirms that branch ownership is already returned and no recovery is required.
+3. **Replace, do not build on.**
+   Custody recovery settles branch ownership, not content.
+   Replace the obsolete work from the correct pre-invalidation base rather than building on top of the recovered-but-obsolete head, keeping the obsolete run's own pipeline-fix commits out of what gets validated and shipped.
+4. **Revalidate once.**
+   Once ownership is settled, validate exactly once against that final head so no obsolete or intermediate head is ever treated as authoritative.
+
+## Duplicated pipeline ownership
+
+A worker hand-editing, committing, aborting, or restarting during an active validation run outside the sequence above has duplicated pipeline ownership.
+Steer it back to the gate response flow rather than letting a second owner accumulate commits under the run.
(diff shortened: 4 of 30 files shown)


lmktechnology:
Confirmed and fixed in 9256182 - thank you, this was a real defect and the reasoning was exactly right.

I verified both call sites, then audited every reader of `state/.lock` rather than fixing only the two named, and found four more instances of the same mistake:

| Gate | Effect of a tagged identity before the fix |
| --- | --- |
| `bin/fm-session-start.sh` completion record | never written |
| `bin/fm-sessionstart-run.sh` `session_start_completed` | clear/compact reran the full startup |
| `bin/fm-startup-network.sh` `lock_unchanged` | network sweeps silently downgraded |
| `bin/fm-bootstrap.sh` `network_mutation_authorized` | false "ownership changed" every session |
| `bin/fm-claude-stop-autoarm.sh` | a dead Windows session read as an unreadable lock, so never recoverable |
| `bin/fm-lease.sh` | `tr -cd '0-9'` reduced `win:7204` to `7204` |

The lease one is the one I would flag back to a reviewer. It did not reject the value, it *mangled* it into a number naming an unrelated process in the local table. It happened to fail closed downstream, because the equality check against the untouched lock value could never match, but deriving a wrong-namespace pid is precisely what the tag exists to prevent. It now takes the value whole and requires a local pid.

Root cause was structural rather than six separate oversights: "is this lock value usable at all" had six independent inline implementations, so adding a second identity shape was guaranteed to split them. `fm_session_pid_valid` now owns that question and all six delegate to it.

Verified against the shipped bytes of each gate with a tagged lock in place, contrasted with the same function on `main`:

```
session_start_completed   branch: COMPLETED -> compact re-emits only
                          main:   NOT COMPLETED -> compact reruns the full startup
lock_unchanged            ok -> sweeps run
network_mutation_authorized  ok -> sweeps authorized
completion record         ok -> writes win:7204
lease holder pid          branch: empty, falls back to this shell
                          main:   7204   <- wrong-namespace pid
```

Added `a published identity is accepted by every gate that reads the lock`, which pins the round trip: the identity a session publishes must be accepted by the predicate that reads it back, local pids stay valid, and torn or bare-tag values still fail closed. Full results are in the updated description.

cr101:
@lmktechnology, thanks for the thorough Windows investigation and fixes here.

Before resolving the current conflicts, we need to establish a clear update path so completed work does not become stranded:

   1. Are you available to refresh this PR onto current `main`?
   2. If not, do you approve a clean replacement PR that preserves your authorship and credits this work?

   The refreshed branch should contain only the three Windows session-identification commits:

   - `8b281b7b5507`
   - `0baf64f71ffe`
   - `92561828fb95`

The three earlier commits appear unrelated and currently expand the PR to 30 files, making conflict resolution substantially harder.

One additional requirement surfaced on native Windows: the current implementation recognizes `CLAUDE_PID`, but Pi does not publish that variable. The final solution should either include a verified Pi process-ID path or explicitly leave Pi support to a linked follow-up.

If you are unavailable and approve the replacement route, we will prepare a clean PR from current `main`, preserve attribution, and link this PR as its source. No replacement PR will be opened without your approval.

cr101:
Following up: this remains blocked, and I’m holding off on implementation until there’s an agreed landing path.

@lmktechnology: are you available to refresh this PR, or do you approve a clean replacement preserving your authorship?

@kunchenguid: alternatively, would you approve a maintainer-assisted refresh of this existing PR, limited to the three Windows session-identification commits identified above, with Pi support tracked separately?

No implementation work or replacement PR will begin pending that confirmation.

MLA82:
Tested this branch on native Windows (Git Bash) against a current upstream base.

**Environment:** Windows 11 (build 26120), Git for Windows 2.55.0, Cygwin `ps` 3.6.9, Claude Code primary (publishes `CLAUDE_PID`), Developer Mode on with `MSYS=winsymlinks:nativestrict`.

**Applying to current main:** cherry-picked `8b281b7`, `0baf64f`, `9256182` onto `ecfe071981b1`. Two conflicts, both mechanical:

- `bin/fm-session-lock-lib.sh`, walk loop: main now examines pid 1 (a harness that is pid 1 of its own namespace). I kept that rule and swapped in the portable accessor:
  ```sh
  pid=$(fm_ps_ppid "$pid")
  case "$pid" in '' | *[!0-9]*) break ;; esac
  [ "$pid" -ge 1 ] || break
  ```
- `bin/fm-sessionstart-nudge.sh`: resolved directly to the `0baf64f` shape (local walk through `fm_ps_ppid`, a `win:` holder deferred to `fm_session_lock_owned_by_self`), again keeping main's pid-1 allowance, so `0baf64f` then applies empty.

**Results on the Windows host:**

- `bin/fm-lock.sh`: `lock acquired: harness pid win:7808` (the real `claude.exe`); `status` reports `held by live harness pid win:7808`; re-acquire is idempotent. The same session on unpatched main gets `cannot locate harness process in ancestry`.
- `bin/fm-session-start.sh`: takes the lock and runs the locked bootstrap sweeps, the wake drain and the deferred network checks.
- `bin/fm-watch-arm.sh`: `watcher: started ... (beacon fresh)`, delivered a wake, and the `fm-wake-drain.sh --ack-through` acknowledgement went through.
- Suites:
  - `tests/fm-session-lock-ancestry.test.sh`: 13/13, including the three e2e cases (they can exec their fixtures here because symlinks work with Developer Mode plus `nativestrict`).
  - `tests/fm-claude-stop-autoarm.test.sh`: 19/20. The one failure is `fm-check-register.sh could not register the custom check`, the NTFS mode-600 check that #4135 addresses, not this change.
  - `tests/fm-sessionstart-nudge.test.sh`: 7/8 plus one skip. The failure is `OpenCode exact nudge delivery` (no `opencode` installed), matching the note in the description.

**Outside this PR's scope, seen in the same run** (for whoever tracks native Windows):

- `bin/fm-fleet-sync.sh` compares `git rev-parse --show-toplevel` (`D:/...`) with `pwd -P` (`/d/...`), so every clone is skipped as `not a clone root` (observed; the Windows sibling of #3448).
- `bin/fm-procevent.sh` `require_isolated_group` reads `ps -o pgid=`, which Cygwin `ps` rejects, so a process-event runner would die with `cannot inspect runner process group` (by inspection, not yet run).

Happy to rerun anything on this host if it helps the refresh.

lmktechnology:
@cr101. due to time i approve a clean replacement preserving my authorship. Will still try to

cr101:
The replacement PR, https://github.com/kunchenguid/firstmate/pull/4803, is now ready for review. It preserves Lloyd's original commits and attribution, along with the subsequent fixes and history-preserving upstream integration. Please continue review there. It has not been merged.

PR #3560, opened 2026-09-02
fix(bin): never let a skipped live run hand a branch to older history

kuan0808:
nm_runs_status_for_branch answers "is a run for this branch active right now" from the newest-first `no-mistakes runs` list. A row whose recorded head does not match the worktree is skipped as a reused-branch mismatch. When that skipped row was the LIVE run (the shape right after the local branch advanced past the run's head), the walk continued to an older terminal row for the same branch and reported its `failed` as this crew's current state.

Remember that a live row was skipped. When only older terminal history would then answer, answer nothing, so the caller's pane and log fallback decides. A genuine newest failure with no live run still reports failed, and exact-head attribution is unchanged.

tests/fm-crew-state.test.sh covers the live-plus-older-failed shape. On the previous script it asserted `state: failed`.

### DIFF
--- a/bin/fm-crew-state.sh
+++ b/bin/fm-crew-state.sh
@@ -375,9 +375,11 @@ nm_ci_checks_state() {
 # is exact) - but branch + coarse status is exactly what this predicate needs:
 # is a run for THIS branch active right now. Echoes the first (most recent)
 # matching row's status word (running/completed/cancelled/failed), or empty
-# when the branch has no run within FM_CREW_STATE_RUNS_LIMIT rows.
+# when the branch has no run within FM_CREW_STATE_RUNS_LIMIT rows, when a head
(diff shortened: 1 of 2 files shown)


kuan0808:
Closing as superseded by merged #3681.

Rechecked against main `8f7b79c77c2198a71a01082215227a64500015e3`: the shared `fm_nm_runs_status_for_worktree` in `bin/fm-nm-run-lib.sh` lets only the newest same-branch ledger row decide, rather than falling through to older terminal history. This PR's unchanged `test_live_coarse_row_is_not_replaced_by_older_terminal_history` passes against that upstream production code, including the live-row/older-failure case, an attributable genuine failure, and an unresolvable newest terminal row.

That covers this PR's specific regression. Broader full-status/PR provenance work remains separate; this does not claim those issues are resolved. The branch and its history are retained.

PR #3574, opened 2026-09-03
Make push scans fail closed when no terms are loaded

jleemcf:
## Summary

Push scans resolved their term list through a caller-relative path. When that path did not exist in the working directory, the scan loaded no terms and still reported a pass, so a scan could report clean without having checked anything.

This change makes that outcome impossible:

- The term list is resolved by absolute path from the tool's own root rather than from the caller's working directory.
- A missing, unreadable, or empty list is a hard failure that stops the operation instead of reporting a pass.
- Every scan record states how many terms were loaded, so a zero-term load is visible rather than silent.
- The scan covers the full change set, the branch name, commit messages, commit authorship, and the request title and body.
- The scan step lives in the shared delivery contract, so ordinary and promoted deliveries both carry it.

## Validation

Regression coverage asserts the failure paths directly. A missing list, an unreadable list, an empty list, and a list containing only comments each fail closed rather than reporting a pass. Further cases cover a planted match failing the scan, and a clean set passing while reporting its loaded-term count.


jleemcf:
Withdrawing this pull request; we're no longer pursuing this change.

PR #3576, opened 2026-09-03
fix(bin): detect missing perl JSON::PP and fail loudly in captain-hold (#2)

NewAiCoder:
## What Changed

- `bin/fm-bootstrap.sh` now runs `perl -MJSON::PP -e1` during local tool detection and reports `MISSING_MANUAL: perl JSON::PP module` with per-platform package commands when it fails, matching the existing bootstrap-diagnostics contract. `JSON::PP` ships with a full perl distribution but is packaged separately on minimal installs (hit live on a minimal Fedora host: bare perl, module absent, hand-fixed with `sudo dnf install perl-JSON-PP`).
- `bin/fm-captain-hold.sh`'s decoding subcommands (`answer`, `answers`, `complete`, `verify`, `diverged`, and the `--identity` branch of `open`) now fail loudly with the same install hint when the JSON::PP module is missing, instead of silently reading empty fields. `bind`, `unbind`, and `binding` stay usable without the module since they never decode JSON.
- The guard is scoped to only the code paths that actually decode JSON: `open` needs it only for its `--identity` branch (not the ordinary `fm_backlog_row_probe` path), and `verify`/`reconcile` need it only when `decision_keys` is non-empty (the common case for a scout task with no captain calls at all needs no decoder at all). An earlier round of this change called the guard unconditionally ahead of all three subcommands, which broke `command_open`'s and `command_verify`'s exit-code contract on a JSON::PP-less host even when no decoding was needed; this revision fixes that.
- `.agents/skills/bootstrap-diagnostics/SKILL.md` and `docs/configuration.md` document perl JSON::PP in the universal toolchain and its `MISSING_MANUAL` instructions field.
- Test fixtures (`tests/lib.sh`, `tests/fm-bootstrap.test.sh`, `tests/fm-secondmate-sync.test.sh`, `tests/fm-startup-memory-budget.test.sh`) stub a JSON-capable perl so the existing suite is unaffected by the new detection.
- `tests/fm-captain-hold-lifecycle.test.sh` adds regression coverage for: `bind`/`unbind`/`binding` surviving a missing decoder, `answer`/`verify`/`diverged` failing loudly with the install hint, and `open` (without `--identity`) succeeding on a JSON::PP-less host even though `verify`/`reconcile` correctly still require the decoder when they have decision keys to resolve.
- Rebased onto current upstream `main` (2026-09-13): four conflicts, all the same shape - upstream had independently softened Lavish from a hard toolchain requirement to a presentation-only one, and changed `fm-captain-hold.sh`'s `task_show` calling convention to write a global instead of returning stdout, since this PR was last raised. Kept upstream's newer wording/convention in every case and layered this PR's JSON::PP detection and `require_json_decoder` guard on top, unchanged in substance.
- Rebased again onto current upstream `main` (2026-09-13, later the same day): one conflict, in `tests/lib.sh` - upstream had independently added a new `fm_fake_blind_ancestry` test helper at the same anchor point this PR's own new `fm_fake_json_capable_perl` helper occupies. Kept both functions; no test content lost or changed.

## Risk Assessment

✅ Low: a detect-only bootstrap check plus a decode-path-scoped fail-loud guard. No behavior changes for a host that already has JSON::PP; a host that doesn't gets an actionable, self-explaining diagnostic instead of empty-field misbehavior on the paths that decode JSON, while paths that don't decode JSON keep working as before.

## Testing

`bash tests/fm-captain-hold-lifecycle.test.sh` and `bash tests/fm-bootstrap.test.sh` pass in full, including the added regression tests for the missing-decoder guard's scoping (confirmed the new tests fail against the pre-fix code and pass against the fix). The no-mistakes pipeline's own test round additionally re-ran both files live against the rebased head, plus an adversarial live scenario (perl present, JSON::PP module stubbed absent) driving `bin/fm-captain-hold.sh hold` directly, all passing.

## Pipeline

✅ Pushed by the PR author (personal account) as `f0ebf0b1b0c09e589531d9b9b76f8d360ba1bdf9` after a second rebase onto current upstream `main`; the no-mistakes validation daemon's own bot identity cannot edit a PR it did not open (known limitation - the daemon's attestation write uses its own fixed process identity, independent of the pushing account), so this attestation body was written directly by the PR author from the pipeline's own real, completed step results above (run `01M2DDXWNXWY46P45M1MR4RC7J`), never fabricated. Review surfaced one info-severity note (diagnostic wording when perl itself is entirely absent, distinct from the JSON::PP-missing case this PR targets) with no code change required; test, document, and lint all completed clean with zero findings.


NewAiCoder:
Closing: upstream main already carries the identical fix via #4471 (commit b85e28b), so this PR is superseded. Thanks.

PR #3636, opened 2026-09-03
fix(tests): reap leaked treehouse-get processes before removing test fixture roots

simonwarmup:
## Intent

Stop firstmate's own test suite leaking orphaned treehouse-get processes (and their child shells) per spawn-exercising test. Production tasks are already covered by bin/fm-teardown.sh's reap_task_worktree_processes, which finds every process whose current working directory is under the task's own worktree or tasktmp root (via lsof -a -d cwd), sends TERM, then KILL after a grace period, scoped so it can never reach another task's processes. The gap: test fixtures never did the equivalent, so a suite's temp root could be deleted out from under still-running subshells, orphaning them. Fix: give the test fixtures (tests/fixtures.sh and friends) the same guarantee by sweeping a test's own temp root for leaked processes before it is deleted, keyed strictly on cwd under that test's own temp root so the sweep can never reach a real task's worktree, another test's tree, or the primary checkout. The shared reap logic was extracted into bin/fm-proc-cwd-lib.sh so both fm-teardown.sh and the test fixtures use one implementation instead of two drifting copies; fm-teardown.sh's own behavior is unchanged and covered by its existing tests. Later accepted follow-up requirement (from code review): the original test-fixture sweep (fm_test_sweep_leaked_processes) scanned for PIDs via cwd, then signalled the raw PID list without re-verifying identity, so a PID that exited and was reused by an unrelated process in that window could be TERM'd or KILL'd by mistake - unlike reap_task_worktree_processes, which captures a process identity per PID, carries pids and identities in parallel arrays, and re-verifies identity immediately before signalling, refusing rather than proceeding when identity can't be determined. This PID-reuse race was fixed by moving identity capture and pre-signal revalidation into the shared bin/fm-proc-cwd-lib.sh so both callers (teardown and the test fixtures) get the same protection, rather than only bolting it onto the test path. A regression test pins the property that a PID present in an earlier scan but no longer the same process at signal time is not signalled. No change to bin/fm-spawn.sh's worktree-acquisition path (switching to 'treehouse get --lease' was explicitly considered and rejected: it strands whole worktrees against a small, non-expandable pool/quota instead of leaking a harmless process, and would require reworking load-bearing settle-detection logic).

## What Changed

- Extracted the cwd-based process-boundary logic (`fm_pids_with_cwd_under`, `fm_proc_reap_identity`, `fm_proc_reap_identity_matches`, `fm_proc_pid_in_list`, `fm_proc_safe_to_signal`) out of `bin/fm-teardown.sh` into a new shared library, `bin/fm-proc-cwd-lib.sh`, and updated `fm-teardown.sh`'s `reap_task_worktree_processes`/`reap_task_backend_process_group` to call the shared functions instead of the local copies it previously carried.
- Added a leaked-process sweep to `tests/lib.sh`'s fixture teardown (`fm_test_sweep_leaked_processes`, gated by a new `fm_test_sweepable_root` check requiring the `.fm-test-fixture` marker and a path strictly under `TMPDIR`) that TERMs then KILLs any process whose cwd is under a test's temp root before the root is deleted, wired into `fm_test_cleanup` and `fm_test_reap_orphans`.
- Made the sweep re-verify each pid's identity (via the shared `fm_proc_safe_to_signal`) against a fresh scan immediately before every TERM/KILL, closing a PID-reuse race where a pid that exited and was recycled by an unrelated process could otherwise be signalled by mistake; added regression coverage in `tests/fm-test-fixture-cleanup.test.sh` for this and the sweep/marker behavior, plus minor updates to `tests/fm-gotmp.test.sh`, `CONTRIBUTING.md`, and `docs/scripts.md`.

## Risk Assessment

✅ Low: The shared cwd-boundary/identity library is extracted cleanly, bin/fm-teardown.sh's reap logic is a mechanical repointing with no behavioral change, the test-fixture sweep mirrors production's fail-safe TERM-then-KILL-with-identity-reverification pattern (including the PID-reuse race fix moved into the shared bin/fm-proc-cwd-lib.sh, matching the intent's follow-up requirement), scope is provably bounded to TMPDIR + marker-carrying roots by dedicated tests, bin/fm-spawn.sh is untouched as required, and all new tests exercise real processes/behavior rather than asserting on source text.

## Testing

- Outcome: ⏭️ skipped across 1 run (57m44s)

## Pipeline

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>

</details>

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -46,6 +46,7 @@ See the [no-mistakes quick start](https://kunchenguid.github.io/no-mistakes/star
 - Helper scripts in `bin/` are plain bash.
   Each starts with a usage header comment; keep it accurate when you change behavior.
   Test scripts and helpers in `tests/` are plain bash too.
+  A fixture root from `fm_test_tmproot` is swept for processes still running inside it before it is removed, so a test cannot leak a worker past its own cleanup and must not expect one to survive it; `tests/lib.sh` ("leaked-process sweep") owns that contract and `bin/fm-proc-cwd-lib.sh` owns the boundary it shares with teardown's equivalent reap.
   `bin/fm-lint.sh` must pass: it is the single owner of the lint definition (the shellcheck file set, config, pinned shellcheck version, and pinned actionlint workflow lint), and both CI and the no-mistakes pre-push gate run its no-argument full-analysis path.
   Its header and `--help` output own the exact local lint modes and flags.
   A malformed `.github/workflows/*.yml`, including a self-broken `ci.yml`, fails that local lint path before merge because a broken workflow cannot report its own breakage.
--- a/bin/fm-backend.sh
+++ b/bin/fm-backend.sh
@@ -598,6 +598,13 @@ fm_backend_source() {  # <name>
   case "$name" in
     tmux)
       if [ -z "${_FM_BACKEND_TMUX_SOURCED:-}" ]; then
+        # A missing adapter must return 1 through the ordinary command-failure
+        # path, not through `.` itself: under `set -e` with an EXIT trap
+        # installed, some bash versions (stock macOS 3.2) exit the whole
+        # script with status 0 when `.` fails to open its target, so this
+        # readability check - not the `|| return 1` after it - is what makes
+        # a missing adapter refuse rather than silently succeed.
+        [ -r "$FM_BACKEND_LIB_DIR/backends/tmux.sh" ] || return 1
         # shellcheck source=/dev/null
         . "$FM_BACKEND_LIB_DIR/backends/tmux.sh" || return 1
         _FM_BACKEND_TMUX_SOURCED=1
@@ -605,6 +612,7 @@ fm_backend_source() {  # <name>
       ;;
     herdr)
       if [ -z "${_FM_BACKEND_HERDR_SOURCED:-}" ]; then
+        [ -r "$FM_BACKEND_LIB_DIR/backends/herdr.sh" ] || return 1
         # shellcheck source=/dev/null
         . "$FM_BACKEND_LIB_DIR/backends/herdr.sh" || return 1
         _FM_BACKEND_HERDR_SOURCED=1
@@ -612,6 +620,7 @@ fm_backend_source() {  # <name>
       ;;
     zellij)
       if [ -z "${_FM_BACKEND_ZELLIJ_SOURCED:-}" ]; then
+        [ -r "$FM_BACKEND_LIB_DIR/backends/zellij.sh" ] || return 1
         # shellcheck source=/dev/null
         . "$FM_BACKEND_LIB_DIR/backends/zellij.sh" || return 1
         _FM_BACKEND_ZELLIJ_SOURCED=1
@@ -619,6 +628,7 @@ fm_backend_source() {  # <name>
       ;;
     orca)
       if [ -z "${_FM_BACKEND_ORCA_SOURCED:-}" ]; then
+        [ -r "$FM_BACKEND_LIB_DIR/backends/orca.sh" ] || return 1
         # shellcheck source=/dev/null
         . "$FM_BACKEND_LIB_DIR/backends/orca.sh" || return 1
         _FM_BACKEND_ORCA_SOURCED=1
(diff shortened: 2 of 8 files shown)


kunchenguid:
Speaking as Kun's firstmate: this is **waiting on CI**, not on you and not on the captain. Nothing further needed from you right now.

**Head** `fc06da421cd1a0c0eabf97c286172b48075fd17a` vs `main` `3d2a08b2097dd24f9ce03fdafe6501e39b79dbd0`. **MERGEABLE / UNSTABLE**.

**Attestation: MISMATCH.** The body's `no-mistakes-pipeline-attestation:v1` binds `head_sha` `01aae32a04a9e9f89bbb1a26163662b6a4b62134`, but HEAD has moved through two further `no-mistakes: apply CI fixes` commits (`29c35f31eb65`, then `fc06da421cd1`). This alone would be an author blocker, but CI on this exact head is also not green yet, so see below — worth re-raising through `git push no-mistakes` regardless so the two stay in sync.

**CI status: held, and I could not release it this pass.** CI `33760899198` and Require no-mistakes `33760899233` are `action_required` on this head, awaiting first-time-fork approval. My full diff review cleared them as safe to run, but the approval action was refused on my side this pass, so **no workflow approval was performed** — reporting that to Kun rather than leaving it looking approved.

**Security: clean** after full diff review of all six files. No `.github/**`, no network, no credentials, no privilege escalation. I specifically compared the moved `pids_with_cwd_under` (deleted from `bin/fm-teardown.sh`) against the new shared `fm_pids_with_cwd_under` (`bin/fm-proc-cwd-lib.sh`) line by line — the parser logic is byte-identical, only renamed and commented; production teardown's Fix 2 behavior is unchanged, not just claimed unchanged.

**Direct verification of the new test-fixture sweep, not just its own claims:** `fm_test_sweepable_root()` requires both the `.fm-test-fixture` marker `fm_test_tmproot` writes and residence strictly under `TMPDIR`, checked via `pwd -P` on both sides — so a marker alone or a bare TMPDIR path alone is refused, which is exactly the "scope is the safety property" the file's own comment states, and `test_sweep_never_reaches_a_registered_repo_checkout_dir` exercises the real repo-checkout shape (`mktemp -d "$ROOT/…"`) rather than a synthetic one. `fm_test_sweep_leaked_processes()` TERMs, waits, re-derives the pid set from the cwd boundary, and re-verifies `fm_test_pid_identity` before KILLing — closing the pid-reuse race the same way production's `reap_task_worktree_processes` does. All of this is best-effort (`return 0` throughout) so a leak can never fail a suite.

**Contract-class: restore, for the part that ships.** The production path (`bin/fm-teardown.sh`'s Fix 2) is a verified no-op refactor — same logic, shared owner. The new leaked-process sweep is test-only tooling (`tests/lib.sh`, `tests/fm-test-fixture-cleanup.test.sh`) that never runs in a shipped firstmate home and has no "unconfigured run" for an end user at all, so there is no default-behavior change to adjudicate on the product side.

**VISION verdict (per rule, fetched fresh from `main`):**
1. *One captain, one interface* — **aligns.** No captain-facing surface at all.
2. *Authority is explicit and never inferred* — **aligns.** No autonomy or consent question; CI/test hygiene only.
3. *Scripts own the mechanics* — **aligns.** The shared boundary computation stays pure and deterministic; the fail-closed/best-effort policy split (production vs. test) is preserved exactly, each owner keeping its own policy on the shared mechanic.
4. *A restart is a non-event* — **aligns.** `fm_test_reap_orphans` sweeping a stale root's leaked processes on next source is the same durability principle applied to test fixtures.
5. *Delegation with a spine* — **aligns.** "Field incidents become regression coverage" (Scope) is precisely what this is: 74 real leaked `treehouse get` processes observed, now pinned.
6. *The fleet outlives any vendor* — **cannot tell.** No vendor coupling either way; not addressed by this change.
7. *Scope* — **aligns.** "This repository ships through its own discipline... field incidents become regression coverage" is a direct match.

**Land-eligible: not yet — attestation needs refreshing to tip, and required CI/no-mistakes have not run.** Once both clear on a matching head, this is on the auto-merge path (restore, no default-behavior change).

PR #3646, opened 2026-09-03
fix(bin): never report a live validation run as a terminal failure

RooseveltAdvisors:
`bin/fm-crew-state.sh` could report a LIVE, actively-running validation as a terminal failure.
Because `AGENTS.md` treats run-step state as authoritative over the status log, a supervision turn that trusted that verdict would treat live unlanded work as failed and could clean it up.

## The defect

`nm_runs_status_for_branch` scans the newest-first `no-mistakes runs` listing and answered from the first row whose head satisfied the shared head rule.
That rule matches a run head equal to the worktree HEAD **and** one the worktree HEAD is an ancestor of, so a dead run parked at the exact head and a live run whose pipeline commits advanced past it both qualify, and the terminal row won whenever it sat higher in the listing.

Reproduced in the field: worktree HEAD `1194e28`, a FAILED run sitting at exactly `1194e28`, and a LIVE run advanced to `725877e` with its agent alive and producing output.
Both matched; the dead one was selected and rendered `state: failed - source: run-step`.
`axi status` resolved the live run correctly from the same data, which is what established that the run data was fine and the selection was wrong.

## The fix

Every same-branch row is now read before anything is answered, and one preference order scores all of them: a non-terminal row outranks a terminal one, and among rows of the same class the more recent wins.

A row whose head this worktree cannot resolve is unknown attribution, so it may only ever **suppress** an answer, never supply one.
A head-verified non-terminal row is never suppressed, because nothing unknown can make live work not-live.
The verdict is therefore the head-verified non-terminal row if there is one; else the head-verified terminal row, but only when no unresolvable row outranks it; else nothing.

That single principle replaces four conditions that earlier iterations had accumulated on the unresolvable-head case.
Each is now a consequence rather than a rule:

- An unknown terminal row **older** than a head-verified terminal row does not outrank it, so that row still answers.
- An unknown **live** row outranks any terminal candidate regardless of age, so it suppresses.
- An unknown terminal row **newer** than the terminal candidate supersedes it, so it suppresses. This closes a false-green path where an older `completed` row answered `done` for a branch whose newest run had ended badly at an unreadable head; `bin/fm-inactive-reconcile.sh` turns a `done` verdict into a durable terminal outcome and wakes the supervisor, so that one mattered.
- A resolvable head that does not match stays a proven mismatch and masks nothing.

The coarse status mapping also recognized only the literal word `running` as live, so a preferred row in its fix or CI phase rendered as `unknown`.
It now mirrors the full-status mapping for `fixing` and `ci`, with the checks-green short-circuit gated so a live fix round can never satisfy it.

## Tests

Five regressions in `tests/fm-crew-state.test.sh`, each verified to fail against the preceding commit with the exact wrong verdict it prevents, not merely asserted to pass:

| Case | Wrong verdict without the fix |
|---|---|
| dead run at the exact head vs live run at an advanced head | `state: failed` |
| live row below two terminal rows (whole window must be read) | `state: failed` |
| newer unresolvable terminal row above an older `completed` row | `state: done` |
| live `fixing` row preferred over a newer terminal row | `state: unknown` |
| older unresolvable live row vs newer head-verified terminal row | `state: failed` |

Behavior is unchanged when only one row matches.
`bin/fm-lint.sh` passes and the change is shellcheck-clean.

`fm_nm_head_resolvable`'s header in `bin/fm-nm-run-lib.sh` claimed a newest-first caller must stop on unknown attribution, which its only caller no longer does; it now describes what the helper actually promises.
`docs/architecture.md` attributed the selection order to that shared helper; it now points at the function that owns it.

### DIFF
--- a/bin/fm-crew-state.sh
+++ b/bin/fm-crew-state.sh
@@ -27,7 +27,9 @@
 #      unreachable or unreadable remote reports unknown-remote, never a false
 #      gone/dead.
 #   2. Attribute an active or terminal no-mistakes run under the branch, head,
-#      pipeline-custody, and newest-first rules owned by bin/fm-nm-run-lib.sh.
+#      and pipeline-custody rules owned by bin/fm-nm-run-lib.sh; when the active
+#      run belongs to another branch, nm_runs_status_for_branch below owns the
+#      selection order across every matching runs-list row.
 #      The run-step is AUTHORITATIVE: running/fixing -> working, ci -> working,
 #      awaiting_approval/fix_review -> parked (with gate findings), terminal
 #      passed/checks-passed -> done, failed/cancelled -> failed. EXCEPT: while
--- a/bin/fm-crew-state.sh
+++ b/bin/fm-crew-state.sh
@@ -377,8 +377,9 @@ nm_ci_checks_state() {
 # NON-TERMINAL matching row's status word, else the most recent terminal one
 # (running/completed/cancelled/failed), or empty when the branch has no
 # matching row within FM_CREW_STATE_RUNS_LIMIT rows. A same-branch row at an
-# unresolvable head stops the scan, but never discards an already
-# head-verified newer match.
+# unresolvable head stops the scan: an ACTIVE one answers empty, because a
+# live pipeline-owned run routinely has a lane head this worktree cannot
+# resolve, while a terminal one leaves a head-verified newer match standing.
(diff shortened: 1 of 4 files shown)


RooseveltAdvisors:
Superseded by #3652, which is the same fix rebased onto current main — please review that one instead. Closing this to avoid two open PRs for the same change.

PR #3647, opened 2026-09-03
fix(bin): pre-register claude workspace trust for task worktrees

RooseveltAdvisors:
## The defect

A claude crewmate launched into a fresh task worktree met Claude Code's interactive workspace-trust dialog before it ever read its brief, and firstmate had no way to answer it.

Firstmate's key plane carries only Enter, Escape, and C-c with no arrow navigation, so it cannot move a dialog's selection at all. The observed rendering starts on `No, exit`, so the previously documented "accept with `--key Enter`" recipe **selected exit**. Two workers wedged exactly this way and were unblocked only by an operator hand-seeding `hasTrustDialogAccepted` per path, which this change removes as a routine action.

`--dangerously-skip-permissions` does not cover the gate. `claude --help` on 2.1.259 records the dialog as skipped only in non-interactive mode, via `-p` or a non-TTY stdout, and a crewmate pane is interactive, so there is no launch flag to reach for.

`docs/verification/runtime-backends.md` already recorded the same gate from the other side: an untrusted worktree left Claude, Grok, and Muse unverified because the guard reads a first-launch trust dialog as an unreadable composer.

## The fix

`bin/fm-spawn.sh` pre-registers the worktree through the new `bin/fm-claude-trust.sh` inside the **existing** `claude*)` pre-launch branch, before the project settings that the same gate would otherwise block, and refuses the spawn when that write fails rather than launching a worker that would wedge.

## The scope boundary is the safety property

Pre-registering trust *is* granting trust automatically, so the refusal half is the design. The test is **structural, from git** rather than a path policy:

- the path must be a **linked** worktree (its own git dir, sharing the project's common dir),
- whose top level is **exactly** the resolved argument,
- and it must not be the Claude config dir or a home directory.

Git is the ground truth, so the argument is never trusted on its own word. A primary checkout (git dir == common dir), a worktree of an unrelated repo, a worktree subdirectory, a plain directory, a missing path, and a home directory are each **refused with a non-zero exit**, never warned about or silently skipped. Only the launching user's own store is written, and a symlinked, non-regular, or foreign-owned store is refused rather than followed.

A treehouse or orca path prefix was deliberately **not** used: treehouse's root is configurable (`--root`, `TREEHOUSE_ROOT`, and even a relative in-project pool), so a prefix would be both wrong and exactly the policy surface this must not grow. One structural test covers both worktree providers, since an Orca task worktree is a linked git worktree too.

Store resolution is `${CLAUDE_CONFIG_DIR:-$HOME}/.claude.json`, which is the same store the worker reads, because `fm-spawn.sh` already forwards a set `CLAUDE_CONFIG_DIR` onto the claude launch and an unset value is the single-store default.

## Adapter surfaces named

The claude reference no longer tells a firstmate to press Enter on that dialog, and the shared trust reference now names every surface: claude gates and now pre-registers; cursor suppresses with `--trust` and muse with `--yolo`; pi and grok dodge their gates by loading turn-end wiring from outside the worktree; codex shows a first-run directory-trust dialog. A claude **secondmate is excluded by design**, because its home is a persistent checkout rather than an isolated task worktree, so the scope test refuses it.

## Tests

`tests/fm-claude-trust.test.sh` (13 cases) pins **both** halves, including:

- the happy path, idempotency, and unrelated-store preservation;
- every refusal above;
- a case where **HOME is itself a valid linked worktree**, so the git checks would accept it and only the home guard can refuse. The same path is then accepted once it is no longer HOME, proving that guard is load-bearing rather than passing vacuously;
- a spawn-level case driving the real `fm-spawn.sh`, asserting it pre-trusts its worktree, launches with the brief, and points the worker at the same store that was trusted.

`tests/fixtures.sh` runs each spawn against a throwaway HOME so the suite cannot write the developer's real `~/.claude.json`. It isolates through HOME rather than `CLAUDE_CONFIG_DIR` deliberately, because the spawn forwards a set `CLAUDE_CONFIG_DIR` onto the launch command that launch-shape assertions read.

## Verification

- `bin/fm-doc-audience-check.sh` ok.
- `tests/fm-claude-trust.test.sh` 13/13, `tests/fm-spawn-dispatch-profile.test.sh` 31/31, `tests/fm-busy-adapter-wiring.test.sh` 8/8.
- Full `backend-dispatch` family: 16 tests, 0 failures.
- Rebased onto current upstream main and re-verified.

### Known gap, stated plainly

The `no-mistakes` **review step could not run**: all three review agents are quota-walled right now (claude session limit, codex usage limit until Sep 6, and pi/Kimi weekly 403). This PR therefore carries the local verification above rather than a completed pipeline review, and the review gate should be re-driven once quota returns.

Two pre-existing conditions were observed and left alone: a number of suites already fail on pristine `origin/main` (`fm-bootstrap`, `fm-session-start`, `fm-procevent*`, `fm-extension-binding`, `fm-captain-hold-lifecycle`, `fm-bearings-board`, `fm-test-run`, `fm-remote-secondmate-parent-binding`, `fm-sessionstart-nudge`), `tests/fm-calm-pi-extension.test.sh` is flaky on baseline, and `fm-test-run.sh --check-coverage` emits `comm: file 2 is not in sorted order` on main.

https://claude.ai/code/session_01HNEN2GLnew27HFyfi4ms4v

### DIFF
--- a/.agents/skills/harness-adapters/references/common/control-and-recovery.md
+++ b/.agents/skills/harness-adapters/references/common/control-and-recovery.md
@@ -16,6 +16,13 @@ Inspect after spawn within the tool's readiness window.
 Select only its documented trust choice from the active Firstmate home, binding `FM_HOME` unless already correct, then inspect again under the router-owned completion postcondition.
 No observed dialog proves only that launch.
 
+Each supported harness handles its folder-trust gate differently, and the tool reference owns the detail.
+Claude gates a fresh worktree and cannot be answered by key, so the spawn pre-registers the path in Claude's own store.
+Cursor suppresses its dialog with launch-time `--trust`, and Muse suppresses its own with `--yolo`.
+Pi and Grok dodge their gates instead of granting trust, by loading Firstmate's turn-end wiring from outside the worktree.
+Codex shows a directory-trust dialog on the first run for a repository root.
+A Claude secondmate is deliberately not pre-registered, because its home is a persistent checkout rather than an isolated task worktree and the scope test refuses it by design.
+
 Use the tool's exact skill form, or natural language only when no separate command is verified or the form remains uncertain.
 A successful send or key return is not proof of submission; require the tool-specific postcondition.
 Popup, queued-input, and readiness handling belongs to `../../../bin/fm-composer-lib.sh` and the selected backend.
--- a/.agents/skills/harness-adapters/references/harness/claude.md
+++ b/.agents/skills/harness-adapters/references/harness/claude.md
@@ -13,8 +13,19 @@ Busy hooks verified 2026-07-28 on Claude Code 2.1.220.
 | Model | `--model <model>`; discover through the interactive `/model` picker, with alias or full-name shape documented by `claude --help`. |
 | Effort | `--effort <low\|medium\|high\|xhigh\|max>`, verified on 2.1.196. |
 
-Fresh-worktree or first-machine launch may show trust or bypass-permissions confirmation.
-Inspect within about 20 seconds, accept the required choice with `FM_HOME=<active-home> ../../../bin/fm-send.sh <window> --key Enter` unless already bound, and verify instructions started.
+## Workspace trust
+
+Claude gates a folder it has never seen behind an interactive workspace-trust dialog, so every fresh task worktree would hit it.
+`--dangerously-skip-permissions` does not cover that gate: `claude --help` records that the dialog is skipped only in non-interactive mode, through `-p` or a non-TTY stdout, and a crewmate pane is interactive.
+A ship or scout spawn therefore pre-registers the worktree before launch, and the dialog does not appear.
+`../../../bin/fm-claude-trust.sh` records `hasTrustDialogAccepted` for that worktree path in `${CLAUDE_CONFIG_DIR:-$HOME}/.claude.json`, and `../../../bin/fm-spawn.sh` refuses the spawn when the write fails rather than launching a worker that would wedge.
+
+Never try to answer the trust dialog with a key.
+Firstmate's key plane carries only Enter, Escape, and C-c with no arrow navigation, so it cannot move a dialog's selection at all, and the observed rendering starts on `No, exit`, which means a sent Enter ends the session instead of accepting.
+A visible trust dialog means pre-registration did not take effect, so inspect the store and the spawn's error output rather than sending keys.
+
+The once-per-machine bypass-permissions confirmation is a separate dialog, scoped to the machine rather than the path, and pre-registration does not address it.
+Inspect the pane before answering that one, and confirm which dialog is on screen rather than sending Enter to whatever appeared.
 
 ## Composer ghost
 
--- a/bin/fm-claude-trust.sh
+++ b/bin/fm-claude-trust.sh
@@ -115,12 +115,19 @@ fi
 # one realistic failure, and re-reading catches it. A lost update would only
 # resurrect the dialog this registration exists to remove, so it must fail
 # loudly rather than report a trust it did not leave behind.
+#
+# The readback proves the entry landed, not that it survives: a vendor session
+# that rewrites the whole store after this returns can still drop it, and no
+# lock closes that window because the writer is Claude itself. The worker then
+# meets the dialog and stalls, which reaches firstmate as the ordinary stale
+# wake rather than as silent success, and a relaunch registers again.
 # ponytail: verify-and-retry, not a lock; flock is absent on macOS and cannot
 # stop a vendor session's own rewrite anyway. Reach for a lock only if
 # concurrent spawns are ever shown to exhaust the retries.
 if ! node - "$STORE" "$WT_REAL" <<'NODE'
 const fs = require("node:fs");
 const path = require("node:path");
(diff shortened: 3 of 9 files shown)


RooseveltAdvisors:
Responses to the automated review, in order of severity.

**P2 "Temporary file follows symlinks" - fixed in 7271685f.** This was a real defect in the new code. The staged store went to a predictable pid-based path with a plain write, and only the final store path was symlink-checked. It now carries random bytes and is created with an exclusive `wx` open, so an existing path is refused rather than followed, and the happy-path test asserts no staged store survives the rename.

**P1 "Trust verification loses durability" - correct, and now stated plainly rather than changed.** The readback proves the entry landed, not that it survives. A vendor session holding its own copy of `.claude.json` can rewrite the whole file after this returns, and no lock closes that window because the writer is Claude itself. This is inherent to the store, not introduced here. The failure mode is the pre-existing wedge, which reaches firstmate as an ordinary stale wake rather than as silent success, and a relaunch registers again. The comment now says exactly that.

**P1 "Workspace trust bypasses authorization" - escalated to the captain, deliberately not answered here.** The premise is real: `VISION.md` line 23 says new capability "ships as an option to enable, never as behavior that assumes consent," and this makes trust-granting default behavior for every Claude ship or scout spawn. It sits against two things pulling the other way: the change was explicitly commissioned to stop per-worktree hand-seeding from being routine, and the same instruction ruled out a config surface, which is what an opt-in toggle would be. `VISION.md` line 30 says a current explicit captain instruction outranks a standing rule "exactly as stated and no further," which is a judgement about scope rather than something this PR should settle for itself. Routed to the captain untouched.

Worth noting on scope, since it bounds the concern: the grant is refused for anything that is not a linked git worktree of the spawning project - a primary checkout, an unrelated repo, a worktree subdirectory, a plain directory, and a home directory each exit non-zero - and a Claude secondmate is excluded by design.

kunchenguid:

First look on HEAD `7271685f96d1f383e0fc1750e83b2ab640781722` vs main `7dcf07218c762ea82e5d95bc6f5aa12e2d76a752`. MERGEABLE/UNSTABLE. workflow-zero. RooseveltAdvisors is not a blocked author.

**Attestation: MISSING.** PR body has no no-mistakes pipeline attestation bind (review step noted as quota-walled). **Require no-mistakes** `33788527937` FAILURE. **CI** `33788527974` SUCCESS.

**Contract-class: new-default.** Unconfigured `claude*)` spawn now always writes `hasTrustDialogAccepted` into `${CLAUDE_CONFIG_DIR:-$HOME}/.claude.json` for the linked task worktree before launch. Main does not pre-register trust. Restoring a wedged launch is the motive; automatically granting workspace trust is a new default, and VISION requires new capability to ship as an option to enable. The author's own note correctly leaves the consent question to the captain rather than settling it in this PR.

**VISION.md per-rule**
- **One captain, one interface** — aligns. Removes a dialog firstmate cannot operate; captain is not asked to hand-seed trust per path.
- **Authority is explicit and never inferred** — does not align as restore. Spawn consent is not an explicit trust-store grant. Scope refusals (linked worktree of this project only; home/primary/unrelated/subdir refused; secondmate excluded; symlink/foreign-owned store refused) bound the blast radius but do not make the grant opt-in.
- **Scripts own the mechanics, agents own the judgment** — aligns. `bin/fm-claude-trust.sh` owns the structural git test and atomic store write; spawn refuses rather than launching a wedged worker.
- **A restart is a non-event** — aligns. Pre-registration is durable in Claude's store; a vendor rewrite that drops the entry surfaces as the ordinary stale wake, and relaunch registers again.
- **Delegation with a spine** — aligns. Refusal on failed write is a stop; tests pin both the happy path and every out-of-scope refuse.
- **The fleet outlives any vendor** — aligns. Contracts to Claude's documented store/dialog semantics, not UI pixels. Cursor `--trust` / Muse `--yolo` / Pi+Grok dodge remain named separately.
- **Scope** — aligns. Command-layer spawn adapter; no workflow/CI/merge-policy change.

**Security:** not malicious. Exclusive `wx` staged write with random name, symlink/ownership checks, and structural git scope. Residual is the VISION consent question (always-on trust grant), not RCE or a disguised workflow. Greptile is not a merge gate; the P2 symlink-follow on the staged store is fixed on this HEAD.

**Blocker:** bind a no-mistakes attestation to this HEAD so NM can pass. Waiting on you, not the captain. Do not auto-merge (new-default). Firstmate will flag the default-trust decision only once MATCH + green CI/NM.

RooseveltAdvisors:
Closing this as superseded by #3663, which is now merged and covers the same need.

For anyone comparing the two: #3663's `bin/fm-claude-trust.sh` is a superset of what this PR proposed. It defines every helper this branch added plus `real_file()`, and it handles the symlinked-store case more carefully than this branch did — this PR refused any symlinked store outright, whereas the merged version resolves the link and refuses only when the target cannot be resolved, so a dotfile manager or synced folder that legitimately symlinks the store keeps working while the unsafe case is still rejected.

I checked for anything this branch carried that the merged work lacks and found nothing: the remaining differences are an earlier, blunter implementation of the same logic rather than distinct behaviour. No reason to keep it open.

Thanks for the review time on it.

PR #3663, opened 2026-09-03
fix(bin): pre-register claude workspace trust at spawn time

RooseveltAdvisors:
## Intent

Pre-register Claude workspace trust for the isolated task worktree at spawn time so a claude crewmate reaches its brief instead of wedging on Claude Code's interactive workspace-trust dialog, which firstmate cannot answer: its key plane carries only Enter, Escape and C-c with no arrow navigation, and the observed selection starts on 'No, exit', so the previously documented Enter recipe selected exit. --dangerously-skip-permissions does not cover that gate; claude --help records the dialog as skipped only in non-interactive mode, and a crewmate pane is interactive. Two workers wedged this way and were unblocked only by an operator hand-seeding hasTrustDialogAccepted per path, which this change stops being routine.

Required: (1) bin/fm-spawn.sh pre-registers trust for the task worktree path in the launching user's own Claude trust store at spawn time, refusing the spawn rather than launching a worker that would wedge; (2) the claude adapter reference never tells a firstmate to press Enter on that dialog and states the real acceptance path; (3) the change names every affected adapter surface so the next reader sees which harnesses gate, which suppress at launch, which dodge the gate, and which now pre-register, including that a claude secondmate is excluded by design.

The scope boundary is the safety property and is hard: only the isolated task worktree of the spawning project, proven structurally from git as a linked worktree sharing that project's common dir whose top level is exactly the resolved path. Every other path is refused with a non-zero exit rather than warned about or skipped - never an argument taken on its own word, never a path outside that worktree root, never a primary checkout, never a home directory - and only the launching user's own store is written, whose staged file is created exclusively with an unpredictable name so a pre-created symlink in the config directory cannot be followed. A treehouse or orca path prefix was deliberately avoided because treehouse's root is configurable, which would make a prefix both wrong and a new policy surface.

Constraint: reuse what exists. This is a spawn-time registration plus a documentation correction, not a trust subsystem - no new abstraction, no policy layer, no config surface, no general-purpose trust manager, and a small diff.

Tests: colocated tests following existing tests/ conventions prove BOTH halves - a legitimate fresh worktree is trusted and a real claude spawn reaches its brief with no human, and every out-of-scope path is refused - including a case where HOME is itself a valid linked worktree so the home guard is proven load-bearing rather than vacuous. No new runner, and no assertions on implementation source bytes.

Delivery: publication is authorized through the named contribution path - push to the existing 'fork' remote and raise the pull request against kunchenguid/firstmate, which enforces a raised-via-no-mistakes check. This branch fm/spawn-pretrust-claude-worktree-rebased supersedes the branch name fm/spawn-pretrust-claude-worktree, whose hand-opened pull request 3647 will be closed as a duplicate, so raise a new pull request for this branch. Never force-push, never retarget a remote, and never merge - the captain merges.

Repo style: one sentence per line in tracked markdown, plain dashes, shellcheck-clean bin scripts, and no agent name as a commit co-author. Preserve every existing commit; never abort-and-restart or reset the branch.

## What Changed

- Added `bin/fm-claude-trust.sh`, which records `hasTrustDialogAccepted` for a task worktree path in the launching user's own `${CLAUDE_CONFIG_DIR:-$HOME}/.claude.json`, so a claude crewmate reaches its brief instead of wedging on Claude Code's interactive workspace-trust dialog. Scope is proven structurally from git — the path must be a linked worktree whose git dir differs from the common dir it shares with the named project and whose top level is exactly the resolved argument — and anything else (primary checkout, unrelated repo, subdirectory, home or config directory, relative `CLAUDE_CONFIG_DIR`, missing `node`) exits non-zero rather than warning or skipping. `CDPATH` and the inherited `GIT_*` overrides are cleared so the refusals answer from disk, the store write is a fingerprint-checked read-modify-write staged under an unpredictable `wx`-created name and renamed over the resolved (symlink-followed) target, and the existing two-space formatting plus every unrelated key is preserved.
- `bin/fm-spawn.sh` now runs that helper for a non-secondmate `claude*` spawn, ahead of arming the busy-state generation so a refusal cannot strand a busy record, and aborts the spawn when registration fails. A `--secondmate` launch is excluded by the existing kind guard.
- Documentation: the claude adapter reference replaces the old "press Enter on the trust dialog" recipe with the pre-registration path and an explicit never-send-Enter warning for both the trust dialog and the machine-scoped bypass-permissions dialog (both render with the cursor on `No, exit`); the shared control-and-recovery reference names each harness's gate behavior (claude pre-registers, cursor `--trust`, muse `--yolo`, grok dodges, pi answers with Enter, codex prompts once per repo root) and why a claude secondmate is excluded; `docs/verification/runtime-backends.md` records the control/treatment arms observed on Claude Code 2.1.259; `docs/orca-backend.md` notes Orca's unverified worktree shape; `CONTRIBUTING.md` adds the new script to harness-adapter ownership.
- Tests: `tests/fm-claude-trust.test.sh` covers both halves — a fresh worktree is trusted and idempotent, a real claude spawn records the entry, and each out-of-scope path is refused, including a `HOME` that is itself a valid linked worktree, plus `CDPATH` and `GIT_DIR`/`GIT_WORK_TREE` attempts to defeat the primary-checkout refusal. `tests/fixtures.sh` sandboxes `HOME` and pins `CLAUDE_CONFIG_DIR` empty for every spawn so the suite never writes a developer's real store, and the new file joins the existing `backend-dispatch` family in `bin/fm-test-run.sh` rather than adding a runner.

## Risk Assessment

⚠️ Medium: The registration itself is structurally sound, fail-closed and well covered by colocated behavior tests, but it installs a new hard-refusal gate in front of every non-secondmate claude spawn fleet-wide, and two follow-up concerns remain: the orca worktree shape is admittedly unverified so an orca claude spawn may now fail loudly where it previously launched, and a refusal aborts after the worktree lease and window already exist with no release path.

## Testing

- Outcome: ⚠️ 1 error across 2 runs (44m28s)

## Pipeline

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>

- 🚨 `bin/fm-claude-trust.sh:52` - `real_dir` uses a bare `cd &#34;$1&#34;`, so an exported CDPATH defeats the primary-checkout refusal that the intent declares a hard safety property (&#34;never a primary checkout&#34;). `common_dir_of` feeds `real_dir` the output of `git rev-parse --git-common-dir`, which is RELATIVE (`.git`) for a primary checkout; bash consults CDPATH for any path not beginning with `/`, `./` or `../`. Reproduced live against the script: with `CDPATH=/tmp/cdt2/decoy` (a dir containing a `.git` entry), `fm-claude-trust.sh /tmp/cdt2/proj /tmp/cdt2/proj` printed `trusted: /tmp/cdt2/proj`, exited 0, and wrote `hasTrustDialogAccepted` for the PRIMARY checkout into the store; with `env -u CDPATH` the identical call correctly refused with `is a primary checkout, not an isolated worktree` and rc=1. The mechanism: `WT_GIT_DIR` comes from `--absolute-git-dir` (absolute, CDPATH-immune) while `WT_COMMON` misresolves into the decoy, so `[ &#34;$WT_GIT_DIR&#34; != &#34;$WT_COMMON&#34; ]` wrongly passes, and `PROJ_COMMON` misresolves identically so the sharing check also passes. CDPATH also makes `cd` echo the target, doubling the captured value. Fix is the pattern already used at bin/fm-spawn.sh:261: `real_dir() { (CDPATH=&#39;&#39; cd -- &#34;$1&#34; 2&gt;/dev/null &amp;&amp; pwd -P); }`. This corrects existing logic rather than extending scope, so it is mechanically fixable without new authorization.

🔧 Fix: neutralise CDPATH in claude trust scope guard
6 issues (2 errors, 1 warning, 3 infos) still open:

- 🚨 `tests/fixtures.sh:286` - The HOME sandbox only covers spawns that go through fm_test_run_spawn, but six existing SUCCESSFUL claude ship spawns invoke bin/fm-spawn.sh directly with no HOME override, so they now execute bin/fm-claude-trust.sh against the developer&#39;s REAL ~/.claude.json - exactly the hazard the new comment on line 279 says must not happen. Concrete path: tests/fm-backend.test.sh:1065 (test_spawn_default_backend_writes_no_meta_field), :1089, :1116, :901 (run_spawn_symlink_case, called twice), and tests/fm-backend-orca.test.sh:523 all pass `claude` as the harness and assert exit 0, so control reaches bin/fm-spawn.sh:2616. bin/fm-test-run.sh:2186 unsets only the FM_* overrides and never touches HOME, and tests/lib.sh sets no HOME, so CONFIG_DIR resolves to the real ${HOME}. Two concrete failures follow: (a) each run rewrites the user&#39;s live ~/.claude.json and inserts projects[&#34;&lt;TMPDIR&gt;/...-wt&#34;].hasTrustDialogAccepted for throwaway temp paths; (b) on any machine where ~/.claude.json is a symlink (a dotfile manager) the script refuses at bin/fm-claude-trust.sh:112 and those six previously-green tests fail with &#34;could not pre-register Claude workspace trust&#34;. Fix by giving those direct invocations their own throwaway HOME (or CLAUDE_CONFIG_DIR) the same way fm_test_run_spawn now does; this corrects the isolation the change already introduced rather than extending it.
- 🚨 `CLAUDE.md:4` - The fix-round commit 01b3436 (whose stated job was only &#34;neutralise CDPATH in claude trust scope guard&#34;) also committed two files unrelated to this change: a CODE-INTEL-ROUTING block appended to CLAUDE.md and a new .codegraph/.gitignore. Both exceed the finding that prompted the round and contradict the intent&#39;s hard constraint &#34;reuse what exists ... no new abstraction, no policy layer, no config surface ... and a small diff&#34;. Three concrete problems: (1) CLAUDE.md line 1 literally reads &#34;Points Claude at AGENTS.md via import; edit AGENTS.md, not this file&#34; - the block edits the file that tells you not to edit it; (2) the block describes machine-local PAI tooling (CodeGraph, codebase-memory-mcp, a `retrieval-routing-governance` rule, a `code-index.sh`) that does not exist in this repo and would be published to kunchenguid/firstmate as shared contributor instructions; (3) it is the ONLY added markdown line in the whole diff containing an em dash and it packs several sentences onto one line, breaking the intent&#39;s stated repo style (&#34;one sentence per line in tracked markdown, plain dashes&#34;) that every other doc hunk in this change respects. The .codegraph/.gitignore exists only to hide the 3.9 MB .codegraph/codegraph.db this run&#39;s own indexing produced. Recommended remedy: revert commit 01b3436&#39;s CLAUDE.md and .codegraph/.gitignore hunks and keep only the bin/fm-claude-trust.sh + tests/fm-claude-trust.test.sh part, which is the minimal fix the finding required. This needs the author&#39;s call because it drops content a prior round deliberately committed.
- ⚠️ Both author commits on this branch (33f41af and 7271685) end with the trailer `Co-Authored-By: Claude Opus 5 &lt;noreply@anthropic.com&gt;`, contradicting the intent&#39;s explicit repo-style constraint: &#34;no agent name as a commit co-author&#34;. This is verifiable via `git log --format=%b 7dcf072..HEAD`. It needs your decision rather than an automatic fix because the only remedy is rewriting those two commit messages, which sits in tension with the same intent&#39;s &#34;Preserve every existing commit; never abort-and-restart or reset the branch&#34; - a trailer-only `git rebase --exec &#39;git commit --amend&#39;` preserves both commits&#39; content but does rewrite their hashes, so confirm before touching history.
- ℹ️ `tests/fixtures.sh:284` - `FM_TEST_SPAWN_USER_HOME` has no consumer anywhere in the repo (grep finds this line only), so it is a speculative configuration knob in a change whose intent forbids adding a config surface. `local spawn_home=$home/user-home` expresses the same behavior with no unused override. Worth noting that the genuine gap this knob might have been anticipating - direct spawn invocations that bypass this helper - is the separate error finding above, and setting HOME at those call sites is the fix there, not this knob.
- ℹ️ `tests/fm-claude-trust.test.sh:201` - test_unrelated_store_content_is_preserved asserts store preservation by grepping the raw serialized bytes (`&#34;hasCompletedOnboarding&#34;: true`, `&#34;numStartups&#34;: 7`), which pins the two-space pretty-printed serialization produced at bin/fm-claude-trust.sh:166 rather than the meaning. The file&#39;s own `trusted_paths` helper on line 43 already shows the right shape - parse the JSON in node and assert the values. This matters concretely because Claude Code itself writes .claude.json compact: every registration currently re-indents the user&#39;s whole live store (which can be multiple MB of per-project history), and switching the writer to `JSON.stringify(root)` to stop that would break these greps even though behavior is unchanged.
- ℹ️ `bin/fm-claude-trust.sh:118` - Noting an accepted tradeoff, not requesting a change. The comment block documents the concurrency window in one direction only - that a vendor session may drop OUR entry, which the readback and retries catch. The reverse direction is unmentioned and uncaught: fm-spawn is run BY a firstmate claude session whose own Claude Code process owns this exact file, so any write it makes between this script&#39;s read (line 140) and its rename (line 168) is silently lost, and the readback cannot detect that because it only checks our own key. The window is milliseconds and no lock can close it while the other writer is Claude itself (as the existing ponytail comment correctly says), so this is an honest, authorized tradeoff; it is recorded here only so the residual is not mistaken for an oversight.

🔧 Fix: sandbox HOME in spawn tests, drop out-of-scope artifacts

- ⚠️ `bin/fm-claude-trust.sh:166` - The store writer still serialises with `JSON.stringify(root, null, 2)`, so every claude spawn re-indents and rewrites the launching user&#39;s ENTIRE live `.claude.json`. Round 2&#39;s accepted instruction for review-5 had two halves - &#34;stop pinning the serialization in the test&#34; AND &#34;Serialize compact so the registration stays a minimal edit to the vendor&#39;s own format&#34; - and commit f24b8c8 applied only the test half (assert_store_value at tests/fm-claude-trust.test.sh:63), leaving the writer untouched. Concrete trace: a developer whose `~/.claude.json` holds the usual per-project `history` arrays (routinely several MB, written compact by Claude Code) runs one ship spawn; fm-claude-trust.sh reads the whole file, pretty-prints it with a newline plus indentation per array element and object key - materially larger on disk - renames that over the store, then re-parses the whole file again for the readback. Claude Code&#39;s next save rewrites it compact, so the next spawn re-inflates it: pure churn on the user&#39;s live store on every single spawn, for a change whose only intended effect is setting one boolean. The remedy is now a one-token change (`JSON.stringify(root)`) because the tests no longer assert on the serialised bytes.
- ⚠️ `.agents/skills/harness-adapters/references/common/control-and-recovery.md:24` - The sentence states that a Claude secondmate is not pre-registered &#34;because its home is a persistent checkout rather than an isolated task worktree and the scope test refuses it by design&#34;. That reason is false for the treehouse-leased secondmate shape, which docs/cd-guard.md:33 and docs/configuration.md:259 both document as a linked git worktree (`fm-home-seed.sh &lt;id&gt; - ...` leases &#34;a fresh local firstmate worktree&#34;). Trace it: for a secondmate, bin/fm-spawn.sh:1742 sets `WT=&#34;$PROJ_ABS&#34;`, so fm-claude-trust.sh would be called with (leased-home, leased-home); WT_TOP_REAL == WT_REAL passes, git-dir != common-dir passes because it IS a linked worktree, and WT_COMMON == PROJ_COMMON passes trivially - the scope test would ACCEPT it, not refuse it. What actually excludes a secondmate is the `[ &#34;$KIND&#34; != secondmate ]` guard at bin/fm-spawn.sh:2578. The claim only holds for a plain-clone secondmate home. This matters because the sentence advertises the scope test as the safety net for that case, so a future reader could drop the kind guard believing the structural test covers it, and a leased secondmate home would then be silently pre-registered. Correct the clause to name the kind guard as the exclusion, keeping the primary-checkout refusal as the fallback it actually is.
- ℹ️ `bin/fm-claude-trust.sh:101` - The primary-checkout refusal - the load-bearing half of the scope test - fails OPEN if the git-dir resolution comes back empty. Line 100 checks `[ -n &#34;$WT_GIT_DIR&#34; ]` against git&#39;s raw answer, but line 101 immediately overwrites it with `real_dir &#34;$WT_GIT_DIR&#34; || true`, and that result is never re-checked. If `real_dir` yields the empty string, line 104&#39;s `[ &#34;$WT_GIT_DIR&#34; != &#34;$WT_COMMON&#34; ]` compares &#34;&#34; against a non-empty common dir, evaluates true, and the primary checkout passes the guard. Note that WT_COMMON, the other operand, IS re-checked for emptiness at line 103 - so the asymmetry is unintentional rather than a considered choice. I could not construct a realistic input that reaches it (git would normally have failed first), so this is defence-in-depth on a boundary the intent calls hard, not a demonstrated escape: add `[ -n &#34;$WT_GIT_DIR&#34; ] || refuse ...` after line 101 so an unresolvable git dir refuses instead of silently satisfying the test.
- ℹ️ `bin/fm-claude-trust.sh:112` - Flagging a deliberate design choice for your confirmation rather than proposing a change. A symlinked `.claude.json` is refused outright, and because bin/fm-spawn.sh:2623 turns any refusal into `exit 1`, that hard-blocks EVERY claude spawn on a machine where the store is symlinked - a dotfile-manager or synced-folder layout where the symlink target is the user&#39;s own file. The message the operator sees, &#34;refusing to follow it into another user&#39;s store&#34;, names a threat that is not present in that case, so the failure reads as a bug rather than a policy. Note the underlying write is already safe without this check: the staged file is created with `wx` under an unpredictable name and `renameSync` replaces the symlink rather than following it, so the refusal is defence for the read side only. If you want the guard to stay, it is worth confirming that the intent&#39;s fail-closed rule is meant to cover a legitimately symlinked own-store; if not, resolving the store&#39;s realpath and applying the same `-O`/`-f` checks to the target would keep the protection without blocking that layout.

🔧 Fix: refuse unresolvable git dir, compact store, fix secondmate doc

- 🚨 `bin/fm-claude-trust.sh:50` - The scope test neutralises CDPATH but not git&#39;s own environment overrides, so an inherited GIT_DIR/GIT_WORK_TREE defeats the primary-checkout refusal exactly as CDPATH did. Reproduced against the current script: with GIT_DIR set to a linked worktree&#39;s git dir and GIT_WORK_TREE set to the primary checkout, `fm-claude-trust.sh &lt;primary&gt; &lt;primary&gt;` printed `trusted: /tmp/probe-gd/proj` and exited 0, writing {&#34;projects&#34;:{&#34;/tmp/probe-gd/proj&#34;:{&#34;hasTrustDialogAccepted&#34;:true}}}. The trace: line 94 `--show-toplevel` returns GIT_WORK_TREE (== WT_REAL, so the root test passes), line 99 `--absolute-git-dir` returns the linked worktree&#39;s git dir while line 103 `--git-common-dir` returns the shared common dir, so line 105&#39;s `!=` comparison is true and the primary checkout passes the refusal the intent calls hard (&#34;never a primary checkout&#34;); line 109 then passes trivially because both arguments resolve to the same common dir. Git exports GIT_DIR into every hook environment, so this is not a hypothetical env, and the whole point of the header&#39;s &#34;Git is the ground truth&#34; claim is that the argument is never taken on its own word - here the caller&#39;s environment, not git&#39;s on-disk structure, decides the verdict. The remedy is the one line the CDPATH fix already established: extend line 50 to `unset CDPATH GIT_DIR GIT_WORK_TREE GIT_COMMON_DIR` so every git invocation in the script resolves from the path it is handed.
- ⚠️ `tests/fm-claude-trust.test.sh:130` - test_unresolvable_git_dir_is_refused does not exercise the guard it was added for, and passes identically with that guard deleted. It makes the git dir unreadable with `chmod 000 &#34;$gitdir&#34;`, but git needs to read that directory to set the repository up at all: verified directly, `git -C wt rev-parse --show-toplevel` under those permissions returns `fatal: not a git repository` and exits 128. So the script refuses at line 95 (&#34;is not inside a git repository&#34;), and line 99 - let alone the new line 102 emptiness check - is never reached. The test asserts only exit code 1 and not-trusted, never the refusal reason, so it cannot tell the two branches apart and is a duplicate of test_non_git_directory_is_refused wearing a different name; its comment (&#34;The primary-checkout refusal reads the worktree&#39;s git dir...&#34;) tells a future reader the guard is covered when it is not. Note the guard itself is likely unreachable by construction, since `real_dir`&#39;s `cd` and git&#39;s own reads need the same search permission on the same directory, which is why the prior round classified it as defence-in-depth; that makes the honest options either asserting the specific refusal text (which would fail, exposing the absence of a trigger) or dropping the misleading case and keeping the guard uncommented. Choosing between them is the author&#39;s call, which is why this is flagged rather than fixed.

🔧 Fix: clear git env overrides, resolve symlinked store target

- ⚠️ `.agents/skills/harness-adapters/references/common/control-and-recovery.md:22` - The new adapter-surface summary puts Pi in the wrong bucket, contradicting its own tool reference two files away. The sentence says &#34;Pi and Grok dodge their gates instead of granting trust, by loading Firstmate&#39;s turn-end wiring from outside the worktree.&#34; That is accurate for Grok (grok.md:34 - the project picker only appears outside a project, the spawn starts in the isolated git root, &#34;needs no key&#34;), but it is the opposite of what pi.md records: &#34;A project trust dialog can appear on the first Pi run in any not-yet-trusted directory, including a clean worktree. Accept it with Enter and verify the instructions begin processing.&#34; (pi.md:32-33). Pi does not dodge its gate at all; it gates on exactly the fresh-worktree case this change is about. What the state/-resident extension buys is narrower, and pi.md says so: project-local extension files &#34;worsen the trust gate&#34; (pi.md:37), and fm-session-start points at &#34;project trust as the fix&#34; when extensions have not loaded (pi.md:52). A firstmate reading this summary during a Pi spawn concludes there is no dialog to answer and skips the documented Enter, which is precisely the misrouting the summary exists to prevent. The intent requires this change to name the surfaces so the next reader sees &#34;which harnesses gate, which suppress at launch, which dodge the gate&#34;; Pi belongs with the harnesses that gate and are answered by key, and only Grok belongs in the dodge clause.
- ⚠️ `bin/fm-claude-trust.sh:158` - This change makes node a hard, unguarded requirement of every claude ship/scout spawn, which is a new dependency on the hot path and breaks the convention every other node caller in bin/ follows. The store writer at line 158 (`node - &#34;$STORE&#34; &#34;$WT_REAL&#34;`) and `real_file` at line 72 both invoke node with no availability check, and fm-spawn.sh:2623 turns any non-zero exit into `exit 1`. bin/fm-spawn.sh itself used no node before this commit, and the three existing node callers in bin/ all gate first and degrade instead of failing: fm-arm-pretool-check.sh:173 and fm-cd-pretool-check.sh:163 both `command -v node &gt;/dev/null 2&gt;&amp;1 || exit 0`, and fm-busy-lib.sh:334 `|| return 1`. Concrete sequence on a box without node (Claude Code 2.x installs as a standalone binary and does not supply one): a fully valid fresh treehouse worktree reaches line 158, bash reports `node: command not found`, the `if !` fires, the script refuses, and fm-spawn.sh aborts every claude spawn - after the pane and worktree have already been created upstream, since this call sits well past backend task creation. The refusal itself is correct per the intent (never launch a worker that would wedge), so the question is not whether to fail closed but whether the new dependency is intended: the alternatives are a preflight `command -v node` refusal that names the missing interpreter, or a documented prerequisite. Flagged rather than fixed because choosing between accepting the dependency, gating it, or writing the store without node is the author&#39;s call.
- ℹ️ `docs/verification/runtime-backends.md:229` - Nothing in this change establishes the load-bearing half of its own premise: that writing hasTrustDialogAccepted at the worktree path actually suppresses the dialog for a real Claude Code process. The verification record at lines 209-222 proves only the negative - that `claude --help` scopes the bypass to non-interactive mode - and the spawn test (tests/fm-claude-trust.test.sh:307) runs against the fake claude from make_spawn_fakebin, so it asserts the store entry and the launch command string, not that a real harness reaches its brief. Two assumptions ride on this untested: that Claude keys its projects entry by the physically resolved cwd (the script registers WT_REAL from `pwd -P`, so a harness keying by the logical launch path would silently miss), and that the flag alone clears the gate. Either being wrong yields the exact wrong-result-without-error shape - the script prints `trusted:` and exits 0 while the worker still wedges. The intent&#39;s test clause reads &#34;a real claude spawn reaches its brief with no human&#34;, which may mean a real fm-spawn run (what the test does) or a real harness launch; the repo has a separate gated convention for the latter (tests/fm-cmux-claude-composer-live-e2e.test.sh). Raising rather than fixing because the remedy - a gated live-e2e or an added Verified-on-version observation of a pre-registered worktree launching dialog-free - extends the change, and the intent also says &#34;no new runner&#34;.

🔧 Fix: degrade without node, fix Pi gate claim, record trust proof

- 🚨 `bin/fm-claude-trust.sh:131` - The fix round&#39;s node-missing branch launches exactly the worker the change exists to prevent, contradicting a REQUIRED criterion. The intent states: &#34;bin/fm-spawn.sh pre-registers trust for the task worktree path in the launching user&#39;s own Claude trust store at spawn time, refusing the spawn rather than launching a worker that would wedge&#34;. Lines 131-133 now do the opposite: `if ! command -v node &gt;/dev/null 2&gt;&amp;1; then echo &#34;warning: ...&#34; &gt;&amp;2; exit 0; fi`. fm-spawn.sh:2623 tests only the exit status, so a zero exit lets the launch proceed; the crewmate then meets the interactive workspace-trust dialog, whose selection starts on `No, exit`, and firstmate&#39;s key plane cannot answer it - the exact wedge the change was written to remove, now reached silently on any box without node. Two documents also became false the moment this landed: .agents/skills/harness-adapters/references/harness/claude.md:21 still tells the firstmate &#34;bin/fm-spawn.sh refuses the spawn when the write fails rather than launching a worker that would wedge&#34;, and docs/verification/runtime-backends.md&#39;s new &#39;Claude workspace trust&#39; section never mentions a node prerequisite or the degrade at all, so a firstmate reading either one will conclude a visible dialog means the store write failed rather than that node was absent. The prior round&#39;s finding that raised this named three options - accept the dependency, gate it with a preflight refusal that names the missing interpreter, or write the store without node - and explicitly left the choice to the author; the fix round instead picked a fourth that the intent forbids. The remedy is a policy decision (refuse loudly and require node, or keep the degrade and amend the intent plus both docs), not a mechanical correction, so it needs your call rather than an auto-fix.
- ⚠️ `tests/fixtures.sh:286` - The new test sandboxing pins HOME but not CLAUDE_CONFIG_DIR, so it does not achieve what its own comment claims (&#34;without it the suite would write the developer&#39;s real ~/.claude.json&#34;). bin/fm-claude-trust.sh:87 resolves the store as `${CLAUDE_CONFIG_DIR:-${HOME:-}}`, so CLAUDE_CONFIG_DIR wins over the throwaway HOME whenever the invoking shell exports it - which is routine on a machine that splits work/personal Claude subscriptions, the very setup fm-spawn.sh:3061 exists to serve. Concrete sequence: a developer with CLAUDE_CONFIG_DIR=~/.claude-work runs bin/fm-test-run.sh; every claude spawn in tests/fm-backend.test.sh (SPAWN_HOME at line 43), tests/fm-backend-orca.test.sh (line 13), tests/fm-control-relaunch.test.sh, tests/fm-trace-context-spawn.test.sh and every caller of fm_test_run_spawn writes `projects[&lt;temp worktree&gt;].hasTrustDialogAccepted=true` into that real store. The tests still pass, so the pollution is silent, and it accumulates dead entries pointing at deleted temp paths. The repo already has the correct idiom two files away: tests/fm-spawn-dispatch-profile.test.sh:94 pins `CLAUDE_CONFIG_DIR=&#34;${FM_TEST_CLAUDE_CONFIG_DIR:-}&#34;` for exactly this reason, and tests/fm-backlog-atomicity.test.sh already sets `CLAUDE_CONFIG_DIR=&#39;&#39;`. Adding the same empty pin alongside the HOME assignment in the four suites that lack it closes the leak without touching any launch-shape assertion, since an empty value produces the same no-prefix launch command those assertions already read.

🔧 Fix: refuse without node, pin CLAUDE_CONFIG_DIR in spawn tests
4 issues (1 error, 2 warnings, 1 info) still open:

- 🚨 `CLAUDE.md:4` - The last fix round (a20ad27) re-added two files that an earlier fix round (f24b8c8) had deliberately deleted as out-of-scope: the `CODE-INTEL-ROUTING` block appended to `CLAUDE.md` (lines 3-12) and the new `.codegraph/.gitignore`. Neither has anything to do with this change. Both are artifacts of the reviewing machine&#39;s local CodeGraph tooling — the block literally self-identifies as &#34;(managed by code-index.sh)&#34; and documents MCP servers (`codegraph_explore`, `codebase-memory-mcp`) and a `retrieval-routing-governance` rule that exist nowhere in this repo, and the block also violates the intent&#39;s stated repo style (&#34;one sentence per line in tracked markdown&#34;, plain dashes). The intent scopes this change to &#34;a spawn-time registration plus a documentation correction... and a small diff&#34;; these two files are outside it, and this branch is published to the third-party upstream `kunchenguid/firstmate`, so the local tooling config would land in someone else&#39;s repo. The history shows the exact regression: 01b3436 added them, f24b8c8 removed them, a20ad27 added them back — most likely because the CodeGraph daemon rewrites `CLAUDE.md` on save and a broad `git add` swept it up again. Remedy: restore both paths to their base state (`git checkout 7dcf072 -- CLAUDE.md` and `git rm -r --cached .codegraph`), and stage files explicitly on any further fix round so the daemon&#39;s rewrite cannot re-enter the commit a third time.
- ⚠️ `bin/fm-claude-trust.sh:87` - The script resolves the store as `${CLAUDE_CONFIG_DIR:-${HOME:-}}` and then hardens it with `real_dir`, which resolves a relative value against fm-spawn&#39;s own cwd. `bin/fm-spawn.sh:3062` forwards the SAME variable onto the launch verbatim (`CLAUDE_CONFIG_DIR=$(shell_quote &#34;$CLAUDE_CONFIG_DIR&#34;)`), and the crewmate pane is created with cwd set to the task worktree. Concrete sequence: firstmate runs with `CLAUDE_CONFIG_DIR=.claude-work` (relative) from `~`; the trust script writes `hasTrustDialogAccepted` into `~/.claude-work/.claude.json`, prints `trusted: &lt;wt&gt;`, and exits 0, so fm-spawn:2623 lets the launch proceed; the pane then reads `&lt;worktree&gt;/.claude-work/.claude.json`, finds no entry, and the worker wedges on exactly the dialog this change exists to remove — with the spawn reporting success. This is a wrong result that does not error. The script&#39;s own header at line 42 already asserts the invariant it does not hold: &#34;fm-spawn.sh forwards its own resolved CLAUDE_CONFIG_DIR onto the launch&#34; — fm-spawn does no resolution. The earliest shared boundary is fm-spawn: resolve `CLAUDE_CONFIG_DIR` to an absolute path once, before both the `fm-claude-trust.sh` call at line 2623 and the launch prefix at line 3062, so the two sides provably name the same store. Every absolute value (the only realistic configuration, and the one `tests/fm-spawn-dispatch-profile.test.sh:734` asserts) produces a byte-identical launch command, so no existing launch-shape assertion changes. The same class applies to an unset `CLAUDE_CONFIG_DIR` when firstmate&#39;s HOME differs from the tmux daemon&#39;s — the comment at fm-spawn.sh:3054 already notes the pane does not inherit firstmate&#39;s environment — but that one has no in-repo fix and is far less reachable.
- ⚠️ `bin/fm-claude-trust.sh:200` - `attempt()` reads the whole store at time T, mutates the in-memory object, and renames a full re-serialisation over it. Any write Claude Code itself makes between the read and the rename is lost in full — not just a projects entry, but whatever else moved in that file (`oauthAccount`, user-scope `mcpServers`, onboarding flags, every other project&#39;s history). The readback at line 207-208 only confirms OUR key survived; it cannot detect that we dropped the vendor&#39;s update, so the failure is silent and the script still prints `trusted:`. This matters because fm-spawn is invoked from a live firstmate Claude Code session that is itself an active writer of that exact file, and this now runs on every claude ship/scout spawn. The comment at lines 152-166 reasons carefully about the opposite direction only (&#34;a vendor session that rewrites the whole store after this returns can still drop it&#34;) and the ponytail note scopes the deferred lock to &#34;concurrent spawns&#34;, so this direction appears unconsidered rather than accepted. The window is milliseconds and the vendor already races itself across its own sessions, so this is not a merge blocker — but the honest remedy is either a lock or a compare-and-swap on the pre-read bytes, which is new machinery beyond &#34;a spawn-time registration, not a trust subsystem&#34;. Flagging the remedy, not the defect, as the thing needing your authorization: either accept and record this direction in the comment alongside the one already documented, or authorize the guard.
- ℹ️ `tests/fm-claude-trust.test.sh:321` - The comment above `test_scope_refusal_stays_fail_closed_without_node` still reads &#34;Degrading on a missing interpreter must not soften the scope boundary&#34;, left over from the reverted degrade-on-missing-node round. The code it documents now refuses rather than degrades (`bin/fm-claude-trust.sh:132`), and the sibling comment at lines 303-305 states the correct behaviour, so a reader hitting line 321 first is told the opposite of what the suite proves. Reword to &#34;A missing interpreter must not soften the scope boundary&#34; or similar.

🔧 Fix: refuse relative config dir and concurrent store modification

- ⚠️ Intent contradiction (metadata, not behavior). The intent&#39;s Repo style clause states: &#34;one sentence per line in tracked markdown, plain dashes, shellcheck-clean bin scripts, and no agent name as a commit co-author.&#34; Both of the author&#39;s own commits in this range carry an agent co-author trailer: 33f41af and 7271685 each end with `Co-Authored-By: Claude Opus 5 &lt;noreply@anthropic.com&gt;`. (The seven pipeline fix-round commits are clean — they carry no trailers at all — so this is confined to the two original commits.) I am not resolving it, because the only remedy conflicts with two other binding constraints from the same intent: &#34;Preserve every existing commit; never abort-and-restart or reset the branch&#34; and &#34;Never force-push&#34;. Amending a message rewrites both commits. This needs your decision: either authorize the rewrite before push (the branch is not published yet, so no force-push would be needed), or accept the trailers and waive the style clause for this branch. Note the repo&#39;s own history is not uniform here — 5 commits before the base carry `Co-Authored-By: Cursor &lt;cursoragent@cursor.com&gt;` — so precedent exists for accepting it.
- ℹ️ `bin/fm-claude-trust.sh:36` - The header asserts as fact: &#34;One structural test covers both worktree providers because Orca&#39;s task worktree is a linked git worktree too.&#34; I could not find in-repo evidence for that. docs/orca-backend.md:48 says spawn &#34;registers the repository, creates an independent worktree&#34; and never states the worktree shares the project&#39;s git common dir; docs/verification/runtime-backends.md:770-795 verifies Orca&#39;s readiness, terminal handle, and worktree id/path against the real binary, but records nothing about the worktree&#39;s git relationship to the project. The existing pre-check the Orca path already runs (validate_spawn_worktree, bin/fm-spawn.sh:1901) only requires a git toplevel that is not the primary — it never compares common dirs, so it would pass an Orca-managed clone. The fake-Orca suite cannot detect a divergence either: tests/fm-backend-orca.test.sh:517 builds its worktree with fm_git_worktree, i.e. a real `git worktree add`, so the fixture satisfies the invariant by construction. If Orca clones rather than links, every `--backend orca` claude spawn now hard-fails at bin/fm-spawn.sh:2621 with &#34;is not a worktree of project&#34;. That failure is loud and correct per the intent&#39;s stated policy, and orca is experimental and macOS-only, so this is not a merge blocker — but the header states an unproven assumption as established fact, and a reader will rely on it. Either soften it to a stated assumption or add the one-line Orca evidence to the verification doc.
- ℹ️ `bin/fm-claude-trust.sh:229` - In the fix round&#39;s new concurrency guard, the staged file is created at line 228 and only removed on two paths: the `moved` branch (line 230) and a failed rename (line 236). If `readStore()` inside `fingerprint(readStore())` at line 229 throws anything other than ENOENT — EISDIR if the store was replaced by a directory, EACCES if its mode changed under us, EIO — the throw propagates to the outer catch at line 251 and the script exits 1 with the staged `.claude.json.fm-trust.&lt;pid&gt;.&lt;hex&gt;` left in the config directory, holding a full copy of the user&#39;s store. Concrete sequence: store exists and is read at line 199, tmp written at 228, the store&#39;s mode is changed to 000 by anything else on the box, line 229 throws EACCES, exit 1, temp file orphaned. Every other exit from attempt() cleans up, so this is an omission rather than a tradeoff; the fix is mechanical — wrap lines 229-238 so the tmp is removed in a finally. Low reachability, and the contents are mode 0600, so this is litter rather than exposure.

🔧 Fix: correct orca worktree claim, clean staged store on failure

- ⚠️ `bin/fm-claude-trust.sh:238` - A prior fix round regressed the store serialisation and the reason is not recorded anywhere. The author&#39;s original code was `JSON.stringify(root, null, 2)`; commit 9ac8b44 (&#34;refuse unresolvable git dir, compact store, fix secondmate doc&#34;) changed it to `JSON.stringify(root)` with no code comment and no commit body explaining why. Claude Code writes this file pretty-printed - verified on this box: `head -c 3 ~/.claude.json` is `{\n  ` and the file is 345 KB - so every claude spawn now rewrites the operator&#39;s entire 345 KB config as a single line, and the vendor&#39;s next write expands it again. That churn is not hypothetical for the layout this script goes out of its way to support: lines 155-164 and tests/fm-claude-trust.test.sh:303 exist specifically for a dotfile-manager or synced-folder store, where a version-controlled `.claude.json` gets a whole-file diff on every crewmate spawn. Restore `JSON.stringify(root, null, 2)` so the write matches the format the vendor itself uses; that is a revert to the author&#39;s own line, not a new behavior.
- ⚠️ `docs/verification/runtime-backends.md:268` - The new trust-proof record claims &#34;All test entries were removed from the store afterwards and the scratch repo deleted.&#34; The second half holds - /tmp/trustproof no longer exists - but the first half does not: `grep -o &#39;&#34;[^&#34;]*trustproof[^&#34;]*&#34;&#39; ~/.claude.json` still returns `&#34;/tmp/trustproof/proj&#34;`, a live project key in the operator&#39;s real store pointing at a path that no longer exists. The wt-a and wt-c arms were cleaned; the proj entry was missed. This is a verification record whose whole value is that its claims are checkable, and it is the one claim in the section that a reader can check and find wrong. The remedy is bounded to the sentence, because deleting the residual entry means writing outside this worktree: state honestly that the worktree arms were removed and one project entry for the deleted scratch repo remains, or drop the cleanup sentence.
- ⚠️ `.agents/skills/harness-adapters/references/harness/claude.md:28` - The change removes the Enter recipe for the trust dialog and states plainly why Enter is wrong there (&#34;the observed rendering starts on `No, exit`&#34;), then reintroduces the same hazard one dialog over. Line 27 says pre-registration does not address the machine-scoped bypass-permissions confirmation, and line 28 says &#34;Inspect the pane before answering that one, and confirm which dialog is on screen rather than sending Enter to whatever appeared.&#34; A firstmate that follows this confirms the bypass dialog is on screen and then answers it with the only key it has, Enter - and this same change&#39;s own evidence at docs/verification/runtime-backends.md:267 records that the Bypass Permissions warning &#34;also defaults to `No, exit`&#34;, so that Enter exits the worker exactly as the trust-dialog Enter did. The reference states the No-exit default for one dialog and withholds it for the other. This needs your call rather than my edit, because the honest remedy is a policy statement about what a firstmate may do - most likely that it cannot accept this one either and an operator must accept it once per machine - which is guidance, not a mechanical correction.
- ℹ️ `.agents/skills/harness-adapters/references/common/control-and-recovery.md:26` - The secondmate exclusion is named as the intent requires, but only as a mechanism: &#34;fm-spawn.sh runs its per-harness pre-launch setup only for non-secondmate kinds, so the registration is never invoked for one,&#34; and &#34;that kind guard is the whole exclusion.&#34; What a reader is not told is the consequence. bin/fm-spawn.sh:1259 is the single claude launch template used for every kind, so a claude secondmate launches into an interactive pane in a not-yet-trusted home the same way a crewmate does, and by this change&#39;s own argument firstmate cannot answer that dialog. A reader deciding whether a claude secondmate can be spawned unattended would conclude from this paragraph that the case is handled. This is not a regression - the intent excludes secondmates by design and the behavior is unchanged - so I am not proposing to extend registration to them; the ask is only whether to add the residual-gate sentence, which is your wording decision.

🔧 Fix: restore pretty-printed store, correct trust dialog docs
1 warning still open:

- ⚠️ `bin/fm-spawn.sh:2625` - The new trust refusal exits after the busy-state generation is armed and before its rollback is armed, so a fresh claude spawn that cannot pre-register trust leaves orphaned busy records behind. Concrete sequence: `fm-spawn.sh &lt;id&gt; &lt;proj&gt; claude --mode no-mistakes --yolo off` on a box where `~/.claude.json` is root-owned (or `node` is missing, or the backend is orca and its worktree is a clone). Line 2597 runs `fm-busy-event.sh arm`, which writes `state/&lt;id&gt;.busy-gen` and `state/&lt;id&gt;.busy-state`. Line 2623 then refuses and line 2625 exits 1. `spawn_abort_cleanup` retires the generation only when `RELAUNCH_REPLACEMENT_PENDING` is set (relaunch only, line 801) or when `SPAWN_FRESH_COMMIT_PENDING` is set, and that flag is not set until line 2919 - far past this exit. So both files survive for a task id that has no `state/&lt;id&gt;.meta`. This is a new hole, not a pre-existing one: the other exits in the same block (2571, 2591, 2599, 2610) all run before the arm or before it succeeds, and this is the only `exit 1` between line 2597 and line 2919. The invariant is one the repo already enforces - `fm_backlog_dispatch_rollback` (bin/fm-backlog-transition-lib.sh:556) fails loudly if `.busy-state` or `.busy-gen` survive a failed dispatch. Minimal remedy, no new machinery: move the `fm-claude-trust.sh` call above the `BUSY_GEN=` arm at line 2596, where it sits alongside the codex and kimi refusals that already exit before arming; it needs only `$WT` and `$PROJ_ABS`, both settled by line 2536. Retiring `$BUSY_GEN` immediately before the `exit 1` is an equivalent one-line alternative.

🔧 Fix: arm trust gate before busy state to avoid orphans
1 info still open:

- ℹ️ `tests/fixtures.sh:273` - The header comment on `fm_test_run_spawn` still says &#34;Extra variables in the caller (GROK_HOME, FM_FAKE_LAUNCH_LOG, CLAUDE_CONFIG_DIR, ...) are inherited&#34;, but line 291 now hard-assigns `CLAUDE_CONFIG_DIR=&#34;${FM_TEST_CLAUDE_CONFIG_DIR:-}&#34;` and line 290 hard-assigns `HOME=&#34;$spawn_home&#34;`, so neither is inherited any more. A future test author who follows the header and writes `CLAUDE_CONFIG_DIR=/x fm_test_run_spawn ...` gets the empty value silently: the spawn succeeds, no error is raised, and the launch-shape assertion they wrote to check the forwarding prefix just fails with a confusing message. The correct opt-in is `FM_TEST_CLAUDE_CONFIG_DIR`, which the in-body comment documents but the header contradicts. Fix is to update the header to name `CLAUDE_CONFIG_DIR` and `HOME` as pinned (opt in via `FM_TEST_CLAUDE_CONFIG_DIR`) rather than inherited. Worth noting alongside: `tests/fm-spawn-dispatch-profile.test.sh:94` now pins `CLAUDE_CONFIG_DIR=&#34;${FM_TEST_CLAUDE_CONFIG_DIR:-}&#34;` immediately before calling `fm_test_run_spawn`, which does the same pin, so that line is redundant and can go.

- ℹ️ `tests/fixtures.sh:273` - The `fm_test_run_spawn` header still says &#34;Extra variables in the caller (GROK_HOME, FM_FAKE_LAUNCH_LOG, CLAUDE_CONFIG_DIR, ...) are inherited&#34;, but the body now hard-assigns both `HOME=&#34;$spawn_home&#34;` (line 290) and `CLAUDE_CONFIG_DIR=&#34;${FM_TEST_CLAUDE_CONFIG_DIR:-}&#34;` (line 291), so neither is inherited any more. Concrete failure: a test author follows the header and writes `CLAUDE_CONFIG_DIR=/x fm_test_run_spawn ...`; the value is silently replaced by the empty default, the spawn still succeeds, and the launch-shape assertion they wrote to prove the forwarding prefix fails with a message that points at fm-spawn rather than at the fixture that dropped their value. The documented opt-in is `FM_TEST_CLAUDE_CONFIG_DIR`, which only the in-body comment mentions. Remedy is to correct the header to name `HOME` and `CLAUDE_CONFIG_DIR` as pinned (opt in via `FM_TEST_CLAUDE_CONFIG_DIR`). Worth folding in: `tests/fm-spawn-dispatch-profile.test.sh:94` now sets `CLAUDE_CONFIG_DIR=&#34;${FM_TEST_CLAUDE_CONFIG_DIR:-}&#34;` immediately before calling `fm_test_run_spawn`, which applies the identical pin and is therefore redundant.
- ℹ️ `bin/fm-spawn.sh:2598` - Informational, no change requested. The trust gate sits after `treehouse get` has already leased a worktree in the pane (line 2508) and after the task window exists, and `spawn_abort_cleanup` (line 792) releases neither: it handles orca terminals/worktrees and the herdr projection, but there is no treehouse release anywhere in this script. A refusal therefore exits with a leased pooled worktree and an orphan window, and no `state/&lt;id&gt;.meta` for teardown to find them by. This is not a hole this change opens - the pre-existing exits at lines 2550, 2554, 2571 and 2610 leak identically - but the trust gate makes that class materially more reachable, because it is the first gate that fails for environmental reasons (a root-owned or unwritable `~/.claude.json`, a missing node, an orca worktree that turns out to be a clone) rather than for a bad argument. Recording it so the exposure is a known tradeoff rather than a surprise; the remedy would be new lease-release machinery in the abort path, which is outside this change&#39;s stated scope.
</details>

<details>

</details>

<details>

- ℹ️ `.agents/skills/harness-adapters/references/common/control-and-recovery.md:19` - Judgment call worth surfacing rather than a defect I should fix further. The new folder-trust overview paragraph exists because the intent requires one place naming which harnesses gate, suppress, dodge, and pre-register, but that requirement is in tension with the one-owner placement policy: each line necessarily restates a fact owned by its harness reference. I trimmed the Pi entry, which had grown into two sentences carrying fully owned detail including the ~/.pi/agent/trust.json path and the state/-wiring rationale, and left the Grok, Cursor, Muse, and Codex entries as single category clauses because that is the minimum needed for the overview to be intelligible. The residual risk is that a later change to grok.md line 34 or cursor.md line 19 leaves this paragraph&#39;s clause stale, since nothing enforces the pairing. If that becomes real, the follow-up is to strip every rationale clause here down to a bare category plus owner pointer, which is a consolidation pass over the whole paragraph rather than part of this change.
- ℹ️ `docs/scripts.md:8` - bin/fm-claude-trust.sh is not listed in the bin/ toolbelt table, and I deliberately did not add it. The table is already selective rather than an enforced inventory: 34 tracked bin/ entries are absent from it, including non-lib entrypoints such as fm-lint.sh, fm-doc-audience-check.sh, and fm-stow-cascade.sh, and no test or check verifies coverage. fm-claude-trust.sh is an internal pre-launch helper invoked only by fm-spawn.sh, so omitting it matches how comparable helpers are already treated, and adding one row would imply a completeness contract the table does not hold. Flagging it so the omission is a recorded decision rather than an oversight; if you want the table to become exhaustive, that is a separate pass with a drift check to keep it honest.

- ℹ️ `docs/fm-test-portable-shards.md:73` - This change adds tests/fm-claude-trust.test.sh to the portable serial lane, and docs/fm-test-portable-shards.md instructs &#34;Refresh the hints whenever the serial lane gains scripts, rather than waiting for that bound to trip.&#34; I did not perform that refresh, and I judged hand-editing it worse than leaving it. The doc&#39;s numbers are dated measurement evidence, not prose: the recorded &#34;139 scripts&#34;, the 3809887 ms weight total, and the per-lane script counts and estimated durations all come from the slowest of three named green CI runs on 2026-09-01, and the authoritative hint values live in bin/fm-test-run.sh:portable_serial_weight_hints, not in the doc. Producing an honest update needs fm-test-timing-portable-serial-* artifacts from a green run of this branch, which I have no way to generate from this phase, and inventing a weight for the new script would put an unmeasured number where the doc promises a measured one. The consequence of leaving it is bounded and the doc says so itself: the new script falls back to PORTABLE_SERIAL_DEFAULT_WEIGHT_MS, hints affect balance only, and the coverage guard keeps the partition complete and disjoint regardless, so this costs shard balance rather than coverage. The follow-up is to refresh the hints in bin/fm-test-run.sh and the evidence tables in this doc together from the next green run&#39;s timing artifacts.
</details>

<details>

✅ No issues found.

✅ No issues found.
</details>

<details>

✅ No issues found.

✅ No issues found.
</details>


kunchenguid:

First look on HEAD `4c3239d4d54b736975c70d8d74a0207ee2634ec8` vs main `31cba0db93f08f5d7ce87442d716fcb97fc00d5d`. MERGEABLE. RooseveltAdvisors is not a blocked author. No `.github/**` edits. Workflow approvals this pass: no (CI/NM already running on tip).

**Attestation: MATCH** (body binds tip). CI `33827955478` SUCCESS across lint/coverage/portable/Herdr/timing/macOS/invariants; Require no-mistakes `33827994206` SUCCESS. Greptile FAIL with a real P1.

**Contract-class: restore.** Claude crewmates were wedging on the interactive workspace-trust dialog (Enter selects `No, exit`; `--dangerously-skip-permissions` does not skip it in an interactive pane). Pre-registering `hasTrustDialogAccepted` for the isolated task worktree restores the already-promised path that a spawned worker reaches its brief. Scope is structural (linked worktree of the spawning project only); refusal is fail-closed. Secondmate exclusion is documented.

**VISION.md per-rule**
- **One captain, one interface** — aligns.
- **Authority is explicit and never inferred** — aligns as restore of reachability; trust write is scoped to the task worktree the captain already authorized this spawn into.
- **Scripts own the mechanics, agents own the judgment** — aligns (`fm-claude-trust.sh` deterministic).
- **A restart is a non-event** — aligns.
- **Delegation with a spine** — aligns (unblocks supervised workers).
- **The fleet outlives any vendor** — aligns (Claude-specific adapter; other harnesses unchanged).
- **Scope** — aligns.

**Author blocker (verified):** Greptile P1 on `bin/fm-spawn.sh` — trust registration runs after backend endpoint creation (tmux/Herdr/Zellij/cmux create_task) and after `TASK_TMP=/tmp/fm-$ID`, but `spawn_abort_cleanup` only specially recovers Orca / Herdr-projection aborts. A trust refusal therefore leaves an orphaned non-Orca endpoint + `/tmp/fm-<id>` without published meta for normal teardown. Fix the abort path (or move registration before endpoint creation) before merge.

Do not merge while that orphan path is open. Not a Firstmate flag (still author-blocked despite green CI).

kunchenguid:
Speaking as Kun's firstmate: this is merged. Thank you @RooseveltAdvisors — really appreciate you taking the time on this.

Re-triage note on tip `651c6a5a3b3ec8bd71dd7864558bbc6ef8ac8181` (after the prior waiting-author stamp on `4c3239d4…`): attestation MATCH; CI `33829698539` + NM `33829698526` green; contract-class restore. Greptile still flags the endpoint left on trust refusal — accepted as the same pre-existing refusal class as the treehouse-get / validate_spawn_worktree exits above it (`inspect window $T`; no new temp-root orphan after the ordering move).

PR #3676, opened 2026-09-04
fix(bin): detect away mode flagged with no live supervisor daemon

krakns:
## Intent

Find out why firstmate's away-mode supervisor daemon stops, and fix it. Away mode is the captain's most-used supervision mode and it failed three times in one evening on 2026-09-02, each time leaving the fleet in the worst possible shape: state/.afk present with no daemon and no watcher, so away mode claimed supervision it was not providing.

This is firstmate's shared tracked material, so the firstmate-coding-guidelines contract applies: one sentence per line in Markdown, plain dash never an em dash, no agent co-author on commits, shellcheck-clean bin scripts, colocated tests.

TWO PIECES OF WORK, IN ORDER.

ITEM 1 - determine why the daemon stops. This is the one that matters. The captain required reproduction first, launched the way it actually runs (through the harness's own tracked background job, FM_AFK_STATE_PREPARED=1 bin/fm-afk-start.sh, not a bare shell launch), to establish whether the daemon is reaped by the host, exiting on its own, or dying to a signal. Three given data points: all three stops happened after tens of minutes with work in flight; only the third produced any captured error; the host harness marked the job [killed] each time, which is the harness's own marker rather than daemon output and so does not by itself identify the killer. The captain stated explicitly that if the host can reap that job then the away-mode contract itself is wrong and a fix inside the daemon will not hold, and that in that case I should say so plainly and propose the structural answer rather than patching the symptom.

That is what the investigation found, and it is the decision behind this diff. The daemon does not crash and does not exit on its own. The host harness (Claude Code) reaps the tracked background job with SIGTERM; the harness's captured job output ends with "[killed] / Terminated: 15", and the daemon log shows "daemon shutting down" at each reap time, which is its own SIGTERM cleanup trap running: flush buffered escalations, kill the watcher child, release the lock, remove the pidfile, exit 0. cleanup deliberately never clears state/.afk, because clearing the flag last is bin/fm-afk-launch.sh stop's job. A bare harness SIGTERM is not that stop path, so the flag is left set with nothing running - silently, which is why the first two stops produced no error at all. The sleep hypothesis firstmate raised was checked against pmset -g log and disconfirmed: the one daytime system sleep on 2026-09-02 (17:23-17:56 clamshell) has no reap near it, and the daemon alive across that sleep survived it and kept logging real work mid-sleep. What is deliberately NOT claimed as established is the exact internal Claude Code condition that triggers the reap (lifetime cap, idle reaping, resource pressure, or a compaction boundary); that lives in harness internals not in this repo. Per the captain's instruction, that is reported as not established rather than guessed, and it does not change the conclusion: the host can and does reap the job, the daemon cannot self-protect, so the answer is structural detection, not a patch inside the daemon.

Hence the required outcome the captain named regardless of cause: state/.afk present with no live daemon must be detected and reported loudly, because right now nothing notices until a human looks, and that is the difference between away mode being off (safe) and away mode lying (not safe). Implemented as three coordinated changes so every supervision-health surface stops trusting a lingering beacon:
- bin/fm-wake-lib.sh: new fm_afk_flag_without_live_daemon, and fm_watcher_supervision_verdict is now away-mode-aware. A fresh beacon alone is deliberately no longer accepted as proof of health in away mode, because the daemon's watcher child leaves a fresh .last-watcher-beat for up to the grace window after the daemon is reaped; the verdict now also requires a live identity-matched daemon via fm_afk_daemon_owns_supervision (the same test the turn-end guard already used) and returns a new no-afk-daemon reason otherwise. This closes the fresh-beacon blind window where a just-reaped daemon still read as healthy.
- bin/fm-guard.sh: the watcher-down banner now names the away-mode case explicitly, with the header AWAY MODE FLAGGED, BUT NO SUPERVISOR IS RUNNING and a cause line. It fires from the pull/peek/merge guards and from session start's read-only guard.
- bin/fm-session-start.sh: the away-mode digest subsection no longer unconditionally claims away-mode supervision is active; when the flag is set with no live daemon it says plainly that no supervisor is running and to relaunch or exit away mode. This catches the idle-fleet case the guard's needed gate skips.
The turn-end guard was deliberately left unchanged because it already blocked a blind turn end in this exact case; it is the model the other surfaces now follow.

ITEM 2 - fix the empty-mtime read at the housekeeping gate in bin/fm-supervise-daemon.sh. The captured error from the third stop was "bin/fm-supervise-daemon.sh: line 1720: [: : integer expression expected". That line gates the housekeeping tick on _file_age "$STATE/.subsuper-last-housekeep"; _file_age returned empty, which its own contract says cannot happen since it returns 999999 when the file is missing. This is a real defect on its own account: on that tick the housekeeping gate silently fails, so batch flushes, stale rechecks and the catch-all scan do not fire. The captain's instruction was to make the age read fail closed when its result is empty as well as when it errors, and to check whether _stat_file_mtime can return success with no output.

Reproduction refined the captain's stated premise, and that refinement is deliberate, not an oversight: under the daemon's set -u a truly EMPTY mtime does not produce the observed error, because $(( now - m )) with m empty treats it as 0 and prints a large number, so the gate would simply run. The observed empty _file_age output requires m to be non-empty but non-numeric (for example "abc" or "12x"): then the arithmetic itself aborts inside the command substitution, _file_age prints nothing, and [ "" -ge 15 ] raises exactly the reported error. The abort is confined to the $(...) subshell, which is why it was non-fatal. The fix covers all of those cases the same way: _file_age now validates that both the mtime and the clock are plain digit runs and returns 999999 otherwise, failing closed to "very old" so housekeeping RUNS rather than silently skipping - deliberately chosen as the safe direction, since running housekeeping unnecessarily is much cheaper than silently skipping supervision work.

The captain warned explicitly not to assume item 2 causes item 1, and it does not: the [ failure is inside an if condition and non-fatal, and the first two stops produced no error at all. The third stop's adjacency between the two is coincidence, and treating the visible error as the cause was named as the trap to avoid. Both items are fixed on their own merits.

ACCEPTANCE CRITERIA the captain set: item 1's cause established with a reproduction or explicitly reported as not established with what was ruled out (a well-evidenced could-not-reproduce is acceptable, a guess dressed as a finding is not); away mode flagged with no live daemon detected and surfaced loudly, with a regression test; item 2 fixed with a colocated test covering the empty-result case; existing suites still pass. Tests were added in tests/fm-daemon.test.sh, tests/fm-guard-stale-banner.test.sh and tests/fm-session-start.test.sh accordingly.

CONSTRAINTS the captain set, which the diff must not violate: do not weaken the daemon's existing single-instance lock, crash-loop backoff, pane-gone guard or shutdown flush; and do not change what away mode is allowed to approve - it batches notifications, it never expands authority. Neither was touched.

=== ACCEPTED DECISIONS FROM NINE COMPLETED REVIEW ROUNDS - DO NOT RE-LITIGATE ===

A previous run of this branch completed nine review-and-fix rounds before its review agent was killed by the host (exit 143, SIGTERM) and the run timed out. Every fix from those rounds is present in this branch's history. The following were reviewed, decided by the supervising firstmate, implemented and verified. They are deliberate accepted intent, not oversights, and they should not be reopened:

1. Ownership before freshness. fm_watcher_supervision_verdict tests fm_afk_daemon_owns_supervision BEFORE beacon freshness, so a reaped away-mode daemon reads no-afk-daemon at ANY beacon age. The original fix had this ordering backwards and only caught the fresh-beacon window, which is the rare case; every real incident sat past the grace window. Non-away-mode paths are deliberately untouched.

2. Age helpers guarantee a postcondition, not a list of rejected inputs. _file_age, _epoch_age, _oldest_line_age and fm_path_age all return through a single result gate guaranteeing either a real non-negative base-10 age or the UNREADABLE_AGE sentinel. This was reached after three separate input classes defeated narrower checks in sequence: empty/non-numeric, leading-zero octal tokens, and out-of-order operands producing a negative age. The gate closes the property rather than the instances, and also catches overflow. Failing closed to the sentinel is deliberate and the safe direction: an unreadable age must make guarded work RUN and escalations FIRE rather than silently skip.

3. The sentinel never reaches captain-facing text as a number. One _age_text renderer turns it into a plain statement that the elapsed time could not be read. The fail-closed behaviour is unchanged - the gate still clears and the escalation still fires - and every measured-age message is byte-identical to the base. That _age_text keys off the sentinel value, so a genuine 999999-second age renders as unreadable, is a knowingly accepted tradeoff.

4. Provability split, decided deliberately after the detector was found overstating in BOTH directions. Where absence is genuinely proven - no lock, no recorded pid, a dead pid, or a recorded identity that demonstrably mismatches - the surfaces say no supervisor is running. Where a live daemon exists but its ownership cannot be established, principally the deliberate case at fm-supervise-daemon.sh where a daemon runs without recording its identity, they say supervision could not be confirmed. Both cases stay loud and both carry the repair instruction. A detector built because away mode asserted supervision it did not have must not itself assert absence it has not proven, and a recycled pid must land in proven absence rather than softening a real alarm.

5. Every away-mode surface reads the one owner. The guard banner header and cause line, the repair line, the session-start subsection, the NEXT STEP block and the turn-end guard banner all derive from fm_afk_daemon_owns_supervision / fm_afk_flag_without_live_daemon rather than the raw flag. fm-supervision-instructions.sh remains a pure renderer taking parameters, with safe defaults and no state access or library sourcing.

6. Read-only outranks the repair instruction. A session that did not acquire the fleet lock is still told loudly about the half-state - suppressing it would reintroduce the silent failure this change exists to remove - but is not told to relaunch anything it cannot act on, following the existing worktree-tangle precedent.

7. The reap evidence is recorded in docs/watcher-continuity.md because this ships to a public repository and the code comment would otherwise assert a causal claim no maintainer could check. It states what was observed, how it was traced, a one-line reproduction of the MECHANISM, that the sleep hypothesis was tested and disconfirmed against pmset, and explicitly that the internal harness condition triggering the reap is NOT established. That last omission is deliberate honesty, not an unfinished sentence.

8. Test standard, required and verified: every test added or changed in this branch was proven to fail against the defect it covers by actually reverting the relevant fix and running it, never by reasoning about it. This was imposed after two tests in this branch looked green while proving nothing - one where two cases shared a state directory through a deterministic name collision, and one that passed with the fix reverted because an upstream guard already short-circuited the path. Commits record what each pre-fix failure looked like.

9. Deliberately OUT OF SCOPE, tracked as separate work, and not to be raised as findings here:
- bin/fm-claude-stop-autoarm.sh and bin/fm-turnend-guard-cursor.sh use a bare .afk presence test as a proxy for ownership and therefore stand down recovery after a reap. This is real and is filed as fm-afk-flag-ownership-proxy. It changes WHEN supervision resumes, needs its own reproduction to prove recovery does not fight a genuinely live daemon, and does not belong in this branch. (bin/fm-turnend-guard.sh is NOT excluded and was fixed here.)
- No library owns computing an age safely, because the daemon sources fm-classify-lib.sh rather than fm-wake-lib.sh and fm-supervision-lib.sh is separate again. That structural gap is why the age family regenerated across six of nine rounds. It is filed as fm-age-helper-no-owner. Any further age-helper instance found now belongs to that work: note it, do not fix it here.

This branch delivers the DETECTION half and that half is complete and valuable on its own. Every failure direction in it is fail-closed toward doing the work and raising the alarm, so an undiscovered residual instance costs a wasted tick or an early warning, never silent supervision - which is the harm this task existed to remove.

## What Changed

- `bin/fm-wake-lib.sh` gains `fm_afk_flag_without_live_daemon` and `fm_afk_daemon_absence_is_proven`, and `fm_watcher_supervision_verdict` now short-circuits while `state/.afk` exists: it requires a live identity-matched daemon via `fm_afk_daemon_owns_supervision` before it looks at the beacon at all, returning a new `no-afk-daemon` reason at any beacon age, so a reaped daemon no longer reads as healthy behind its watcher child's still-fresh beacon.
- Every away-mode surface now derives its wording from those predicates instead of the raw flag, split by how strong the evidence is: `bin/fm-guard.sh` and `bin/fm-turnend-guard.sh` print `AWAY MODE FLAGGED, BUT NO SUPERVISOR IS RUNNING` for proven absence and `... SUPERVISION CANNOT BE CONFIRMED` for a live daemon with unverifiable ownership, `bin/fm-session-start.sh` replaces the unconditional "away-mode supervision is active" digest line and adds a matching NEXT STEP block, and `bin/fm-supervision-instructions.sh` takes new `--afk-daemon-down` / `--afk-absence-proven` parameters that default to the weaker claim.
- The daemon's age helpers now fail closed: `_file_age` and the new `_epoch_age` validate both operands as base-10 digit runs and return through a single `_age_result` gate yielding either a real non-negative age or the `UNREADABLE_AGE` sentinel, fixing the empty `_file_age` read that made the housekeeping gate at `bin/fm-supervise-daemon.sh` error with `[: : integer expression expected` and silently skip batch flushes, stale rechecks, and the catch-all scan; `_oldest_line_age` and the marker-age reads delegate to it, `_age_text` keeps the sentinel out of captain-facing text, and `bin/fm-supervision-lib.sh` drops its own mtime helper for `fm_path_age`, which carries the same contract. Regression coverage was added in `tests/fm-daemon.test.sh`, `tests/fm-guard-stale-banner.test.sh`, `tests/fm-session-start.test.sh`, `tests/fm-supervision-instructions.test.sh`, and `tests/fm-turnend-guard.test.sh`, and `docs/watcher-continuity.md` records the observed `SIGTERM` reap evidence and the disconfirmed system-sleep hypothesis.

## Risk Assessment

✅ Low: The two commits under review close their prompting defects without opening a reachable new one - I traced every branch of the rewritten absence proof and the deferred wake-lib load and found them correct and fail-closed at each boundary - and the only surviving issue is contract text that understates what the code does.

## Testing

- Outcome: ⚠️ 1 error across 2 runs (1h21m50s)

## Pipeline

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>

- ⚠️ `bin/fm-supervision-lib.sh:27` - The new `if ! command -v fm_path_age` source block justifies itself with a comment at line 21 that the code contradicts: &#34;Sourced only when absent, so the four scripts that already load bin/fm-wake-lib.sh are unaffected.&#34;

bin/fm-turnend-guard.sh sources bin/fm-supervision-lib.sh at line 102 and bin/fm-wake-lib.sh only at line 156, so at line 102 `fm_path_age` is NOT yet defined, the guard fires, and wake-lib is sourced early - then parsed a second time at 156. One of the four consumers is affected, and it is the one on the per-turn-end hook path.

Verified empirically (standalone `set -u` script that sets STATE, then sources only bin/fm-supervision-lib.sh): the state directory does not exist before the source and does exist after, because bin/fm-wake-lib.sh:16 runs `mkdir -p &#34;$STATE&#34;` at source time.

Concrete consequences in bin/fm-turnend-guard.sh, none of which existed before this commit:
- `fm_primary_scope_matches &#34;$FM_ROOT&#34; &#34;$STATE&#34;` at line 152 ends with `[ -d &#34;$state&#34; ] || return 1` (bin/fm-primary-scope-lib.sh:32). That arm can no longer fail, because the state dir was created 50 lines earlier. A plain checkout with AGENTS.md and bin/ but no state dir previously exited 0 at the scope check and now proceeds into the guard.
- The write also lands ahead of the payload guard (line 112), the missing-jq degrade (line 122), and the foreign-host stand-down (line 123-125), so a checkout this hook deliberately stands down on now gets a `state/` directory created on every turn end.
- wake-lib is parsed twice per invocation here, and likewise in bin/fm-claude-stop-autoarm.sh and bin/fm-turnend-guard-cursor.sh, which source supervision-lib first.

The delegation itself is right and is what the previous round asked for; only the placement is wrong. Earliest supported boundary: move the source inside `fm_supervision_status` (its only user of `fm_path_age`), which keeps the single owner, restores the previous load order in every consumer - fm_supervision_status is not called until bin/fm-turnend-guard.sh:171, well after wake-lib is in scope - and removes the double parse. Then delete the line-21 claim rather than rewording it.
- ℹ️ `bin/fm-supervision-lib.sh:33` - `fm_sup_stat_mtime` lost its only caller in this commit: `fm_supervision_status` now reads the beacon age through `fm_path_age`, and the helper is referenced nowhere else in bin/, tests/, or docs/ (`grep -rn fm_sup_stat_mtime` returns only its own definition).

It was added purely to give this library a portable mtime read of its own; that responsibility now belongs to `fm_path_mtime` in bin/fm-wake-lib.sh, which the library already pulls into scope. Leaving it behind re-creates in miniature the thing the delegation was meant to remove: a second portable-stat implementation with no consumer to keep it honest. Delete it along with its &#34;Portable mtime; Linux stat lacks -f, macOS stat lacks -c&#34; comment.
- ℹ️ `bin/fm-guard.sh:176` - `episode_key=$(fm_guard_stale_episode_key &#34;$watcher_down_reason&#34;)` keys the once-per-episode full banner on the reason string alone, and `fm_guard_stale_episode_key` (line 62) just echoes it. The change added a second dimension to the banner - proven absence versus unconfirmed ownership, selected by `fm_afk_daemon_absence_is_proven` at line 190 - that the key does not carry.

Concrete path: away mode is flagged and a live daemon that never recorded its identity is supervising. The first guard call prints the full banner with `AWAY MODE FLAGGED, BUT SUPERVISION CANNOT BE CONFIRMED` and claims a supervisor process is running. That daemon then dies. `watcher_down_reason` is still `no-afk-daemon`, the episode marker still matches, so `print_full_banner` stays 0 and every later call prints only `WARNING: watcher still down (same stale episode...)`. The corrected, now-provable claim - that nothing is supervising - is never surfaced for the rest of that episode.

The reader is still loudly alarmed either way and the repair instruction is on both branches, so this is not a silent failure; it is the first-printed claim outliving the evidence that selected it. Including the provability flag in the episode key would let the stronger claim re-announce itself once. Recording it rather than prescribing it, since re-printing a full bordered banner mid-episode is a deliberate noise tradeoff the author owns.

🔧 Fix: defer wake-lib load in supervision-lib; drop dead mtime helper
1 info still open:

- ℹ️ `bin/fm-wake-lib.sh:333` - The contract paragraph for `fm_afk_daemon_absence_is_proven` still states the pre-fix rule and directly contradicts the code the same round shipped.

bin/fm-wake-lib.sh:333-337 opens with &#34;True only when away mode is flagged and NO daemon process is running at all: the singleton lock is absent, records no pid, or records a pid that is not alive. False when a live daemon pid IS present but its ownership could not be verified.&#34; The implementation at :344-357 now also returns true (proven absence) for a live pid whose recorded identity demonstrably mismatches - the recycled-pid case that was the whole point of the round-9 fix. So the opening sentence is false, and the &#34;False when a live daemon pid IS present&#34; clause is wrong for that path. The paragraph added immediately below it in the same commit states the correct rule, so one comment block now carries both.

bin/fm-supervision-instructions.sh:29 repeats the stale list verbatim in user-visible usage text: &#34;--afk-absence-proven reports that the absence is PROVEN - no lock, no recorded pid, or no live process&#34;, omitting the fourth condition. docs/turnend-guard.md:53 has it right and names all four, so a maintainer reading the library or the CLI help gets a contract that the doc and a one-line test both falsify - on exactly the predicate this branch exists to make honest.

Separately and in the same class: bin/fm-supervision-lib.sh:49 still documents FM_SUP_BEACON_DESC as &#34;human-readable beacon age, for banners (\&#34;never\&#34; if absent)&#34; while :89 now also emits `unknown` for an unmeasurable beacon.

All three are text-only; no behavior is wrong and nothing computes a wrong value. Mechanical fix: rewrite the wake-lib opening sentence to state the four proven-absence conditions (no lock, no recorded pid, dead pid, verified identity mismatch) and the single unconfirmed condition (live pid whose identity is unrecorded or unreadable), delete the now-contradictory &#34;False when a live daemon pid IS present&#34; clause rather than rewording it, extend the --afk-absence-proven usage text the same way, and add `unknown` to the FM_SUP_BEACON_DESC line.
</details>

<details>

🔧 Fix: no fix needed: 3 test failures pre-existing at base
1 error still open:

</details>

<details>

- ℹ️ `docs/turnend-guard.md:45` - Judgment call, recorded rather than acted on. The change added away-mode half-state wording to three surfaces, but only two of them have a documentation owner: docs/turnend-guard.md documents the pull-guard and turn-end-guard headers and the provability split, while bin/fm-session-start.sh&#39;s AFK digest subsection, its NEXT STEP block, and the rendered `- Away mode: FLAGGED, BUT ...` lines from bin/fm-supervision-instructions.sh have no prose owner in docs/. That is not staleness introduced here - the session-start digest&#39;s wording was never documented (the prior `present - away-mode supervision is active` line had no owner either), and docs/sessionstart-nudge.md owns the nudge transport and runtime bound, not digest content. I deliberately did not create a session-start documentation surface to close it, per the placement policy&#39;s rule against new surfaces for perceived gaps; instead docs/watcher-continuity.md now names bin/fm-session-start.sh as one of the three consumers of the detection. If a future change wants the session-start digest contract written down, the follow-up is to extend an existing operator-current owner rather than to add a new page.
</details>

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>


kunchenguid:

First look on HEAD `a3625e5daff4b4522286a524379a7a5f1f861068` vs main `f4d7875824ecc5e274b4bb896f10c1e1f207b7e4`. MERGEABLE/UNSTABLE. krakns is not a blocked author. Workflows were already running (not first-time action_required this pass).

**Attestation: MATCH.** Body binds `a3625e5daff4b4522286a524379a7a5f1f861068`. Require no-mistakes SUCCESS.

**CI: FAILURE** run `33821497921` — Behavior portable serial 2 failed in `tests/fm-turnend-guard.test.sh`: `not ok - the beacon is not actually stamped in the future, so this case proves nothing`, with `integer expression expected` while reading a future beacon mtime (stat output looked like `stat` file-info dump rather than a bare epoch). Other shards/Lint/Herdr/macOS/Repo invariants/coverage succeeded. Greptile P2 notes captain-facing copy still names daemon-identity machinery in places; that is a presentation polish request, not the merge gate here.

**Contract-class: restore.** Away mode already promised live supervision when `.afk` is present. Main still reports `present - away-mode supervision is active` in the half-state where the flag is set and no owning daemon is proven. Correcting that false active claim to a fail-closed half-state detection (and matching guard / turn-end / session-start surfaces) restores the existing honesty contract rather than inventing a new default path. Non-afk paths stay on the prior watcher story.

**VISION.md per-rule**
- **One captain, one interface** — aligns on honesty (no false “active”); Greptile’s P2 is real presentation debt where some new lines still expose process-identity machinery instead of outcome-only language.
- **Authority is explicit and never inferred** — aligns. Detection does not grant autonomy; it refuses to claim supervision that is not there.
- **Scripts own the mechanics, agents own the judgment** — aligns. Ownership proof lives in `fm-wake-lib.sh` / supervision helpers.
- **A restart is a non-event** — aligns. Half-state is derived from durable `.afk` + lock/identity records.
- **Delegation with a spine** — aligns. Supervision contract stays explicit.
- **The fleet outlives any vendor** — aligns. Daemon ownership is harness-agnostic.
- **Scope** — aligns. Shared tracked bin/docs/tests only.

Security review: no workflow edits; fail-closed absence proof; not malicious.

Do not merge while CI is red. Do not escalate to Firstmate for a red CI / author fix. After CI is green this remains a restore candidate (still subject to the captain-facing wording polish if you choose to address Greptile’s P2).

kunchenguid:

Re-check after newer activity on HEAD `ec7f4bc3dc5873e723b814afb8dd18dd8b31adca` vs main `a5c64a0b6791b7643024436af9f737059d15cef2`. MERGEABLE/UNSTABLE. krakns is not a blocked author. No `.github/workflows/*` edits. Workflow approvals this pass: no (CI already running).

**Attestation: MISMATCH (stale).** Body still binds `a3625e5daff4b4522286a524379a7a5f1f861068` after later `no-mistakes: apply CI fixes` tips. Require no-mistakes run `33862009820` FAILURE. Rebind the attestation HTML to the current HEAD and re-push through no-mistakes.

**CI:** run `33862009807` still in progress (several portable serial shards / Lint / Herdr pending). Already green: portable parallel 1+2, macOS snapshot, Repo invariants, Test coverage guard, Greptile. Do not treat this as merge-ready while NM is red and CI unfinished.

**Contract-class: restore.** Away mode already promised live supervision while `state/.afk` is present. Main still surfaces an active/away claim in the half-state where the flag remains after a host reap and no owning daemon is proven. Fail-closed detection on guard / turn-end / session-start / wake-lib restores that honesty contract; non-afk watcher paths stay on the prior story. Age-helper fail-closed is corrective for an existing housekeeping gate defect.

**VISION.md per-rule**
- **One captain, one interface** — aligns (loud half-state instead of a false “active”).
- **Authority is explicit and never inferred** — aligns (detection does not widen away-mode approval authority).
- **Scripts own the mechanics, agents own the judgment** — aligns (`fm-wake-lib.sh` / supervision helpers own the proof).
- **A restart is a non-event** — aligns (derived from durable `.afk` + lock/identity).
- **Delegation with a spine** — aligns (supervision contract stays explicit).
- **The fleet outlives any vendor** — aligns (ownership check is harness-agnostic; reap evidence is documented without pretending the Claude Code trigger is fully known).
- **Scope** — aligns (shared bin/docs/tests only).

Security review: no workflow edits; fail-closed absence proof; not malicious.

Blocked on author (attestation rebind) and CI finishing green. Not escalating to Firstmate while those blockers remain. After MATCH + green CI/NM this stays a restore auto-merge candidate.

PR #3679, opened 2026-09-04
fix(bin): declare a wait after finishing, and honor the declaration

synchronizeaicontact7919:
## Motivation

Twice in one afternoon on 2026-09-03, a lane finished its work correctly, appended `done: PR <url>`, and stopped - with its validation run still parked at its checks gate because GitHub Actions was halted account-wide. Both lanes had done exactly the right thing, including not looping the CI fixer. The watcher flagged each idle pane as a possible wedge anyway, and each cost firstmate a supervision turn to diagnose a completely healthy state. A third lane hit the same shape mid-run.

The generated crewmate status-protocol block defines `done:` as the terminal definition-of-done report, and separately explains `paused: {why}` for a deliberate external wait firstmate should recheck on a long cadence. Both descriptions are individually correct, but neither says what to do when both are true at once - the work is finished AND something external is still outstanding - so a worker reasonably emitted only the terminal `done:` line and stopped, leaving its idle pane indistinguishable from a wedge.

## Behavior change

Inside the existing status-protocol block generated by `bin/fm-brief.sh` (no new section), the block now states that `done:` is a terminal report and not a declaration of a wait, and that a worker finishing into an external wait (merge authority, a halted build service, an upstream branch landing) appends a `paused: <why>` line after its `done:` line and then stops. `paused:` and `blocked:` keep their existing meanings, and the mid-task rule that `working:` is nonterminal is unchanged. The sentence is verified present in every generated variant that carries the status protocol (ship, scout, and the charter/secondmate brief) via `bin/fm-brief.test.sh`.

## Also in this PR (follow-on fixes accepted during validation)

- A shared `status_terminal_then_paused` predicate in `bin/fm-classify-lib.sh`, used by both `bin/fm-watch.sh` and `bin/fm-crew-state.sh`, so a declared `paused:` after a `done:` line is read consistently and does not silently flip the CI-ready verdict back to false.
- `FM_TERMINAL_PAUSE_SCAN_MAX` as a bounded, named-default lookback for that scan, documented in `docs/configuration.md` alongside its sibling watcher knobs.
- `docs/architecture.md` updated to describe the declared-wait fast path and its inbox-freshness exit.
- `bin/fm-crew-state.sh` reads the status log once instead of twice, so the last-line read and the terminal-pause scan can never see different snapshots.
- `.omc/` added to `.gitignore` so local agent-tooling session state is never swept into a firstmate commit.

Deliberately left out: an array-based efficiency rewrite of the pause-scan loop was considered and dropped as separable from correctness - the existing per-iteration scan is retained, only shared.

## Addressed Greptile review finding

Greptile flagged that `no_inbox_activity_since_status` in `bin/fm-watch.sh` compared coarse integer-second mtimes with a strict `>`, so a steer landing in the *same* second as the terminal-then-paused status append was read as "no activity" and the stale `paused:` classification survived. Fixed by using `>=` instead, with a regression test (`test_same_second_inbox_activity_restores_wedge_detection`) pinned alongside the existing strictly-newer case in `tests/fm-watch-triage.test.sh`.

## Known pre-existing test flake (stepped past, not fixed here)

`tests/fm-composer-lib.test.sh` - "a half-block rule row must count as a structural edge" - fails reproducibly (3/3) on an untouched worktree at this branch's head, and this branch never touches `fm-composer-lib.sh` or its test (confirmed via `git log`). It is unrelated to this change and was approved past during the no-mistakes test gate rather than fed to the auto-fixer out of scope. Filed separately for its own owner.


synchronizeaicontact7919:
The same-second tie is removed by comparing at sub-second (nanosecond) mtime resolution instead of whole seconds (both platforms' stat report it - see the updated no_inbox_activity_since_status in 9431b0e). `>=` remains as the documented fallback only for the residual case where a filesystem genuinely cannot report sub-second precision, biased toward treating an unresolvable tie as activity (a silent miss that hides real work forever is worse than one extra supervision check).

kunchenguid:
Speaking as Kun's firstmate:

**Triage** (tip `9431b0e4af7f127d37678099cbe5333b744c1bb5`): fork CI is not cleanly running on this tip (NM/CI runs concluded without usable Behavior jobs; approve/rerun hit workflow-file / not-waiting errors). Greptile still flags a race where a later status append can make a newer steering record look older than the pause being classified (`no_inbox_activity_since_status` / status mtime vs snapshot). Your nanosecond mtime fix addresses the same-second case; that residual still needs a clear answer.

**Contract-class:** **new-default** — unconfigured watcher now trusts a `done:`+`paused:` shape as a non-wedge fast path, and every brief always instructs that shape. Not restore of an already-specified default path (the protocol previously left the both-true case unspecified).

**VISION:** R1 (peace/wedge noise) motive aligns; R2 Authority — always-on watcher fast-path is a default-behavior change, so no auto-merge; flag Firstmate for a captain card only once CI is green and Greptile residuals are closed. R3 scripts-vs-judgment aligns (deterministic classify/watch).

**Waiting on the author** (not a captain hold yet): get CI actually running green on this tip, and close or refute the status-mtime-vs-snapshot Greptile P1. When otherwise ready except the new-default decision, say so and I will flag Firstmate — I will not auto-merge.

kunchenguid:
Speaking as Kun's firstmate:

Follow-up: CI run `33904051235` is now attempt 2 and Behavior jobs are queued (NM body-compliance still red / attestation MISSING). Prior stamp's \"CI broken\" line is superseded — still **waiting on the author** for the Greptile status-mtime residual and for green CI; still **new-default** (no auto-merge; captain card only once otherwise ready).

PR #3685, opened 2026-09-04
fix(bin): bound shellcheck memory per lint shard

ikehoff:
## Intent

The captain reported (2026-09-03): "we just suffered through three OOM crashes due to spellcheck being unbounded. There is a fix that landed locally but you are not tracking that fix yet." The linter bin/fm-lint.sh runs shellcheck per shard with no memory bound; a pathological analysis grew until the kernel OOM killer took down the machine three times. The fix already exists locally as an uncommitted change (bounding each shard's shellcheck with ulimit -v under FM_LINT_SHELLCHECK_MAX_MB, default 4096 MB) and must be shipped so every firstmate home gets it.

## What Changed

- `bin/fm-lint.sh` now runs each shard's ShellCheck inside a subshell under `ulimit -v`, capped at `FM_LINT_SHELLCHECK_MAX_MB` (default 4096 MB) so a pathological analysis fails that shard instead of triggering the kernel OOM killer machine-wide. The cap is best-effort: it is enforced where the platform honors RLIMIT_AS (Linux) and silently ignored elsewhere (macOS).
- The parent validates `FM_LINT_SHELLCHECK_MAX_MB` as a positive integer (exit 2 with a clear message on `abc`, `4G`, `0`, `-1`), converts it to KiB, and passes it to workers via `FM_LINT_INTERNAL_SHELLCHECK_MAX_KIB`, which the `--internal-worker` entry point now requires.
- `tests/fm-lint.test.sh` extends the ShellCheck stub to record its `ulimit -v`, and adds tests asserting the default 4096 MB cap, the `FM_LINT_SHELLCHECK_MAX_MB=512` override, and rejection of malformed values without running ShellCheck. The cap test self-skips on platforms where `ulimit -v` is unsupported.

## Risk Assessment

✅ Low: The change is a contained two-file hotfix: the ulimit wrapper preserves the worker's PID/trap semantics, the parent-side validation rejects every malformed input the prior round identified before any ShellCheck runs, and the new tests assert observable behavior through the real script.

## Testing

- Outcome: ⚠️ 1 error across 1 run (3m40s)

## Pipeline

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>

- ⚠️ `bin/fm-lint.sh:88` - FM_LINT_SHELLCHECK_MAX_MB is fed straight into arithmetic and ulimit with no validation, unlike FM_LINT_JOBS which is validated with a clear message at bin/fm-lint.sh:179-181. Traced concrete inputs: FM_LINT_SHELLCHECK_MAX_MB=abc under `set -u` aborts the subshell with &#39;abc: unbound variable&#39; before ShellCheck runs (shard rc=1, message replayed as if it were a lint finding); FM_LINT_SHELLCHECK_MAX_MB=4G fails with &#39;value too great for base&#39; (rc=1); FM_LINT_SHELLCHECK_MAX_MB=0 sets `ulimit -v 0` so the exec itself dies (observed rc=139 segfault). None of these mention the memory cap, so a user who mistypes the override sees an opaque shard failure and every lint run fails. Fix: validate the value once in the parent alongside JOBS (digits only, greater than zero; reject or treat 0/unlimited explicitly with a clear &#39;fm-lint.sh: FM_LINT_SHELLCHECK_MAX_MB must be a positive integer&#39; message) and pass the validated KiB value to the worker.
- ℹ️ `bin/fm-lint.sh:34` - On platforms where RLIMIT_AS cannot be set (bash on macOS accepts `ulimit -v` but the kernel does not enforce it, and some containers pin a lower hard limit), the `|| true` and `2&gt;/dev/null` mean the cap is silently not applied. That is the right choice for keeping lint runnable, and the test correctly skips there, but the header comment at line 34-36 states the cap unconditionally. Consider wording it as best-effort so an operator on macOS does not assume they are protected. No code change required.

🔧 Fix: validate FM_LINT_SHELLCHECK_MAX_MB and document best-effort cap
1 info still open:

- ℹ️ `bin/fm-lint.sh:196` - The validation accepts any digit string, so a value beyond 64-bit range wraps silently in `$(( SHELLCHECK_MAX_MB * 1024 ))`. Traced: FM_LINT_SHELLCHECK_MAX_MB=99999999999999999999 passes the case guard, the arithmetic yields 2123646838278978560 KiB, `ulimit -v` accepts it, and ShellCheck runs effectively uncapped with no error. This only affects absurd operator input and the documented cases (abc, 4G, 0, -1) are all rejected correctly, so no change is required for this hotfix; an upper bound on the digit count would close it if desired.
</details>

<details>

</details>

<details>

✅ No issues found.
</details>

<details>

</details>

<details>

✅ No issues found.
</details>

### DIFF
--- a/bin/fm-lint.sh
+++ b/bin/fm-lint.sh
@@ -31,6 +31,9 @@
 # Each shard writes separate diagnostics, and the parent replays those outputs in
 # deterministic shard and root order after every worker finishes. FM_LINT_JOBS=1
 # runs the same shards serially with byte-identical diagnostics and exit selection.
+# Each shard's ShellCheck runs under a virtual-address-space cap of
+# FM_LINT_SHELLCHECK_MAX_MB (default 4096) so a pathological analysis fails that
+# shard instead of triggering the kernel OOM killer machine-wide.
 #
 # Optional quiet telemetry writes one bounded TSV snapshot of content and source
 # graph identity, wall/CPU/RSS, shard load, and competing ShellCheck processes.
@@ -79,7 +82,12 @@ fm_lint_worker() {  # <manifest> <output-dir> <shard-index>
     if [ "${FM_LINT_INTERNAL_FAST:-0}" -eq 1 ]; then
       shellcheck_args+=(--extended-analysis=false)
     fi
-    "$FM_LINT_SHELLCHECK" "${shellcheck_args[@]}" -- "${roots[@]}" > "$output.out" 2>&1 &
+    # Bound shellcheck's address space so a pathological analysis fails this
+    # shard loudly instead of triggering the kernel OOM killer machine-wide.
+    (
+      ulimit -v "$(( ${FM_LINT_SHELLCHECK_MAX_MB:-4096} * 1024 ))" 2>/dev/null || true
+      exec "$FM_LINT_SHELLCHECK" "${shellcheck_args[@]}" -- "${roots[@]}"
+    ) > "$output.out" 2>&1 &
     FM_LINT_WORKER_SHELLCHECK_PID=$!
     wait "$FM_LINT_WORKER_SHELLCHECK_PID" || rc=$?
     FM_LINT_WORKER_SHELLCHECK_PID=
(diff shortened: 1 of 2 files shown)


ikehoff:
Rebased onto current main (daaffdb5); conflicts resolved in bin/fm-lint.sh and tests/fm-lint.test.sh. Behavior unchanged: main's per-path ShellCheck loop (added by #3778) now runs each invocation under the same FM_LINT_SHELLCHECK_MAX_MB cap as the source-following path. Commit author email rewritten to the GitHub noreply address so the push clears email-privacy checks.

PR #3698, opened 2026-09-04
feat: surface detached background work in fleet snapshots

att430:
## Intent

Make detached background work visible.

The captain, 2026-09-04: "i would like to still have some indication that its a process in action, i have zero visibility, can we add it to the fleet dashboard"

THE CONCRETE CASE, running right now: a volumebot data collector, `artifacts/collect_E.py`, started 2026-09-03 22:25 UTC with a 26-hour budget. It is `setsid`-detached with no agent attached, writing to `data/volumebot-older-cohort-collection/artifacts/sampleE.jsonl` outside any worktree so it survives teardown. Nearly seven hours in it has written 25,484 rows across 411 cycles.

It is doing real work the captain is waiting on - it makes the difference between an underpowered null and a real verdict - and **nothing anywhere shows it exists.** It is not an agent, so it has no pane, no status file and no row in any fleet view. If it died, no stale alarm would fire and no wake would reach firstmate. Firstmate has been checking its row count by hand.

He understood the architecture correctly and asked the right question: this pattern is CHEAPER and more robust than keeping an agent alive for 26 hours, and firstmate expects more of it - so the gap is worth closing properly rather than special-casing this one process.

## What Changed

- Add a background-work registry and CLI for registering, observing, listing, and retiring detached non-agent processes using PID identity and bounded progress probes.
- Include tracked background work in bearings snapshots with explicit liveness and progress states while clearly marking it as visibility-only, unsupervised work.
- Document the operator and agent workflows, including safe registration, stale-progress handling, and the distinction from process-event sources.

## Risk Assessment

✅ Low: Captain, the change is bounded and the reviewed source preserves registered identities, explicit unknown/degraded states, lock-free reads, and the public collection deadline without introducing a substantiated merge-blocking risk.

## Testing

- Outcome: ⚠️ 1 error across 1 run (8m53s)

## Pipeline

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>
<summary>🔧 **Review** - 1 issue found → auto-fixed (12) ✅</summary>

- 🚨 `bin/fm-background-work.sh:29` - The required behavior is “add it to the fleet dashboard,” but this change only introduces a JSON interface described as being “for dashboards”; no fleet snapshot, Bearings projection, dashboard payload, or renderer consumes it. A registered collector therefore remains absent from every existing fleet view unless someone manually runs this new command. Integrate this interface at the canonical fleet snapshot boundary and project the records into the dashboard, with an executable dashboard/snapshot behavior test.

🔧 Fix: Project background work into fleet snapshots
2 errors still open:

- 🚨 `bin/fm-background-work.sh:479` - `list --json` probes records serially, each with its own configurable timeout and with no total deadline or record bound. For example, 20 live records with hung probes and the default 2-second timeout delay every Bearings/dashboard snapshot by roughly 40 seconds; registration permits arbitrarily many records and larger timeouts. Fixing this requires authorizing a bounded/concurrent collection contract that still emits honest per-record unknown states instead of silently dropping work.
- 🚨 `bin/fm-background-work.sh:287` - A `process_is_zombie` result of 2 (status unreadable) is ignored, after which a readable matching identity is labeled `alive`. The same unchecked result at registration can adopt such a process. Handle the unknown return explicitly: refuse registration and project existing records as unknown, preserving the requirement that unresolved process state never looks healthy.

🔧 Fix: Bound concurrent background-work observation honestly
1 error still open:

- 🚨 `bin/fm-background-work.sh:551` - The promised total collection deadline still starts only inside `collect_records`, after this unbounded serial loop has encoded every registry entry with multiple `jq` processes; the later serial re-read and final `jq -s` are also outside the watchdog. A registry with sufficiently many records can therefore keep Bearings/dashboard collection blocked well beyond `collection.budget_seconds`, even when no progress probe hangs. Apply the authorized whole-collection bound at the earliest list/snapshot boundary while preserving an explicit unknown row for every registered item and disclosing truncation.

🔧 Fix: Bound complete background-work snapshot collection
1 error still open:

- 🚨 `bin/fm-background-work.sh:570` - The total deadline still does not include the serial placeholder-encoding loop. With 32 valid records, line 570 repeatedly parses metadata and launches multiple `jq` processes before checking the deadline at line 573; if this consumes the budget, collection has already exceeded its promised bound, and final assembly is then granted an additional second. Apply the deadline/watchdog around the entire list operation, including placeholder generation, enumeration, rereading, and assembly, while retaining an unknown row for every record.

🔧 Fix: Bound fallback encoding within collection deadline
1 error still open:

- 🚨 `bin/fm-background-work.sh:570` - Registry membership is enumerated twice without holding `.registry.lock`, and the second pass stores live `fallback.json` paths for later assembly. A concurrent `retire` can delete one of those files after line 587 but before `jq` reads it, causing the entire list to fail; Bearings then replaces all tracked work with one `(registry)` error row, hiding unrelated active processes. Capture a stable record set and fallback data under the registry lock, then perform bounded observation against that snapshot.

🔧 Fix: Stabilize background-work snapshots during retirement
1 error still open:

- 🚨 `bin/fm-background-work.sh:570` - The collection deadline begins at line 561, but this blocking registry-lock acquisition has no bound. A paused or wedged `register`/`retire` holder can therefore keep `list --json`, Bearings, and the dashboard blocked indefinitely, contradicting the authorized whole-operation deadline. Use the existing `fm_lock_acquire_wait_bounded` with the remaining collection budget and, on expiry, return a disclosed collection-level unknown/error result rather than waiting past the deadline.

🔧 Fix: Bound registry lock within collection deadline
1 error still open:

- 🚨 `bin/fm-background-work.sh:604` - The authorized whole-operation deadline is still bypassed while the registry lock is held: fallback rows are copied serially with ordinary `cp`, outside any watchdog. A blocked copy (for example, a replaced `fallback.json` FIFO or stalled filesystem read) keeps `list --json`, Bearings, and the dashboard blocked indefinitely. The pre-lock fallback encoding is likewise unbounded. Put the entire list implementation behind one deadline at its public boundary so setup, stable snapshot capture, observation, and assembly share the same hard budget and timeout returns the disclosed unknown collection result.

🔧 Fix: Enforce public background-work collection deadline
1 error still open:

- 🚨 `bin/fm-background-work.sh:662` - The hard-timeout fallback emits only a synthetic `(registry)` row, so the concrete blocked-`fallback.json` sequence now tested makes the registered `blocked` item disappear. This contradicts the authorized contract that timeout “must not hide registered work as healthy or absent” and must preserve an explicit unknown row for every registered record. The honest remedy likely needs a deadline-safe membership snapshot or other durable cached index, which extends the machinery and therefore requires authorization.

🔧 Fix: Preserve blocked work identities through collection timeouts
2 errors still open:

- 🚨 `bin/fm-background-work.sh:599` - On registry-lock timeout, the list still replaces all registered identities with a single `(registry)` row. The authorized invariant says “Every registered record appears in the output, always”; the existing lock-contention test concretely registers `visible`, holds this lock, and then accepts output where `visible` disappears. Since the captain already authorized a cached membership index if enumeration cannot be deadline-safe, implement that authorized fallback so lock contention returns every known identity as unknown and discloses degradation.
- 🚨 `bin/fm-background-work.sh:561` - The new unknown-row encoder interpolates a directory-derived ID directly into JSON. A malformed registry directory such as `bad&#34;id` with no regular fallback produces invalid staged JSON; final `jq -s` then fails and hides every otherwise valid record. Encode the row with `jq --arg` (as other JSON paths do) so corrupt record names remain safely represented as unknown.

🔧 Fix: Preserve identities during registry lock degradation
2 errors still open:

- 🚨 `bin/fm-background-work.sh:667` - The prior fix removed the public hard-timeout wrapper, leaving snapshot capture unbounded. A registered record whose `fallback.json` is a FIFO passes the file checks and blocks forever in `cp`, so `list --json`, Bearings, and the dashboard exceed the promised total deadline; the updated regression test itself constructs this exact state but would hang. Reconcile the authorized identity-preserving fallback with a hard timeout covering the entire public list operation.
- 🚨 `bin/fm-background-work.sh:758` - Registry membership and the cached index are not updated transactionally. During retirement the record is deleted before `refresh_membership_index`; if that refresh fails, the command errors but a retry returns `absent` without repairing the cache, leaving the retired identity permanently visible during later lock contention. Registration has the inverse partial-failure risk if rollback removal fails. Making the cache a dependable fallback requires authorized crash/error-consistent update semantics rather than another local symptom patch.

🔧 Fix: Bound lock-free background-work visibility
1 error still open:

- 🚨 `bin/fm-background-work.sh:667` - The timeout fallback is not guaranteed any execution time. The outer `fm_run_timed` allows `_list` `COLLECTION_BUDGET - 1` seconds, but its timeout implementation may spend an additional one-second TERM-to-KILL grace period. A registered probe that ignores TERM can therefore make the inner call return at the overall deadline; `remaining` becomes zero here, `list --json` errors, and every registered row disappears instead of being returned as UNKNOWN. Capture lock-free membership before starting the timed probing phase, then use that captured membership directly if probing times out, so fallback encoding does not depend on budget remaining after timeout cleanup.

🔧 Fix: Precompute timeout-safe background-work membership
1 error still open:

- 🚨 `bin/fm-background-work.sh:650` - Valid record IDs may begin with `-`, but cached fallback membership passes IDs to `jq` without an option terminator. For example, registering `-x` makes `jq` parse it as an unknown option, so every `list` call fails while precomputing the fallback; `--version` can instead produce non-JSON fallback output. Insert `--` before the expanded ID arguments so all IDs accepted by `valid_id` remain positional strings.

🔧 Fix: Handle dash-prefixed background-work IDs safely
</details>

<details>

</details>

<details>

✅ No issues found.
</details>

<details>

🔧 Fix: Suppress intentional jq variable expansion warning
</details>

<details>

✅ No issues found.
</details>

### DIFF
--- a/.agents/skills/background-work/SKILL.md
+++ b/.agents/skills/background-work/SKILL.md
@@ -1,0 +1,60 @@
+---
+name: background-work
+description: >-
+  Agent-only procedure for making detached non-agent work visible.
+  Use before launching or adopting a detached process that should remain visible
+  after its agent or worktree goes away, and before retiring that visibility.
+user-invocable: false
+metadata:
+  internal: true
+---
+
+# background-work
+
+Load this before launching or adopting detached non-agent work that the captain may wait on.
+
+Tracked background work is visibility, not supervision.
+It has no agent endpoint, watcher dependency, automatic wake, restart, or recovery behavior.
+Use a process-event source instead when the requirement is to turn a blocking external result into a durable wake.
+
+## Register visibility
+
+Start the detached process through the task's authorized workflow, capture its actual PID, and then register that already-running process through the supported command:
+
+```sh
+bin/fm-background-work.sh register <id> \
+  --description "<plain-language activity>" \
+  --task "<owning task or investigation>" \
+  --pid <pid> \
+  --started-at <UTC-RFC3339> \
+  [--expected-finish-at <UTC-RFC3339>] \
+  [--cwd <directory>] \
+  [--stale-after <seconds>] \
+  [--progress-timeout <seconds>] \
+  --progress <argv>...
+```
+
+Registration adopts a live process and refuses when its process identity cannot be recorded.
+Give the progress probe the cheapest deterministic argv that returns one short human-readable line, such as a row count or cycle number.
+Use absolute data paths or set `--cwd` explicitly when the process outlives a disposable worktree.
+Never put secrets in the description, task, working directory, or progress argv because the private record is still operational state that diagnostics may expose.
+The command header and `--help` own exact validation, timing, storage, and output semantics.
+
+## Read and retire visibility
+
+Use the supported list rather than reading private state:
+
+```sh
+bin/fm-background-work.sh list
+bin/fm-background-work.sh list --json
+```
+
+Treat `dead`, `stalled`, and `unknown` as observations requiring judgment, not as automatic recovery instructions.
+The first successful progress sample is `unknown` because one value alone cannot prove motion.
+Retire a record only when its work no longer needs to remain visible:
+
+```sh
+bin/fm-background-work.sh retire <id>
+```
+
+Retirement removes only the visibility record and never signals the process.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -377,8 +377,8 @@ jobs:
           snapshot_output=$(/bin/bash tests/fm-fleet-snapshot-view.test.sh)
           printf '%s\n' "$snapshot_output"
           snapshot_count=$(printf '%s\n' "$snapshot_output" | grep -c '^ok - ')
-          [ "$snapshot_count" -eq 16 ] || {
-            echo "::error::expected 16 snapshot/fleet-view tests, got $snapshot_count"
+          [ "$snapshot_count" -eq 17 ] || {
+            echo "::error::expected 17 snapshot/fleet-view tests, got $snapshot_count"
             exit 1
           }
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -385,8 +385,8 @@ jobs:
           bearings_output=$(/bin/bash tests/fm-bearings-snapshot.test.sh)
           printf '%s\n' "$bearings_output"
           bearings_count=$(printf '%s\n' "$bearings_output" | grep -c '^ok - ')
-          [ "$bearings_count" -eq 49 ] || {
-            echo "::error::expected 49 Bearings tests, got $bearings_count"
+          [ "$bearings_count" -eq 50 ] || {
+            echo "::error::expected 50 Bearings tests, got $bearings_count"
             exit 1
           }
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -116,6 +116,7 @@ state/               runtime records and signals; gitignored
   x-watch.check.sh   generated Relay poll shim; present only when opted in (section 14)
   tool-updates.check.sh  generated watched-tool update poll shim and its .check-trust binding; present only after bin/fm-tool-update-check.sh arm; its report record .tool-updates is what keeps one pending update from being reported on every poll
   pending-replies/   parent-owned secondmate pending-reply records (correlation id, delivery vs reply, recovery, escalation); fm-pending-reply-lib.sh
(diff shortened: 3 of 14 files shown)


att430:
Withdrawn by the author.

PR #3699, opened 2026-09-04
feat(bin): add outward Telegram notifications

att430:
## Intent

Build phase 1 of the MIST Telegram integration: outward notifications only.

The captain, 2026-09-04: "go with your telegram recommendation" - approving option A from the scope, which is phase 1 outward only.

WHY IT EXISTS, in his own words earlier tonight: "this is part of the problem of i guess my own drift from the project and why we have the dashboard." The dashboard's decision list closed part of that gap, but a dashboard is PULL. This is PUSH. Tonight a five-hour session had four decisions waiting on him and nothing could reach him away from the terminal.

THE SCOPE IS ALREADY WRITTEN AND YOU MUST FOLLOW IT: `data/fm-telegram-scope/report.md`. Read it before writing code - especially sections 4, 9 and 10. It is 56KB and it did the design work; you are implementing its conclusions, not re-deciding them.

PHASE 1 ONLY. The bot SENDS AND NEVER LISTENS. `getUpdates` is never called. Inbound notation is phase 2 and is NOT authorised. Actions over Telegram are permanently out of scope by the captain's decision.

The bot token is already stored at `~/.mist-telegram-token`, mode 0600, and is currently inert.

## What Changed

- Add opt-in, send-only Telegram notifications for captain decisions, PRs ready for review, failed work, and merged work.
- Queue deduplicated cards in a durable outbox for asynchronous delivery, with credential refusal, internal-identifier redaction, bounded retries, and arm/disarm/status commands.
- Wire notifications into existing outcome publishers and document configuration, operational behavior, and failure handling with dedicated test coverage.

## Risk Assessment

✅ Low: The outward-only Telegram integration is well-bounded, preserves publisher behavior, keeps network delivery asynchronous, and the reviewed fixes structurally preserve URLs while redacting required internal identifiers without a substantiated remaining defect.

## Testing

- Outcome: ⚠️ 1 error across 1 run (17m20s)

## Pipeline

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>
<summary>🔧 **Review** - 2 issues found → auto-fixed (9) ✅</summary>

- 🚨 `bin/fm-pr-check.sh:157` - Notification deduplication uses only the task ID. After `pr-t1` has been delivered, registering a replacement PR for `t1` silently queues nothing because the delivered ledger still matches; `fm-merge-outcome-lib.sh:94` similarly uses `merged-$id`, despite the existing merge marker explicitly supporting distinct canonical PR identities. Silent inactive failures also omit their incarnation at `fm-inactive-reconcile.sh:526`. Include the canonical PR identity—or failure fingerprint/incarnation—in each publisher key so distinct outcomes remain reachable while exact retries deduplicate.
- 🚨 `bin/fm-telegram-lib.sh:258` - The scrubber does not uphold the stated rule that branch and harness names never reach captain-facing cards. For example, `note=build failed on branch fm/task-x under claude` passes unchanged because only `branch=...`/`harness=...` fragments are removed. Failure notes come from free-form status text, so this path is reachable without error. Redact actual known internal identifiers at the shared card-safety boundary, or restrict publishers to typed safe fields rather than relying only on `key=value` syntax.

🔧 Fix: Fix Telegram card deduplication and identifier redaction

- 🚨 `bin/fm-inactive-reconcile.sh:526` - The secondmate fallback still keys failures only by secondmate/task/state. If a configured secondmate relaunches the same task ID and both incarnations fail before writing terminal ledger lines, the second card is silently deduplicated. Append the already-computed fingerprint/incarnation to the Telegram card key while preserving parent-channel semantics.
- 🚨 `bin/fm-telegram-send.sh:216` - A valid HTTP 200 response such as `{&#34;ok&#34;: true}` is treated as rejection because success is recognized only by the exact substring `&#34;ok&#34;:true`. Parse the JSON response semantically so harmless whitespace or field formatting cannot leave a successfully sent card queued and emit a false configuration error.
- ⚠️ `bin/fm-merge-outcome-lib.sh:94` - If digest generation fails after the merge lock is acquired, this new early return leaves the lock held and skips the existing outcome publication. Release the lock before returning, or compute the digest before lock acquisition while retaining the locally-held-input constraint.

🔧 Fix: Decouple Telegram delivery from outcome recording

- 🚨 `bin/fm-telegram-send.sh:132` - `telegram_response_ok` still does not parse the response semantically: a valid 200 body such as `{&#34;result&#34;:{},&#34;ok&#34;:true}` is rejected, while malformed input beginning `{&#34;ok&#34;:true,` is accepted and causes the queued card to be deleted. Recognize the top-level `ok` boolean independent of field order while rejecting malformed JSON, without adding a required dependency.
- 🚨 `bin/fm-telegram-send.sh:295` - This contradicts the stated invariant “Absent configuration stays the absent feature: inert, exit 0, no warning”: invoking `arm` without configuration prints an error and returns 1. The implementation and documentation also claim an unconfigured home is a hard no-op. Confirm whether explicit `arm` is intended as an exception; otherwise make this path silent and successful.
- ⚠️ `bin/fm-telegram-lib.sh:289` - Both field and final-card limits use byte-oriented `cut -c` under the caller&#39;s locale. A value ending near the limit with a multibyte character can be split into invalid UTF-8, producing a queued card Telegram rejects even though its content is valid. Truncate on character boundaries (and retain a final byte-size guard appropriate for Telegram).

🔧 Fix: Harden Telegram responses, UTF-8 bounds, and arm guidance
2 errors still open:

- 🚨 `bin/fm-telegram-lib.sh:207` - A configured extra credential file is reduced to only its first line. For example, if `config/telegram-secret-files` names a multi-line `.env` and a card contains a credential from line 2, the value is absent from `fm_telegram_secret_values` and the card is sent. Check every distinctive credential line/value in each configured file, not only `head -1`.
- 🚨 `bin/fm-pr-check.sh:156` - After the PR metadata and merge poll have already been durably published, failure to compute the Telegram-only digest exits the command with status 1 and skips the parent-channel ready notification and final success output. This makes optional notification machinery change the outcome of PR registration when neither `shasum` nor `sha256sum` is available—the same coupling already removed from the merge path. Treat digest failure as card-only failure and continue the existing PR-ready publication/result path.

🔧 Fix: Decouple PR cards and enforce complete secret scanning
2 errors still open:

- 🚨 `bin/fm-telegram-lib.sh:214` - A path explicitly listed in `config/telegram-secret-files` is silently skipped when it is missing, unreadable, or a symlink. For example, a card queued while the file exists can later be sent after that credential file becomes temporarily unreadable, because the send-time scan treats the configured credential as absent and returns success. Return the existing incomplete-scan status instead so the card remains refused and the local diagnostic is emitted.
- 🚨 `bin/fm-telegram-lib.sh:222` - For a common dotenv entry such as `API_KEY=&#34;super-secret-value&#34;`, the scanner records the whole line and the quoted substring `&#34;super-secret-value&#34;`, but not the actual credential value `super-secret-value`. A card containing that raw credential therefore passes the positive check. Normalize matching paired single or double quotes around the value while retaining the existing whole-line and unmodified-value candidates.

🔧 Fix: Enforce complete Telegram credential verification
2 errors still open:

- 🚨 `bin/fm-telegram-send.sh:90` - `FM_TELEGRAM_API_BASE` is inherited by the armed sender and accepted without validation, so an ambient value such as `https://attacker.example` sends both the bot token (embedded in the URL) and private card text to that host. The code describes this as a test-only override; restrict non-default endpoints to an explicit test context or safe loopback endpoint.
- 🚨 `bin/fm-telegram-lib.sh:159` - Configuration loading accepts any readable token file without enforcing the documented mode-0600 requirement. A custom token path with mode 0644 therefore passes `arm` despite its message claiming a private file is required, leaving the bot credential exposed to other local users. Validate private ownership/permissions at the existing configuration boundary.

🔧 Fix: Enforce Telegram endpoint and token-file safety
1 error still open:

- 🚨 `bin/fm-telegram-lib.sh:156` - A configured token can still fail silently or select the wrong credential. If `config/telegram-token-path` exists but is unreadable, empty, or a symlink, `_fm_telegram_config_read` fails and this falls back to `$HOME/.mist-telegram-token`; if the configured target later disappears or becomes empty, lines 161/166 return the same silent “unconfigured” status. Thus the drain can silently stop with queued cards—or send through an unintended default bot—despite explicit configuration. Distinguish an absent token-path setting from an invalid one, and return the visible configuration-error status for every unusable explicitly selected token and for a configured token that becomes missing/empty at send time.

🔧 Fix: Distinguish invalid Telegram tokens from absent configuration
3 errors still open:

- 🚨 `bin/fm-telegram-lib.sh:155` - A present but empty, unreadable, symlinked, or nonnumeric `telegram-chat-id` returns the same silent status as an entirely unconfigured home. Concretely, an armed home whose chat-id becomes `abc` leaves queued cards undelivered forever without the required one-per-problem wake. Distinguish an absent setting from a present invalid setting at this configuration boundary, as already done for token configuration.
- 🚨 `bin/fm-telegram-lib.sh:358` - The harness-name scrubber also operates inside URLs because `/` satisfies both regex boundaries. A canonical PR URL such as `https://github.com/codex/repo/pull/1` is silently rendered as `https://github.com//repo/pull/1`, breaking both PR-ready and landed cards. Preserve complete URL fields while still scrubbing harness names from prose.
- 🚨 `bin/fm-telegram-lib.sh:356` - Absolute paths are removed only when preceded by whitespace or the start of the field. Common text such as `build failed (~/private/task)` therefore sends the private path unchanged, violating the documented identifier-redaction guarantee. Recognize punctuation-delimited absolute paths as well.

🔧 Fix: Preserve Telegram URLs and harden configuration redaction
2 errors still open:

- 🚨 `bin/fm-telegram-lib.sh:370` - The new unconditional parenthesized-path substitution also edits URL-shaped tokens, contradicting the required byte-preservation of URLs in free text. For example, `See https://example.test/(draft)/details` becomes `See https://example.test/details` without error. Preserve URL tokens before applying path redaction, then scrub only the surrounding prose.
- 🚨 `bin/fm-telegram-lib.sh:371` - Absolute-path redaction still only recognizes a limited punctuation set. Concrete notes such as `artifact=~/private/result` or `failed:&#34;~/private/result&#34;` retain the full private path, violating the documented guarantee that absolute paths never reach Telegram. Apply path recognition at the shared scrub boundary for all non-URL token delimiters.

🔧 Fix: Replace regex scrubbing with structural token classification
</details>

<details>

</details>

<details>

✅ No issues found.
</details>

<details>

🔧 Fix: Annotate Telegram configuration error output global
</details>

<details>

✅ No issues found.
</details>


att430:
Withdrawn by the author.

PR #3703, opened 2026-09-04
feat(harness): add dormant OMP candidate adapter with proven tool containment

Lakescape:
Adds the OMP candidate adapter, kept hard dormant, plus a regression guard proving its rendered tool surface cannot inherit the operator's Claude tools directory.

## What this contains

Two commits, replayed cleanly onto current `main`:

1. `feat(harness): re-land dormant OMP candidate` - the adapter itself, reconciled against the narrower non-executing contract.
2. `test(omp): prove rendered tool containment` - the containment regression guard.

## Dormancy is unchanged

OMP remains hard dormant. `refuse_omp_unverified_gates` still refuses every runnable launch. This PR waives nothing: executable provenance, effective configuration, and lifecycle proof all still gate activation. The binary is never executed by anything here.

The provenance guard refuses in ~20ms with `family=live-harness-optin` under `FM_OMP_TOOLS_LIVE_E2E=1`, and skips cleanly with `gate_skip=optin` without it. The refusal duration is itself evidence that no binary is resolved.

## Why the containment test exists

This is not a hypothetical. An accidental ambient OMP run enumerated `~/.claude/tools` and attempted to load two operator scripts as its own tools. Both failed on format, but the attempt proves the tool-inheritance hazard is real.

The test builds a fake operator home containing those two actual filenames, renders the manifest with `HOME` pointed at it, and asserts from known-good literals that `PI_CODING_AGENT_DIR` is the isolated directory and that neither the tools path nor either filename appears anywhere in the rendered output.

**Honest red-state note:** the adapter already isolated correctly, so no authentic pre-fix failure was observable without falsifying correct behavior. This is a regression guard over already-correct behavior, not a red-to-green fix, and no invented red state is claimed. That the guard *binds* was proven instead by controlled mutation: changing the expected literal made the suite exit 1 with `not ok - candidate OMP manifest inherited the captain Claude tools directory`, then it was reverted.

Isolation is argument-driven rather than `HOME`-derived, so if the renderer were ever changed to read `HOME`, the assertion fires.

## Documentation

`docs/verification/runtime-backends.md` states what is proven (isolated settings rendered, tool surface requested disabled, version pinned) and what is not (effective configuration, fallback isolation, effective tool construction). The 2026-08-24 RPC-invalidation finding and the Bun-cache `@oh-my-pi/pi-coding-agent` 18.0.4 finding are both preserved deliberately, so neither is re-attempted.

## Checks run locally

| Check | Result |
| --- | --- |
| `bin/fm-doc-audience-check.sh` | exit 0 |
| `bin/fm-lint.sh` over changed shell files | exit 0 |
| `bin/fm-test-run.sh tests/fm-omp-harness.test.sh` | exit 0, all passed |
| provenance guard, opt-in set | exit 1 in 20ms, designed refusal |
| provenance guard, opt-in absent | exit 0, `gate_skip=optin` |

https://claude.ai/code/session_01VneT4KxJB4wQCaK2FV6Ka1

### DIFF
--- a/.agents/skills/harness-adapters/SKILL.md
+++ b/.agents/skills/harness-adapters/SKILL.md
@@ -3,7 +3,7 @@ name: harness-adapters
 description: >-
   Agent-only reference for firstmate harness operations.
   Use before spawning or recovering a crewmate or secondmate, handling a trust dialog, sending a harness-specific skill invocation, interrupting or exiting an agent, resuming an exited agent, or verifying a new harness adapter.
-  Contains verified facts for claude, codex, opencode, pi, pi-signed, grok, kimi, cursor, gemini, and muse.
+  Contains verified facts for claude, codex, opencode, pi, pi-signed, grok, kimi, cursor, gemini, and muse, plus the dormant OMP candidate boundary.
 user-invocable: false
 metadata:
   internal: true
@@ -36,6 +36,8 @@ Deliver lifecycle actions only through `../../../bin/fm-control.sh <task-id> int
 Never type an interrupt key or exit command through `fm-send`, where routing-marked lifecycle text becomes chat.
 Trust handling is complete only when inspection proves the target started processing its instructions; delivery success alone is not proof.
 Muse and Gemini are verified only for crewmate and scout work, never a secondmate or primary.
+OMP is an unverified crewmate/scout candidate and remains hard dormant.
+Never resolve, probe, execute, dispatch, or describe OMP as live-ready; its selected reference owns the bounded tmux-only development contract and independent activation gates.
 
 ## Detection
 
@@ -90,7 +92,8 @@ A new tool remains undispatchable until the `verify` plan, its harness entry, ev
     "kimi": "references/harness/kimi.md",
     "cursor": "references/harness/cursor.md",
     "gemini": "references/harness/gemini.md",
-    "muse": "references/harness/muse.md"
+    "muse": "references/harness/muse.md",
+    "omp": "references/harness/omp.md"
   }
 }
 ```
--- a/.agents/skills/harness-adapters/references/harness/omp.md
+++ b/.agents/skills/harness-adapters/references/harness/omp.md
@@ -1,0 +1,16 @@
+# OMP candidate
+
+Oh My Pi 17.2.9 is a review-only candidate for crewmate and scout work, not a verified worker.
+Every runnable selection remains refused until immutable executable provenance is established, a supported session-free consumer proves effective configuration and tool containment against that exact artifact, and ATX-2170 independently proves First Mate interrupt, exit, and relaunch control.
+No mandatory prerequisite substitutes for another.
+
+The candidate requires caller-explicit `--harness omp` selection and a structurally qualified `--model <provider>/<model>` value.
+It is never an implicit default, secondmate, coordinator, or primary.
+The bounded development exception permits only the tmux backend and changes no fleet default or other adapter's backend resolution.
+An Orca selection must refuse and name tmux.
+
+`../../../bin/fm-spawn.sh --help` owns selection, model and backend validation, dormancy, trace ordering, and the non-executing refusal.
+`../../../bin/fm-omp-candidate-artifacts.sh` owns the exact requested environment, isolated settings, narrower manifest, input policy, and lifecycle extension.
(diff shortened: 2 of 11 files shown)


Lakescape:
Confirming the raw-launch finding: it reproduces, and it is a real gap in the dormancy claim this PR rests on.

`raw_launch_selects_omp` matches a literal `omp` token in the raw text, but that text is typed into the pane shell via `spawn_send_literal` and evaluated there. Any construct that *expands* to `omp` therefore runs the binary while evading the check.

Tested the predicate in isolation, without resolving or executing anything:

| raw launch text | detected |
| --- | --- |
| `omp` | yes |
| `/usr/bin/omp` | yes |
| `o"m"p` | **no** |
| `\omp` | **no** |
| `'o''m''p'` | **no** |
| `${X}mp` | **no** |

Four of seven bypass it.

I am not patching this in place, because tightening the regex does not fix it. Deciding what a shell command will execute from its text is undecidable, so any pattern is an arms race that loses to the next quoting form. The honest options are narrower:

1. Treat raw launch as what it is already documented to be - the operator's deliberate escape hatch for unverified adapters - and state the dormancy claim precisely as covering *selected* launches, keeping this predicate as a best-effort guard against the accidental case it was actually written for.
2. Expand the contract so raw launch is refused entirely while any dormant candidate adapter exists, accepting that this changes behaviour for every other unverified adapter too.

That is a scope decision rather than a code fix, so it is going to the repository owner rather than being resolved unilaterally in this branch. Tracked on our side; this PR should not be read as claiming raw-launch coverage until it is settled.

Everything else in the PR is unaffected: the containment guard, the selected-launch refusals, and the provenance gate all still hold.

kunchenguid:

First look on HEAD `3313e8243d37cfecde6b2c80856ffe937883c188` vs main `a5c64a0b6791b7643024436af9f737059d15cef2`. MERGEABLE/UNSTABLE. Lakescape is not a blocked author. No `.github/workflows/*` edits.

**Fork CI approval this pass: yes.** After diff review (dormant candidate + containment tests; no workflow/RCE surface), approved pending runs `33854580381` (CI) and `33854580405` (Require no-mistakes).

**Attestation: MISSING.** No `no-mistakes-pipeline-attestation:v1` HTML comment binds this HEAD. NM cannot pass until the PR is raised/rebound through no-mistakes with a MATCH attestation.

**CI/NM:** were `action_required` before approval; Greptile currently FAIL on the raw-launch dormancy gap. Re-check after the approved runs finish.

**Contract-class: opt-in.** OMP stays hard dormant: every runnable selected launch hits `refuse_omp_unverified_gates`, requires explicit `--harness omp` + qualified `--model`, tmux-only, never secondmate, never inherits default/trace selection. Unconfigured fleets do not start running OMP. The containment regression and provenance opt-in gate (`FM_OMP_TOOLS_LIVE_E2E`) match that posture.

**VISION.md per-rule**
- **One captain, one interface** — aligns (candidate stays below deck until gates clear).
- **Authority is explicit and never inferred** — aligns for selected launches (explicit harness + refused runnable path). **Open gap:** author-confirmed raw-launch expansion forms (`o"m"p`, `\omp`, `'o''m''p'`, `${X}mp`) bypass `raw_launch_selects_omp` while the pane shell can still execute `omp`, so the “hard dormant” claim does not currently cover the raw-launch escape hatch.
- **Scripts own the mechanics, agents own the judgment** — aligns (spawn refusals + artifact renderer are scripted).
- **A restart is a non-event** — aligns / N/A (no new durable obligation beyond adapter artifacts).
- **Delegation with a spine** — aligns (activation gates are explicit and non-substitutable).
- **The fleet outlives any vendor** — aligns (candidate adapter; other harnesses unchanged).
- **Scope** — aligns as harness-adapter surface with verification docs.

Security review for fork CI: no workflows; selected-path refusals and containment look intentional; approved runs. Product dormancy gap on raw launch is real (author reproduced it) but is a contract/scope issue, not a disguised workflow RCE — not a blocked-author / malicious close.

Author already said they will not patch the regex arms race in-place and want a scope choice (narrow the dormancy claim to selected launches vs refuse raw launch while any dormant candidate exists). That becomes a captain hold only once this is otherwise ready (MATCH attestation + green CI/NM). Until then it stays on the author: bind attestation via no-mistakes, let approved CI finish, and either land a claim-accurate dormancy boundary or leave the PR blocked on that settlement.

PR #3713, opened 2026-09-04
feat(bin): add agy (Antigravity CLI) as a crewmate/scout worker runtime

RooseveltAdvisors:
## Intent

Make agy (Antigravity CLI) usable as a firstmate worker runtime: wire the adapter so a crewmate or scout can be spawned on agy, reach its brief without a human answering a workspace-trust dialog, and report turn end back to firstmate's supervision.

## What Changed

- Registered `agy` across the adapter plane: harness detection now tests `ANTIGRAVITY_CONVERSATION_ID` before `CLAUDECODE` (agy does not clear an inherited marker), and `fm-spawn.sh` gained an agy launch template (`--dangerously-skip-permissions`, prompt-last `-i` flag ordering, `--effort low|medium|high` only) plus exit/interrupt/select keys, quota mapping, dispatch validation, process classification, and an `esc to cancel` delivery-busy guard. `--secondmate` on agy is refused, since there is no primary supervision protocol.
- Added `bin/fm-agy-trust.sh`, which pre-registers the task worktree in `trustedWorkspaces` in `~/.gemini/antigravity-cli/settings.json` so a fresh worktree does not park on the workspace-trust dialog (which renders no status text). Registration is scope-tested to linked worktrees of the named project, recorded in `state/<id>.agy-trust`, and withdrawn — by the exact spellings the spawn added — on teardown and on aborted-spawn cleanup.
- Added `bin/fm-agy-turnend-hook.sh`, which installs one key-preserving `firstmate-turn-end` entry in `~/.gemini/config/hooks.json` with a private token registry and per-worktree `.fm-agy-turnend` pointer; it touches the turn-end state file only when the Stop payload's `workspacePaths`, pointer, and token agree and `fullyIdle` is true. Teardown retires grok/kimi/agy registry entries together. New `tests/fm-agy-harness.test.sh` and `tests/fm-agy-signals-live-e2e.test.sh` are wired into `fm-test-run.sh`, and the adapter's verified facts (including the declined-tool-call and Escape-interrupt paths where no Stop fires) are documented in the harness-adapters skill reference and the runtime-backend docs.

## Risk Assessment

✅ Low: The change is an additive new crewmate/scout adapter that mirrors the established kimi/grok/muse patterns, its four cross-cutting edits (marker precedence, launch-boundary scrub, shared delivery-regex union, shared turn-end retirement loop) are each narrow and covered by behavior tests against real processes and real store/hook files, and the only mutation outside the firstmate home is guarded by a structural scope test, a fingerprint-and-refuse write, and a withdrawal on every abort and teardown path I could trace.

## Testing

- Outcome: ⚠️ 1 error across 1 run (27m54s)

## Pipeline

<details>

🔧 Fix: clear inherited agy marker, fix secondmate arg and live-guard cleanup

- ⚠️ `bin/backends/herdr.sh:1459` - fm_backend_herdr_server_ensure scrubs every harness identity marker from the long-lived herdr server&#39;s startup environment except agy&#39;s newly added ANTIGRAVITY_CONVERSATION_ID. docs/herdr-backend.md:209 states the adapter launches the server &#39;without inherited Firstmate home and directory overrides, harness identity markers, or the supervision-model override&#39; because &#39;Herdr passes its server startup environment to every later pane&#39;, and tests/fm-backend-herdr.test.sh:559 (test_server_ensure_scrubs_home_and_harness_identity) asserts exactly that. ANTIGRAVITY_CONVERSATION_ID is now a harness identity marker and is tested FIRST in bin/fm-harness.sh:63, ahead of CLAUDECODE. Failing sequence: a shell carrying ANTIGRAVITY_CONVERSATION_ID (an agy pane, or a terminal that ran agy) invokes a herdr-backed firstmate command; fm_backend_herdr_server_ensure starts the server, which inherits the marker and injects it into every pane it later creates. A claude session started by hand in such a pane resolves detect_own -&gt; agy, so `fm-harness.sh crew` on `default` dispatches every crewmate on agy and the new agy secondmate gate at bin/fm-spawn.sh:1407 hard-refuses secondmate launches. I confirmed the detection flip directly: with the marker exported, fm-harness.sh prints &#39;agy&#39; where it printed &#39;claude&#39; before. The launch-boundary clear added by the fix round at bin/fm-spawn.sh:3124 covers only the agent process fm-spawn itself starts, which is why the repo maintains both layers for CLAUDECODE/CURSOR_AGENT/PI_CODING_AGENT/GROK_AGENT. Remedy: add ANTIGRAVITY_CONVERSATION_ID to the unset list at bin/backends/herdr.sh:1459 and to the enumeration the fixture records at tests/fm-backend-herdr.test.sh:84.
- ⚠️ `bin/fm-teardown.sh:2573` - cleanup_firstmate_home_children removes each child task&#39;s grok, kimi, muse, cursor and pi state sidecars but was not extended with the new &lt;id&gt;.agy-turnend-token, although the top-level teardown list at bin/fm-teardown.sh:2919 was. Failing sequence: a secondmate home runs an agy crewmate, so fm-spawn writes $sub_state/&lt;child-id&gt;.agy-turnend-token (bin/fm-spawn.sh:2925). The home is torn down with --force; cleanup_firstmate_home_children calls remove_all_turnend_auth (which does correctly retire the global registry entry) and then deletes the child&#39;s .meta, but the explicit rm -f list at lines 2571-2575 omits the agy token, so that file outlives the task that owned it. The sub-state directory is not removed wholesale - that is precisely why the sibling rm -f list exists - so the orphan persists. The leftover is inert (its registry entry is already gone), but leaving state behind for a task whose records are being removed is the exact condition this list is maintained to prevent. Remedy: add &#34;$sub_state/$child_id.agy-turnend-token&#34; to the rm -f list at line 2573, matching line 2919.
- ℹ️ `bin/fm-control.sh:182` - The &#39;resume is not a control verb&#39; refusal enumerates the per-adapter resume contracts and was not extended with agy, although docs/agent-control.md:51 was updated in this same change to say &#39;codex, grok, and agy resume only from a session id printed at exit&#39; plus agy&#39;s extra caveat that --conversation=&lt;id&gt; restores the conversation but not the workspace. Failing sequence: an operator runs `bin/fm-control.sh &lt;id&gt; resume` on an agy task and gets a message listing codex/grok on one side and claude/pi/pi-signed/kimi on the other, with agy in neither clause, so it says nothing about the adapter they asked about. No wrong action results (resume is refused either way), but the runtime message and the doc the author wrote in this same change now disagree. Remedy: mirror the doc&#39;s wording into the message.

🔧 Fix: wire agy into herdr scrub, teardown, quota, dispatch and trust

- ⚠️ `bin/fm-quota-choose.sh:316` - The fix round resolved the prior ask-user finding by inventing a quota-axi provider family named &#39;agy&#39; rather than establishing the real one. Nothing in this branch verifies that quota-axi emits a provider row named &#39;agy&#39;: docs/verification/runtime-backends.md&#39;s new agy section records the trust dialog, both status bars, the Stop payload, the environment marker, resume and launch shapes from live observation, but no quota-axi evidence at all, and .agents/skills/harness-adapters/references/harness/agy.md has no quota row either. The existing families are vendor/credential names (muse -&gt; &#39;meta&#39;, grok, kimi, cursor, codex, claude); agy is a Gemini/Antigravity credential, so its family is most likely &#39;google&#39;/&#39;gemini&#39;, not &#39;agy&#39;. Failing sequence with a wrong name: `fm-quota-choose.sh --candidate agy:gemini-3.8-flash` against a real snapshot -&gt; effective_for_provider_model finds no `.providers[] | select(.provider == &#34;agy&#34;)` -&gt; returns {status:&#34;unknown&#34;} -&gt; the jq eligibility test at line 371 takes the `.status == &#34;unknown&#34;` branch and returns false -&gt; the candidate is skipped with no diagnostic. With agy as the only candidate the helper prints `none` and exits 1, telling the agent no candidate is quota-eligible while agy&#39;s actual provider family has full runway; in an ordered list it silently demotes agy behind every later candidate. That is a wrong routing answer produced without any error. The new test at tests/fm-quota-choose.test.sh:544-559 cannot catch this: it synthesizes the `agy` provider row into the fixture itself with jq, so it passes for any name the code and the fixture happen to agree on, and its comment asserts &#39;quota-axi reports it as its own provider family&#39; as fact. Remedy is a verification decision, not a mechanical correction: either capture a real `quota-axi` snapshot for the Antigravity credential and use the family it actually reports (rewriting the fixture to that name), or leave agy unmapped as the pre-fix code did and accept the loud `unknown harness: agy` refusal. Both change which routing answers the helper gives, so this needs the author&#39;s call.
- ⚠️ `bin/fm-spawn.sh:2598` - Every agy spawn permanently appends its worktree path (one or two spellings) to `trustedWorkspaces` in the operator&#39;s real ~/.gemini/antigravity-cli/settings.json, and nothing ever removes it. bin/fm-teardown.sh is otherwise symmetric for this adapter - it removes `$WT/.fm-agy-turnend` (lines 2798, 2785, 2576), `$STATE/$ID.agy-turnend-token` (2920, 2574), and the private registry entry via remove_all_turnend_auth - but there is no de-registration counterpart for the trust write, and fm-agy-trust.sh offers no remove verb. Concrete growth path: on the orca backend the task worktree is created as `fm-$ID` (bin/fm-spawn.sh:2171 -&gt; 2390 -&gt; bin/backends/orca.sh:134), so each task gets a distinct absolute path; after teardown removes that worktree the path is dead, yet the entry stays. N agy tasks leave N dead absolute paths in a vendor-owned config file the change explicitly promises to leave otherwise intact. Two consequences: unbounded growth of a file firstmate does not own, and a stale trusted path that a later, unrelated checkout at the same path would inherit without a dialog. Remedy is not a one-line correction - it needs a new teardown-side de-registration path carrying the same read-fingerprint-rename concurrency guard the registration has, plus the both-spellings removal - so it extends the change rather than correcting it, and needs authorization.
- ℹ️ `tests/fm-agy-signals-live-e2e.test.sh:149` - The final assertion `grep -Fq &#34;$WT&#34; &#34;$STORE&#34; &amp;&amp; fail &#34;the guard left its worktree in the operator&#39;s trusted workspaces&#34;` runs one line AFTER `restore_store` (line 148), which copies the pre-test backup over $STORE. $WT is a fresh mktemp path that by construction never appeared in that backup, so the grep can never match and the check can never fail - it proves nothing about whether the guard leaked, which is exactly what the accompanying `pass` message claims it proved. The check that would actually fail on a leak is the same grep run BEFORE restore_store, i.e. against the store as the guard left it. Remedy: move the grep above line 148 (or grep the live store before restoring). Non-functional and no product decision involved.

🔧 Fix: retire agy workspace trust at teardown

- ⚠️ `bin/fm-spawn.sh:2602` - The teardown-side withdrawal added by the last fix round closes the accumulation only for tasks that reach a published meta record; a fresh agy spawn that aborts AFTER the registration leaks the entry permanently. Concrete sequence: the operator&#39;s ~/.gemini/config/hooks.json is malformed JSON (a case bin/fm-agy-turnend-hook.sh explicitly refuses on, readConfig -&gt; `agy hooks config is malformed JSON`). fm-spawn.sh:2598 has already registered the worktree in trustedWorkspaces; line 2602 then fails and line 2604 exits 1. No $STATE/$ID.meta is written (meta publish is ~400 lines later, near bin/fm-spawn.sh:3016), and bin/fm-teardown.sh:281 refuses outright for a task with no meta record, so retire_agy_workspace_trust can never run for that id. The path stays in the operator&#39;s vendor settings.json forever, which is exactly the invariant docs/configuration.md now claims (&#39;so a settings file firstmate does not own does not accumulate one dead worktree path per task&#39;). The same holds for every other abort between 2598 and meta publish (busy-arm failure at 2664, the AGY_AUTH_DIR mktemp at 2925, backend pane creation); spawn_abort_cleanup (bin/fm-spawn.sh:783, EXIT trap at 894) clears relaunch replacement wiring but has no agy arm. The new test test_refused_spawn_leaves_no_task_state exercises a trust REFUSAL (which registers nothing), so it cannot catch this. Remedy corrects the register/retire symmetry the change already commits to using the --remove verb it already ships: set a flag when the registration succeeds and add an arm to spawn_abort_cleanup that runs `fm-agy-trust.sh --remove &#34;$WT&#34;` when meta was never published (at minimum, withdraw in the hook-install failure branch at 2602).
- ℹ️ `bin/fm-spawn.sh:114` - The usage header enumerates the bare adapter names the harness positional accepts as `(claude|codex|opencode|pi|pi-signed|grok|kimi|cursor|muse)`, but this change added agy to that accepted set at bin/fm-spawn.sh:1194 and bin/fm-spawn.sh:1327, and updated every user-facing doc (AGENTS.md:202, docs/configuration.md:299, docs/architecture.md:235) accordingly. The adjacent per-harness wiring notes at bin/fm-spawn.sh:181-190, which describe kimi&#39;s global hook + registry + pointer, grok&#39;s global hook + pointer + state token, muse&#39;s session sidecar and cursor&#39;s transcript sidecar, likewise have no agy row despite agy using the same global-hook-plus-token shape as grok and kimi. No wrong behavior results; the in-file contract an author reads before editing this script now disagrees with the code and with the docs written in the same change.

🔧 Fix: withdraw agy trust when a spawn aborts
2 warnings still open:

- ⚠️ `bin/fm-teardown.sh:2800` - The teardown-side withdrawal is unreachable for exactly the case `fm-agy-trust.sh --remove` was built to handle: a task whose worktree directory is already gone. All four call sites sit inside a `[ -d &#34;$WT&#34; ]` / `[ -d &#34;$child_wt&#34; ]` guard - bin/fm-teardown.sh:2790 gates the orca call at 2800, bin/fm-teardown.sh:2804 gates the treehouse call at 2814, and bin/fm-teardown.sh:2547 and 2555 gate the two child calls at 2552 and 2561. Concrete sequence: an agy ship task registers `$WT` in the operator&#39;s `~/.gemini/antigravity-cli/settings.json` (bin/fm-spawn.sh:2619); the worktree is then removed out of band (a captain `rm -rf`, an orca worktree already reclaimed, a pool worktree pruned) before teardown runs; `fm-teardown.sh &lt;id&gt; --force` proceeds, `[ -d &#34;$WT&#34; ]` is false, neither branch fires, and no other code path withdraws the entry. `remove_all_turnend_auth` and every `state/&lt;id&gt;.*` removal below still run unconditionally, so the task is fully torn down while the vendor settings keep a dead absolute path forever - the precise accumulation docs/configuration.md:337 now claims cannot happen. That the gone-worktree case is supported is not in doubt: bin/fm-agy-trust.sh:172-178 falls back to the absolute argument when the path no longer resolves, and tests/fm-agy-harness.test.sh:195 (`test_removal_works_after_the_worktree_is_gone`) pins that behaviour - the capability is shipped and tested but its only production caller is gated away from it. Remedy is a placement correction using machinery this change already ships, not new machinery: move `retire_agy_workspace_trust &#34;$WT&#34;` out of the `[ -d &#34;$WT&#34; ]` blocks to run unconditionally for a non-secondmate teardown (for example alongside `remove_all_turnend_auth &#34;$STATE&#34; &#34;$ID&#34;` at bin/fm-teardown.sh:2926), and likewise for the two child call sites. `$WT` comes from the task record and is absolute, and the helper is already best-effort (`|| true`), so no teardown can be stranded by the move.
- ⚠️ `bin/fm-agy-turnend-hook.sh:200` - `if (registryInfo === null) fs.mkdirSync(REGISTRY, { mode: 0o700 });` is a check-then-create with no EEXIST tolerance, so two agy spawns racing the first-ever install on a box abort one of them. Concrete sequence: no `~/.gemini/antigravity-cli/fm-turn-end.d` exists yet; a captain dispatches two agy crewmates concurrently; both reach `lstatOrNull(REGISTRY) === null`; the loser&#39;s `mkdirSync` throws `EEXIST`, which the outer `catch` at line 251 turns into `fm-agy-turnend-hook: refused: EEXIST: file already exists, mkdir ...`, exit 1; bin/fm-spawn.sh:2625 then aborts the spawn with &#39;refusing agy spawn because the global turn-end hook could not be installed safely&#39;. That is a spurious refusal - the directory the spawn needed now exists and is correct. The immediately preceding line 194 already uses `fs.mkdirSync(STATE_DIR, { recursive: true })`, and the established sibling installer uses the race-safe form too (bin/fm-kimi-turnend-hook.sh:254, `os.makedirs(REGISTRY, mode=0o700, exist_ok=True)` followed by an explicit chmod), so this is an omission rather than a deliberate divergence. Remedy: `fs.mkdirSync(REGISTRY, { recursive: true, mode: 0o700 })` (optionally with an explicit `fs.chmodSync(REGISTRY, 0o700)` to match kimi, since `recursive` mkdir does not re-apply the mode to a pre-existing directory). The symlink/non-directory refusal at lines 197-199 still runs first, so the safety property is unchanged.

🔧 Fix: retire agy trust by record, fix registry mkdir race
2 warnings still open:

- ⚠️ `bin/fm-teardown.sh:2927` - The last fix round moved `retire_agy_workspace_trust &#34;$WT&#34;` out of the `[ -d &#34;$WT&#34; ]` branches (correct) but placed it at line 2927, AFTER the task has already released the worktree - `fm_backend_remove_worktree` at bin/fm-teardown.sh:2804 for orca and `teardown_treehouse_return` at bin/fm-teardown.sh:2825 for a pool worktree. Two defects follow from that placement.

(1) POOL-REUSE RACE, and it is no longer agy-specific. Treehouse pool worktrees are reused at the same path - the adjacent comment at bin/fm-teardown.sh:2810 says so outright (&#34;Remove our hook file so a reused pool worktree cannot fire signals for a dead task&#34;). Concrete sequence: task A (any harness) tears down, returns `&lt;pool&gt;/wt-03` at line 2825; a captain&#39;s concurrent dispatch runs `treehouse get` (bin/fm-spawn.sh:2559), receives `wt-03`, and registers agy workspace trust for it at bin/fm-spawn.sh:2619; task A&#39;s teardown then reaches line 2927 and, because the call is now unconditional and harness-agnostic, removes `wt-03` from the operator&#39;s `trustedWorkspaces`. Task B&#39;s agy worker launches into an untrusted folder and parks on the workspace-trust dialog, which bin/fm-agy-trust.sh:28-31 records as drawing NO status-bar text at all - a pane indistinguishable from idle. That is precisely the wedge this whole feature exists to prevent, and it is now reachable from a NON-agy task&#39;s teardown. No shared lock closes the window: the task-set lock is taken by teardown only on the forced secondmate-home path (bin/fm-teardown.sh:2242), not on an ordinary ship teardown. Before this fix round the call ran inside the branch while the task still owned the worktree, so the race did not exist.

(2) DETERMINISTIC PARTIAL LEAK FOR ORCA. Registration records BOTH spellings of the worktree - resolved and logical (bin/fm-agy-trust.sh:165-171). `fm_backend_remove_worktree` deletes the directory, so by line 2927 `real_dir` fails and only the `WT_ARG` literal survives via the absolute-path fallback at bin/fm-agy-trust.sh:172-178. When the recorded spelling traverses a symlink, the resolved spelling stays in the operator&#39;s vendor settings permanently - the exact accumulation docs/configuration.md:337 now claims cannot happen.

Remedy is a placement correction using machinery the change already ships, not new machinery: move the unconditional `retire_agy_workspace_trust &#34;$WT&#34;` to just BEFORE the worktree-release block (above the `if [ &#34;$BACKEND&#34; = orca ] &amp;&amp; [ &#34;$KIND&#34; != secondmate ]` at bin/fm-teardown.sh:2787), where the task still owns the worktree and the directory has not yet been deleted. That keeps the gone-worktree case the round-5 finding asked for (the call is still unconditional, and `--remove` still falls back to the absolute argument) while restoring the ownership invariant. Apply the same move to the child call at bin/fm-teardown.sh:2578, which sits after the child&#39;s own `teardown_treehouse_return` at 2562. Neither new test catches this: `test_teardown_retires_the_agy_workspace_trust_entry` and `test_teardown_retires_the_agy_trust_entry_when_the_worktree_is_gone` are both single-task, and both still pass under the earlier placement.
- ⚠️ `bin/fm-spawn.sh:823` - The abort-cleanup arm added by the previous fix round guards its withdrawal on `[ ! -e &#34;$STATE/$ID.meta&#34; ]`, but it sits at bin/fm-spawn.sh:823 - BEFORE `spawn_fresh_commit_rollback` at bin/fm-spawn.sh:891, which is what deletes that very record (`fm_backlog_dispatch_rollback` -&gt; `fm_backlog_record_remove`, bin/fm-backlog-transition-lib.sh:553). So the guard reads an intermediate state, not the final one, and the leak it was written to close stays reachable.

Concrete sequence: a fresh (non-relaunch) agy ship spawn registers trust and sets `AGY_TRUST_REGISTERED=1` at bin/fm-spawn.sh:2624; the provisional task record is published at bin/fm-spawn.sh:3080 and `SPAWN_FRESH_COMMIT_PENDING=1` is set at bin/fm-spawn.sh:3015; the spawn then fails after publication - for example the trace-context clear at bin/fm-spawn.sh:3228 (`refusing to append the launch command`, exit 1), or the backlog dispatch failure whose own message at bin/fm-spawn.sh:3301 says &#34;its record was removed&#34;. The EXIT trap runs `spawn_abort_cleanup`: at line 825 `$STATE/$ID.meta` still EXISTS, so the withdrawal is skipped and `AGY_TRUST_REGISTERED` is cleared to 0; execution continues to line 891 where `spawn_fresh_commit_rollback` removes the meta. The run ends with no task record and the worktree path still in the operator&#39;s `~/.gemini/antigravity-cli/settings.json`. bin/fm-teardown.sh:281 refuses outright for an id with no meta, so no supported command can ever withdraw it - the same permanent leak the round-4 finding identified, just through the post-publish door instead of the pre-publish one. The new test `test_aborted_spawn_withdraws_the_trust_it_registered` cannot catch it: it aborts on a malformed hooks config at bin/fm-spawn.sh:2625, which is before the publish, and it asserts `[ ! -e &#34;$home/state/abortedspawn.meta&#34; ]` to confirm exactly that.

Remedy is an ordering correction inside the code this fix round already added, with no new state or machinery: move the `AGY_TRUST_REGISTERED` block from bin/fm-spawn.sh:819-829 to after the `SPAWN_FRESH_COMMIT_PENDING` block that ends at bin/fm-spawn.sh:895, so the meta-existence test observes the final record state. Both directions then stay correct - a surviving record (including the orca recovery record republished at bin/fm-spawn.sh:880, and a failed rollback) leaves the entry for teardown, and a rolled-back record is withdrawn here.

🔧 Fix: withdraw agy trust before release and after rollback
3 warnings still open:

- ⚠️ `bin/backends/tmux.sh:173` - `fm_backend_tmux_classify_process_name` was not taught agy&#39;s process name, so every control verb this change advertises for agy refuses on the default backend.

Concrete trace. bin/fm-spawn.sh:1350 launches `env -u CLAUDECODE ... agy --dangerously-skip-permissions ...`; `env` execs, so the pane&#39;s foreground comm and argv[0] are both the literal `agy`. In the classifier at bin/backends/tmux.sh:160 that base hits neither `muse|muse-bin-*` (172) nor `*claude*|*codex*|*opencode*|*grok*|*kimi*|pi|...` (173), is not a shell, and falls to the default arm: `fm_harness_path_name` scans FM_HARNESS_NAMES=(claude codex opencode grok kimi pi-signed pi) (bin/fm-session-lock-lib.sh:26) for a whole path COMPONENT and misses, and `fm_cursor_process_matches` misses. Verdict `other`. In `fm_backend_tmux_agent_state` (bin/backends/tmux.sh:265) `other` sets `fg_other=1`, so line 332 takes the else branch and the endpoint reads **`ambiguous`, never `alive`**.

What that breaks, on tmux, for a healthy agy crewmate:
- `fm-control.sh &lt;id&gt; interrupt` dies at bin/fm-control.sh:869 - &#34;endpoint reads &#39;ambiguous&#39; rather than a positively classified state; refusing to send a lifecycle key into an unattributed endpoint&#34;.
- `fm-control.sh &lt;id&gt; exit` dies the same way at bin/fm-control.sh:874, so the agent cannot be stopped through the control plane.
- `fm-control.sh &lt;id&gt; relaunch` cannot get past that exit, and its post-launch gate `wait_agent_state &#34;$LAUNCH_WAIT&#34; alive` (bin/fm-control.sh:844) could never be satisfied either.

This is a gap the change itself opened rather than a pre-existing one: it registered agy in `fm_control_harness_supported`, `fm_control_interrupt_key`, `fm_control_exit_command` and `fm_control_harness_supports_kind agy ship|scout` (bin/fm-control-lib.sh:66-108), and documented agy relaunch in docs/agent-control.md:51 - claiming a control surface that `fm_control_backend_state_verified tmux` (bin/fm-control-lib.sh:194) says is provable but the classifier cannot deliver. Herdr is unaffected (its classifier is API-based, bin/backends/herdr.sh:1896); zellij/orca/cmux already report `unverified`. No test covers it: `test_control_tables_carry_agys_verified_mechanics` exercises only the fm-control-lib tables, never the backend probe.

Remedy is the one-line parity muse got for the same reason, in the shared classifier rather than at any call site: add `agy) printf &#39;agent&#39; ;;` beside line 172, anchored exactly (never `*agy*`) - the same anchoring bin/fm-harness.sh:103 already applies to agy so `legacy`/`agyneja` cannot be misread.
- ⚠️ `bin/fm-quota-choose.sh:316` - `provider_for_harness` maps `agy` to a quota-axi provider family literally named `agy`, and nothing in the change establishes that quota-axi reports such a provider.

docs/verification/dispatch-auth.md is this repo&#39;s record of the provider names quota-axi actually emits (it names `codex`, `claude`, `cursor`, `copilot`, `grok` from live captures) and this change adds no agy evidence to it. The change also repurposed `agy` away from being the codebase&#39;s canonical *unknown harness* example - tests/fm-quota-choose.test.sh:234 and :538 previously asserted `error: unknown harness: agy` and were rewritten to use `copilot` - so the only thing now standing behind the new family name is a synthesized fixture: the new cases at tests/fm-quota-choose.test.sh:545 and :553 `jq &#39;.providers += [{&#34;provider&#34;:&#34;agy&#34;,...}]&#39;` into the snapshot, which proves the lookup routes to whatever string this line prints and cannot prove that string is the one quota-axi uses.

Concrete wrong result if the family is really `google`, `gemini`, or `antigravity`: `effective_for_provider_model agy default` finds no matching entry, `($p // null) == null` returns `{status:&#34;unknown&#34;}` (bin/fm-quota-choose.sh:328), the `.status == &#34;unknown&#34;` arm rejects the candidate, and an ordered dispatch that lists `agy:default` silently prints `none` and exits non-zero even when Antigravity quota is completely fresh. It fails closed and loudly at the exit code, so nothing is corrupted - but agy becomes permanently undispatchable through the quota helper the quota-array-dispatch skill points at (.agents/skills/quota-array-dispatch/SKILL.md:24), for a reason no message names.

The remedy is not a code correction - it is capturing `quota-axi --json` against a signed-in agy install and recording the observed provider name in docs/verification/dispatch-auth.md alongside the other families, then keeping or changing this mapping to match. That is new verification evidence rather than a fix to what the change already does, so it needs the author&#39;s call rather than an automated edit.
- ⚠️ `bin/fm-spawn.sh:891` - Moving the abort withdrawal after the fresh-commit rollback (correct, and it does close the post-publish leak) also moved it past the orca worktree removal, which reintroduces the partial-leak half of the same defect the previous round fixed in teardown.

Registration records BOTH spellings of the directory - the fully resolved `WT_REAL` and the caller&#39;s logical `WT_LOGICAL` (bin/fm-agy-trust.sh:165-171). `--remove` recomputes that pair only while the directory still resolves; once it is gone, the absolute-argument fallback at bin/fm-agy-trust.sh:172-178 sets BOTH to the literal `WT_ARG`, so only the logical spelling is withdrawn.

Concrete sequence: an orca agy ship spawn registers trust at bin/fm-spawn.sh:2621-2626 with `AGY_TRUST_WT=$WT`, where `$WT` is the path orca reported (bin/backends/orca.sh:140) and traverses a symlink, so `WT_REAL != WT_LOGICAL` and two entries are written. The spawn then aborts with `ORCA_ABORT_CLEANUP=1`; `fm_backend_remove_worktree orca` at bin/fm-spawn.sh:836 SUCCEEDS, so no orca recovery record is published and the directory is deleted. Execution reaches line 891, `$STATE/$ID.meta` is correctly absent, and `--remove &#34;$AGY_TRUST_WT&#34;` withdraws only the literal spelling. `WT_REAL` stays in the operator&#39;s `~/.gemini/antigravity-cli/settings.json` with no task record any supported command could key off - the accumulation docs/configuration.md:337 states cannot happen. Before this fix round the block sat at line 819, ahead of the orca removal, where the directory still resolved and both spellings were withdrawn.

Both constraints cannot be met by moving one call: the meta guard needs the state AFTER the rollback (line 889), and the rollback runs after the orca removal. The honest remedy is to stop depending on the directory still existing - capture the resolved spelling the registration already prints (`trusted: $WT_REAL`, bin/fm-agy-trust.sh:335) into a second spawn variable beside `AGY_TRUST_WT`, and withdraw both at line 891. That adds new state to machinery a prior fix round introduced rather than correcting what it already does, so it needs authorization rather than an automated edit. Note the narrow trigger: with no symlink in the orca worktree path both spellings coincide and nothing leaks.

🔧 Fix: classify agy panes, withdraw both trust spellings
1 warning still open:

- ⚠️ `bin/fm-teardown.sh:2791` - `retire_agy_workspace_trust &#34;$WT&#34;` runs on EVERY teardown regardless of the task&#39;s harness, and `--remove` runs no scope test by design, so a teardown can silently delete a `trustedWorkspaces` entry the operator registered themselves for a path firstmate never registered.

Concrete reachable path. For a secondmate, `bin/fm-spawn.sh:1822` sets `WT=&#34;$PROJ_ABS&#34;` — the firstmate home. agy can never be a secondmate: `bin/fm-spawn.sh:1438` refuses it and `fm_control_harness_supports_kind agy secondmate` returns 1, so no spawn ever registers trust for that path. But line 2791 sits above the `[ &#34;$KIND&#34; != secondmate ]` guards at 2793 and 2818, so it still fires. If the captain has hand-trusted their firstmate home in agy (the live store on this box carries exactly this shape of operator entry — `~/antigravity`, `/opt/ra`), tearing the secondmate down drops it, and the next hand-run `agy` there parks on the workspace-trust dialog that this whole feature exists to remove. The same shape reaches a crewmate: a treehouse pool worktree is a stable path reused across tasks, so tearing down a *claude* task there revokes an entry the operator made for that directory. `cleanup_firstmate_home_children` at line 2547 has the identical property for `$child_wt`.

It fails quietly in both directions — the removal writes nothing when the path is absent, and reports nothing when it does write (`&gt;/dev/null || true`), so neither outcome is distinguishable from the other.

The harness-blindness is deliberate and load-bearing: a relaunch from agy to another harness leaves `harness=claude` in meta while the trust entry is still registered, so gating on the recorded harness would strand it. The honest remedy is therefore not to gate on harness but to record, per task, that THIS spawn registered trust — new durable task state beyond what the change already does. That is why this needs the author&#39;s decision rather than an automated edit: the remedy, not the defect, is what needs authorization. A cheaper partial step the author may prefer is skipping the call when `KIND = secondmate`, which is provably always collateral.

🔧 Fix: withdraw only agy trust this task registered
2 infos still open:

- ℹ️ `bin/fm-spawn.sh:2652` - The new state/&lt;id&gt;.agy-trust record is written unconditionally, but the registration it records is idempotent and writes nothing when the path is ALREADY in trustedWorkspaces (bin/fm-agy-trust.sh:273 `if (!trusted.includes(worktree)) trusted.push(worktree)`, proven by test_registration_is_idempotent). The report line is `trusted: &lt;path&gt;` either way (bin/fm-agy-trust.sh:335), so the spawn cannot tell &#34;I added this&#34; from &#34;it was already there&#34; and records it as this task&#39;s own registration regardless.

Concrete sequence: the operator runs `agy` by hand in a treehouse pool worktree W and answers its trust dialog, so W is in their store. Later a firstmate agy ship spawn lands on that same reused pool path; fm-agy-trust.sh finds W present, adds nothing, prints `trusted: W`, and fm-spawn.sh:2652 writes W into state/&lt;id&gt;.agy-trust. At teardown, retire_agy_workspace_trust (bin/fm-teardown.sh:1004-1008) reads the record and calls `--remove W`, deleting the operator&#39;s own entry. Their next hand-run `agy` there parks on the workspace-trust dialog that draws no status text - exactly the outcome the fm-teardown.sh:987 comment names as the reason the record exists (&#34;a pool worktree is reused across tasks and harnesses&#34;).

This is a much narrower residual than the path-keyed withdrawal it replaced: it now requires the operator to have hand-trusted the exact path a firstmate agy task later runs in, rather than any shared path with any harness. docs/configuration.md:338 is honest about the narrower claim it makes (&#34;a task with no record touches nothing&#34;), so nothing documented is false.

The smallest honest remedy is not a correction to what the change already does: it would have fm-agy-trust.sh report which spellings it actually ADDED (a new output contract on the registration report) and have the spawn record only those. That extends the fix-round machinery rather than fixing it, so it needs the author&#39;s call.
- ℹ️ `AGENTS.md:99` - AGENTS.md carries the agent-facing inventory of per-task files under `state/`, and this change added `&lt;id&gt;.agy-turnend-token` to it but not `&lt;id&gt;.agy-trust`, which the last fix round introduced as a new durable per-task state file (written at bin/fm-spawn.sh:2652, read at bin/fm-teardown.sh:1002, removed at bin/fm-teardown.sh:2956 and bin/fm-teardown.sh:2607).

Every sibling per-task state file in that block is listed - `&lt;id&gt;.grok-turnend-token`, `&lt;id&gt;.kimi-turnend-token`, `&lt;id&gt;.muse-session`, `&lt;id&gt;.cursor-session`, `&lt;id&gt;.reconcile-nudged` - so an agent reading the contract sees an undocumented file appear in `state/` with no owner named. No test enforces this inventory, so nothing will catch it downstream. Add one line alongside line 99: `&lt;id&gt;.agy-trust   worktree paths this task&#39;s agy spawn registered in agy&#39;s trustedWorkspaces; withdrawn and removed by teardown`.

🔧 Fix: record only agy trust the registration added
1 warning still open:

- ⚠️ `bin/fm-agy-trust.sh:176` - `--remove` still re-derives BOTH spellings of the argument, which undoes the guarantee the last fix round built: the spawn now records only the spellings its registration actually ADDED (bin/fm-spawn.sh:2645, from the new `added:` report at bin/fm-agy-trust.sh:326), but the removal expands whatever single spelling it is handed back into the resolved/logical pair at lines 176-177 and filters both out of `trustedWorkspaces` (line 277). So a withdrawal keyed on one recorded path can delete a second entry this task never registered.

Concrete reachable sequence, reproduced against the script on this box:
- The operator has hand-trusted the RESOLVED worktree path `/…/real/wt` (their own entry; the store starts as `{&#34;trustedWorkspaces&#34;:[&#34;/…/real/wt&#34;]}`).
- A firstmate agy spawn launches into the same directory by its symlinked spelling — `$WT` reaches bin/fm-spawn.sh:2629 unresolved, which is exactly why the dual-spelling handling exists. Registration finds `/…/real/wt` already present, adds only `/…/link/wt`, and reports `added: /…/link/wt`. `state/&lt;id&gt;.agy-trust` therefore holds one line: the link spelling. Correct so far.
- Teardown calls `fm-agy-trust.sh --remove /…/link/wt` (bin/fm-teardown.sh:1006), deliberately BEFORE the worktree is released, so the directory still resolves. Line 120 sets `WT_REAL=/…/real/wt`, line 176 sets `WT_LOGICAL=/…/link/wt`, and node receives both. Observed result: the store goes from `[&#34;/…/real/wt&#34;]` to `[]` — the operator&#39;s own entry is gone.

It fails silently in both directions: the removal reports only `untrusted: &lt;resolved&gt;` and teardown discards even that (`&gt;/dev/null || true`), so nothing distinguishes withdrawing one entry from withdrawing two. The operator&#39;s next hand-run `agy` in that directory parks on the workspace-trust dialog that draws no status text — the exact outcome bin/fm-teardown.sh:987 names as the reason the record exists. bin/fm-spawn.sh:900 has the same property on the abort path whenever the worktree still exists.

The safe direction holds (operator trusted the LINK spelling, task launches at the link spelling: registration adds only the resolved one, and `--remove &lt;resolved&gt;` cannot re-expand because a resolved path has no symlink components), so the trigger is narrow: it needs a symlinked worktree path AND a pre-existing operator entry on the resolved spelling.

The smallest honest remedy is to stop expanding in remove mode and withdraw exactly the literal argument — both callers now pass an exact spelling the registration reported adding, so the pair-expansion has no remaining job. That is a deletion, not new machinery, but it contradicts a deliberately documented contract (&#34;The same pair is retired&#34;, lines 171-175) and retires `test_removal_withdraws_both_spellings_and_keeps_the_rest`, and it sits inside machinery three prior fix rounds authored rather than in the author&#39;s original change. That is why this needs the author&#39;s decision rather than an automated edit — the remedy, not the defect, is what needs authorization. If the author prefers not to touch the removal contract, the alternative is to have the record carry the pair only when the registration added the pair.

🔧 Fix: withdraw exactly the agy trust spelling named
3 warnings still open:

- ⚠️ `CLAUDE.md:4` - The last fix round (d841aab, &#34;withdraw exactly the agy trust spelling named&#34;) committed two files that have nothing to do with the finding it was fixing, or with agy at all: a `CODE-INTEL-ROUTING` block appended to `CLAUDE.md`, and a new `.codegraph/.gitignore`.

`CLAUDE.md` line 1 is `&lt;!-- Points Claude at AGENTS.md via import; edit AGENTS.md, not this file. --&gt;` - the file states its own contract that it holds nothing but the import, and the fix round edited it anyway. The added block also asserts a machine-local fact as repo-wide guidance (&#34;## Code-intelligence routing (this repo is indexed)&#34;, &#34;Reindex is automatic (CodeGraph auto-syncs on save)&#34;): it is true only on a box where CodeGraph and codebase-memory-mcp are installed, registered in `.mcp.json`, and have indexed this checkout. Every other contributor&#39;s agent reads it as a standing instruction to route questions at MCP organs it does not have. `.codegraph/.gitignore` is the same class - it commits a local index directory&#39;s ignore rules into a repo that never carried one.

Neither file appears in the author&#39;s original commit (bdb58dc), and neither is referenced by any other file in the change. This is pipeline-authored scope creep beyond what the finding required, so the remedy is to revert that fix round to its minimal part - the `bin/fm-agy-trust.sh` remove-mode change and its two tests - and drop both files. That is the author&#39;s call rather than an automated edit, because keeping a repo-level CLAUDE.md routing block is a project-policy decision.
- ⚠️ `tests/fm-agy-signals-live-e2e.test.sh:149` - The last fix round changed `--remove` to withdraw exactly the spelling it is named (bin/fm-agy-trust.sh:184-189), and updated the two production callers to withdraw per recorded `added:` spelling (bin/fm-spawn.sh:899, bin/fm-teardown.sh:1006). This live guard was not updated: it still issues one `--remove &#34;$WT&#34;` for a registration that can add two spellings.

Concrete sequence, on any box whose TMPDIR resolves through a symlink - macOS default, where $TMPDIR is `/var/folders/.../T` and `/var` is a symlink to `/private/var`:
- Line 62 builds LAB with `mktemp -d &#34;${TMPDIR:-/tmp}/fm-agy-signals.XXXXXX&#34;` and does NOT resolve it (unlike tests/lib.sh:104, which resolves with `pwd -P`), so `$WT` = `/var/folders/.../fm-agy-signals.XXXX/wt`.
- Line 108 registers it. `WT_REAL` resolves to `/private/var/folders/.../wt` and `WT_LOGICAL` stays `/var/folders/.../wt`, so the store gains BOTH entries.
- Line 149 calls `--remove &#34;$WT&#34;`. Remove mode now takes the absolute argument literally (bin/fm-agy-trust.sh:186) and filters only that one string, so `/private/var/folders/.../wt` survives.
- Line 150 is `grep -Fq &#34;$WT&#34; &#34;$STORE&#34; &amp;&amp; fail`. The surviving entry `/private/var/folders/.../wt` CONTAINS `$WT` as a substring, so the grep matches and the guard fails its own final assertion, &#34;the guard leaves the operator&#39;s store and hooks exactly as it found them&#34; - a false failure that names the wrong cause, on the one test whose job is to refresh agy evidence after an agy self-update.

On Linux with a real /tmp the two spellings coincide, the Set in bin/fm-agy-trust.sh:251 dedups them to one, and the guard passes - which is why this survived the fix round.

The remedy is a correction, not an extension: capture the registration&#39;s `added:` lines at line 108 and issue one `--remove` per line, which is exactly what bin/fm-spawn.sh:2645 and bin/fm-teardown.sh:1004 already do.
- ⚠️ `.agents/skills/harness-adapters/references/harness/agy.md:20` - agy is the first adapter added whose composer shape is not in the shape catalogue, and the resulting backend boundary is not recorded anywhere - unlike the identical cursor case, which is.

The reference records agy&#39;s composer as &#34;Bare `&gt;` prompt inside horizontal rules; no bordered box&#34;. `&gt;` is declared in `FM_COMPOSER_SHELL_PROMPT_GLYPHS` (bin/fm-composer-lib.sh:375), and THE SAFETY RULE (bin/fm-composer-lib.sh:71-75) is explicit that a bare shell glyph outside a bordered container is a dead-shell prompt that classifies `unknown`, never `empty`. The only other shape that could match - pi&#39;s `separated` region between two rules - resolves through `_fm_composer_classify_bare_pi_overlap`, which falls back to `_fm_composer_classify_bare_row` for any identity that is not pi. So an agy pane&#39;s composer verdict is ALWAYS `unknown`.

What that costs, traced per backend:
- tmux: rescued. `fm_tmux_submit_enter_core` (bin/fm-tmux-lib.sh:250-263) converts `unknown` to `empty` when the pane was idle before typing and turns busy after Enter, and `fm_pane_is_busy` is called with NO harness argument, so it uses `FM_DELIVERY_BUSY_REGEX_DEFAULT` - which this change extended with `esc to cancel` (bin/fm-composer-lib.sh:315). agy&#39;s `? for shortcuts` reads idle and `esc to cancel` reads busy, so the transition confirms.
- cmux, orca, zellij: not rescued. `fm_composer_submit_retry_core` (bin/fm-composer-lib.sh:1317) returns any non-pending verdict immediately, so `unknown` comes straight back, and bin/fm-send.sh:1028 treats anything other than exact `empty` as `error: text not submitted ... (delivery unconfirmed)` and exits non-zero. Every steer to an agy crewmate on those three backends reports unconfirmed even when it landed.

This does not block the stated intent - the brief rides the launch command (`-i &#34;&lt;prompt&gt;&#34;`), not fm-send - but it does bound later supervision, and the change ships no note of it. The precedent is right there: docs/configuration.md:301 records &#34;Cursor typed-submit confirmation is verified on tmux and Herdr only&#34; with a pointer to the evidence, for a harness whose composer verdict is `unknown` for the same structural reason.

The remedy is the author&#39;s call rather than an automated edit, because it is a claim about verified scope: either record agy&#39;s backend boundary alongside cursor&#39;s, or register agy&#39;s composer shape in the catalogue so the verdict is provable on every backend. The second is a real extension to the shape classifier, not a correction.

🔧 Fix: drop tooling creep, fix live guard, record agy submit boundary
2 infos still open:

- ℹ️ `bin/fm-teardown.sh:1006` - `retire_agy_workspace_trust` swallows a refused withdrawal with `|| true` and then unconditionally `rm -f`s the record on line 1009, so a withdrawal that fails leaves the trust entry in the operator&#39;s vendor settings with nothing left that can key its removal.

Concrete sequence: a task ran on agy, so `state/&lt;id&gt;.agy-trust` names `/…/wt`. Teardown calls `fm-agy-trust.sh --remove /…/wt` while the operator&#39;s own agy session writes `~/.gemini/antigravity-cli/settings.json`. The node writer fingerprints the bytes before its rename (bin/fm-agy-trust.sh:317); a store that moved returns `moved`, and on the second such attempt it refuses with &#34;was modified while trust was being recorded&#34; (bin/fm-agy-trust.sh:339) rather than clobbering the vendor&#39;s write. Teardown discards that exit status, deletes the record, and line 2957 would delete it anyway. The entry survives; `state/&lt;id&gt;` is gone; nothing can name it again. That is exactly the accumulating dead-path leak the `--remove` mode exists to prevent (bin/fm-agy-trust.sh:18-20).

The inline justification - &#34;a retry would face the same store&#34; - holds for the permanent refusals (not owned, not writable, non-array `trustedWorkspaces`) but not for the concurrent-modification refusal, which is the one failure the trust script itself models as transient by retrying three times.

Impact is small: the refusal message still reaches teardown&#39;s stderr, so it is not silent, and the window is three consecutive fingerprint checks. I am flagging it as informational rather than as a blocker because the smallest durable remedy - retaining the record for a later sweep - needs a sweeper that does not exist, which is new machinery beyond this change&#39;s stated intent. The in-scope alternative is to keep the failure loud and actionable (name the stranded path and the `fm-agy-trust.sh --remove &lt;path&gt;` a captain can run by hand) instead of `|| true`. Either way it is a call about the author&#39;s deliberate best-effort tradeoff, which the comment on lines 995-997 already states.
- ℹ️ `tests/fm-agy-signals-live-e2e.test.sh:48` - The live guard deliberately operates on the operator&#39;s REAL agy store (line 76 explains why: a separate HOME lands on an unauthenticated agy). It unwinds by `cp`ing a snapshot back over that store - in the EXIT trap on line 48 and on the success path on line 164.

A plain `cp` of a snapshot is the one thing the code this guard exists to prove explicitly refuses to do. `bin/fm-agy-trust.sh` fingerprints the bytes it read and refuses rather than overwrite when the store moved underneath it (bin/fm-agy-trust.sh:222-244, 317), because losing the vendor&#39;s write means losing a telemetry choice or another workspace&#39;s trust in a format firstmate does not own.

Concrete sequence: the guard runs for roughly three to four minutes (three 60s status-bar waits plus a 120s answer wait). The snapshot is taken at line 77. If the operator&#39;s own agy session trusts a folder or changes a setting during that window, line 164 reverts it with no signal. The withdrawal loop added on lines 156-162 already restores the store to the correct state on its own; the `restore_store` that follows it is what discards the concurrent write.

Low impact and opt-in - the operator runs this deliberately after an agy upgrade, and the file states up front that it writes their real store. Raising it because the guard&#39;s own teardown contradicts the invariant it is guarding, which is worth an explicit decision rather than an automated edit.

🔧 Fix: name unwithdrawn agy trust, unwind live guard precisely
2 warnings still open:

- ⚠️ `tests/fm-secondmate-harness.test.sh:60` - `ANTIGRAVITY_CONVERSATION_ID` is now the HIGHEST-precedence marker in `bin/fm-harness.sh:detect_own` (bin/fm-harness.sh:63, tested before `CLAUDECODE`), but only the new `tests/fm-agy-harness.test.sh:19` was given the corresponding `unset`. Three existing suites scrub the old marker set and were not extended:

- tests/fm-secondmate-harness.test.sh:60
- tests/fm-kimi-harness.test.sh:12
- tests/fm-muse-harness.test.sh:21

All three carry the same header comment explaining exactly why the scrub exists (&#34;so the asserted verdict does not depend on which harness launched the suite&#34;), and this change adds a new member of that class without updating them.

Concrete failing sequence: run `bin/fm-test-run.sh tests/fm-secondmate-harness.test.sh` from an agy pane — which this change makes a supported crewmate/scout runtime. `ANTIGRAVITY_CONVERSATION_ID` is inherited (the reference at .agents/skills/harness-adapters/references/harness/agy.md:24 records that agy exports it to every tool subprocess). tests/fm-secondmate-harness.test.sh:86 runs `CLAUDECODE=1 FM_CONFIG_OVERRIDE=... fm-harness.sh secondmate`; `detect_own` now returns `agy` before it ever reads `CLAUDECODE`, so the `own` fallback resolves to `agy` instead of `claude` and case A1 fails. The same applies to tests/fm-kimi-harness.test.sh:551-554 and tests/fm-muse-harness.test.sh:188.

Before this change the marker did not exist in `detect_own`, so those suites were hermetic; after it they are not. The remedy is mechanical and non-behavioral: add `ANTIGRAVITY_CONVERSATION_ID` to the three `unset` lines, exactly as tests/fm-agy-harness.test.sh:19 already does.
- ⚠️ `docs/verification/runtime-backends.md:1088` - The new agy submit-boundary paragraph enumerates only three backends as reporting unconfirmed delivery and then calls it &#34;the same boundary Cursor has, for the same structural reason.&#34; Both halves are wrong about Herdr, and Herdr is the one backend a reader is most likely to act on.

Source evidence that Herdr does NOT rescue an agy submit:
- agy&#39;s composer verdict is always `unknown`, never `pending` — pinned by tests/fm-agy-harness.test.sh:776 and by the bare-`&gt;`/`FM_COMPOSER_SHELL_PROMPT_GLYPHS` safety rule (bin/fm-composer-lib.sh:375).
- Herdr&#39;s rendered-footer rescue is gated on `pending` only: `if [ &#34;$verdict&#34; = pending ] &amp;&amp; ...` (bin/backends/herdr.sh:2836). An `unknown` verdict falls straight through to `unknown) printf &#39;unknown&#39;; return 0` (bin/backends/herdr.sh:2844).
- The idle-native branch is no better: a native-idle pane falls to `fm_backend_herdr_composer_state`, and `*) printf &#39;%s&#39; &#34;$verdict&#34;` (bin/backends/herdr.sh:2828) returns `unknown` as-is.

So an agy typed send on Herdr lands but `bin/fm-send.sh` reports `delivery unconfirmed` and exits non-zero — identical to Zellij/cmux/Orca. Cursor&#39;s boundary is genuinely different: docs/verification/runtime-backends.md:969-970 and docs/configuration.md:301 both record Cursor as &#34;verified on tmux and Herdr only&#34;, because Cursor&#39;s verdict IS `pending` and therefore is rescued by the line above.

Failure this produces: a firstmate reading line 1088 sees Herdr absent from the unconfirmed list and an explicit &#34;same boundary Cursor has&#34; pointer at a doc that names Herdr as working, and dispatches an agy crewmate on Herdr expecting confirmed steers. Every `fm-send` typed steer to that worker then exits non-zero on a message that actually landed, with no doc explaining it. docs/configuration.md:302 states the correct scope (&#34;verified on tmux only&#34;) but attributes it to &#34;only tmux&#39;s submit core resolves that through the busy footer&#34;, which Herdr&#39;s footer path at bin/backends/herdr.sh:2836 also does — just not for this verdict.

The remedy is a doc correction, not new machinery: name Herdr alongside Zellij, cmux, and Orca, and either drop the Cursor analogy or state the actual difference (Cursor&#39;s `pending` is rescued on Herdr; agy&#39;s `unknown` is not).

🔧 Fix: scrub agy marker in sibling suites, fix stale launch pins
2 infos still open:

- ℹ️ `AGENTS.md:99` - The same hunk that added `&lt;id&gt;.agy-turnend-token` to the `state/` inventory did not add `&lt;id&gt;.agy-trust`, and that record is the more consequential of the two.

`&lt;id&gt;.agy-trust` is the ONLY per-task state file that keys the undo of a mutation written OUTSIDE the firstmate home: `bin/fm-spawn.sh:2645` appends the spellings the registration reported adding to it, and `bin/fm-teardown.sh:1006` (`retire_agy_workspace_trust`) withdraws exactly those from the operator&#39;s `~/.gemini/antigravity-cli/settings.json`. `bin/fm-teardown.sh:995-1000` states the invariant plainly - &#34;No record means this task registered nothing and nothing is touched.&#34;

Concrete consequence: an agent or captain recovering a wedged home by clearing `state/&lt;id&gt;.*` by hand has AGENTS.md as its inventory of what each file means. `&lt;id&gt;.agy-trust` reads as an undocumented stray; deleting it strands an absolute worktree path in a vendor settings file firstmate does not own, with no supported command left that can name it - the exact accumulating leak `--remove` exists to prevent (`bin/fm-agy-trust.sh:18-20`).

The inventory is already non-exhaustive (`&lt;id&gt;.muse-session-current` and `&lt;id&gt;.control-relaunch` are absent too), so this is not a broken contract - which is why it is informational rather than a blocker. The remedy is one line beside the sibling this change already added.
- ℹ️ `tests/fm-agy-harness.test.sh:785` - `fm_composer_classify_screen &#34;&#34; &#34;$screen&#34; &#34;&#34; agy` passes `agy` as the function&#39;s FOURTH parameter, which is `identity`, not a harness selector: `bin/fm-composer-lib.sh:1199-1200` declares `# &lt;caps&gt; &lt;screen&gt; [cursor_row] [identity]`, and `identity` is read only by `_fm_composer_pi_verdict` / `_fm_composer_classify_bare_pi_overlap`, both gated on `has_identity` being set from the `caps` argument - which is empty here. The value is therefore never read; the same call with `claude`, `bogus`, or nothing at all returns `unknown` identically.

The assertion itself is correct and behavior-based - the `unknown` verdict comes from the real bare-`&gt;` shell-glyph safety rule (`bin/fm-composer-lib.sh:375` `FM_COMPOSER_SHELL_PROMPT_GLYPHS`) - so this is not a test that can pass with the code wrong. The problem is what it implies. The test is named `test_agy_composer_verdict_is_unknown_not_empty` and the identical snippet is published as reproducible evidence at docs/verification/runtime-backends.md:1080 under &#34;an agy pane&#39;s composer verdict is always `unknown`&#34;. Both read as though the classifier has an agy-specific branch.

Failure this sets up: a maintainer refactoring `fm_composer_classify_screen` sees a harness name threaded in and preserves a per-harness path that does not exist, or conversely assumes agy&#39;s verdict is pinned by harness dispatch and stops checking that the bare-glyph rule still covers `&gt;`. The verdict this whole tmux-only submit boundary rests on is owned by that one shape rule and nothing else.

Remedy is mechanical and non-functional: drop the inert argument in both places (or leave it and name in a comment that it is the pi `identity` slot and is ignored without `caps`), so the evidence says what actually decides the verdict.

</details>

<details>

</details>

<details>

- ℹ️ `tests/fm-harness-liveness-drift-live-e2e.test.sh:103` - docs/tmux-backend.md now records agy among the process identities the tmux liveness probe classifies `alive`, and states directly beneath that roster: &#34;Run the real-harness guard after any harness upgrade and before trusting refreshed evidence.&#34; For agy that instruction currently resolves to nothing. The opt-in drift guard iterates a hard-coded list — `for harness in claude codex opencode pi pi-signed grok kimi cursor muse` (tests/fm-harness-liveness-drift-live-e2e.test.sh:103) — and agy is not in it, so running the guard after an agy self-update silently exercises no agy identity and still reports a pass.

That matters more for agy than for most entries, because agy is the adapter this change documents as self-updating without being asked (&#34;Verified on 2026-09-03 with Antigravity CLI 1.1.24, which self-updated to 1.1.25 mid-investigation&#34;), and its classification depends on an exact `agy` process-name match with no versioned or path-component fallback (bin/backends/tmux.sh). A vendor rename to a version-suffixed or wrapper name would drop a healthy agy crewmate to `ambiguous` — every control verb then refuses to touch it — with no guard failing to say so.

I did not fix this because the remedy is a test change, which is outside this phase&#39;s edit scope, and because it is the author&#39;s call whether agy&#39;s drift coverage belongs in the shared liveness guard or stays with the new FM_AGY_SIGNALS_LIVE_E2E suite (which covers the trust dialog, status bars, typed submission, and the Stop hook, but not process-name classification). The documentation I wrote does not overclaim: it describes the classifier&#39;s behavior, not the guard&#39;s coverage.

🔧 Fix: record agy&#39;s liveness drift-guard coverage boundary
1 info still open:

- ℹ️ `tests/fm-harness-liveness-drift-live-e2e.test.sh:103` - Carried over from the previous document round, where it was selected to be fixed. The documentation half is now resolved: docs/tmux-backend.md records that agy sits outside the drift guard&#39;s installed-harness roster, so the &#34;run the real-harness guard after any harness upgrade&#34; instruction no longer reads as covering agy. The underlying coverage gap is unchanged and is a test edit this phase is not permitted to make.

The opt-in guard still iterates a hard-coded list — `for harness in claude codex opencode pi pi-signed grok kimi cursor muse` — so agy is the only entry in the tmux liveness roster (docs/tmux-backend.md:51) that no live run exercises. That matters more for agy than for its siblings because agy self-updates without being asked (recorded at docs/verification/runtime-backends.md:982, where 1.1.24 became 1.1.25 mid-investigation) and its classification depends on an exact `agy` process-name match with no versioned or path-component fallback (bin/backends/tmux.sh). A vendor rename to a version-suffixed or wrapper name would drop a healthy agy crewmate to `ambiguous`, at which point every control verb refuses to touch it, and the guard would still report a pass.

The remedy is one word added to that list, but it is the author&#39;s call whether agy&#39;s process-name drift coverage belongs in the shared liveness guard or stays with tests/fm-agy-signals-live-e2e.test.sh, which today covers the trust dialog, both status bars, typed submission, and the Stop hook but not process-name classification. Whichever way it lands, the tmux-backend sentence added this round should be updated to match.

</details>

<details>

🔧 Fix: fix shellcheck findings in agy harness test

</details>

<details>

✅ No issues found.

</details>


RooseveltAdvisors:
Closing this — it was opened against the wrong base by our pipeline while the work is still in progress. We're re-raising it on our own fork and will bring a clean copy here once it's complete and validated. Apologies for the noise.

PR #3717, opened 2026-09-04
fix: allow Herdr relaunch after worker exit

cam-em:
## Intent

Captain: "Yes clean it up, I want to relaunch and get this setup for antigravity over with. Use GPT." Clean up the Herdr lifecycle issue that prevents relaunch after the cancelled worker has exited, so the antigravity setup can proceed. Captain: "I have cancelled it and we should no longer use claude." Use GPT and do not use Claude.

## What Changed

- Classify a stale Herdr agent registration as stopped when process evidence proves the pane has returned to its idle shell topology, allowing safe worker relaunch.
- Preserve fail-closed behavior for active, branching, malformed, or unreadable process states while sharing the stricter classifier across lifecycle and cleanup paths.
- Add portable classifier coverage, a GPT-only live Pi exit-and-relaunch guard, and updated Herdr verification documentation.

## Risk Assessment

✅ Low: The change is narrowly scoped, preserves fail-closed behavior for malformed or competing process topologies, and adds behavioral coverage for the durable stale-registration fix and live-agent counterexample.

## Testing

- Outcome: ⚠️ 1 error across 1 run (12m13s)

## Pipeline

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>

</details>

<details>

✅ No issues found.
</details>

<details>

</details>

<details>

✅ No issues found.
</details>

### DIFF
--- a/bin/backends/herdr.sh
+++ b/bin/backends/herdr.sh
@@ -1207,9 +1207,8 @@ fm_backend_herdr_pane_idle_shell_pid() {  # <session> <pane-id>
 # It prints the foreground shell pid. With <root-only>=1, that shell must also
 # be Herdr's pane root shell; presentation cleanup depends on that stronger
 # identity because it signals the returned pid to remove the pane. With
-# <root-only>=0, the root-to-foreground subtree may be one unbranched wrapper
-# chain (the normal spawn shape is root shell -> treehouse -> task shell), but
-# no sibling or descendant process may survive anywhere under the pane root.
+# <root-only>=0, the only accepted subtree is root shell -> treehouse -> task
+# shell, with no sibling or descendant process anywhere under the pane root.
 fm_backend_herdr_idle_shell_from_process_info() {  # <json> <pane-id> <root-only>
   local info=$1 pane=$2 root_only=$3 shell_pid foreground_pgid count
   local process_pid name argv0 shell_name rows stat ps_bin
@@ -1248,11 +1247,12 @@ fm_backend_herdr_idle_shell_from_process_info() {  # <json> <pane-id> <root-only
     fm_backend_herdr_pid_is_bare_shell "$ps_bin" "$shell_pid" || return 1
     fm_backend_herdr_pid_is_bare_shell "$ps_bin" "$process_pid" || return 1
   fi
-  rows=$(LC_ALL=C "$ps_bin" -axo pid=,ppid= 2>/dev/null) || return 1
+  rows=$(LC_ALL=C "$ps_bin" -axo pid=,ppid=,comm= 2>/dev/null) || return 1
   printf '%s\n' "$rows" | awk -v root="$shell_pid" -v foreground="$process_pid" '
     {
       pid[NR] = $1
       parent[$1] = $2
+      command[$1] = $3
       present[$1]++
     }
     END {
@@ -1264,6 +1264,12 @@ fm_backend_herdr_idle_shell_from_process_info() {  # <json> <pane-id> <root-only
         current = parent[current]
       }
       path[root] = 1
+      if (foreground != root) {
+        wrapper = parent[foreground]
+        if (wrapper == root || parent[wrapper] != root) exit 1
+        sub(/^.*\//, "", command[wrapper])
+        if (command[wrapper] != "treehouse") exit 1
+      }
       for (i = 1; i <= NR; i++) {
         current = pid[i]
         walk++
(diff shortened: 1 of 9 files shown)


tiago-peixoto (commented):
Thanks for this fix.
I reviewed head 2552938dc58d4d3eddfa8b9a5043c18786e9ace4 and ran its classifier functions, byte-identical to this head, against a real Herdr 0.9.0 session with real `treehouse get` panes on macOS (Pi 0.85.1, treehouse v2.3.0).
It fixes the stale registration from #4115: after Pi `/quit`s in a pooled pane, the pane reads `no-agent` / `dead` where main reads `alive`, and a live agent under the launch wrapper still reads `alive`.
I found one case where it reads a live agent as stopped.

## A live agent whose launch wrapper exits reads as stopped

`fm_backend_herdr_idle_shell_from_process_info` proves the pane idle by walking only the pane root shell's descendants (`bin/backends/herdr.sh` lines 1273-1283 at this head).
`fm-spawn` starts the agent under a `/bin/sh -c '...'` wrapper.
If that wrapper exits while the agent keeps running, the agent is re-parented to pid 1 and leaves the root's subtree.
The task shell takes the foreground back, process-info then lists only that idle shell, and the proof passes.
So `fm_backend_herdr_pane_agent_state` returns `no-agent` and the recovery read returns `dead`, while the agent is still running on the pane's terminal and its registration still stands.

To reproduce:

1. In a Herdr pane, run `treehouse get`, then start Pi the way `fm-spawn` does: `/bin/sh -c 'unset TRACEPARENT; pi'`.
   With the registration present, the pane reads `active` / `live` / `alive`.
   The scratch Pi here had no credentials, so the registration was reported with `herdr pane report-agent --agent pi --state idle`.
2. Send SIGTERM to the wrapper only: `kill -TERM <foreground_process_group_id>`.
3. Pi keeps running, re-parented to pid 1 and still attached to the pane's terminal, and the classifier now reads `idle-shell` / `no-agent` / `dead`.

```text
before, pane root 6389 (pid ppid pgid stat comm):
   6389 92964  6389 Ss   -zsh
   8689  6389  8689 S    treehouse
   8902  8689  8902 S    /bin/zsh
  13267  8902 13267 S+   /bin/sh
  13279 13267 13267 S+   pi
classifier: active / live / alive

after kill -TERM 13267 (pid ppid pgid stat tty comm):
  13279     1 13267 S    ttys034  pi
classifier: idle-shell / no-agent / dead
Pi was still alive 5 seconds later.
```

The same happens with a busy stand-in agent (`python3 -c 'import time; time.sleep(900)'`) and in a root-only pane with no `treehouse` level.

Relaunch and the spawn precondition treat `dead` as safe to start a new agent, so a second agent can start in the same copy while the first is still running.
The orphaned agent can no longer read input from the pane, but one that was mid-turn keeps running tool calls in a copy that may hold unlanded work until it next reads.
The trigger is external, a wrapper killed from outside or one that crashes, so it should be rare.
On current main, which trusts the registration alone, this pane reads `alive`.

## A check that separates the two cases

After the idle proof passes, also require every process whose controlling terminal is the pane root's terminal to lie on the proven path, and read the pane as live otherwise.
Herdr's `process_info.tty` came back `null` on macOS here, and its socket API docs do not say when it is populated, so the terminal has to come from `ps -o tty= -p <shell_pid>`.
Its processes then come from `ps -t <tty>`; macOS `ps(1)` documents `-t` as "Display information about processes attached to the specified terminal devices".
In the same session, the idle nested pane lists exactly the three processes on the proven path, and the orphan case adds the orphan:

```text
nested pane, idle (ps -t ttys021):
   3422  3033  3422 Ss   -zsh
  12368  3422 12368 S    treehouse
  12863 12368 12863 S+   /bin/zsh

same pane after a launch wrapper was killed (ps -t ttys021):
   3422  3033  3422 Ss   -zsh
  12368  3422 12368 S    treehouse
  12863 12368 12863 S+   /bin/zsh
  17563     1 17557 S    python3
```

The trade-off is that a helper an agent detached onto the terminal, for example `(sleep 600 &)`, which this head reads as stopped, would then read as live.
That errs toward refusing a relaunch rather than starting a second agent.

We carry this PR on our fork, and I can follow up here with a tested patch for this check once we have one.
The PR currently conflicts with main, so the line numbers above refer to 2552938.

kunchenguid:
Closing as superseded by #4191 (merged 2026-09-11), which already handles the herdr stale-registration recovery contract via the narrow process-level liveness proof. Per the repo's one-owner rule we avoid landing a second implementation of the same contract, so this is closed in favor of the landed fix. Thanks for the work here.

PR #3723, opened 2026-09-04
fix(bin): refuse teardown when the recorded worktree is not a Treehouse pool slot, and guard Claude hook removal

MLA82:
## Summary

This PR closes two teardown safety gaps:

1. **Prove a recorded worktree before reaping it.** `bin/fm-teardown.sh` used
   to trust a task's recorded `worktree=` (and, for Orca, its recorded
   worktree id) without proof. If that field was ever stale or wrong -
   pointing at the wrong directory, or even at the primary firstmate
   checkout - teardown could kill live processes and hard-reset the wrong
   worktree before any later check caught the mistake.

2. **Guard Claude hook removal.** `remove_claude_hook_file` deleted a
   worktree's `.claude/settings.local.json` without checking whether that
   worktree was still in use, and treated a missing `treehouse` binary the
   same as a confirmed "not in use" - silently removing a live hook.

## Behavior change

Teardown now refuses, before touching a single process or file, whenever a
task's recorded worktree cannot be proven to be a real Treehouse pool slot
(same project's Git common directory, present pool state). The refusal is
non-mutating - the task record, backlog row, and worktree are all left
exactly as found - and explains itself the same way every other refusal in
this file does. This is a deliberate new restriction: a mis-recorded
`worktree=` or `project=` must never be reaped, only reported.

Claude hook removal now checks the worktree's actual in-use state before
deleting the hook file, and treats an inconclusive check (including a
missing `treehouse` binary) as "do not remove" rather than "safe to remove".

## Testing

Verified locally, including under a `PATH` with `treehouse` and `herdr`
removed to simulate a runner without them installed:

- `tests/fm-teardown.test.sh` - the full teardown safety suite, including the
  Treehouse pool-slot proof, the Claude hook removal guard (in-use, not
  in-use, inconclusive, and missing-binary cases), and the forced
  secondmate-teardown fixtures.
- `tests/fm-backlog-atomicity.test.sh` - the backlog/record pairing
  invariant, including interrupted-cleanup recovery.
- `tests/fm-teardown-endpoint-safety.test.sh` and
  `tests/fm-captain-hold-lifecycle.test.sh` - adjacent teardown-related
  suites, unaffected by this change.

All four suites pass. `shellcheck` on the changed scripts is clean (no new
findings versus the pre-change baseline).


kunchenguid:
Speaking as Kun's firstmate:

**Triage** (tip `083a7ae2618a9c712a1c64660de58d21192744d8`): fork CI approved (CI `33904191117` / NM `33904191091`). Direction is right — prove treehouse membership before reaping, and stop blind Claude hook `rm`. Attestation **MISSING**.

**Real residuals (waiting on you, not a captain hold):**
1. `remove_claude_hook_file` still fails open when `treehouse` is missing: the in-use check is skipped and the hook is removed anyway, contrary to the stated “inconclusive → do not remove” rule.
2. For `backend=orca`, Fix 0 is skipped and `reap_task_worktree_processes` still runs before Orca's own path-match proof — a stale recorded worktree can still be TERM/KILL'd first.
3. Related open hold: #3168 (teardown duplicate unvalidated reap) — do not treat this PR as clearing that card.

**Contract-class:** restore (promised teardown safety before destructive reap).

**VISION:** R1–R7 align for fail-closed teardown; Authority/refusal paths strengthen when the residuals above are closed.

MLA82:
Both confirmed real, verified against the actual code (not just the summary).

**1. Hook removal on inconclusive status.** `worktree_is_in_use` returned "not in use" both when the real path couldn't be resolved *and* when `treehouse status` itself errored (its own command-substitution failure was silently swallowed). `remove_claude_hook_file` treats "not in use" as license to delete the hook, so an inconclusive check - not just a genuine "not in use" - let it proceed and remove an active task's hook. Unknown must fail closed toward "in use" (preserve), the opposite of what a destructive path's unknown state should default to.

**2. Unconditional treehouse-membership guard blocking Orca.** Confirmed from `docs/architecture.md`: "treehouse pools clean worktrees for tmux, herdr, zellij, and cmux tasks, while Orca creates its own worktrees for `backend=orca`." The new `worktree_is_treehouse_managed` guard ran for every non-secondmate task with no backend check, so it would refuse *every* Orca teardown before ever reaching Orca's own worktree-path verification (`require_orca_worktree_path_match_if_present`) further down - Orca worktrees never appear in `treehouse status` by design. The existing Orca teardown test coverage didn't catch this because it always registers the worktree with the fake treehouse pool via `fm_fake_treehouse_pool "$FB" "$wt"`, which isn't representative of a real Orca worktree.

Both fixed in `083a7ae2`:
- `worktree_is_in_use` now returns 0 (in-use/unknown) on either failure mode, with an updated doc comment on the intended fail-closed direction.
- The membership guard now skips `backend=orca`, relying on Orca's own downstream path-match check instead.

Added two regression tests: one that makes `treehouse status` error specifically on the in-use check (while still passing the earlier membership check, via a call-counting fake) and asserts the hook survives; one that runs an Orca teardown against an *empty* treehouse pool (the realistic case) and asserts it still succeeds. Re-ran `fm-teardown`, `fm-teardown-endpoint-safety`, and `fm-backend-orca`: all pass, 0 failures, including both new tests.

MLA82:
Closing in favour of #4838, which replaces this proposal.

The change here sits on a base from 2026-09-04 and had grown to 15 files for a change of this size. #4838 is cut fresh against current `main` and comes to 3 files.

Both residuals raised on this pull request are addressed there: the first no longer reproduces on current `main`, because ownership is now established through `teardown_owns_worktree`, which does not treat an unresolved check as permission to delete; the second — reaping an Orca-owned worktree before Orca's own path-match proof — is fixed, with regression tests that fail without the fix.

PR #3725, opened 2026-09-04
fix(bin): harden lock waits, install checks, and test aggregation against silent failures

MLA82:
## What Changed

- The wake acknowledgement (`bin/fm-wake-drain.sh --ack-through`), branch-grant (`bin/fm-wake-grant.sh`), and queued-key (`fm_wake_queued_keys` in `bin/fm-wake-lib.sh`) callers now wait for the wake-queue lock through the existing `fm_lock_acquire_wait_bounded` (default 30s, `FM_WAKE_QUEUE_LOCK_WAIT`) and refuse (exit or return non-zero, with a diagnostic in the drain) instead of mutating queue state without the lock. `fm_lock_acquire_wait` is unchanged: it still waits indefinitely and cannot fail, so its other callers keep their behavior.
- `bin/backends/cmux.sh` surface-existence check now captures the CLI call's own exit status before piping to `jq -e`, avoiding a false "surface exists" result when the CLI call fails and jq 1.6 treats empty stdin as success.
- `bin/fm-install-herdr.sh` and `bin/fm-install-treehouse.sh` capture `--version` stderr to a separate file (instead of merging with stdout) so incidental stderr output can't corrupt the parsed version string, and surface captured stderr in the `die` messages on failure.
- `bin/fm-test-run.sh` aggregation now skips unreadable/truncated per-lane timing JSON artifacts (logging a warning) instead of aborting the whole aggregate step; `bin/fm-afk-return.sh`'s `print_evidence` now propagates a write failure via its return status.
- Updated `docs/watcher-continuity.md` and `docs/configuration.md` for the bounded queue-lock wait, and added/extended tests covering the cmux surface-exists fix, test-run aggregation skip behavior, and the queue-lock callers refusing when the lock stays held.

## Risk Assessment

Low: small, independent robustness fixes with behavioral test coverage for each. The lock wait keeps its existing unbounded contract; only the three caller groups above use the bounded variant, and each handles its failure explicitly.

## Testing

- Outcome: ⚠️ 1 error across 1 run (26m8s)

## Pipeline

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>

- ℹ️ `bin/fm-wake-lib.sh:932` - This branch (41a2035) changes fm_lock_acquire_wait&#39;s long-standing infinite-wait contract to a 30s default timeout that returns 1 on expiry. Only a handful of call sites (fm-wake-grant.sh, fm-wake-drain.sh, fm_wake_append, fm_wake_queued_keys) were updated in this branch to check the new return value and fail closed; roughly 80 other call sites across the codebase (e.g. fm-branch-outcome.sh, fm-captain-hold.sh, fm-teardown.sh:296, fm-promote.sh, fm-pr-check.sh, fm-startup-network.sh, fm-x-lib.sh, fm-backlog-handoff.sh) still call it bare and unconditionally set their own *_LOCK_HELD=1 afterward, so after 30s of contention they would proceed to mutate shared state believing they hold a lock they do not. This is explicitly acknowledged and scoped as intentional, proportionate containment in the ca08147 commit message and in code comments in fm-wake-lib.sh (a stderr diagnostic is the stated mitigation, with a full caller audit explicitly deferred as a separate, larger change), so this is not a new unauthorized regression from this diff - noting it for visibility only.
</details>

<details>

</details>

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

### DIFF
--- a/bin/backends/cmux.sh
+++ b/bin/backends/cmux.sh
@@ -398,8 +398,15 @@ fm_backend_cmux_parse_target() {  # <target>
 # (fm_backend_zellij_pane_exists) rather than the design sketch's original
 # read-screen-based suggestion.
 fm_backend_cmux_surface_exists() {  # <workspace_id> <surface_id>
-  local wsid=$1 sfid=$2
-  fm_backend_cmux_cli list-panes --workspace "$wsid" --json --id-format uuids 2>/dev/null \
+  local wsid=$1 sfid=$2 out
+  # Capture and check the CLI call's own exit status before piping to jq: a
+  # bare `cli ... | jq -e ...` pipe (without pipefail) takes its exit status
+  # from jq alone, and jq 1.6 (still the apt-installed default on several
+  # supported Linux distros) reports success (exit 0) for `-e` on completely
+  # empty stdin instead of jq 1.7+'s failure (exit 4) - so a failed,
+  # output-less list-panes call would misreport the surface as existing.
+  out=$(fm_backend_cmux_cli list-panes --workspace "$wsid" --json --id-format uuids 2>/dev/null) || return 1
+  printf '%s' "$out" \
     | jq -e --arg s "$sfid" '[.panes[]? | select(.surface_ids // [] | index($s))] | length > 0' >/dev/null 2>&1
 }
 
--- a/bin/fm-afk-return.sh
+++ b/bin/fm-afk-return.sh
@@ -104,11 +104,12 @@ write_gate() {  # <evidence-file> <blockers-file>
 }
 
 print_evidence() {  # <file>
-  local file=$1 kind text
+  local file=$1 kind text status=0
   while IFS="$(printf '\t')" read -r tag kind text; do
     [ "$tag" = evidence ] || continue
-    printf 'catch-up %s: %s\n' "$kind" "$text"
+    printf 'catch-up %s: %s\n' "$kind" "$text" || status=1
   done < "$file"
+  return "$status"
 }
 
 print_blockers() {  # <file>
--- a/bin/fm-install-herdr.sh
+++ b/bin/fm-install-herdr.sh
@@ -77,12 +77,15 @@ mkdir -p "$DESTINATION"
 install -m 0755 "$TMP/$ASSET" "$DESTINATION/herdr"
 
 # Post-install version and protocol gates (no floating latest).
-installed_version=$("$DESTINATION/herdr" --version 2>/dev/null | awk '{print $2; exit}')
+version_output=$("$DESTINATION/herdr" --version 2>&1) || {
+  die "'$DESTINATION/herdr --version' exited $? with output: ${version_output:-<none>}"
+}
+installed_version=$(printf '%s' "$version_output" | awk '{print $2; exit}')
 [ "$installed_version" = "$FM_HERDR_CI_VERSION" ] \
   || die "installed herdr version is '${installed_version:-<empty>}', expected exact pin $FM_HERDR_CI_VERSION"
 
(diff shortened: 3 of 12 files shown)


kunchenguid:
Speaking as Kun's firstmate:

**Triage** (tip `ca08147128878784cd51f1d0a680ca525adee4d1`): fork CI approved for this tip (CI `33904943320` / NM `33904943315`). Attestation **MISSING** (NM body-compliance fail). Greptile still right that a timeout return does not stop existing callers that ignore `fm_lock_acquire_wait`'s status — they can keep mutating after the wait fails; the new stderr diagnostic helps visibility but does not close that. Zero-timeout is fixed on this tip. Closes link to #3279 is real in the author body (bounded wait), but the unlock-without-ownership residual is still on you.

**Contract-class:** restore (bound the forever-spin #3279 promised to end) with an incomplete caller surface — treat as **waiting on author**, not merge-ready.

**VISION:** R1–R7 align for a lock/robustness fix; Authority stays explicit (no new consent assumed). Scripts-own-mechanics aligns.

**Waiting on the author** (not a captain hold): make callers honor a failed acquire (or otherwise refuse to proceed without the lock), then rebind attestation / green NM. Do not merge until that lands.

MLA82:
Looked into both parts of this against the actual code and against how the fix already runs elsewhere.

**Zero-timeout behavior: confirmed real, and fixed.** The expiry check was guarded by `[ "$timeout" -gt 0 ]`, so `fm_lock_acquire_wait lockdir 0` - documented as "try once and return immediately" - actually disabled the bound entirely and fell back to waiting forever. Fixed in `ca081471` by dropping that guard so elapsed time alone decides. Added a direct unit test (`test_lock_acquire_wait_zero_timeout_tries_once`) proving a `timeout=0` call against a live-held lock returns within ~1s instead of hanging.

**Unchecked callers: real, but scoped deliberately.** `grep -rn 'fm_lock_acquire_wait' bin/` turns up on the order of 40 call sites across the codebase, and it's true that a good number of them (`fm-lock.sh`, `fm-afk-return.sh`, `fm-branch-outcome.sh`, `fm-startup-network.sh`, several in `fm-spawn.sh`, etc.) call it bare, with no `||` check - because before this change the function *could not fail* (its own doc, and the sibling `fm_lock_acquire_wait_bounded`'s comment "mutation-critical callers that can safely block keep fm_lock_acquire_wait", made blocking-forever the documented contract). Adding a default bound necessarily changes that contract for every caller at once.

Auditing and fixing all ~40 call sites is a much larger, separate change than this PR's scope (five small, independent fixes): it would mean touching many files this PR otherwise has no reason to change, and each caller needs its own judgment call about whether blocking-forever or a bounded failure is actually correct for that lock. As an in-scope mitigation here, `fm_lock_acquire_wait` now also prints a timeout diagnostic to stderr on expiry (`fm_lock_acquire_wait: timed out after ${timeout}s waiting for $lockdir`), so a caller that doesn't check the return value still leaves visible evidence in logs instead of silently proceeding as if the lock had been acquired. Added a second unit test (`test_lock_acquire_wait_times_out_and_reports`) asserting that message appears on an actual timeout.

If a full caller audit is wanted, I'm glad to scope that as its own follow-up - happy to hear whether that's of interest.

Re-ran `fm-backend-cmux`, `fm-test-run`, `fm-wake-queue`, `fm-home-summary-refresh`, `fm-inactive-reconcile`, `fm-afk-return`, and `fm-watcher-lock` (which carries the two new tests): all pass, 0 failures.

PR #3736, opened 2026-09-04
fix(secondmate): bind worker pools to their owning homes

wjkawecki-jt:
## Intent

Every Firstmate secondmate home cannot launch a Claude worker for its own project clones, and its non-Claude workers have been silently working in worktrees of the primary home's clone instead of their own.
The captain's standing instruction is that firstmate owns the whole internal process and fixes such defects without asking him; this task is that fix.
Observed on 2026-09-04 in the design-system secondmate home at ~/.treehouse/firstmate-7bab20/6/firstmate: its projects/web-app-design-system clone has the same remote as the primary's projects/web-app-design-system, treehouse keys the pool by that shared identity, so `treehouse get` run from the secondmate clone hands out ~/.treehouse/web-app-design-system-c96007/N/web-app-design-system whose `.git` file reads `gitdir: ~/Development/firstmate/projects/web-app-design-system/.git/worktrees/...`, a linked worktree of the PRIMARY clone.
bin/fm-claude-trust.sh's structural scope test then correctly refuses, because the worktree does not share the secondmate clone's common dir, and bin/fm-spawn.sh refuses the Claude launch rather than wedging on the trust dialog.
Codex, Gemini, Cursor and the other harnesses have no equivalent pre-registration, so their workers launched from that home have been working in the primary clone's worktrees all along, which only works because the remote is the same.

## What Changed

- Added shared Treehouse setup and pool-binding helpers, and updated bootstrap, install, spawn, trust, and teardown flows to provision and clean up secondmate-owned worker pools safely.
- Documented the secondmate pool-binding configuration and architecture changes.
- Expanded fixtures and regression coverage for worker spawning, worktree settlement, teardown safety, backend behavior, and secondmate lifecycle scenarios.

## Risk Assessment

✅ Low: The change is focused on binding non-primary worker acquisition and teardown to a home-local Treehouse root, with fail-closed behavior for unsupported Treehouse versions and corresponding coverage updates.

## Testing

- Outcome: ⚠️ 1 error across 1 run (51m53s)

## Pipeline

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>

</details>

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

### DIFF
--- a/bin/fm-bootstrap.sh
+++ b/bin/fm-bootstrap.sh
@@ -52,7 +52,9 @@
 #          on a feature branch instead of its default branch - a crewmate's work
 #          landed in the primary instead of its own worktree; restore it per the line.
 #          treehouse is also MISSING when its installed version lacks
-#          "treehouse get --lease" support.
+#          "treehouse get --lease" support, or when this home needs its own pool
+#          root and the installed version lacks the global --root option
+#          (docs/configuration.md owns the per-home pool contract).
 #          no-mistakes is also MISSING when its installed version is older than
 #          1.46.0 (structured pipeline attestation floor; see CONTRIBUTING.md).
 #          The AXI-family floor policy is owned beside GH_AXI_MIN and
@@ -181,6 +183,8 @@ DATA="${FM_DATA_OVERRIDE:-$FM_HOME/data}"
 # deferred network stage sets, so an ordinary bootstrap run records nothing.
 # shellcheck source=bin/fm-timing-lib.sh disable=SC1091
 . "$SCRIPT_DIR/fm-timing-lib.sh"
+# shellcheck source=bin/fm-treehouse-lib.sh disable=SC1091
+. "$SCRIPT_DIR/fm-treehouse-lib.sh"
 
 # Network-phase selection (see the header). An unrecognized value resolves to
 # `all` so a malformed override runs every step rather than silently dropping a
@@ -911,6 +915,16 @@ treehouse_supports_lease() {
   treehouse get --help 2>&1 | grep -Eq '(^|[^[:alnum:]_-])--lease([^[:alnum:]_-]|$)'
 }
 
+# This home needs Treehouse's --root option only when it resolves a pool root of
+# its own; the primary home keeps Treehouse's default precedence and never does.
+# An unresolvable root is its own detect-only diagnostic, so treat it as needing
+# the option rather than quietly clearing the gate.
+treehouse_root_option_required() {
+  local root
+  root=$(fm_treehouse_root_for_home "$FM_ROOT" "$FM_HOME") || return 0
+  [ -n "$root" ]
+}
+
 # Shared semantic-version floor for the tool gates below. A version string that
 # cannot be parsed into exactly one major.minor.patch triple is incompatible,
 # never assumed current, so a development or vendored build cannot pass a floor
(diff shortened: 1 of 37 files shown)


wjkawecki-jt:
Answering both review remarks.

**P1 - `bin/fm-spawn.sh` unguarded assignment.** Fixed. The call site is now `if ! TREEHOUSE_HOME_ROOT=$(fm_treehouse_root_for_home ...); then` with an explicit diagnostic and `exit 1`, so an unresolvable root reports the problem instead of aborting silently under `errexit`.

**P2 - `bin/fm-treehouse-lib.sh` "root resolution failures are masked".** Working as intended, and deliberately not changed. The function is total over home shapes rather than partial:

- An empty result on success means "this home uses Treehouse's default root", which is the correct and only safe answer for a shape the structural test does not recognise as a non-primary home (a plain directory, an unrelated clone, a home nested inside some other repository). It preserves today's behaviour rather than inventing a pool location.
- Genuine failures do return non-zero: an unreadable home or code-root path, and unreadable Git metadata for a home that reached the metadata stage, all `return 1`. A home carrying a valid secondmate marker never falls back - if its structure cannot be resolved, it fails rather than resolving to the default pool.

The masking concern assumed acquisition and return could disagree. They cannot: `bin/fm-spawn.sh` and `bin/fm-teardown.sh` call the same function with the same inputs, so a home that acquired from the default pool also returns to the default pool. Making the unrecognised-shape case an error instead would refuse teardown for every home that works correctly today.

**Additional fix in this revision - the real cause of the red Herdr lane.** The three failing behaviour tests all ended in `treehouse get did not enter a worktree within 60s` when spawning from a secondmate-shaped home. Treehouse's global `--root` option was added in **v2.2.0**, and CI pinned **v2.0.1**, which rejects the flag (and ignores `TREEHOUSE_ROOT` as well, so there is no environment-variable fallback on that build).

- `bin/fm-install-treehouse.sh` now pins v2.3.0, the version the local real-Herdr suite actually exercises, with fresh checksums for all four assets.
- `fm_treehouse_supports_root` probes `treehouse get --help` the same way the existing `--lease` gate does. A home that resolves a pool root of its own and finds no `--root` support now refuses at spawn, refuses at teardown with its own exit code so the child path can never fall back to raw removal, and is reported by the session-start bootstrap as needing a treehouse upgrade. Pooling into another home's clone - the defect this PR fixes - is never the fallback.

kunchenguid:
Speaking as Kun's firstmate:

First look on HEAD `16b054e532231bd81a4322cc793095a2fafd3eb8` vs main (now includes squash merges of #3783/#3785). Attestation **MATCH**. Tip NM `33969928495` SUCCESS (earlier fail run superseded). CI `33969917248` SUCCESS. Greptile 5/5; author answered P1/P2 (2026-09-05T14:06Z).

**Contract-class: restore** — secondmate/non-primary homes must not pool into the primary's Treehouse clone (isolation already promised). Primary keeps default pool; non-primary binds `FM_HOME/config` via `--root` and refuses without v2.2.0+. Treehouse pin v2.0.1→v2.3.0 is enablement for that promised binding, not a new default for unconfigured primary.

**VISION:** (1) OK (2) authority explicit via structural home root OK (3) scripts own mechanics OK (4) restart non-event OK (5) OK (6) vendor pin only as isolation floor OK (7) scope OK for secondmate isolation.

**Outcome: waiting-author** — `mergeable=CONFLICTING` / DIRTY after main moved. Please rebase onto current main and re-raise via no-mistakes so tip attestation + NM stay MATCH/green. Not waiting-captain (author still needs to clear conflicts). No security FYI.

wjkawecki-jt:
Superseded by #3873, which carries the same pool-binding fix rebased onto current main, raised through no-mistakes, with the fish-shell quoting defect in the worktree-root argument fixed and covered by a test. Closing this one to avoid two open pull requests for the same change.

PR #3822, opened 2026-09-06
Fix two steer-delivery defects in firstmate's own tooling, plus the test isolation hole that surfaced them

jordanhindo:
## 1. A busy cursor pane's empty composer read `pending`, so steers lost their doorbell

While cursor-agent is working it right-aligns a dim `ctrl+c to stop` on the composer row itself, beside its dim `Add a follow-up` placeholder. Ghost stripping removes both and leaves only the reverse-video remnant, but that remnant is proven against the *unstripped* plain row, which then ends in the hint and no longer matches the anchored placeholder - so an empty composer classified `pending`. `fm_task_inbox_ring` skips the doorbell on exactly that verdict, so for as long as a worker stayed busy every steer went in durably and silently, and the watcher's re-ring skipped for the same reason; five records in a row on one task, one delivered about fifteen minutes late. The fix proves the remnant against the plain row's **first column** only. With no gap that is the whole row, so every existing verdict is unchanged, and the length guard still compares against that column, so text typed to exactly match the placeholder stays `pending`. Reproduced and verified against live tmux Cursor panes: the false ones now read `empty`, while a pane genuinely holding an unsent line still reads `pending`. Covered by a new case in `tests/fm-composer-lib.test.sh` built from the byte-exact live capture, with non-vacuity anchors on both the surviving remnant and the hint.

This corrects a prior expectation rather than only adding one. `tests/fm-backend-herdr.test.sh` pinned the same mid-turn row as `pending` and recorded that the shared verdict was "deliberately NOT relaxed - a right-aligned status token on the composer row is content". That token is furniture, never user input, so the expectation pinned this defect; it was scoped around at the time by adding herdr's rendered-footer signal instead. The third commit asserts the corrected verdict, tells the placeholder row and a typed row apart, and re-anchors both herdr submit cases on a typed capture so the rendered-footer transition they exercise is still reached through a genuinely pending composer rather than passing vacuously. Please sanity-check that call - it is the one judgement here that overturns a documented decision.

## 2. A recycled pid from a sibling home was reported as this home's watcher

The watcher's singleton check treated any live pid recorded in the lock as a running peer. When a watcher dies and the OS recycles its pid onto a sibling firstmate home's `fm-watch.sh` - routine on a machine running several homes - the stale lock still named that number, so `fm-watch.sh` reported `already running pid <N>` and declined. The arm layer applies the identity rule (`fm_watcher_healthy`), correctly saw no watcher, started a child that immediately declined, and reported `FAILED - cycle ended without an actionable reason`, leaving the home unsupervised; observed twice in five minutes. The fix requires the recorded pid to still be the process this home recorded (`fm_watcher_lock_matches_pid`, the same rule the arm layer already uses) before treating it as a peer, and clears the proven-stale lock through the existing recovery transition so the downtime it records is preserved. An absent recorded identity proves nothing and still defers exactly as before. `tests/fm-watcher-lock.test.sh` gains a case with a live foreign pid, a mismatched recorded identity and a *fresh* beacon - so only the identity rule can let it through - which fails without the fix and passes with it.

## 3. The test suite could run against a real firstmate home

Every test isolates itself by *inventing* `FM_HOME` and friends, and an inherited value silently wins - `tests/wake-helpers.sh` only invents `FM_ROOT_OVERRIDE` when none is set. A shell that had exported them therefore aimed the whole suite at that live home: real state files written, config deleted, and the running watcher's lock stolen out from under it. This is not hypothetical - it is how the defect was found. `tests/lib.sh` now refuses outright when one of those overrides is set, so a direct `bash tests/<x>.test.sh` cannot slip past; the 26 smoke and e2e scripts do not source it, so `bin/fm-test-run.sh` clears the same overrides for every script it launches. Nothing in the suite, CI, or the runner sets them for a test, so there is nothing legitimate to preserve. One test in `tests/fm-test-run.test.sh` proves both halves, with a probe that first demonstrates it can observe an inherited value so the scrubbed reading cannot pass vacuously.

## Verification

`bin/fm-lint.sh` clean throughout (ShellCheck 0.11.0, actionlint 1.7.12).

Full-suite counts are **not** included, and I would rather say so than round up: the host ran out of memory and killed `--all` four separate times (twice with another suite alongside, once alone, once as serial `--jobs 1` lanes). Please let CI's regression lanes be the authority here. What did complete, all under a scrubbed environment:

- `--all` on this branch: **444 ok, 0 failures** before the kill.
- `--lane portable-parallel-1 --jobs 1`: **171 ok, 0 failures** before the kill.
- The suites covering this diff, run together under `--jobs 1`: **269 ok, 2 failures**, both reproduced on untouched `upstream/main` (see below). Per suite: `fm-composer-ghost` exit 0, `fm-watcher-lock` exit 0 (97 ok), `fm-backend-herdr` exit 0 (182 ok), `fm-send-inbox` exit 0 (12 ok), `fm-test-run` exit 0 (30 ok).
- A clean `upstream/main` baseline run for comparison: 701 ok, 3 failures, none of them related to these changes. Two failures reproduce identically on this branch and on untouched `upstream/main` on macOS Bash 3.2: `a half-block rule row must count as a structural edge` (an existing fixture uses `\uXXXX` escapes, which Bash 3.2's `printf` does not expand) and `guard repair line did not source the X-mode cadence config`. They are pre-existing and out of scope here. A third, `re-arm stayed live instead of surfacing durable wakes and the still-open remote decision`, is the host OOM killer SIGKILLing the watcher child (`Killed: 9` in the log); it reproduces identically on untouched `upstream/main` under the same memory pressure, so it is environmental rather than a regression in the watcher change.

Defect 1 was additionally verified against live tmux Cursor panes rather than fixtures alone: the panes that were falsely skipping doorbells now read `empty`, while a pane genuinely holding an unsent line still reads `pending`.

### DIFF
--- a/bin/fm-composer-lib.sh
+++ b/bin/fm-composer-lib.sh
@@ -530,6 +530,17 @@ fm_composer_classify_content() {  # <bordered> <content> [idle_re] [idle_case] [
       plain_body=${plain_body#*"$plain_glyph"}
     fi
     fm_composer_normalize_trim_var plain_body
+    # A harness may right-align a status hint on the composer row itself,
+    # separated from the input by a run of spaces: cursor-agent draws
+    # `ctrl+c to stop` there for as long as it is working. That furniture is not
+    # input, and leaving it on the row defeats the anchored placeholder match
+    # below, so a busy cursor pane's idle composer read `pending` and every
+    # steer sent while it worked skipped its doorbell. Keep the row's FIRST
+    # column only: with no gap this is the whole row, so every gapless verdict
+    # is unchanged, and the length guard still compares against that column, so
+    # text typed to exactly match the placeholder stays `pending`.
+    plain_body=${plain_body%%"  "*}
+    fm_composer_normalize_trim_var plain_body
     if [ "${#content}" -lt "${#plain_body}" ] \
        && fm_composer_idle_matches "$plain_body" "$idle_re" "$idle_case"; then
       case "$plain_body" in
--- a/bin/fm-test-run.sh
+++ b/bin/fm-test-run.sh
@@ -123,6 +123,15 @@ RUN_STARTED_MS=$(now_ms)
 ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
 cd "$ROOT" || exit 1
 
+# Scrub the home-selecting overrides out of every test's environment. Each test
+# isolates itself by inventing these, so an inherited value silently wins and
+# aims the suite at a real firstmate home. tests/lib.sh refuses outright when it
+# sees one, but the smoke and e2e scripts here do not source it, so the runner
+# clears them for every script it launches instead. Nothing downstream sets
+# them, so there is nothing legitimate to preserve.
+unset FM_HOME FM_ROOT_OVERRIDE FM_STATE_OVERRIDE FM_DATA_OVERRIDE \
+  FM_PROJECTS_OVERRIDE FM_CONFIG_OVERRIDE
+
 MODE=
 LIST_ONLY=0
 LIST_SCHEDULED=0
--- a/bin/fm-watch.sh
+++ b/bin/fm-watch.sh
@@ -1406,9 +1406,20 @@ elif [ -n "${FM_LOCK_HELD_PID:-}" ] && [ -s "$WATCH_LOCK/pid-identity" ] \
   # not matching is positive proof the holder is gone, so clear the stale lock
   # through the recovery transition (the downtime it records is real) and take
   # it; an absent identity proves nothing and still defers below.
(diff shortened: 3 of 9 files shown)


jordanhindo:
Serialized identity-based stale-lock reclamation in bin/fm-watch.sh under the existing steal mutex through recheck, removal, and acquisition, with a two-reclaimer regression in tests/fm-watcher-lock.test.sh.

jordanhindo:
Updated test_foreign_home_watcher_pid_is_not_a_peer in tests/fm-watcher-lock.test.sh so the watcher and lock share one explicit FM_HOME and the fixture now fails only on PID identity mismatch.

jordanhindo:
The required composer suite's half-block structural-edge assertion also fails on untouched upstream/main at f9aca25, so that failure is pre-existing and unrelated to these changes.

PR #3851, opened 2026-09-06
docs: honour a project's own substituted review as part of its route

ashkonmousavi:
Resolves the conflict between a project's maintenance route that substitutes one focused independent review for the no-mistakes Review/Test/Document phases and AGENTS.md section 7's 'follow the faster path without adding an independent reviewer'.


ashkonmousavi:
Opened against upstream by mistake; this change lands in the fork's local main. Closing.

PR #3867, opened 2026-09-06
feat(bin): add verified omp (Oh My Pi) harness adapter for crew, secondmate, and primary

danielkuykendall23-boop:
## What Changed

- Registers `omp` as a verified harness across the `bin/` toolbelt: `fm-harness.sh` detects it through the Firstmate-owned `FM_OMP_HARNESS=omp` marker gated on an anchored `omp` process ancestor, session-lock and tmux liveness recognize the exact process name, `fm-control-lib.sh` gains its interrupt (`Escape`) and exit (`/quit`) contract, `fm-composer-lib.sh` accepts the `Working…` busy footer and treats omp's status row as a composer boundary, and `fm-spawn.sh` adds a launch template that clears foreign markers, passes the tracked `.omp/fm-worker-overlay.yml` with `--auto-approve --cwd`, validates `<provider>/<id>` models against `omp models --json`, and writes a per-task `state/<id>.omp-ext.ts` busy-state extension (busy on `agent_start`, idle on `agent_end` without `willContinue`).
- Adds two tracked primary extensions under `.omp/extensions/`: a watcher bridge that arms `fm-watch-arm.sh --restart` through the `fm_watch_arm_omp` tool and owns successor and replacement handoff, and a turn-end guard that answers omp's blocking `session_stop` hook by forcing one continuation when `fm-turnend-guard.sh` returns 2. `fm-wake-lib.sh` classifies omp under the `extension` supervision model with its own ownership proof, `fm-session-start.sh` reports when both extensions are not loaded, and `fm-supervision-instructions.sh` renders the new `docs/supervision-protocols/omp.md`.
- Extends `fm-quota-choose.sh` to key `omp:` candidates on the model prefix (`openai-codex/` to the codex row, `claude-bridge/` to the claude row, any other prefix refused), adds the harness-adapter reference and README/docs coverage, and adds a portable regression suite (`tests/fm-omp-harness.test.sh`) plus an opt-in live end-to-end test gated by `FM_OMP_LIVE_E2E=1`.

## Risk Assessment

✅ Low: The fix round is a small, bounded change to the omp arm of an optional selection helper: unmapped omp prefixes now fail closed before selection with the prescribed message, the bare model id is what is compared against model and product scopes (traced: omp:openai-codex/codex_bengalfox is vetoed by the exhausted model scope while omp:openai-codex/codex_other selects on the provider-wide row), the new test exercises the public argv interface and would fail on the pre-fix code, and the supervision.md omp guard row now claims only what the live test asserts (tool invocation count of at least two and a live watcher pid after the thaw).

## Testing

- Outcome: ⚠️ 1 error across 1 run (51m41s)

## Pipeline

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>
<summary>🔧 **Review** - 6 issues found → auto-fixed (5) ✅</summary>

- ⚠️ `bin/fm-remote-secondmate-control.sh:230` - cmd_relaunch still uses the pre-omp harness list while cmd_launch (line 161) accepts omp. Concrete path: a remote omp secondmate is launched successfully, then bin/fm-secondmate-restart.sh:167 calls `fm-remote-secondmate-control.sh relaunch &lt;id&gt; omp ...`, which dies with &#34;unverified remote secondmate harness: omp&#34;. Remote omp secondmates therefore cannot be restarted through the control plane. Add `omp` to the case at line 230.
- ⚠️ `bin/fm-spawn.sh:3591` - The new `pi|pi-signed|omp) supervision_model=extension` arm also changes existing Pi and pi-signed secondmates: at the base commit they received FM_SUPERVISION_MODEL=persistent (only claude|cursor were autoarm, everything else persistent). Because fm_supervision_model in bin/fm-wake-lib.sh:186 honours the env override before detection, every Pi secondmate&#39;s pull-guard verdict now tolerates an unheld lock with a fresh beacon when the Pi ownership proof holds, where it previously alarmed. This is probably the intended alignment with fm_supervision_model&#39;s own table, but it is an unstated behaviour change to an existing adapter, not part of the omp intent. Confirm it is deliberate and name it in the change description.
- ℹ️ `bin/fm-remote-doctor.sh:60` - HARNESS_TOOLS still lists claude codex opencode pi pi-signed grok kimi. The readiness check at line 329 requires at least one of these on the remote host, so a host that only has omp installed reports no required harness even though fm-remote-secondmate-control.sh now accepts omp launches there. Adding omp to the array also routes it through repair_tool_wrapper (line 432-438), which has not been exercised for omp, so this needs a decision rather than a mechanical edit.
- ⚠️ `bin/fm-composer-lib.sh:327` - Simplification: FM_DELIVERY_OMP_BUSY_REGEX_DEFAULT accepts both `Working…` and `Working...`. The three-dot form is documented as omp&#39;s headless `-p` stderr output, but fm-spawn never launches omp with -p and every supervised pane is the TUI, which renders U+2026. No intent requirement needs the second spelling for the omp-keyed regex; the narrower form `Working…` plus the spinner alternative satisfies the change. (Widening the harness-less union default is fine and unaffected.)
- ⚠️ `.omp/fm-worker-overlay.yml:19` - Simplification: `tools.approvalMode: yolo` in the overlay is a second definition of the rule the launch line already pins with `--auto-approve`; the comment itself says the two agree even if one is dropped. The intent requires one unattended-approval pin. omp&#39;s help lists `--auto-approve` and `--approval-mode` as separate mechanisms, so confirm they are equivalent and, if so, keep only one.
- ℹ️ `.omp/extensions/fm-primary-omp-watch.ts:3` - The omp watcher is a ~1000-line near-verbatim copy of .pi/extensions/fm-primary-pi-watch.ts (arm, successor, retry, replacement handoff) with branch/calm removed and sendUserMessage treated as synchronous. No shared abstraction is requested here, but every future fix to the Pi watcher&#39;s continuity logic now has to be mirrored by hand; the header states this is deliberate.

🔧 Fix: omp: unverify remote routes, narrow busy regex, drop overlay approval pin

- ⚠️ `.agents/skills/harness-adapters/references/harness/omp.md:56` - The line added by the fix round names `../../../bin/fm-spawn.sh --remote` as a remote secondmate route, but fm-spawn.sh has no --remote flag: remote placement is selected by the `remote=1` field of the secondmate&#39;s row in data/secondmates.md (bin/fm-spawn.sh:560, read inside spawn_remote_secondmate during an ordinary --secondmate spawn). This is an agent-runtime reference loaded before every spawn, so an agent following it would pass a flag that does not exist. Rewrite the sentence to say the remote route is a secondmate registered remote in data/secondmates.md spawned through fm-spawn.sh --secondmate, plus fm-remote-secondmate-control.sh launch.
- ⚠️ `bin/fm-spawn.sh:1638` - omp_model_validate runs with $MODEL as resolved so far, but the durable secondmate model pin from config/secondmate-harness is applied only afterwards (SM_MODEL at line 1650 sets MODEL when no per-spawn harness was given). Concrete sequence: config/secondmate-harness reads `omp openai-codex/gpt-nope`; `fm-spawn.sh sm &lt;home&gt; --secondmate` reaches line 1638 with MODEL empty, validation returns 0, MODEL is then set to openai-codex/gpt-nope, and the launch line carries `--model &#39;openai-codex/gpt-nope&#39;` even though the provider is listed and the id is not. The header promise that a listed provider&#39;s unknown id refuses before endpoint creation does not hold on the very path the header calls durable across respawns. Cursor&#39;s catalog check at line 1621 shares this ordering (pre-existing, out of scope). Move the omp validation below the secondmate token resolution block (guarded on HARNESS=omp) so the config-sourced model is checked too.
- ℹ️ `.agents/skills/harness-adapters/references/harness/omp.md:33` - The Detection section states that every non-omp launch template clears FM_OMP_HARNESS, but no template does: the per-template env lines clear CLAUDECODE, PI_CODING_AGENT, GROK_AGENT, FM_PI_HARNESS and Cursor/Gemini markers, and the shared boundary at bin/fm-spawn.sh:3572 clears only CURSOR_AGENT, CURSOR_INVOKED_AS, and GEMINI_CLI. Behaviour is unaffected today because daemon-created panes do not inherit the spawner&#39;s environment and bin/fm-harness.sh treats the marker as inert without an omp ancestor, but the documented invariant is not what the code enforces. Either add `-u FM_OMP_HARNESS` to the shared clear at line 3572 (consistent with how the sibling Firstmate-owned FM_PI_HARNESS marker is cleared) or correct the sentence to describe the ancestry requirement as the real defence.

🔧 Fix: omp: validate config-pinned model, correct remote and marker docs
2 warnings still open:

- ⚠️ `tests/fm-omp-harness.test.sh:182` - The fix round moved omp_model_validate below the config/secondmate-harness token block (bin/fm-spawn.sh:1662) so a durably pinned model is checked, and I verified the new ordering is correct: for `omp openai-codex/gpt-nope` in config/secondmate-harness with `fm-spawn.sh sm &lt;home&gt; --secondmate`, HARNESS resolves to omp, OMP_BIN is set in the case arm, SM_MODEL sets MODEL, and the validation now refuses. But no test exercises that path. Every case in test_spawn_model_validation_scoped_to_listed_providers passes `--harness omp --model ...` explicitly, and test_secondmate_launch_relies_on_discovery passes `omp` as the positional harness, so SM_MODEL never applies and the whole suite passes with the call in either position (it passed before the fix). Add one case in the secondmate-world shape of test_secondmate_launch_relies_on_discovery: write `omp openai-codex/gpt-nope` to `$world/home/config/secondmate-harness`, spawn `fm-spawn.sh sm &lt;home&gt; --secondmate` with no harness positional and the fake omp on PATH, and assert exit 1, the &#34;is not listed by &#39;omp models --json&#39; although provider &#39;openai-codex&#39; is&#34; refusal, and no sm.meta. That test fails at the pre-fix ordering and passes now.
- ⚠️ `.omp/fm-worker-overlay.yml:30` - Simplification: the overlay pins `steeringMode: one-at-a-time` and `followUpMode: one-at-a-time`, which the overlay&#39;s own comment says are omp&#39;s shipped defaults. They fall outside the overlay&#39;s stated scope (settings whose captain-level value would park a worker on a prompt, change its pinned model, or make its composer unreadable), and they are absent from the pin list in the bin/fm-spawn.sh header at line 132 (&#34;composer shape, plan mode off, prewalk off, and the non-interactive usage-reserve policy&#34;), which omp.md declares to be the exact list. The comment cites &#34;the ordering the delivery guard expects&#34;, but nothing in bin/ or docs/ references steering or follow-up mode, so no stated requirement needs these two pins. Recommend removing both keys and their comment so the overlay matches the header&#39;s list; if the pins are wanted, that is the user&#39;s call and the header list should then name them.

🔧 Fix: omp: pin config-model validation with a test, trim overlay

- ⚠️ `docs/verification/supervision.md:251` - The omp turn-end guard row states the lab evidence was produced &#34;with the extension&#39;s retry bound at zero and the successor watcher killed&#34;, and the continuity row at line 467 states &#34;closing rpc stdin retired the session cleanly&#34;. The shipped live test contradicts both: stage 3 of tests/fm-omp-primary-live-e2e.test.sh freezes the successor watcher with SIGSTOP until the beacon passes FM_GUARD_GRACE (line 251, added by commit 7e163fc because a killed watcher closes its arm child and the extension re-arms before the guard can fire), the omp launch sets no FM_WATCH_REARM_RETRY_LIMIT at all (and positiveInteger in the watch extension maps 0 to the fallback of 5, so a zero bound is not expressible), and the shutdown stage records that omp 18.1.11 outlived its closed stdin for more than 30s as a note rather than asserting a clean exit (line 302, commit 0509741). The test&#39;s own header item 4 (line 14, &#34;with retries exhausted and the watcher gone&#34;) carries the same stale description. This is the maintainer verification record docs/turnend-guard.md and README point to, so it currently documents a mechanism the cited test does not exercise. Rewrite the two rows and the header item to describe the frozen-watcher stage (beacon stale past guard grace with the arm child still attached, guard rc=2 followed by a stop_hook_active stop, model repair through fm_watch_arm_omp returning the ownership no-op) and to say the rpc shutdown is observed and noted, not asserted.
- ℹ️ `.omp/extensions/fm-primary-omp-watch.ts:576` - Simplification: deliverActionableWake still takes the repairFailed parameter inherited from the Pi port, but with the supervision-branch offer removed its only use is `void repairFailed;` at line 592, and the sole caller at line 708 computes Boolean(restoration.failure) just to pass it. Drop the parameter and that argument; the ported comment &#34;No supervision branch on omp: every actionable wake goes to main&#34; already records why.
- ℹ️ `bin/fm-quota-choose.sh:306` - provider_for_harness has no omp arm, so `bin/fm-quota-choose.sh --candidate omp:&lt;model&gt;` now fails at line 355 with &#34;unknown harness: omp&#34; even though fm_control_harness_supported accepts omp and crew_dispatch_validate in bin/fm-bootstrap.sh accepts omp rules in config/crew-dispatch.json. The quota-array-dispatch skill names this helper as the canonical worker-side chooser, so a captain who writes an omp dispatch rule and a worker that follows the skill hits a misleading refusal. gemini and rovo share this pre-existing gap, and omp has no single primary provider family (openai-codex, claude-bridge, and ollama all run under it), so leaving it unmapped may be deliberate. This is a policy decision rather than a defect: either map omp to a family (for example codex, matching the verified openai-codex default) or leave it unmapped and accept the current refusal.

🔧 Fix: omp: sync guard evidence, drop dead param, map quota family

- ⚠️ `bin/fm-quota-choose.sh:322` - The new omp arm returns the provider name `omp` for any model that is not openai-codex/ or claude-bridge/, and the user instruction for this fix stated that such a candidate should be handled as disclosed uncertainty that keeps the candidate eligible, never a refusal. The script&#39;s existing handling of a provider the snapshot does not report is the opposite: effective_for_provider_model returns {status: &#34;unknown&#34;} at line 345 for a provider absent from the snapshot, and the eligibility jq at line 387 maps .status == &#34;unknown&#34; to false, so the candidate is skipped. Concrete trace: `bin/fm-quota-choose.sh --snapshot &lt;any valid snapshot&gt; --candidate omp:ollama/qwen3:8b` prints `none` and exits 1, which the header at line 19 defines as an unsafe-dispatch signal; a bare `omp` candidate (model default) does the same. `omp:openai-codex/&lt;id&gt;` on a host where quota-axi reports codex unavailable follows the same path (codex row absent or unknown with no applicable scope) and is likewise excluded rather than surfaced as uncertainty. The header text at lines 40-45 (&#34;reads as unknown quota exactly like any provider quota-axi does not report&#34;, &#34;stays disclosed uncertainty&#34;) is mechanically accurate but does not say that the consequence is exclusion, so a captain who writes an omp dispatch rule with a local ollama model and a worker that follows the quota-array-dispatch skill&#39;s pointer to this helper gets `none` for a model that has no quota at all. This is a policy decision rather than a mechanical defect: either accept exclusion and state it plainly in the header (the chooser only ever selects known positive quota, matching the cursor provider-level-unknown behavior the existing tests pin), or, if unmeasured omp candidates must stay eligible, that is a new eligibility semantic for this script that goes beyond the prior finding&#39;s minimal fix and needs authorization. The remedy, not the defect, is what needs the user&#39;s call.
- ⚠️ `bin/fm-quota-choose.sh:381` - When the omp arm maps `openai-codex/&lt;id&gt;` to the codex row or `claude-bridge/&lt;id&gt;` to the claude row, the selection loop still passes the full prefixed model to effective_for_provider_model, whose $model_token is compared verbatim against `model:`/`product:` scopes at line 351. The quota-axi scopes carry bare ids (the shipped fixtures use `model:codex_bengalfox` and `model:fable`), so `openai-codex/codex_bengalfox` never equals `codex_bengalfox` and only all_models/all_products rows ever apply to an omp candidate. Concrete trace: a snapshot whose codex provider carries all_models known 40 percent through_reset plus model:codex_bengalfox known 0 percent exhausted_now; `--candidate codex:codex_bengalfox` is vetoed (applicable includes the exhausted scope, line 355), while `--candidate omp:openai-codex/codex_bengalfox` on the same snapshot is chosen because the applicable set is only the all_models row. The arm therefore declares that the prefix selects the family but leaves the family&#39;s model-scoped exhaustion unapplied, a wrong result with no error. Fix inside the arm the prior round added: when the omp arm consumes a prefix, pass the bare id (`${model#*/}`) as the scope token to effective_for_provider_model while still printing the original `&lt;harness&gt; &lt;model&gt;` on selection.
- ℹ️ `docs/verification/supervision.md:251` - The rewritten omp guard row states that fm_watch_arm_omp &#34;answered with the extension&#39;s ownership no-op because it still owned the frozen arm&#34;. The cited live test asserts only that the tool was called at least twice (tests/fm-omp-primary-live-e2e.test.sh lines 281-289) and that a live watcher exists after thaw (line 293); it never captures the tool&#39;s reply, and the no-op text lives only in the stage comment at lines 254-255. The reply is what the extension&#39;s lockOwnership branch at .omp/extensions/fm-primary-omp-watch.ts line 877 would produce, so the clause is expected by design, not observed evidence. In a verification record, reword it to say the repair call is expected to hit the ownership no-op (the lab does not capture the reply), or drop the clause. The test&#39;s own header item 4 at line 14 (&#34;with retries exhausted and the watcher gone&#34;) still carries the superseded description the earlier finding named; it was outside the files the user scoped for that fix, so it is noted here rather than reported again.

🔧 Fix: omp quota: refuse unmapped prefixes, match bare model scopes
</details>

<details>

</details>

<details>
<summary>🔧 **Document** - 2 issues found → auto-fixed ✅</summary>

- ℹ️ `tests/fm-omp-primary-live-e2e.test.sh:14` - The live test&#39;s header item 4 still says the guard stage runs &#34;with retries exhausted and the watcher gone&#34;, but the shipped stage freezes the successor watcher with SIGSTOP until its beacon passes the lab FM_GUARD_GRACE, as docs/verification/supervision.md now records. This is a comment in a test file, which the document phase may not edit; a one-line rewrite of that header item would close it.
- ℹ️ `docs/subagent-guard.md:178` - The harness wiring table states &#34;Every supported primary harness was reviewed&#34; for built-in delegation tools, but it has no omp row now that omp is a verified primary (Cursor is also absent, a pre-existing gap). The omp turn-end guard extension header mentions task/subagent sessions, so omp may expose a delegation surface. Establishing its status needs a live enumeration of omp&#39;s tools, as the Codex row was established, not a documentation edit; until then the sentence overclaims.

🔧 Fix: docs: add omp subagent-guard row, fix live test header
</details>

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>


kunchenguid:
Speaking as Kun's firstmate: this is merged. Thank you @danielkuykendall23-boop — really appreciate you taking the time on this.

HEAD `e5eb039c3776153b36ff0801b322df36a0670616` · attestation MATCH · CI 34062226406 SUCCESS · NM edited 34062227086 SUCCESS (synchronize 34062226345 was a body-race MISMATCH against older `cd51cf4b`, superseded by the edited raise) · contract-class opt-in (omp only when launched / detected; live E2E gated by `FM_OMP_LIVE_E2E=1`) · workflow-zero.

Security FYI (tip only, not a merge gate): omp has no project-trust dialog and the worker overlay uses `--auto-approve` / yolo for unattended sessions — same shape as other verified unattended harnesses; captain should know before choosing omp as primary.

PR #3765, opened 2026-09-05
feat(bin): add IMAP/SMTP mail plane with standing poll

feilipu:
## What Changed
- Add `bin/fm-mail.sh` and `bin/fm-mail.py` as a general-purpose mail plane: read unseen IMAP mail, send one SMTP message, and `poll` new UNSEEN messages into durable `check: mail <uid>` wakes keyed by IMAP UID and mailbox generation, with a per-poll wake cap, degraded fetch plus retry recovery, and journal-first exactly-once-recovering emission.
- Add `bin/fm-mail-check.sh` to arm/disarm a standing watcher check that runs that poll on the slow-check cadence and prints a wake line unless the poll is a proven no-op; timeouts and fail-closed polls that already queued mail always doorbell.
- Document the gitignored `.env` mail schema, expose `fm_wake_append_locked` so poll can commit a wake row atomically with its journal, and add colocated mail and mail-check behavior tests.

## Risk Assessment

⚠️ Medium: The persist-order and error-path logout fixes match the claimed invariants, but this remains a credentialed IMAP/SMTP plane with multi-phase wake, cursor, and retry state.

## Testing

Ran the mail and mail-check behavior suites, then exercised the public CLI and the two poll_list contracts from this change: a post-login IMAP failure still logs out, and a fail-closed retry-pos write leaves the cap-1 retry turn unspent so the next poll recovers uid 82 instead of new mail.

<details>
<summary>Evidence: Public fm-mail and fm-mail-check CLI surface</summary>

Source: [Public fm-mail and fm-mail-check CLI surface](https://github.com/feilipu/firstmate/blob/7dde6638f0b4471c4d56ce14e60910fe971156ac/.no-mistakes/evidence/fm/ship-bin-fm-mail-sh-email-plane-through-a4/cli-public-surface.txt)

```text
=== fm-mail.sh --help ===
fm-mail.sh read
fm-mail.sh send <to> <subject> <body | ->
fm-mail.sh poll
fm-mail.sh status

=== fm-mail.sh status ===
mail account: captain@example.com
imap: imap.example.com:993 smtp: smtp.example.com:465
cursor: (none)

=== fm-mail.sh status without config (stderr) ===
fm-mail: missing required $FM_HOME/.env value: FM_MAIL_USER
fm-mail: add FM_MAIL_USER (and the other three FM_MAIL_* values) to /tmp/tmp.yQAry6Xv23/.env

=== fm-mail-check.sh --help ===
Usage:
[...]

PR #3915, opened 2026-09-07
fix(bin): rotate small retry window and harden poll persist ordering

feilipu:
Fixes the Greptile P1 'Retry cursor ignores small windows' on https://github.com/kunchenguid/firstmate/pull/3765 and the two related runtime-contract findings from the no-mistakes review.

Changes:
-  now rotates the small-window branch by the durable position.
- The cap-one turn flag is persisted only after emit/flush.
- The retry-scan position advances only up to the first emitted retry uid, never past an unpublished wake.
- Add regression tests for all three findings.

Verification:
- FM_TEST_BEGIN 2026-09-07T11:04:51Z tests/fm-mail-check.test.sh family=watcher-wake-lock expected_gate_skip=none
FM_TEST_BEGIN 2026-09-07T11:04:51Z tests/fm-mail.test.sh family=watcher-wake-lock expected_gate_skip=none
ok - fm-mail-check: help and usage plumbing
ok - fm-mail-check: arm writes and binds, re-arm is idempotent, disarm removes
ok - fm-mail-check: arm resolves a relative home into the shim
ok - fm-mail-check: arm refuses a symlink at the shim path
ok - fm-mail-check: arm refuses without the mail plane
ok - fm-mail-check: healthy poll stays silent while the poll still surfaces new mail
ok - fm-mail-check: a poll failure is reported once and re-reported after recovery
ok - fm-mail-check: an unconfigured home is reported once, not every poll
ok - fm-mail-check: a slow poll times out into a one-line report
ok - fm-mail-check: a fail-closed poll after a wake reports the failure, not the wake
ok - fm-mail-check: a missing mail plane is reported, not assumed
FM_TEST_END 2026-09-07T11:04:58Z tests/fm-mail-check.test.sh exit=0 duration_ms=7343 gate_skip=false
ok - fm-mail: missing required configuration fails cleanly naming the variable
ok - fm-mail: environment values override the .env file
ok - fm-mail: status succeeds without network and prints configuration
ok - fm-mail: --help prints usage for every subcommand
ok - fm-mail: unknown subcommand prints usage and exits non-zero
ok - fm-mail: status never prints the password
ok - fm-mail: send passes body not empty through stdin
ok - fm-mail: poll propagates errors instead of swallowing them
ok - fm-mail: poll surfaces each new uid exactly once
ok - fm-mail: generation change prevents a reused uid from being suppressed
ok - fm-mail: poll heals a wake whose cursor record was interrupted
ok - fm-mail: a poll interrupted before evidence writes does not duplicate on recovery
ok - fm-mail: the poll lock serializes overlapping polls so mail wakes exactly once
ok - fm-mail: journal recovers a wake the drain already acknowledged
ok - fm-mail: an acknowledged wake whose cursor record was lost is recovered from the journal
ok - fm-mail: a legacy generation-less wake never marks a reused uid surfaced
ok - fm-mail: a missing wake library fails the poll instead of suppressing mail
ok - fm-mail: a wake with no durable record is rolled back, not left ackable
ok - fm-mail: a rollback failure never releases a wake the drain could acknowledge without a durable record
ok - fm-mail: per-poll wake cap bounds the durable queue without missing mail
ok - fm-mail: poll sanitizes tabs and newlines in header fields
ok - fm-mail: bounded fetch makes progress through a large backlog without re-surfacing mail
ok - fm-mail: an unfetchable uid is surfaced degraded and cannot starve later mail
ok - fm-mail: a transient fetch failure recovers real metadata on a later poll
ok - fm-mail: the retry-scan cursor advances past persistent failures
ok - fm-mail: retry-scan position does not advance past an unpublished wake
ok - fm-mail: retry budget advances the scan by examined candidates, not the window
ok - fm-mail: a retry window of unseen uids never stalls the cursor
ok - fm-mail: hung logout cannot advance the retry-scan position
ok - fm-mail: retry-scan position is not saved when a row is emitted
ok - fm-mail: small retry window rotates past an unfetchable prefix
ok - fm-mail: cap-one turn is not saved before emit/flush
ok - fm-mail: the reserved retry budget survives a new-mail flood
ok - fm-mail: a retry uid whose degraded wake never recorded is re-surfaced, not dropped
ok - fm-mail: a cap of one never suppresses new mail while retries exist
ok - fm-mail: a single contended slot alternates between new mail and retry recovery
ok - fm-mail: a zero retry budget leaves the retry-scan position unchanged
ok - fm-mail: a failed retry write fails the poll instead of losing recovery
ok - fm-mail: a journal write failure never cursor-records a suppressed mail
ok - fm-mail: a failed retry clear fails the poll; the published wake stays and the retry entry remains
ok - fm-mail: a failed stale-retry clear fails the poll instead of silently leaving a duplicate-wake entry
ok - tests/lib: assert_equals fails on mismatch so cursor assertions cannot false-pass
ok - fm-mail: a failed poll_list fails closed, wakes nothing, and releases the lock
ok - fm-mail: a recovered wake that cannot append restores the retry instead of stranding the metadata
ok - fm-mail: a death between retry remove and wake publish cannot strand recovered metadata
ok - fm-mail: a raised FETCH surfaces that uid degraded and advances
ok - fm-mail: the journal survives when the heal cannot commit a uid
ok - fm-mail: heal failure with unseen mail fails the poll instead of re-waking
ok - fm-mail: an empty plain-text alternative falls back to the html preview
ok - fm-mail: body_preview does not crash on a None payload
ok - fm-mail: read tolerates a None payload without crashing
ok - fm-mail: read does not silently hide an unfetchable uid
ok - fm-mail: a non-numeric port fails cleanly in bash
FM_TEST_END 2026-09-07T11:05:00Z tests/fm-mail.test.sh exit=0 duration_ms=9219 gate_skip=false
FM_TEST_SUMMARY total=2 failed=0 skipped_gate=0 duration_ms=9363
FM_TEST_SUMMARY_FAMILY family=watcher-wake-lock count=2 duration_ms=16562 failed=0
FM_TEST_SLOWEST rank=1 script=tests/fm-mail.test.sh duration_ms=9219
FM_TEST_SLOWEST rank=2 script=tests/fm-mail-check.test.sh duration_ms=7343 passes.
- fm-lint-workflows.sh: 3 workflow files valid passes.


feilipu:
Recreating with a clean PR body.

PR #3847, opened 2026-09-06
feat(bin): add opt-in branch-currency refresh dispatch

dbeihl-utilicast:
## Intent

The captain has now told us twice that he has to press Update branch himself before he can merge anything we hand him. His words on 2026-09-06: "i have had to update the branch EVERY time, firstmate. i cant merge if it isn't up to date (i guess i could but this seems risky)". He named a specific case the first time, on 2026-09-05, and it was filed and left queued. It kept happening. That is the shape of a reminder rather than a control, which is exactly what he holds we should not ship.

He is right that merging without being up to date is risky, and he should not have to choose between that risk and doing our work for us.

Verified fact behind it: all three active repositories require a branch to be current with its base before it can merge. utilicast-design-system, utilicast-plugins and utilicast-triage all have the strict up-to-date requirement enabled on their default branch. So this is not a preference; it is a hard gate on every single pull request we produce.

The captain also reviewed a proposed state-machine design (Observed/Refused/Dispatched/Resolved for a single branch-currency refresh) and said, verbatim, "build it." He separately decided, after a review finding showed a relaunch unrelated to a task's PR could permanently disable this control, to relax the worker-generation binding rather than keep it, and required: a test proving a refresh still dispatches after the worker generation changes, deduplication so no refusal path can wake him on a loop, and a PR note documenting that the generation binding was deliberately relaxed and why. After a further reviewer objection that always-on dispatch is new default behavior needing explicit consent, and that a crash before a delivery record is confirmed must not be mistaken for a completed delivery, the captain said, verbatim, "rebuild it rather than patch it again," and asked for: an explicit Observed/Refused/Dispatched/Resolved lifecycle where nothing is recorded as done until it is done, a restart that reaches the same conclusion as an uninterrupted run, opt-in per home rather than always-on, every refusal deduplicated, and a case shown where the control correctly refuses to act.

## What Changed

- `bin/fm-watch.sh` gains a new opt-in (`config/pr-refresh`) four-state Observed/Refused/Dispatched/Resolved branch-currency lifecycle (`pr_refresh_dispatch`) that reactivates a `done` GitHub PR worker when the poll reports its head as `behind` or `conflict`, defers on active work or unsupported states, deduplicates refusals per (head, reason), and never records Dispatched/Resolved until the inbox delivery record is confirmed on disk.
[...]

PR #3932, opened 2026-09-07
feat(bin): auto-refresh a waiting pull request that falls behind its base

dbeihl-utilicast:
## Intent

Captain, 2026-09-07: "dispatch an agent to fix https://github.com/kunchenguid/firstmate/pull/3847".

That pull request is the branch-currency rebuild. Its whole subject is keeping a pull request that waits on the captain from going stale. It was designed as a state machine after an earlier version accumulated six review findings, and the captain chose to rebuild it rather than patch it again.

It is currently failing two checks and has been sitting that way since roughly 06:20 this morning.

## What Changed

- `bin/fm-pr-poll.sh` now emits `behind <head>` or `conflict <head>` for an open GitHub PR in addition to `merged`, reading conflict from the PR's own `mergeable` field and behind-ness from a `compare/base...head` divergence check rather than `mergeStateStatus`, which collapses BEHIND under BLOCKED.
- `bin/fm-watch.sh` gains `pr_refresh_dispatch`, a four-state (Observed/Refused/Dispatched/Resolved) lifecycle keyed on task plus current head that reactivates a finished worker to bring its branch current: one dispatch per head, a quiet defer on `working` crew state or an already-dispatched head, and a deduped one-refusal-per-head for unsupported mode, remote-secondmate tasks, a status log predating the current `spawn_gen`, or a failed send. The whole path is opt-in via the new `config/pr-refresh` flag; without it, behind/conflict is absorbed into the triage log with nothing dispatched or recorded. `bin/fm-teardown.sh` clears the two new state markers.
- `bin/fm-send.sh` adds `FM_SEND_PRINT_INBOX_RECORD=1` to print the durable inbox record path after enqueue, so the dispatcher can bind its state to a confirmed record; `bin/fm-backend.sh` guards each backend source with a file-existence check. Tests cover the dispatch lifecycle in `tests/fm-pr-check-security.test.sh` and the send receipt in `tests/fm-send-inbox.test.sh`; `AGENTS.md`, `docs/architecture.md`, and `docs/configuration.md` document the flag, the state machine, and the state files.

## Risk Assessment

✅ Low: The dispatch state machine is coherent under concrete traces (crash-between-send-and-marker, corrupted marker, acknowledge-mid-rebase, head-moves-and-returns all converge without duplicating an instruction), the whole feature is behind the default-off config/pr-refresh flag, the poll fails closed on every unreadable forge field, and the only surviving findings are inaccurate header and inventory prose.

## Testing

Ran the two test files that cover every changed script (`fm-pr-check-security` 40/40 and `fm-send-inbox` 6/6, both exit 0), then produced product-level evidence by driving a real PR through the whole branch-currency lifecycle with the actual watcher and poll binaries against the suite's forge and worker fakes: poll 1 wakes the captain with `branch-refresh-dispatched` and queues the exact `FIRSTMATE_OP: v1 branch-currency` instruction into the worker's inbox; poll 2, after the worker acknowledges but the head has not moved, is silent to the captain and records `branch-refresh-deferred ... reason=dispatch-resolved` in triage; poll 3 at a genuinely moved head dispatches again; poll 4, once the PR is current, is silent — two instructions total for two heads. A home without `config/pr-refresh` stayed entirely silent and logged only the raw `behind` poll for triage. The change is a shell watcher with no rendered surface, so the reviewer-visible artifact is the CLI transcript rather than a screenshot. No failures, no flakes, nothing left in the working tree.

<details>
<summary>Evidence: Branch-currency lifecycle transcript (captain wakes, worker instruction, triage log, opt-out)</summary>

Source: [Branch-currency lifecycle transcript (captain wakes, worker instruction, triage log, opt-out)](https://github.com/dbeihl-utilicast/firstmate/blob/2734e9e9dd2677bdf45f7f321c2a247eaa275ad7/.no-mistakes/evidence/fm-brief-branch-currency/branch-currency-lifecycle.txt)

<code>===== poll 1: base moved, PR is behind at head 0123456 =====&#10;poll 1 captain wake:&#10;check: .../task-a.check.sh: branch-refresh-dispatched pr=https://github.com/o/r/pull/42 head=0123456789abcdef0123456789abcdef01234567 condition=behind&#10;&#10;instruction delivered to the worker inbox (.../task-a.inbox/001.msg):&#10;FIRSTMATE_OP: v1 branch-currency: Your open pull request https://github.com/o/r/pull/42 is not current with its base (behind or in conflict) at head 0123456789abcdef0123456789abcdef01234567. Re-run this brief&#39;s no-mistakes delivery contract with its exact serialized captain intent. ...&#10;&#10;===== poll 2: worker acknowledged, rebase still in flight, same head =====&#10;poll 2 captain wake:&#10;check: .../z-stop.check.sh: stop-cycle&#10;poll 2 captain triage log:&#10;branch-refresh-deferred pr=https://github.com/o/r/pull/42 head=0123456789abcdef0123456789abcdef01234567 condition=behind reason=dispatch-resolved&#10;&#10;===== poll 3: base advanced again, head genuinely moved to 2222222 =====&#10;poll 3 captain wake:&#10;check: .../task-a.check.sh: branch-refresh-dispatched pr=https://github.com/o/r/pull/42 head=2222222222222222222222222222222222222222 condition=behind&#10;&#10;===== poll 4: rebase landed, PR current at 89abcde =====&#10;poll 4 captain wake:&#10;check: .../z-stop.check.sh: stop-cycle&#10;&#10;===== instructions sent across the whole lifecycle =====&#10;one per moved head, 2 total:&#10;at head 0123456789abcdef0123456789abcdef01234567&#10;at head 2222222222222222222222222222222222222222&#10;&#10;===== a home that never opted in (no config/pr-refresh) =====&#10;opted-out poll captain wake: (silent - nothing reached the captain)&#10;opted-out triage log:&#10;check: .../task-a.check.sh: behind 0123456789abcdef0123456789abcdef01234567</code>

```text

===== poll 1: base moved, PR is behind at head 0123456 =====
poll 1 captain wake:
    check: /private/var/folders/j5/kgdthy6n37j7dgv0r7qr3qd80000gn/T/fm-pr-check-security.qpeyBF/evidence-lifecycle/home/state/task-a.check.sh: branch-refresh-dispatched pr=https://github.com/o/r/pull/42 head=0123456789abcdef0123456789abcdef01234567 condition=behind

instruction delivered to the worker inbox (/private/var/folders/j5/kgdthy6n37j7dgv0r7qr3qd80000gn/T/fm-pr-check-security.qpeyBF/evidence-lifecycle/home/state/task-a.inbox/001.msg):
    schema=fm-task-inbox.v1
    at=fixture
    --
    FIRSTMATE_OP: v1 branch-currency: Your open pull request https://github.com/o/r/pull/42 is not current with its base (behind or in conflict) at head 0123456789abcdef0123456789abcdef01234567. Re-run this brief's no-mistakes delivery contract with its exact serialized captain intent. Before any edit or branch movement, inspect no-mistakes axi status and do not act while an active run owns the branch. Let the pipeline's rebase step bring the branch current and re-establish every check on the resulting head. If rebase conflicts, report blocked and name the conflicted paths. Report done again only after the current-head checks are green.

===== poll 2: worker acknowledged, rebase still in flight, same head =====
poll 2 captain wake:
    check: /private/var/folders/j5/kgdthy6n37j7dgv0r7qr3qd80000gn/T/fm-pr-check-security.qpeyBF/evidence-lifecycle/home/state/z-stop.check.sh: stop-cycle
poll 2 captain triage log:
    [2026-09-07T11:28:04-0400] branch-refresh-deferred pr=https://github.com/o/r/pull/42 head=0123456789abcdef0123456789abcdef01234567 condition=behind reason=dispatch-resolved

===== poll 3: base advanced again, head genuinely moved to 2222222 =====
poll 3 captain wake:
    check: /private/var/folders/j5/kgdthy6n37j7dgv0r7qr3qd80000gn/T/fm-pr-check-security.qpeyBF/evidence-lifecycle/home/state/task-a.check.sh: branch-refresh-dispatched pr=https://github.com/o/r/pull/42 head=2222222222222222222222222222222222222222 condition=behind

===== poll 4: rebase landed, PR current at 89abcde =====
poll 4 captain wake:
    check: /private/var/folders/j5/kgdthy6n37j7dgv0r7qr3qd80000gn/T/fm-pr-check-security.qpeyBF/evidence-lifecycle/home/state/z-stop.check.sh: stop-cycle

===== instructions sent across the whole lifecycle =====
one per moved head, 2 total:
    at head 0123456789abcdef0123456789abcdef01234567
    at head 2222222222222222222222222222222222222222

===== a home that never opted in (no config/pr-refresh) =====
opted-out poll captain wake: (silent - nothing reached the captain)
opted-out triage log:
    check: /private/var/folders/j5/kgdthy6n37j7dgv0r7qr3qd80000gn/T/fm-pr-check-security.qpeyBF/evidence-opted-out/home/state/task-a.check.sh: behind 0123456789abcdef0123456789abcdef01234567
```
</details>
<details>
<summary>Evidence: Evidence driver script (reproduces the transcript above)</summary>

Source: [Evidence driver script (reproduces the transcript above)](https://github.com/dbeihl-utilicast/firstmate/blob/2734e9e9dd2677bdf45f7f321c2a247eaa275ad7/.no-mistakes/evidence/fm-brief-branch-currency/branch-currency-lifecycle.sh)

```text
#!/usr/bin/env bash
# Evidence driver: walks one PR through the branch-currency lifecycle end to end
# using the real fm-pr-check.sh / fm-pr-poll.sh / fm-watch.sh binaries and the
# suite's forge + worker fakes, printing what the captain and the worker see.
set -u
REPO=~/.no-mistakes/worktrees/2f2b4426b91c/01M1XVNPHKWC8D7NFYYCJX47TD
SUITE="$REPO/tests/fm-pr-check-security.test.sh"
shim=$(mktemp "${TMPDIR:-/tmp}/fm-evidence-shim.XXXXXX")
trap 'rm -f "$shim"' EXIT
sed -e 's|^\. "\$(dirname "\${BASH_SOURCE\[0\]}")/lib\.sh"|. "'"$REPO"'/tests/lib.sh"|' \
    -e '/^test_branch_currency_dispatch_and_active_refusal$/,$d' "$SUITE" > "$shim"
# shellcheck disable=SC1090
. "$shim"

PR=https://github.com/o/r/pull/42
HEAD_A=0123456789abcdef0123456789abcdef01234567
HEAD_B=2222222222222222222222222222222222222222
HEAD_C=89abcdef0123456789abcdef0123456789abcdef

banner() { printf '\n===== %s =====\n' "$*"; }
captain_view() {  # <label> <stdout-file>
  if [ -s "$2" ]; then printf '%s captain wake:\n' "$1"; sed 's/^/    /' "$2"
  else printf '%s captain wake: (silent - nothing reached the captain)\n' "$1"; fi
}

dir=$(make_case evidence-lifecycle)
state="$dir/home/state"
write_task_meta "$dir"
enable_pr_refresh "$dir"
run_check_entry "$dir" task-a "$PR" >/dev/null || { echo "arm failed"; exit 1; }
cat > "$dir/fakebin/fm-crew-state.sh" <<'SH'
#!/usr/bin/env bash
printf 'state: done \302\267 source: run-step \302\267 checks green: PR ready for review\n'
SH
chmod +x "$dir/fakebin/fm-crew-state.sh"
fake_idempotent_refresh_send "$dir"
: > "$dir/refresh-send.log"

cycle() {  # <out-name> <BEHIND_BY> <HEAD>
  FM_TEST_GH_STATE=OPEN FM_TEST_GH_BEHIND_BY="$2" FM_TEST_GH_HEAD="$3" \
    FM_CREW_STATE_BIN="$dir/fakebin/fm-crew-state.sh" \
    FM_PR_REFRESH_SEND_BIN="$dir/fakebin/fm-refresh-send.sh" \
    FM_TEST_REFRESH_SEND_LOG="$dir/refresh-send.log" \
    run_watcher_bounded "$dir/home" "$dir/fakebin" > "$dir/$1.out" 2> "$dir/$1.err"
}

banner "poll 1: base moved, PR is behind at head ${HEAD_A:0:7}"
cycle c1 1 "$HEAD_A"
captain_view "poll 1" "$dir/c1.out"
printf '\ninstruction delivered to the worker inbox (%s):\n' "$(ls "$state/task-a.inbox"/*.msg)"
sed 's/^/    /' "$state/task-a.inbox"/*.msg; printf '\n'

banner "poll 2: worker acknowledged, rebase still in flight, same head"
mv "$state/task-a.inbox"/*.msg "$state/task-a.inbox/handled/"
ack_watcher_cycle "$state" >/dev/null || echo "ack failed"
add_stop_custom_check "$dir"
cycle c2 1 "$HEAD_A"
captain_view "poll 2" "$dir/c2.out"
printf 'poll 2 captain triage log:\n'; grep 'branch-refresh' "$state/.watch-triage.log" | tail -1 | sed 's/^/    /'

banner "poll 3: base advanced again, head genuinely moved to ${HEAD_B:0:7}"
ack_watcher_cycle "$state" >/dev/null || true
cycle c3 1 "$HEAD_B"
captain_view "poll 3" "$dir/c3.out"

banner "poll 4: rebase landed, PR current at ${HEAD_C:0:7}"
mv "$state/task-a.inbox"/*.msg "$state/task-a.inbox/handled/" 2>/dev/null
ack_watcher_cycle "$state" >/dev/null || true
add_stop_custom_check "$dir"
cycle c4 0 "$HEAD_C"
captain_view "poll 4" "$dir/c4.out"

banner "instructions sent across the whole lifecycle"
printf 'one per moved head, %s total:\n' "$(grep -c . "$dir/refresh-send.log")"
grep -o 'at head [0-9a-f]*' "$dir/refresh-send.log" | sed 's/^/    /'

banner "a home that never opted in (no config/pr-refresh)"
out=$(make_case evidence-opted-out)
write_task_meta "$out"
run_check_entry "$out" task-a "$PR" >/dev/null || echo "arm failed"
FM_TEST_GH_STATE=OPEN FM_TEST_GH_BEHIND_BY=1 \
  run_watcher_bounded "$out/home" "$out/fakebin" > "$out/o1.out" 2> "$out/o1.err"
captain_view "opted-out poll" "$out/o1.out"
printf 'opted-out triage log:\n'; grep -o 'check:.*' "$out/home/state/.watch-triage.log" | tail -1 | sed 's/^/    /'
printf '\n'
```
</details>
<details>
<summary>Evidence: Targeted suite results</summary>

```text
tests/fm-pr-check-security.test.sh exit=0 40 ok, 0 failed (13 branch-currency cases)
tests/fm-send-inbox.test.sh exit=0 6 ok, 0 failed
```
</details>

## Pipeline

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>

- 🚨 `bin/fm-watch.sh:2025` - The `config/pr-refresh` gate covers only the dispatch, not the new poll output, so an opted-out home gets a repeating captain wake it never had before. `bin/fm-pr-poll.sh:81-83` emits `behind &lt;sha&gt;`/`conflict &lt;sha&gt;` unconditionally; with the flag absent this block is skipped and `out` falls through to `fm_wake_append check` + `wake` at bin/fm-watch.sh:2040-2042. Concrete path: home with no `config/pr-refresh`, task-a armed with an open GitHub PR whose mergeStateStatus is BEHIND. Cycle 1 emits `check: &lt;c&gt;: behind &lt;sha&gt;` and wakes (a full firstmate turn); `.last-check` is touched, the watcher exits, and the next run re-runs the check after CHECK_INTERVAL (default 300s) and wakes again with the identical reason. `fm_wake_append` has no dedup and the poll is never retired while the PR is open, so this repeats every ~5 minutes for as long as the PR sits behind - exactly the long-lived state this feature targets. Before the change the poll printed nothing for any non-merged PR, so this contradicts docs/configuration.md:208 (&#34;a home that has not asked for it sees no behavior change at all&#34;) and the diff&#39;s own test `test_branch_currency_opt_out_by_default`, which asserts the behind line in the watcher&#39;s woken stdout. Whether the right answer is to gate the emission on the flag, or to keep detection but dedup/triage-log it instead of waking, is a product call on the opt-out contract.
- ⚠️ `bin/fm-watch.sh:1488` - Re-dispatch after a resolved attempt has no cap or escalation, so a worker that acknowledges the instruction without moving the head produces an unbounded loop. Concrete path: attempt 1 dispatched at head H, worker mv&#39;s the record to handled/ and reports `done:` again without the rebase landing (a gate refusal reported as done rather than blocked). Next poll: head still H, marker status is `resolved`, so this branch computes attempt=2, builds a body that differs only in the attempt number - which defeats `fm_task_inbox_write_idempotent`&#39;s all-history dedup by design - writes a new inbox record, and returns 0, which the caller turns into a wake. That repeats every CHECK_INTERVAL indefinitely: one new durable inbox record and one captain wake per cycle, with no attempt ceiling and no pruning of `&lt;task&gt;.inbox/handled/`. The inbox re-ring ladder cannot backstop it because each record is genuinely handled. The earlier iteration of this feature bounded the equivalent case with FM_PR_REFRESH_STALE_SECS; the rebuild dropped it. Flagging the remedy rather than the defect as the thing needing authorization: capping attempts or escalating on a stuck head adds new policy/state beyond what this change currently does.
- ⚠️ `bin/fm-pr-check.sh:3` - The rewritten header now claims the script stores &#34;its worker generation&#34; alongside `pr=&lt;url&gt;` and `pr_head=&lt;sha&gt;`, but nothing in bin/fm-pr-check.sh writes or reads any generation - `grep -n &#39;spawn_gen\|generation&#39; bin/fm-pr-check.sh` matches only this comment line. The rebuild moved to reading `spawn_gen` live from `&lt;task&gt;.meta` inside `pr_refresh_dispatch`, so the arm-time pinning this sentence describes was abandoned; the header is now a false contract statement in a script whose whole purpose is a validated, non-interpolated sidecar. The same edit also dropped the accurate &#34;atomically arm a static merge poll&#34; wording for &#34;arm its poll&#34; even though the arming is still atomic and the poll program still static. Restore the accurate description and drop the generation claim.
- ℹ️ `docs/architecture.md:43` - &#34;A `working` crew state always defers without persisting anything, unconditionally&#34; is contradicted by the code and by the diff&#39;s own test. The dispatched-record block at bin/fm-watch.sh:1410-1436 runs before the crew-state check, so when a previously dispatched record has moved to `handled/` while the worker is now `working`, `pr_refresh_state_write ... resolved` persists the marker and returns 2 without the crew state ever being read - which is precisely what `test_branch_currency_dispatch_and_active_refusal` asserts (crew-state working, then `resolved\t&lt;head&gt;\t1\t001.msg` in `task-a.pr-refresh-state`). Reword the claim to scope it to the refusal path, since the state machine&#39;s ordering is deliberate.
- ℹ️ `bin/fm-watch.sh:1486` - This second `pr_refresh_state_read &#34;$marker&#34;` re-reads and re-validates a file already read at line 1410. Every way the first condition can fail (status is `resolved`, head mismatch) leaves `PR_REFRESH_STATUS`/`PR_REFRESH_HEAD`/`PR_REFRESH_ATTEMPT` already populated from that read, and the one case where they are not (the read itself failed) makes the second read fail identically. Recording the first read&#39;s outcome in a local and reusing it removes a duplicate file read plus a duplicate copy of the head/status validation. Non-functional.

🔧 Fix: bound branch-currency attempts, absorb opted-out behind polls

- ⚠️ `bin/fm-pr-poll.sh:82` - `behind` detection is read off `mergeStateStatus`, which GitHub collapses into ONE value with a fixed precedence (DIRTY &gt; UNKNOWN &gt; BLOCKED &gt; BEHIND &gt; UNSTABLE &gt; HAS_HOOKS &gt; CLEAN). A pull request that is behind its base AND has any required check pending/failing, or an outstanding required review, reports BLOCKED - not BEHIND - so `bin/fm-pr-poll.sh:82` emits nothing and the whole branch-currency lifecycle never fires. Concrete instance from the stated intent itself: PR #3847 &#39;is currently failing two checks and has been sitting that way since roughly 06:20&#39; - a PR in exactly that state reports BLOCKED, so the feature built to keep it current would not act on it. `BEHIND` additionally only ever appears when &#39;Require branches to be up to date before merging&#39; is enabled; without that setting a behind PR reads CLEAN. Note the asymmetry the same hunk already avoids for conflicts: line 80 uses the independent `mergeable` field (CONFLICTING) alongside `merge_state = DIRTY`, so conflict detection is NOT masked - only `behind` depends on the precedence-collapsed field alone. Flagging the remedy rather than the defect as what needs authorization: closing this means deciding what an additional behind-signal is (treating BLOCKED/UNSTABLE plus a base-comparison as behind, or querying `baseRefOid`/`compare` to determine behind-ness directly), which adds a new detection input beyond what this change does, and it is a product call on how aggressively an opted-in home should refresh a PR whose checks are red.
- ⚠️ `bin/fm-pr-poll.sh:64` - The pre-existing merged-PR detection lost its independence. Before this change the poll ran `gh pr view &#34;$url&#34; --json state -q .state`; it now runs a single widened query for `state,mergeStateStatus,mergeable,headRefOid` and, per `|| exit 0` on line 65, ANY failure of that one query silently yields no output. Because the poll is deliberately silent on error, a failure is indistinguishable from &#39;not merged&#39;: the task&#39;s merge poll never retires, `fm_merge_outcome_report` never runs, and the captain is never told the PR landed - forever, since every subsequent poll repeats the same failing query. Concrete trigger: `mergeStateStatus` is permission-sensitive (it is the GraphQL analogue of REST `mergeable_state`, surfaced only to identities with push access), so a poll running under a token with read-only access to the PR&#39;s repo - a fork, or a scoped PAT - fails the whole query where the old narrow `state` query succeeded. Same class for any gh build whose `pr view --json` field list predates `mergeStateStatus`, which errors client-side on the unknown field. Nothing covers this: the test fake at tests/fm-pr-check-security.test.sh:149 answers the widened query unconditionally, and `FM_TEST_GH_FAIL=1` fails both query shapes, so no test distinguishes them. Smallest correction is to keep merged detection on its own narrow `state` query (or fall back to it when the widened query fails), so a behind/conflict field problem cannot take merge detection down with it.
- ℹ️ `AGENTS.md:383` - The new sentence &#39;The merge poll also reactivates a finished GitHub PR worker when its head falls behind or conflicts with its base...&#39; states the behavior unconditionally, but the entire control is default-off and gated on `config/pr-refresh` (bin/fm-watch.sh:2034). AGENTS.md is firstmate&#39;s own always-loaded operating manual, and the surrounding &#39;PR ready, landing, and teardown&#39; block is instruction, not background - so in a home that never created the flag, firstmate is told to expect a reactivation that will never happen, and the correct absent-flag behavior (detected, absorbed into the triage log, never woken) is stated nowhere in this section. AGENTS.md:83 does document the flag, but in the config inventory far from this instruction. Add the opt-in condition to the sentence, matching docs/configuration.md:207 and docs/architecture.md:40.
- ℹ️ `AGENTS.md:117` - Two drifts in the one new `state/` inventory line. First, `&lt;id&gt;.pr-refresh-refused` is missing entirely: pr_refresh_refuse (bin/fm-watch.sh:1400) creates it as a durable per-task state file, bin/fm-teardown.sh:3347 cleans it alongside `&lt;id&gt;.pr-refresh-state`, and docs/architecture.md:39 names it as the Refused state&#39;s record - but the AGENTS.md inventory, which is the canonical index of what lives under state/, lists only the one file. Second, the entry it does add describes the file as a &#39;prepared/dispatched/resolved receipt&#39;, while pr_refresh_state_read (bin/fm-watch.sh:1368) accepts only `dispatched|resolved` and nothing anywhere writes `prepared` - a marker carrying that word would be rejected as corrupt. Add the missing entry and drop `prepared/`.

🔧 Fix: detect behind from base comparison, not merge state

- ⚠️ `bin/fm-pr-poll.sh:92` - `behind` now fires on any nonzero divergence, not on a merge-blocking condition. The replaced signal, `mergeStateStatus = BEHIND`, is only ever returned when the base branch has &#34;Require branches to be up to date before merging&#34; enabled - that is, only when being behind actually blocks the merge. `behind_by` from the compare endpoint carries no such qualifier. Concrete trace in an opted-in home: task-a&#39;s PR waits on the captain at head H; an unrelated PR merges into `main`; the next poll reads `behind_by=1` and emits `behind H`; `pr_refresh_dispatch` (bin/fm-watch.sh:1489) finds no marker for H, sets attempt=1, and sends the full reactivation instruction, so the finished worker re-enters its delivery contract, rebases, force-pushes, and re-runs every check on a PR GitHub would have merged cleanly. The attempt cap does not bound this: the rebase produces head H2, so the next base merge finds `PR_REFRESH_HEAD != head`, takes the `else attempt=1` path at bin/fm-watch.sh:1497, and starts a fresh ladder. In a repo with ordinary merge traffic that is one worker reactivation and one full CI cycle per base merge, indefinitely. The base-comparison mechanism itself was explicitly authorized in the prior round&#39;s instruction (&#34;Read the relationship between the branch and its base directly (e.g. ahead/behind commit counts against the base ref)&#34;), so this is not a challenge to that decision - what was silently dropped alongside it is the &#34;does being behind actually block this merge&#34; qualifier the old signal carried for free. Flagging the remedy rather than the defect as what needs authorization: closing this means deciding what the new qualifier is (reading the base branch&#39;s protection rules for the required-up-to-date setting, or a divergence threshold before dispatching), which adds a new detection input beyond what this change does and is a product call on how aggressively an opted-in home should churn a waiting PR.
- ⚠️ `bin/fm-pr-poll.sh:90` - The behind check adds a second, unconditional GitHub round trip per poll and downloads the entire compare payload to read one integer. `gh api repos/$owner/$repo/compare/$base...$head -q .behind_by` applies the jq filter client-side, so the full response body arrives first: up to 250 commit objects and up to 300 file entries with their patches. For a PR whose base has moved substantially in a large repo that is a multi-megabyte body fetched every CHECK_INTERVAL (default 300s, docs/configuration.md:891) for the life of the PR. It also runs in homes that never created `config/pr-refresh`, because the poll sidecar is task-static and cannot read the home config - the gate lives downstream at bin/fm-watch.sh:2037, so an opted-out home pays the request and discards the answer. The smallest correction is to bound the commits array with a query parameter (`repos/$owner/$repo/compare/$base...$head?per_page=1`); `behind_by`, `ahead_by`, `status`, and `total_commits` are computed independently of commit pagination, so the reported value is unchanged. The test fake at tests/fm-pr-check-security.test.sh:148 matches on `api:*/compare/*`, so the added query string is still covered.
- ℹ️ `docs/configuration.md:208` - &#34;a home that has not asked for it sees no behavior change at all&#34; is no longer accurate for an opted-out home. Two changes reach it regardless of the flag: the poll now issues an extra `gh api .../compare/...` request per cycle for every open non-conflicting PR (bin/fm-pr-poll.sh:90), and the watcher writes a triage-log line on every cycle for every behind or conflicting PR (bin/fm-watch.sh:2037-2039) where before this change the poll emitted nothing but `merged`. The wake claim in the preceding sentence is correct and verified - the opted-out path never reaches `fm_wake_append` or `wake` - so the accurate scope is that the captain&#39;s cadence is unchanged, not that nothing changes. Narrow the sentence to the wake and dispatch claim rather than asserting no observable change at all.
- ℹ️ `docs/configuration.md:907` - Two new watcher environment knobs are absent from the canonical env inventory. `FM_PR_REFRESH_ATTEMPT_MAX` (bin/fm-watch.sh:241) and `FM_PR_REFRESH_SEND_BIN` (bin/fm-watch.sh:100) are read only at watcher process start, exactly like every entry in that block, and the immediate sibling test override `FM_CREW_STATE_BIN` is listed on this line. `FM_PR_REFRESH_ATTEMPT_MAX` appears once in docs/architecture.md:47 as prose but nowhere an operator would look for a tunable; `FM_PR_REFRESH_SEND_BIN` is documented nowhere at all. Add both alongside the other watcher knobs, matching the existing `&lt;name&gt;=&lt;default&gt;   # &lt;meaning&gt;` form.

🔧 Fix: fire behind only when base enforces branch currency

- 🚨 `bin/fm-pr-poll.sh:91` - `gh api repos/$owner/$repo/branches/$base -q .protection.required_status_checks.strict` reads a field this endpoint does not return, so `behind` is now unreachable on every repository. Verified live against two repos: `gh api repos/cli/cli/branches/trunk` returns `&#34;protection&#34;:{&#34;enabled&#34;:true,&#34;required_status_checks&#34;:{&#34;enforcement_level&#34;:&#34;non_admins&#34;,&#34;contexts&#34;:[...],&#34;checks&#34;:[...]}}` (no `strict` key), and `gh api repos/kunchenguid/firstmate/branches/main` returns the same legacy shape. Both queries for `.strict` printed an empty string with exit 0, so `|| exit 0` on line 92 does not fire and `[ &#34;$strict&#34; = true ]` on line 93 always fails, exiting before the compare on line 94. cli/cli trunk genuinely enforces required status checks and still yields nothing, so this is not a fixture of unprotected bases: `strict` is only surfaced by `GET /repos/{o}/{r}/branches/{b}/protection` (admin-gated, and 404s on a repository governed by rulesets rather than classic protection - firstmate&#39;s own `main` is exactly that case) or, for rulesets, by `GET /repos/{o}/{r}/rules/branches/{b}` as a `required_status_checks` rule with `parameters.strict_required_status_checks_policy`. Concrete trace in an opted-in home: task-a&#39;s PR waits on the captain, an unrelated commit lands on a base that DOES require branches to be up to date, the next poll reads `strict=&#34;&#34;` and exits silently, `pr_refresh_dispatch` is never called, no triage line is written, and the branch stays stale forever - the exact failure the change exists to prevent. This also contradicts the authoritative intent, which states the PR&#39;s &#34;whole subject is keeping a pull request that waits on the captain from going stale&#34;: with `behind` unreachable, only the `conflict` path survives and the stated subject is not delivered. The remedy, not the defect, is what needs authorization: neither replacement endpoint covers both classic protection and rulesets, and `/branches/{b}/protection` needs admin, so closing this means deciding which source(s) to read (possibly both, or accepting an unreadable-strictness fallback policy) and what a home whose token cannot read either should do - a new detection input and a product call on fail-open versus fail-closed, beyond what this change currently does. The opted-out cost also grows: every poll in a home without `config/pr-refresh` now pays this request for a signal that can never fire.
- ⚠️ `tests/fm-pr-check-security.test.sh:154` - The gh fake&#39;s new `api:*/branches/*` arm prints `${FM_TEST_GH_STRICT:-true}`, a bare `true` at a path where the real API returns a `protection` object with no `strict` key at all (verified live above). The fixture therefore pins a response shape GitHub does not produce, and it defaults to the value that makes dispatch fire. `test_branch_currency_blocked_pr_still_dispatches` (line 1315) passes `FM_TEST_GH_STRICT=true` and asserts a dispatch that cannot happen against the real forge, so the suite reports green on behavior that is dead in production. The new negative case `test_branch_currency_lenient_base_is_not_dispatched` also passes vacuously: with the real API, no dispatch occurs regardless of the `FM_TEST_GH_STRICT=false` it sets, so it cannot distinguish the fix from the bug. Correcting the fixture requires knowing the real shape and endpoint, which is blocked on the same decision as the source finding; once that is settled, the fake should reproduce the actual JSON of whichever endpoint is chosen (including the absent-`strict` case) rather than a synthetic scalar.

🔧 Fix: drop branch-protection read, keep divergence as behind signal
5 issues (1 warning, 4 infos) still open:

- ⚠️ `bin/fm-watch.sh:1497` - The attempt bound does not cover the head-moves case, and dropping the branch-protection qualifier made that case apply to every base. Concrete trace, opted-in home, base receives commits faster than the pipeline finishes: poll 1 sees `behind H0`, marker absent, line 1497 sets attempt=1, dispatch fires, worker rebases to H1 and re-runs the whole no-mistakes contract. Poll 2 sees `behind H1`; the marker holds head=H0, so the guard at 1489 (`[ &#34;$PR_REFRESH_HEAD&#34; = &#34;$head&#34; ]`) is false, line 1497 sets attempt=1 again, and the marker is overwritten with H1. `FM_PR_REFRESH_ATTEMPT_MAX` (line 243, checked at 1492) is only reachable when the head does NOT move, which docs/architecture.md:49 states explicitly (&#34;a worker that keeps acknowledging the instruction without moving the head&#34;). So a pull request parked on the captain re-validates indefinitely, one full pipeline run per base commit, with no ceiling and one wake each time. Before cfa3a67 the `required_status_checks.strict` read confined this to bases that actually enforce currency; the captain settled that the sole condition is divergence, and that decision is not what this finding contests - the gap is that the bound introduced in c9215a6 to stop exactly this class of runaway does not reach the reachable half of it. The remedy, not the defect, is what needs authorization: an honest fix adds new durable state (a per-PR refresh budget over a time window, or an attempt count that survives a head change) rather than correcting what the current code does, and it is a product call how many rebases a waiting PR may earn per day. Mitigating: the control is default-off behind config/pr-refresh, and a worker mid-rebase reports `working`, which defers at line 1436 instead of stacking a second dispatch.
- ℹ️ `tests/fm-pr-check-security.test.sh:158` - The replacement for the deleted `strict` arm overshoots: `case &#34;${1:-}&#34; in api) exit 1 ;; esac` fails EVERY `gh api` call except graphql and `*/compare/*`. `bin/fm-pr-merge.sh:451` calls `gh api --paginate &#34;repos/$PR_OWNER/$PR_REPO/rules/branches/$branch_path&#34; --jq ...`, and that endpoint is one GitHub really answers: a branch with no rulesets returns `200 []`, so the `--jq` filter emits nothing and gh exits 0. The old fake reproduced exactly that (no arm matched `&#34; api --paginate repos/o/r/rules/branches/main --jq ... &#34;`, the fake fell through and exited 0 with empty stdout), giving `FM_PR_GITHUB_QUEUE_STATUS=none` at fm-pr-merge.sh:484. The new arm gives `unreadable` (early `return 0` at line 456). fm-pr-merge.sh IS exercised from this file via `run_merge_entry` (lines 463, 475, 506, 551, 587, 636, 2073, 2293, 2326), so the fixture change is live; no assertion flips today only because `github_read_queue_method` is reached solely from `github_report_queue_rules` (line 549) via `github_report_unmerged_outcome` (line 605), and the graphql arm at test line 139 answers `merged=true`, keeping every one of those calls off the unmerged path. Net effect: the audit asked for arms that fabricate shapes GitHub does not produce, and this one now fabricates a failure where GitHub succeeds. Narrow it to the endpoints the poll actually queries (e.g. `api:*/branches/*) exit 1` for the protection read specifically), or give the rules endpoint an arm that prints nothing and exits 0.
- ℹ️ `bin/fm-pr-check.sh:2` - The rewritten contract header says the script stores &#34;one validated canonical pr=&lt;url&gt;, its worker generation, and the forge&#39;s exact pr_head=&lt;sha&gt;&#34;. It does not store a worker generation: lines 114-115 append only `pr=%s` and, conditionally, `pr_head=%s`, and the filter at line 110 (`pr=*|pr_head=*`) names those two keys alone. `spawn_gen` is written by the spawn path into `&lt;id&gt;.meta` and merely survives this script&#39;s rewrite; `bin/fm-watch.sh:1444` reads it straight from the meta file, never from anything fm-pr-check.sh recorded. A reader trusting this header would look for a generation this script guarantees and find none. Restore the accurate wording (pr= plus optional pr_head=).
- ℹ️ `AGENTS.md:117` - The new state-inventory entry describes `&lt;id&gt;.pr-refresh-state` as a &#34;prepared/dispatched/resolved receipt&#34;, but `prepared` is never written and is actively rejected on read: `pr_refresh_state_write` (bin/fm-watch.sh:1389) is only ever called with `dispatched` or `resolved` (lines 1462, 1503), and `pr_refresh_state_read` (line 1405) fails any status outside `case ... in dispatched|resolved`. The same entry is also the only new one, while the Refused half of the lifecycle has its own file: `pr_refresh_refuse` writes `$STATE/$id.pr-refresh-refused` (line 1441), docs/architecture.md:44 names it as where the Refused state lives, and bin/fm-teardown.sh:3348 removes it alongside `.pr-refresh-state`. Correct the status list to dispatched/resolved and add the `&lt;id&gt;.pr-refresh-refused` line so the canonical state inventory covers both files this change creates.
- ℹ️ `AGENTS.md:383` - The new operating line states unconditionally that &#34;The merge poll also reactivates a finished GitHub PR worker when its head falls behind or conflicts with its base&#34;. That behavior is default-off: bin/fm-watch.sh:2037 returns to the triage log without ever calling `pr_refresh_dispatch` unless `$CONFIG/pr-refresh` exists, which AGENTS.md:83 itself documents as an opt-in presence flag, and docs/configuration.md:207 confirms an absent flag means no dispatch, no state, and no wake. Every other opt-in control in this file names its flag where the behavior is described (see the config/turnend-churn-absorb treatment). As written, an agent reading the PR-handoff section will tell the captain a waiting PR is kept current automatically in a home that never created the flag. Add the `config/pr-refresh` condition to the sentence.

🔧 Fix: bind refresh budget to the PR, block when spent

- ⚠️ `bin/fm-watch.sh:1495` - The per-PR budget is spent by successful refreshes and never reset, so a healthy long-parked PR permanently disables its own branch-currency dispatch and is reported with a label that misdescribes what happened. Concrete trace, opted-in home, FM_PR_REFRESH_ATTEMPT_MAX at its default 3: (1) base moves, poll emits `behind H0`, no marker, line 1509 sets attempt=1, dispatch, marker becomes `dispatched H0 1 rec1`. (2) The worker rebases and the PR is current at H1. `bin/fm-pr-poll.sh:93` only prints when `behind_by &gt; 0`, so `out` is empty, the whole `if [ -n &#34;$out&#34; ]` block at bin/fm-watch.sh:2024 is skipped, and nothing clears `$STATE/$id.pr-refresh-state`. Grep confirms the marker is written only by `pr_refresh_state_write` inside `pr_refresh_dispatch` and removed only by `bin/fm-teardown.sh:3347`. (3) Base moves again, `behind H1`: the guard at 1421 is skipped because the head differs, line 1495 reads the surviving marker, attempt=2, dispatch, success again. (4) Third cycle, attempt=3, dispatch, success again. (5) Fourth cycle, attempt=4 &gt; 3: line 1498 writes the terminal `exhausted` status, prints `branch-refresh-blocked pr=... reason=attempts-exhausted attempts=3` as an actionable wake, and every later poll hits the early guard at 1414 and repeats that same line into triage forever. Nothing was exhausted: all three attempts were acknowledged and effective. The feature is now off for the rest of that PR&#39;s life, on exactly the case it exists for (a PR waiting on the captain while a busy base advances). The prior round&#39;s finding asked only that a moving base not RESET the count; binding it to the PR was correct, but no rule was added for what SPENDS it, and `resolved` alone cannot separate &#39;acknowledged and effective&#39; from &#39;acknowledged and ineffective&#39; because a successful refresh and a runaway base both present as a moved head. The remedy, not the defect, is what needs authorization: the honest fix is a reset rule keyed on observing the PR current, which requires `bin/fm-pr-poll.sh` to emit a distinguishable current signal (its silence today also covers every `gh` failure, `bin/fm-pr-poll.sh:64`), so it extends the change rather than correcting it; alternatively the captain may want the lifetime cap kept as-is with an honest reason string instead of `attempts-exhausted`. Mitigating: default-off behind config/pr-refresh, and the terminal state is quiet after the single wake.
- ℹ️ `bin/fm-watch.sh:241` - The new terminal `exhausted` status and the re-keyed budget left four contract statements factually wrong, two of which now contradict each other inside the same document section. (a) bin/fm-watch.sh:241 still documents the knob as &#39;Branch-currency attempts ONE HEAD may consume before pr_refresh_dispatch stops re-instructing a worker that keeps acknowledging without moving it&#39; - after this round the count is per pull request and is consumed by a moving head too, which is the whole point of the change. (b) bin/fm-watch.sh:1350 header says &#39;Four-state branch-currency dispatch (Observed/Refused/Dispatched/Resolved)&#39; and bin/fm-watch.sh:1352&#39;s signature comment says `# &lt;path&gt; &lt;dispatched|resolved&gt; &lt;head&gt; &lt;attempt&gt; &lt;record&gt;`, while `pr_refresh_state_read` (1367) now accepts `exhausted` and `pr_refresh_state_write` is called with it at 1498. (c) docs/architecture.md:44 enumerates the same four states and defines Resolved as reached when &#39;the head moved&#39; - head-move is no longer a resolution, it is a budget-consuming attempt - and docs/architecture.md:49, in the same section, correctly describes the terminal `exhausted` status, so the section contradicts itself. (d) docs/configuration.md:207 repeats &#39;the four-state dispatch lifecycle (Observed, Refused, Dispatched, Resolved)&#39;, and the new &#39;Branch-currency dispatch&#39; section never names `FM_PR_REFRESH_ATTEMPT_MAX` even though sibling tunables are enumerated in that file&#39;s env block (docs/configuration.md:940-941) - an operator cannot discover the ceiling that permanently ends dispatch for a PR. Correct the four enumerations to five states and re-word the line-241 comment to say the budget belongs to the pull request.

🔧 Fix: one refresh dispatch per head, blocker when unmoved
1 warning still open:

- ⚠️ `bin/fm-watch.sh:1429` - The new one-dispatch-per-head guard treats an inbox acknowledgement as proof the worker refused to refresh, and wakes the captain with a blocker while the worker is provably still doing the work. Concrete trace, opted-in home: (1) poll reports `behind H0`, pr_refresh_dispatch sends the instruction, marker = `dispatched H0 001.msg`. (2) The worker reads the instruction and moves 001.msg to handled/ - bin/fm-task-inbox-lib.sh:27 states the `mv` into handled/ IS the acknowledgement, not evidence the branch moved, and the ladder at fm-task-inbox-lib.sh:47 re-rings and finally escalates any record left unhandled past FM_TASK_INBOX_GRACE_SECS, so a compliant worker must acknowledge before finishing a multi-minute no-mistakes re-run. (3) 300s later (CHECK_INTERVAL, bin/fm-watch.sh:172) the poll still reports `behind H0` because the rebase has not been pushed yet. pr_refresh_dispatch takes the marker branch at line 1411, finds record_state=resolved, writes the terminal `resolved` marker and returns 1 at line 1431, which the caller at bin/fm-watch.sh:2030-2036 routes to wake(): `branch-refresh-blocked pr=... reason=acknowledged-without-refresh`. The label is false - the worker acknowledged and is mid-refresh - and the head then never earns another dispatch. The `working` crew-state defer at lines 1445-1452 is unreachable on this path because the marker branch short-circuits 30 lines above it, which also makes docs/architecture.md:46 (&#34;A `working` crew state always defers without persisting anything, unconditionally&#34;) false for the changed code. This round&#39;s own test pins the wrong outcome as correct: tests/fm-pr-check-security.test.sh:925-944 sets the fake crew state to `state: working - validating (running)`, then asserts the blocker appears in the woken stdout; the prior version of that same fixture asserted a quiet deferral, so the false wake is new in this round. Smallest honest remedy is to consult crew state (defer quietly on `working`, and only declare the blocker for a worker that acknowledged and then went idle at an unmoved head) before line 1429, but that changes the blocker semantics the captain prescribed verbatim in round 6 (&#34;the record is resolved (worker acknowledged the instruction without moving the head) ... report a blocker&#34;), so the remedy, not the defect, is what needs authorization. Mitigating: default-off behind config/pr-refresh, and later polls at the same head stay quiet in triage.

🔧 Fix: drop refresh blocker, defer quietly on dispatched head
3 infos still open:

- ℹ️ `docs/architecture.md:44` - After cfb415c dropped the blocker branch, the marker&#39;s status field is written but never consumed, so the documented four-state machine has three effective states. `pr_refresh_state_read` (bin/fm-watch.sh:1365) validates `PR_REFRESH_STATUS` against `dispatched|resolved` and then nothing reads it: the only consumer of the marker branch is `[ &#34;$PR_REFRESH_HEAD&#34; = &#34;$head&#34; ]` at bin/fm-watch.sh:1411, and both statuses fall into the same `pending|resolved` defer at 1413-1418. Two docs still describe Resolved as a state with its own behavior: docs/architecture.md:44 defines it as &#34;the worker had already acknowledged that record by the time the dispatch was confirmed&#34;, and docs/configuration.md:207 enumerates &#34;the four-state dispatch lifecycle (Observed, Refused, Dispatched, Resolved)&#34;, as does the function header at bin/fm-watch.sh:1347. Separately, docs/architecture.md:51 still explains the `generation-unconfirmed` refusal by saying &#34;the mid-check compare above never fires&#34; - that spawn_gen mid-check compare was removed in an earlier round, and the immediately preceding line 48 now states the opposite (&#34;nothing here persists that value across polls or compares it to an earlier attempt&#34;), so the paragraph contradicts itself about a mechanism the code no longer contains. Remedy is mechanical and non-user-visible: reword the three enumerations to the states the code actually distinguishes, and drop the dangling &#34;mid-check compare above&#34; clause.
- ℹ️ `bin/fm-watch.sh:1394` - `pr_refresh_refuse` remembers exactly one (head, reason) pair, so two refusal reasons that alternate at the same head wake the captain on every poll instead of once. Concrete path: `$FM_CREW_STATE_BIN` fails transiently on one poll, so `state_line` is empty at bin/fm-watch.sh:1431, `state` falls through the `*)` arm to `unknown` at 1440, and the function refuses `task-state-unknown` and stores that pair. The next poll&#39;s crew-state read succeeds and reports `done`, the status-log mtime check at 1450-1465 fires, and the function refuses `generation-unconfirmed`; the stored reason differs, so line 1400 rewrites the marker and returns 1, which the caller routes through `wake()` at bin/fm-watch.sh:2022-2026. Flip back on the following poll and it wakes again, indefinitely, at one full captain turn per CHECK_INTERVAL - the exact repeat-wake noise the dedup at 1394-1399 exists to prevent, and the header comment&#39;s &#34;an unchanged (head, reason) never wakes twice&#34; is true only for a reason that never alternates. Probability is low (it needs the crew-state read to flap), and the head-move path is unaffected. The remedy, not the defect, is what needs authorization: honestly fixing it means remembering a set of already-woken reasons per head rather than one pair, which adds durable state beyond what this change currently keeps, so the alternative of deduping on head alone (one wake per head whatever the reason) is a product call about how much refusal detail the captain should see.
- ℹ️ `bin/fm-pr-poll.sh:91` - The branch-divergence lookup runs for every home, including the majority that never create `config/pr-refresh`, because the opt-out gate lives in the watcher (bin/fm-watch.sh:2012-2015) rather than in the poll. Every CHECK_INTERVAL, each armed poll for an open non-conflicting PR now spends a second forge call (`gh api repos/.../compare/$base...$head`) whose result the watcher discards into the triage log and drops. This is an acknowledged tradeoff - docs/architecture.md:43 states &#34;Detection costs one commit comparison per poll for an open PR that is not already conflicting&#34; - but docs/configuration.md:207 tells a default home that with the flag absent &#34;the captain&#39;s cadence is exactly what it was before this control existed&#34;, which is true of wakes and not of forge traffic. Noting it rather than asking for a change: moving the gate into the poll would make a deliberately task-static, config-free script read home configuration, which is a larger design change than the cost justifies.

🔧 Fix: dedup branch-refresh refusals on head alone
3 infos still open:

- ℹ️ `bin/fm-pr-check.sh:2` - The rewritten header claims the script stores &#34;one validated canonical pr=&lt;url&gt;, its worker generation, and the forge&#39;s exact pr_head=&lt;sha&gt;&#34;, but nothing in bin/fm-pr-check.sh writes any generation: the meta rewrite at bin/fm-pr-check.sh:107-118 strips and re-adds only `pr=` and `pr_head=`, and `grep -n &#39;spawn_gen&#39; bin/fm-pr-check.sh` matches the comment alone. The generation this feature actually uses is read live from the task meta by pr_refresh_dispatch (bin/fm-watch.sh:1429) and handed to fm-send as a delivery-time guard; it is deliberately never persisted at PR-check time, which is the property docs/architecture.md relies on (&#34;nothing here persists that value across polls&#34;). A reader trusting this header would look for a stored generation that does not exist and could conclude the arming step already pinned an incarnation. Remedy is to drop the &#34;its worker generation&#34; clause.
- ℹ️ `AGENTS.md:117` - The state inventory entry describes `&lt;id&gt;.pr-refresh-state` as a &#34;prepared/dispatched/resolved receipt&#34;, but `prepared` is not a status this change ever writes or accepts: pr_refresh_state_write is only ever called with the result of the `pending)-&gt;dispatched` / `resolved)` mapping at bin/fm-watch.sh:1503-1506, and pr_refresh_state_read rejects anything outside `dispatched|resolved` at bin/fm-watch.sh:1365, so a marker reading `prepared` would be discarded as corrupt and silently re-dispatched. The same entry omits `&lt;id&gt;.pr-refresh-refused` entirely, even though this change creates it (bin/fm-watch.sh:1399) and adds it to teardown&#39;s removal list (bin/fm-teardown.sh:3347) alongside `&lt;id&gt;.pr-refresh-state`; every other per-task marker in that block is listed, so an operator reading the inventory to audit leftover state after a crash will not know to look for it. Remedy is mechanical: correct the status list to dispatched/resolved and add the missing marker line.
- ℹ️ `AGENTS.md:383` - The added sentence promises &#34;every non-active refusal surfaces for supervision&#34;, which the head-only refusal dedup introduced in a69c91a makes false. Concrete path: at head H0 a poll refuses `unsupported-mode` and writes H0 into $id.pr-refresh-refused (bin/fm-watch.sh:1399), waking the captain. On a later poll at the same H0 the crew-state read reports `blocked`, so pr_refresh_dispatch refuses `task-state-blocked` (bin/fm-watch.sh:1443) - a genuinely different, genuinely new refusal - but the dedup at bin/fm-watch.sh:1394-1400 matches on head alone, returns 2, and the caller routes it to triage_log rather than wake (bin/fm-watch.sh:2018-2021). Only the first refusal per head ever reaches the captain; the rest are visible only to someone reading .watch-triage.log. That behavior is the explicitly authorized round-8 tradeoff and should stay - the sentence is what needs to match it, e.g. &#34;the first refusal at each head surfaces for supervision and later refusals at that head defer into triage&#34;. Remedy is a reword, no behavior change.
</details>

<details>

✅ No issues found.
- <code>`bash bin/fm-test-run.sh tests/fm-pr-check-security.test.sh` — 40/40 ok, exit 0 (all 13 `test_branch_currency_*` cases: dispatch/active refusal, generation race, stale status log, relaunch, remote-secondmate refusal, restart before state recorded, refusal dedup on head alone, blocked PR, no branch-protection read, moved head each time, acknowledged head quiet, normal refresh never reaches captain, opt-out by default)</code>
- <code>`bash bin/fm-test-run.sh tests/fm-send-inbox.test.sh` — 6/6 ok, exit 0 (inbox delivery plane touched by the diff)</code>
- <code>Manual end-to-end lifecycle run: `bash ~/.no-mistakes/evidence/01M1XVNPHKWC8D7NFYYCJX47TD/branch-currency-lifecycle.sh` — drives four watcher polls (behind at head A, acknowledged-unmoved, moved to head B, current at head C) plus an opted-out home, capturing captain wake output, the triage log, the worker inbox instruction, and the per-head instruction count</code>
- <code>`git status --porcelain` — worktree clean, no transient artifacts left behind</code>
</details>

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>


dbeihl-utilicast:
Closing: this pull request targets the upstream repository. Our standing rule is that we never publish changes there; the work belongs on our own fork, dbeihl-utilicast/firstmate. The cause was a missing default repository on our machines, now fixed, and a guard so a wrong target refuses is filed. Closed on the captain instruction, 2026-09-15. Nothing was merged upstream.

PR #3962, opened 2026-09-08
fix(bin): surface a process-event state root that stops being private

dbeihl-utilicast:
## Intent

The captain asked whether we are using the local Qwen model yet. The honest answer is that it was proved this evening and has been idle since, and one change it produced and passed is sitting on that machine as a local branch that nobody has done anything with.

That is the gap to close. He did not want a proof, he wanted it building things.

Two parts.

First, land the one that already passed. Case firstmate-3927 reached ACCEPT at commit 7912898f0862a2e72b54e1f45e48e055c3b03a92, on branch qwen/firstmate-3927-20260907T230721Z-2569952 with its bundle beside it, at ~/qwen-build-pipeline/runs/firstmate-3927-20260907T230721Z-2569952. Read the change yourself before you move it. An independent judge passing it is evidence, not a reason to skip reading it, and the whole point of that judge is that the model's own account of its work carries no weight. If it is genuinely good, take it through the ordinary path to a pull request and say plainly in the body that a local model wrote it and what judged it. If it is not, say why, and that is a more useful result than the accept was.

Second, put more work through it. There are ten firstmate defects sitting queued that nobody is working on, and this is exactly the machine to grind them overnight. Read the backlog, pick the ones whose shape matches what actually survived: exact file scope, a focused existing suite, and a check that can prove the change is real rather than plausible. Build a case for each, using the existing template and cases as the pattern, and run them.

Do not stretch it. Two of the four trials were rejected on both attempts because they were genuinely harder than this path handles today, and that is the pipeline working. A case you cannot write a strong judge for is a case that does not go in, because a weak judge turns this from useful into dangerous.

The captain later decided, on the review finding "reconcile-swallow-still-silent" against the arm-time fix for issue 3929: this is in scope rather than an expansion, because issue 3929's reported failure is the silent outage of nothing starting for two days with idle lanes as the only symptom, and refusing at arm time closes only one door into that room while the finding shows the room is still reachable through another door. Fixing that completes the accepted intent rather than adding a new guarantee. The captain wants the swallowed failure made visible as one durable, visible record, not a stream of them on every poll -- no general alerting mechanism, and not every reconcile error turned into noise, only the fact that the state root has stopped being private. The same swallowing pattern appears in the board script as well as the watcher, and the captain asked to judge whether one fix covers both, doing it once if so, or explaining why not rather than fixing one and leaving its twin. The captain required it be provable with a test that arms while the directory is private, relaxes it afterwards, runs the reconcile, and shows that the failure now surfaces where it previously vanished.

## Provenance

The initial candidate for this change (the arm-time refusal, commit `06c945db`) was built by a local Qwen Coder model on `qwen-build-pipeline` (fm-spark), as case firstmate-3929. It was judged by that pipeline's independent verdict step — not the model's own account of its work — before I read the diff by hand, verified it, and fixed a leaked stdout line before landing it. Everything from the second commit onward (the surfacing fix and four review-driven hardening rounds below) is normal no-mistakes review against that starting point, same as any other change.

This change took four review rounds, and every one of them found that the previous round's fix did not cover its own case: symlinks, then an empty directory, then a non-empty directory, then a diagnostic left unreachable by an earlier round of the same effort. The change that ships is better for it. The pattern is also the reason a fifth round would have been a rewrite rather than more fixing, so it stopped here — two residual findings are filed as issues #3995 and #3996 rather than fixed in this diff.

## What Changed

- `bin/fm-procevent-when.sh` now resolves and validates the process-event state root during `arm`, so arming against a non-private root dies before any spec, trust, or registration record is written instead of half-arming a source.
- `bin/fm-procevent.sh` records the first swallowed privacy refusal once at `$FM_HOME/.procevent-state-insecure` (staged via `mktemp` + rename, keeping the original detection on repeats) and clears it plus its suppressor on the first command that finds the root private again; `bin/fm-watch.sh` turns that record into exactly one `check: procevent-state-insecure` wake, latching `.procevent-state-insecure-surfaced` only after the wake is delivered, and `bin/fm-bearings-board.sh` names the same record when a source is not listening after reconcile. New `fm_marker_settled` / `fm_marker_clear` helpers in `bin/fm-pr-lib.sh` back both marker paths, rejecting symlinks and directories and moving a non-empty directory aside rather than deleting it.
- `tests/fm-procevent-when.test.sh` drives the real `when` adapter and watcher to prove the arm-time refusal writes nothing, that relaxing the root after arming leaves one durable record that clears on repair, that the wake is delivered once and not re-reported (anchored by a queued process-event wake proving the later cycle ran past the check), and that planted symlinks and directories at either marker path cannot silence or redirect it; `tests/fm-gitignore-config.test.sh` pins that the markers do not dirty the checkout, matching the new `.gitignore` entry. `AGENTS.md`, `docs/configuration.md`, `docs/verification/process-event-sources.md`, and the `process-event-sources` skill document the markers and the new wake.

## Risk Assessment

⚠️ Medium: The core mechanism (durable one-shot record, directory/symlink-hardened markers, delivery-latched suppressor) is correct and genuinely test-pinned, but the board half of the intent's two-caller requirement sits on a branch the insecure-root case cannot reach, so that caller still reports the wrong cause.

## Testing

Both targeted suites pass (`tests/fm-procevent-when.test.sh`, `tests/fm-gitignore-config.test.sh`), and beyond that I drove the intent's exact scenario by hand against both the base tree and this change: on base the relaxed state root produces no record and no watcher wake (the silent outage of #3929), while on this change it produces one durable record naming the root, exactly one delivered `check: procevent-state-insecure` wake, no repeat on a later cycle anchored by a separate process-event wake, and automatic clearing once the root is private again. I also reproduced the round-5 non-empty-directory marker silence on the pre-fix commit and showed the same case now refused with the planted directory's contents preserved. No UI surface is involved, so the evidence is CLI transcripts rather than screenshots; the board-side diagnostic was left untested because review round 5 declined it as unreachable.

<details>
<summary>Evidence: Operator transcript, base commit 72bfdd0 (failure vanishes)</summary>

Source: [Operator transcript, base commit 72bfdd0 (failure vanishes)](https://github.com/dbeihl-utilicast/firstmate/blob/7671a3b114c022dc32aebfff56506a9aa074bb10/.no-mistakes/evidence/fm/qwen-overnight-batch/before-fix-transcript.txt)

<code>=== 3. the watcher&#39;s own reconcile call, verbatim: exit and output discarded&#10;operator sees: (nothing - that is the swallow)&#10;&#10;=== 4. is there any durable record of the refusal?&#10;NONE - the failure vanished&#10;&#10;=== 5. what the captain actually sees from the real watcher (one poll window)&#10;no wake about the state root at all</code>

```text
tree under test: base commit 72bfdd0 (before the #3929 fix)

=== 1. arm a condition->action watch while state/ is private (700)
    armed: when-demo-watch
    starts on the watcher's next cycle; or run: bin/fm-procevent.sh reconcile
    reminder: deterministic, safe, reversible actions only; judgment and destructive actions stay on the wake-and-decide path

=== 2. an operator relaxes the state root (umask, chmod -R, a helper tool)
    drwxrwxr-x@ 4 davidsair  staff  128 Sep  8 08:52 /var/folders/j5/kgdthy6n37j7dgv0r7qr3qd80000gn/T//fm-3929-demo.CkLBVJ/state

=== 3. the watcher's own reconcile call, verbatim: exit and output discarded
    operator sees: (nothing - that is the swallow)

=== 4. is there any durable record of the refusal?
    NONE - the failure vanished

=== 5. what the captain actually sees from the real watcher (one poll window)
    no wake about the state root at all

=== 6. does it become a stream? second watcher window, same broken root
    anchor proving the cycle ran past the check: 1
    repeat insecure-root wakes this window: 0

=== 7. operator restores private permissions and reconciles
    reconcile: ok
    nothing to clear - there never was a record
```
</details>
<details>
<summary>Evidence: Operator transcript, this change 6b937c8 (surfaces once, clears itself)</summary>

Source: [Operator transcript, this change 6b937c8 (surfaces once, clears itself)](https://github.com/dbeihl-utilicast/firstmate/blob/7671a3b114c022dc32aebfff56506a9aa074bb10/.no-mistakes/evidence/fm/qwen-overnight-batch/after-fix-transcript.txt)

<code>=== 4. is there any durable record of the refusal?&#10;.../.procevent-state-insecure:&#10;fm-procevent-state-insecure-v1&#10;detected=1788872040&#10;state=.../state&#10;&#10;=== 5. what the captain actually sees from the real watcher (one poll window)&#10;check: procevent-state-insecure&#10;&#10;=== 6. does it become a stream? second watcher window, same broken root&#10;anchor proving the cycle ran past the check: 1&#10;repeat insecure-root wakes this window: 0&#10;&#10;=== 7. operator restores private permissions and reconciles&#10;reconcile: ok&#10;record cleared automatically</code>

```text
tree under test: 6b937c8 (this change)

=== 1. arm a condition->action watch while state/ is private (700)
    armed: when-demo-watch
    starts on the watcher's next cycle; or run: bin/fm-procevent.sh reconcile
    reminder: deterministic, safe, reversible actions only; judgment and destructive actions stay on the wake-and-decide path

=== 2. an operator relaxes the state root (umask, chmod -R, a helper tool)
    drwxrwxr-x@ 4 davidsair  staff  128 Sep  8 08:53 /var/folders/j5/kgdthy6n37j7dgv0r7qr3qd80000gn/T//fm-3929-demo.Pcdvov/state

=== 3. the watcher's own reconcile call, verbatim: exit and output discarded
    operator sees: (nothing - that is the swallow)

=== 4. is there any durable record of the refusal?
    /var/folders/j5/kgdthy6n37j7dgv0r7qr3qd80000gn/T//fm-3929-demo.Pcdvov/.procevent-state-insecure:
      fm-procevent-state-insecure-v1
      detected=1788872040
      state=/var/folders/j5/kgdthy6n37j7dgv0r7qr3qd80000gn/T//fm-3929-demo.Pcdvov/state

=== 5. what the captain actually sees from the real watcher (one poll window)
    check: procevent-state-insecure

=== 6. does it become a stream? second watcher window, same broken root
    anchor proving the cycle ran past the check: 1
    repeat insecure-root wakes this window: 0

=== 7. operator restores private permissions and reconciles
    reconcile: ok
    record cleared automatically
```
</details>
<details>
<summary>Evidence: Non-empty directory at both marker paths, pre-fix bfb9cf6 (bug reproduces)</summary>

Source: [Non-empty directory at both marker paths, pre-fix bfb9cf6 (bug reproduces)](https://github.com/dbeihl-utilicast/firstmate/blob/7671a3b114c022dc32aebfff56506a9aa074bb10/.no-mistakes/evidence/fm/qwen-overnight-batch/nonempty-dir-marker-before.txt)

<code>=== relax the state root, then run the swallowed reconcile&#10;NO record - the planted directory was read as a valid one&#10;&#10;=== one watcher window: does the captain hear about the broken root?&#10;anchor proving the cycle ran past the check: 1&#10;SILENT - the planted directory latched the wake away</code>

```text
tree under test: bfb9cf6 (before "reject any directory as a settled marker")

=== plant a non-empty directory at BOTH marker paths
    .procevent-state-insecure
    .procevent-state-insecure-surfaced

=== relax the state root, then run the swallowed reconcile
    NO record - the planted directory was read as a valid one

=== one watcher window: does the captain hear about the broken root?
    anchor proving the cycle ran past the check: 1
    SILENT - the planted directory latched the wake away

=== was the planted directory content preserved?
    left in place untouched

home: /var/folders/j5/kgdthy6n37j7dgv0r7qr3qd80000gn/T//fm-3929-dirmarker.yH4mwE
```
</details>
<details>
<summary>Evidence: Non-empty directory at both marker paths, this change 6b937c8 (refused, contents preserved)</summary>

Source: [Non-empty directory at both marker paths, this change 6b937c8 (refused, contents preserved)](https://github.com/dbeihl-utilicast/firstmate/blob/7671a3b114c022dc32aebfff56506a9aa074bb10/.no-mistakes/evidence/fm/qwen-overnight-batch/nonempty-dir-marker-after.txt)

<code>=== relax the state root, then run the swallowed reconcile&#10;durable record written:&#10;fm-procevent-state-insecure-v1&#10;detected=1788873198&#10;state=.../state&#10;&#10;=== one watcher window: does the captain hear about the broken root?&#10;check: procevent-state-insecure&#10;&#10;=== was the planted directory content preserved?&#10;moved aside to .procevent-state-insecure.displaced-68262-30150, payload intact: do not lose me</code>

```text
tree under test: 6b937c8 (this change)

=== plant a non-empty directory at BOTH marker paths
    .procevent-state-insecure
    .procevent-state-insecure-surfaced

=== relax the state root, then run the swallowed reconcile
    durable record written:
      fm-procevent-state-insecure-v1
      detected=1788873198
      state=/var/folders/j5/kgdthy6n37j7dgv0r7qr3qd80000gn/T//fm-3929-dirmarker.ZIZvMS/state

=== one watcher window: does the captain hear about the broken root?
    anchor proving the cycle ran past the check: 0
    check: procevent-state-insecure

=== was the planted directory content preserved?
    moved aside to .procevent-state-insecure.displaced-68262-30150, payload intact: do not lose me

home: /var/folders/j5/kgdthy6n37j7dgv0r7qr3qd80000gn/T//fm-3929-dirmarker.ZIZvMS
```
</details>
<details>
<summary>Evidence: Reusable demo script: #3929 operator scenario</summary>

Source: [Reusable demo script: #3929 operator scenario](https://github.com/dbeihl-utilicast/firstmate/blob/7671a3b114c022dc32aebfff56506a9aa074bb10/.no-mistakes/evidence/fm/qwen-overnight-batch/demo-insecure-state-root.sh)

```text
#!/usr/bin/env bash
# Walk the exact scenario issue #3929 reports, as an operator would:
# arm a process-event source while the state root is private, relax the root
# afterwards, run the reconcile call whose exit the caller swallows, and see
# whether the failure reaches anyone. FM_ROOT points at the tree under test.
set -u
FM_TREE=$1
H=$(mktemp -d "${TMPDIR:-/tmp}/fm-3929-demo.XXXXXX")
mkdir -p "$H/state"

say() { printf '\n=== %s\n' "$*"; }

say "1. arm a condition->action watch while state/ is private (700)"
chmod 700 "$H/state"
FM_HOME="$H" "$FM_TREE/bin/fm-procevent-when.sh" arm demo-watch --interval 0.1 \
  --condition false --action true 2>&1 | sed 's/^/    /'

say "2. an operator relaxes the state root (umask, chmod -R, a helper tool)"
chmod 775 "$H/state"
ls -ld "$H/state" | sed 's/^/    /'

say "3. the watcher's own reconcile call, verbatim: exit and output discarded"
FM_HOME="$H" "$FM_TREE/bin/fm-procevent.sh" reconcile >/dev/null 2>&1 || true
printf '    operator sees: (nothing - that is the swallow)\n'

say "4. is there any durable record of the refusal?"
HAD_RECORD=
if [ -f "$H/.procevent-state-insecure" ] && [ ! -L "$H/.procevent-state-insecure" ]; then
  printf '    %s:\n' "$H/.procevent-state-insecure"
  sed 's/^/      /' "$H/.procevent-state-insecure"
  HAD_RECORD=yes
else
  printf '    NONE - the failure vanished\n'
fi

run_watcher() {  # <out> <tenths>
  local out=$1 tenths=$2 pid= i=0 pf="$H/.demo-watch.pid"
  rm -f -- "$pf"
  ( FM_HOME="$H" FM_STATE_OVERRIDE="$H/state" FM_POLL=1 FM_SIGNAL_GRACE=1 \
      FM_CHECK_INTERVAL=999999 FM_HEARTBEAT=999999 "$FM_TREE/bin/fm-watch.sh" > "$out" 2>/dev/null &
    echo $! > "$pf"; wait ) &
  local wrapper=$!
  while [ "$i" -lt 50 ] && [ -z "$pid" ]; do pid=$(cat "$pf" 2>/dev/null); [ -n "$pid" ] && break; sleep 0.1; i=$((i+1)); done
  i=0
  while [ "$i" -lt "$tenths" ]; do kill -0 "$pid" 2>/dev/null || break; sleep 0.1; i=$((i+1)); done
  kill "$pid" 2>/dev/null || true; wait "$wrapper" 2>/dev/null || true; rm -f -- "$pf"
}

say "5. what the captain actually sees from the real watcher (one poll window)"
run_watcher "$H/watch-1.out" 100
if grep -qF 'procevent-state-insecure' "$H/watch-1.out"; then
  grep -F 'procevent-state-insecure' "$H/watch-1.out" | sed 's/^/    /'
else
  printf '    no wake about the state root at all\n'
fi

say "6. does it become a stream? second watcher window, same broken root"
FM_STATE_OVERRIDE="$H/state" bash -c '. "$1"; fm_wake_append check "procevent:demo-anchor" "check: process-event result captured: demo-anchor"' \
  _ "$FM_TREE/bin/fm-wake-lib.sh" 2>/dev/null || true
run_watcher "$H/watch-2.out" 100
printf '    anchor proving the cycle ran past the check: '
grep -cF 'process-event result captured' "$H/watch-2.out"
printf '    repeat insecure-root wakes this window: '
grep -cF 'check: procevent-state-insecure' "$H/watch-2.out"

say "7. operator restores private permissions and reconciles"
chmod 700 "$H/state"
FM_HOME="$H" "$FM_TREE/bin/fm-procevent.sh" reconcile >/dev/null 2>&1 \
  && printf '    reconcile: ok\n' || printf '    reconcile: FAILED\n'
if [ -e "$H/.procevent-state-insecure" ]; then
  printf '    record still present (stale)\n'
elif [ -n "$HAD_RECORD" ]; then
  printf '    record cleared automatically\n'
else
  printf '    nothing to clear - there never was a record\n'
fi

FM_HOME="$H" "$FM_TREE/bin/fm-procevent.sh" sweep-home >/dev/null 2>&1 || true
rm -rf "$H"
```
</details>
<details>
<summary>Evidence: Reusable demo script: non-empty directory marker case</summary>

Source: [Reusable demo script: non-empty directory marker case](https://github.com/dbeihl-utilicast/firstmate/blob/7671a3b114c022dc32aebfff56506a9aa074bb10/.no-mistakes/evidence/fm/qwen-overnight-batch/demo-nonempty-dir-marker.sh)

```text
#!/usr/bin/env bash
# The round-5 hardening: a NON-EMPTY directory planted at either marker path.
# Plant one at both paths, break the state root, and see whether the captain
# still gets the wake. FM_TREE points at the tree under test.
set -u
FM_TREE=$1
H=$(mktemp -d "${TMPDIR:-/tmp}/fm-3929-dirmarker.XXXXXX")
mkdir -p "$H/state"
chmod 700 "$H/state"
say() { printf '\n=== %s\n' "$*"; }

FM_HOME="$H" "$FM_TREE/bin/fm-procevent-when.sh" arm dir-demo --interval 0.1 \
  --condition false --action true >/dev/null 2>&1

say "plant a non-empty directory at BOTH marker paths"
mkdir -p "$H/.procevent-state-insecure/kept" "$H/.procevent-state-insecure-surfaced/kept"
printf 'do not lose me\n' > "$H/.procevent-state-insecure/kept/payload"
ls -d "$H"/.procevent-state-insecure* | sed 's|.*/|    |'

say "relax the state root, then run the swallowed reconcile"
chmod 775 "$H/state"
FM_HOME="$H" "$FM_TREE/bin/fm-procevent.sh" reconcile >/dev/null 2>&1 || true
if [ -f "$H/.procevent-state-insecure" ] && [ ! -L "$H/.procevent-state-insecure" ]; then
  printf '    durable record written:\n'; sed 's/^/      /' "$H/.procevent-state-insecure"
else
  printf '    NO record - the planted directory was read as a valid one\n'
fi

run_watcher() {  # <out> <tenths>
  local out=$1 tenths=$2 pid= i=0 pf="$H/.demo-watch.pid"
  rm -f -- "$pf"
  ( FM_HOME="$H" FM_STATE_OVERRIDE="$H/state" FM_POLL=1 FM_SIGNAL_GRACE=1 \
      FM_CHECK_INTERVAL=999999 FM_HEARTBEAT=999999 "$FM_TREE/bin/fm-watch.sh" > "$out" 2>/dev/null &
    echo $! > "$pf"; wait ) &
  local wrapper=$!
  while [ "$i" -lt 50 ] && [ -z "$pid" ]; do pid=$(cat "$pf" 2>/dev/null); [ -n "$pid" ] && break; sleep 0.1; i=$((i+1)); done
  i=0
  while [ "$i" -lt "$tenths" ]; do kill -0 "$pid" 2>/dev/null || break; sleep 0.1; i=$((i+1)); done
  kill "$pid" 2>/dev/null || true; wait "$wrapper" 2>/dev/null || true; rm -f -- "$pf"
}

say "one watcher window: does the captain hear about the broken root?"
FM_STATE_OVERRIDE="$H/state" bash -c '. "$1"; fm_wake_append check "procevent:dir-anchor" "check: process-event result captured: dir-anchor"' \
  _ "$FM_TREE/bin/fm-wake-lib.sh" 2>/dev/null || true
run_watcher "$H/watch.out" 250
printf '    anchor proving the cycle ran past the check: '; grep -cF 'process-event result captured' "$H/watch.out"
if grep -qF 'check: procevent-state-insecure' "$H/watch.out"; then
  grep -F 'check: procevent-state-insecure' "$H/watch.out" | sed 's/^/    /'
else
  printf '    SILENT - the planted directory latched the wake away\n'
fi

say "was the planted directory content preserved?"
d=$(printf '%s\n' "$H"/.procevent-state-insecure.displaced-* 2>/dev/null | head -1)
if [ -f "$d/kept/payload" ]; then
  printf '    moved aside to %s, payload intact: %s\n' "${d##*/}" "$(cat "$d/kept/payload")"
elif [ -f "$H/.procevent-state-insecure/kept/payload" ]; then
  printf '    left in place untouched\n'
else
  printf '    payload LOST\n'
fi

FM_HOME="$H" "$FM_TREE/bin/fm-procevent.sh" sweep-home >/dev/null 2>&1 || true
chmod -R 700 "$H/state" 2>/dev/null || true
printf "\nhome: %s\n" "$H"
```
</details>

## Pipeline

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>
<summary>⚠️ **Review** - 2 issues (1 warning, 1 info)</summary>

- ⚠️ `bin/fm-procevent.sh:206` - Both new runtime markers land in FM_HOME, which defaults to FM_ROOT (the firstmate checkout), and neither name is in .gitignore - unlike the sibling FM_HOME-root markers .fm-secondmate-home and .fm-secondmate-parent, which both are. Concrete failure: state/ stops being private, fm-procevent.sh:212-215 writes $FM_ROOT/.procevent-state-insecure and fm-watch.sh:1825 writes $FM_ROOT/.procevent-state-insecure-surfaced; bin/fm-update.sh:80 calls ff_target &#34;$FM_ROOT&#34; firstmate origin no no, whose dirty check (bin/fm-ff-lib.sh:345 via dirty_status with ignore_seed_marker=no) now sees `?? .procevent-state-insecure` and prints &#34;firstmate: skipped: dirty working tree&#34; instead of fast-forwarding. The markers are deliberately left in place while the root stays non-private, so firstmate self-update stays disabled until someone deletes them by hand. tests/fm-gitignore-config.test.sh states this exact invariant. Remedy is mechanical: add .procevent-state-insecure and .procevent-state-insecure-surfaced to .gitignore.
- ⚠️ `bin/fm-watch.sh:1825` - The one-shot latch is written before wake runs, so a wake that never delivers is lost permanently while the latch suppresses every later cycle - the same silent-loss class this change exists to close. Concrete path: watcher stdout is a pipe whose reader is gone; line 1824 creates .procevent-state-insecure-surfaced, then wake (bin/fm-push-transition-lib.sh:88-109) hits EPIPE on `echo &#34;$1&#34;`. Because FM_WAKE_POST_OUTPUT_ACTION is empty, PIPE is not trapped (line 95), so the watcher dies on SIGPIPE, or on the handled path sets output_status=1 and exits without calling watch_delivery_publish. Nothing was delivered, the latch is on disk, and every restarted watcher short-circuits at line 1824 forever (only a bearings-board build would still report it). The adjacent procevent path in the same file already solves this: procevent_surface_queued sets FM_WAKE_POST_OUTPUT_ACTION=procevent_surface_after_output (fm-watch.sh:1328-1330), which commits its surfaced markers only when output_status -eq 0 (fm-watch.sh:1297-1308). Use that existing hook rather than latching first.
- ⚠️ `bin/fm-watch.sh:1824` - No test exercises the watcher branch that turns the durable marker into a `check: procevent-state-insecure` wake, which is the only place the failure becomes visible to firstmate. tests/fm-procevent-when.test.sh:420-434 proves the marker file appears, stays first-write-wins, and clears - all through fm-procevent.sh - so deleting bin/fm-watch.sh:1824-1827 entirely leaves the suite green and nothing pins the surfacing half the intent asks to prove. The harness already exists and is cheap: tests/fm-inactive-reconcile.test.sh:619-638 backgrounds $WATCH with FM_POLL=1 FM_CHECK_INTERVAL=999999 and greps `check: inactive-outcome` from the captured watcher output; the same shape would assert `check: procevent-state-insecure` fires once and not on the following cycle.
- ℹ️ `bin/fm-procevent.sh:212` - Both marker paths are probed with bare `-e`, unlike the `[ -e &#34;$STATE&#34; ] || [ -L &#34;$STATE&#34; ]` idiom four lines above at fm-procevent.sh:208. A dangling symlink at $FM_HOME/.procevent-state-insecure makes `[ ! -e &#34;$INSECURE_MARKER&#34; ]` true and `printf ... &gt; &#34;$INSECURE_MARKER&#34;` then creates and writes the symlink&#39;s target at any path the user can write; a symlink at $FM_HOME/.procevent-state-insecure-surfaced pointing at any existing file makes fm-watch.sh:1824&#39;s `[ ! -e ... ]` false forever and suppresses the wake. The sibling scan marker is symlink-hardened with a dedicated regression (tests/fm-inactive-reconcile.test.sh:592-603). Fix is the established idiom: treat a symlink at either path as present-and-invalid and stage via mktemp+chmod+mv instead of redirecting onto the path.

🔧 Fix: gitignore markers, latch wake after delivery, harden symlinks

- ⚠️ `bin/fm-watch.sh:1320` - The new symlink hardening stages via mktemp+mv but never unlinks the target, and `mv` resolves a symlink-to-directory as a directory rather than replacing the link. Verified on this platform: `ln -s realdir link; mv -f -- tmpfile link` exits 0, drops tmpfile INSIDE realdir, and leaves `link` a symlink. Concrete failure at bin/fm-watch.sh:1320: a symlink at $FM_HOME/.procevent-state-insecure-surfaced pointing at any existing directory passes the watcher&#39;s `[ -L ... ]` disjunct (fm-watch.sh:1838) so the wake fires, then procevent_state_insecure_after_output&#39;s `mv` moves the empty temp into the target directory and returns 0 while the suppressor path is STILL a symlink - so the next poll takes the `-L` branch again and wakes again, forever, dropping another temp file into the attacker-chosen directory each cycle. That is precisely the &#34;stream of them on every poll&#34; the intent forbids, and it is self-sustaining because the latch can never commit. The twin at bin/fm-procevent.sh:213 has the same hole with a quieter consequence: a symlink-to-directory at $FM_HOME/.procevent-state-insecure passes the `[ -L &#34;$INSECURE_MARKER&#34; ]` disjunct at line 210, the record is written into the symlink target instead, insecure_marker_record returns 0, and the durable record the board&#39;s fail message at fm-bearings-board.sh:430 points a human at never exists. The new tests only plant symlinks to regular FILES (tests/fm-procevent-when.test.sh:483,491), where `mv` does replace the link, so neither case is covered. Remedy is mechanical and matches the finding that prompted this hardening (&#34;treat a symlink at either path as present-and-invalid&#34;): `rm -f -- &#34;$marker&#34;` immediately before each `mv` so the rename lands on a nonexistent path, plus a symlink-to-directory case in both regressions. `mv -T` is not portable to the macOS path this repo supports.
- ⚠️ `tests/fm-procevent-when.test.sh:451` - The idempotence assertion cannot fail, so the first-write-wins invariant is unproven. `detected=` is written with `date +%s` (bin/fm-procevent.sh:212), one-second resolution, and the two `pe &#34;$H&#34; reconcile` calls at lines 445 and 449 are separated only by an awk and an assert_grep - microseconds. Delete the `[ ! -e &#34;$INSECURE_MARKER&#34; ] || [ -L &#34;$INSECURE_MARKER&#34; ] || return 0` guard at bin/fm-procevent.sh:210 and the second reconcile rewrites the record with a fresh timestamp that lands in the same wall-clock second, `state=` is identical either way, and line 451 still passes. Nothing else in the suite pins &#34;one durable record, not a stream&#34; - the intent&#39;s explicit requirement. Mechanical fix with no behavior change: after line 447 append a sentinel line to the marker (`printf &#39;sentinel=1\n&#39; &gt;&gt; &#34;$H/.procevent-state-insecure&#34;`), run the second reconcile, and assert_grep the sentinel survived; that fails the moment the record is replaced, regardless of clock resolution.
- ℹ️ `bin/fm-procevent.sh:57` - The header comment is emitted verbatim as `bin/fm-procevent.sh --help` output (usage() sed-extracts lines 2..`set -u`), and it tells the operator the record lives at `state/../.procevent-state-insecure`. The actual path is `$FM_HOME/.procevent-state-insecure` (line 206). Those coincide only when STATE is the default `$FM_HOME/state`; with FM_STATE_OVERRIDE set - which is how fm-watch.sh, the test suite, and every non-default home invoke this script - `state/..` resolves somewhere the file will never be, sending someone chasing the marker to the wrong directory. Replace with `$FM_HOME/.procevent-state-insecure`, which is what the code computes and what fm-bearings-board.sh:430 already prints.
- ℹ️ `tests/fm-procevent-when.test.sh:107` - In the closed-reader branch, `$!` is the pid of the `( ... | true )` subshell, not of fm-watch.sh, so the `kill &#34;$pid&#34;` at line 109 and the `kill -0` liveness loop both target the wrapper. Today the watcher exits on its own (wake&#39;s `exit 1` after the EPIPE), so the leak stays latent - but the moment the wake does not fire within the 10s budget (slower machine, or the insecure block moving behind a step that returns early), the subshell is killed and fm-watch.sh survives as an orphan still holding the home&#39;s watcher lock and polling a $TMP_ROOT that teardown then deletes. The next `run_watcher` call on the same home at line 469 self-evicts against that orphan&#39;s lock and reads an empty out file, so the failure surfaces as an unexplained missing `check: procevent-state-insecure` rather than as a leaked process. The file-output branch has the pid right; make the pipe branch match by capturing the watcher&#39;s own pid, e.g. `( ... &amp; echo $! &gt; &#34;$home/.watch.pid&#34;; wait ) | true &amp;` and killing the recorded pid.

🔧 Fix: unlink marker before mv, pin one-record invariant, fix helper pid

- ⚠️ `tests/fm-procevent-when.test.sh:492` - The two &#34;the watcher does not re-report&#34; assertions give the watcher 3 seconds (run_watcher ... 30) while the positive assertion on the very same home needed 10 seconds (run_watcher ... 100) to observe one completed poll cycle. Nothing in the 3-second run anchors that a cycle ran at all: the assertion is a bare `grep -Fq ... &amp;&amp; fail` over an output file that is also empty when the watcher never reached the check. Concrete failure: on a loaded machine where the first cycle (lock acquisition, recovery-marker reopen, fm_pending_reply_tick, secondmate_wake_stall_tick, the reconcile subprocess) exceeds 3s, delete the `-surfaced` disjunct at bin/fm-watch.sh:1838 so the watcher re-wakes every cycle, and line 493 still finds nothing and passes. The positive run proving 10s was needed is exactly the evidence that 3s is too short. Same shape at line 520 for watch-symlink-dir-again.out (whose fail message is &#34;the suppressor never committed, so the watcher woke again every cycle&#34;). Mechanical fix with no behavior change: give both negative runs the same 100-tenths budget the positive runs used, so the run that must not wake gets at least as long as the run that did.
- ⚠️ `AGENTS.md:147` - The change introduces two new files at the FM_HOME root - .procevent-state-insecure (written by bin/fm-procevent.sh:206-215) and .procevent-state-insecure-surfaced (written by bin/fm-watch.sh:1316-1323) - plus their mktemp siblings, and neither is listed in the home-level state inventory in AGENTS.md. That inventory is exhaustive for this level: it documents config/, data/, projects/, state/ and .no-mistakes/ at line 147, and every state/ entry carries its owning script and mutation contract (&#34;written only by bin/fm-procevent.sh&#34;, &#34;never touch&#34;, &#34;safe to delete&#34;). An operator who finds .procevent-state-insecure appear in their home has no documented owner, no format, and no answer to whether deleting it is safe - and deleting the -surfaced half specifically re-arms a wake, which is exactly the kind of consequence the surrounding entries spell out. Add two lines beside .no-mistakes/ naming bin/fm-procevent.sh as the record&#39;s owner, bin/fm-watch.sh as the suppressor&#39;s owner, and that both are cleared automatically once the state root is private again.
- ℹ️ `docs/verification/process-event-sources.md:131` - This file is the repo&#39;s per-scenario evidence ledger, and the prior round shipped a dedicated commit (5b6174c) to record the arm-time refusal in it. This round adds three more proved scenarios to the same suite that are not recorded: the durable record survives a repeat reconcile with its first detection intact and clears once the root is private again (tests/fm-procevent-when.test.sh:453-465), the watcher turns the swallowed reconcile failure into exactly one wake and latches the suppressor only after delivery (476-495), and a planted symlink - including one pointing at a directory - is replaced at either marker path rather than written through (500-548). The row updated at line 131 covers only the arm refusal. Add the surfacing and marker-integrity scenarios as their own rows so the ledger still describes what the suite actually proves.
- ℹ️ `bin/fm-procevent.sh:210` - The hardening covers symlinks at both marker paths but not a plain directory. At bin/fm-procevent.sh:210 the guard `[ ! -e &#34;$INSECURE_MARKER&#34; ] || [ -L &#34;$INSECURE_MARKER&#34; ] || return 0` reads a real directory as &#34;a valid record already exists&#34; and returns without recording; the clearing `rm -f` at line 224 also fails on it, so it never goes away. The sharper half is the suppressor: `mkdir $FM_HOME/.procevent-state-insecure-surfaced` makes the watcher&#39;s condition at bin/fm-watch.sh:1838-1839 permanently false (`-e` true, `-L` false), so the state-root-not-private wake can never fire again for that home. Noting it rather than asking for a fix because the obvious remedy conflicts with an explicit acceptance constraint: relaxing the suppressor test to `[ ! -f ]` makes the wake fire, but procevent_state_insecure_after_output&#39;s `rm -f`/`mv` cannot replace a directory either, so the suppressor never commits and the watcher wakes every cycle - precisely the &#34;stream of them on every poll&#34; the intent forbids. Replacing it would need `rm -rf` on an attacker-chosen path. Recording the residual boundary of the symlink hardening.

🔧 Fix: reject directory markers, anchor negative runs, document markers
1 warning still open:

- ⚠️ `bin/fm-pr-lib.sh:277` - fm_marker_settled accepts a NON-EMPTY directory as a settled marker (`[ -d &#34;$path&#34; ] &amp;&amp; [ -n &#34;$(ls -A ...)&#34; ]`), so the directory hardening this round claims only holds for the empty-directory case the suite plants. Two concrete wrong outcomes, neither of which errors. (1) Silence: `mkdir -p $FM_HOME/.procevent-state-insecure-surfaced/x`, then relax the state root. The watcher&#39;s reconcile fails silently, insecure_marker_record writes the record, but at bin/fm-watch.sh:1836-1837 `fm_marker_settled record` is true and `fm_marker_settled surfaced` is ALSO true, so the `check: procevent-state-insecure` wake never fires for that home again — exactly the silent outage of issue #3929, re-entered one layer up. procevent_state_insecure_after_output&#39;s fm_marker_clear could not replace it either, which is the tradeoff the helper comment documents, but the prior round&#39;s instruction was that the suppressor check must not treat a directory as a valid latch, and for a non-empty directory it still does. (2) Wrong label: `mkdir -p $FM_HOME/.procevent-state-insecure/x` on a perfectly private root makes the watcher wake claiming the root stopped being private, insecure_marker_record return early without writing a real record, and fm_marker_clear at bin/fm-procevent.sh:221 fail forever — so bin/fm-bearings-board.sh:429&#39;s `[ -e ]` check permanently misattributes every later not-listening failure to a non-private state root. The docs assert the handled case without qualification: docs/verification/process-event-sources.md:134 says a plain directory is proved &#39;replaced rather than written through or into, and that neither shape can silence the wake&#39;, and AGENTS.md:148 says the suppressor is &#39;cleared with the record&#39;. Both are false for a non-empty directory, and no test plants one. Remedy that preserves behavior: narrow the ledger row and the AGENTS.md line to the shapes actually handled (regular file, symlink, symlink-to-directory, empty directory) and state the unreplaceable-path tradeoff the helper comment already names. Rejecting a non-empty directory outright would reintroduce the wake-every-cycle stream the intent forbids, so do not do that without asking.

🔧 Fix: reject any directory as a settled marker

- ⚠️ `bin/fm-bearings-board.sh:429` - The board-side diagnostic added for issue #3929 is placed on a branch the insecure-root case cannot reach, so `fm-bearings-board.sh build` still misattributes the failure it was added to explain. Reaching line 429 requires `owner` to be non-empty at line 408, but `source_owner` (line 323) runs `fm-procevent.sh list 2&gt;/dev/null`, and every fm-procevent.sh command now dies at the top-level state-root check on a non-private root, so `list` prints nothing and `owner` is empty. Concrete trace for exactly the intent&#39;s scenario (root armed private, then relaxed to 775): line 392 `pre_reopen_owner=$(source_owner)` -&gt; &#34;&#34; (this call also writes the marker); line 408 `owner=$(source_owner)` -&gt; &#34;&#34;; the `elif [ -n &#34;$owner&#34; ]` already-armed branch is skipped; the `else` branch runs `fm-procevent-lavish.sh arm`, whose `fm-procevent.sh register` dies on the same check, and the board exits at `fail &#34;cannot arm the board as a process-event source&#34;` (line 417) - never reaching the reconcile at 426 or the new `-e` check at 429. With BOARD_SESSION_REOPENED=1 it exits even earlier at `fail &#34;cannot retire the pre-reopen source generation&#34;` (line 395). Line 429 fires only if the root flips from private to insecure inside the few hundred milliseconds between line 408 and line 426, because a private root at 408 means the same `list` call just cleared the marker. No test covers this branch, consistent with it being unreachable. The cause is on disk at $FM_HOME/.procevent-state-insecure at both earlier exits (the failed `list` and the failed `register` each write it), so the remedy is placement, not new machinery: consult the marker at the arm/retire failure exits too (or once in `command_build` around the procevent calls), using `fm_marker_settled` rather than `-e` so the shared predicate this change introduced is the one deciding.
- ℹ️ `bin/fm-watch.sh:1839` - Recording the residual boundary of the surfacing design, not asking for a change. `check: procevent-state-insecure` is a bare `wake` echo with no durable wake-queue row, unlike its sibling `check: inactive-outcome`, whose findings are appended via `fm_wake_append` and re-announced until acknowledged. `procevent_state_insecure_after_output` latches the suppressor on successful delivery (output_status 0), not on the captain acting, so a wake that is printed and then scrolled past (see issue #3984 on Stop-hook feedback blocks getting lost) is never re-announced. The state root stays non-private, every reconcile keeps failing silently, and the remaining surfaces are the marker file itself plus a board build failure whose message does not name the cause (see the board finding above). This is a faithful reading of the captain&#39;s explicit constraint - &#34;one durable, visible record, not a stream of them on every poll, no general alerting mechanism&#34; - and the smallest alternative (a queue row with ack semantics) would add persistence machinery beyond this change, so noting the boundary rather than proposing a fix.
</details>

<details>

✅ No issues found.
- <code>`bash tests/fm-procevent-when.test.sh` (21 scenarios, all pass, incl. the five new state-root/marker scenarios)</code>
- <code>`bash tests/fm-gitignore-config.test.sh` (6 scenarios, incl. the new marker-porcelain case)</code>
- <code>Manual operator walkthrough on the base tree `72bfdd0` (extracted via `git archive`): arm while `state/` is 700, `chmod 775`, run `fm-procevent.sh reconcile &gt;/dev/null 2&gt;&amp;1 || true`, then a real `bin/fm-watch.sh` poll window — no record, no wake</code>
- <code>Same walkthrough on `6b937c8`: durable record written, one `check: procevent-state-insecure` wake delivered, no repeat on a second window anchored by `fm_wake_append check procevent:demo-anchor`, record auto-cleared after `chmod 700` + reconcile</code>
- <code>Non-empty-directory marker case on `bfb9cf6` (pre round-5 fix): planted `.procevent-state-insecure/kept/payload` and `.procevent-state-insecure-surfaced/kept` reproduce the silence (no record, no wake, anchor=1 proving the cycle ran past the check)</code>
- <code>Same non-empty-directory case on `6b937c8`: record written, wake fires, directory moved aside to `.procevent-state-insecure.displaced-*` with payload intact</code>
</details>

<details>

✅ No issues found.
</details>

<details>

🔧 Fix: fix SC1007 empty local assignment in when test
</details>

<details>

✅ No issues found.
</details>


kunchenguid:
Speaking as Kun's firstmate:

**Triage:** Tip `5b6174c2816a9018ca5bfe705dee67db21b801b7`. Attestation **MATCH**. The change is a restore of the process-event private-directory contract. Not merged this pass: hosted CI had not run yet. First-time fork workflows are approved; waiting on CI.

**What I inspected:** `bin/fm-procevent-when.sh` `cmd_arm` on main only checks that `$STATE` is a directory and not a symlink, then writes spec, trust, and registry files. `bin/fm-procevent-lib.sh` already owns `fm_procevent_state_root_resolve` / `fm_procevent_private_directory_valid` (owner-only, not group-writable, not a symlink). This tip calls that resolve before any `when` arm write, and the new test refuses a mode-775 state root before spec/trust/registry files exist. Docs line in `docs/verification/process-event-sources.md` matches that refusal. No Greptile line comments. Greptile Review check passed.

**VISION (per rule):**
- One captain, one interface — **aligns**. A refused arm is a below-deck stop, not new captain noise.
- Authority is explicit — **aligns**. A `when` arm can run a condition and an action. Refusing a non-private state root is the existing grant boundary, not a new one.
- Scripts own the mechanics — **aligns**. The private-directory check is exact and already owned by the process-event lib.
- A restart is a non-event — **aligns**. Watch records stay in the home's private state. Arming onto a shared-writable root would not survive as a private record.
- Delegation with a spine — **aligns**. Not a new task shape. The watch still fires only under the existing `when` contract.
- The fleet outlives any vendor — **aligns**. Local state-root privacy, not a vendor UI.
- Scope — **aligns**. Command-layer process-event trust. Field hole → regression coverage.

**Contract-class:** restore. Other process-event paths already refuse a non-private state root. `when` arm was the path that skipped that check. An unconfigured home with a private `state/` still arms as before.

**Security:** this is the hardening, not a tip. No captain card.

Waiting on CI I just approved: CI `34177027014`, Require no-mistakes `34177026888`. Not a Firstmate flag.

dbeihl-utilicast:
Closing: this pull request targets the upstream repository, and our standing rule is that we never publish changes to kunchenguid/firstmate. The work belongs on our own fork, dbeihl-utilicast/firstmate. Closed on the captain instruction, 2026-09-15. No changes were merged upstream.

PR #3978, opened 2026-09-08
fix(bin): keep a declared external wait from re-arming the stale alarm (#3942)

### LINKED ISSUE #3942: Declared `paused:` external wait does not suppress the watcher's pane-based stale alarm
## Summary

`AGENTS.md` section 8 defines `paused:` as "a bounded external wait expected to clear on its own", explicitly distinct from `blocked:` ("firstmate action is needed"). A worker that correctly declares such a wait still gets flagged `stale:` by the watcher on the normal cadence, so **the documented mechanism for declaring a legitimate long wait does not suppress the alarm it exists to prevent**.

## Repro

1. Spawn a ship task.
2. Have the worker start a genuinely long background job (here: a ~3h data acquisition) and append a well-formed `paused:` line naming the clearing condition and the pid it is waiting on.
3. Let the agent's turn end while the background job runs.

`bin/fm-crew-state.sh <id>` correctly reports `state: paused · source: status-log · ...`. The watcher raises `stale: <pane>` anyway, every poll, for the whole duration of the wait.

## Observed

Task `ind-bhavcopy-cash-universe`, backend herdr, endpoint `default:wB:p2`.

- `paused:` status appended at ~epoch **1788707984**, text: `paused: waiting on the 2006-2026 bhavcopy acquisition (pid 57641) to finish - 3729 of ~5,400 dates probed ... Clears on its own when the fetch exits; a background waiter on that pid will wake me.`
- `stale:` wake seq **613** fired at epoch **1788708070** — 86 seconds *after* the pause was declared, with `fm-crew-state.sh` already reporting `paused`.
- Same episode: seq 607 (1788707790), 608 (1788707895), 618 (1788708426), 621 (1788708958) — and onward for the full ~90 minutes of the download.

Verified throughout that the wait was genuine, not a wedge: the store grew by +33 sessions and +55,831 rows in a measured 20-second window while the alarm was firing.

## Likely cause

The pane-liveness staleness path does not consult the status-log-reconciled state. A declared external wait and an actually-wedged worker are indistinguishable to it, which is precisely the distinction `paused:` was introduced to express.

## Suggested direction

Have the staleness check consult reconciled state before raising, and suppress (or substantially lengthen the grace for) a task whose latest status event is the configured paused verb — ideally bounded, so a `paused:` that never clears eventually escalates rather than silencing the task forever. `FM_CLASSIFY_PAUSED_VERB` already names the verb.

## Operator-visible consequence

Two costs, beyond the noise:

1. **It made the situation less legible to the captain, not more.** Having been told the work was "paused", the captain twice asked why it was paused and whether it was stuck — because the alarms kept arriving alongside that word. The vocabulary meant to reassure ended up reading as a problem.
2. **Habituation.** A repeating alarm on a task known to be fine trains the operator to acknowledge without looking, and a real stall presents identically.

Shares a likely root cause with the "delivered ship task awaiting merge has no quiet state" issue filed alongside this one (#3941): in both cases a correctly-declared non-working state fails to suppress a pane-based staleness alarm.

## Environment

macOS, herdr backend, primary harness claude, Claude Stop-hook-owned supervision (`bin/fm-claude-stop-autoarm.sh`).

scfconnect:
Fixes #3942.

## What was wrong

`paused:` is documented as a bounded external wait that must not be treated as a possible wedge, and the suppression path for it already existed. The defect was not a missing check: the path was being erased underneath itself.

A declared wait records a re-surface throttle, `.paused-resurfaced-<key>`, which is scoped to the DECLARATION rather than to the pane. Two branches of the staleness loop called `clear_pause_tracking` whenever a poll found the pane busy, and that deletes the throttle. A worker parked on a declared wait still renders its harness busy signature whenever it does anything at all, such as taking a steer or reacting to its background job, so one such poll handed the unchanged declaration a fresh window. The next idle sighting then read the absent throttle as a first sight and alarmed again, far inside `FM_PAUSE_RESURFACE_SECS`. Every blip cost one more `stale:` wake, which is how an operator ends up being told the work is paused while being paged that it is stuck.

Review of the first commit found the same defect reachable by a second, narrower route. `bin/fm-crew-state.sh` emits `state: working source: pane` before its `paused:` status-log fallback, and it takes its own pane capture a moment after the staleness loop already read that same pane idle. A parked worker that blips between those two captures therefore arrives as a `working` verdict, and the three sibling `working` arms retired the same throttle.

## The change

- The busy branches clear only the per-hash bookkeeping while the status line still declares a wait. Withdrawing a declaration keeps its single owner, the loop-top reconciliation.
- The `working` arms still end pause mode, because an authoritative active run really is evidence the crew is not idle, but they hold back the re-surface marker. That is the same split `surface_nonterminal_stale` already makes, for the same reason.
- The duplicated not-paused arm is dropped from both busy branches, which also removes a redundant `window_to_task` subprocess from the busy poll's hot path.

Bounded escalation is unchanged. An absorbed sighting never refreshes the throttle, so the window still elapses and a wait that never clears still re-surfaces once, and a crew that declared nothing keeps alarming on every new stale hash.

## Verification

- The regression drives the real `bin/fm-watch.sh` poll loop and asserts on the durable wake queue and the marker files. Both halves fail against unfixed code and pass with the change: the busy-pane blip and the racing working-verdict each reproduce the reported alarm on their own.
- A disconfirming half asserts that an undeclared silent crew keeps alarming on every new stale hash, so the bound cannot swallow a genuinely stopped worker.
- The full `watcher-wake-lock` family runs green, 18 suites with 0 failures, including the pre-existing tests that pin the authoritative-active-run and wedge-timer behavior. `bin/fm-lint.sh` is clean.

`docs/architecture.md` is updated to record that a provably-working verdict does not retire the throttle either, and why.


kunchenguid:
Speaking as Kun's firstmate:

Reviewed the tip `be2e0592dd8aaea261ecb137aeebcfa4700ef990` on main `b84e0e362face25f3dd8945297a3df1320d7668c`. Diff is `bin/fm-watch.sh` + docs + regression in `tests/fm-watch-triage.test.sh` — no `.github/workflows` changes. This restores the already-documented `paused:` / captain-held re-surface throttle so a busy-pane blip or racing `working` crew-state verdict cannot erase `.paused-resurfaced-*` and re-arm `stale:` inside `FM_PAUSE_RESURFACE_SECS`. Fixes #3942 as claimed.

**VISION.md (each rule)**
- One captain, one interface: **aligns** — stops paging the captain that a declared wait is both paused and stuck.
- Authority is explicit: **aligns** — no new grant; bounded re-surface cadence unchanged.
- Scripts own the mechanics: **aligns** — throttle ownership stays in `fm-watch.sh`.
- A restart is a non-event: **aligns** — durable pause markers remain the source of truth.
- Delegation with a spine: **aligns** — corrective to existing pause contract.
- The fleet outlives any vendor: **aligns** — harness-agnostic watcher.
- Scope: **aligns** — supervision command layer.

Contract-class: **restore**. Attestation: **MISSING** (no no-mistakes raise / attestation marker on tip). Fork CI+NM approved this pass (runs `34195553856` / `34195553968`). Waiting on author for attestation MATCH and green CI/NM. Security: clean. Do not auto-merge yet.

PR #4026, opened 2026-09-08
fix(bin): break the Claude Stop guard/auto-arm deadlock that refused every turn end

MortenGad:
## Intent

The captain authorised fixing the supervision defects that keep interrupting his fleet, and this is
one of them. It is the defect that has cost the most working time: over a single day it repeatedly
refused every turn end in the supervising session, and the only operator remedy found was restarting
that session.

WHAT GOES WRONG

The Stop boundary registers two hooks in order, `bin/fm-turnend-guard.sh --claude` first and
`bin/fm-claude-stop-autoarm.sh` (asyncRewake) second. The guard refuses a stop when no live watcher
holds the home lock. The arm only runs when a turn actually ends. So once the watcher has exited and
the guard starts refusing, the arm can never run, which guarantees the next refusal. The two hooks
deadlock each other.

Observed 2026-09-06 05:00-05:40: every turn end refused with "TURN WOULD END BLIND - SUPERVISION IS
OFF ... The Stop-owned auto-arm did not claim this home either". `state/.claude-autoarm-epoch` stayed
frozen at `epoch=165 owner_pid=41746 outcome=rewake updated_at=1788664081` for over forty minutes,
while `state/.turnend-claude-blocks` stayed `count=1 epoch=165` and its bounded fail-open never
advanced, because it is pinned to an epoch that cannot change.

Entry condition: the watcher exits normally after delivering an actionable wake. Everything else was
healthy. The session lock was held by a live harness pid, hook registration was correct and complete,
no autoarm failure-notified or failure-alarmed markers existed, and `state/.watch-triage.log` showed
the watcher absorbing four parked tasks correctly before each exit.

WHY IT CANNOT BE REPAIRED FROM INSIDE THE SESSION, established by three attempts:
1. `bin/fm-watch-arm.sh` restores a watcher that beats, but it is one-shot. It exits at its next
actionable close and the deadlock returns.
2. A second arm when no recheck was due held about thirty minutes, then reported a due recheck,
exited, and the deadlock returned. The duration varies; the outcome does not.
3. Running `bin/fm-claude-stop-autoarm.sh` directly exits 0 immediately, prints nothing, and does NOT
advance the epoch or claim the home. The arm appears to require the harness Stop invocation
context, so the owner cannot be made to claim from an ordinary tool call.

The chain can therefore only re-engage from a genuine harness Stop event, which the guard is
preventing.

TWO VARIANTS, AND WHY A FIX KEYED ON EPOCH STALENESS ALONE IS WRONG

A second episode the same day, 07:30-09:00 in the same home, recovered on its own several times
instead of locking up, so the deadlock is intermittent. In that window the epoch ADVANCED (166 failed
to 167 rewake) rather than freezing. A frozen epoch is one presentation of this defect, not its
definition.

That window also showed a third symptom: the arm can fail while a live watcher process holds the
singleton lock with a heartbeat stale past the 300s grace. It reports "lock held by live pid <n> but
heartbeat is stale for <n>s; inspect or stop that watcher before re-arming", then "watcher cycle
exited 1 without an actionable reason". Reproduced twice with different pids about an hour apart.
Both times the named pid no longer existed by the time the handling turn ran `ps` against it, the
beacon was fresh well inside the 300s grace, and the drain printed no watcher-down banner. So the arm
and the real state are racing: the arm reports a broken mechanism while supervision is in fact alive.
These alarms are largely spurious and must be verified before being acted on.

CONTRIBUTING LOAD WORTH RULING OUT AS A CAUSE

Through that window the watcher was re-absorbing two captain-held parked tasks roughly every
seventeen seconds for hours, because their wait was never declared in their status. Once each worker
appended its own `paused:` line the churn stopped and the recheck moved to the intended long cadence.
Whether that churn was driving the collapses is untested, but the two coincided closely.

## What Changed

- `bin/fm-turnend-guard.sh` now primes `bin/fm-claude-stop-autoarm.sh --ensure-watcher` on the refusal path only, placed below the attended fail-open and immediately before the final `block_stop`, so a refused Stop no longer aborts the only process that could restore a watcher. The new `--ensure-watcher` mode in `bin/fm-claude-stop-autoarm.sh` reuses the existing identity, AFK, and need gates, then setsid-detaches `bin/fm-watch-arm.sh` as a handling successor (`FM_WATCH_PREDECESSOR_ARM_PID`) carrying the guard's derived `FM_GUARD_GRACE`, takes no generation claim, and discards the cycle's stdout so `state/.watch-cycle-exits.log` stays its single record.
- Block-budget accounting takes a `blocked-stop` argument: a Stop that reaches the refusal now advances the count even when the auto-arm epoch has not changed, so a frozen epoch can no longer pin `state/.turnend-claude-blocks` at `count=1` and starve the bounded fail-open. The Claude refusal banner collapses to one unconditional sentence that reports the missing auto-arm claim without asserting anything about the detached cycle.
- Adds five `--ensure-watcher` cases to `tests/fm-claude-stop-autoarm.test.sh` (detached start without a claim, AFK and session-lock inertness, refused-cycle ledger row, derived grace) and six guard cases to `tests/fm-turnend-guard.test.sh` (ordinary Stop does not prime, frozen/advancing/no-progress refusal variants, fail-open allows without priming, frozen epoch still reaches the fail-open); replaces GNU-only `touch -d @<epoch>` in the three away-mode grace cases with a portable `set_mtime` helper, and records the contract and verification runs in `docs/turnend-guard.md`, `docs/watcher-continuity.md`, and `docs/verification/supervision.md`.

## Risk Assessment

⚠️ Medium: The change is well-targeted and now correct on the paths I traced - priming is genuinely refusal-only, the frozen-epoch count no longer pins the bound, and the successor disposition only defers rather than destroys downtime re-presentation - but it adds a detached process fork from a Stop hook and changes when the loud attended fail-open becomes reachable (consecutive refusals rather than distinct epochs), so it warrants the follow-ups above rather than being trivially safe.

## Testing

I ran the two smallest relevant suites (fm-turnend-guard, fm-claude-stop-autoarm) and then demonstrated the intent the way an operator experiences it: a reproducer stands up real Firstmate homes carrying the real bin/ tree of the branch base and of the branch head, seeded with the recorded 2026-09-06 state (task in flight, no watcher, epoch frozen at 165 rewake, blocks count=1 epoch=165), and fires genuine fm-turnend-guard.sh --claude Stop hooks under a harness process owning state/.lock, without the sibling asyncRewake arm — exactly what Claude Code does when a Stop is blocked. At base, three turn ends in a row refuse, no watcher ever appears, the ledger stays empty and the block count stays pinned at 1; at head the first refusal forks a real fm-watch-arm.sh/fm-watch.sh that takes the home lock, the count advances 1→2, the auto-arm epoch is left byte-identical (no generation claim), and the next Stop is allowed silently. The same before/after holds for the advancing-epoch variant (166 failed → 167 rewake), where head also leaves the allowed handoff Stop unprimed. Running the head test file against the base bin/ tree fails at the first priming assertion, so the added tests are true regressions. The first suite run surfaced one failure unrelated to this change — a pre-existing away-mode case using GNU-only `touch -d @epoch` on a BSD-touch host, which also made two neighbouring cases assert nothing; I fixed it with the repo's existing portable set_mtime idiom and re-ran, after which both suites are green (139 ok, 0 failures). No screenshots apply: the end-user surface here is the Stop-hook CLI transcript and the persisted state files, both captured verbatim.

<details>
<summary>Evidence: End-to-end before/after transcript of the deadlock (both variants)</summary>

Source: [End-to-end before/after transcript of the deadlock (both variants)](https://github.com/MortenGad/firstmate/blob/57848237b1691a643ee260e4b27507f5c739a1ae/.no-mistakes/evidence/fm/fm-turnend-arm-deadlock-d28/deadlock-before-after.txt)

```text
==============================================================================
BEFORE the fix - branch base b84e0e3
==============================================================================
seeded incident state:
    state/.claude-autoarm-epoch : epoch=165 owner_pid=41746 outcome=rewake updated_at=1788664081
    state/.turnend-claude-blocks: session=sess-deadlock count=1 epoch=165 
    watcher holding home lock   : none
    primed cycle processes      : 

--- Stop 1 ---------------------------------------------------------------
guard exit=2  (REFUSED - the turn cannot end)
●  1 task(s) in flight, but no live watcher holds this home lock (last beat: never).
●  The Stop-owned auto-arm did not claim this home either, so recovery is NOT already under way.
state immediately after the refusal:
    state/.claude-autoarm-epoch : epoch=165 owner_pid=41746 outcome=rewake updated_at=1788664081
    state/.turnend-claude-blocks: session=sess-deadlock count=1 epoch=165 
    watcher holding home lock   : none
    primed cycle processes      : 
waiting up to 10s for a watcher this Stop may have primed...
RESULT: no watcher ever came up.
state before the next Stop:
    state/.claude-autoarm-epoch : epoch=165 owner_pid=41746 outcome=rewake updated_at=1788664081
    state/.turnend-claude-blocks: session=sess-deadlock count=1 epoch=165 
    watcher holding home lock   : none
    primed cycle processes      : 

--- Stop 2 ---------------------------------------------------------------
guard exit=2  (REFUSED - the turn cannot end)
●  1 task(s) in flight, but no live watcher holds this home lock (last beat: never).
●  The Stop-owned auto-arm did not claim this home either, so recovery is NOT already under way.
state before the next Stop:
    state/.claude-autoarm-epoch : epoch=165 owner_pid=41746 outcome=rewake updated_at=1788664081
    state/.turnend-claude-blocks: session=sess-deadlock count=1 epoch=165 
    watcher holding home lock   : none
    primed cycle processes      : 

--- Stop 3 ---------------------------------------------------------------
guard exit=2  (REFUSED - the turn cannot end)
●  1 task(s) in flight, but no live watcher holds this home lock (last beat: never).
●  The Stop-owned auto-arm did not claim this home either, so recovery is NOT already under way.
state before the next Stop:
    state/.claude-autoarm-epoch : epoch=165 owner_pid=41746 outcome=rewake updated_at=1788664081
    state/.turnend-claude-blocks: session=sess-deadlock count=1 epoch=165 
    watcher holding home lock   : none
    primed cycle processes      : 

live watcher process:
    <none>
watcher lifecycle ledger (state/.watch-cycle-exits.log) after the cycle ends:
    <no cycle ever ran>

==============================================================================
AFTER the fix - branch head 915761e
==============================================================================
seeded incident state:
    state/.claude-autoarm-epoch : epoch=165 owner_pid=41746 outcome=rewake updated_at=1788664081
    state/.turnend-claude-blocks: session=sess-deadlock count=1 epoch=165 
    watcher holding home lock   : none
    primed cycle processes      : 

--- Stop 1 ---------------------------------------------------------------
guard exit=2  (REFUSED - the turn cannot end)
●  1 task(s) in flight, but no live watcher holds this home lock (last beat: never).
●  The Stop-owned auto-arm did not claim this home, so nothing owns recovery and restoring supervision is on you.
state immediately after the refusal:
    state/.claude-autoarm-epoch : epoch=165 owner_pid=41746 outcome=rewake updated_at=1788664081
    state/.turnend-claude-blocks: session=sess-deadlock count=2 epoch=165 
    watcher holding home lock   : none
    primed cycle processes      : 21638 
waiting up to 10s for a watcher this Stop may have primed...
RESULT: a watcher came up and holds the home lock.
state before the next Stop:
    state/.claude-autoarm-epoch : epoch=165 owner_pid=41746 outcome=rewake updated_at=1788664081
    state/.turnend-claude-blocks: session=sess-deadlock count=2 epoch=165 
    watcher holding home lock   : 21679
    primed cycle processes      : 21638 21679 21917 

--- Stop 2 ---------------------------------------------------------------
guard exit=0  (ALLOWED - the turn ends)
(no output)
state before the next Stop:
    state/.claude-autoarm-epoch : epoch=165 owner_pid=41746 outcome=rewake updated_at=1788664081
    state/.turnend-claude-blocks: <absent, reset by an allowed stop>
    watcher holding home lock   : 21679
    primed cycle processes      : 21638 21679 21917 

The turn ended normally - the deadlock is broken.

live watcher process:
    21679 21638 bash /private/var/folders/hg/5q3jmyw50ln9sxy7c8800vgr0000gn/T/fm-deadlock-repro.1yc8Fl/after/bin/fm-watch.sh
watcher lifecycle ledger (state/.watch-cycle-exits.log) after the cycle ends:
    arm_pid=21638	watcher_pid=21679	origin=started	started_at=1788907574	ended_at=1788907575	exit_code=143	signal=TERM	reason=arm-interrupted	beacon_age=1	lock_before=pid:21679|identity:Wed Sep  9 00:46:14 2026     bash /private/var/folders/hg/5q3jmyw50ln9sxy7c8800vgr0000gn/T/fm-deadlock-repro.1yc8Fl/after/bin/fm-watch.sh	lock_after=pid:none|identity:none	successor=none

==============================================================================
BEFORE the fix - advancing epoch (166 failed -> 167 rewake), branch base b84e0e3
==============================================================================
seeded state: fresh failed epoch 166 with its failure notice
    state/.claude-autoarm-epoch : epoch=166 owner_pid=999 outcome=failed updated_at=1788907576
    state/.turnend-claude-blocks: <absent, reset by an allowed stop>
    watcher holding home lock   : none
    primed cycle processes      : 

--- Stop A (fresh failed epoch 166 spends its one handoff) ----------------
guard exit=0  (ALLOWED - the turn ends)
(no output)
    state/.claude-autoarm-epoch : epoch=166 owner_pid=999 outcome=failed updated_at=1788907576
    state/.turnend-claude-blocks: session=sess-deadlock count=0 epoch=166 
    watcher holding home lock   : none
    primed cycle processes      : 

the auto-arm now advances the epoch to 167 outcome=rewake, which then ages out:
    state/.claude-autoarm-epoch : epoch=167 owner_pid=41746 outcome=rewake updated_at=2
    state/.turnend-claude-blocks: session=sess-deadlock count=0 epoch=166 
    watcher holding home lock   : none
    primed cycle processes      : 

--- Stop B ---------------------------------------------------------------
guard exit=2  (REFUSED - the turn cannot end)
●  1 task(s) in flight, but no live watcher holds this home lock (last beat: never).
●  The Stop-owned auto-arm did not claim this home either, so recovery is NOT already under way.
waiting up to 10s for a watcher this Stop may have primed...
RESULT: no watcher ever came up.
state before the next Stop:
    state/.claude-autoarm-epoch : epoch=167 owner_pid=41746 outcome=rewake updated_at=2
    state/.turnend-claude-blocks: session=sess-deadlock count=1 epoch=167 
    watcher holding home lock   : none
    primed cycle processes      : 

--- Stop C ---------------------------------------------------------------
guard exit=2  (REFUSED - the turn cannot end)
●  1 task(s) in flight, but no live watcher holds this home lock (last beat: never).
●  The Stop-owned auto-arm did not claim this home either, so recovery is NOT already under way.
state before the next Stop:
    state/.claude-autoarm-epoch : epoch=167 owner_pid=41746 outcome=rewake updated_at=2
    state/.turnend-claude-blocks: session=sess-deadlock count=1 epoch=167 
    watcher holding home lock   : none
    primed cycle processes      : 

--- Stop D ---------------------------------------------------------------
guard exit=2  (REFUSED - the turn cannot end)
●  1 task(s) in flight, but no live watcher holds this home lock (last beat: never).
●  The Stop-owned auto-arm did not claim this home either, so recovery is NOT already under way.
state before the next Stop:
    state/.claude-autoarm-epoch : epoch=167 owner_pid=41746 outcome=rewake updated_at=2
    state/.turnend-claude-blocks: session=sess-deadlock count=1 epoch=167 
    watcher holding home lock   : none
    primed cycle processes      : 

live watcher process:
    <none>

==============================================================================
AFTER the fix - advancing epoch (166 failed -> 167 rewake), branch head 915761e
==============================================================================
seeded state: fresh failed epoch 166 with its failure notice
    state/.claude-autoarm-epoch : epoch=166 owner_pid=999 outcome=failed updated_at=1788907626
    state/.turnend-claude-blocks: <absent, reset by an allowed stop>
    watcher holding home lock   : none
    primed cycle processes      : 

--- Stop A (fresh failed epoch 166 spends its one handoff) ----------------
guard exit=0  (ALLOWED - the turn ends)
(no output)
    state/.claude-autoarm-epoch : epoch=166 owner_pid=999 outcome=failed updated_at=1788907626
    state/.turnend-claude-blocks: session=sess-deadlock count=0 epoch=166 
    watcher holding home lock   : none
    primed cycle processes      : 

the auto-arm now advances the epoch to 167 outcome=rewake, which then ages out:
    state/.claude-autoarm-epoch : epoch=167 owner_pid=41746 outcome=rewake updated_at=2
    state/.turnend-claude-blocks: session=sess-deadlock count=0 epoch=166 
    watcher holding home lock   : none
    primed cycle processes      : 

--- Stop B ---------------------------------------------------------------
guard exit=2  (REFUSED - the turn cannot end)
●  1 task(s) in flight, but no live watcher holds this home lock (last beat: never).
●  The Stop-owned auto-arm did not claim this home, so nothing owns recovery and restoring supervision is on you.
waiting up to 10s for a watcher this Stop may have primed...
RESULT: a watcher came up and holds the home lock.
state before the next Stop:
    state/.claude-autoarm-epoch : epoch=167 owner_pid=41746 outcome=rewake updated_at=2
    state/.turnend-claude-blocks: session=sess-deadlock count=1 epoch=167 
    watcher holding home lock   : 28913
    primed cycle processes      : 28889 28913 29130 29168 

--- Stop C ---------------------------------------------------------------
guard exit=0  (ALLOWED - the turn ends)
(no output)
state before the next Stop:
    state/.claude-autoarm-epoch : epoch=167 owner_pid=41746 outcome=rewake updated_at=2
    state/.turnend-claude-blocks: <absent, reset by an allowed stop>
    watcher holding home lock   : 28913
    primed cycle processes      : 28889 28913 29130 

The turn ended normally - the deadlock is broken.

live watcher process:
    28913 28889 bash /private/var/folders/hg/5q3jmyw50ln9sxy7c8800vgr0000gn/T/fm-deadlock-repro.1yc8Fl/after-adv/bin/fm-watch.sh
```
</details>
<details>
<summary>Evidence: Reproducer that produced the transcript</summary>

Source: [Reproducer that produced the transcript](https://github.com/MortenGad/firstmate/blob/57848237b1691a643ee260e4b27507f5c739a1ae/.no-mistakes/evidence/fm/fm-turnend-arm-deadlock-d28/deadlock-repro.sh)

```text
#!/usr/bin/env bash
# End-to-end reproduction of the 2026-09-06 Claude Stop guard/auto-arm deadlock,
# run against the branch base (b84e0e3) and against the branch head (915761e).
#
# usage: deadlock-repro.sh <firstmate-checkout> <base-commit>
#
# Both homes are real primary-shaped Firstmate homes carrying the REAL bin/ tree
# of their respective commit (guard, auto-arm, fm-watch-arm.sh, fm-watch.sh) and
# the incident state observed on 2026-09-06:
#   - one task in flight
#   - no live watcher holding the home lock
#   - state/.claude-autoarm-epoch frozen at epoch=165 outcome=rewake
#   - state/.turnend-claude-blocks at count=1 epoch=165
# Each "Stop" is a genuine Claude Stop hook invocation: the guard runs as a child
# of a harness process that owns state/.lock, with the Stop payload on stdin.
# The registered asyncRewake auto-arm hook is NOT run after a refusal, because
# Claude Code cancels sibling hooks when a Stop is blocked - that cancellation is
# the deadlock this change is about.
set -u

ROOT=$1
BASE=$2
WORK=$(mktemp -d "${TMPDIR:-/tmp}/fm-deadlock-repro.XXXXXX")
WORK=$(cd "$WORK" && pwd -P)

cleanup() {
  local p
  for p in $(pgrep -f "$WORK" 2>/dev/null || true); do kill "$p" 2>/dev/null || true; done
  sleep 0.5
  for p in $(pgrep -f "$WORK" 2>/dev/null || true); do kill -9 "$p" 2>/dev/null || true; done
  rm -rf "$WORK"
}
trap cleanup EXIT

seed_home() {  # <dir> base|head
  local dir=$1 which=$2
  mkdir -p "$dir/state" "$dir/bin" "$dir/docs"
  git init -q "$dir"
  git -C "$dir" commit -q --allow-empty -m init
  : > "$dir/AGENTS.md"
  if [ "$which" = base ]; then
    git -C "$ROOT" archive "$BASE" bin | tar -x -C "$dir"
    git -C "$ROOT" archive "$BASE" docs/supervision-protocols | tar -x -C "$dir"
  else
    cp -R "$ROOT/bin/." "$dir/bin/"
    cp -R "$ROOT/docs/supervision-protocols" "$dir/docs/supervision-protocols"
  fi
  chmod +x "$dir"/bin/*.sh
  ln -s /bin/bash "$dir/fake-claude"
  : > "$dir/state/task1.meta"
  printf 'epoch=165 owner_pid=41746 outcome=rewake updated_at=1788664081\n' \
    > "$dir/state/.claude-autoarm-epoch"
  touch -t 202001010000 "$dir/state/.claude-autoarm-epoch"
  printf 'session=sess-deadlock\ncount=1\nepoch=165\n' > "$dir/state/.turnend-claude-blocks"
}

stop() {  # <dir> -> prints hook output on stdout, returns the hook's exit code
  local dir=$1 home
  home=$(cd "$dir" && pwd -P)
  # shellcheck disable=SC2016 # the fake harness expands FM_HOME inside its child shell
  printf '{"stop_hook_active":true,"session_id":"sess-deadlock"}' \
    | CLAUDECODE=1 FM_HOME="$home" "$dir/fake-claude" -c '
        printf "%s\n" "$$" > "$FM_HOME/state/.lock"
        bash "$FM_HOME/bin/fm-turnend-guard.sh" --claude
      ' 2>&1
}

report() {  # <dir>
  local dir=$1 pid
  pid=$(cat "$dir/state/.watch.lock/pid" 2>/dev/null || true)
  printf '    state/.claude-autoarm-epoch : %s\n' \
    "$(head -1 "$dir/state/.claude-autoarm-epoch" 2>/dev/null || echo '<absent, reset by an allowed stop>')"
  if [ -e "$dir/state/.turnend-claude-blocks" ]; then
    printf '    state/.turnend-claude-blocks: %s\n' "$(tr '\n' ' ' < "$dir/state/.turnend-claude-blocks")"
  else
    printf '    state/.turnend-claude-blocks: <absent, reset by an allowed stop>\n'
  fi
  printf '    watcher holding home lock   : %s\n' "${pid:-none}"
  printf '    primed cycle processes      : %s\n' \
    "$(pgrep -f "$dir/bin/fm-watch" 2>/dev/null | tr '\n' ' ' || true)"
}

await_watcher() {  # <dir>
  local dir=$1 i=0 pid
  while [ "$i" -lt 200 ]; do
    pid=$(cat "$dir/state/.watch.lock/pid" 2>/dev/null || true)
    if [ -n "$pid" ] && kill -0 "$pid" 2>/dev/null && [ -e "$dir/state/.last-watcher-beat" ]; then
      return 0
    fi
    sleep 0.05
    i=$((i + 1))
  done
  return 1
}

run_home() {  # <dir> <title>
  local dir=$1 title=$2 out status n pid p
  printf '==============================================================================\n'
  printf '%s\n' "$title"
  printf '==============================================================================\n'
  printf 'seeded incident state:\n'
  report "$dir"
  printf '\n'
  for n in 1 2 3; do
    out=$(stop "$dir"); status=$?
    printf -- '--- Stop %s ---------------------------------------------------------------\n' "$n"
    printf 'guard exit=%s  (%s)\n' "$status" \
      "$([ "$status" -eq 2 ] && echo 'REFUSED - the turn cannot end' || echo 'ALLOWED - the turn ends')"
    if [ -n "$out" ]; then printf '%s\n' "$out"; else printf '(no output)\n'; fi
    if [ "$n" -eq 1 ]; then
      printf 'state immediately after the refusal:\n'
      report "$dir"
      printf 'waiting up to 10s for a watcher this Stop may have primed...\n'
      if await_watcher "$dir"; then
        printf 'RESULT: a watcher came up and holds the home lock.\n'
      else
        printf 'RESULT: no watcher ever came up.\n'
      fi
    fi
    printf 'state before the next Stop:\n'
    report "$dir"
    printf '\n'
    if [ "$status" -eq 0 ]; then
      printf 'The turn ended normally - the deadlock is broken.\n\n'
      break
    fi
  done
  pid=$(cat "$dir/state/.watch.lock/pid" 2>/dev/null || true)
  printf 'live watcher process:\n'
  if [ -n "$pid" ] && kill -0 "$pid" 2>/dev/null; then
    ps -o pid=,ppid=,command= -p "$pid" | sed 's/^/    /'
  else
    printf '    <none>\n'
  fi
  for p in $(pgrep -f "$dir/bin/fm-watch" 2>/dev/null || true); do kill "$p" 2>/dev/null || true; done
  sleep 1
  printf 'watcher lifecycle ledger (state/.watch-cycle-exits.log) after the cycle ends:\n'
  if [ -s "$dir/state/.watch-cycle-exits.log" ]; then
    sed 's/^/    /' "$dir/state/.watch-cycle-exits.log"
  else
    printf '    <no cycle ever ran>\n'
  fi
  printf '\n'
}

seed_home "$WORK/before" base
run_home "$WORK/before" "BEFORE the fix - branch base b84e0e3"
seed_home "$WORK/after" head
run_home "$WORK/after" "AFTER the fix - branch head 915761e"

# --- second variant: the epoch ADVANCES (166 failed -> 167 rewake) -------------
# The 07:30-09:00 episode the same day recovered on its own several times instead
# of locking up, and its epoch advanced rather than freezing. The fresh failed
# epoch 166 spends its one automatic handoff (a Stop the guard ALLOWS); once the
# rewake epoch 167 ages past the freshness window, every later Stop refuses.
seed_advancing() {  # <dir> base|head
  local dir=$1 which=$2
  seed_home "$dir" "$which"
  rm -f "$dir/state/.turnend-claude-blocks"
  : > "$dir/state/.claude-autoarm-failure-notified"
  printf 'epoch=166 owner_pid=999 outcome=failed updated_at=%s\n' "$(date +%s)" \
    > "$dir/state/.claude-autoarm-epoch"
}

run_advancing() {  # <dir> <title>
  local dir=$1 title=$2 out status n p pid
  printf '==============================================================================\n'
  printf '%s\n' "$title"
  printf '==============================================================================\n'
  printf 'seeded state: fresh failed epoch 166 with its failure notice\n'
  report "$dir"
  printf '\n'
  out=$(stop "$dir"); status=$?
  printf -- '--- Stop A (fresh failed epoch 166 spends its one handoff) ----------------\n'
  printf 'guard exit=%s  (%s)\n' "$status" \
    "$([ "$status" -eq 2 ] && echo 'REFUSED - the turn cannot end' || echo 'ALLOWED - the turn ends')"
  if [ -n "$out" ]; then printf '%s\n' "$out"; else printf '(no output)\n'; fi
  report "$dir"
  printf '\n'
  printf 'the auto-arm now advances the epoch to 167 outcome=rewake, which then ages out:\n'
  printf 'epoch=167 owner_pid=41746 outcome=rewake updated_at=2\n' > "$dir/state/.claude-autoarm-epoch"
  touch -t 202001010000 "$dir/state/.claude-autoarm-epoch"
  report "$dir"
  printf '\n'
  for n in B C D; do
    out=$(stop "$dir"); status=$?
    printf -- '--- Stop %s ---------------------------------------------------------------\n' "$n"
    printf 'guard exit=%s  (%s)\n' "$status" \
      "$([ "$status" -eq 2 ] && echo 'REFUSED - the turn cannot end' || echo 'ALLOWED - the turn ends')"
    if [ -n "$out" ]; then printf '%s\n' "$out"; else printf '(no output)\n'; fi
    if [ "$n" = B ]; then
      printf 'waiting up to 10s for a watcher this Stop may have primed...\n'
      if await_watcher "$dir"; then
        printf 'RESULT: a watcher came up and holds the home lock.\n'
      else
        printf 'RESULT: no watcher ever came up.\n'
      fi
    fi
    printf 'state before the next Stop:\n'
    report "$dir"
    printf '\n'
    if [ "$status" -eq 0 ] && [ "$n" != A ]; then
      printf 'The turn ended normally - the deadlock is broken.\n\n'
      break
    fi
  done
  pid=$(cat "$dir/state/.watch.lock/pid" 2>/dev/null || true)
  printf 'live watcher process:\n'
  if [ -n "$pid" ] && kill -0 "$pid" 2>/dev/null; then
    ps -o pid=,ppid=,command= -p "$pid" | sed 's/^/    /'
  else
    printf '    <none>\n'
  fi
  for p in $(pgrep -f "$dir/bin/fm-watch" 2>/dev/null || true); do kill "$p" 2>/dev/null || true; done
  sleep 1
  printf '\n'
}

# Each home is seeded immediately before its own run: the failed epoch 166 must
# still be inside FM_CLAUDE_AUTOARM_EPOCH_FRESH (15s) when its first Stop lands.
seed_advancing "$WORK/before-adv" base
run_advancing "$WORK/before-adv" "BEFORE the fix - advancing epoch (166 failed -> 167 rewake), branch base b84e0e3"
seed_advancing "$WORK/after-adv" head
run_advancing "$WORK/after-adv" "AFTER the fix - advancing epoch (166 failed -> 167 rewake), branch head 915761e"
```
</details>
<details>
<summary>Evidence: Evidence index (marks two earlier-round artifacts superseded)</summary>

Source: [Evidence index (marks two earlier-round artifacts superseded)](https://github.com/MortenGad/firstmate/blob/57848237b1691a643ee260e4b27507f5c739a1ae/.no-mistakes/evidence/fm/fm-turnend-arm-deadlock-d28/README-evidence.md)

```text
# Evidence index

## Current for branch head 915761e

- `deadlock-before-after.txt` - end-to-end before/after transcript of the
  2026-09-06 deadlock, produced by `deadlock-repro.sh`. Real Firstmate homes
  carrying the real `bin/` tree of the branch base (b84e0e3) and of the branch
  head (915761e); every "Stop" is a real `fm-turnend-guard.sh --claude` hook
  invocation under a harness process that owns `state/.lock`. Covers both
  variants the intent names: the frozen epoch (165 rewake) and the advancing
  epoch (166 failed -> 167 rewake).
- `deadlock-repro.sh` - the reproducer. Run as
  `deadlock-repro.sh <firstmate-checkout> b84e0e362face25f3dd8945297a3df1320d7668c`.

## Superseded - do not read as head behaviour

- `banner-honesty-e2e.txt` / `banner-honesty-e2e.sh` (captured at e823344)
- `turnend-deadlock-e2e.txt` / `turnend-deadlock-e2e.sh` (captured against base 6d396da)

Both were captured before d8e407d collapsed the refusal banner to one
unconditional sentence, so they show a banner line that no longer exists:
"This refusal primed a detached watcher start that survives it ...". They also
predate 915761e, which moved priming below the attended fail-open. The banner
text at head is the one in `deadlock-before-after.txt`.
```
</details>
<details>
<summary>Evidence: Key excerpt: same seeded incident state, base vs head</summary>

```text
BEFORE the fix - branch base b84e0e3
seeded: epoch=165 owner_pid=41746 outcome=rewake | blocks count=1 epoch=165 | no watcher
Stop 1: guard exit=2 (REFUSED) -> RESULT: no watcher ever came up. blocks count=1 epoch=165
Stop 2: guard exit=2 (REFUSED) blocks count=1 epoch=165
Stop 3: guard exit=2 (REFUSED) blocks count=1 epoch=165
watcher lifecycle ledger: <no cycle ever ran>

AFTER the fix - branch head 915761e
seeded: epoch=165 owner_pid=41746 outcome=rewake | blocks count=1 epoch=165 | no watcher
Stop 1: guard exit=2 (REFUSED) primed cycle processes: 21638
-> RESULT: a watcher came up and holds the home lock (pid 21679)
blocks count=2 epoch=165 (the bound is no longer pinned to a frozen epoch)
epoch unchanged: epoch=165 owner_pid=41746 outcome=rewake (no generation claim taken)
Stop 2: guard exit=0 (ALLOWED - the turn ends), no output
live watcher: 21679 21638 bash <home>/bin/fm-watch.sh
ledger: arm_pid=21638 watcher_pid=21679 origin=started ... successor=none

ADVANCING EPOCH (166 failed -> 167 rewake)
base: Stop A allowed (handoff) -> Stop B/C/D all REFUSED, no watcher ever
head: Stop A allowed with no primed cycle -> Stop B REFUSED and primes -> Stop C allowed
```
</details>
<details>
<summary>Evidence: New coverage fails against the base bin/ tree (regression proof)</summary>

```text
$ cp <base>/bin/fm-turnend-guard.sh bin/ ; cp <base>/bin/fm-claude-stop-autoarm.sh bin/
$ bash tests/fm-turnend-guard.test.sh
...
not ok - the watcher primed by the refused control stop never beat:
suite exit=1

$ git checkout -- bin/ && bin/fm-test-run.sh tests/fm-turnend-guard.test.sh tests/fm-claude-stop-autoarm.test.sh
ok - fm-turnend-guard --claude: an ordinary stop does not prime and leaves downtime a primed cycle would consume
ok - fm-turnend-guard --claude: a refused stop cannot guarantee the next refusal (frozen-epoch variant)
ok - fm-turnend-guard --claude: a refused stop still recovers when the epoch does not advance
ok - fm-turnend-guard --claude: a refused stop cannot guarantee the next refusal (advancing-epoch variant)
ok - fm-turnend-guard --claude: the attended fail-open allows without priming a handling successor
FM_TEST_SUMMARY total=2 failed=0 skipped_gate=0 duration_ms=143408
```
</details>
- Outcome: ⚠️ 2 infos across 2 runs (56m6s)

## Pipeline

_... (51 earlier update rounds omitted to keep the PR body within GitHub's 65536-char limit; full history is in the run log.)_

<details>
<summary>⚠️ **Review** - 4 infos</summary>

Stating the trade-off honestly so you can decide rather than me: the un-primed wording (&#34;recovery is NOT already under way&#34;) is a genuinely stronger act-now signal than the primed one, so collapsing to a single sentence does lose a real distinction. The narrower forms are either (a) keep only the original wording and drop the whole contract, or (b) print the neutral &#34;no generation claim owns recovery yet&#34; unconditionally, which is true whether or not a cycle was detached and also needs no exit status. Recommending removal of the component rather than hardening it; this is deliberate authored content you have shaped across several rounds, so it is your call.
- ℹ️ `bin/fm-turnend-guard.sh:298` - A blocked Stop still fails to advance state/.turnend-claude-blocks when autoarm_owns_recovery already accounted the same epoch earlier in that same Stop without incrementing, and docs/turnend-guard.md:129 now asserts the opposite. (Raised in round 9 and deferred by your &#34;leave them for a later response&#34;; code unchanged.)

Concrete state: session S, tasks in flight, no watcher, state/.claude-autoarm-failure-notified present, state/.claude-autoarm-epoch = `epoch=42 owner_pid=&lt;dead&gt; outcome=failed updated_at=T` with mtime T, and state/.turnend-claude-blocks = `session=S count=1 epoch=42` from the previous Stop. A Stop at T+3: the `failed` branch of autoarm_owns_recovery (:367-372) sees age 3 &lt; EPOCH_FRESH (15) and the notice present, so it calls budget_account_current_epoch with no argument. Inside, blocked=0 and old_epoch == current_epoch == 42, so :297-299 takes the no-op branch and COUNT stays 1 - yet :327 still sets BUDGET_ACCOUNTED=1. BUDGET_INITIALIZED_FAILURE is 0 (the file exists and the session matches), so the branch falls through, the guard refuses, and at :509 `budget_account_current_epoch blocked-stop` finds old_epoch == current_epoch and BUDGET_ACCOUNTED == 1 and skips the increment as well. The Stop is a consecutive block that advanced nothing. The `failed-suppressed` branch at :374-379 has the identical shape.

docs/turnend-guard.md:129 states &#34;a Stop that reaches the refusal with no epoch progress still advances the count&#34;, which this sequence contradicts.

Impact is bounded, which is why this is info rather than a repeat of the 2026-09-06 pin: the epoch file&#39;s mtime ages past EPOCH_FRESH within 15 seconds and nothing refreshes it without a new generation, so accounting resumes on the next Stop; the cost is at most one extra blocked Stop before the attended fail-open, far below Claude&#39;s 8-block override. The observed 2026-09-06 incident (outcome=rewake, no notice) never reaches the accounting call at all and is fixed correctly by this change.

Remedy: set BUDGET_ACCOUNTED only when COUNT was actually incremented, rather than whenever accounting ran, and leave docs/turnend-guard.md:129 as written.
- ℹ️ `bin/fm-claude-stop-autoarm.sh:84` - The --ensure-watcher header block states a fact about the shipped banner that stopped being true two commits ago, and it is the stated justification for the whole exit-status contract above it.

Line 84-85 reads &#34;The guard&#39;s refusal banner tells the model a recovery cycle is in flight only on the 0, so that claim can never outrun the fork.&#34; Commit 1474acb removed the printf at bin/fm-turnend-guard.sh:256 that made that claim. The banner the guard now prints on PRIMED_RECOVERY=1 is only &#34;The Stop-owned auto-arm did not claim this home, so no generation claim owns recovery yet.&#34; - it does not tell the model a recovery cycle is in flight, and docs/turnend-guard.md:104 explicitly records that it &#34;deliberately says nothing about the detached start&#39;s prospects&#34;.

A maintainer reading this header concludes the guard announces an in-flight recovery on exit 0 and may either rely on that announcement or reintroduce it. Correct the sentence to say what the exit status is actually keyed to - which of the two missing-claim wordings the banner selects - or drop it entirely if the wording split is removed per the simplification finding above.
- ℹ️ `tests/fm-turnend-guard.test.sh:31` - The new file-level trap overrides EXIT, INT, and TERM but leaves HUP and QUIT pointing at the handler in tests/lib.sh, which removes the temp root without stopping the real detached cycle the new case starts.

tests/lib.sh:177-181 installs handlers for EXIT, INT, TERM, HUP, and QUIT. tests/fm-turnend-guard.test.sh:31-33 replaces only the first three with turnend_guard_cleanup, so `trap &#39;fm_test_cleanup; exit 129&#39; HUP` and the QUIT equivalent survive from the lib and never call stop_real_primed_cycle.

Concrete sequence: test_hook_claude_mode_ordinary_stop_does_not_prime has set PRIMED_CYCLE_HOME and is in its positive-control phase, so a real bin/fm-watch-arm.sh and bin/fm-watch.sh copied by install_real_watch_stack are running against the fixture home. The terminal hangs up (SIGHUP) or the user sends SIGQUIT. The lib handler runs fm_test_cleanup, which rm -rf&#39;s the fixture directory, and exits - while the primed cycle keeps running against the deleted tree. It does not receive the signal itself, because the priming path deliberately setsid&#39;s it into its own session (bin/fm-claude-stop-autoarm.sh:215), which is exactly the property the production change relies on.

Remedy: add the two missing lines next to :32-33, `trap &#39;turnend_guard_cleanup; exit 129&#39; HUP` and `trap &#39;turnend_guard_cleanup; exit 131&#39; QUIT`, matching the exit codes tests/lib.sh already uses.

🔧 Fix: collapse the refusal banner to one unconditional sentence
2 warnings still open:

- ⚠️ `bin/fm-turnend-guard.sh:491` - Priming is placed above `budget_account_current_epoch blocked-stop || block_stop`, exactly as instructed, but that line is not the refusal. `terminal_fail_open` runs afterwards and can ALLOW the same Stop, so a primed handling successor still lands on an allowed Stop - the outcome the recorded decision required be confined to refusals.

Concrete sequence: session S, tasks in flight, no watcher, `state/.claude-autoarm-failure-notified` present, epoch `outcome=failed-suppressed` aged past `FM_CLAUDE_AUTOARM_EPOCH_FRESH`, `state/.turnend-claude-blocks` at `count=3 epoch=E`, no `.afk`, no `FAILURE_ALARM`, and `state/.watcher-down` holding `pending:downtime:&lt;gen&gt;`. `autoarm_owns_recovery` returns 1 (stale epoch takes no accounting branch), so line 491 primes a detached arm with `FM_WATCH_PREDECESSOR_ARM_PID` set. Line 501 raises COUNT to 4. `terminal_fail_open` then passes every gate (`4 &gt; 3`, `failure_episode_verified` true, no alarm, no open claim), writes `FAILURE_ALARM`, and returns 0, so line 515 exits 0 - the Stop is ALLOWED. Claude therefore does not cancel the registered asyncRewake sibling; it arms, finds the primed watcher already holding `state/.watch.lock`, and attaches (bin/fm-watch-arm.sh:438). That watcher is running with `FM_WATCH_HANDLING_SUCCESSOR=1` (bin/fm-watch-arm.sh:481-482), so `fm_recovery_marker_arm_check` still flips the marker to `announced:downtime:&lt;gen&gt;` (bin/fm-watch.sh:1635) while `WATCHER_RECOVERY_PENDING` is forced to 0 (:1639-1640) and `resurface_after_downtime` returns immediately for the whole cycle (:1757-1763). The once-per-generation re-presentation of the unacknowledged downtime episode never fires, with no error anywhere.

`terminal_fail_open` returning 2 reaches the same place through line 517, and there the priming call can also change the verdict of the Stop that started it: if the primed watcher wins the lock before the health re-check at :419, that check reports healthy, `fm_failure_episode_reset` runs, and the loud attended fail-open is skipped entirely.

This contradicts the invariant the change asserts in three places - bin/fm-turnend-guard.sh:83 and :477 (&#34;Priming is refusal-path only&#34;, &#34;Prime recovery only on the refusal path&#34;) and docs/turnend-guard.md:98.

Remedy needs your call because it moves the block you explicitly placed: either relocate priming to immediately before the final `block_stop` at line 518 (true refusal-path only, at the cost of the fail-open decision no longer seeing a cycle this Stop started), or keep the placement and correct the three invariant statements to say priming runs on the terminal path, which usually but not always refuses.
- ⚠️ `tests/fm-claude-stop-autoarm.test.sh:1354` - Removing the `--ensure-watcher` exit-status contract in d8e407d left two inertness tests asserting nothing that can fail. Their only remaining evidence is a filesystem check evaluated one statement after the hook returns, but priming is asynchronous by construction, so both would pass with the gate they exist to prove deleted.

`test_ensure_watcher_inert_when_afk` previously asserted `expect_code 1` - which only the non-forking gates produced - and now asserts `expect_code 0`, the value every path returns. The surviving assertion at :1354 is `[ ! -e &#34;$dir/state/.watch.lock/pid&#34; ]`. If the `[ -e &#34;$STATE/.afk&#34; ] &amp;&amp; exit 0` gate at bin/fm-claude-stop-autoarm.sh:152 were removed, the hook would fork through python3, exec `bin/fm-watch-arm.sh`, and only then reach the fixture watcher that creates the lock dir - hundreds of milliseconds later. The file&#39;s own positive control proves the timing: `test_ensure_watcher_starts_detached_watcher_without_claim` needs a 50-iteration, 5-second poll loop to observe that same lock appear. `test_ensure_watcher_inert_without_session_lock` (:1264-1265) has the identical shape against the session-lock gate at :142-149.

This also removed the last non-racy proof of away-mode priming inertness: d8e407d deleted `test_hook_claude_mode_away_refusal_claims_no_primed_start`, which checked `primed_cycle_pids` and the lock through the guard&#39;s real refusal path. The same pattern exists at tests/fm-turnend-guard.test.sh:2092, where the allowed handoff Stop&#39;s `primed_cycle_pids` check runs immediately rather than over a window.

Remedy is mechanical: give each of the three a bounded observation window instead of a single instantaneous check, matching what `test_hook_claude_mode_ordinary_stop_does_not_prime` already does at tests/fm-turnend-guard.test.sh:1927-1939 (watch for ~1s and fail on the first trace that appears), and in the autoarm suite watch for the arm process as well as the lock so the check is not blind to a fork whose watcher has not started yet.

🔧 Fix: prime only on the true refusal, below the fail-open
4 infos still open:

- ℹ️ `bin/fm-turnend-guard.sh:479` - Relocating the priming block in 915761e opened one refusal path that no longer primes. `budget_account_current_epoch blocked-stop || block_stop` refuses directly on the `||` branch, and that branch is now ABOVE the priming block at :497 - before the move (d8e407d) priming ran first, so this same refusal did prime.

Reachable sequence: the registered asyncRewake sibling fires on the same Stop and calls `fm_failure_episode_reset` (bin/fm-wake-lib.sh:1205-1207), which takes `state/.turnend-claude-blocks.lock`. If it holds that lock at the moment the guard reaches :272, `fm_lock_try_acquire` returns 1, `budget_account_current_epoch` returns 1, and the guard refuses with no detached cycle started - the one outcome the change exists to prevent. The same happens if the `printf`/`mv -f` of the budget file fails (:308-313).

It is self-healing rather than a durable deadlock: `fm_lock_try_acquire` steals a lock whose holder is dead (bin/fm-wake-lib.sh:913-930), so the next Stop primes normally. Remedy is mechanical - factor the priming block at :497-521 into a function and call it from both refusal sites, e.g. `budget_account_current_epoch blocked-stop || { prime_recovery; block_stop; }` - and does not move the block back above `terminal_fail_open`.
- ℹ️ `tests/fm-turnend-guard.test.sh:2145` - Two of the three assertions in `test_hook_claude_mode_fail_open_allows_without_priming_a_cycle` cannot fail, and they claim coverage the same change argues elsewhere is impossible with a stub.

The case uses `write_gated_watch_fixture` (:1955), whose fm-watch.sh spins on `state/release-watch` - which this case never creates - and which never calls `fm_recovery_marker_arm_check`. So `[ ! -e &#34;$dir/state/.watch.lock/pid&#34; ]` (:2145) is unreachable even if priming ran, and the `pending:downtime:failopen.1.aaa` marker check (:2146-2149) is unreachable for the same reason. The change states this itself at `install_real_watch_stack` (:1830-1834): &#34;A stub watcher cannot perform it, so &#39;the downtime episode is still unconsumed&#39; is only a real assertion over the real watcher.&#34;

The case still catches its regression through `primed_cycle_pids` (:2143), which matches the detached arm as soon as it execs - and docs/verification/supervision.md:462 records exactly that, &#34;the forked arm and watcher pids named in its message&#34;. So the test is not broken; its stated coverage is wider than what it proves. Smallest honest remedy: drop the two assertions that cannot fail, or switch this case to `install_real_watch_stack` so they mean what the comment at :2120-2125 says.
- ℹ️ `tests/fm-turnend-guard.test.sh:2116` - The new test was inserted between an existing comment and the function it documents, so the comment now heads the wrong case.

Lines 2116-2119 (&#34;The 2026-09-06 budget half of the same deadlock: state/.turnend-claude-blocks stayed count=1 epoch=165 for forty minutes ... A frozen auto-arm epoch must not pin the bound.&#34;) described `test_hook_claude_mode_frozen_epoch_still_reaches_the_bounded_fail_open`, which is now at :2153. It reads as a preamble to `test_hook_claude_mode_fail_open_allows_without_priming_a_cycle` (:2126), which is about priming placement and not about the pinned count at all. Move lines 2116-2119 down to immediately above :2153, leaving the new comment at :2120-2125 with its own case.
- ℹ️ `tests/fm-turnend-guard.test.sh:33` - The new file-level trap overrides EXIT, INT, and TERM but leaves HUP and QUIT pointing at the handler in tests/lib.sh, which removes the temp root without stopping the real detached cycle the new case starts.

tests/lib.sh:177-181 installs handlers for EXIT, INT, TERM, HUP, and QUIT. tests/fm-turnend-guard.test.sh:31-33 replaces only the first three with `turnend_guard_cleanup`, so `trap &#39;fm_test_cleanup; exit 129&#39; HUP` and the QUIT equivalent survive from the lib and never call `stop_real_primed_cycle`.

Concrete sequence: `test_hook_claude_mode_ordinary_stop_does_not_prime` has set `PRIMED_CYCLE_HOME` and is in its positive-control phase, so a real bin/fm-watch-arm.sh and bin/fm-watch.sh copied by `install_real_watch_stack` are running against the fixture home. The terminal hangs up (SIGHUP) or the user sends SIGQUIT. The lib handler runs `fm_test_cleanup`, which rm -rf&#39;s the fixture directory, and exits - while the primed cycle keeps running against the deleted tree. It does not receive the signal itself, because the priming path deliberately setsid&#39;s it into its own session (bin/fm-claude-stop-autoarm.sh:196-208), which is exactly the property the production change relies on.

Remedy: add the two missing lines next to :32-33, `trap &#39;turnend_guard_cleanup; exit 129&#39; HUP` and `trap &#39;turnend_guard_cleanup; exit 131&#39; QUIT`, matching the exit codes tests/lib.sh already uses. (Raised in round 10 and never selected by a human; still present at branch head.)

</details>

<details>
<summary>⚠️ **Test** - 2 infos</summary>

✅ No issues found.
- <code>`bash tests/fm-turnend-guard.test.sh` — full guard suite including the six new `--claude` cases (ordinary stop does not prime, frozen-epoch/advancing-epoch/no-epoch-progress refusal variants, away-mode refusal claims no primed start, frozen epoch still reaches the bounded fail-open)</code>
- <code>`bash tests/fm-claude-stop-autoarm.test.sh` — full auto-arm suite including the four new `--ensure-watcher` cases (detaches a watcher with no generation claim, records a refused primed cycle in the lifecycle ledger, inert without a session lock, inert while away mode owns supervision)</code>
- <code>Manual end-to-end incident replay against real `bin/` at base 6d396da vs HEAD e823344 over real Claude Stop boundaries: `bash ~/.no-mistakes/evidence/01M1XBNK8V52GE36GK363C65BD/turnend-deadlock-e2e.sh &lt;worktree&gt; 6d396da`</code>
- <code>Manual end-to-end banner-honesty check at HEAD across a primeable home, an away-mode home, and a home whose `state/.lock` names a live foreign claude-shaped pid: `bash ~/.no-mistakes/evidence/01M1XBNK8V52GE36GK363C65BD/banner-honesty-e2e.sh &lt;worktree&gt;`</code>
- <code>Verified the primed cycle is genuinely detached and live (`ps -o pid=,ppid=,args=` on the process holding `state/.watch.lock`) and that a primed cycle blocked by a wedged stale-beacon holder still writes its `state/.watch-cycle-exits.log` row naming that holder</code>

- ℹ️ `tests/fm-turnend-guard.test.sh:2204` - Working-tree change made by this test phase: tests/fm-turnend-guard.test.sh gained the repo&#39;s existing portable `set_mtime` helper (copied in shape from tests/fm-inactive-reconcile.test.sh) and its three `touch -d &#34;@$beat&#34;` call sites now use it. `touch -d @&lt;epoch&gt;` is GNU-only; on this macOS host BSD touch rejects it, left the beacon fresh, and made test_hook_away_daemon_blocks_beacon_older_than_poll_derived_grace fail while two neighbouring poll-derived-grace cases passed vacuously. This is unrelated to the deadlock fix and pre-exists the branch base; it is a host-portability fix, not a behaviour change, and it lands in this PR&#39;s diff.
- ℹ️ Two evidence artifacts from earlier rounds of this run are stale and will be linked from the PR: banner-honesty-e2e.txt (captured at e823344) and turnend-deadlock-e2e.txt (captured against base 6d396da) both show the refusal-banner line &#34;This refusal primed a detached watcher start that survives it ...&#34;, which d8e407d removed when the banner collapsed to one unconditional sentence. They also predate 915761e moving priming below the attended fail-open. I left them in place and added README-evidence.md to the evidence directory marking them superseded and naming deadlock-before-after.txt as the head-accurate transcript; deleting them is the user&#39;s call.
- <code>`bin/fm-test-run.sh tests/fm-turnend-guard.test.sh tests/fm-claude-stop-autoarm.test.sh` — 139 ok, FM_TEST_SUMMARY total=2 failed=0 skipped_gate=0</code>
- <code>End-to-end before/after reproduction of the 2026-09-06 incident with real homes and real Stop hooks: `deadlock-repro.sh &lt;checkout&gt; b84e0e362face25f3dd8945297a3df1320d7668c` (frozen epoch 165 rewake, and advancing epoch 166 failed -&gt; 167 rewake), transcript captured</code>
- <code>Regression check that the new coverage fails before the fix: copied the base-commit `bin/fm-turnend-guard.sh` and `bin/fm-claude-stop-autoarm.sh` into the worktree, ran `bash tests/fm-turnend-guard.test.sh` (first priming assertion fails: `not ok - the watcher primed by the refused control stop never beat`), then restored with `git checkout -- bin/`</code>
- <code>Diagnosed the initial `test_hook_away_daemon_blocks_beacon_older_than_poll_derived_grace` failure to BSD touch (`touch -d &#34;@1700000000&#34;` → `out of range or illegal time specification`), applied the repo&#39;s portable `set_mtime` idiom in `tests/fm-turnend-guard.test.sh`, and re-ran the suite (`bin/fm-test-run.sh tests/fm-turnend-guard.test.sh` → 93 ok, exit 0)</code>
- <code>Verified no primed watcher/arm processes and no temp homes were left behind after each repro run (`ps -eo pid,command | grep fm-deadlock-repro`), and that the worktree carries only the intentional test change (`git status --porcelain`)</code>

</details>

<details>
<summary>🔧 **Document** - 1 issue found → auto-fixed (2) ✅</summary>

- ℹ️ `docs/verification/supervision.md:386` - Judgment call left for the phase that actually runs the suites: docs/verification/supervision.md records dated evidence entries for the two closest precedents to this change - the 2026-08-02 Claude auto-arm false-failure, guard-predicate, and bounded fail-open correction (line 356) and the 2026-08-04 model-aware pull-guard correction (line 372) - each with the exact commands (bin/fm-lint.sh, bin/fm-doc-audience-check.sh, bin/fm-test-run.sh over the four isolated supervision suites) and observed output. This deadlock fix is the same class of guard/auto-arm correction and has no such entry. I did not add one: an honest entry needs a real lint, doc-check, and suite run with captured output, which belongs to the validation phase rather than the documentation phase, and inventing a date or a FM_TEST_SUMMARY line would make the file lie. The deterministic proof itself is already recorded in docs/turnend-guard.md &#34;Regression coverage&#34; and docs/watcher-continuity.md, so nothing is undocumented - the open question is only whether this change also warrants a dated maintainer-verification record alongside its two precedents.

🔧 Fix: record 2026-09-08 deadlock-fix verification evidence

- ⚠️ `docs/verification/supervision.md:416` - The working tree carries an uncommitted change to tests/fm-turnend-guard.test.sh (beyond the stated target commit 915761e) that replaces the GNU-only `touch -d &#34;@&lt;epoch&gt;&#34;` calls in the three away-mode poll-derived-grace cases with a portable `set_mtime` helper. That is exactly the host-portability failure this verification entry narrates at length. Two present-tense claims in the entry become false the moment that change is committed: line 416 says the failure &#34;remains open against tests/fm-turnend-guard.test.sh rather than against this correction&#34;, and line 460 says the observed `touch -d` failure &#34;is still the only observation of that case&#34;. I did not edit them. The historical narration of the 2026-09-08 run stays true either way, but an honest update needs a real full-script run of tests/fm-turnend-guard.test.sh on the tree that carries the helper, and inventing an FM_TEST_SUMMARY line or a second `ok` block would make this evidence file lie - the same reason Round 1 of this step left the entry to a phase that actually runs the suites. Decision needed: either the test fix is dropped (entry is already correct), or it lands and someone re-runs `bin/fm-test-run.sh tests/fm-claude-stop-autoarm.test.sh tests/fm-guard-stale-banner.test.sh tests/fm-turnend-guard.test.sh tests/fm-supervision-instructions.test.sh` so lines 414-417 and 460 can be replaced with observed output.
- ℹ️ `docs/turnend-guard.md:97` - Line 97 states that when the wait produces no recovery proof the guard &#34;primes bin/fm-claude-stop-autoarm.sh --ensure-watcher immediately before refusing&#34;. After 915761e moved the priming call below terminal_fail_open, one Claude-path refusal no longer primes: `budget_account_current_epoch blocked-stop || block_stop` (bin/fm-turnend-guard.sh:479) refuses directly when fm_lock_try_acquire on state/.turnend-claude-blocks.lock fails, and that block_stop returns before the priming block at :513. Before the relocation that refusal did prime. The consequence is bounded and self-correcting - the budget lock is a micro-mutex held across a few file operations, and the next Stop primes normally - so it is not a fifth unbounded entry condition alongside the four the doc enumerates. I did not add a qualifier to line 97 because the recorded round-11 decision instructs that the surrounding invariant statements be re-read and confirmed rather than rewritten, and I would rather surface the imprecision than edit against that instruction. Your call whether the doc should name the exception or the code should prime on that path too.

🔧 Fix: record the passing full-script re-run; name the priming exception

</details>

<details>
<summary>🔧 **Lint** - 1 issue found ✅</summary>

✅ No issues found.

</details>

<details>

✅ No issues found.

✅ No issues found.

</details>


MortenGad:
Handing this over for review. It is rebased onto current `main` and **13 of 14 checks pass**.

The one red check is **"PR must be raised via no-mistakes"**, and it is red for a reason
outside this branch: `no-mistakes` refuses to start its pipeline against this repository at
all. Every agent it names as supported — `grok`, `codex`, `claude` — is refused before the
first step with:

```
gate agent "<agent>" does not neutralize the target repository's project
agent-instruction files (AGENTS.md/CLAUDE.md); refusing to launch it in the target checkout
```

Reproduced on v1.64.0 and again on v1.72.0. The same install runs fine against another
repository minutes later, so it is repo-specific rather than a version or configuration
problem. Details and the likely cause are in #4152.

Because the pipeline could not run, the rebase was pushed directly, which is precisely what
leaves that provenance check red. No behaviour changed in the rebase; the reviewed content
is unchanged from before `main` moved.

One correction to the record on this PR: an earlier run here was reported with
`Behavior portable serial 4` cancelled on a 20-minute timeout, and that shard **passes on
this head**. The full run is green apart from the provenance check above.

Nothing here needs action from me that I can take — clearing that check needs either a fix
for #4152 or a maintainer with a working pipeline. Happy to rebase again or revalidate the
moment it is runnable.

PR #4030, opened 2026-09-08
fix(bin): reclaim locks whose owning process is gone

dbeihl-utilicast:
## Intent

The captain has asked for this fixed: a lock whose owning process has died is never reclaimed, so whatever tries to take it next waits out its full timeout and then fails.

It is filed as issue 3966 against the status-presentation lock, but that issue understates it. This is the shared lock helper rather than one lock. In the last twelve hours it has stranded at least five different locks: the status-presentation lock, the home-summary refresh lock, a task metadata lock, a pending-reply lock, and a remote-reply lifecycle lock. Each time the owner directory held a pid that was long gone.

What it has cost, so you understand why this is worth doing properly rather than quickly. Supervision died overnight and nothing watched the fleet for about nine hours. It has failed six or seven more times today, each one a window where nothing was supervising. Every recovery was manual: read the lock, confirm the owner is dead, remove it by hand.

The fix is in the shared helper, not in any one caller. A waiter that finds a lock held by a process that no longer exists should be able to reclaim it rather than waiting for a timeout that will never be satisfied.

Three things it must get right, and they are the whole difficulty.

Reclaiming must not race. Two waiters can decide simultaneously that an owner is dead, and both must not end up believing they hold the lock. Whatever you build, show what happens when two arrive at once rather than reasoning that it is unlikely.

A pid alone is not proof of death. Pids are reused, so a live unrelated process can occupy a dead owner's number and make a stale lock look held forever. Decide what actually identifies the owner and say why it is sound.

Do not turn a stale lock into a stolen one. A slow owner is not a dead owner. Whatever test you use has to distinguish a process that is busy from one that no longer exists, and getting that wrong is worse than the bug, because it means two writers in a section meant for one.

Prove it can fail. Plant a lock owned by a dead process, show a waiter hanging on it today, then show it reclaimed after the fix. Then plant one owned by a live but slow process and show it is still respected. The second test is the one that matters; without it the fix is indistinguishable from removing the lock.

## What Changed

- `fm_lock_try_acquire` in `bin/fm-wake-lib.sh` now reclaims a lock only when the recorded owner is provably gone: acquisition writes a `pid-identity` record beside `pid`, and `fm_lock_owner_is_abandoned` treats an owner as abandoned only when its pid is dead or its start-time identity no longer matches. A live pid, or one whose identity cannot be read, stays held, so a slow owner is never stolen from. Reclaim still runs through the existing `.steal` lock, so two waiters that decide simultaneously cannot both take ownership.
- Owner identity dropped its argv component in both `bin/fm-wake-lib.sh` (`fm_pid_identity`) and `bin/fm-extension.mjs` (`pidIdentity`), leaving process start time alone: `exec` keeps the pid and start time while replacing argv, so an argv-based identity read a live holder as dead. Start time still makes a reused pid a mismatch. Owner-record writes clear the identity before changing the pid and are read back for verification.
- `fm_lock_try_create` / `fm_lock_try_acquire` / `fm_lock_acquire_wait{,_bounded}` now separate contention (rc 1) from owner-record creation failure (rc 2), and ~20 callers across `bin/` check the acquisition result before entering the protected section instead of assuming success. `docs/verification/supervision.md` documents the failure contract and the per-caller behavior; `tests/fm-lock-failure.test.sh` (new) and `tests/fm-watcher-lock.test.sh` add coverage for dead-owner reclaim, a reused pid, a live-but-slow owner, a holder that `exec`s, missing identity, lost creation races, and interrupted handoff.

## Risk Assessment

⚠️ Medium: The reclaim mechanism, the rc=1/2 contract and all three intent constraints verify correct under direct probing with strong executable regressions, but this is a shared lock-helper contract change rewired through roughly forty call sites across the whole fleet, so residual blast radius warrants a follow-up pass rather than a clean-bill low.

## Testing

Ran the two lock suites (fm-watcher-lock, fm-lock-failure) and the wake-queue suite that regressed in round 1 — all green, including the presentation-lock case that failed last round. Then demonstrated the intent end to end at the CLI level by planting a lock whose owner was SIGKILLed and whose pid number was taken over by a live unrelated process, and running the same five scenarios against a base b84e0e3 copy and against HEAD: base leaves `bin/fm-watch.sh` printing \"watcher: already running pid 99747\" with nothing supervising and a bounded waiter failing at 124 after its full window, while HEAD reclaims and supervises and the waiter acquires in 0s; on both trees a busy live owner and a live owner that exec()s keep their locks, and at HEAD eight simultaneous reclaimers of one stale lock yield exactly one winner where base yields none. Both new regressions were proven capable of red against the code they were written against (rc=124 at base for reused-pid reclaim, rc=0 lock theft at 4a32ff1 for the exec case). No UI surface is involved, so the artifacts are CLI transcripts rather than screenshots. One pre-existing flaky test in tests/fm-wake-queue.test.sh reproduces at base too and is already tracked as issue 4012.

<details>
<summary>Evidence: Before/after CLI transcript: stale lock reclaim across base b84e0e3 and HEAD 6ac892f</summary>

Source: [Before/after CLI transcript: stale lock reclaim across base b84e0e3 and HEAD 6ac892f](https://github.com/dbeihl-utilicast/firstmate/blob/c925bf363a41cb75b076572025759780fb004e82/.no-mistakes/evidence/fm/fm-stale-lock-reclaim/stale-lock-before-after.txt)

<code>############ BEFORE the fix - base b84e0e3&#10;&#10;--- 1. Supervision start behind a stale lock whose pid got reused (issue 3966) ---&#10;the watcher that owned this lock (pid 95831) was killed; its pid slot is now&#10;occupied by an unrelated live process, pid 99747&#10;$ bin/fm-watch.sh&#10;watcher: already running pid 99747&#10;RESULT: watcher refused to start; the stale lock still reads as held -&gt; NOTHING IS SUPERVISING&#10;&#10;--- 2. A waiter blocking on that same stale lock ---&#10;$ fm_lock_acquire_wait_bounded state/.presentation.lock 5 # 5 second deadline&#10;exit=124 after 5s (0 = acquired, 124 = deadline expired)&#10;RESULT: waiter waited out its whole window and FAILED on a lock nobody holds&#10;&#10;--- 5. Eight processes reclaim the SAME stale lock simultaneously ---&#10;RESULT: 0 of 8 simultaneous reclaimers acquired the lock&#10;&#10;############ AFTER the fix - HEAD 6ac892f&#10;&#10;--- 1. Supervision start behind a stale lock whose pid got reused (issue 3966) ---&#10;the watcher that owned this lock (pid 42014) was killed; its pid slot is now&#10;occupied by an unrelated live process, pid 44599&#10;$ bin/fm-watch.sh&#10;check: rearm-resurface&#10;RESULT: watcher RECLAIMED the stale lock and is supervising (lock pid now 44600)&#10;&#10;--- 2. A waiter blocking on that same stale lock ---&#10;$ fm_lock_acquire_wait_bounded state/.presentation.lock 5 # 5 second deadline&#10;exit=0 after 0s (0 = acquired, 124 = deadline expired)&#10;RESULT: waiter RECLAIMED the stale lock&#10;&#10;--- 3. A LIVE but slow owner must still be respected ---&#10;$ FM_LOCK_STALE_AFTER=0 fm_lock_try_acquire state/.slow.lock&#10;exit=1 held-by=49398&#10;RESULT: slow live owner KEPT its lock (lock pid still 49398)&#10;&#10;--- 4. A LIVE owner that exec()s must still be respected ---&#10;live owner pid 49769 has exec&#39;d &#39;sleep 60&#39;: same pid, same start time, new argv&#10;$ FM_LOCK_STALE_AFTER=0 fm_lock_try_acquire state/.exec.lock&#10;exit=1 held-by=49769&#10;RESULT: exec&#39;ing live owner KEPT its lock (lock pid still 49769)&#10;&#10;--- 5. Eight processes reclaim the SAME stale lock simultaneously ---&#10;winner: pid 52592&#10;RESULT: 1 of 8 simultaneous reclaimers acquired the lock&#10;&#10;--- 6. Control: an owner killed with no pid reuse ---&#10;$ fm_lock_try_acquire state/.plain.lock&#10;exit=0 RESULT: reclaimed</code>

```text
Issue 3966 - stale lock reclaim, end to end on this machine (2026-09-08 18:25 EDT, darwin 25.5.0, no /proc)

############ BEFORE the fix - base b84e0e3

--- 1. Supervision start behind a stale lock whose pid got reused (issue 3966) ---
    the watcher that owned this lock (pid 95831) was killed; its pid slot is now
    occupied by an unrelated live process, pid 99747
$ bin/fm-watch.sh
    /var/folders/j5/kgdthy6n37j7dgv0r7qr3qd80000gn/T//fm3966-7qahl3/s1.out:watcher: already running pid 99747
    RESULT: watcher refused to start; the stale lock still reads as held -> NOTHING IS SUPERVISING

--- 2. A waiter blocking on that same stale lock ---
    owner pid 211 is gone; its number is now held by unrelated live pid 3013
$ fm_lock_acquire_wait_bounded state/.presentation.lock 5      # 5 second deadline
    exit=124 after 5s        (0 = acquired, 124 = deadline expired)
    RESULT: waiter waited out its whole window and FAILED on a lock nobody holds

--- 3. A LIVE but slow owner must still be respected ---
    live owner pid 9181 is busy inside its critical section, not releasing
$ FM_LOCK_STALE_AFTER=0 fm_lock_try_acquire state/.slow.lock   # most aggressive reclaim setting
    exit=1 held-by=9181
    RESULT: slow live owner KEPT its lock (lock pid still 9181)

--- 4. A LIVE owner that exec()s must still be respected ---
    live owner pid 9485 has exec'd 'sleep 60': same pid, same start time, new argv
$ FM_LOCK_STALE_AFTER=0 fm_lock_try_acquire state/.exec.lock
    exit=1 held-by=9485
    RESULT: exec'ing live owner KEPT its lock (lock pid still 9485)

--- 5. Eight processes reclaim the SAME stale lock simultaneously ---
    RESULT: 0 of 8 simultaneous reclaimers acquired the lock

--- 6. Control: an owner killed with no pid reuse ---
    owner pid 41448 was SIGKILLed; nothing has taken its pid number
$ fm_lock_try_acquire state/.plain.lock
    exit=0  RESULT: reclaimed


############ AFTER the fix - HEAD 6ac892f

--- 1. Supervision start behind a stale lock whose pid got reused (issue 3966) ---
    the watcher that owned this lock (pid 42014) was killed; its pid slot is now
    occupied by an unrelated live process, pid 44599
$ bin/fm-watch.sh
    /var/folders/j5/kgdthy6n37j7dgv0r7qr3qd80000gn/T//fm3966-SSkwb0/s1.out:check: rearm-resurface
    RESULT: watcher RECLAIMED the stale lock and is supervising (lock pid now 44600)

--- 2. A waiter blocking on that same stale lock ---
    owner pid 45779 is gone; its number is now held by unrelated live pid 48918
$ fm_lock_acquire_wait_bounded state/.presentation.lock 5      # 5 second deadline
    exit=0 after 0s        (0 = acquired, 124 = deadline expired)
    RESULT: waiter RECLAIMED the stale lock

--- 3. A LIVE but slow owner must still be respected ---
    live owner pid 49398 is busy inside its critical section, not releasing
$ FM_LOCK_STALE_AFTER=0 fm_lock_try_acquire state/.slow.lock   # most aggressive reclaim setting
    exit=1 held-by=49398
    RESULT: slow live owner KEPT its lock (lock pid still 49398)

--- 4. A LIVE owner that exec()s must still be respected ---
    live owner pid 49769 has exec'd 'sleep 60': same pid, same start time, new argv
$ FM_LOCK_STALE_AFTER=0 fm_lock_try_acquire state/.exec.lock
    exit=1 held-by=49769
    RESULT: exec'ing live owner KEPT its lock (lock pid still 49769)

--- 5. Eight processes reclaim the SAME stale lock simultaneously ---
    winner: pid 52592
    RESULT: 1 of 8 simultaneous reclaimers acquired the lock

--- 6. Control: an owner killed with no pid reuse ---
    owner pid 95764 was SIGKILLed; nothing has taken its pid number
$ fm_lock_try_acquire state/.plain.lock
    exit=0  RESULT: reclaimed
```
</details>
<details>
<summary>Evidence: Negative controls: each new regression run against the code it was written against</summary>

Source: [Negative controls: each new regression run against the code it was written against](https://github.com/dbeihl-utilicast/firstmate/blob/c925bf363a41cb75b076572025759780fb004e82/.no-mistakes/evidence/fm/fm-stale-lock-reclaim/regression-negative-controls.txt)

<code>=== 1. Reused-pid reclaim, HEAD test vs BASE library (b84e0e3) ===&#10;not ok - reused-pid lock waited instead of reclaiming: rc=124 held=8555&#10;&#10;=== 1b. Same test against HEAD library (6ac892f) ===&#10;ok - reused-pid stale lock reclaims while an identity-matched slow owner holds&#10;&#10;=== 2. Live exec-ing holder, HEAD test vs the PREVIOUS commit library (4a32ff1) ===&#10;not ok - a live holder lost its lock after exec replaced its command line: rc=0 held=&#10;&#10;=== 2b. Same test against HEAD library (6ac892f) ===&#10;ok - a live holder that execs keeps its lock</code>

```text
Negative controls: each new regression test run against the library it was written
against, to show it can return red. The test file is HEAD's; only bin/ is swapped.

=== 1. Reused-pid reclaim, HEAD test vs BASE library (b84e0e3) ===
$ cd <base tree> && bash tests/fm-watcher-lock.test.sh   # test_lock_reclaims_reused_pid_and_respects_slow_owner
    not ok - reused-pid lock waited instead of reclaiming: rc=124 held=8555

=== 1b. Same test against HEAD library (6ac892f) ===
    ok - reused-pid stale lock reclaims while an identity-matched slow owner holds

=== 2. Live exec-ing holder, HEAD test vs the PREVIOUS commit library (4a32ff1) ===
$ cd <4a32ff1 tree> && bash tests/fm-watcher-lock.test.sh   # test_live_holder_that_execs_keeps_its_lock
    not ok - a live holder lost its lock after exec replaced its command line: rc=0 held=

=== 2b. Same test against HEAD library (6ac892f) ===
    ok - a live holder that execs keeps its lock
```
</details>
- Outcome: ⚠️ 1 info across 2 runs (1h40m11s)

## Pipeline

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>

- 🚨 `bin/fm-wake-lib.sh:471` - `fm_lock_prepare_owner` writes the new short-format `fm_lock_pid_identity` value into `$ownerdir/pid-identity`, a file that already has a different semantic owner: `fm_pid_identity` output (which appends `cmdline-hex=`/`command`). `bin/fm-supervise-daemon.sh:1529` and `bin/fm-watch.sh:1731` both overwrite that same file with the LONG format immediately after acquiring their lock through `fm_lock_try_acquire`. Failing sequence, deterministic and race-free: (1) fm-supervise-daemon.sh:1515 acquires `.supervise-daemon.lock`, prepare_owner records `linux-starttime=99123`; (2) line 1529 overwrites it with `linux-starttime=99123 cmdline-hex=...`; (3) a second daemon startup calls `fm_lock_try_acquire`, and `fm_lock_owner_is_abandoned` (line 592) compares the recorded LONG string against `fm_lock_pid_identity`&#39;s SHORT string, which can never match, so the live daemon&#39;s singleton lock is declared abandoned and stolen. The identical sequence applies to `.watch.lock` via fm-watch.sh:1731, and on that path the steal also publishes a false `.watcher-down` downtime marker (fm-wake-lib.sh:1000) before evicting the live watcher. There is also a narrower window between lock creation and the overwrite in which `fm_watcher_lock_matches_pid` (line 174) reads the SHORT format and compares it against `fm_pid_identity`, so a healthy watcher reads as unhealthy. This directly produces the outcome the intent forbids (&#34;Do not turn a stale lock into a stolen one... two writers in a section meant for one&#34;). Earliest shared boundary for the fix: give the lock-ownership identity its own filename (e.g. `lock-owner-identity`), add it to `fm_lock_clean_known_files` (line 421), and update the two new tests; alternatively make the lock path consume `fm_pid_identity` so one format serves one file. Remedy is mechanical and corrects what the change already does.
- ⚠️ `bin/fm-wake-lib.sh:113` - `fm_lock_pid_identity`&#39;s non-/proc fallback pins `LC_ALL=C` but not `TZ`, and `ps -o lstart=` renders the process start instant in local wall-clock time. This is the branch that runs on Darwin (no /proc), the fleet&#39;s primary platform. If the recorded identity is written under one TZ and re-read under another — a DST transition while a long-lived watcher holds `.watch.lock`, a TZ=UTC launchd/CI process writing while an interactive shell reads, or a manual timezone change — the rendered string differs for the same live process, `fm_lock_owner_is_abandoned` reports abandoned, and a live owner&#39;s lock is stolen. The sibling `fm_pid_identity` already documents this hazard in its own comment (&#34;immune to the wall-clock steps that re-render the ps lstart fallback&#39;s date ... and would evict a live watcher&#34;), but the new function inherits the fallback without the mitigation. Fix is one token: `LC_ALL=C TZ=UTC ps -p &#34;$pid&#34; -o lstart=`, which makes the rendering both locale- and offset-invariant.
- ⚠️ `bin/fm-wake-lib.sh:592` - `IFS= read -r recorded &lt; &#34;$lockdir/pid-identity&#34; || return 1` has no `2&gt;/dev/null`, so when the file is absent bash emits `fm-wake-lib.sh: line 592: &lt;path&gt;/pid-identity: No such file or directory` to stderr (verified with a standalone bash snippet: message printed, rc=1). The absent-file state is reachable in current code, not just across an upgrade: `bin/fm-supervise-daemon.sh:1530` explicitly `rm -f`s `$LOCK/pid-identity` when `fm_pid_identity` fails, leaving a held lock with `pid` but no identity. Every waiter then polls through `fm_lock_acquire_wait` at 10 Hz and each poll writes an error line, flooding stderr in hook and log contexts that treat stderr as a failure channel. The abandonment verdict itself is correct (conservative: treated as held), so this is noise rather than a wrong result. Suppress the redirect diagnostic, e.g. guard with `[ -r &#34;$lockdir/pid-identity&#34; ] || return 1` before the read.
- ℹ️ `bin/fm-wake-lib.sh:469` - `fm_lock_prepare_owner` previously verified its write by reading the pid back (`back=$(cat &#34;$ownerdir/pid&#34;); [ &#34;$back&#34; = &#34;$mypid&#34; ]`); the rewrite checks only that the two `printf` redirections returned success. `fm_lock_claim` also dropped its own write-plus-readback of the pid. On a filesystem where a write reports success but does not land (the case the original readback appears to have been defending against), the lock is now published with an unverified or empty owner record, and `fm_lock_owner_is_abandoned` then reads an empty pid and reports abandoned. No concrete reachable path is proven here, so this is a note about a deliberately removed defensive check rather than a demonstrated defect; confirm the removal was intended.
- ℹ️ `bin/fm-wake-lib.sh:1053` - `_fm_lock_acquire_wait_handoff` updates `$ownerdir/pid` to the caller and only then `$ownerdir/pid-identity`, while the lock is already published. A waiter that lands between the two writes reads pid=caller (alive) against the helper&#39;s still-recorded identity, computes a mismatch, and enters the steal path. Traced through: the steal path re-reads the identity after acquiring `$lockdir.steal` (several forks later) and normally sees the completed write, and even if the helper is descheduled long enough for the steal to complete, the handoff&#39;s own readback at line 1055-1056 finds its owner dir gone, releases, and returns 1, so double ownership does not result. The cost is a spurious handoff failure and lock churn rather than a correctness break. Recording pid and identity in a single atomic record would close it, but that is a state-shape change beyond this fix&#39;s scope; noting the tradeoff only.
- ℹ️ `tests/fm-watcher-lock.test.sh:241` - `test_lock_stale_steal_single_winner_under_concurrency` was converted from a genuinely dead pid to a live pid with a planted mismatched identity, so no concurrency test now covers a lock that has `pid` but no `pid-identity` file at all. That shape is reachable today (fm-supervise-daemon.sh:1530 removes the file; locks held across a deploy of this change never had it) and it is the shape whose abandonment verdict depends entirely on the `read` failure path at fm-wake-lib.sh:592. A case planting `pid` alone with a live reused pid, asserting the lock is respected rather than stolen, would pin that fallback.

🔧 Fix: restore owner-record readbacks in lock prepare and claim
2 errors still open:

- 🚨 `bin/fm-wake-lib.sh:484` - `fm_lock_prepare_owner` now hard-fails whenever `fm_lock_pid_identity` cannot answer for SELF (line 484 -&gt; `fm_lock_cache_self_identity` line 464). Because `fm_lock_try_create` cannot publish an owner record, NO lock can ever be acquired on a host where process identity is unreadable, and `fm_lock_acquire_wait` (line 1041, `while ! fm_lock_try_acquire; do sleep 0.1; done`) then spins forever. Reproduced directly, not inferred: with `ps` stubbed to `exit 1` and `FM_PROC_ROOT_OVERRIDE` at a nonexistent dir, base commit b84e0e3 prints `BASE try_acquire=ok pid=95021`, while HEAD prints `identity_rc=1` and then blows up (see the companion recursion finding); `fm_lock_acquire_wait` was killed after 120s without acquiring. That exact environment is a supported, documented configuration (`docs/configuration.md:878`) and is installed by the repo&#39;s own fixtures at `tests/fm-procevent.test.sh:1367-1374` and `:1445`, which drive code that calls the unbounded `fm_lock_acquire_wait &#34;$EXTENSION_LIFECYCLE_LOCK&#34;` at `bin/fm-procevent.sh:228`. It is also reachable in production: on Darwin there is no `/proc`, so any hook or launchd context whose PATH lacks `ps` loses every lock permanently. This inverts the intent&#39;s own goal - `&#34;a waiter ... waits out its full timeout and then fails&#34;` becomes a waiter that never returns at all. Fix at the write boundary rather than each caller: when identity is unavailable, publish the pid record with an absent/empty identity instead of refusing to publish. `fm_lock_owner_is_abandoned` (line 603) already handles that shape conservatively - a missing or empty `pid-identity` returns 1 (treated as held) - so degrading preserves both base&#39;s acquisition behavior and the new identity-based reclaim, and no test currently covers the identity-unavailable acquire path.
- 🚨 `bin/fm-wake-lib.sh:979` - `fm_lock_try_acquire` recurses into `fm_lock_try_acquire &#34;$steal&#34;` (line 979) whenever `fm_lock_try_create` fails, without distinguishing &#34;another holder owns this lock&#34; from &#34;this process cannot create a lock at all&#34;. When creation fails for a non-contention reason, every recursion level fails identically and recurses again on `$lockdir.steal.steal.steal...`, so the depth is bounded only by PATH_MAX. Observed in the probe above: one `fm_lock_try_acquire` call produced roughly 250 recursion levels and 4.4 MB of `basename: ... File name too long` on stderr before unwinding, and `fm_lock_acquire_wait` repeats that storm every 100ms forever. The shape predates this branch, but it was only reachable through rare write failures; the new identity hard-fail at line 484 makes it reachable on an ordinary supported configuration, and it stays reachable independently of that fix through any create failure on a full or read-only state dir (`mktemp -d` at line 456, or the record write at line 470). Bound the steal recursion - e.g. refuse to recurse when `$lockdir` does not exist, since a lock nobody holds needs no steal mutex - so a create failure returns 1 instead of amplifying into a fork-and-stderr storm.

🔧 Fix: degrade missing pid identity and bound steal recursion
1 warning still open:

- ⚠️ `bin/fm-wake-lib.sh:1059` - `_fm_lock_acquire_wait_handoff` still hard-fails when process birth identity is unreadable: `caller_identity=$(fm_lock_pid_identity &#34;$caller_pid&#34;) || return 1` returns before the helper ever calls `fm_lock_acquire_wait`. The round-2 fix degraded exactly this shape at `fm_lock_cache_self_identity` (:464), `fm_lock_prepare_owner` (:484) and `fm_lock_claim` (:532), but left the handoff sibling on the old hard-fail, so `fm_lock_acquire_wait_bounded` loses its whole bounded wait on that configuration. Reproduced against three trees with a stubbed failing `ps` plus `FM_PROC_ROOT_OVERRIDE` at an absent dir, contending on a live-held lock with a 3s window: base b84e0e3 returns `rc=124 held=&lt;pid&gt;` after 3s; HEAD returns `rc=1 held=` in 0s; HEAD with only :1059 changed to `fm_lock_pid_identity &#34;$caller_pid&#34; 2&gt;/dev/null || caller_identity=` returns `rc=124 held=&lt;pid&gt;` after 3s again. The wrong rc changes caller behavior, it is not just a timing loss: `bin/fm-wake-drain.sh:629-638` treats 124 as the benign `WAKE DRAIN SKIPPED: queue lock remains held by live pid ... retry on the next drain` + `exit 0`, and anything else as `wake drain: queue lock could not be acquired safely` on stderr + `exit 1`. `print_status_presentation` (fm-wake-drain.sh:583-593) and `bin/fm-x-lib.sh:1021` degrade the same way. The new test `test_lock_acquires_and_stays_closed_without_process_identity` (tests/fm-watcher-lock.test.sh:415) sets the no-identity environment only for the `fm_lock_try_acquire` leg; its `fm_lock_acquire_wait_bounded` leg runs with identity available, so this path is uncovered. Remedy is the one-line degrade above, mirroring what prepare/claim already do; empty identity is already handled conservatively by `fm_lock_owner_is_abandoned` (:609-610, treated as held), so it cannot widen the steal window.

🔧 Fix: degrade handoff identity to preserve bounded wait
3 issues (1 error, 1 warning, 1 info) still open:

- 🚨 `bin/fm-wake-lib.sh:613` - `fm_lock_owner_is_abandoned` compares the recorded `$lockdir/pid-identity` against `fm_lock_pid_identity &#34;$pid&#34;`, but that file is a PRE-EXISTING contract owned by `fm_pid_identity`, whose format is strictly different. `fm_pid_identity` (:58) emits `linux-starttime=&lt;ticks&gt; cmdline-hex=&lt;hex&gt;` on /proc hosts and `ps -o lstart= -o command=` elsewhere; the new `fm_lock_pid_identity` (:97) emits `linux-starttime=&lt;ticks&gt;` / `ps -o lstart=` only. Measured on this machine for the same pid: rich=[Tue Sep  8 13:38:18 2026     sleep 20] vs lock=[Tue Sep  8 13:38:18 2026] -&gt; DIFFERENT. Two production holders acquire their lock through the shared helper and then OVERWRITE `pid-identity` in the rich format: `bin/fm-watch.sh:1606` acquires `$STATE/.watch.lock` and `:1731` writes `fm_pid_identity &#34;$WATCHER_PID&#34;` into it; `bin/fm-supervise-daemon.sh:1515` acquires `$STATE/.supervise-daemon.lock` and `:1529` writes `fm_pid_identity` into it. Any later `fm_lock_try_acquire` on those locks then reads a rich-format record, computes the short format for the same live pid, sees a mismatch, and returns &#34;abandoned&#34; for a fully healthy owner. Reproduced end-to-end at HEAD with a holder that acquires via `fm_lock_acquire_wait` and then performs the fm-watch.sh:1731 overwrite: the holder stayed alive (`alive=yes`, `lock pid=[89488]`) and a second process printed `WAITER: STOLE lock from live holder, new pid=89602`. For `.watch.lock` the steal path additionally publishes a `.watcher-down` downtime recovery marker (fm_lock_try_acquire, :1020-1023), so a healthy watcher is evicted AND recorded as an outage. This contradicts the intent&#39;s explicit forbidden behavior: &#34;Do not turn a stale lock into a stolen one. A slow owner is not a dead owner... getting that wrong is worse than the bug, because it means two writers in a section meant for one.&#34; The collision is symmetric and also breaks the other direction: `fm_lock_write_owner_record` (:469) now plants the SHORT format in `$lockdir/pid-identity` at acquire time, so `fm_watcher_lock_matches_pid` (:174-179, same file) and `fm_watcher_healthy` report a live identity-matched watcher as unmatched until fm-watch.sh:1731 overwrites it, and a lock stolen by a non-watcher keeps the short format permanently. REMEDY NEEDS AUTHORIZATION, not the defect: two owners of one filename now compete, and picking between them is a design call. Option A - drop `fm_lock_pid_identity` and have `fm_lock_owner_is_abandoned`/`fm_lock_write_owner_record` use the existing `fm_pid_identity`, matching every other reader and writer in the repo (fm-afk-start.sh:88, fm-watch-arm.sh:229, fm-afk-launch.sh:97, fm-procevent-lib.sh:636), at the cost of the &#34;survives exec&#34; property the new function&#39;s comment claims. Option B - give the lock helper its own filename (e.g. `$lockdir/lock-identity`, added to `fm_lock_clean_known_files`) so the two identity contracts stop sharing storage. Either way the shared-boundary fix belongs here, not in another per-caller patch.
- ⚠️ `tests/fm-watcher-lock.test.sh:345` - No test in the change plants a `pid-identity` in the format the production writers actually use, which is why the live-lock steal above survived three fix rounds. Every new case either plants a literal dummy (`&#39;dead-owner-identity&#39;`, :356/:249), plants an empty file (:466), or lets the lock helper write its own short-format record (`test_lock_reclaims_reused_pid_and_respects_slow_owner`&#39;s slow-owner leg, :396-406). The repo&#39;s own shared fixtures already build the untested shape: `tests/lib.sh:91`, `tests/fm-turnend-guard.test.sh:276-287` and `tests/fm-secondmate-harness.test.sh:1022-1028` all plant `$STATE/.watch.lock/pid-identity` using `fm_pid_identity`, and that is exactly the state a running `fm-watch.sh` leaves behind. The missing negative control is: hold a lock with a live process, write `fm_pid_identity &#34;$holder&#34;` into `$lockdir/pid-identity` (mirroring fm-watch.sh:1731), then assert `fm_lock_try_acquire` returns 1 with `FM_LOCK_HELD_PID=$holder` and the lock pid is unchanged. Against HEAD that assertion fails. This test is gated on which remedy is chosen for the collision finding, so it needs the same authorization.
- ℹ️ `bin/fm-wake-lib.sh:613` - `fm_lock_owner_is_abandoned` replaces a fork-free `fm_pid_alive` (`kill -0`, a builtin) with a command substitution around `fm_lock_pid_identity`, which forks `cat /proc/&lt;pid&gt;/stat` on Linux and `ps -p &lt;pid&gt;` on Darwin/Git Bash. It sits on `fm_lock_acquire_wait`&#39;s 10 Hz spin (:1047-1051) and runs on every poll for every waiter, and up to twice per poll when the recorded pid equals the caller&#39;s (:956 short-circuits into the recheck at :972). This file already documents fork cost on that hot path as a deliberate concern (&#34;forking uname per call is a measurable cost on the platform (Git Bash/MSYS) that already pays the highest fork price&#34;, :11-13). The added cost is real but modest relative to the mktemp/dirname/basename forks `fm_lock_try_create` already pays each poll, and the correctness need for the probe is genuine, so this is a noted tradeoff rather than something to fix now; a per-(pid,identity) memo inside the wait loop would remove it if the spin ever shows up in a profile.

🔧 Fix: use fm_pid_identity for lock owner identity
1 warning still open:

- ⚠️ `bin/fm-wake-lib.sh:585` - `fm_lock_owner_is_abandoned` reads the owner identity with `IFS= read -r recorded &lt; &#34;$lockdir/pid-identity&#34; || return 1` and no stderr guard, so a missing file makes bash print `bash: &lt;path&gt;/pid-identity: No such file or directory` on stderr. The verdict is still correct (the redirect failure makes `read` fail, so the function returns 1 = held), so this is pure noise, not a wrong result. Two reachable paths. (1) Routine release race on the 10 Hz spin: `fm_lock_try_acquire` reads `$lockdir/pid` with `2&gt;/dev/null`, sees a live pid, and only then reaches line 585; if the holder&#39;s `fm_lock_release` executes `rm -f &#34;$lockdir&#34;` in between, `$lockdir/pid-identity` no longer resolves and every `fm_lock_acquire_wait` poll landing in that window emits the error. (2) Deterministic: `bin/fm-supervise-daemon.sh:1529-1531` does `rm -f &#34;$LOCK/pid-identity&#34;` when `fm_pid_identity` fails and then keeps running as the live lock holder, so every waiter poll on `.supervise-daemon.lock` emits the error for the daemon&#39;s whole lifetime. That stderr reaches user-facing callers: `bin/fm-guard.sh:101`, `bin/fm-captain-hold.sh:281`, `bin/fm-backlog-handoff.sh:703`. The file&#39;s own convention already guards this exact shape: `fm_recovery_marker_read` (bin/fm-wake-lib.sh:607) tests `[ -f &#34;$marker&#34; ]` before its `read &lt;`. Remedy: add `[ -r &#34;$lockdir/pid-identity&#34; ] || return 1` immediately before the read. Do NOT append `2&gt;/dev/null` after the `&lt;` redirect - redirections are processed left to right, so the redirect error fires before stderr is redirected. Verified on this machine: `bash -c &#39;IFS= read -r r &lt; /tmp/missing-abc 2&gt;/dev/null&#39;` still prints the error, while `bash -c &#39;IFS= read -r r 2&gt;/dev/null &lt; /tmp/missing-abc&#39;` is silent.

🔧 Fix: Silence missing lock identity reads without weakening ownership checks

- 🚨 `tests/zz-tmp-runner.test.sh:1` - `tests/zz-tmp-runner.test.sh` is a stray scratch artifact committed by the last fix round (c678953). It is a byte-for-byte copy of `tests/fm-watcher-lock.test.sh` lines 1-1498 with every test invocation stripped except a single trailing `test_lock_owner_without_identity_is_held_quietly`; its own header comment still reads `# tests/fm-watcher-lock.test.sh`, and it is mode 100644 while every other test file in the repo is 100755. It is not inert: `bin/fm-test-run.sh:1076` selects the suite with `for f in tests/*.test.sh`, `family_for_basename` leaves it `unclassified` so it lands in the derived portable-serial lane (the coverage guard at :968 therefore does not reject it), and `:2187` invokes it with `bash &#34;$script&#34;` regardless of the exec bit. Concrete effect on the next CI run: 1499 duplicated lines of watcher-lock test source get parsed and one test re-executed under `TMP_ROOT=$(fm_test_tmproot fm-watcher-lock-tests)` - the same tmproot name the real file claims - and every future edit to `tests/fm-watcher-lock.test.sh` silently diverges from this copy. Remedy: `git rm tests/zz-tmp-runner.test.sh`. Nothing in the change references it and no assertion in it is unique to it.
- ⚠️ `bin/fm-wake-lib.sh:925` - The recursion guard added by an earlier fix round (`if [ ! -e &#34;$lockdir&#34; ] &amp;&amp; [ ! -L &#34;$lockdir&#34; ]; then return 1; fi`) is broader than the failure it was added for and converts a lost creation race into a wrong &#34;held by nobody&#34; refusal. Demonstrated on this machine by sourcing the base library and the HEAD library into the same fresh state dir and forcing exactly one `ln -s &lt;lockdir&gt;` failure (the EEXIST a losing racer sees): base printed `rc=0 acquired pid=1167`, HEAD printed `rc=1 held=[]`. The reachable sequence: process A passes `fm_lock_try_create`&#39;s `[ -e &#34;$lockdir&#34; ]` precheck and writes its owner record; B publishes the symlink; A&#39;s `ln -s` fails EEXIST; A runs `fm_lock_remove_stray_owner_link` (a `basename` fork) and `fm_lock_discard_owner`; B releases (`rm -f &#34;$lockdir&#34;`); A reaches :925, finds no lockdir, and returns 1 with `FM_LOCK_HELD_PID` empty. Against the base library A instead fell through to the steal path, where `fm_lock_recheck_stale_owner` with an empty expected pid succeeds and `fm_lock_try_create` acquires. Wrong-result-without-erroring at the single-shot callers: `bin/fm-watch.sh:1606` sees rc=1 with an empty `FM_LOCK_HELD_PID`, prints `watcher: already running`, and `exit 0` while no process holds `.watch.lock` - the unsupervised window the intent is filed against; `bin/fm-supervise-daemon.sh:1515` exits 1 with `another fm-supervise-daemon is already running (lock ... held)` for a lock nobody holds. Wait-loop callers (`fm_lock_acquire_wait`, `fm_lock_acquire_wait_bounded`) only lose one poll. Remedy is a narrowing of this guard, not a revert - reverting restores the unbounded `$lockdir.steal.steal...` recursion the guard was added for. Either retry `fm_lock_try_create` once before bailing (still bounded: `tests/fm-watcher-lock.test.sh` `test_lock_create_failure_does_not_recurse_through_steal` asserts at most 2 mktemp attempts, which a single retry satisfies), or have `fm_lock_try_create` signal &#34;the owner record could not be established&#34; (mktemp or `fm_lock_prepare_owner` failed, as opposed to the lock already existing) and gate :925 on that signal alone.

🔧 Fix: Retry lock creation races and remove duplicate scratch test

- ⚠️ `bin/fm-wake-lib.sh:929` - The narrowed recursion guard bounds the recursion correctly, but its exhausted branch still reports a lock nobody holds as held. When both `fm_lock_try_create` attempts fail and `$lockdir` is still absent, line 929 sets `FM_LOCK_HELD_PID` from a nonexistent `$lockdir/pid`, so the caller sees `rc=1` with an empty held pid — indistinguishable from real contention. Verified on this machine by stubbing `mktemp() { return 1; }` and sourcing the HEAD library: `rc=1 held=[]`. Under the same stub the base library instead recursed unbounded (`.watch.lock.steal.steal…: File name too long`), so this is an improvement, not a regression — but the residual is a wrong label emitted without an error. Concrete effect: on a full or read-only state filesystem, `bin/fm-watch.sh:1605-1622` takes the empty-`FM_LOCK_HELD_PID` branch, prints `watcher: already running` and `exit 0`, so supervision believes a watcher exists when none does — the same unsupervised window the intent is filed against, now silent instead of noisy. `bin/fm-supervise-daemon.sh:1515-1520` similarly prints `another fm-supervise-daemon is already running (lock ... held)` and exits 1. The lost-double-race variant (two consecutive lost `ln -s` races, then a release) reaches the same branch but is far narrower than the single lost race the retry already covers. Note the remedy, not the defect, is what needs authorization: correcting this means giving `fm_lock_try_create` a distinguishable &#34;owner record could not be established&#34; signal (a new variable or return code) that line 929 and the single-shot callers can render as an error, which extends the change beyond the retry the last round was asked for.
- ℹ️ `bin/fm-wake-lib.sh:1051` - `_fm_lock_acquire_wait_handoff` open-codes the exact write-then-read-back sequence that `fm_lock_write_owner_record` (:442) already implements — two `printf` redirects into `pid` and `pid-identity` followed by two `cat` read-backs — differing only in that it writes the caller&#39;s pid/identity instead of the current process&#39;s. The four clauses at :1053-1056 collapse to `|| ! fm_lock_write_owner_record &#34;$ownerdir&#34; &#34;$caller_pid&#34; &#34;$caller_identity&#34;`, keeping the preceding `[ &#34;$pid_back&#34; != &#34;$current&#34; ]` ownership check and identical file ordering and semantics. That also removes `identity_back`, declared at :1036 and never read. Non-functional dedup; no behavior change.

🔧 Fix: Propagate distinct lock creation failures through shared wait helpers

- 🚨 `bin/fm-wake-lib.sh:1062` - `_fm_lock_acquire_wait_handoff` overwrites `$ownerdir/pid` with the caller&#39;s pid (:1062) BEFORE writing `$ownerdir/pid-identity` (:1063). If anything interrupts between those two writes, the lock is left with a live pid and the dead helper&#39;s identity, and `fm_lock_owner_is_abandoned` (:585-595) reads exactly that pair as ABANDONED: pid is alive, recorded identity != current identity -&gt; return 0. Concrete sequence: caller C runs `fm_lock_acquire_wait_bounded lock N`; helper H acquires and writes pid=C; the deadline fires and `fm_run_timed` sends TERM (bin/fm-timeout-lib.sh:135 sends TERM then KILL 0.2s later); H&#39;s trap runs `fm_lock_release`, which reads `$ownerdir/pid`=C, compares it to H&#39;s own pid at :1150 (`[ &#34;$pid&#34; = &#34;$current&#34; ] || return 0`) and REFUSES to clean up; the parent then reads `owner_pid`=caller_pid at :1108 and returns 0, so C enters the critical section believing it holds the lock. Any other waiter arriving next reclaims that lock out from under live C via the steal path - `fm_lock_recheck_stale_owner` re-runs the same `fm_lock_owner_is_abandoned` and agrees - producing two writers in a section meant for one, the outcome the intent calls &#39;worse than the bug&#39;. The repo&#39;s own new test `tests/fm-watcher-lock.test.sh:346` (`test_lock_reclaims_reused_pid_and_respects_slow_owner`) proves that a live pid with a mismatched identity IS reclaimed (rc=0), so the consequence is already demonstrated; only the reachability of the mismatched state is new. The same window is reachable without a signal: if the `pid` write lands and the `pid-identity` write fails (ENOSPC, remount read-only), the `||` chain falls to `fm_lock_release` at :1066, which is a no-op for the identical reason, and the parent still returns 0. Remedy is a two-line reorder: write `pid-identity` first, then `pid`. Then an interruption leaves pid=helper&#39;s own pid, `fm_lock_release` succeeds, the parent&#39;s `owner_pid != caller_pid` branch falls through to its final `fm_lock_try_acquire`, and ordinary dead-owner recovery handles it. No behavior change on the success path.
- ⚠️ `bin/fm-wake-lib.sh:1035` - This round turned `fm_lock_acquire_wait` from an infallible blocking call into one that can return 2 (:1031-1037 propagate any non-1 status out of the poll loop), and `fm_lock_try_acquire` from 0/1 into 0/1/2. Nothing in bin/ reads the new code, so one class of defect surfaces two ways. (a) Silent unlocked mutation: `bin/fm-wake-drain.sh` is `set -u` with no `-e`, and calls `fm_lock_acquire_wait &#34;$FM_WAKE_QUEUE_LOCK&#34;` unchecked at :628 and :684; on rc=2 the script continues, sets `DRAIN_LOCK_HELD=true`, and rewrites `$FM_WAKE_QUEUE` through its ack tempfile with no lock held. `bin/fm-startup-network.sh` (also `set -u` only) has the same unchecked shape at :225, :337, :372, :387, :429, :452, :480, :596, as do bin/fm-x-lib.sh:934,965,1023 and bin/fm-lease-lib.sh:182. (b) The original symptom is unchanged at the single-shot callers: `bin/fm-watch.sh:1606` takes `if ! fm_lock_try_acquire`, sees rc=2 with an empty `FM_LOCK_HELD_PID`, prints `watcher: already running` and `exit 0`, so supervision believes a watcher exists when none does - the unsupervised window the intent is filed against; `bin/fm-supervise-daemon.sh:1515` likewise reports &#39;already running (lock ... held)&#39; and exits 1 for a lock nobody holds. rc=2 is reachable two ways: an unwritable or full state filesystem (mktemp or the owner-record write fails, deterministic), and a lost creation race where the winner releases inside the retry window - `fm_lock_try_acquire` :922-929 checks absence, retries `fm_lock_try_create`, then returns 2 if the path is absent again, which the hot `$FM_WAKE_QUEUE_LOCK` (contended by fm-wake-drain, fm-watch, fm-wake-grant) can hit under churn without any filesystem fault. Note the remedy, not the defect, needs authorization: the previous round&#39;s instruction said to keep the distinction central to the helper and not to alter individual lock callers, so honestly closing this means either auditing and guarding the ~40 unchecked `fm_lock_acquire_wait` call sites or deciding which of rc=2&#39;s two triggers should keep blocking - both extend this change beyond the retry-and-propagate work that was asked for.

🔧 Fix: Guard failed lock acquisitions and preserve handoff ownership
1 error still open:

- 🚨 `bin/fm-wake-lib.sh:930` - rc=2 (&#34;cannot establish owner record&#34;) is decided by probing whether $lockdir exists after fm_lock_try_create fails, but an absent lock is the NORMAL end state of ordinary contention during a reclaim, so genuine contention is now reported as an unrecoverable creation failure. Mechanism: while a stealer holds $lockdir.steal and has removed the dead owner&#39;s lock (:1019 fm_lock_remove_path), any other acquirer&#39;s ln -s succeeds, then fm_lock_claim -&gt; fm_lock_claim_blocked_by_steal (:492-500) refuses and rm -f&#39;s its own link (:514). fm_lock_try_create therefore returns 1 with $lockdir absent, and both new probes (:930 and the post-reclaim one at :1027) read that as creation failure. The retry at :927 does not help: it is blocked by the same steal mutex and removes its link again. Reproduced deterministically on this machine, no race required - one process holding `&lt;lock&gt;.steal` with the primary lock absent, a second calling fm_lock_try_acquire: HEAD prints `lock: cannot establish owner record for .../state/.c.lock` and returns `rc=2 held=[]`; the identical scenario against base b84e0e3 returns `rc=1 held=[]`. Consequence, given commit 895308a wired ~40 callers to refuse on rc!=1 and fm_lock_acquire_wait (:1040) no longer retries rc=2: during the very stale-lock reclaim window this change exists to create, a concurrent fm_wake_append returns 2 and bin/fm-watch.sh:1743 (`fm_wake_append check pr-poll-retirement ... || exit 1`) kills the watcher; bin/fm-wake-drain.sh:628 exits 2 leaving acknowledged rows in the queue; bin/fm-supervise-daemon.sh:1518 exits 2 at startup. Base blocked ~50ms and succeeded. The same misread also reaches the steal recursion at :969-975, which returns 2 with FM_LOCK_HELD_PID still set, contradicting the rc=2 contract documented in docs/verification/supervision.md. Earliest shared boundary for the fix: fm_lock_try_create already knows which branch it failed on - only fm_lock_owner_dir (:527) and fm_lock_prepare_owner (:534) are genuine owner-record failures, while lock-present, ln-lost, and claim-refused are contention. Have it publish that distinction (e.g. a FM_LOCK_CREATE_* status the caller reads) and key rc=2 off it instead of the filesystem probe at :930, :954 and :1027. Note the existing coverage cannot catch this: tests/fm-lock-failure.test.sh and test_lock_create_failure_does_not_recurse_through_steal force failure only by faulting mktemp, so no case exercises a concurrent steal window.

🔧 Fix: Distinguish reclaim contention from owner-record creation failures
2 infos still open:

- ℹ️ `bin/fm-task-inbox-lib.sh:138` - `fm_lock_acquire_wait` gained an `[ &#34;$rc&#34; -eq 1 ] || return &#34;$rc&#34;` escape so a permanent owner-record failure returns immediately, but the five hand-rolled equivalents were not given it: `bin/fm-task-inbox-lib.sh:138` (`while ! fm_lock_try_acquire &#34;$lock&#34;; do ... sleep 0.1` against a 5s default deadline), `bin/fm-watch-arm.sh:149,198,283` and `bin/fm-push-transition-lib.sh:48` (20 attempts each). With the state filesystem unwritable, `fm_lock_try_acquire` returns 2 on every attempt and now prints `lock: cannot establish owner record for &lt;path&gt;` each time, so one inbox write emits ~50 identical stderr lines and each watcher-arm cycle emits 20. No wrong state results: every loop is bounded and refuses the protected work on exhaustion, which satisfies the round-8 instruction; this is retry waste and log noise on a failure that cannot clear. Remedy is the same two-line escape already in `fm_lock_acquire_wait`.
- ℹ️ `bin/fm-control.sh:272` - `fm-watch.sh`, `fm-supervise-daemon.sh` and `fm-lock.sh` were taught to separate rc=2 from contention, but four sibling single-shot callers still print a holder-exists message for a lock nobody holds: `bin/fm-control.sh:272` and `bin/fm-promote.sh:119` / `bin/fm-teardown.sh:388` (&#34;another lifecycle action is already running for task $ID&#34;) and `bin/fm-spawn.sh:1227` (&#34;another spawn is already creating task $ID&#34;). On an unwritable state directory the operator is told to hunt for a process that does not exist, which is the same manual-recovery cost the intent describes. Materially mitigated: `fm_lock_try_acquire` prints its own accurate `lock: cannot establish owner record for &lt;path&gt;` line immediately before, and all four refuse the protected work rather than proceeding, so no state is mutated and no held flag is set.
</details>

<details>

- 🚨 `bin/fm-wake-lib.sh:585` - A LIVE lock holder that exec()s loses its lock to the next acquirer. fm_pid_identity mixes process start time with the full argv, but exec preserves pid and start time while replacing argv, so fm_lock_owner_is_abandoned (bin/fm-wake-lib.sh:585-595) reads a live owner as dead and the steal path hands its lock away. Reproduced deterministically: a process acquires .probe.lock, execs `sleep 30`, and a second fm_lock_try_acquire reports &#34;RECLAIMED the live holder&#39;s lock&#34; while the original holder is still running inside its critical section - two owners in a section meant for one, the outcome the intent calls worse than the bug. This also breaks a pre-existing regression test that was not touched by this change: tests/fm-wake-queue.test.sh::test_live_presentation_holder_is_deadlined_without_weakening_ack, whose fixture holds the queue lock and then `exec sleep 30` (tests/fm-wake-queue.test.sh:1721) to model a live-but-stuck holder. It asserts one WAKE DRAIN SKIPPED advisory at line 1742; at HEAD the drain reclaims the lock instead and the assertion fails on 2 of 2 runs. The whole file passes at base b84e0e3 (109519 ms, exit 0). The remedy is a product decision, not a test fix, so I did not change code: either narrow the identity to what exec preserves (pid start time, which already survives exec and pid reuse) and accept the tick-collision weakening the cmdline component was added for, or keep the cmdline component and decide that an exec&#39;ing holder is out of contract - in which case the pre-existing test&#39;s fixture needs the author&#39;s ruling. Evidence: ~/.no-mistakes/evidence/01M20Z1MYB2PMK55PN53DN8XGH/live-holder-exec-steal.txt
- <code>`bin/fm-test-run.sh tests/fm-watcher-lock.test.sh tests/fm-lock-failure.test.sh` — 2 scripts, 0 failed (covers reused-pid reclaim, slow-owner respect, concurrent single-winner, creation-failure propagation, caller refusal)</code>
- <code>`bin/fm-test-run.sh tests/fm-wake-queue.test.sh` — FAILS at HEAD: `not ok - queue deadline did not emit exactly one holder advisory` (tests/fm-wake-queue.test.sh:1742), reproduced on 2 of 2 runs</code>
- <code>Same file at base: `git archive b84e0e3 | tar -x -C &lt;scratch&gt; &amp;&amp; bin/fm-test-run.sh tests/fm-wake-queue.test.sh` — exit 0, whole file passes</code>
- <code>End-to-end base-vs-HEAD demo at the operator surface: `FM_DEMO_WORKTREE=&lt;worktree&gt; bash stale-lock-reclaim-demo.sh` running the real `bin/fm-wake-drain.sh` against identical planted state (dead-owner-with-reused-pid lock; live slow holder; 8 concurrent `fm_wake_append` writers)</code>
- <code>Minimal mechanism probe for the regression: `bash live-holder-exec-steal.sh` — live holder execs, second `fm_lock_try_acquire` takes its lock</code>

🔧 Fix: Identify lock owners by start time, not argv
1 info still open:

- ℹ️ `tests/fm-wake-queue.test.sh:1922` - tests/fm-wake-queue.test.sh::test_secondmate_reprovisioned_queue_starts_a_fresh_interval is flaky and is not caused by this change. It failed once at HEAD (`not ok - a frozen reprovisioned queue generation was hidden: checkpoint: no actionable wake within 1s`) and reproduces at base b84e0e3 on 1 of 2 runs with the identical message. Already tracked as open issue 4012, &#34;test(wake): stabilize foreign-stall checkpoint fixture&#34;. A clean re-run of the same file at HEAD is green (41 ok, exit 0), including the presentation-lock case that regressed in round 1.
- <code>`bash tests/fm-watcher-lock.test.sh` — 40 cases green, incl. `test_live_holder_that_execs_keeps_its_lock`, `test_lock_reclaims_reused_pid_and_respects_slow_owner`, `test_lock_steals_dead_pid_lock`, `test_lock_does_not_steal_live_lock`, `test_lock_single_winner_under_concurrency`, `test_lock_stale_steal_single_winner_under_concurrency`</code>
- <code>`bash tests/fm-lock-failure.test.sh` — 16 cases green, incl. the forced-acquisition-failure regressions through fm-wake-drain, fm-wake-grant, fm-watch.sh, fm-supervise-daemon.sh, the lease guard and the startup harvest</code>
- <code>`bash tests/fm-wake-queue.test.sh` — whole file green on a clean run (41 ok, exit 0); confirms the round-1 regression `test_live_presentation_holder_is_deadlined_without_weakening_ack` now passes</code>
- <code>Manual end-to-end before/after: planted a real acquired lock, SIGKILLed its owner, then gave the dead owner&#39;s pid number to an unrelated live process, and ran the identical five scenarios against a base b84e0e3 copy and against HEAD — `bin/fm-watch.sh` startup, `fm_lock_acquire_wait_bounded &lt;lock&gt; 5`, a busy live owner under `FM_LOCK_STALE_AFTER=0`, a live owner that `exec`s, and 8 concurrent reclaimers of one stale lock, plus a no-reuse control</code>
- <code>Negative control: ran HEAD&#39;s `test_lock_reclaims_reused_pid_and_respects_slow_owner` against the base b84e0e3 library — red with `rc=124 held=8555` (waited out the window instead of reclaiming); green against HEAD</code>
- <code>Negative control: ran HEAD&#39;s `test_live_holder_that_execs_keeps_its_lock` against the 4a32ff1 library — red with `rc=0 held=` (a live holder&#39;s lock taken after exec); green against HEAD</code>
- <code>Flake attribution: ran `bash tests/fm-wake-queue.test.sh` twice against the base b84e0e3 copy — `test_secondmate_reprovisioned_queue_starts_a_fresh_interval` failed 1 of 2 runs at base with the same message seen at HEAD</code>
</details>

<details>

✅ No issues found.
</details>

<details>

🔧 Fix: silence intentional SC2016 in lock test snippets
</details>

<details>

✅ No issues found.
</details>


dbeihl-utilicast:
Closing: this pull request targets the upstream repository. Our standing rule is that we never publish changes there; the work belongs on our own fork, dbeihl-utilicast/firstmate. The cause was a missing default repository on our machines, now fixed, and a guard so a wrong target refuses is filed. Closed on the captain instruction, 2026-09-15. Nothing was merged upstream.

PR #4067, opened 2026-09-09
fix(spawn): keep a claude worker's account across relaunch

strandborg:
## Why

A claude worker's account was silently lost on relaunch.

`bin/fm-spawn.sh` already forwarded a set `CLAUDE_CONFIG_DIR` onto a claude worker's launch, and `bin/fm-claude-trust.sh` already wrote that worker's worktree trust entry into the same account's store. But nothing recorded *which* account a task was launched against, so `bin/fm-control.sh relaunch` — which rebuilds every identity axis from `state/<id>.meta` — had nothing to rebuild it from. A relaunched worker therefore inherited firstmate's own ambient account instead, where its worktree trust record may not exist, leaving it facing the trust dialog firstmate cannot answer.

The failure is silent, which is why it needed fixing rather than remembering.

## What changed

- The account is resolved **once**, at the first point the harness is final and before either consumer runs, so the trust pre-registration and the launch prefix can never name different stores.
- A fresh spawn takes it from the environment; a `--relaunch` takes it from the task's own record, exactly as it already takes backend, kind, mode, yolo and worktree.
- `claude_config_dir=` is recorded for a caller-chosen account. An absent field means the default store on **both** sides: the trust helper falls through to its own `HOME` resolution and the launch carries no prefix, so the two always agree.
- A relative value is refused rather than resolved, on both the environment and the record paths, because the two sides would otherwise name different stores.
- `bin/fm-bootstrap.sh`'s secondmate liveness sweep respawns an existing task as a fresh spawn with no `--relaunch`, so it reached the same silent loss. Those call sites now read the recorded account and pass it, keeping account selection with the caller.
- `.gitignore` gains a blanket `.omc/` rule. Agent runtime state was being swept into commits — it happened during this change's own review rounds.

Firstmate deliberately gains **no** account picker and no dependency on any external routing tool. Selection stays with the caller; these changes only stop the choice being lost.

## Evidence

Driven end to end on a private tmux server against the real product: real `bin/fm-spawn.sh`, real `bin/fm-claude-trust.sh`, the real `bin/fm-bootstrap.sh` sweep, real git worktrees and real per-account `.claude.json` stores. The only stand-ins are the two programs a worker pane runs, `treehouse` and `claude`, with the `claude` stand-in serving as the observation point reporting which account the worker process actually received.

Seven live scenarios pass. The headline:

> On the fixed build, a task dispatched to account 2 stays on account 2 through a relaunch, with the trust entry only ever in account 2. On the pre-fix baseline the same drive moves the replacement worker to account 1 and writes the trust record there.

Portable regressions extend `tests/fm-claude-trust.test.sh`, `tests/fm-control-relaunch.test.sh` and the secondmate liveness suite. The relaunch tests deliberately set the ambient account to a *different* one, so they cannot pass vacuously by both paths agreeing.


strandborg:
Closing: opened against the wrong repository by mistake. This belongs on my own fork, not upstream.

PR #4078, opened 2026-09-09
fix(backend): validate Orca's composite worktree id structurally

lelouch:
## Problem

`bin/fm-teardown.sh` refuses **every** Orca-backed task:

```
REFUSED: Orca endpoint metadata for task <id> is malformed or inconsistent; preserving task state.
```

Orca's worktree ids are natively composite, `<repo id>::<absolute worktree path>`, but `fm_backend_endpoint_atom_valid` accepts only `[A-Za-z0-9._@%+-]`. The `::` and `/` present in every real Orca id fail that check, so no Orca task can ever be torn down. Completed workers strand, holding a terminal and a worktree, and an idle finished worker keeps raising stale wakes.

The recorded metadata was always correct: `bin/fm-spawn.sh`'s `parse_orca_worktree_result` faithfully stores what `orca worktree create --json` returns. The validator was the wrong half.

This shipped because every Orca fixture used a fabricated simple id (`wt-teardown`, `wt-spawn`, `worktree-9`). None exercised the shape Orca actually returns, so the suite passed against a fiction.

## Fix

The shared atom class is **not** widened. It guards the tmux, Herdr, Zellij, and cmux endpoint atoms, and loosening it to admit `:` and `/` would weaken every other backend to fix one.

Instead the composite id gets its own structural rule, `fm_backend_orca_worktree_id_valid`:

1. splits on the first `::` into two non-empty halves;
2. the repo half must still pass `fm_backend_endpoint_atom_valid` (reused, not re-implemented);
3. the path half must be absolute and free of control characters;
4. the path half must equal the task's recorded `worktree=`.

Point 4 is a cross-record consistency check the character class could never express: a record whose id names a different directory than the task's own worktree now refuses, where previously the check was simply unreachable. The result is strictly stronger than what it replaces. Every failure mode remains a refusal that preserves task state, and each names its concrete reason and the offending value.

## Tests

Fixtures across `tests/fm-backend-orca.test.sh`, `tests/fm-teardown-endpoint-safety.test.sh`, and `tests/fm-control.test.sh` are converted from fabricated atoms to the real composite shape. New cases cover a valid id accepted, a mismatched path half, a missing `::`, a relative path half, empty halves, control characters, and the repo half still bound by the atom rule, plus a companion case pinning the shared atom class as unchanged.

Verified: the new tests fail against the old validator and pass against the new one. Full regression sweep green across all five backends. `bin/fm-lint.sh` exits 0 (ShellCheck 0.11.0, actionlint 1.7.12). The pipeline's test step independently reproduced the defect on the base commit, confirmed the fix, and ran a hostile-id matrix with zero destructive Orca calls.

No same-class defect exists in the other backends: Herdr pre-strips colons, and cmux and Zellij ids are uuid/numeric and pass the atom class unchanged.

https://claude.ai/code/session_017f1DofkK6VLYWvtXwyk53a

### DIFF
--- a/bin/fm-backend.sh
+++ b/bin/fm-backend.sh
@@ -382,6 +382,13 @@ fm_backend_meta_exact_value() {  # <meta-file> <key>
   printf '%s' "$value"
 }
 
+# Refusals name the value they rejected, but an endpoint value only ever reaches
+# a refusal because it is malformed, so it can carry control bytes an operator's
+# terminal would interpret. Render it printable before quoting it into stderr.
+fm_backend_printable() {  # <value>
+  printf '%s' "$1" | tr '[:cntrl:]' '?'
+}
+
 fm_backend_endpoint_atom_valid() {  # <value>
   case "$1" in
     ''|*[!A-Za-z0-9._@%+-]*) return 1 ;;
@@ -560,7 +567,7 @@ fm_backend_validate_task_endpoint() {  # <meta-file> <task-id>
         return 1
       fi
       if ! fm_backend_orca_worktree_id_valid "$worktree_id" "$worktree"; then
-        echo "REFUSED: Orca worktree id for task $id $FM_BACKEND_ORCA_ID_REASON; preserving task state." >&2
+        echo "REFUSED: Orca worktree id for task $id $FM_BACKEND_ORCA_ID_REASON; id '$(fm_backend_printable "$worktree_id")', recorded worktree '$(fm_backend_printable "$worktree")'; preserving task state." >&2
         return 1
       fi
       window=$terminal
--- a/bin/fm-backend.sh
+++ b/bin/fm-backend.sh
@@ -388,6 +388,54 @@ fm_backend_endpoint_atom_valid() {  # <value>
   esac
 }
 
+# Orca worktree ids are natively composite, unlike every other backend's opaque
+# endpoint atoms: docs/orca-backend.md "Task shape and metadata" owns that format.
+# Validating its structure here keeps fm_backend_endpoint_atom_valid's character
+# class intact for the tmux, Herdr, Zellij, and cmux atoms that share it, and adds
+# a cross-record check the character class could never express: the id's own path
+# half must be the task's recorded worktree, so a record whose id names another
+# directory refuses instead of releasing the wrong worktree. Every failure mode is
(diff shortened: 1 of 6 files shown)


kunchenguid:
Speaking as Kun's firstmate:

First look on HEAD `32715fdd58df88f2c5556e609bb75485c23d7431` vs main `40c50ea8843c5b6a5351db8352675537252b653e`.

**Contract-class: restore.** Unconfigured Orca teardown already promised path-matched release of the recorded worktree; the shared endpoint-atom character class wrongly refused every real Orca composite id (`<repo>::<abs path>`). Tip adds `fm_backend_orca_worktree_id_valid` (split on first `::`, atom repo half, absolute control-free path half must equal recorded `worktree=`) and keeps other backends on the unwidened atom rule. Failures still refuse and preserve task state. Docs + `tests/fm-teardown-endpoint-safety.test.sh` pin it.

**Attestation: MISSING** — PR body has no `no-mistakes-pipeline-attestation:v1`. Raise/refresh via `git push no-mistakes` so the marker MATCHES tip before merge can be considered.

**Workflow-zero: yes** (no `.github/workflows/*`). File list: `bin/fm-backend.sh`, Orca docs, backend/teardown tests. Diff security skim: structural validation only; printable refusal rendering; no credential/workflow payload. **Approved** first-time fork CI `34359573769` and Require no-mistakes `34359574000` this pass.

**VISION (per rule):**
- One captain, one interface: **aligns** — teardown honesty under Orca; no new captain surface.
- Authority is explicit and never inferred: **aligns** — still refuse-preserve on bad metadata; no new autonomy.
- Scripts own the mechanics, agents own the judgment: **aligns** — deterministic id validation in `fm-backend.sh`.
- A restart is a non-event: **aligns** — cleanup can complete for real Orca records instead of false-malformed refuse.
- Delegation with a spine: **aligns** — ship teardown stays refuse-closed on mismatch.
- The fleet outlives any vendor: **aligns** — binds to Orca's documented composite id semantics without widening other backends.
- Scope: **aligns** — backend/command-layer validation + regression; not workshop.

Waiting on hosted CI (just approved), not on an author code fix. Attestation remains an author-side merge gate after green checks. Merge-eligible **N**. No Firstmate captain card.

PR #4107, opened 2026-09-10
fix(spawn): let a home scope the treehouse pool its workers take worktrees from

synchronizeaicontact7919:
## The defect

Treehouse names a pool `<repo>-<sha256(remote-url)[:6]>` beneath one root, so a pool is keyed on the **remote** and not on the clone. Every clone of one project on a machine therefore resolves to the same pool, while that pool's worktrees stay linked to whichever clone created them.

Two firstmate homes that each hold a clone of one project collide there. Reproduced on a machine with two homes holding clones of one project:

```
$ bin/fm-spawn.sh <task> projects/<proj> --mode no-mistakes --yolo on
error: refusing to pre-register Claude trust:
  '/Users/…/.treehouse/<proj>-11c658/1/<proj>' is not a worktree of project
  '/Users/…/<second-home>/projects/<proj>'
```

The pool key is the remote URL, confirmed directly:

```
sha256("https://github.com/<owner>/<proj>.git")[:6] == 11c658
```

`bin/fm-claude-trust.sh` is right to refuse: the acquired worktree does not share the spawning project's git common dir, and its structural test is the fence that keeps a worker out of another home's checkout. The second home simply cannot dispatch at all — every ship and scout spawn fails here.

## The change

The fence stays exactly as it is; the **pool** moves. `FM_TREEHOUSE_ROOT`, else `config/treehouse-root`'s first non-blank, non-`#` line, is carried onto `treehouse get --root`, so the acquired worktree is linked to this home's own clone and the structural test then admits it.

Verified after the change on the same machine:

```
worktree common dir: /Users/…/<second-home>/projects/<proj>/.git   # this home's clone
other home's pool:   unchanged, 11 worktrees, untouched
```

## What does not change

Absent, a bare `treehouse get` is sent exactly as before. Changing an existing home's root would orphan the pool it already has, so a home opts in and every existing home sees no difference.

Two refusals rather than silent fallbacks:

- A **relative** value is refused. Treehouse resolves a relative `--root` from the *repository* root, so a relative value would put the pool inside the project clone — a write into a project, and a pool shared by every home that clones it, which is the collision this setting exists to end.
- A **present file naming no root** is refused rather than treated as absent, so an emptied file cannot silently return the home to the shared pool it was configured to leave.

## Tests

`tests/fm-spawn-treehouse-root.test.sh`, five cases, asserting the command the spawn actually sends to the pane: bare `treehouse get` when unconfigured, `--root` carried when configured, `FM_TREEHOUSE_ROOT` winning over the file, and both refusals. Confirmed to be a real oracle — the configured-root case fails against the pre-change script and passes after.

The spawn tmux fixture gains `FM_FAKE_CMD_LOG` for the submitted text lines a spawn types into the pane; the existing `FM_FAKE_LAUNCH_LOG` only ever saw `send-keys -l` payloads, which a typed shell command is not.

Run on this branch, rebased on current `main`:

- `bin/fm-lint.sh` — ShellCheck 0.11.0, actionlint 1.7.12, clean
- `bin/fm-doc-audience-check.sh` — ok, surfaces=97 local_links=365
- `bin/fm-test-run.sh` over `fm-spawn-treehouse-root`, `fm-claude-trust`, `fm-spawn-worktree-settle`, `fm-test-fixtures`, `fm-spawn-dispatch-profile` — 5 suites, 0 failed

One pre-existing unrelated failure on this machine, present on a clean tree before this change and untouched by it: `fm-muse-harness.test.sh` — `fm-harness.sh under process 'muse-bin-0.1.0-R708.1' reported ''`.

## Documentation

`docs/configuration.md` gains "Treehouse pool root" as the owner; `AGENTS.md`'s config inventory gains the one-line pointer; `bin/fm-spawn.sh`'s header carries the mechanics.


kunchenguid:
Speaking as Kun's firstmate:

Diff review: `FM_TREEHOUSE_ROOT` / `config/treehouse-root` scopes `treehouse get --root` per home; absent = today's default root. Absolute-path only. Security clean (no credential paths; prevents cross-home pool collision without relaxing trust).

**VISION per-rule:** One captain/interface — aligns. Authority — aligns (explicit config). Scripts vs agents — aligns. Restart — aligns. Delegation — aligns (safer multi-home). Fleet/vendor — aligns. Scope — aligns.

**contract-class:** opt-in — unconfigured spawn unchanged; home opts into its own pool root.

**attestation:** MISSING — please raise through no-mistakes so the tip `09ceca3445278583f9b0c747f746e4b48d0a6a50` carries a matching pipeline attestation.
**workflow-approval:** fork CI + Require no-mistakes approved this tip (`34422270218`, `34422270326`).
**next:** waiting on the author (attestation), not a captain hold.

PR #4127, opened 2026-09-10
feat(skills): add RevoCall-family local e2e provision/cleanup pair

Hao-Siong:
## What this adds

A shared skill pair for standing up and tearing down local Docker stacks used for
end-to-end testing across the RevoCall family (RevoCall, RevoCall-AI-Handler,
Conversational-AI-Framework), plus the deterministic core the pair drives.

- `skills/revocall-e2e-provision` — provisions a task-scoped stack, seeds the
  minimum data for a named scenario, and records what it created.
- `skills/revocall-e2e-teardown` — releases exactly what was recorded.
- `bin/fm-e2e-stack.sh` — the record/gate/down core, with four ownership gates.
- `bin/fm-teardown.sh` — Fix 4 releases any recorded stack, so a worker that dies
  mid-test still gets cleaned up.

## Why

Container leaks from local e2e runs were taking a forensic session to attribute:
most stray containers carried no Compose project label, so recovering ownership
meant reading Postgres logs, checking whether worktrees still existed, and
grepping scout reports. Fifteen such containers were removed by hand before this
work started, some more than five days old.

Root cause was three concrete Compose defects rather than carelessness:
`container_name:` surviving `-p` and hiding project identity; list-valued keys
appending instead of overriding without `!override`; and `external: true` volumes
that Compose neither creates on `up` nor removes on `down -v`.

## Design

Identity comes from the Compose project name (`fm-<task>-<stack>`), so ownership
is readable from a bare `docker ps` with no forensic work; labels serve as
cleanup filters rather than as the identity itself. Isolation uses a generated
override Compose file written into the per-task tmp directory, reusing RevoCall's
existing `gen-slot-env.sh` port-offset scheme — no project repo is modified.

Seeding routes to each repo's own tooling (golang-migrate plus the Go seeder for
RevoCall, Alembic plus pytest fixtures for CAF, in-memory SQLite for AI-Handler)
with a thin per-scenario SQL delta layer on top.

The core safety rule is that **removal always needs a record, never an
inference**. An anonymous container is not evidence of abandonment — during this
work five containers that looked like strays turned out to be live pipeline work
and self-cleaned.

## Verification

The full provision → seed → verify → cleanup cycle was run live against CAF's
real Compose file and RevoCall's real 101-migration schema, with zero residue and
zero collateral damage. `docs/verification/local-e2e-infra.md` records the
procedure.

Tests: 13 new unit cases in `tests/fm-e2e-stack.test.sh` (driven by a fake
`docker` fed from a TSV inventory) and 3 integration cases in
`tests/fm-teardown.test.sh` covering release-before-worktree-return ordering,
refusal never blocking teardown, and no-record never invoking Docker.

## Notes for the reviewer

- Stacks are isolated per task, but LiveKit is not — that caveat is documented in
  the provision skill rather than papered over.
- The design deliberately ships one shared skill pair rather than per-repo
  variants; the reasoning and the 15 real per-repo differences it has to
  accommodate are in `skills/revocall-e2e-provision/references/profiles.md`.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -105,6 +105,7 @@ state/               runtime records and signals; gitignored
   <id>.cursor-session  cursor busy-source binding (projects root, task worktree, prior conversations) written by fm-spawn; removed by teardown
   <id>.reconcile-nudged  epoch second of the last inventory-reconcile nudge sent to this secondmate; bin/fm-secondmate-reconcile.sh owns its per-home cooldown window
   <id>.backlog-close  the exact backlog transition a teardown recorded before removing the task's record, so an interrupted cleanup can still be finished at the next session start; bin/fm-backlog-transition-lib.sh owns its format and replay, and a landed transition removes it
+  <id>.e2e-stack     durable record of the local docker stacks a task provisioned for end-to-end testing, so cleanup never has to rediscover them; written only by bin/fm-e2e-stack.sh, which owns the block format and the four ownership gates, and removed by that script's verified `down`
   <id>.inbox/          durable steering inbox: sequenced firstmate instruction records the worker acknowledges by moving them into its handled/ subdirectory; written by fm-send, with ordinary records re-rung and escalated by the watcher while explicit fire-and-forget records are excluded from that ladder, and removed by teardown (bin/fm-task-inbox-lib.sh)
   <id>.meta          task metadata; each producer script's header owns its exact fields and mutation contract, with docs/configuration.md routing operator-facing backend and trace-context details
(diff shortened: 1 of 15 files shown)


Hao-Siong:
Closing unmerged: this skill pair is being kept local to the originating firstmate home rather than published to the shared repo, so it is no longer proposed for this repository. Nothing here needs review or action. Note for anyone who finds this later: a fork PR cannot satisfy this repo's no-mistakes provenance gate by construction, since that gate requires an attestation the pipeline can only produce when it is able to push to the base repository.

PR #4133, opened 2026-09-10
fix(herdr): reconcile stale Pi registrations on recovery

### LINKED ISSUE #4115: Recovery classifier trusts a stale herdr agent registration as 'live', blocking relaunch/spawn after a Pi crew exits to a shell
## What happened

When a Pi crew exits to a plain shell but its herdr pane remains, herdr keeps the last Pi hook registration as `agent=pi`, `agent_status=idle` (it does not release the agent on TUI shutdown). The recovery classifier then treats that stale registration as a live agent and blocks all in-policy recovery:

- `fm-control.sh <id> relaunch` sends `/quit` into the dead shell, waits `FM_CONTROL_EXIT_WAIT` for a dead transition that never comes, and refuses.
- `fm-spawn.sh`'s positively-dead precondition then also refuses (the pane is not positively dead).
- Raising `FM_CONTROL_EXIT_WAIT` cannot help — the stale registration persists for hours.
- The raw `herdr pane release-agent` path returns success but is a **no-op** on the affected herdr version, so there is no in-policy self-recovery.

Impact: crew recovery is wedged. Multiple validations sit parked with preserved (uncommitted) work that cannot resume.

## Root cause (confirmed present on current `main`)

`fm_backend_herdr_pane_agent_state` (bin/backends/herdr.sh) classifies a pane from the JSON body of `pane get` / `agent get`:

- `pane get` → `pane_not_found` ⇒ `dead`
- `agent get` → `agent_not_found` ⇒ `no-agent` (⇒ dead)
- `agent get` succeeds with **any** registered `agent_status` (`working|idle|done|blocked`) ⇒ `live`

The `live` branch never verifies that the OS process still exists. The code's stated assumption is that herdr reaps the agent when its process dies ("killing a pane's shell pid on a live server makes herdr immediately drop both the pane and agent"). That assumption does **not** hold when a Pi agent exits to a shell and the pane persists: the registration survives, so the classifier reports `live` for a dead agent, and `fm_backend_herdr_agent_state` maps that to `alive`, blocking relaunch and the spawn dead-precondition.

Note the existing pane-liveness helpers (`fm_backend_herdr_pane_idle_shell_pid` and the death-close path) are **top-shell-only**; a crew pane contains a `treehouse get` subshell, so a top-shell check is not sufficient here even if it were wired into the recovery classifier (it is not).

Related: #3954 (herdr keeps escalating "possible wedge" after an agent has cleanly exited — likely the same underlying "agent not released on clean exit" root, different downstream effect) and #4091 (recovery classified `unreadable` instead of `missing` when the pane is confirmed gone).

## Suggested fix — guarded stale-registration reconciliation in the control owner (NOT a general force flag)

When `agent get` reports a Pi agent `idle|done|blocked`, also read pane process-info and attribute processes to the exact pane shell / descendant tree (descendant-aware, since the crew pane holds a `treehouse get` subshell — the top-shell-only helper is insufficient):

- Pane has a real Pi descendant ⇒ keep `alive`.
- Pane has only the preserved worktree shell chain ⇒ re-read identities, release only the matching `source=herdr:pi` / `agent=pi` authority, require `agent_not_found`, then return `dead`.
- Do all of this under the control owner's per-task lock so release and replacement cannot race.
- Any ambiguity, identity change between proof and release, an active foreground process, or a failed release ⇒ stay `unreadable` / refuse.

Separately, herdr's Pi integration should send `pane.release_agent` on TUI shutdown — there is currently no exit-release handler, which is what leaves the stale registration behind in the first place.

## Suggested regression tests

- Stale `idle` registration over a `treehouse` → `bash` descendant chain ⇒ reconciled to dead.
- A real idle Pi ⇒ stays alive (no false release).
- process-info read failure ⇒ stays unreadable / refuses (no release).
- Identity change between the liveness proof and the release ⇒ refuses (no release).
- Preserved uncommitted work survives across the relaunch that follows the reconciliation.

kunchenguid:
Fixes #4115.

When a Pi crew exits to a shell, Herdr can keep `agent=pi` / `agent_status=idle` registered on the still-present pane with no Pi OS process. The recovery-grade classifier treated any registered status as live, so `fm-control` relaunch sent `/quit` into a dead shell and `fm-spawn --relaunch` refused the positively-dead precondition.

This change keeps the husk classifier strict (closing a pane still refuses a registered idle Pi). Only the recovery-grade read proves descendant-aware process liveness:

- A real Pi descendant stays alive.
- A treehouse-to-shell worktree chain may release only matching `source=herdr:pi` / `agent=pi` authority, and only under the control plane's per-task lock, and only after `agent get` reports `agent_not_found`.
- Any ambiguity, identity change, process-info failure, missing lock, or failed release stays unreadable and does not release.

There is no general force-override. Sending `pane.release_agent` on Pi TUI shutdown remains an upstream Herdr concern.

## Test plan

- [x] Stale idle registration over a treehouse->bash descendant chain reconciles to dead
- [x] A real idle Pi stays alive (no false release)
- [x] Process-info read failure stays unreadable (no release)
- [x] Identity change between the liveness proof and the release refuses (no release)
- [x] Preserved uncommitted work survives the reconciliation, and a real Herdr relaunch of an agent-free pane keeps the worktree file
- [x] `bin/fm-test-run.sh tests/fm-backend-herdr.test.sh tests/fm-secondmate-liveness.test.sh tests/fm-control.test.sh tests/fm-control-relaunch.test.sh tests/fm-control-herdr-smoke.test.sh`

### DIFF
--- a/bin/backends/herdr.sh
+++ b/bin/backends/herdr.sh
@@ -142,6 +142,17 @@ FM_BACKEND_HERDR_SECONDMATE_MARKER=".fm-secondmate-home"
 # No send, capture, Treehouse, or general task-ownership path reads it.
 FM_BACKEND_HERDR_PRESENTATION_JOURNAL_SUFFIX=".herdr-presentation"
 
+# Recovery-grade stale-Pi reconciliation (issue #4115) is owned by
+# fm_backend_herdr_recovery_live_state. Callers that hold the per-task
+# control lock (fm-control exit, fm-spawn --relaunch) export
+# FM_BACKEND_HERDR_CONTROL_LOCK to that lock directory and
+# FM_BACKEND_HERDR_RECONCILE_STALE_PI=1 so a proven worktree shell chain
+# may release only source=herdr:pi / agent=pi. Interrupt, husk detection,
+# and every other reader leave those unset, so a stale registration stays
+# unreadable rather than released. FM_BACKEND_HERDR_STALE_PI_PROOF_POLLS
+# bounds the process-tree settle (default 10). FM_HERDR_PS_BIN overrides
+# the process-list binary in tests.
+
 # The config item a home writes to opt out of, or explicitly in to, the
 # projection.
 FM_BACKEND_HERDR_PRESENTATION_CONFIG="herdr-presentation-spaces"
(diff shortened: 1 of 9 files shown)


kunchenguid:
Superseded: this direct-PR was opened before the delivery path switched to no-mistakes. The pipeline will raise the attested PR for the same branch. Closing so the no-mistakes gate can own publication.

Issue #1796, opened 2026-08-06
fm-spawn.sh cannot detect treehouse worktree entry on Windows: nested cmd.exe hides the cwd move


jk1rby:
## Summary

On a Windows / Git Bash host, `fm-spawn.sh` always fails at the `treehouse get` step with:

```
error: treehouse get did not enter a worktree within 60s; inspect window default:wE:p4
```

`treehouse get` itself **succeeds**. The worktree is created and is a valid checkout. The spawn fails only because firstmate cannot *observe* that it happened.

## Root cause

`fm-spawn.sh` detects worktree entry by polling the pane's reported cwd and waiting for it to move away from the project directory (`bin/fm-spawn.sh:1296-1341`).

On Windows, `treehouse get` enters the worktree by launching a **nested `cmd.exe` subshell** inside the outer PowerShell pane. herdr tracks the cwd of the pane's top-level shell, so the reported cwd never changes.

Observed simultaneously:

```
herdr pane list  -> wE:p4 cwd = C:\Users\jk\Documents\programming\firstmate\projects\aim
herdr pane read  -> C:\Users\jk\.treehouse\aim-59788f\1\aim>
```

The terminal is visibly inside the worktree while the structured cwd still reports the project dir. The detection loop polls all 60 iterations and times out.

The worktree is real:

```
$ git -C /c/Users/jk/.treehouse/aim-59788f/1/aim log --oneline -1
6bb45826 Merge branch 'feature/07-mac-T1651-skills-audit-2026-08-04' into 'develop'
```

## Why the current approach cannot work here

The two-consecutive-reads confirmation in that loop guards against a *transient stale* path. This is not transient: the outer shell's cwd is correct and permanent. No amount of polling will see the move, because the move happened in a child process the pane does not report on.

## Suggested directions

Not prescribing a fix, but the options seem to be:

1. Ask the pane for the worktree path directly rather than inferring it from cwd (send a command that prints `git rev-parse --show-toplevel` and read it back).
2. Derive the expected treehouse path and verify it on disk, rather than watching the pane at all.
3. Have `treehouse get` report the path it entered in a machine-readable way.

Option 1 or 2 keeps the worktree validation that `validate_spawn_worktree` performs, which matters: the existing comment explains that accepting the wrong path silently records a bad `worktree=` in `state/<id>.meta`.

## Scope

[...]

PR #4170, opened 2026-09-10
fix(backends): launch headless Zellij sessions via PowerShell on native Windows

MLA82:
## Intent

`fm_backend_zellij_server_ensure` in `bin/backends/zellij.sh` launches a headless Zellij session with `( nohup zellij attach -b "$session" </dev/null >/dev/null 2>&1 & )`. On native Windows (Git Bash/MSYS, not Cygwin), this session never comes up, even though Zellij itself is fully functional there.

## Root cause

Git Bash's `&` backgrounding does not achieve real OS-level process detachment the way a Unix double-fork/`setsid` does. The backgrounded `zellij attach -b` process is tied to the invoking one-shot bash subshell's lifetime and dies the instant that subshell exits — before the poll loop that waits for the session to appear ever gets a chance to find it.

Verified live by elimination on Windows 11 (Git Bash, Zellij 0.45.1):
- The identical `zellij attach -b <name>` command, run interactively with an attached client, works fine (creates the session, renders correctly, exits cleanly).
- The same command launched instead as a genuinely separate Windows process via PowerShell's `Start-Process` persists independently, and `zellij list-sessions` finds it afterward.

This proves Zellij's own background-session support is fine on Windows — only this script's Unix-style backgrounding idiom fails to survive there.

## Fix

On `msys*`/`mingw*` only (Cygwin has a real fork/setsid and keeps the existing `nohup ... &` path, since it never had this bug to begin with), launch the headless session via `powershell.exe -NoProfile -NonInteractive -Command "Start-Process -FilePath 'zellij.exe' -ArgumentList 'attach','-b','$session' -WindowStyle Hidden"` instead. Since the session name is interpolated into a single-quoted PowerShell string, it is charset-guarded (`[A-Za-z0-9._-]+`) first — the session name is normally only the fixed default `"firstmate"` or an operator-supplied `FM_ZELLIJ_SESSION`, never arbitrary task-controlled content, but the guard is there regardless.

## Testing

Added to `tests/fm-backend-zellij.test.sh`:
- `fm_backend_zellij_server_ensure` launches via PowerShell `Start-Process` on a forced `OSTYPE=msys` fixture (fake `zellij`/`powershell.exe` binaries), and the session appearing in the simulated `list-sessions` output lets the existing poll loop return promptly.
- The same forced-Windows fixture refuses a session name containing a single quote before ever building the PowerShell command line, and never invokes `powershell.exe` in that case.
- A forced `OSTYPE=cygwin` fixture keeps the original `nohup`-based path, proving Cygwin is unaffected.
- A default (non-Windows) `OSTYPE` fixture still accepts a session name with a space, proving the charset guard is scoped to the Windows branch only.

Full suite: `tests/fm-backend-zellij.test.sh` — 68/68 passing.

Since this environment has no native Windows host, the Windows branch itself is exercised only through the forced-`OSTYPE` fixtures above; the actual PowerShell `Start-Process` behavior and the root-cause diagnosis were verified live on a real Windows 11 machine (Git Bash, Zellij 0.45.1) before writing this fix.

## Risk

Low: the change is fully scoped to native-Windows `fm_backend_zellij_server_ensure`, guarded by a narrow `OSTYPE` case that Cygwin and every non-Windows platform fall through unchanged, with an added defensive charset check that has no effect off Windows.

### DIFF
--- a/bin/backends/zellij.sh
+++ b/bin/backends/zellij.sh
@@ -227,16 +227,49 @@ fm_backend_zellij_session_exists() {  # <session>
 # fm_backend_zellij_server_ensure: create the named session in the background,
 # headless (no attached client), if it does not already exist - mirrors
 # tmux's `tmux has-session || tmux new-session -d` and herdr's server_ensure.
-# Verified: `zellij attach -b <name>` with stdin redirected from /dev/null and
-# no controlling TTY creates the session and returns promptly (it cannot
-# actually attach without a TTY, so it exits after creating); running it again
-# against an EXISTING session prints "Session already exists" and exits 1 -
-# harmless here because existence is checked first and the launch is
-# backgrounded, its exit status never inspected.
+# On the default branch - everything but Git Bash/MSYS, Cygwin included,
+# whose real fork/setsid makes `&` detach properly - verified: `zellij attach
+# -b <name>` with stdin redirected from /dev/null and no controlling TTY
+# creates the session and returns promptly (it cannot actually attach without
+# a TTY, so it exits after creating); running it again against an EXISTING
+# session prints "Session already exists" and exits 1 - harmless here because
+# existence is checked first and the launch is backgrounded, its exit status
+# never inspected.
+#
+# On native Windows (Git Bash/MSYS, not Cygwin), that same command never
+# comes up: Git Bash's `&` backgrounding does not achieve real OS-level
+# process detachment the way a Unix double-fork/setsid does, so the
+# backgrounded zellij server is tied to the invoking bash process's lifetime
+# and dies the instant this one-shot function's subshell exits, before the
+# poll loop below ever finds it. Verified live by elimination on Windows 11
+# (Developer Mode on, Zellij 0.45.1): the identical `zellij attach -b <name>`
+# launched instead as a genuinely separate Windows process via PowerShell's
+# Start-Process persists independently and is found by `zellij list-sessions`
+# afterward, proving Zellij's own background-session support works fine there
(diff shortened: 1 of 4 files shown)


MLA82:
Follow-up after a live retest on native Windows (Windows 11 build 26120, Git for Windows 2.55.0, Zellij 0.45.1).

**Detection fix, pushed as de327cb:** the `OSTYPE` match never fired on current Git for Windows. Its bash is now a Cygwin-runtime build (`bash --version`: `x86_64-pc-cygwin`), so Git Bash reports `OSTYPE=cygwin` and `server_ensure` silently took the `nohup ... &` path, whose session died immediately. The branch now keys on `uname -s` (`MINGW*`/`MSYS*` vs `CYGWIN_NT-*`), and the tests stub `uname` to pin that exact combination. `tests/fm-backend-zellij.test.sh` passes 68/68 on the Windows host.

**Separate finding, not addressed here:** with the PowerShell launch working, the headless session (`zellij attach -b`) still cannot host task tabs on Windows. Every `new-tab`, with any layout and with `-WindowStyle Hidden` or `Minimized`, logs `Failed to apply layout: Not enough room for panes` and gets no terminal pane, so `fm-spawn` stops at `could not find a terminal pane for zellij tab 1`. The same `new-tab` works immediately once a real client is attached (`zellij attach --create <name>` in a minimized console). The 2026-09-10 check only proved that the session persists, not that tabs can be created in it. Past that point, the spawn then hits #1796: panes start `cmd.exe`, so the `treehouse get` move is not observable. I'd keep this PR to the launch fix and follow up on those separately.

PR #4183, opened 2026-09-10
fix(test): symlink bash in muse harness fixtures to prevent macOS SIGKILL 137

lmuehleisen:
## Summary
Fixes `tests/fm-muse-harness.test.sh` failures on macOS by replacing `cp "$(command -v bash)"` in process detection and launch fixtures with `ln -sf "$(command -v bash)"`, and unsetting ambient `JETSKI_APP_DATA_DIR` and `GEMINI_CLI` markers.

## Problem & Root Cause
The failure `fm-harness.sh under process 'muse-bin-0.1.0-R708.1' reported '', expected muse` was previously assumed to be caused by `muse` not being installed on the host. Empirical testing confirmed that `muse` installation status is not the cause:
1. On macOS, `/bin/bash` is a platform-signed system binary (`Identifier=com.apple.bash`).
2. When `/bin/bash` is copied via `cp` to a new filename such as `muse-bin-0.1.0-R708.1` or `muse-bin-test-version`, macOS AMFI (code-signing enforcement) terminates execution of the copied binary immediately with `SIGKILL` (exit code 137).
3. Because the child process was terminated before running `fm-harness.sh`, the output was empty string `''`.
4. In Linux CI, platform binary copying does not trigger code-signing termination, which is why the suite passed in CI but failed consistently on macOS.

## Fix
1. **Use `ln -sf` for fixtures**: Symlink the system bash binary instead of copying it in `make_spawn_fakebin`, `test_detects_versioned_process_ancestor`, and `test_detection_is_anchored`. Symlinking preserves the valid code signature while allowing `ps -o comm=` to report the symlinked process name (matching the pattern already documented and used in `tests/fm-omp-harness.test.sh`).
2. **Add documentation comments**: Included comments above each fixture symlink explaining that copied platform-signed binaries get SIGKILLed under macOS code-signing enforcement so future refactors do not revert them to `cp`.
3. **Ambient marker hygiene**: Added `JETSKI_APP_DATA_DIR` and `GEMINI_CLI` to the suite's ambient marker unset line alongside existing markers (`CLAUDECODE`, `PI_CODING_AGENT`, etc.) so that suite runs invoked from within `agy` or `gemini` do not leak foreign marker identities into ancestry detection.
4. **No gate needed**: No `command -v muse` gate is needed; all 26 assertions are pure contract/fixture tests that run and pass on any host with or without `muse` installed.

## Verification
- `bin/fm-test-run.sh tests/fm-muse-harness.test.sh`: all 26 tests pass cleanly without skips or failures (`total=1 failed=0 skipped_gate=0`).
- `bin/fm-test-run.sh --check-coverage`: passes (`FM_TEST_COVERAGE ok total=194 parallel=24 serial=157 serial_shards=5 serial_unhinted=12 herdr=13`).
- `PATH="/private/tmp/fm-muse-test-gate-m3-tools:$PATH" bin/fm-lint.sh`: passes cleanly (`ShellCheck 0.11.0`, `actionlint 1.7.12`).


lmuehleisen:
Closing: opened unintentionally against upstream. This is a macOS-specific test-fixture fix meant for my own fork, not an upstream contribution, and it was raised by automation that resolved the wrong remote. Apologies for the noise.

PR #4197, opened 2026-09-11
chore: reconcile with kunchenguid/firstmate upstream

uayyagari:
## What is the problem

`applypass/firstmate` had drifted 329 commits behind canonical upstream `kunchenguid/firstmate` while carrying 11 local customizations. The merge base is `a5fe1bcc`. Every upstream fix, backend, and safety change since then was unavailable here, and the gap was growing faster than the fork was.

The customizations were not all worth keeping. Several solved a problem upstream has since solved differently and better, so a straight "merge and preserve everything" would have re-introduced machinery upstream deliberately does not have, on top of a 329-commit conflict resolution nobody could review.

## What was the fix

**Base chosen: current `upstream/main` (`4768e98d`), not the parked branch.** The parked `origin/reconcile/kunchenguid-upstream-20260821` was built for the old "preserve all 11" plan and was already ~173 behind upstream. Because the decided end-state drops five customizations and reduces a sixth, continuing that branch would have meant folding everything in and then deleting it - more work and more ways to leave something behind. Basing on current upstream and re-adding only the kept delta makes the drops self-evident: a dropped customization is simply never re-added, and there is nothing to prove absent by inspection.

**The merge.** `f1f110a2` records `origin/main` as a second parent, so the fork's history is preserved and this lands as an ordinary merge, while the tree is upstream's exactly. The four commits on top are one per kept concern, so this is reviewable a concern at a time rather than as one 329-commit conflict resolution.

**Per customization:**

| # | Customization | How it landed |
|---|---|---|
| 1 | `sc-`/type branch convention | **Kept, re-ported inline** (`0cb7ebb8`). The branch and PR-title convention, the `+ticket:<prefix>` registry flag, `fm-merge-local` branch discovery with `--branch` recovery, the same recovery in `fm-review-diff`, `fm-promote`'s ticket-aware ship-branch instruction, and the `fm-bearings.v1` `task` -> `branch` field rename. Two obsolete parts dropped: pre-window fail-closed mode resolution in `fm-spawn.sh` (upstream now requires `--mode`) and the explicit meta-write failure check (upstream publishes meta atomically under lock). |
| 2 | Batched defect-class decision | **Kept, reconciled** (`6a6b9748`). Re-added on top of upstream's 5-step `ask-user-authority` rewrite, cross-reference moved off the retired step 7 onto upstream's step 4. |
| 3 | `cursor-agent` adapter | **Dropped - took upstream's `cursor` adapter**, including `docs/supervision-protocols/cursor.md` and upstream's restructured harness references. |
| 4 | Context-budget guard (180k) | **Dropped entirely.** No `bin/fm-context-budget.sh`, no ceiling stub, no hook slot. |
| 5 | Handover + helm takeover | **Reshaped to its manual core** (`b8c77aba`). Kept: `bin/fm-handover.sh` (prepare / check / release / show / consume / pending, helm-gated, captain-invoked), `bin/fm-lock.sh release` and `clear --pid N`, and one line naming the exact clear command in every live-holder refusal and in `status`. Dropped: the automatic trigger (`fm-session-pulse.sh`), all token and silence measurement (`fm-context-measure-lib.sh`, transcript mtime), automatic helm takeover (`fm-helm-lib.sh`), the `.helm-activity` markers, and the awaiting-the-captain block with its `fm-decided.sh` / `fm-backlog-record-lib.sh` dependencies. |
| 6 | `resolve --no-work` | **Dropped - took upstream** `bin/fm-captain-hold.sh` and both the `decision-hold-lifecycle` and `captain-hold-lifecycle` skills. |
| 7+11 | PR body sections | **Kept, Option A.** Upstream's captain-words-only `--intent` contract taken as-is; the three sections are shaped after the PR opens, in the same post-open edit that sets the `sc-` title prefix. This PR body is that shape. |
| 8 | Trace stand-down failures | **Dropped** (rode inside #4). |
| 9 | Exclude zero-usage entries | **Dropped** (rode inside #5's measurement). |
| 10 | Worktree isolation identity | **Kept, reconciled inline** (`2d99bcd3`). Upstream's `spawn_worktree_isolated` kept as the base, including its git-common-dir check; its three path-text comparisons replaced with device-and-inode identity tests. |

**The six expected conflict files.** `bin/fm-spawn.sh` and `bin/fm-supervision-instructions.sh` stopped being collisions when #3 was dropped and #1's spawn hunks were dropped as obsolete; both are upstream's verbatim, with #10's identity change applied on top of upstream's current predicate in `fm-spawn.sh`. `bin/fm-brief.sh`, `bin/fm-merge-local.sh`, `bin/fm-bearings-snapshot.sh`, and `bin/fm-fleet-sync.sh` are upstream's current versions with #1's changes hand-applied - no textual merge was attempted on any of them.

## Proof of work

**Lint.** `bin/fm-lint.sh` exits 0: ShellCheck 0.11.0 (pinned), actionlint 1.7.12 (pinned), 3 workflow files valid.

**Tests.** `bin/fm-test-run.sh --changed --base f1f110a2` selects 75 files. Three of them drive real Herdr lifecycle behaviour, which this task is not scaffolded to run, so 72 were run: **63 pass, 9 fail, and every one of the 9 is an environment limit on this machine, not a regression.**

| Failing test | Why |
|---|---|
| `fm-captain-hold-lifecycle`, `fm-pr-merge`, `fm-bearings-board`, `fm-teardown`, `fm-backlog-atomicity`, `fm-bearings-snapshot`, `fm-backend`, `fm-control-relaunch` | `bin/fm-tasks-axi-lib.sh` sets `FM_TASKS_AXI_MIN=0.2.4`; the installed `tasks-axi` is 0.2.3. Each stops on the same refusal, `fm-captain-hold: compatible tasks-axi is required` or `automatic backlog transitions require tasks-axi 0.2.4 or newer`. None of these scripts is touched by this branch. |
| `fm-calm-pi-extension` | One case, `render_export_dom ... on a Chrome that never finished`. The whole fake-Chrome harness is defined inside `tests/fm-calm-pi-extension.test.sh`, which is upstream's at `746fc0aa` and unchanged here; nothing on this branch is in its dependency path. |

Four more failed on this machine's `asdf` shims and pass once a real interpreter is on PATH, so they are listed as passing above: `fm-claude-trust`, `fm-spawn-dispatch-profile`, `fm-trace-context-spawn`, and `fm-backend`'s trust step all hit `node` exiting 126 under the overridden `HOME` these tests use, and `fm-kimi-harness` needs a `python3` with `tomllib` that survives the same override (the shim exits 126, `/usr/bin/python3` is 3.9). `fm-test-run.test.sh` failed once on a debug file I had left in `tests/`; removed, it exits 0.

`fm-bearings-snapshot` stops at a `tasks-axi hold` before reaching the case this branch changed, so that case was run on its own: `--include-prs is the only path that fetches, and it enriches correctly` passes, including the `.branch == "fm/ship-task"` assertion.

Directly relevant suites, all green: `fm-session-handover` (10), `fm-session-start` (53), `fm-tangle-guard`, `fm-brief`, `fm-merge-local`, `fm-project-mode`, `fm-review-diff`, `fm-ask-user-authority`, `fm-task-delivery`, `fm-documentation-audiences`, `fm-lint`, `fm-test-isolation-proof`.

**Tests whose behavior changed, and why:**

- `tests/fm-task-delivery.test.sh` - `fm-promote.sh` no longer emits `git checkout -b fm/<id>`, so the assertion on that literal was replaced with assertions on the new branch-hint wording, and the promote fixture's meta gained `project=` so the ticket flag resolves. A new case pins the ticketed variant. The contract changed; the assertions were re-pointed at it, not weakened.
- `tests/fm-tangle-guard.test.sh` - same reason, one grep re-pointed from `git checkout -b fm/` to the new first-action line.
- `tests/fm-bearings-snapshot.test.sh` - `.task == "ship-task"` became `.branch == "fm/ship-task"` for the `fm-bearings.v1` field rename.
- `tests/fm-brief.test.sh` - four new cases for the ticketless, ticketed, non-Shortcut-prefix, and `--help` forms of the branch convention.
- `tests/fm-session-start.test.sh` - the truncation-banner case asserts the stage list, which gains `handover`. One new case covers a replacement session finding a released handover.
- `tests/fm-ask-user-authority.test.sh` and `tests/fm-review-diff.test.sh` - additions only. The first pins #2's batched-defect-class section on upstream's rewritten skill, the second pins `--branch` recovery. No existing assertion in either file was touched.
- New: `tests/fm-project-mode.test.sh`, `tests/fm-merge-local.test.sh`, `tests/fm-session-handover.test.sh`, all registered in `bin/fm-test-run.sh`'s family map and deliberately not in the proven-isolated parallel shard set.

No test was deleted or weakened to make anything pass.

**Kept items are functional.** #1: `bin/fm-brief.sh` renders both branch rules and `--help` states the convention; `fm-project-mode.sh` emits the third `+ticket:` field; `fm-merge-local.sh` and `fm-review-diff.sh` both discover the branch from the worktree HEAD and both accept `--branch` recovery; `fm-promote.sh` resolves the ticket flag before flipping state; `fm-bearings-snapshot.sh` emits `branch:`. #2: the section is present in upstream's rewritten skill with the AGENTS.md line re-anchored. #5: `fm-handover.sh` runs its whole lifecycle and `fm-lock.sh` names the clear command. #10: `tests/fm-tangle-guard.test.sh`'s `test_spawn_isolation_is_identity_not_path_text` passes all four rows - the canonical worktree spawns, the same worktree under a case-variant spelling spawns instead of being refused as "not a worktree root", and the spawning project is refused under both spellings by the reason that names it.

**Dropped items are gone, with no orphan references.** A tree-wide grep for `helm-activity`, `helm-takeover`, `context-budget`, `fm-decided`, `fm-session-pulse`, `fm-awaiting-captain`, `fm-context-measure`, `fm-helm-lib`, `fm-backlog-record-lib`, and `handover-due` returns nothing outside `.git`. No `AGENTS.md` state-inventory line, skill, doc, or `.claude/settings.json` hook slot describes machinery that is not here. The remaining `cursor-agent` hits are upstream's own references to the CLI binary name, which is still `cursor-agent` even though the adapter is `cursor`.

**#5 is reduced to its manual core.** `bin/` contains `fm-handover.sh`, `fm-lock.sh`, `fm-lock-lib.sh`, and `fm-session-lock-lib.sh`, and nothing else from the fork's handover or helm machinery. Nothing measures a session, and no code path takes the helm from a live holder.

**This rendered PR body shows all three sections.** Confirmed after opening; see the follow-up edit.

## Two things outside this PR

- `data/captain.md` still says to set `config/crew-harness=cursor-agent`; the value is now `cursor`. That file is gitignored and lives outside this worktree, so it is a follow-up for the captain's home rather than a change here.
- `bin/fm-project-mode.sh`'s unknown-mode `exit 3` refusal was **not** re-ported. Upstream deliberately owns warn-and-default there and its header states the fallback; changing it is a cross-consumer contract change (`fm-remote-home-seed.sh`, `fm-spawn.sh`, `fm-fleet-sync.sh`, `fm-home-seed.sh`) that the agreed dispositions do not sanction. Consequence: `fm-fleet-sync.sh`'s fail-closed guard now fires only when `fm-project-mode.sh` itself crashes or is missing, so a mistyped `local-only` in the registry is still not caught there.


uayyagari:
oops, this was supposed to be a PR against my fork. Sorry.

PR #4204, opened 2026-09-11
feat: add daily deployment secret parity check

bingb0t5:
## CEO overview

- **What is changing:** Add a scheduled/operator-armed daily parity check for approved Render, Coolify, and n8n deployment secrets and fixed signals URLs.
- **Why it matters:** Deployment drift is detected before one environment silently stops sharing the contract.
- **Customer or business impact:** No provider values are changed; operators receive one redacted alert for each newly observed mismatch set.
- **Risk and rollout:** The check is read-only, opt-in through the existing operator-home registration path, and skips the n8n credential-store assistant bearer.

## What changed technically

Added `bin/fm-secret-parity-check.sh` with separate Coolify and Render adapters, in-memory comparison, Coolify quote unwrapping, n8n token-name membership validation, fixed URL pins, duplicate-alert suppression, and value-free private finding records.
Added executable-interface coverage for equal, mismatch, missing, quoted-Coolify, membership, pins, duplicate-alert, redaction, and `fm-check-register.sh` operator-home registration behavior.

## Validation

- **Checks passed:** `bin/fm-test-run.sh tests/fm-secret-parity-check.test.sh`; ShellCheck 0.11.0; `bin/fm-lint.sh`; `bin/fm-doc-audience-check.sh`; one read-only live provider sweep with no alert output.
- **Checks not run:** `/no-mistakes`, per the task instructions.
- **Evidence and limitations:** Tests generate temporary provider values and assert that alerts and finding records contain only approved names and environments.

## Module-boundary decision

Not applicable: documentation/config-only change.

## Decision needed

No decision required.


bingb0t5:
Opened against the upstream parent by mistake; this work belongs in the bingb0t5 fork.

PR #4217, opened 2026-09-11
fix(opencode): deliver actionable wakes to most recently idle session

AndrewGoldfinch:
In a home with more than one OpenCode session, the primary watcher delivered every actionable wake to the session that first armed it, not the one in use. This tracks the most recently idle top-level session (resolving subagent child idles to the delegating top-level session via a bounded parentID walk) and uses it for actionable-wake delivery; a child/subagent idle never retargets delivery.

Coverage: behavioral plugin tests (most-recent-idle delivery, subagent-child skip) plus neighboring continuity tests green; busy-adapter wiring suite green; lint and doc-audience checks clean.


AndrewGoldfinch:
Superseded by https://github.com/kunchenguid/firstmate/pull/4249, which raises this same fix through no-mistakes.

PR #4230, opened 2026-09-11
test(watch): gate backlog-hold away-record fixture on tasks-axi

cipherholdingsllc:
One test was crashing on machines without `tasks-axi` installed instead of politely skipping like its siblings.

## The problem

`test_backlog_hold_never_rechecked_while_away_record_exists` in `tests/fm-watch-triage.test.sh` calls `make_hold_home`, which shells out to `tasks-axi` to build the backlog fixture. Every other `make_hold_home` caller in the file guards with `command -v tasks-axi` and prints `skip: tasks-axi not found (...)` — this one didn't, so on a host without the tool the suite reports `not ok - could not build the backlog-hold fixture` instead of a clean skip.

## The fix

Adds the same `command -v tasks-axi` guard at the top of the test, matching the sibling pattern verbatim. Two lines, test-only, no production code.

## Verified

On a host without `tasks-axi` (macOS), the test now reports `skip: tasks-axi not found (away-record backlog hold)` and exits clean; previously it failed. `bash -n` clean.

## Note on the no-mistakes gate

This PR was prepared manually outside the no-mistakes pipeline, so the `PR must be raised via no-mistakes` check will fail — flagging in advance so the red X isn't a surprise. Happy to re-submit through the pipeline or hand this off as a patch if you prefer; the diff is a two-line test guard.

### DIFF
--- a/tests/fm-watch-triage.test.sh
+++ b/tests/fm-watch-triage.test.sh
@@ -4665,6 +4665,8 @@ test_live_captain_held_first_sight_silenced_by_away_record() {
 
 test_backlog_hold_never_rechecked_while_away_record_exists() {
   local dir out capture wakes
+  command -v tasks-axi >/dev/null 2>&1 \
+    || { echo "skip: tasks-axi not found (away-record backlog hold)"; return 0; }
   dir=$(make_hold_home away-record-backlog-hold 'done: PR https://example.test/pr/9 checks green' hold) \
     || fail "could not build the backlog-hold fixture"
   out="$dir/watch.out"; capture="$dir/pane.txt"


cipherholdingsllc:
**Author review**

**What landed:** a `command -v tasks-axi` skip-guard at the top of `test_backlog_hold_never_rechecked_while_away_record_exists`, matching the four other `make_hold_home` callers in the same file.

**Residual risk:** none structural — the only question is the skip label wording (`away-record backlog hold`); it follows the file's existing naming style but is easy to bikeshed.

**Verified:** on a macOS host without `tasks-axi` the test now prints `skip: tasks-axi not found (away-record backlog hold)` and exits clean (previously `not ok - could not build the backlog-hold fixture`); `bash -n` clean. Not run: the full suite on a host *with* tasks-axi — the guarded path is byte-identical to siblings that already exercise it.

cipherholdingsllc:
Closing this — resubmitting through the required no-mistakes pipeline so the gate check is real rather than manually flagged. Thanks for the patience.

PR #4244, opened 2026-09-11
fix(bin): generalize the watcher-kill seatbelt to deny any broad process kill

christophmeise:
## Intent

Goal: close the class of defect where a test inside one lane's no-mistakes pipeline can kill work in another lane, not merely narrow the one pattern that caused the 2026-09-10 incident.

Incident, established (not assumed): the no-mistakes TEST-STEP pipeline agent - not a committed test - ran an ad-hoc `pgrep -f "lavish-axi poll"` during cleanup and killed a live sibling crewmate whose launch-brief argv contained that phrase. `pgrep -f` matches the full command line, so an accidentally-broad phrase reached across the shared process table. Worktrees are isolated; the process table is not.

Sweep (acceptance criterion: sweep, do not spot-fix): I swept all of bin/ and tests/ for process matching (pkill/pgrep/killall/ps/kill). No committed test does an unscoped pattern-kill: every one scopes by ancestry (pgrep -P / pkill -P), process group (kill -- -$pgid), or a unique per-invocation marker, or matches read-only. The one machine-wide-SHAPED reap (tests/fm-omp-primary-live-e2e.test.sh reap_lab: ps -axo | awk 'index($0,lab)') is keyed to a unique per-invocation marker LAB=$ROOT/.omp-live-e2e.$$ (repo root plus the test's own PID), which the task brief explicitly permits; it is justified and deliberately left unchanged (changing a live-e2e reap is risk with no safety gain). bin/ has no pattern-kill. The committed fixture the killer hand-copied (fm-bearings-board-render.test.sh) is already safe: bounded self-terminating stub plus home-scoped reap; the agent improvised the unscoped kill, so there is no fixture bug to fix.

Guard - is a general one achievable (acceptance criterion): argued, not assumed. The only mechanism that intercepts arbitrary agent shell commands is the PreToolUse seatbelt (bin/fm-arm-command-policy.mjs). It is intentionally disabled inside no-mistakes gate agents (.no-mistakes.yaml disable_project_settings: true, so gate agents do not adopt the fleet-captain identity), and re-enabling it would reintroduce that worse problem. So a harness guard cannot reach the exact incident surface. A helper only protects opt-in callers; shimming kill (a shell builtin) is infeasible. Therefore a fully-general deterministic guard for the gate-agent surface is not achievable from within firstmate. Shipped the strongest achievable layered containment:

1. Generalized the seatbelt from the watcher-only broad-watcher-kill rule to a general broad-process-kill rule (new stable reason code). It denies pkill/killall that do not select by the caller's own ancestry/process-group/session, and a kill fed by an unscoped pgrep (command substitution, variable, or the pgrep | xargs kill pipe). It allows caller-scoped forms (-P/-g/-s), a specific kill <pid>, read-only discovery, and quoted data. It judges shape, not runtime identity - exactly as the existing watcher rule permits pkill -P <pid> regardless of the pid value. This also closes a pre-existing gap where pgrep -f fm-watch | xargs kill slipped past the watcher rule. Coverage claim, stated honestly after review: the seatbelt denies the known broad-kill forms enumerated in docs/arm-pretool-check.md for every agent that loads project settings (crewmates - the class that reported this - primaries, secondmates). It is a denylist over shell command text, and a denylist over shell has no completeness floor, so the residual list in that doc names the known gaps and is not a proof that nothing else gets through. The AGENTS.md shared-process-table rule is the containment for the no-mistakes gate-agent surface (disable_project_settings), which this seatbelt never reaches, and remains the containment for any bypass the residual list does not name; a deny-by-default allowlist redesign is filed as a separate follow-up.
2. Widened the transport prefilter (bin/fm-arm-pretool-check.sh) trigger set to fm-watch/pkill/killall/pgrep so these delegate to the classifier instead of being fast-allowed; the trigger set is coupled to the classifier and documented as such.
3. Widened AGENTS.md section 8's watcher prohibition to the general shared-process-table rule (always loaded; reaches every firstmate/crewmate/secondmate agent, and is the containment for the gate-agent surface).
4. Updated docs/arm-pretool-check.md (the one-owner contract): purpose/boundary, prefilter trigger set, reason-code table, a new "General broad process kill" section, and the disable_project_settings coverage limit.

Honest limitations, stated in the doc and PR: (a) the seatbelt is inactive inside gate agents (disable_project_settings) - containment there is the AGENTS.md rule plus the crewmate brief's wait-discipline rule plus the already-self-terminating fixtures; (b) a discover-then-LITERAL-kill across two commands cannot be caught by any command-shape seatbelt, because a literal kill <pid> carries no evidence of foreign origin - the AGENTS.md rule is the containment for it.

Regression coverage (acceptance criterion: prove the guard refuses an out-of-group match, assert the refusal not merely that nothing crashed): tests/fm-arm-pretool-check.test.sh gains matrix K01-K09 (run through all five harness entry forms), test_broad_process_kill_contract (paired allow/deny direct assertions asserting the exact broad-process-kill verdict, including the exact incident shape, and asserting caller-scoped/literal forms stay allowed), widened prefilter strict-superset assertions, and a watcher pipe-xargs lock-in. I mutation-tested it: removing the deny makes the affected assertions fail (the guarded shapes wrongly allow), proving the tests are non-vacuous.

Constraint honored: this edits firstmate's shared tracked material; firstmate-coding-guidelines was loaded before editing. No process command was run machine-wide during the work.

## What Changed

- `bin/fm-arm-command-policy.mjs` adds a `broad-process-kill` reason code alongside `broad-watcher-kill`: it denies `pkill`/`killall` that do not select by the caller's own `-P`/`-g`/`-s` (parent/pgroup/session) scope, `fuser -k`, `kill` of pid `-1`, and any `kill` fed by unscoped discovery (`pgrep -f`, `ps`, `lsof`, `pidof`, `fuser`) via command substitution, a variable hop, or a `| xargs kill`/`xargs pkill` pipe; caller-scoped forms, a specific `kill <pid>`, `kill -0` probes, read-only discovery, `command -v`/`--help` query forms, and quoted data stay allowed. The parser also unwraps a leading `!` and `builtin`, and unsupported loop/`if`/`case` grammar falls back to a raw byte check for the same forms.
- `bin/fm-arm-pretool-check.sh` widens the fast-allow prefilter's trigger set from `fm-watch` alone to `fm-watch`/`kill`/`pgrep`/`ps`/`lsof`/`pidof`/`fuser`/`xargs`, so those commands delegate to the classifier instead of bypassing it; `AGENTS.md` section 8 widens the watcher-only prohibition to the general shared-process-table rule, and `docs/arm-pretool-check.md`, `docs/scripts.md`, and the harness-adapter reference document the new rule, prefilter coupling, reason-code row, enumerated residual gaps, and the `disable_project_settings` gate-agent coverage limit.
- `tests/fm-arm-pretool-check.test.sh` adds the K01–K09 matrix rows (run through all five harness entry forms), a `test_broad_process_kill_contract` set of paired allow/deny assertions on the exact `broad-process-kill` verdict including the `kill $(pgrep -f ...)` incident shape, widened prefilter strict-superset assertions, and a `pgrep -f fm-watch | xargs kill` watcher lock-in.

## Risk Assessment

⚠️ Medium: The change is clean after eight fix rounds (both round-7 regressions verified fixed and all prior deny/allow claims re-verified at HEAD), but it is a ~630-line extension to the PreToolUse seatbelt that now classifies every agent shell command carrying kill/ps/lsof/xargs bytes, with a documented set of accepted false positives and residual gaps that warrant follow-up observation rather than a block.

## Testing

Ran the committed seatbelt test file (all 342 assertions pass, including the new K01-K48 cross-harness matrix and the paired allow/deny broad-process-kill contract), proved the regression is real by comparing base vs target classifier verdicts (incident shapes and the pgrep|xargs watcher gap flip allow→deny; scoped forms unchanged) and by a mutation run that fails once the deny is removed, captured the agent-facing hook JSON for denied and allowed commands, and demonstrated the class closure live: with a real out-of-tree sibling process whose argv contained "lavish-axi poll", every incident-shaped cleanup was refused and the sibling survived, while the caller-scoped `pkill -P $$` was allowed and reaped only the caller's own child. No UI surface is involved, so evidence is CLI transcripts and logs. Worktree left clean; no stray processes.

<details>
<summary>Evidence: Live process-table demo: sibling survives denied incident cleanups, scoped kill reaps only own child</summary>

<code>step 1: two processes whose argv both contain &#39;lavish-axi poll&#39; (pid ppid command)&#10;54715 1 crewmate-brief lavish-axi poll fm-evidence-54650 sibling 120&#10;54717 54650 fixture-stub lavish-axi poll fm-evidence-54650 mine 120&#10;&#10;step 2: pkill -f &#34;lavish-axi poll&#34;&#10;hook -&gt; DENY (exit 2) [broad-process-kill] ...&#10;step 2b: kill $(pgrep -f &#34;lavish-axi poll&#34;)&#10;hook -&gt; DENY (exit 2) [broad-process-kill] ...&#10;step 2c: pgrep -f &#34;lavish-axi poll&#34; | xargs kill&#10;hook -&gt; DENY (exit 2) [broad-process-kill] ...&#10;none executed; both still alive&#10;&#10;step 3: pkill -P $$&#10;hook -&gt; ALLOW (exit 0, silent)&#10;RESULT: sibling 54715 SURVIVED&#10;RESULT: my stub 54717 reaped&#10;&#10;step 4: kill 54715&#10;hook -&gt; ALLOW (exit 0, silent)&#10;sibling gone; process table clean</code>

```text
# Live process-table demo. This shell (the 'test-step agent') is pid 54650; marker fm-evidence-54650

step 1: two processes whose argv both contain 'lavish-axi poll' (pid ppid command)
  54715     1 crewmate-brief lavish-axi poll fm-evidence-54650 sibling 120
  54717 54650 fixture-stub lavish-axi poll fm-evidence-54650 mine 120
  -> sibling ppid != 54650 (another lane); my stub ppid == 54650 (mine)

step 2: the incident cleanup, exactly as the 2026-09-10 agent ran it: pkill -f "lavish-axi poll"
  hook -> DENY  (exit 2) [broad-process-kill] a process kill that matches by command line or na...
step 2b: kill $(pgrep -f "lavish-axi poll")
  hook -> DENY  (exit 2) [broad-process-kill] a process kill that matches by command line or na...
step 2c: pgrep -f "lavish-axi poll" | xargs kill
  hook -> DENY  (exit 2) [broad-process-kill] a process kill that matches by command line or na...
  none executed (harness blocks a denied Bash call); both still alive:
    54715 crewmate-brief lavish-axi poll fm-evidence-54650 sibling 120
    54717 fixture-stub lavish-axi poll fm-evidence-54650 mine 120

step 3: the caller-scoped cleanup: pkill -P $$
  hook -> ALLOW (exit 0, silent)
  executing pkill -P $$ ...
  alive after the scoped kill:
    54715     1 crewmate-brief lavish-axi poll fm-evidence-54650 sibling 120
  RESULT: sibling 54715 SURVIVED
  RESULT: my stub 54717 reaped

step 4: tear down the sibling by explicit pid: kill 54715
  hook -> ALLOW (exit 0, silent)
  executed
  sibling gone; process table clean
```
</details>
<details>
<summary>Evidence: Before/after classifier verdicts (base 9074f9d vs target 7cbef3e)</summary>

<code>command | BEFORE | AFTER&#10;pkill -f &#34;lavish-axi poll&#34; | allow | deny broad-process-kill&#10;kill $(pgrep -f &#34;lavish-axi poll&#34;) | allow | deny broad-process-kill&#10;pgrep -f &#34;lavish-axi poll&#34; | xargs kill | allow | deny broad-process-kill&#10;killall claude | allow | deny broad-process-kill&#10;pkill node | allow | deny broad-process-kill&#10;pgrep -f fm-watch | xargs kill | allow | deny broad-watcher-kill&#10;kill -9 -1 | allow | deny broad-process-kill&#10;pkill -P 12345 | allow | allow&#10;kill $(pgrep -P $$) | allow | allow&#10;pgrep -f &#34;lavish-axi poll&#34; | head | allow | allow&#10;kill 76803 | allow | allow&#10;kill -- -$pgid | allow | allow</code>

```text
# Before/after verdicts: base 9074f9d classifier vs target 7cbef3e classifier
# (node bin/fm-arm-command-policy.mjs --root . --home . --command <cmd>)
command                                              | BEFORE (base)                | AFTER (target)
-----------------------------------------------------|------------------------------|------------------------------
pkill -f "lavish-axi poll"                           | allow                        | deny broad-process-kill a process kill that matches by command line or name across the shared process table, or signals every process via pid -1, is forbidden; target your own process group, your own descendants, or a specific pid
kill $(pgrep -f "lavish-axi poll")                   | allow                        | deny broad-process-kill a process kill that matches by command line or name across the shared process table, or signals every process via pid -1, is forbidden; target your own process group, your own descendants, or a specific pid
pgrep -f "lavish-axi poll" | xargs kill              | allow                        | deny broad-process-kill a process kill that matches by command line or name across the shared process table, or signals every process via pid -1, is forbidden; target your own process group, your own descendants, or a specific pid
killall claude                                       | allow                        | deny broad-process-kill a process kill that matches by command line or name across the shared process table, or signals every process via pid -1, is forbidden; target your own process group, your own descendants, or a specific pid
pkill node                                           | allow                        | deny broad-process-kill a process kill that matches by command line or name across the shared process table, or signals every process via pid -1, is forbidden; target your own process group, your own descendants, or a specific pid
pgrep -f fm-watch | xargs kill                       | allow                        | deny broad-watcher-kill a broad process kill targeting the firstmate watcher is forbidden
kill -9 -1                                           | allow                        | deny broad-process-kill a process kill that matches by command line or name across the shared process table, or signals every process via pid -1, is forbidden; target your own process group, your own descendants, or a specific pid
pkill -P 12345                                       | allow                        | allow
kill $(pgrep -P $$)                                  | allow                        | allow
pgrep -f "lavish-axi poll" | head                    | allow                        | allow
kill 76803                                           | allow                        | allow
kill -- -$pgid                                       | allow                        | allow
```
</details>
<details>
<summary>Evidence: PreToolUse hook transcript (Claude entry form) showing deny JSON and silent allow</summary>

```text
# PreToolUse hook transcript (Claude Code entry form: JSON on stdin, --claude)
# Exit 2 + hookSpecificOutput deny JSON on stderr = the harness blocks the Bash call and shows the agent the reason.

$ echo '{"tool_name":"Bash","tool_input":{"command":"pkill -f \"lavish-axi poll\""}}' | bin/fm-arm-pretool-check.sh --claude
exit=2
stderr:
{
  "hookSpecificOutput": {
    "hookEventName": "PreToolUse",
    "permissionDecision": "deny"
  },
  "systemMessage": "[broad-process-kill] a process kill that matches by command line or name across the shared process table, or signals every process via pid -1, is forbidden; target your own process group, your own descendants, or a specific pid"
}

$ echo '{"tool_name":"Bash","tool_input":{"command":"kill $(pgrep -f \"lavish-axi poll\")"}}' | bin/fm-arm-pretool-check.sh --claude
exit=2
stderr:
{
  "hookSpecificOutput": {
    "hookEventName": "PreToolUse",
    "permissionDecision": "deny"
  },
  "systemMessage": "[broad-process-kill] a process kill that matches by command line or name across the shared process table, or signals every process via pid -1, is forbidden; target your own process group, your own descendants, or a specific pid"
}

$ echo '{"tool_name":"Bash","tool_input":{"command":"pgrep -f \"lavish-axi poll\" | xargs kill"}}' | bin/fm-arm-pretool-check.sh --claude
exit=2
stderr:
{
  "hookSpecificOutput": {
    "hookEventName": "PreToolUse",
    "permissionDecision": "deny"
  },
  "systemMessage": "[broad-process-kill] a process kill that matches by command line or name across the shared process table, or signals every process via pid -1, is forbidden; target your own process group, your own descendants, or a specific pid"
}

$ echo '{"tool_name":"Bash","tool_input":{"command":"pkill -P $$"}}' | bin/fm-arm-pretool-check.sh --claude
exit=0
(silent allow: no stdout, no stderr)

$ echo '{"tool_name":"Bash","tool_input":{"command":"kill 76803"}}' | bin/fm-arm-pretool-check.sh --claude
exit=0
(silent allow: no stdout, no stderr)
```
</details>
<details>
<summary>Evidence: Mutation check: deny line removed → test fails at K01</summary>

`not ok - K01 via codex must deny, got exit 0`

```text
# Mutation check: broad-process-kill deny line removed from classifier, then tests/fm-arm-pretool-check.test.sh run
not ok - K01 via codex must deny, got exit 0
...
exit=
```
</details>
<details>
<summary>Evidence: Full tests/fm-arm-pretool-check.test.sh log (342 ok)</summary>

```text
ok - matrix A01: allow through all five entry forms
ok - matrix A02: allow through all five entry forms
ok - matrix A03: allow through all five entry forms
ok - matrix A04: allow through all five entry forms
ok - matrix A05: allow through all five entry forms
ok - matrix A06: allow through all five entry forms
ok - matrix A07: allow through all five entry forms
ok - matrix A08: allow through all five entry forms
ok - matrix A09: allow through all five entry forms
ok - matrix A10: allow through all five entry forms
ok - matrix A11: allow through all five entry forms
ok - matrix A12: allow through all five entry forms
ok - matrix A13: allow through all five entry forms
ok - matrix A14: allow through all five entry forms
ok - matrix A15: allow through all five entry forms
ok - matrix A16: allow through all five entry forms
ok - matrix A17: allow through all five entry forms
ok - matrix R01: allow through all five entry forms
ok - matrix R02: allow through all five entry forms
ok - matrix R03: allow through all five entry forms
ok - matrix R04: allow through all five entry forms
ok - matrix R05: allow through all five entry forms
ok - matrix R06: allow through all five entry forms
ok - matrix R07: allow through all five entry forms
ok - matrix R08: allow through all five entry forms
ok - matrix R09: allow through all five entry forms
ok - matrix R10: allow through all five entry forms
ok - matrix R11: allow through all five entry forms
ok - matrix R12: allow through all five entry forms
ok - matrix R13: allow through all five entry forms
ok - matrix R14: allow through all five entry forms
ok - matrix R15: allow through all five entry forms
ok - matrix R16: allow through all five entry forms
ok - matrix R17: allow through all five entry forms
ok - matrix R18: allow through all five entry forms
ok - matrix R19: allow through all five entry forms
ok - matrix D01: deny through all five entry forms
ok - matrix D02: deny through all five entry forms
ok - matrix D03: deny through all five entry forms
ok - matrix D04: deny through all five entry forms
ok - matrix D05: deny through all five entry forms
ok - matrix D06: deny through all five entry forms
ok - matrix D07: deny through all five entry forms
ok - matrix D08: deny through all five entry forms
ok - matrix D09: deny through all five entry forms
ok - matrix D10: deny through all five entry forms
ok - matrix D11: deny through all five entry forms
ok - matrix D12: deny through all five entry forms
ok - matrix D13: deny through all five entry forms
ok - matrix D14: deny through all five entry forms
ok - matrix D15: deny through all five entry forms
ok - matrix D16: deny through all five entry forms
ok - matrix D17: deny through all five entry forms
ok - matrix D18: deny through all five entry forms
ok - matrix D19: deny through all five entry forms
ok - matrix D20: deny through all five entry forms
ok - matrix D21: deny through all five entry forms
ok - matrix D22: deny through all five entry forms
ok - matrix D23: deny through all five entry forms
ok - matrix D24: deny through all five entry forms
ok - matrix D25: deny through all five entry forms
ok - matrix D26: deny through all five entry forms
ok - matrix D27: deny through all five entry forms
ok - matrix D28: deny through all five entry forms
ok - matrix D29: deny through all five entry forms
ok - matrix D30: deny through all five entry forms
ok - matrix D31: deny through all five entry forms
ok - matrix D32: deny through all five entry forms
ok - matrix D33: deny through all five entry forms
ok - matrix D34: deny through all five entry forms
ok - matrix D35: deny through all five entry forms
ok - matrix D36: deny through all five entry forms
ok - matrix D37: deny through all five entry forms
ok - matrix D38: deny through all five entry forms
ok - matrix D39: deny through all five entry forms
ok - matrix D40: deny through all five entry forms
ok - matrix D41: deny through all five entry forms
ok - matrix D42: deny through all five entry forms
ok - matrix D43: deny through all five entry forms
ok - matrix D44: deny through all five entry forms
ok - matrix D45: deny through all five entry forms
ok - matrix D46: deny through all five entry forms
ok - matrix D47: deny through all five entry forms
ok - matrix D48: deny through all five entry forms
ok - matrix D49: deny through all five entry forms
ok - matrix D50: deny through all five entry forms
ok - matrix D51: deny through all five entry forms
ok - matrix D52: deny through all five entry forms
ok - matrix D53: deny through all five entry forms
ok - matrix D54: deny through all five entry forms
ok - matrix D55: deny through all five entry forms
ok - matrix D56: deny through all five entry forms
ok - matrix D57: deny through all five entry forms
ok - matrix D58: deny through all five entry forms
ok - matrix E01: allow through all five entry forms
ok - matrix E02: allow through all five entry forms
ok - matrix E03: allow through all five entry forms
ok - matrix E04: allow through all five entry forms
ok - matrix E05: deny through all five entry forms
ok - matrix E06: deny through all five entry forms
ok - matrix E07: deny through all five entry forms
ok - matrix E08: deny through all five entry forms
ok - matrix E09: deny through all five entry forms
ok - matrix E10: deny through all five entry forms
ok - matrix E11: deny through all five entry forms
ok - matrix E12: allow through all five entry forms
ok - matrix E13: allow through all five entry forms
ok - matrix E14: allow through all five entry forms
ok - matrix E15: allow through all five entry forms
ok - matrix E16: allow through all five entry forms
ok - matrix E17: allow through all five entry forms
ok - matrix K01: deny through all five entry forms
ok - matrix K02: deny through all five entry forms
ok - matrix K03: deny through all five entry forms
ok - matrix K04: deny through all five entry forms
ok - matrix K05: deny through all five entry forms
ok - matrix K06: allow through all five entry forms
ok - matrix K07: allow through all five entry forms
ok - matrix K08: allow through all five entry forms
ok - matrix K09: allow through all five entry forms
ok - matrix K10: deny through all five entry forms
ok - matrix K11: deny through all five entry forms
ok - matrix K12: deny through all five entry forms
ok - matrix K13: deny through all five entry forms
ok - matrix K14: deny through all five entry forms
ok - matrix K15: deny through all five entry forms
ok - matrix K16: deny through all five entry forms
ok - matrix K17: deny through all five entry forms
ok - matrix K18: allow through all five entry forms
ok - matrix K19: allow through all five entry forms
ok - matrix K20: allow through all five entry forms
ok - matrix K21: allow through all five entry forms
ok - matrix K22: allow through all five entry forms
ok - matrix K23: allow through all five entry forms
ok - matrix K24: deny through all five entry forms
ok - matrix K25: deny through all five entry forms
ok - matrix K26: deny through all five entry forms
ok - matrix K27: allow through all five entry forms
ok - matrix K28: allow through all five entry forms
ok - matrix K29: deny through all five entry forms
ok - matrix K30: deny through all five entry forms
ok - matrix K31: deny through all five entry forms
ok - matrix K32: deny through all five entry forms
ok - matrix K33: allow through all five entry forms
ok - matrix K34: allow through all five entry forms
ok - matrix K35: allow through all five entry forms
ok - matrix K36: allow through all five entry forms
ok - matrix K37: deny through all five entry forms
ok - matrix K38: deny through all five entry forms
ok - matrix K39: allow through all five entry forms
ok - matrix K40: allow through all five entry forms
ok - matrix K41: deny through all five entry forms
ok - matrix K42: deny through all five entry forms
ok - matrix K43: allow through all five entry forms
ok - matrix K44: deny through all five entry forms
ok - matrix K45: deny through all five entry forms
ok - matrix K46: deny through all five entry forms
ok - matrix K47: allow through all five entry forms
ok - matrix K48: allow through all five entry forms
ok - direct poli

... [2860 bytes truncated] ...

of-xargs: deny	broad-process-kill
ok - direct policy bpk-lsof-cmdsub: deny	broad-process-kill
ok - direct policy bpk-pidof-cmdsub: deny	broad-process-kill
ok - direct policy bpk-fuser-xargs: deny	broad-process-kill
ok - direct policy bpk-ps-var: deny	broad-process-kill
ok - direct policy bpk-fuser-k: deny	broad-process-kill
ok - direct policy bpk-fuser-k-namespace: deny	broad-process-kill
ok - direct policy bpk-fuser-k-signal: deny	broad-process-kill
ok - direct policy bpk-fuser-kill-long: deny	broad-process-kill
ok - direct policy bpk-loop-fuser-k: deny	broad-process-kill
ok - direct policy bpk-loop-ps-xargs: deny	broad-process-kill
ok - direct policy bpk-if-lsof-xargs: deny	broad-process-kill
ok - direct policy bpk-loop-pidof-cmdsub: deny	broad-process-kill
ok - direct policy bpk-loop-ps-cmdsub: deny	broad-process-kill
ok - direct policy bpk-kill-all-signal: deny	broad-process-kill
ok - direct policy bpk-kill-all-dashdash: deny	broad-process-kill
ok - direct policy bpk-kill-all-named-signal: deny	broad-process-kill
ok - direct policy bpk-kill-all-signal-dashdash: deny	broad-process-kill
ok - direct policy bpk-loop-kill-all: deny	broad-process-kill
ok - direct policy bpk-if-kill-all-dashdash: deny	broad-process-kill
ok - direct policy bpk-case-kill-all: deny	broad-process-kill
ok - direct policy bpk-subshell-pgrep-xargs: deny	broad-process-kill
ok - direct policy bpk-group-pgrep-xargs: deny	broad-process-kill
ok - direct policy bpk-cmdsub-pgrep-echo-xargs: deny	broad-process-kill
ok - direct policy bpk-subshell-ps-xargs: deny	broad-process-kill
ok - direct policy bpk-xargs-pkill-by-name: deny	broad-process-kill
ok - direct policy bpk-xargs-pkill-n1: deny	broad-process-kill
ok - direct policy bpk-xargs-killall: deny	broad-process-kill
ok - direct policy bpk-xargs-sudo-kill: deny	broad-process-kill
ok - direct policy bpk-xargs-command-kill: deny	broad-process-kill
ok - direct policy bpk-xargs-env-kill: deny	broad-process-kill
ok - direct policy bpk-xargs-long-option-kill: deny	broad-process-kill
ok - direct policy bpk-xargs-max-lines-bare: deny	broad-process-kill
ok - direct policy bpk-xargs-eof-bare: deny	broad-process-kill
ok - direct policy bpk-xargs-max-lines-attached: deny	broad-process-kill
ok - direct policy bpk-xargs-kill-signal: deny	broad-process-kill
ok - direct policy bpk-if-pkill-executed: deny	broad-process-kill
ok - direct policy bpk-if-killall-list-then-kill: deny	broad-process-kill
ok - direct policy bpk-loop-quoted-pattern-P: deny	broad-process-kill
ok - direct policy bpk-loop-quoted-pattern-g: deny	broad-process-kill
ok - direct policy bpk-loop-quoted-pattern-pgrep: deny	broad-process-kill
ok - direct policy bpk-lsof-table-awk-xargs: deny	broad-process-kill
ok - direct policy bpk-lsof-table-awk-cmdsub: deny	broad-process-kill
ok - direct policy bpk-if-lsof-table-awk: deny	broad-process-kill
ok - direct policy bpk-negated-pkill: deny	broad-process-kill
ok - direct policy bpk-double-negated-pkill: deny	broad-process-kill
ok - direct policy bpk-negated-kill-all: deny	broad-process-kill
ok - direct policy bpk-builtin-kill-all: deny	broad-process-kill
ok - direct policy bpk-time-pkill: deny	broad-process-kill
ok - direct policy bpk-negated-time-pkill: deny	broad-process-kill
ok - direct policy bpk-negated-time-kill-all: deny	broad-process-kill
ok - direct policy bpk-xargs-kill-all-dashdash: deny	broad-process-kill
ok - direct policy bpk-xargs-kill-all-signal: deny	broad-process-kill
ok - direct policy bpk-loop-kill-all-dashdash: deny	broad-process-kill
ok - direct policy bpk-for-kill-all-signal: deny	broad-process-kill
ok - direct policy bpk-if-kill-all-named: deny	broad-process-kill
ok - direct policy bpk-allow-parent: allow
ok - direct policy bpk-allow-pgroup: allow
ok - direct policy bpk-allow-parent-long: allow
ok - direct policy bpk-allow-session: allow
ok - direct policy bpk-allow-kill-parent-pgrep: allow
ok - direct policy bpk-allow-var-scoped: allow
ok - direct policy bpk-allow-pgrep-scoped-xargs: allow
ok - direct policy bpk-allow-pgrep-scoped-xargs-sep: allow
ok - direct policy bpk-allow-literal-pid: allow
ok - direct policy bpk-allow-kill-pgroup-var: allow
ok - direct policy bpk-allow-kill-pgroup-literal: allow
ok - direct policy bpk-allow-kill-pgroup-attached: allow
ok - direct policy bpk-allow-kill-sighup-pid: allow
ok - direct policy bpk-allow-kill-plain-pid: allow
ok - direct policy bpk-allow-loop-kill-sighup-pid: allow
ok - direct policy bpk-allow-loop-kill-pgroup: allow
ok - direct policy bpk-allow-wait-loop-tail: allow
ok - direct policy bpk-allow-wait-loop-head: allow
ok - direct policy bpk-allow-if-kill-git-log: allow
ok - direct policy bpk-allow-loop-kill-sleep: allow
ok - direct policy bpk-allow-negated-scoped-pkill: allow
ok - direct policy bpk-allow-negated-kill-probe: allow
ok - direct policy bpk-allow-negated-pgrep: allow
ok - direct policy bpk-allow-negated-test: allow
ok - direct policy bpk-allow-negated-time-sleep: allow
ok - direct policy bpk-allow-negated-time-scoped-pkill: allow
ok - direct policy bpk-allow-command-v-pkill: allow
ok - direct policy bpk-allow-command-v-pkill-redirected: allow
ok - direct policy bpk-allow-command-v-killall: allow
ok - direct policy bpk-allow-command-v-pgrep: allow
ok - direct policy bpk-allow-command-v-lsof: allow
ok - direct policy bpk-allow-pkill-help: allow
ok - direct policy bpk-allow-pkill-version: allow
ok - direct policy bpk-allow-killall-list: allow
ok - direct policy bpk-allow-if-command-v-pkill: allow
ok - direct policy bpk-allow-if-command-v-killall: allow
ok - direct policy bpk-allow-if-which-pkill: allow
ok - direct policy bpk-allow-if-pkill-help: allow
ok - direct policy bpk-allow-xargs-kill-zero-probe: allow
ok - direct policy bpk-allow-xargs-n1-kill-zero-probe: allow
ok - direct policy bpk-allow-xargs-pkill-scoped: allow
ok - direct policy bpk-allow-xargs-kill-signal-only: allow
ok - direct policy bpk-allow-xargs-kill-pgroup-var: allow
ok - direct policy bpk-allow-literal-signal: allow
ok - direct policy bpk-allow-parent-attached: allow
ok - direct policy bpk-allow-parent-attached-self: allow
ok - direct policy bpk-allow-parent-attached-var: allow
ok - direct policy bpk-allow-pgrep-attached-xargs: allow
ok - direct policy bpk-allow-kill-pgrep-attached: allow
ok - direct policy bpk-allow-loop-scoped-attached: allow
ok - direct policy bpk-allow-loop-scoped-quoted-value: allow
ok - direct policy bpk-allow-loop-pkill-quoted-value: allow
ok - direct policy bpk-allow-session-signal: allow
ok - direct policy bpk-allow-loop-scoped: allow
ok - direct policy bpk-allow-loop-scoped-pkill: allow
ok - direct policy bpk-allow-kill-pidfile: allow
ok - direct policy bpk-allow-echo-xargs: allow
ok - direct policy bpk-allow-cat-xargs: allow
ok - direct policy bpk-allow-kill-jobs: allow
ok - direct policy bpk-allow-kill-zero-probe: allow
ok - direct policy bpk-allow-readonly-pgrep: allow
ok - direct policy bpk-allow-ps-grep: allow
ok - direct policy bpk-allow-lsof-readonly: allow
ok - direct policy bpk-allow-fuser-readonly: allow
ok - direct policy bpk-allow-loop-ps-readonly: allow
ok - direct policy bpk-allow-data-mention: allow
ok - --command=<val> equals-form parses correctly
ok - --background is accepted for interface parity and is never itself a deny signal
ok - unknown CLI flag is rejected
ok - stdin grok schema (toolInput.command): denied with Grok-shaped stdout JSON
ok - stdin claude/codex schema (tool_input.command): blessed shape allowed
ok - stdin claude/codex schema (tool_input.command): backgrounded shape denied
ok - stdin: unrelated command is a fast allow
ok - transport prefilter is a strict superset: only trigger-free commands fast-allow; every fm-watch, broad-kill, and quoting-decoder-marker command reaches the classifier
ok - fail-open: empty stdin
ok - fail-open: unparseable JSON on stdin
ok - fail-open: missing jq on stdin path
ok - fail-open: missing classifier runtime
ok - --claude: stdout empty, stderr carries hookSpecificOutput deny JSON
ok - default mode: stdout carries Grok-shaped decision JSON on deny
ok - allow is silent on both stdout and stderr in default and --claude mode
ok - bin/fm-arm-pretool-check.sh is clean under bin/fm-lint.sh
```
</details>

## Pipeline

<details>
<summary>🔧 **Review** - 5 issues found → auto-fixed (7) ✅</summary>

🔧 Fix: close signal-flag, xargs-option, and ps/lsof discovery kill gaps
- ⚠️ `bin/fm-arm-command-policy.mjs:783` - `fuser` is classified only as a discovery feed, but `fuser -k` is itself an attribute-selected kill: `fuser -k 3000/tcp`, `fuser -k -n tcp 3000`, and `fuser -KILL -k 3000/tcp` all return `allow` (verified against the target commit), while the equivalent `lsof -ti :3000 | xargs kill -9` is denied. `fuser -k &lt;port&gt;/tcp` is a common &#39;free the port&#39; idiom and reaches every process on that port host-wide, including a sibling lane&#39;s server, so the class the intent claims to close is still reachable through the seatbelt by this shape. Fix within the existing per-node recognition: in the `commandName` block, set `broadProcessKill` when `commandName === &#34;fuser&#34;` and any arg matches `/^-[A-Za-z]*k/` or `--kill`; add the same `fuser` + `-k` case to `rawMentionsGeneralBroadKill` for unsupported grammar; add paired assertions (`fuser -k 3000/tcp` deny, `fuser 3000/tcp` allow) to `test_broad_process_kill_contract` and a matrix row; mention `fuser -k` in the docs bullet list.
- ⚠️ `bin/fm-arm-command-policy.mjs:67` - The unsupported-grammar fallback was not extended alongside the discovery set: it only recognises `pkill`, `killall`, and a `pgrep`+`kill|xargs` pair, so a ps/lsof/pidof/fuser-fed kill inside any loop/`if`/`case` body is allowed. Verified `allow` at the target commit for `while true; do ps aux | grep &#34;lavish-axi poll&#34; | awk &#39;{print $2}&#39; | xargs kill; sleep 1; done`, `if true; then lsof -ti :3000 | xargs kill -9; fi`, `for x in 1; do kill $(pidof node); done`, and `until false; do kill $(ps aux | grep &#34;lavish-axi poll&#34; | awk &#39;{print $2}&#39;); done`, while the same commands with `pgrep -f` in place of `ps`/`lsof`/`pidof` are denied. These are not loop-routed dataflow (docs residual 1) - the pipe/substitution is complete inside one body - so the AST path would catch them if the body were parsed; it is only the `do`/`then` prefix that routes them to the raw check. Fix by mirroring the existing pgrep branch: in `rawMentionsGeneralBroadKill`, also return true when `/\b(?:ps|pidof|fuser)\b|\blsof\b[^;|&amp;\n)`]*\s-[A-Za-z]*t/` matches and `/\b(?:kill|xargs)\b/` matches (same conservative data-vs-command tradeoff already accepted for `pgrep` in docs residual 2). Add deny assertions for the `while`/`if` forms above and update docs/arm-pretool-check.md:157 (&#34;The raw check fires on ...&#34;) and residual 1 to match.
- ℹ️ `bin/fm-arm-command-policy.mjs:763` - The fix-round narrowing of the scope-flag test to `/^-[Pgs][0-9]*$/` now rejects the attached-variable spellings of a caller-scoped select: `pkill -P$$`, `pgrep -P$$ | xargs kill`, `kill $(pgrep -P$$)`, and `pkill -P$pid -f node` all return `deny broad-process-kill` (verified), whereas the pre-fix regex accepted them and `pkill -P $$` / `pkill -P &#34;$pid&#34;` are still allowed. The intent requires caller-scoped `-P/-g/-s` forms to stay allowed, and `-P$$` is that form with no space. No committed script in bin/ or tests/ uses the attached spelling, so impact is low, but it is a false positive on the exact idiom the deny text recommends. Cheap fix that still excludes signal names: `/^-[Pgs](?:[0-9]+|\$\S*)?$/` here and `-[Pgs](?:[0-9]+|\$\S*)?` in `rawArgsSelectByCallerScope` (line 71), plus an allow assertion for `pkill -P$$`.
- ℹ️ `bin/fm-arm-pretool-check.sh:165` - Adding the bare `*ps*` glob to the fast-path trigger set means any command containing the two bytes `ps` (`https://`, `apps/`, `steps`, `deps`, `psql`, `.ps1`, ...) now spawns the Node classifier (~70 ms per Bash tool call measured here) instead of fast-allowing. The prefilter remains a strict superset, so this is correctness-neutral and was explicitly prescribed in the fix instructions; noting only that the fast path now covers a much smaller share of ordinary commands. Tightening `ps` to a word-bounded match is not a drop-in change because the prefilter deletes newlines rather than replacing them with a separator, so leave as-is unless hook latency becomes noticeable.

🔧 Fix: deny fuser -k, loop-body discovery kills; allow -P$$ scope
- ⚠️ `bin/fm-arm-command-policy.mjs:972` - The pipe scan reads `nodeInfos[i].unscopedDiscovery` / `.pgrepWatcher` from `isUnscopedDiscovery(position)` / `isWatcherPgrep(position, ...)`, which look only at the node&#39;s direct command, while the same loop iteration already computed `nodeUnscopedDiscovery` / `nodePgrepWatcher` from the node&#39;s subshell, brace-group, and substitution payloads and then discards them for the pipe scan. So a discovery that is wrapped or embedded in the producer node is allowed into `xargs kill` (verified at HEAD): `(pgrep -f node) | xargs kill`, `{ pgrep -f a; pgrep -f b; } | xargs kill`, `echo $(pgrep -f node) | xargs kill`, `printf &#39;%s\n&#39; $(pgrep -f node) | xargs kill`, and the watcher variant `(pgrep -f fm-watch) | xargs kill` (allowed, whereas `pgrep -f fm-watch | xargs kill` is `broad-watcher-kill`). The substitution form `kill $( (pgrep -f node) )` is correctly denied, so this is an inconsistency in the new pipe-scan code, not a documented residual (the docs residual 1 covers files/arrays/intermediate non-discovery commands, not a subshell or group around the discovery itself). Fix within the existing mechanism: populate the nodeInfos entry with `unscopedDiscovery: nodeUnscopedDiscovery || isUnscopedDiscovery(position)` and `pgrepWatcher: nodePgrepWatcher || isWatcherPgrep(position, nodeContext)`; add deny assertions for `{ pgrep -f a; pgrep -f b; } | xargs kill` (broad-process-kill) and `(pgrep -f fm-watch) | xargs kill` (broad-watcher-kill).
- ℹ️ `bin/fm-arm-command-policy.mjs:824` - `nodeIsXargsKill` tests only the first non-option word after `xargs`, so a kill utility reached through a wrapper or nested shell is allowed (verified at HEAD): `pgrep -f node | xargs -n1 sudo kill`, `... | xargs -n1 env kill`, `... | xargs -n1 command kill`, `... | xargs -I{} sh -c &#34;kill {}&#34;`, `... | xargs sh -c &#39;kill $0&#39;`, the GNU separated long option `... | xargs --max-args 1 kill`, and the consumer-side group `pgrep -f node | (xargs kill)`. Separately, `xargs pkill`/`xargs killall` is only treated as a kill when a discovery feeds it, yet `pkill -f`/`killall` by name is broad regardless of feed: `echo node | xargs pkill -f` and `cat names | xargs -n1 pkill -f` are allowed while `pkill -f node` is denied. None of these is in the docs residual list. Decide explicitly: either extend `nodeIsXargsKill` to skip the classifier&#39;s known wrapper names (`sudo`, `env`, `command`, `nohup`, `exec`, `timeout`) and treat `--&lt;long&gt; &lt;value&gt;` like the short value options, and flag an `xargs pkill`/`xargs killall` utility that lacks a caller-scope flag as broad on its own; or add these to the residuals in docs/arm-pretool-check.md so the coverage statement stays exact.
- ℹ️ `bin/fm-arm-command-policy.mjs:951` - `kill -9 -1`, `kill -- -1`, and `kill -s TERM -1` are allowed (verified at HEAD). pid -1 signals every process the caller may signal, which is the broadest possible cross-lane kill, but it is not a match-by-name/command-line so it falls outside the rule as literally defined in the intent. It is cheap to detect by shape (a `-1` argument after a signal spec or after `--`, which is also where the allowed `kill -- -$pgid` form lives) if you want the class-closing goal to cover it; otherwise it belongs in the residual list so readers do not assume the seatbelt stops it.
- ℹ️ `bin/fm-arm-command-policy.mjs:949` - Non-kill invocations of the kill tools are denied because the pkill/killall check needs no argument evidence: `command -v pkill &gt;/dev/null 2&gt;&amp;1`, `command -v killall`, `pkill --help`, `pkill -V`, and `killall -l` all return `deny broad-process-kill` (verified at HEAD; `command -v` is consumed as a no-argument wrapper option at line 502, mirroring the pre-existing `command -v bin/fm-watch-arm.sh` -&gt; `watcher-nested` behavior). The repo&#39;s own scripts only probe `command -v pgrep`/`command -v lsof` (both allowed), so impact is low, but `command -v pkill` is a standard portability probe an agent may type. If false positives matter, treat `command -v/-V` as a query (do not unwrap to the target) and exempt `--help`/`-V`/`--version`/`-l` for pkill/killall; otherwise leave as-is (fail-closed is the established design).

🔧 Fix: deny kill -1, nested producers, xargs pkill; allow query forms
- ⚠️ `bin/fm-arm-command-policy.mjs:63` - The round-4 `kill -1` (kill-all) deny lives only in the AST path (`isKillAll`); `rawMentionsGeneralBroadKill` was not extended, so any loop/`if`/`case` prefix bypasses it. Verified at HEAD: `while true; do kill -9 -1; done`, `if true; then kill -- -1; fi`, and `case x in x) kill -9 -1 ;; esac` all return `allow`, while the same bodies with `pkill -f node` or `lsof -ti :3000 | xargs kill` are denied. This is the exact inconsistency class the round-3 fix (`raw-fallback-misses-extended-discovery`) closed for the discovery feeds, and pid -1 is the broadest kill the rule covers. docs/arm-pretool-check.md:148 states the kill-all form is denied with no unsupported-grammar caveat, and the raw-check sentence at docs line 160 omits it. Fix within the existing raw fallback: add a target-position `-1` pattern that requires a preceding signal spec or `--` so `kill -1 1234` stays SIGHUP-to-one-pid, e.g. `if (/\bkill\s+(?:-[A-Za-z0-9]+|-[sn]\s+\S+|--)\s+(?:\S+\s+)*-1(?=[\s;|&amp;)]|$)/.test(normalized)) return true;` (validated: matches the three loop forms above and `kill -TERM -- -1`, does not match `kill -1 1234`, `kill -- -$pgid`, `kill -- -12345`, `kill 1234`). Add deny assertions for `while true; do kill -9 -1; done` and `if true; then kill -- -1; fi`, an allow assertion for `for x in 1; do kill -1 1234; done`, and name kill -1 in the docs raw-check sentence.
- ℹ️ `bin/fm-arm-command-policy.mjs:807` - `XARGS_VALUE_LONG_OPTIONS` includes `eof` and `max-lines`, but in GNU xargs those long options take an OPTIONAL argument, which must be attached with `=`; without `=` they consume nothing and the next word is the utility. The set therefore swallows the kill: `pgrep -f node | xargs --max-lines kill` and `pgrep -f node | xargs --eof kill` return `allow` (verified), while `--max-lines=1 kill` and `-l kill` are denied. Obscure spellings, but the fix is a one-line removal of `eof` and `max-lines` from the set (the required-argument short forms `-E`/`-L` are already handled by `XARGS_VALUE_OPTIONS`); add a deny assertion for `pgrep -f node | xargs --max-lines kill`.
- ℹ️ `bin/fm-arm-command-policy.mjs:845` - The `kill -0` liveness-probe exemption the user chose in round 1 (`isKillProbe`) applies only to a direct `kill`; the xargs utility path returns &#34;fed&#34; for any `kill`, so `pgrep -f node | xargs kill -0` and `pgrep -f node | xargs -n1 kill -0` are denied (verified) while `kill -0 $(pgrep -f node)` is allowed. Both send no signal. For consistency, return &#34;&#34; when the utility is `kill` and `isKillProbe(utility.words.slice(utility.index + 1))` holds, and add an allow assertion for `pgrep -f node | xargs kill -0`; or note the asymmetry next to the `kill -0` sentence in the docs.
- ℹ️ `docs/arm-pretool-check.md:157` - The docs state unconditionally that query forms are allowed (`command -v pkill`, ...), but the raw fallback still fires on the bare `pkill`/`killall` token, so the canonical portable-probe idiom `if command -v pkill &gt;/dev/null 2&gt;&amp;1; then echo y; fi` returns `deny broad-process-kill` (verified), whereas the supported-grammar `command -v pkill &gt;/dev/null 2&gt;&amp;1 &amp;&amp; echo y` is allowed. Residual 3 covers &#39;a bare token used as pure data&#39;, which a reader would not naturally map to `command -v`. Decide explicitly: either (a) in `rawMentionsGeneralBroadKill` blank `command\s+-[A-Za-z]*[vV]\S*\s+\S+` spans (and the `--help`/`-V`/`-l`-only invocations) before the pkill/killall match, with a deny-stays assertion for `if true; then pkill -f node; fi` and an allow assertion for the `if command -v pkill ...` idiom; or (b) extend residual 3 and the line-157 sentence to say query/help forms inside unsupported loop/`if`/`case` grammar are conservatively denied.

🔧 Fix: close loop-grammar kill -1 gap; allow query and probe forms
- ⚠️ `bin/fm-arm-command-policy.mjs:80` - The round-2 raw-fallback scope exemption (`rawArgsSelectByCallerScope`) tests the whole pkill/pgrep argument span, including the bytes inside a quoted `-f` pattern, so any command-line pattern that happens to contain ` -P &lt;n&gt;`, ` -g &lt;n&gt;`, or ` -s &lt;n&gt;` is read as caller-scoped and the broad kill is allowed in loop/`if`/`case` grammar. Verified at HEAD through both the classifier and the transport wrapper (rc=0): `while true; do pkill -f &#34;node server.js -P 3000&#34;; done` and `while true; do pkill -f &#39;gunicorn -g 2&#39;; done` return `allow`, while the same commands without the loop prefix return `deny broad-process-kill`. Port/group flags inside a matched command line are plausible (`-P 3000`, `-g 2`), so this is a real inconsistency between the AST path and the raw fallback the fix rounds introduced, not a documented residual (docs line 164 states the raw check fires on a pkill &#39;whose own argument span carries no caller-scope flag&#39;). Fix within the existing raw mechanism: blank quoted spans before testing, e.g. `return /.../.test(args.replace(/&#34;[^&#34;]*&#34;|&#39;[^&#39;]*&#39;/g, &#34; &#34;))` in `rawArgsSelectByCallerScope` (a quoted scope value like `pkill -P &#34;$pid&#34;` still matches because `-P` is then followed by whitespace); add a deny assertion for `while true; do pkill -f &#34;node server.js -P 3000&#34;; done` and keep the allow assertion for `for p in $(pgrep -P &#34;$$&#34;); do kill $p; done`.
- ℹ️ `bin/fm-arm-command-policy.mjs:798` - `lsof` counts as discovery only with a pid-listing flag (`-t`/`-Fp`), as prescribed in round 1, so the awk-extracted table form is allowed while the equivalent `ps` form is denied (verified at HEAD): `lsof -i :3000 | awk &#39;NR&gt;1{print $2}&#39; | xargs kill -9` and `kill -9 $(lsof -i :3000 | awk &#39;NR&gt;1{print $2}&#39;)` return `allow`, whereas `ps aux | grep X | awk &#39;{print $2}&#39; | xargs kill` and `lsof -ti :3000 | xargs kill -9` are denied. &#39;Free the busy port&#39; is a common agent action and a port is exactly the kind of shared resource a sibling lane&#39;s dev server holds, so this shape reaches the class the change claims to close. Since only kill-fed forms are denied (read-only `lsof -i :3000` stays allowed either way), the simplest consistent rule is to treat any `lsof` as discovery exactly as `ps` is (drop the flag condition at this line and the `-t` qualifier in the raw fallback at line 77), or alternatively require `-i`/`-p`-style selection; otherwise add the lsof-table+awk form to the docs residual list so the coverage statement stays exact. Deciding this changes a spec the author set explicitly, hence ask-user.
- ℹ️ `bin/fm-arm-command-policy.mjs:797` - `ps` is discovery on any invocation (as prescribed), so the `ps` spellings of the recommended caller-scoped cleanup are denied as broad (verified at HEAD): `ps -o pid= --ppid $$ | xargs kill`, `kill $(ps -o pid= --ppid $$)`, and `kill $(ps -o pid= -s $$)` all return `deny broad-process-kill`, while the semantically identical `pgrep -P $$ | xargs kill` is allowed. This is a false positive on a caller-scoped form of the same kind the author chose to fix in rounds 2-4 (`-P$$`, `kill -0`, query forms). Impact is low because `pgrep -P`/`pkill -P` are the forms the deny text steers to and no committed script uses `ps --ppid` to kill, and scoping `ps` correctly is platform-sensitive (procps `--ppid`/`--sid`/`-s` vs BSD `-g`), so this may reasonably stay as-is; if accepted, add a sentence to the &#39;General broad process kill&#39; section stating that `ps` is always treated as unscoped even with `--ppid`/`-s`, so a reader can tell without running it.

🔧 Fix: ignore quoted scope text; treat any lsof as discovery
5 issues (2 warnings, 3 infos) still open:
- ⚠️ `bin/fm-arm-command-policy.mjs:72` - The round-5 raw kill-all pattern `/\bkill\s+(?:-[A-Za-z0-9]+|-[sn]\s+\S+|--)\s+(?:\S+\s+)*-1(?=[\s;|&amp;)`]|$)/` lets `(?:\S+\s+)*` run across `;`, `|`, `&amp;`, and `)`, so in unsupported loop/if grammar a `kill -0`/`kill -TERM` of a specific pid followed anywhere later by a `-1` argument to an unrelated command is denied. Verified at HEAD: `while kill -0 &#34;$pid&#34; 2&gt;/dev/null; do tail -1 &#34;$log&#34;; sleep 1; done`, `until ! kill -0 $pid; do sleep 1; done; head -1 out.log`, `if kill -TERM $pid; then git log -1; fi`, and `for x in 1; do kill -9 $pid; sleep -1; done` all return `deny broad-process-kill`, while the supported-grammar `kill -0 &#34;$pid&#34; &amp;&amp; tail -1 &#34;$log&#34;` is allowed. The wait-for-exit loop with a `kill -0` probe plus `tail -1`/`head -1` is a common agent idiom, and docs line 166 says the raw check fires on a kill &#39;whose target after a signal spec or `--` is the literal -1&#39;, which is not what happens. Fix within the existing raw mechanism by bounding the span at separators, mirroring the pkill/pgrep spans: `/\bkill\s+(?:-[A-Za-z0-9]+|-[sn]\s+[^\s;|&amp;)`]+|--)\s+(?:[^\s;|&amp;)`]+\s+)*-1(?=[\s;|&amp;)`]|$)/` (validated: still matches `while true; do kill -9 -1; done`, `if true; then kill -- -1; fi`, `case x in x) kill -9 -1 ;; esac`, `kill -TERM -- -1`, `kill -s TERM -1`, `kill -9 1234 -1`; no longer matches the four false positives above, nor `kill -1 1234`, `kill -- -$pgid`, `kill -- -12345`, `kill 1234`). Add allow assertions for `while kill -0 &#34;$pid&#34; 2&gt;/dev/null; do tail -1 &#34;$log&#34;; sleep 1; done` and `if kill -TERM $pid; then git log -1; fi` next to `bpk-allow-loop-kill-sighup-pid`.
- ⚠️ `bin/fm-arm-command-policy.mjs:908` - A leading `!` (pipeline negation) is treated as the command word, so the real command is never analysed, and `!` is not in the unsupported-keyword list, so the raw fallback never runs either. Verified at HEAD: `! pkill -f node` and `! kill -9 -1` return `allow`, while `pkill -f node || true` and `! pgrep -f node || pkill -f node` are denied. This is a pre-existing parser gap (`! pkill -f fm-watch` and `! bin/fm-watch-arm.sh` are also allowed today) that the new broad-process-kill rule inherits; it is a one-token executed-kill shape the intent claims the seatbelt closes deterministically, and it is not in the docs residual list. Smallest consistent fix: add `&#34;!&#34;` to the unsupported-grammar keyword list on this line so the existing raw fallbacks decide. Verified on a scratch copy: `! pkill -f node` and `! kill -9 -1` -&gt; `broad-process-kill`, `! pkill -f fm-watch` -&gt; `broad-watcher-kill`, `! bin/fm-watch-arm.sh` -&gt; `unclassifiable-protected-command` (fail-closed, consistent with the existing unsupported-grammar rule), and `! pgrep -f &#34;lavish-axi poll&#34;`, `! kill -0 $pid`, `! pkill -P $$`, `! grep -q foo x &amp;&amp; echo y`, `! test -f x` all stay `allow`. This list is private to analyzeProgram, so the exported parser used by the cd-guard is unaffected. Add deny assertions for `! pkill -f node` and `! kill -9 -1` and an allow assertion for `! pgrep -f &#34;lavish-axi poll&#34;`; mention `!` alongside `time` in the docs&#39; unsupported-grammar sentence.
- ℹ️ `docs/arm-pretool-check.md:176` - Round 6 made any `lsof` invocation a discovery feed (bin/fm-arm-command-policy.mjs isUnscopedDiscovery and the raw fallback both dropped the `-t` qualifier, and the section bullet at line 151 says so), but residual 3 on this line still reads &#39;a `pgrep`/`ps`/`lsof -t`/`pidof`/`fuser` alongside a `kill`&#39; and the `broad-process-kill` reason-code row at line 193 still lists the feeds as &#39;(`pgrep`, `ps`, `lsof -t`, `pidof`, `fuser`)&#39;. Both understate coverage: `if true; then lsof -i :3000 | awk &#39;NR&gt;1{print $2}&#39; | xargs kill -9; fi` is denied (asserted by `bpk-if-lsof-table-awk`). Replace `lsof -t` with `lsof` in both places so the coverage statement stays exact, as the round-6 instructions required.
- ℹ️ `bin/fm-arm-command-policy.mjs:858` - `xargsKillKind` returns &#34;fed&#34; for any `kill` utility and never consults `isKillAll` on the utility&#39;s own arguments, so a kill-all reached through xargs is allowed when nothing discovery-shaped feeds it: `echo x | xargs kill -- -1` and `echo x | xargs kill -9 -1` return `allow` (verified at HEAD) although they run `kill -- -1 x` and signal every process, while the direct `kill -- -1` is denied. Not an accident an agent is likely to make, but the fix is one clause in the existing helper: `if (name === &#34;kill&#34;) { const killArgs = utility.words.slice(utility.index + 1); return isKillProbe(killArgs) ? &#34;&#34; : isKillAll(killArgs) ? &#34;broad&#34; : &#34;fed&#34;; }`, plus a deny assertion for `echo x | xargs kill -- -1`.
- ℹ️ `bin/fm-arm-command-policy.mjs:78` - Making every `lsof` a discovery in the raw fallback (round 6) widens the byte-level co-occurrence false positive in unsupported grammar to a plausible test-script shape: `while ! lsof -i :3000 &gt;/dev/null; do sleep 1; done; kill $SERVER_PID` (wait for a port, then kill the caller&#39;s own server pid) now returns `deny broad-process-kill` (verified), whereas the same loop with `until curl -sf localhost:3000` is allowed and the `lsof` loop alone is allowed. This is the tradeoff the user chose in round 6 and it is covered by docs residual 2 (&#39;a discovery command and a kill/xargs anywhere in the same command are denied together&#39;); noting it only so the cost of the lsof-any decision is visible. No change requested; if it proves noisy, the raw `ps|lsof|pidof|fuser` branch could require the discovery and the kill/xargs to share a `;`/`|`-bounded span, but that is a separate design call.

🔧 Fix: unwrap ! and builtin, bound raw kill-all, honest coverage docs
- 🚨 `bin/fm-arm-command-policy.mjs:591` - The new `builtin` wrapper unwrap in the shared `commandPosition` matches on `basename(command.value) === &#34;builtin&#34;`, so a path-qualified `./builtin` or `/tmp/builtin` (an external program, not the shell builtin) is unwrapped too. `bin/fm-cd-command-policy.mjs` imports `commandPosition` and previously unwrapped `builtin` itself by exact value (`command.value === &#34;builtin&#34;`), so this flips its verdicts: verified at HEAD `node bin/fm-cd-command-policy.mjs --command &#39;./builtin cd projects/foo&#39;` and `&#39;/tmp/builtin cd projects/foo&#39;` now return `deny persistent-cd`, while at the previous commit 028bd93 both return `allow`. Both are existing committed assertions in tests/fm-cd-pretool-check.test.sh (`A28 allow &#39;./builtin cd projects/foo&#39;`, `A32 allow &#39;/tmp/builtin cd projects/foo&#39;`), so the cd-guard suite fails at HEAD, and the cd guard&#39;s stated threat model is zero false blocks. Fix in `commandPosition`: test the exact word (`command.value === &#34;builtin&#34;`) rather than the basename, mirroring the cd guard&#39;s own check; for the arm policy this is also the correct semantics (`./builtin pkill` runs an external `builtin`). Note also that `prefixAssignments` is now a count relative to `firstCommandWord` while the cd guard still uses it as a slice start index (`words.slice(position.prefixAssignments, position.index)`); today that only widens the slice by the leading `!` words and changes no verdict, but keep the two in sync if it is touched again.
- ⚠️ `bin/fm-arm-command-policy.mjs:923` - The unsupported-grammar keyword check reads `position.words[0]`, which after this round is the `!` word when a negation is present, so `time` (and any other listed keyword) behind a `!` is not seen as unsupported and the raw fallback never runs; `commandPosition` then lands on `time`, which is not a wrapper, and the real kill is never analysed. Verified at HEAD: `! time pkill -f node` and `! time kill -9 -1` return `allow`, while `time pkill -f node`, `time ! pkill -f node`, and `! pkill -f node` all deny `broad-process-kill`. `! time cmd` is legal bash (checked: `bash -c &#39;! time echo hi&#39;` runs the command), and the docs added this round state that `!` is unwrapped and `time` is routed to the raw fallback, so this is a one-token bypass of a form the docs now claim covered. Fix: compute `firstName` from the first word that is not `!` (e.g. `position.words.find((w) =&gt; w.value !== &#34;!&#34;)`), or have `commandPosition` expose `firstCommandWord` and use it here; add a deny assertion for `! time pkill -f node` (must assert the exact `broad-process-kill` code so it fails when reverted).

🔧 Fix: match builtin exactly; see time behind a negation

</details>

<details>

✅ No issues found.
- <code>`bash tests/fm-arm-pretool-check.test.sh` — full targeted suite for the seatbelt (matrix A/R/E/K through codex/claude/grok/opencode/pi entries, direct policy contract, broad-process-kill contract, prefilter strict-superset, fail-open, output shape); 342 ok, 0 not ok. Note: this file&#39;s last test delegates to bin/fm-lint.sh as part of its own contract; I did not run any linter separately.</code>
- <code>Before/after verdict comparison: base-commit `bin/fm-arm-command-policy.mjs` (extracted via `git show 9074f9d:...`) vs target for 12 commands — incident shapes, killall/pkill-by-name, `pgrep -f fm-watch | xargs kill`, `kill -9 -1` flip allow→deny; scoped/literal/read-only forms stay allow on both</code>
- <code>Hook transcript through the Claude entry form (`printf &lt;json&gt; | bin/fm-arm-pretool-check.sh --claude`) capturing exit code and the hookSpecificOutput deny JSON / silent allow an agent actually receives</code>
- <code>Live process-table demo: spawned a sibling (new session, reparented to pid 1) and an own child, both with `lavish-axi poll` in argv; gated `pkill -f`, `kill $(pgrep -f)`, `pgrep -f | xargs kill` (all denied, nothing killed), then gated and executed `pkill -P $$` (allowed; only own child reaped, sibling survived), then `kill &lt;sibling pid&gt;` (allowed) for teardown</code>
- <code>Mutation check: removed the `deny(&#34;broad-process-kill&#34;)` line, re-ran the test file (fails at `K01 via codex must deny, got exit 0`), restored the file; `git status` clean afterwards</code>
- <code>Read-only sweep: `grep -rnE &#39;\b(pkill|killall)\b&#39; bin tests` minus scoped/-P forms and the guard&#39;s own files, and `pgrep -f` fed to kill — no unscoped pattern-kill found</code>

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>


christophmeise:
Moved to the maintainer's own fork at christophmeise/firstmate#1, unchanged - same branch, same head commit. Closing here to avoid two open pull requests for one change.

Issue #4268, opened 2026-09-12
Second-mate active-turn gate can never see a busy verdict on a tmux-backed mate: --secondmate spawns never arm the busy contract


tiago-peixoto:
Related to #4227, which covers a different blocker on the same gate.

## The gap

`secondmate_in_active_turn` in `bin/fm-watch.sh` exempts a second mate that is demonstrably inside an active turn from the wake-loop stall alarm.
Reaching that verdict requires `window_is_busy`, which reads the semantic busy contract owned by `bin/fm-busy-lib.sh`.

For a second mate, the parent home never holds that contract, so the gate cannot return a busy verdict and the stall alarm fires on the stall interval alone while the mate is mid-turn.

## Evidence, against `a27646c4`

`bin/fm-busy-event.sh` documents itself as the only writer of the semantic busy-state contract.
Its `arm` subcommand has exactly two callers in the repository, `bin/fm-spawn.sh:3300` and `bin/fm-spawn.sh:3308`.

Both sit inside the block that `bin/fm-spawn.sh:3279` opens as `if [ "$KIND" != secondmate ]; then` and line 3643 closes.
A `--secondmate` spawn therefore skips both, and the parent ends up with `state/<id>.meta` but no busy-state record for that mate.

With no record, `fm_busy_record_read` reports `missing`, and `fm_busy_classify_meta` returns `unknown missing`.
That is not a busy verdict, so `window_is_busy` is false and the active-turn gate returns "not in a turn".

The two fallbacks that can reach a busy verdict without the stored record are harness-specific: the rendered-tail readers at `bin/fm-busy-lib.sh:832` (grok) and `:841` (rovo).
The one backend that reports a native semantic state is herdr (`bin/fm-busy-lib.sh:920`, `busy herdr-native`).

So on current main the active-turn gate can only ever pass for a herdr-backed second mate, or a grok one.
For a tmux-backed `claude`, `pi`, `opencode`, or `omp` second mate it cannot pass at all, whatever the mate is actually doing.

### Reproducing the classification

With a `kind=secondmate` meta recording `harness=claude`, `backend=tmux`, and no busy-state record beside it, `fm_busy_classify_meta` returns `unknown missing` for both an empty and a non-empty captured tail.

## Scope

This is deliberately filed as a description of the gap rather than a proposed change.

Closing it means recording real turn evidence for a second mate in the parent home, which is new durable state and new wiring, not a restoration of existing behaviour.
[...]

PR #4270, opened 2026-09-12
fix(bin): stop aging a second mate's active turn from its launch

### LINKED ISSUE #4227: Second-mate wake-loop stall alarm fires for a busy mate launched more than FM_BUSY_TURN_MAX_SECS ago
## What happened

The parent's "secondmate wake-loop stalled" alert fires for a healthy second mate that is in the middle of a turn, whenever that mate was launched more than `FM_BUSY_TURN_MAX_SECS` ago and one of its turns runs past the stall interval.
We saw it twice on 2026-09-11 with `FM_BUSY_TURN_MAX_SECS=3600`, each time seconds after the mate finished a four-minute turn.
Both times the mate's queue drained on its own moments later.

This is the same symptom as #3829, on the same check, but a different cause.
#3829's diagnosis is undrained rows in the child's own wake queue.
This one is the active-turn exemption that #3943 added, which never applies to a mate that was launched long ago.

## Root cause (confirmed present on current `main`, 9074f9d)

`secondmate_in_active_turn` (`bin/fm-watch.sh:740`) exempts a mate that is inside a turn, but it asks `busy_turn_over_age` first:

```sh
secondmate_in_active_turn() {  # <task> <window>
  local task=$1 w=$2 tail40
  [ -n "$w" ] || return 1
  ! busy_turn_over_age "$task" || return 1
  tail40=$(fm_backend_capture "$(window_backend "$w")" "$w" 40 "$(window_label "$w")" 2>/dev/null) || return 1
  window_is_busy "$w" "$tail40"
}
```

`busy_turn_over_age` (`bin/fm-watch.sh:961`) ages the turn from the parent's `$STATE/<task>.turn-ended`, and falls back to `$STATE/<task>.meta` when that file is absent:

```sh
  f="$STATE/$task.turn-ended"
  [ -e "$f" ] || f="$STATE/$task.meta"
  ...
  [ "$(age_of "$f")" -ge "$BUSY_TURN_MAX_SECS" ]
```

A second mate's turns end in its own home, so the parent never holds a `turn-ended` file for it and always falls back to the meta, whose mtime is the mate's last launch.
Once the mate is older than `BUSY_TURN_MAX_SECS`, every check reads it as over age.
The busy verdict is then never consulted, and the alert fires on the stall interval alone.

## Reproduction

Copy `test_secondmate_active_turn_defers_stall_until_the_turn_ends` in `tests/fm-wake-queue.test.sh` and backdate the parent meta before the checkpoint (`touch -t 202001010000 "$STATE/mate.meta"`), which stands in for a mate launched long ago.
The original test raises no alert.
The backdated copy raises `check: secondmate wake-loop stalled` while the mate is still busy.

## Suggested fix

A fix with that reproduction as its regression test is https://github.com/tiago-peixoto/firstmate/pull/46:

- `secondmate_in_active_turn` bounds the busy exemption by how long the mate's queue drain position has been frozen, instead of by when the mate launched.
  A mate stuck busy forever still alerts after `FM_BUSY_TURN_MAX_SECS` with no progress, and an idle mate with unhandled rows still alerts after the stall interval, as it does today.
- `busy_turn_over_age` treats a second mate as always over age, because the parent holds no evidence that its turn is young.
  Its other callers in the main loop reach a mate only while it has a declared wait, and there a busy mate already moved to the declared-wait recheck cadence once its launch was an hour old; this makes that behaviour independent of when the mate launched.

Reading the mate's own busy evidence from its home was considered and rejected: a mate is a primary firstmate session, and a primary home records no harness-neutral per-turn marker the parent could read.

tiago-peixoto:
Fixes #4227.

## The defect

`secondmate_in_active_turn` in `bin/fm-watch.sh` is the gate that keeps the wake-loop stall alarm from firing at a second mate that is demonstrably inside an active turn. It called `busy_turn_over_age` first and returned "not in a turn" whenever that reported the bound crossed, so the busy pane was never consulted.

`busy_turn_over_age` ages from `state/<id>.turn-ended`, falling back to `state/<id>.meta`. A second mate's turns end in its own home, so the parent never receives a `.turn-ended` mark for it and the fallback ages the mate's *last launch*. Every mate launched more than `FM_BUSY_TURN_MAX_SECS` ago was therefore permanently over age, and any turn outlasting `FM_SECONDMATE_WAKE_STALL_SECS` raised a false stall alarm while the mate was working normally.

## Scope: this fixes one of two blockers

Stated up front because it bounds what this PR claims.

**Fixed here:** the gate no longer ages a second mate's turn from its spawn record.

**Still open:** a `--secondmate` spawn never arms the semantic busy contract in the parent home, so the gate cannot obtain a busy verdict for a tmux-backed mate even now. Filed separately as #4268.

The evidence for that second half, so it can be checked without re-deriving it:

- `bin/fm-busy-event.sh` documents itself as the contract's only writer. Its `arm` subcommand has exactly two callers, `bin/fm-spawn.sh:3300` and `:3308`.
- Both sit inside the block `bin/fm-spawn.sh:3279` opens as `if [ "$KIND" != secondmate ]; then` and line 3643 closes, so a `--secondmate` spawn skips both.
- With no record, `fm_busy_classify_meta` returns `unknown missing`, which is not a busy verdict.
- The only record-free busy paths are harness-specific: the rendered-tail readers at `bin/fm-busy-lib.sh:832` (grok) and `:841` (rovo), plus the native herdr verdict at `:920`.

Consequence: today the active-turn gate can pass only for a herdr-backed mate or a grok one. For a tmux-backed `claude`, `pi`, `opencode`, or `omp` mate it cannot pass at all.

Closing that half means recording real turn evidence for a second mate in the parent - new durable state and new wiring rather than a restoration of existing behaviour - so it is left to this project to scope in #4268 rather than folded in here.

## The fix

`secondmate_in_active_turn` now takes the window and the queue's no-progress interval, and bounds the busy exemption on that interval instead of on turn age. The interval is evidence the parent actually holds.

A busy mate keeps the exemption while its queue has been frozen for less than `FM_BUSY_TURN_MAX_SECS`; a mate stuck busy forever still alarms once that bound passes. The bound that stops a busy pane from proving liveness indefinitely is preserved, not removed.

`busy_turn_over_age` itself is unchanged; its remaining callers are the ordinary crew busy-pane path.

## Reproduction, against current `main`

Driven through `bin/fm-watch-checkpoint.sh` over an isolated home, a real local second-mate home with its own durable wake queue, and a real tmux server on a private socket, with the stall interval scaled to 2s so a 12s checkpoint covers the same decision the 180s default makes.

| Base | Fixture | Result |
| --- | --- | --- |
| `a27646c4` | busy mate, spawn record backdated | `check: secondmate wake-loop stalled: mate=mate row=7 idle=2s` - false alarm reproduced |
| `a27646c4` | same mate, spawn record left fresh | silent - isolates the cause to launch-record aging |
| this branch | busy mate, spawn record backdated | silent |
| this branch | still busy, queue frozen past `FM_BUSY_TURN_MAX_SECS=3` | alarms once at `idle=3s` - the bound still holds |

Unchanged behaviour also re-checked live: a frozen queue on a mate whose turn has ended still alarms; a busy record whose endpoint is gone still alarms, because absence of proof is not an active turn; a continuously draining foreign queue stays silent and is never rewritten by the parent.

## What the new test adds

`test_secondmate_long_lived_mate_mid_turn_is_not_a_stall` in `tests/fm-wake-queue.test.sh` pins a mate whose launch record predates `FM_BUSY_TURN_MAX_SECS` and which is demonstrably mid-turn, plus the same mate alarming once when its queue freezes past the bound.

The existing `test_secondmate_active_turn_defers_stall_until_the_turn_ends` uses a freshly written meta and passes either way, so it could not have caught this.

The new case's own comment block records what it does not cover: its fixture arms the busy contract by hand, which a real `--secondmate` spawn does not do, so it pins the launch-aging half only and the tmux gap tracked in #4268 remains reachable.

## Test-timing note

Reaching the alert now costs a pane capture inside the gate. Three checkpoints in `tests/fm-wake-queue.test.sh` that assert an alert move from a 1s to a 4s bound, the value the neighbouring active-turn cases already use. The bound is a ceiling, not a wait - the checkpoint returns on the first actionable wake - and a 1s bound was observed missing the alert repeatedly on a loaded machine where 4s did not miss in 20 runs.

## Pipeline


kunchenguid:
Speaking as Kun's firstmate: first look on PR #4270 (tiago-peixoto). **Draft** — not merge-eligible until marked ready for review.

**Head** `ba63b68c47189ce4f4355ad8bd33336464fd73c3` (`fm/upstream-4227-secondmate-stall-alarm`) vs main `a27646c4eae5`. MERGEABLE / UNSTABLE. Author not blocked. Surface: `secondmate_in_active_turn` in `bin/fm-watch.sh` no longer ages the busy exemption from parent launch meta via `busy_turn_over_age`; bounds it on queue idle vs `BUSY_TURN_MAX_SECS` instead; docs + regression in `tests/fm-wake-queue.test.sh`. **workflow-zero**. +99/−18, 4 files. Targets #4227 (ready-for-pr restore). Explicitly leaves #4268 (missing busy-contract arm on `--secondmate` / tmux) out of scope — correct; do **not** close #4268 with this PR. Please add `Fixes #4227` in the body when you mark it ready so merge closes the issue.

**Attestation: MATCH** (`<!-- no-mistakes-pipeline-attestation:v1` binds tip `ba63b68c…`).

**Contract-class: restore.** Watcher already promised a mate demonstrably inside an active turn does not escalate on the stall interval; aging that exemption from launch meta broke the promise for long-lived mates (same diagnosis as issue #4227 triage). Re-anchoring the bound on queue no-progress idle restores that promise without removing the forever-busy backstop. Honest partial scope (hand-armed busy contract in the regression; real tmux secondmate still blocked by #4268) keeps this restore-class rather than a new durable-state design.

**VISION.md per-rule**
- **One captain, one interface** — aligns (false stall wakes are noise that hide real ones).
- **Authority is explicit and never inferred** — aligns (no new autonomy; supervision gate only).
- **Scripts own the mechanics, agents own the judgment** — aligns (stall/busy gates stay scripted).
- **A restart is a non-event** — aligns (long-lived mates must keep correct supervision after launch age).
- **Delegation with a spine** — aligns (secondmate supervision continuity).
- **The fleet outlives any vendor** — aligns (backend-capture busy check, not vendor UI).
- **Scope** — aligns (watcher continuity; #4268's new busy wiring correctly deferred).

**CI/NM:** Tip CI `34676400920` QUEUED; Require no-mistakes `34676400884` QUEUED (not first-time `action_required` — no approve needed this pass). Merge-eligible **N** while draft (and until tip CI/NM green). Waiting-author: mark ready for review (and prefer `Fixes #4227`). Firstmate flag: **no**.

**Security:** Diff review clean — no secrets, no auth, no network, no workflow changes. No security captain gate. No security FYI.

tiago-peixoto:
Checking in.

Re-verified live:
- Head is still `ba63b68c47189ce4f4355ad8bd33336464fd73c3`.
- GitHub reports MERGEABLE / CLEAN on that head.
- Branch untouched from our side since the reopen cycle that restored green CI.

If anything is still outstanding from your first look, say what you need and we will handle it. Otherwise this can wait on your queue.

kunchenguid:
Speaking as Kun's firstmate: this is merged. Thank you @tiago-peixoto — really appreciate you taking the time on this.

**Outcome:** squash-merged restore `ba63b68c…` → `a8dd08d29d03ff1c88cdf03a462e028824e42e03`. Closes #4227. Leaves #4268 open (honest partial scope — tmux secondmate busy-contract arm).

**Contract-class: restore.** Watcher already promised a mate demonstrably inside an active turn does not escalate on the stall interval; aging that exemption from launch meta broke it for long-lived mates. Re-anchoring the bound on queue no-progress idle restores that promise without removing the forever-busy backstop.

**VISION.md per-rule**
- **One captain, one interface** — aligns (false stall wakes are noise that hide real ones).
- **Authority is explicit and never inferred** — aligns (no new autonomy; supervision gate only).
- **Scripts own the mechanics, agents own the judgment** — aligns (stall/busy gates stay scripted).
- **A restart is a non-event** — aligns (long-lived mates keep correct supervision after launch age).
- **Delegation with a spine** — aligns (secondmate supervision continuity).
- **The fleet outlives any vendor** — aligns (backend-capture busy check, not vendor UI).
- **Scope** — aligns (watcher continuity; #4268's new busy wiring correctly deferred).

**CI/NM:** tip CI `34710349431` SUCCESS; NM `34710349433` SUCCESS; attestation MATCH; MERGEABLE/CLEAN. Security: clean (no secrets/auth/network/workflows). Firstmate flag: **no**.

PR #4305, opened 2026-09-12
fix: wrap the away-daemon command in Git Bash for Windows herdr pane shells

mnsky-tyan:
## Problem

On Windows hosts, `bin/fm-afk-launch.sh start` built the away-daemon terminal command as POSIX (`exec env FM_HOME=... bash bin/fm-afk-start.sh`) and handed it to `herdr pane run`, which types it into the pane's shell. The herdr pane shell on Windows is PowerShell, which has no `exec`, so the daemon never started and the pane sat at a PowerShell prompt (observed live 2026-09-12).

The existing herdr e2e could not catch this: it only asserts launch topology (dedicated non-visible workspace, captain tab untouched, stop tears down by exact id), all of which held even while the pane shell rejected the command.

## Fix

- `bin/fm-afk-launch.sh` now routes the daemon command through a single adapter, `fm_afk_launch_pane_cmd`:
  - POSIX hosts: the bare `exec env ...` command, byte-identical to before.
  - Windows hosts (existing `fm_host_is_windows` detection): the command is wrapped in the Git Bash login invocation PowerShell executes, with PowerShell single-quote literals (`& 'C:\Program Files\Git\bin\bash.exe' -lc '<posix command>'`) so no `$`/backtick interpolation can corrupt the inner command; embedded quotes are doubled. Bash resolution: `FM_AFK_LAUNCH_GIT_BASH` override, then the standard Git for Windows install locations, then the running bash; loud failure (before any record is written) when none is found.
  - tmux path unchanged (tmux runs commands through sh, never a pane shell).
- `tests/fm-afk-launch.test.sh`:
  - Two portable units pin the `pane run` command on both host classes (forced via the memoized host-detection variable, so they hold on any CI host), including the quote-doubling contract.
  - The herdr e2e now proves actual execution: the daemon entry writes a marker file in its pane, and the test fails if that marker never appears, exactly the failure mode topology checks cannot see.

## Verification (live, Windows 11 Home 10.0.26200 x64, Git Bash, herdr protocol 19, zai primary)

Before (unfixed), a real `start` run:

```
fm-afk-launch: daemon did not become ready; closing herdr:default:w1X:p1   # start exits 1
```

and the daemon pane's own capture showed the exact failure:

```
> exec env FM_HOME=/c/Users/tyanw/work/harness/firstmate ... bin/fm-afk-start.sh
exec: The term 'exec' is not recognized as a name of a cmdlet, function, script file, or executable program.
```

After (fixed), `start` succeeds and the pane executes the wrapper:

```
fm-afk-launch: daemon launched in non-visible herdr workspace w1Z (pane default:w1Z:p1), supervising default:w12:pT   # start rc=0
```

```
> & 'C:\Program Files\Git\bin\bash.exe' -lc 'exec env FM_HOME=... bin/fm-afk-start.sh'
afk: starting supervise daemon in foreground; keep this command as a tracked background session
```

The daemon's own log and liveness beacon confirm it really ran:

```
[2026-09-13T01:20:23+0800] daemon starting (pid 119029); target=default:w12:pT; backend=herdr; afk=on; ...
state/.last-watcher-beat  # fresh mtime within one watcher poll
```

and `stop` tore everything down cleanly (`daemon shutting down` in the log, workspace closed by exact id, `.afk` cleared, record archived, `stop rc=0`).

## Checks

- `tests/fm-afk-launch.test.sh`: 61 ok, 0 failures (including the live herdr e2e execution marker on Windows).
- `bin/fm-lint.sh`: shellcheck 0.11.0 + actionlint 1.7.12 clean on the changed files (the two pre-existing SC2016 infos in `bin/fm-windows-process-lib.sh` are untouched by this branch and exist at HEAD).
- `bin/fm-doc-audience-check.sh`: ok (97 surfaces, 363 links).
- `docs/herdr-backend.md` away-mode section updated (one sentence); dated evidence recorded in `docs/verification/runtime-backends.md` under "Away-mode transport".

### DIFF
--- a/bin/backends/herdr.sh
+++ b/bin/backends/herdr.sh
@@ -818,6 +818,13 @@ fm_backend_herdr_canonical_socket_path() {  # <socket-path>
   [ -n "$socket" ] || return 1
   case "$socket" in
     /*) ;;
+    # herdr on Windows injects a drive-letter path (C:\...); convert it to the
+    # MSYS spelling so both sides of a socket-identity comparison canonicalize
+    # the same way (the herdr CLI's own session listing reports the same form).
+    [A-Za-z]:[\\/]*)
+      command -v cygpath >/dev/null 2>&1 || return 1
+      socket=$(cygpath -u "$socket" 2>/dev/null) || return 1
+      ;;
     *) return 1 ;;
   esac
   sock_dir=$(dirname "$socket")
@@ -2137,7 +2144,19 @@ fm_backend_herdr_create_task() {  # <container> <label> <cwd> <seeded_default_ta
 $dup_tabs
 EOF
   fi
-  out=$(fm_backend_herdr_cli "$session" tab create --workspace "$wsid" --cwd "$cwd" --label "$label" --no-focus 2>/dev/null) || return 1
+  # herdr worker panes inherit the SERVER's environment, which on Windows can
+  # predate tool installs the current session already relies on (verified
+  # live: a long-running server's PATH lacked the user's tool directories, so
+  # `treehouse get` failed as an unknown command inside the fresh pane).
+  # Pass this session's own PATH explicitly, already converted to Windows
+  # form (cygpath -wp) because relying on MSYS argument conversion inside a
+  # spawned script context proved unreliable (verified live: the same arg
+  # converted correctly from an interactive shell but arrived empty from
+  # fm-spawn's process tree).
+  worker_path=$PATH
+  command -v cygpath >/dev/null 2>&1 && worker_path=$(cygpath -wp "$PATH" 2>/dev/null) || worker_path=$PATH
+  out=$(fm_backend_herdr_cli "$session" tab create --workspace "$wsid" --cwd "$cwd" --label "$label" --no-focus --env "PATH=$worker_path" 2>/dev/null) || return 1
+  [ -z "${FM_DEBUG_TAB_ENV:-}" ] || { printf 'PATHLEN=%s\n' "${#PATH}" >> "$FM_DEBUG_TAB_ENV"; }
   tab_id=$(printf '%s' "$out" | jq -r '.result.tab.tab_id // empty' 2>/dev/null)
   pane_id=$(printf '%s' "$out" | jq -r '.result.root_pane.pane_id // empty' 2>/dev/null)
   if [ -z "$tab_id" ] || [ -z "$pane_id" ]; then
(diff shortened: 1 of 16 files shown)


mnsky-tyan:
## Windows STOP-path fix: live verification evidence

**Root cause (reproduced live on the target Windows 11 host, real daemon, isolated `fm-lab-*` herdr session, sandbox `FM_HOME`):**
- Bash runs a trapped TERM only after the foreground child it waits on completes; the daemon loop waits on multi-second children, so the daemon deterministically outlived the old stop's 10s SIGTERM grace window (pinned empirically: TERM into `bash -c 'trap ... ; powershell -Command "Start-Sleep 8"'` ran the trap only at the powershell's exit).
- The pane's PowerShell wrapper exits right after launching the daemon (CIM: the daemon winpid's parent pid is dead), and the watcher child carries a stale Win32 parent, so a `taskkill /T` walk alone cannot reach the bash-wrapped tree.

**Fix (60be73d):**
- Record the daemon's `/proc` descendant tree with process identities BEFORE signaling; graceful SIGTERM and its ordering are unchanged.
- After the grace window: `taskkill //PID <winpid> //T //F` (native tree) plus `kill -KILL` (MSYS fork children `/T` cannot see).
- Verify every recorded member exited (identity-matched; pid reuse can never redirect the sweep), fail loudly if a survivor outlives the force, then proceed with pane close, `.afk` clear, archive last. Hosts without `/proc` keep the old single-pid behavior.
- Also repairs the pane-command tests' dead `FM_UNAME_S_CACHE` forcing (`fm_host_is_windows` now reads the memo), which had made those units fail off their native host class.

**Live results (real `fm-afk-start.sh` daemon):**
- Cycle A: start ok -> supervision confirmed (lock held, watcher beat) -> stop hit the grace window, force path completed teardown, rc=0 -> daemon, watcher, transient, and winpid all gone (`/proc` + CIM), workspace closed by exact id, `.afk` + record cleared, posture archived.
- Cycle B (instrumented): same force path, rc=0, all tree members verified gone.
- Cycle C (start->stop->start): relaunch after both forced stops was clean (fresh workspace, supervision confirmed), then a graceful-path stop also verified fully clean.
- `tests/fm-afk-launch.test.sh`: 66 ok / 0 failures on Windows, including the herdr e2e and five new units (force-terminate, descendant sweep, unkillable-survivor surfacing, sweep identity guards).

Full evidence report: `data/firstmate-afk-stop/report.md` in the branch worktree. Neighboring suite failures (`fm-session-lock-ancestry`, `fm-watcher-lock`) reproduce identically at df93004 on this host and are pre-existing environment gaps.

PR #4307, opened 2026-09-12
feat(bin): detached RunPod pod watchdog that outlives the agent session

Alberto-Codes:
## Intent

A rented GPU can outlive the agent session that rented it, and keep billing.

WHY IT EXISTS. Pod deletion has until now run only through the RunPod
integration, which is reachable only from inside an agent session. With no
`runpodctl` on the machine and no credential in the environment, nothing
outside that session could terminate a pod. A calibration run on 2026-09-11
honoured its cost ceiling and its deadline only because the agent stayed
alive — the in-session watchdog it armed was not running once that session was
replaced. That is the gap, demonstrated rather than theorised.

The exposure is straightforward: an unattended H100 at published on-demand
rates accrues cost for as long as nobody notices, and "nobody notices" is
exactly the state an ended agent session produces.

This adds an out-of-band killswitch that survives the session that armed it,
so the meter can be stopped by something other than the agent that started it.
It reads its credential from the repository's gitignored environment file and
is a no-op when none is configured.

## What Changed

- Adds `bin/fm-runpod-watchdog.sh`, an `arm`/`status`/`disarm`/`run` killswitch that `arm` launches under `setsid` in its own session so it survives the agent that rented the pod. It terminates on exactly one condition — a deadline provably passed on the wall clock — enforcing whichever comes first of the deadline the run declared and a `--ceiling-usd`/`--rate-usd-hr` duration anchored to the pod's own start from `runtime { uptimeInSeconds }`; the anchor is persisted per pod at `state/<task>.runpod-watch.observed` so re-arming cannot extend the ceiling, and a carried anchor whose ceiling already fell due is refused as `stale` rather than acted on. Every uncertain state (unreadable record, credential, or clock; unreachable API; a pod id never sighted in the account; an unanchorable ceiling) alarms and keeps polling instead of terminating, and a `podTerminate` is only reported as stopped once the pod is absent from a subsequent account pod listing. `RUNPOD_API_KEY` is parsed out of the gitignored `.env` (never sourced) and handed to `curl` via a config file on stdin, so it stays out of argv and out of everything the script writes.
- Routes alarms to `state/<task>.status` under per-condition decision keys (`runpod-watch-<task>-<condition>`), rate-limited, resolving what it opens and reporting a verified stop as a `note:`; `pod-never-seen` is the one alarm retiring the watch does not close, and alarms carry the pod they are about so news of one pod never answers for another.
- Adds `tests/fm-runpod-watchdog.test.sh` covering both decisive cases plus the states between them (live deadline, unverified termination, unreadable pod list, never-seen pod id, pod-anchored and stale-anchored ceilings, `status` after re-arm, crewmate decisions left intact, corrupt pid file, credential never written, surviving the arming shell's death), with the RunPod API faked at the process boundary by a PATH `curl` shim so no test touches a real account. Documents the script in the new `docs/runpod-pod-watchdog.md` and registers it in `docs/scripts.md`, `docs/documentation-audiences.json`, and the `AGENTS.md` state-file list.

## Risk Assessment

⚠️ Medium: The change is a self-contained new script with no edits to existing behavior and a substantial behavior-level test suite, and it satisfies the intent's required capability (an out-of-band, session-surviving killswitch driven by the .env key); the two remaining issues are both judgment calls on the ceiling machinery that trade money against destroying a healthy run, neither blocks the deadline enforcement that is the intent's core, and both can be settled as follow-ups.

## Testing

Ran the targeted watchdog suite four times (23 behavior cases, green every run, no flakes) and the documentation-audiences test that validates the new operator page's registration, then built end-to-end evidence the suite cannot give: a four-scene CLI transcript driving the real script against a PATH curl shim standing in for RunPod's GraphQL endpoint, showing an agent session arming the killswitch and dying, the watchdog surviving in its own session id, the ceiling falling due unattended, podTerminate firing, absence confirmed by re-listing the account with the unrelated pod untouched, the stop reported on firstmate's wake channel, and the credential absent from every file the script wrote — plus the two refusal states (truncated deadline record, pod id never sighted) terminating nothing and alarming instead. Also verified separately that alarm repetition in those scenes came only from my rate-limit override, not the shipped 900s default. This change is CLI-only with no rendered surface, so the reviewer-visible artifact is the command transcript rather than a screenshot. No real RunPod account was contacted; all tests pass and the worktree is clean.

<details>
<summary>Evidence: End-to-end operator transcript: agent session dies, watchdog still stops the meter (4 scenes)</summary>

Source: [End-to-end operator transcript: agent session dies, watchdog still stops the meter (4 scenes)](https://github.com/Alberto-Codes/firstmate/blob/3a0f230c74b31b1126c28bffc2de7815a85aacbc/.no-mistakes/evidence/fm/fm-runpod-pod-watchdog/runpod-watchdog-e2e-transcript.txt)

```text
RunPod pod watchdog - end-to-end operator transcript
====================================================
bin/fm-runpod-watchdog.sh, branch fm/fm-runpod-pod-watchdog

WHAT THIS SHOWS
  The gap being closed: a rented GPU kept billing after the
  agent session that rented it was replaced, because pod deletion only ever ran
  from inside an agent session. These scenes drive the real executable end to
  end and show a pod being stopped with NO agent alive, plus the states where it
  must refuse to stop anything.

HOW THE ACCOUNT IS FAKED
  RunPod's GraphQL endpoint is replaced at the process boundary by a PATH `curl`
  shim, the same way tests/fm-runpod-watchdog.test.sh does it. Every byte below
  comes from the real script; no real RunPod account was contacted and nothing
  was rented or destroyed. Timings in scenes 2-4 are compressed (poll seconds,
  ceiling in cents) so a six-hour bound can be demonstrated in seconds.


== SCENE 1 - the C6 calibration figures: 20 USD ceiling at 3.49 USD/hr
# the pod has already been up 2h when the run gets around to arming
$ '~/.no-mistakes/worktrees/d58a82b70cc7/01M2B46ZPE0Q8JXMDWY96E6F87/bin/fm-runpod-watchdog.sh' arm --task c6 --pod pod-c6-h100 --deadline '2026-09-12T23:30:27Z' --ceiling-usd 20 --rate-usd-hr 3.49
armed: pod pod-c6-h100, declared deadline 2026-09-12T23:30:27Z, ceiling 20 USD at 3.49 USD/hr caps uptime at 20630s measured from the pod's own start, pid 2350081
# what the operator sees: the ceiling binds BEFORE the declared 6h deadline,
# and it is anchored to the pod's own start, not to arm time
$ '~/.no-mistakes/worktrees/d58a82b70cc7/01M2B46ZPE0Q8JXMDWY96E6F87/bin/fm-runpod-watchdog.sh' status --task c6
c6: pod=pod-c6-h100 deadline=2026-09-12T21:14:17Z source=ceiling anchor=pod-start declared-deadline=2026-09-12T23:30:27Z ceiling=20usd@3.49/hr=20630s-of-uptime watchdog=running pid=2350081
$ '~/.no-mistakes/worktrees/d58a82b70cc7/01M2B46ZPE0Q8JXMDWY96E6F87/bin/fm-runpod-watchdog.sh' disarm --task c6
disarmed: c6

== SCENE 2 - the gap this closes: the agent session dies, the meter does not
# The ceiling is compressed to seconds so the demo can run: 0.02 USD at
# 3.49 USD/hr = 20s of uptime, of which 10s is already burned. The declared
# deadline is an hour out, exactly as in the real 6h/20USD shape.
$ setsid bash '/tmp/fmwd-demo/agent-session.sh' '~/.no-mistakes/worktrees/d58a82b70cc7/01M2B46ZPE0Q8JXMDWY96E6F87/bin/fm-runpod-watchdog.sh' '/tmp/fmwd-demo/agent.pid'
  [agent session pid 2351295] renting the pod and arming the killswitch
armed: pod pod-c6-h100, declared deadline 2026-09-12T18:30:29Z, ceiling 0.02 USD at 3.49 USD/hr caps uptime at 20s measured from the pod's own start, pid 2351369
  [agent session pid 2351295] session replaced - exiting NOW

# the agent session is gone:
$ kill -0 2351295 2>/dev/null && echo 'agent session STILL ALIVE' || echo "agent session pid 2351295: dead"
agent session pid 2351295: dead
# the watchdog is not, and it is in its OWN session id, so a harness
# tearing down the agent's process group cannot take it:
$ ps -o pid=,sid=,cmd= -p 2351369
2351369 2351369 bash ~/.no-mistakes/worktrees/d58a82b70cc7/01M2B46ZPE0Q8JXMDWY96E6F87/bin/fm-runpod-watchdog.sh run --task c6
$ ps -o sid= -p $$ | tr -d ' ' | sed 's/^/this demo shell sid: /'
this demo shell sid: 2349983

# the pod is still billing, and the watchdog knows what it is enforcing:
$ cat /tmp/fmwd-demo/scene2/api/pods
pod-c6-h100
pod-unrelated
$ '~/.no-mistakes/worktrees/d58a82b70cc7/01M2B46ZPE0Q8JXMDWY96E6F87/bin/fm-runpod-watchdog.sh' status --task c6
c6: pod=pod-c6-h100 deadline=2026-09-12T17:30:39Z source=ceiling anchor=pod-start declared-deadline=2026-09-12T18:30:29Z ceiling=0.02usd@3.49/hr=20s-of-uptime watchdog=running pid=2351369

# waiting for the ceiling to fall due, with nobody attending...

# the account's pod list AFTER the deadline passed (the unrelated pod is untouched):
$ cat /tmp/fmwd-demo/scene2/api/pods
pod-unrelated

# what firstmate was told on the wake channel:
$ cat /tmp/fmwd-demo/scene2/state/c6.status
note: rented pod pod-c6-h100 passed the deletion deadline this run declared (in force: ceiling) and has been stopped; absence confirmed by listing the account's pods

# the watchdog's own trail:
$ cat /tmp/fmwd-demo/scene2/state/c6.runpod-watch.log
2026-09-12T17:30:29Z armed pod=pod-c6-h100 deadline=1789237829 ceiling_seconds=20
2026-09-12T17:30:29Z watchdog running pid=2351369
2026-09-12T17:30:29Z pod pod-c6-h100 started at 2026-09-12T17:30:19Z; ceiling anchored there
2026-09-12T17:30:39Z deadline passed; terminating pod=pod-c6-h100
2026-09-12T17:30:39Z podTerminate accepted pod=pod-c6-h100
2026-09-12T17:30:39Z verified absent from the pod list pod=pod-c6-h100
2026-09-12T17:30:39Z note rented pod pod-c6-h100 passed the deletion deadline this run declared (in force: ceiling) and has been stopped; absence confirmed by listing the account's pods
2026-09-12T17:30:39Z watchdog exiting after verified termination

# the API calls the fake endpoint actually received:
$ cat /tmp/fmwd-demo/scene2/api/calls.log
17:30:29Z pod list read
17:30:31Z pod list read
17:30:33Z pod list read
17:30:35Z pod list read
17:30:37Z pod list read
17:30:39Z pod list read
17:30:39Z podTerminate called
17:30:39Z pod list read

# the watchdog exited on its own once the stop was verified:
$ kill -0 2351369 2>/dev/null && echo 'STILL RUNNING' || echo "watchdog pid 2351369: exited"
watchdog pid 2351369: exited

# and the credential never reached anything it wrote:
$ grep -rl 'demo-key-never-logged' /tmp/fmwd-demo/scene2/state 2>/dev/null | sed 's/^/LEAKED: /' || true
$ grep -c 'demo-key-never-logged' /tmp/fmwd-demo/scene2/api/curl-stdin.log | sed 's/^/authorization headers the API received: /'
authorization headers the API received: 8


NOTE on scenes 3 and 4: these two runs set FM_RUNPOD_ALARM_REPEAT_SECONDS=0
to disable per-condition alarm rate limiting, so one alarm is shown once per
poll. At the shipped default (900s) the same 8-poll run appends exactly ONE
alarm line - verified separately.

== SCENE 3 - fail toward NOT killing: the deadline record is unreadable
# a truncated record: the pod is named, but the deadline is gone
$ cat /tmp/fmwd-demo/scene3/state/c6.runpod-watch
fm-runpod-watch-v1
pod=pod-c6-h100
$ timeout 6 '~/.no-mistakes/worktrees/d58a82b70cc7/01M2B46ZPE0Q8JXMDWY96E6F87/bin/fm-runpod-watchdog.sh' run --task c6 >/dev/null 2>&1; echo "loop exit: $? (124 = still watching when the demo cut it off)"
loop exit: 124 (124 = still watching when the demo cut it off)
# no pod was touched:
$ cat /tmp/fmwd-demo/scene3/api/pods
pod-c6-h100
$ grep -c podTerminate /tmp/fmwd-demo/scene3/api/calls.log || echo 'podTerminate calls: 0'
grep: /tmp/fmwd-demo/scene3/api/calls.log: No such file or directory
podTerminate calls: 0
# and the operator was told, on the channel that wakes firstmate:
$ cat /tmp/fmwd-demo/scene3/state/c6.status
blocked [key=runpod-watch-c6-record-unreadable]: the pod watchdog for c6 cannot read the run's deadline record, so it will NOT terminate anything; check the rented pod by hand
blocked [key=runpod-watch-c6-record-unreadable]: the pod watchdog for c6 cannot read the run's deadline record, so it will NOT terminate anything; check the rented pod by hand
blocked [key=runpod-watch-c6-record-unreadable]: the pod watchdog for c6 cannot read the run's deadline record, so it will NOT terminate anything; check the rented pod by hand
blocked [key=runpod-watch-c6-record-unreadable]: the pod watchdog for c6 cannot read the run's deadline record, so it will NOT terminate anything; check the rented pod by hand
blocked [key=runpod-watch-c6-record-unreadable]: the pod watchdog for c6 cannot read the run's deadline record, so it will NOT terminate anything; check the rented pod by hand
blocked [key=runpod-watch-c6-record-unreadable]: the pod watchdog for c6 cannot read the run's deadline record, so it will NOT terminate anything; check the rented pod by hand
$ '~/.no-mistakes/worktrees/d58a82b70cc7/01M2B46ZPE0Q8JXMDWY96E6F87/bin/fm-runpod-watchdog.sh' status --task c6
c6: RECORD UNREADABLE - this watchdog will alarm and will not terminate anything

== SCENE 4 - a pod id that was never in the account keeps alarming
# the deadline passed an hour ago, but this id has never been sighted
$ timeout 6 '~/.no-mistakes/worktrees/d58a82b70cc7/01M2B46ZPE0Q8JXMDWY96E6F87/bin/fm-runpod-watchdog.sh' run --task c6 >/dev/null 2>&1; echo "loop exit: $? (124 = still watching)"
loop exit: 124 (124 = still watching)
$ grep -c podTerminate /tmp/fmwd-demo/scene4/api/calls.log || echo 'podTerminate calls: 0'
0
podTerminate calls: 0
$ cat /tmp/fmwd-demo/scene4/state/c6.status
blocked [key=runpod-watch-c6-pod-never-seen]: the pod watchdog for c6 has never seen pod pod-typo in the account's pod list, so it is guarding nothing and will terminate nothing - the pod id may be wrong or stale while a rented pod keeps billing; check the account by hand
blocked [key=runpod-watch-c6-pod-never-seen]: the pod watchdog for c6 has never seen pod pod-typo in the account's pod list, so it is guarding nothing and will terminate nothing - the pod id may be wrong or stale while a rented pod keeps billing; check the account by hand
blocked [key=runpod-watch-c6-pod-never-seen]: the pod watchdog for c6 has never seen pod pod-typo in the account's pod list, so it is guarding nothing and will terminate nothing - the pod id may be wrong or stale while a rented pod keeps billing; check the account by hand
blocked [key=runpod-watch-c6-pod-never-seen]: the pod watchdog for c6 has never seen pod pod-typo in the account's pod list, so it is guarding nothing and will terminate nothing - the pod id may be wrong or stale while a rented pod keeps billing; check the account by hand
blocked [key=runpod-watch-c6-pod-never-seen]: the pod watchdog for c6 has never seen pod pod-typo in the account's pod list, so it is guarding nothing and will terminate nothing - the pod id may be wrong or stale while a rented pod keeps billing; check the account by hand
blocked [key=runpod-watch-c6-pod-never-seen]: the pod watchdog for c6 has never seen pod pod-typo in the account's pod list, so it is guarding nothing and will terminate nothing - the pod id may be wrong or stale while a rented pod keeps billing; check the account by hand
# and the warning stays OPEN as a decision even after the watch is retired:
$ '~/.no-mistakes/worktrees/d58a82b70cc7/01M2B46ZPE0Q8JXMDWY96E6F87/bin/fm-runpod-watchdog.sh' disarm --task c6
disarmed: c6
$ bash -c '. "$1"; status_open_decisions "$2"' _ '~/.no-mistakes/worktrees/d58a82b70cc7/01M2B46ZPE0Q8JXMDWY96E6F87/bin/fm-classify-lib.sh' '/tmp/fmwd-demo/scene4/state/c6.status'
runpod-watch-c6-pod-never-seen	blocked	the pod watchdog for c6 has never seen pod pod-typo in the account's pod list, so it is guarding nothing and will terminate nothing - the pod id may be wrong or stale while a rented pod keeps billing; check the account by hand
```
</details>
<details>
<summary>Evidence: Targeted suite output (23 behavior cases)</summary>

Source: [Targeted suite output (23 behavior cases)](https://github.com/Alberto-Codes/firstmate/blob/3a0f230c74b31b1126c28bffc2de7815a85aacbc/.no-mistakes/evidence/fm/fm-runpod-pod-watchdog/fm-runpod-watchdog.test.out)

```text
ok - a run whose deadline has passed is terminated, verified by listing, and leaves no open decision
ok - a run whose state cannot be read is NOT terminated, and firstmate is alarmed instead
ok - a run still inside its declared deadline is left alone
ok - a termination the pod list does not confirm is alarmed, never reported as stopped
ok - a failed pod-list read keeps the watch alive instead of concluding anything
ok - a pod that was seen and then disappears ends the watch cleanly
ok - a pod never sighted is never stopped, never reported stopped, and keeps alarming past its deadline
ok - disarming a watch that never sighted its pod leaves that warning standing
ok - a never-sighted warning about one pod is not closed by sighting another
ok - when one unseen pod turns up, the surviving decision names the one still missing
ok - watchdog alarms own their decision key instead of sharing the unkeyed default
ok - an alarm the watchdog raised is closed under its own key once the condition clears
ok - the credential is parsed out of the shared .env rather than executed with it
ok - a corrupt pid file cannot make the watchdog signal the process group that invoked it
ok - the credential reaches the API but never any file the watchdog writes
ok - arm refuses to guess a deadline and refuses one that has already passed
ok - a value-taking flag given as the last argument is refused instead of spun on
ok - the declared ceiling is anchored to the pod start, and re-arming cannot extend it
ok - a pod-start anchor survives a re-arm for the same pod and is not inherited by another
ok - a carried anchor whose ceiling already fell due is refused, alarmed, and terminates nothing
ok - status never reports an instant, source, or anchor from a previous arm
ok - a ceiling whose pod-start anchor cannot be read is not enforced, and status says so
ok - the armed watchdog survives the death of the shell that armed it, in its own session
```
</details>
<details>
<summary>Evidence: The moment the intent is satisfied: the pod is stopped with no agent alive</summary>

```text
[agent session pid 2351295] renting the pod and arming the killswitch
armed: pod pod-c6-h100, declared deadline 2026-09-12T18:30:29Z, ceiling 0.02 USD at 3.49 USD/hr caps uptime at 20s measured from the pod's own start, pid 2351369
[agent session pid 2351295] session replaced - exiting NOW

$ kill -0 2351295 ... -> agent session pid 2351295: dead
$ ps -o pid=,sid=,cmd= -p 2351369
2351369 2351369 bash bin/fm-runpod-watchdog.sh run --task c6 (own session id; demo shell sid: 2349983)

$ fm-runpod-watchdog.sh status --task c6
c6: pod=pod-c6-h100 deadline=2026-09-12T17:30:39Z source=ceiling anchor=pod-start declared-deadline=2026-09-12T18:30:29Z ceiling=0.02usd@3.49/hr=20s-of-uptime watchdog=running pid=2351369

... unattended ...

$ cat state/c6.status
note: rented pod pod-c6-h100 passed the deletion deadline this run declared (in force: ceiling) and has been stopped; absence confirmed by listing the account's pods

$ cat api/pods # the account afterwards - the unrelated pod is untouched
pod-unrelated

$ kill -0 2351369 ... -> watchdog pid 2351369: exited
```
</details>
- Outcome: ⚠️ 1 info across 1 run (12m7s)

## Pipeline

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>
<summary>⚠️ **Review** - 2 warnings</summary>

- ⚠️ `bin/fm-runpod-watchdog.sh:320` - Option parsing uses `shift 2 || true`, which spins forever when a value-taking flag is the last argument. Bash's `shift n` with n > $# leaves the positional parameters unchanged and returns 1, so the `|| true` swallows the failure and `$1` is still the same flag on the next iteration (verified: `set -- --task; while [ $# -gt 0 ]; do case $1 in --task) t=${2:-}; shift 2 || true;; esac; done` never terminates). Concretely, `fm-runpod-watchdog.sh arm --task t1 --pod pod-x --deadline` burns 100% CPU forever instead of printing a usage error; the same pattern is in cmd_disarm (:455), cmd_status (:470) and cmd_run (:548), and in `run` that is a detached, session-less process nothing will notice. The repo already carries the correct guard at bin/fm-afk-contract.sh:870 (`[ "$#" -gt 1 ] || return 2; x=$2; shift 2`); adopt it in all ten arms.
- ⚠️ `bin/fm-runpod-watchdog.sh:222` - `alarm()` appends an unkeyed `blocked: <text>` line to state/<task>.status, but bin/fm-classify-lib.sh:390 `_fm_decision_key` assigns the shared slug `default` to any status line with no `[key=<slug>]` token (none of the seven alarm texts contains a colon in its note, so `_fm_key_at_note_head` fails and `default` is returned), and `_fm_decision_fold_line` (:484) drops the existing `default` record before re-adding it. Failure sequence: the crew for task t1 appends `needs-decision: fp8 or int4 for the C6 sweep?`; one poll later the watchdog cannot reach RunPod and appends `blocked: the pod watchdog for t1 cannot reach RunPod ...`; `status_open_decisions` now folds to a single `default` entry carrying the watchdog note, so the crew's question silently disappears from the open-decision set firstmate surfaces. The reverse also holds: a later `resolved: fp8 chosen` closes the watchdog's still-true `terminate-unverified` alarm. This is exactly the key-takeover hazard documented at fm-classify-lib.sh:424-430, and bin/fm-pending-reply-lib.sh:1262 shows the convention (`blocked [key=<slug>]: ...`) this new independent writer skipped. Emit `blocked [key=runpod-watch-<task>]: ...` so the watchdog owns its own key.
- ⚠️ `bin/fm-runpod-watchdog.sh:356` - `ceiling_deadline=$((now + seconds))` anchors the derived spend ceiling at ARM time rather than pod start, and re-arming re-anchors it, so the effective budget silently exceeds the declared ceiling while `status` keeps reporting it verbatim. Concretely: pod starts at T0, `arm --ceiling-usd 20 --rate-usd-hr 3.49` runs at T0+2h, seconds = 20/3.49*3600 = 20630s, so the ceiling deadline is T0+7.73h = $26.97 of uptime against a declared $20 ceiling. Worse, docs/runpod-pod-watchdog.md:52 explicitly endorses re-arming to extend a deadline; re-arming at T0+3h with the same flags yields T0+8.73h = $30.5, and `fm-runpod-watchdog.sh status` still prints `ceiling=20`. The wrong label is reported without any error. Note this defect lives inside the `--ceiling-usd`/`--rate-usd-hr` component, which is a SECOND deadline source the stated intent does not require - the intent's demonstrated gap is that "the in-session deadline watchdog it armed was NOT running", i.e. deadline enforcement, and `--deadline` alone satisfies that. The smallest honest remedy is removing the ceiling-to-deadline conversion and its two flags, two record fields, and `status` columns. Keeping it requires authorizing a new input (pod start instant or uptime read from the API) to anchor the arithmetic - that extends the change rather than correcting it, which is why this is ask-user.
- ⚠️ `bin/fm-runpod-watchdog.sh:612` - The loop cannot distinguish "the run finished and cleaned up" from "this pod id was never in the account", and both end the watch silently. Sequence: an operator or agent arms with a stale or mistyped pod id (realistic when re-arming after a pod was recreated); `arm` validates only the character class (pod_id_valid, :94) and never asks the API whether the pod exists, even though it has already loaded the key and required curl/jq; it prints `armed: pod <id>, deadline ...`; the detached loop's first poll gets rc=1 from pod_present, logs "no longer listed before its deadline; nothing left to guard", deletes the record and exits 0 with NO alarm. `status` then prints "no armed pod watchdog", and the real rented H100 keeps billing entirely unattended - the exact failure the intent exists to close ("An unattended H100 at 3.49 per hour is roughly half that in eight hours"). The remedy is a product decision, not a mechanical repair: either `arm` verifies presence before publishing (which makes arming fail when the API is briefly unreachable, in tension with the script's own "fail toward not killing" posture at :39-44), or the loop must record first-sighting and alarm rather than exit when a pod was never observed. Both change arm's or the loop's contract, so this needs the author's sign-off.
- ⚠️ `bin/fm-runpod-watchdog.sh:325` - Simplification: the `--progress-file` / `--progress-grace` component is not required by the stated intent. The intent's required goal is an out-of-band way to "stop the meter" after the agent session is gone; this component by design never terminates anything (:626-639, docs/runpod-pod-watchdog.md:85-86), so it cannot contribute to that goal. It costs two CLI flags, two record fields and their validation (:363-367, :189), a record_read branch, a poll-loop branch, and two more alarm conditions. It also adds ongoing noise into the shared status channel: a progress file that does not yet exist, or that lives on the pod rather than on this box, produces a `progress-missing` alarm every ALARM_REPEAT_SECONDS (default 900s) for the run's whole life - 24 `blocked:` lines across a six-hour run, none of which is actionable. Recommended remedy is removing the component rather than hardening or documenting it.
- ⚠️ `bin/fm-runpod-watchdog.sh:294` - Simplification: the `probe` subcommand is not required by the stated intent. The intent states the write path is already established - "THE KEY EXISTS AND IS PROVEN ... a podTerminate mutation against a deliberately bogus pod id was rejected with 'pod not found to terminate' rather than a permission error, which proves write access is granted". `probe` re-performs exactly that one-off manual verification as a permanent subcommand, adding a fifth dispatch arm and its own request/response handling. No intent requirement needs it; recommended remedy is removing it.
- ℹ️ `bin/fm-runpod-watchdog.sh:590` - `[ "$now" -lt "$REC_DEADLINE" ] || deadline_reached=1` conflates "the test was false" with "the test could not be evaluated", and the error branch is the one that terminates a pod. `is_uint` (:101) is the only gate on REC_DEADLINE and accepts any digit string, but bash's `[` rejects a value outside intmax range: `[ 1789000000 -lt 99999999999999999999999 ]` prints "integer expected" and returns non-zero (verified), so such a record makes the loop terminate immediately - the opposite of the script's stated "FAIL TOWARD NOT KILLING" invariant at :39-44. Reachability is limited: `arm` is safe by accident (the same overflow makes the `[ "$declared" -gt "$now" ]` guard at :341 fail closed and refuse), so it takes a hand-edited record file to reach. Flagged at info because the guard is one line - bound the digit length in record_read alongside the existing is_uint check, or test the comparison's rc explicitly - and the failure direction is the money-destroying one.

🔧 Fix: anchor runpod ceiling to pod start; key alarms; drop probe/progress
5 issues (4 warnings, 1 info) still open:

- ⚠️ `bin/fm-runpod-watchdog.sh:307` - `load_key` reads the credential by sourcing `.env` in a subshell (`RUNPOD_API_KEY=$(set -a; . "$file" ...)`) instead of parsing it. Verified directly: with `FM_NOTE=don't` above `RUNPOD_API_KEY=secret123`, the sourced read yields the empty string; with the key line first it yields the key. `.env` is the home's shared multi-key operator file (fm-mail.sh keys, FMX tokens, and now RUNPOD_API_KEY), so an unrelated value containing an apostrophe or unquoted space is a realistic edit. `arm` dies loudly on this, but the loop re-reads `.env` every poll, so an edit AFTER arming turns a live killswitch into a repeating `key-unreadable` alarm that never terminates anything - the exact gap the tool exists to close, with a rented H100 still billing. Sourcing also executes arbitrary `.env` content every 60s. The repo already owns a non-executing reader for this exact format: `fmx_env_get` (bin/fm-x-lib.sh:63) and the inline parser in bin/fm-mail.sh:77-100. Remedy is mechanical - parse the one key line.
- ⚠️ `bin/fm-runpod-watchdog.sh:290` - The alarm rate-limit test `[ "$((now - last))" -lt "$ALARM_REPEAT_SECONDS" ]` reads a non-positive delta as "alarmed recently" and suppresses. Two concrete paths. (1) `now=$(now_epoch) || now=0` at :288: when the clock is unreadable the first `clock-unreadable` alarm emits and appends `clock-unreadable\	0`; every later poll computes `0 - 0 = 0`, sees `0 -lt 900`, and returns before touching the status channel - so the watchdog announces once that it will never terminate anything and then goes PERMANENTLY silent, contradicting this function's own stated contract at :275-278 ("one that alarms once would be missed if that wake was lost"). (2) A backwards clock step (NTP correction, suspend/resume) makes the delta negative for the size of the step, suppressing every condition for that long, including `terminate-unverified` - the one that says a pod may still be billing. Remedy is one added comparison: require `[ "$now" -ge "$last" ]` before applying the window so a non-positive delta falls through to emitting.
- ⚠️ `bin/fm-runpod-watchdog.sh:356` - Simplification: the `lastStartedAt` fallback in `pod_present` is a second acceptance path for the ceiling anchor that no intent requirement needs. The authorization named `runtime { uptimeInSeconds }` as preferred and the fallback as optional ("and/or"), so the narrower uptime-only form satisfies the requirement. It has zero test coverage in either direction - the fixture curl shim always emits `"lastStartedAt":null` (tests/fm-runpod-watchdog.test.sh:95-99) and the `uptime.null` case nulls both - so `parse_when "$started"` at :356-360 is never exercised. It is also the riskier of the two sources for the invariant it serves: `runtime` is null precisely when the pod is NOT running, so this is the only branch that can anchor a spend ceiling from a timestamp taken while the pod was not accruing uptime, and "first sighting wins" (:673-679) then freezes that anchor for the loop's life. `uptimeInSeconds` can only ever be read from a running pod and cannot produce that shape. Recommended remedy is removing the fallback; the absent-anchor case already has defined, tested behavior (anchor=unknown, ceiling not enforced, alarm).
- ⚠️ `bin/fm-runpod-watchdog.sh:730` - The success path reports a verified termination through `alarm`, which at :298 writes `blocked [key=runpod-watch-<task>]: ... has been stopped`. In bin/fm-classify-lib.sh a keyed `blocked` OPENS a decision (:484, :1648, :1711) and only `resolved` or `captain-held` carrying the same key closes it (:1565). The watchdog is the sole writer of that key and it returns and exits two lines later, so it can never close it. Every successful watchdog stop therefore leaves a permanently open blocker on the task's status channel, and `status_open_decisions` keeps surfacing the task as stuck until a human hand-writes `resolved [key=runpod-watch-<task>]`. This challenges a deliberate design choice (one verb for every condition), so it needs the author's call: either a terminal verb for the verified-stop path, or have the watchdog append the matching `resolved` line for conditions that have cleared.
- ℹ️ `bin/fm-runpod-watchdog.sh:487` - `stop_existing` gates only on `is_uint "$pid"`, which accepts `0`. With `0` in `state/<task>.runpod-watch.pid`, `kill -0 0` succeeds (it tests the caller's process group), so :489 runs `kill -TERM 0` and :498 escalates to `kill -KILL 0`, tearing down the whole process group of whoever invoked `arm` or `disarm` - i.e. the agent harness. The script's own writes can never produce `0` (`printf '%s\
' "$$"`), so this needs a corrupt or hand-edited pid file, the same input class the accepted `is_safe_uint` deadline guard already defends against, but with a larger blast radius. Remedy is one comparison: require `[ "$pid" -gt 1 ]` alongside `is_uint` at :487 (and at :471/:541 for consistency).

🔧 Fix: close watchdog alarms it opens; parse .env; guard pid

- ⚠️ `bin/fm-runpod-watchdog.sh:787` - `clear_alarm api-unreachable` is only reachable inside the `present_rc == 0` branch (:769). Concrete sequence: arm with a stale/typo'd pod id; poll 1 hits a transport failure -> `present_rc=2` -> `alarm api-unreachable` appends `blocked [key=runpod-watch-t1-api-unreachable]`; poll 2's list read SUCCEEDS but the pod is absent -> `present_rc=1` with `ever_seen=0` -> the branch at :787 alarms `pod-never-seen` and `continue`s at :791, never reaching :769. A successful, jq-parsed list read is proof the API is reachable, yet the false "cannot reach RunPod" blocker stays open on state/<task>.status for the entire life of the watch and only a human can close it. This defeats the round-2 fix's own stated invariant (script header :67-69; docs/runpod-pod-watchdog.md:76 "That same condition clearing ... closes it"). Remedy is mechanical and stays inside the machinery the fix round already added: hoist the `api-unreachable` clear so it runs for both `present_rc` 0 and 1, since either outcome proves the read succeeded.
- ⚠️ `bin/fm-runpod-watchdog.sh:833` - The `ever_seen == 0` guard short-circuits only the pod-absent branch, not the deadline evaluation. Concrete sequence: arm t1 with pod id `pod-typo` and a 6h deadline. Until the deadline, every poll returns `present_rc=1` and the guard at :787-791 correctly alarms `pod-never-seen` and `continue`s. Once the deadline has passed, a single transient transport failure (near certain over hours of 60s polling) yields `present_rc=2`; the `*)` branch at :799-802 alarms but does NOT `continue`, so execution reaches :833 with `effective=REC_DEADLINE` and `now >= effective`. `terminate_and_verify` calls podTerminate on `pod-typo`; RunPod replies "pod not found to terminate", which terminate_call:465-467 deliberately maps to success; `pod_present` then returns 1 (absent, because it was never there) and :691 logs "verified absent". The loop then calls `retire_alarms` - wiping the standing `pod-never-seen` warning that a real rented pod may still be billing under an id it was never given - appends `note: rented pod pod-typo ... has been stopped; absence confirmed by listing the account's pods`, deletes the record, and exits. A false all-clear on the money channel, contradicting docs/runpod-pod-watchdog.md:94-97 ("It will not end the watch on a pod it never saw") and the assertion intent at tests/fm-runpod-watchdog.test.sh:301. Marked ask-user because every honest remedy changes termination/verification policy (skip the deadline evaluation entirely while `ever_seen == 0`, or refuse to accept absence as proof of a stop for a pod never sighted), which is the author's deliberate design territory rather than a mechanical correction.
- ℹ️ `bin/fm-runpod-watchdog.sh:342` - `clear_alarm` removes the condition's ledger row outright, so a flapping condition re-opens on the very next poll rather than respecting ALARM_REPEAT_SECONDS. An API failing on alternate polls writes a `blocked`/`resolved` pair per pair of polls - roughly 960 status lines over an 8h unattended run against the 32 the 900s limiter was written to allow (header :64-67, docs line 71). Impact is bounded and not fleet-visible: `status_line_is_unread_surface` (bin/fm-classify-lib.sh:1471) admits only `note:` and reserved `pending-reply-` keys, and the OPEN DECISIONS fold reports the correct final state, so this is status-log growth, not wake noise. Recording it as the accepted cost of the round-2 instruction that a closed condition must be able to open again if it recurs.

🔧 Fix: never stop or clear a pod the watch never sighted

- ⚠️ `bin/fm-runpod-watchdog.sh:789` - The `pod-never-seen` alarm is keyed per task (`runpod-watch-<task>-pod-never-seen`, :301) but is cleared by sighting whatever pod the CURRENT record names, so a sighting of a different pod closes a standing warning raised about another id. Concrete reachable sequence: (1) `arm --task t1 --pod pod-typo` -> the loop alarms `blocked [key=runpod-watch-t1-pod-never-seen]: ... has never seen pod pod-typo ...` and writes the ledger row (:806). (2) The operator disarms, or re-arms for the task's next pod. `cmd_disarm` (:622-624) removes the record, pid, and observed files but NOT `state/t1.runpod-watch.alarms`, and `retire_alarms` (:377-380) deliberately writes the `pod-never-seen` row back, so the row and the open decision both survive - exactly as docs/runpod-pod-watchdog.md:84-86 specifies. (3) `arm --task t1 --pod pod-beta` -> `retire_alarms` preserves the row again -> the new loop's first poll finds pod-beta present -> :789 `[ "$ever_seen" -eq 1 ] || clear_alarm "$task" pod-never-seen` finds the surviving row and appends `resolved [key=runpod-watch-t1-pod-never-seen]: the pod watchdog for t1 has now seen pod pod-beta ...`. `status_open_decisions` (bin/fm-classify-lib.sh:510) folds that as a close, so the warning that a real GPU may still be billing under `pod-typo` disappears from OPEN DECISIONS having been closed by an unrelated pod. This defeats the invariant this fix round was built to establish and contradicts docs/runpod-pod-watchdog.md:85 ("only an actual sighting OF THE POD closes it"). The new test at tests/fm-runpod-watchdog.test.sh:320-330 covers disarm-preserves-the-alarm but never the cross-pod clear, so it passes with this hole open. Marked ask-user because the remedy, not the defect, needs authorization: the smallest honest fix binds the alarm to the pod it names - either a per-pod condition slug (`pod-never-seen-<pod>`, which changes the decision-key format documented at docs/runpod-pod-watchdog.md:73,86) or recording the alarmed pod id in the ledger row and clearing only on a matching sighting - and both change the alarm-keying policy the author and the prior round's instructions deliberately specified.
- ⚠️ `bin/fm-runpod-watchdog.sh:795` - `pod_start` lives only in the loop process's memory and `cmd_arm` deletes the derived-state file at :552, so the "first sighting wins" rule the comment at :792-794 states ("a pod that restarts reports fresh uptime, and re-anchoring on that would push the ceiling later than the spend it is meant to bound") is enforced only WITHIN one process and is lost at every re-arm. Concrete sequence with the intent's own figures: pod-alpha's runtime starts at T0; `arm --task t1 --pod pod-alpha --ceiling-usd 20 --rate-usd-hr 3.49` derives ceiling_seconds=20630 and the loop anchors effective=T0+20630. At T0+10000s the pod's runtime restarts (container crash/restart, or a stop-and-resume), so `runtime.uptimeInSeconds` resets. At T0+10100 the author extends the declared deadline and re-arms, which docs/runpod-pod-watchdog.md:43 describes as normal usage. `cmd_arm` kills the old loop and deletes `.observed`, so the new loop starts with `pod_start=''`, reads uptime=100 at :461-466, and sets POD_START_EPOCH=T0+10000 -> effective=T0+30630. Total billed uptime allowed becomes 10000+20630=30630s, about 29.70 USD against the declared 20 USD ceiling, with `status` reporting `anchor=pod-start` as though the bound were intact. This directly contradicts the unconditional claim at docs/runpod-pod-watchdog.md:44 ("Re-arming cannot extend the ceiling: ... a second arm re-derives the same bound rather than buying more of it"), which holds only while uptime increases monotonically. The ceiling test at tests/fm-runpod-watchdog.test.sh:488-496 re-arms with a FIXED uptime fixture (7200), so it cannot catch a runtime reset. Marked ask-user because the remedy extends the change rather than correcting it: it needs the earliest observed pod start to become durable across processes (stop deleting `.observed` at :552 and seed `pod_start` from its `pod_start_epoch`, or track cumulative rather than wall-anchored uptime), which is new persistence and a design decision the author made deliberately when `arm` was given a clean-slate `.observed`.
- ℹ️ `bin/fm-runpod-watchdog.sh:705` - The fix round hoisted `clear_alarm api-unreachable` so any parsed pod-list read closes it (:785), but `terminate_and_verify` runs its own `pod_present` reads (:703-712) with no such clear, and that function never returns while verification keeps failing. Concrete sequence: pod sighted (ever_seen=1), a transport failure lands on the very poll at which the effective deadline passes -> the `*)` branch at :818 opens `blocked [key=runpod-watch-t1-api-unreachable]` and, because :785 was skipped for present_rc=2, it stays open; the same poll falls through to :860 and enters `terminate_and_verify`. The API recovers a minute later, but `podTerminate` does not land (the pod stays listed), so the verify reads succeed with rc=0 forever, `terminate-unverified` alarms at :716, and the loop at :695 never exits. The false "cannot reach RunPod" blocker sits open on state/<task>.status alongside the true one for the rest of the watch, and only a human can close it. This is the same defect class as the round-3 finding and violates the same stated rule that a successful jq-parsed list read proves reachability. Mechanical remedy inside existing machinery: call `clear_alarm "$task" api-unreachable` for rc 0 and rc 1 in the verify `case` at :705-712, exactly as :785 does in the main loop. Low severity because reaching it needs a transport failure on the one poll where the deadline crosses AND a termination that never lands, and the accurate `terminate-unverified` alarm is open beside it throughout.

🔧 Fix: bind watchdog alarms and ceiling anchor to their pod
3 issues (1 error, 2 warnings) still open:

- 🚨 `bin/fm-runpod-watchdog.sh:844` - The persisted pod-start anchor is a WALL-CLOCK instant with no staleness bound, so for a pod that was not accruing uptime during the gap it over-counts on the terminating side. Concrete reachable sequence, all inside documented usage: (1) 08:00 `arm --task t1 --pod pod-alpha --deadline <+6h> --ceiling-usd 20 --rate-usd-hr 3.49` -> ceiling_seconds=20630; the loop sights pod-alpha and anchors pod_start=08:00, written to `.observed` (:926). (2) 10:00 the operator STOPS the pod on RunPod - the id is retained, the volume is kept, the GPU is released and stops billing - and disarms. `cmd_disarm` (:685) deliberately no longer removes `.observed`. (3) The next morning at 09:00 the operator resumes pod-alpha (RunPod restarts `uptimeInSeconds` from 0) and re-arms t1 on the same pod with a fresh deadline. `cmd_arm` (:610) finds `observed_anchor` succeeds for pod-alpha and KEEPS the stored anchor. (4) The new loop at :844 loads pod_start=08:00 yesterday, logs 'anchor carried over', and at :910 computes ceiling_deadline = 08:00+20630s = 13:44 YESTERDAY. :930 compares now >= effective -> true on the FIRST poll -> `terminate_and_verify` destroys a healthy run that has burned roughly 2 hours of GPU, verifies the kill, and reports it as a completed stop. Nothing warns first: `cmd_arm` validates only that the DECLARED deadline is in the future (:572) and then prints 'ceiling 20 USD ... caps uptime at 20630s' as though the bound were intact. This path did not exist before the last fix round, which deleted `.observed` on every arm; it is newly introduced by the persistence that round added, and it inverts the header's own doctrine at :51-57 ('Terminating a healthy run destroys work and money already spent, so every uncertain state alarms and keeps polling instead of terminating'). The distinction the anchor cannot make is container-restart (still billing, reuse is right - the case the fix targeted) versus stop/resume (not billing, reuse is wrong), and `uptimeInSeconds` alone cannot tell them apart after the fact. No test covers it: tests/fm-runpod-watchdog.test.sh:520-559 re-arms with the pod continuously listed and only asserts the deadline did not move LATER, never that a stale anchor cannot put it in the past. Marked ask-user because the remedy changes semantics the author explicitly ordered in the prior round ('reuse the stored anchor rather than re-deriving one from current uptime'), not because the defect is unclear. Two candidate remedies for that decision: (a) the narrow containment - refuse to enforce a carried-over anchor whose ceiling instant is ALREADY past at arm time, alarming `ceiling-anchor-unknown` and falling back to the declared deadline, which keeps persistence and honours fail-toward-not-killing; or (b) the durable form the prior round's own instructions named as an alternative - accumulate observed uptime rather than anchoring to wall clock, so a stopped interval simply does not count.
- ⚠️ `bin/fm-runpod-watchdog.sh:716` - `cmd_status` gates the derived-state read on the pod id matching, but not on the derived state belonging to the CURRENT arm, so it prints an instant and an anchor from a previous watch as though they were in force. Concrete sequence: (1) `arm --task t1 --pod pod-alpha ... --ceiling-usd 20 --rate-usd-hr 3.49`; the loop sights the pod and writes `.observed` with effective_deadline_epoch=E, effective_source=ceiling, anchor=pod-start. (2) `disarm --task t1`, which since the last fix round keeps `.observed` (:685). (3) Later, `arm --task t1 --pod pod-alpha --deadline <new>` with NO ceiling. `cmd_arm` (:610) keeps the file because the pod matches. (4) `status --task t1` now takes the :716 branch, finds the stored pod equals REC_POD, and reads effective=E, source=ceiling, anchor=pod-start. E is a valid uint so the :721 fallback never fires, and REC_CEILING_SECONDS is empty so no ceiling= note contradicts it. Output: `t1: pod=pod-alpha deadline=<E> source=ceiling anchor=pod-start declared-deadline=<new> ...` - a ceiling instant the record no longer even declares, reported as the instant in force. The comment at :710-713 states this is exactly what the branch exists to prevent, and docs/runpod-pod-watchdog.md:57-58 promises `unresolved` until a loop has reported and that 'a ceiling the watchdog is not enforcing can never be mistaken for one it is'. Normally the window is one poll, but when the re-armed pod is never sighted the loop `continue`s at :897-900 and never reaches `observed_write` (:926), so the stale figure stands for the entire life of the watch - precisely the case where the operator most needs the truth, because nothing is being enforced at all. Before the last fix round `arm` deleted `.observed` unconditionally and status correctly fell back to source=declared anchor=unresolved; this is a regression that round introduced. The defect lives in the derived-state file's reporting fields, which are wider than the anchor persistence actually requires: the strictly narrower form is for `cmd_arm` to preserve only what the anchor needs - rewrite `.observed` carrying `pod` and `pod_start_epoch` and reset anchor/effective/source - so `status` prints `unresolved` until the live loop republishes. Mechanical and inside existing machinery (observed_write already takes every field).
- ⚠️ `bin/fm-runpod-watchdog.sh:411` - `pod-never-seen` rows are now per-pod but the DECISION KEY is still per-task (:325, deliberately, per the prior round's instruction), and bin/fm-classify-lib.sh:485-490 keeps only the newest `blocked` note per key. When a second subject clears, `clear_alarm` drops its row and returns early at :411 without restating the note, so the surviving decision keeps displaying the note about the pod that is now fine. Concrete sequence: (1) `arm --task t1 --pod pod-typo`; the pod is not in the account -> :875 alarms, open set shows '... has never seen pod pod-typo ...'. (2) The operator re-arms t1 on pod-beta - typically right after requesting it, before RunPod lists it - so the first poll returns rc=1 and :875 alarms again for pod-beta; the ledger now holds two rows and the open set's note is overwritten to name pod-beta. (3) pod-beta appears; :857 calls `clear_alarm ... pod-never-seen "$REC_POD"`; the pod-beta row is dropped, `remaining=1` because the pod-typo row survives, and :411 returns before any `resolved` line. The decision correctly stays OPEN - but its visible text now reads 'the pod watchdog for t1 has never seen pod pod-beta ... the pod id may be wrong or stale while a rented pod keeps billing; check the account by hand', about a pod the operator can see running fine. The only standing warning that a GPU may be billing under pod-typo is no longer named anywhere in OPEN DECISIONS, so the natural operator response is to close it by hand - exactly the loss of the safety property the last fix round was built to establish, reached by a different route than the cross-pod clear it fixed. docs/runpod-pod-watchdog.md:90-91 promises the opposite ('news of one pod never closes a decision opened about another'). tests/fm-runpod-watchdog.test.sh:332-349 covers only the single-surviving-subject case, so it passes with this hole open. Smallest honest remedy, inside existing machinery and adding no state: when `clear_alarm` drops a subject but `remaining` is non-zero, re-append a `blocked [key=...]` line naming a subject still holding the condition open, so the open decision's note matches a row that is actually still true.

🔧 Fix: refuse stale carried anchors; stop reporting retired figures
2 warnings still open:

- ⚠️ `bin/fm-runpod-watchdog.sh:968` - The new stale-anchor refusal uses a wall-clock proxy (`ceiling_deadline < loop_start`) for a question the loop already has direct evidence about, so it disables the ceiling in the one scenario the intent was written to close. `pod_present` sets POD_START_EPOCH from `uptimeInSeconds` on EVERY poll (:558-560), but :914 only consults it when `pod_start` is empty, so a carried anchor throws that reading away. Concrete reachable sequence, entirely inside documented usage: (1) 08:00 `arm --task t1 --pod pod-alpha --deadline <20:00> --ceiling-usd 20 --rate-usd-hr 3.49` -> ceiling_seconds=20630, the loop anchors pod_start=08:00 and would enforce the ceiling at 13:44. (2) 11:00 the agent session is replaced and the watchdog process dies - this is the gap the intent names verbatim ('the in-session deadline watchdog it armed was NOT running when that session was replaced'). The pod keeps billing, uninterrupted, past 13:44. (3) 14:30 the new agent does the natural remedy and re-arms t1 on the same pod with a fresh declared deadline of 20:00. (4) The loop carries pod_start=08:00 (pod_start_carried=1), loop_start=14:30, ceiling_deadline=13:44 < 14:30 -> :969 sets anchor=stale, :970 alarms, and the ceiling is NOT enforced; only the declared 20:00 deadline is. The pod burns 5.5 more hours at 3.49 = ~19 USD of unintended spend, against a declared 20 USD ceiling that was already provably blown. The signal that settles it was in hand: at 14:30 the same response reports uptimeInSeconds=23400, so POD_START_EPOCH=08:00 matches the carried anchor exactly - proof the pod never stopped. The header comment at :958-967 and the alarm text at :971 both assert 'the pod was almost certainly stopped and resumed', a conclusion contradicted by data the same request already returned; that is a wrong label emitted without erroring. Two compounding consequences: the stale pod_start is re-persisted by `observed_write` (:992) on every poll and carried again by `cmd_arm` (:651), and `cmd_disarm` deliberately keeps `.observed` (:730), so once a pod id is marked stale its ceiling can never be enforced again by any future watch - only arming a DIFFERENT pod or the pod leaving clears it. The mirror case is bounded and fine (a genuinely stopped-and-resumed pod with a long gap always lands far in the past and is correctly refused). Marked ask-user, not auto-fix: the remedy contradicts the rule the author stated explicitly in round 5 ('The watchdog must NEVER terminate on its first poll because of a carried-over anchor'), so it needs the author's decision rather than a mechanical repair. The narrow refinement that stays inside remedy (a) and adds no state: treat a carried anchor as stale only when the freshly-read POD_START_EPOCH DISAGREES with it (the pod's reported uptime shows a restart); when uptime corroborates the carried anchor, the ceiling is genuinely spent and should be enforced. That keeps the fail-toward-not-killing doctrine for the case uptime truly cannot settle, and stops the tool from standing down in the case it exists for.
- ⚠️ `bin/fm-runpod-watchdog.sh:446` - The note-restatement the last fix round added re-opens a `pod-never-seen` decision the operator already closed by hand, and docs/runpod-pod-watchdog.md:101 is what tells them to close it by hand. The ledger keeps `pod-never-seen` rows forever (`retire_alarms` :471-474 preserves them across disarm and re-arm) and has no knowledge of a manual resolution, so the surviving row can be restated indefinitely. Concrete sequence: (1) `arm --task t1 --pod pod-typo`; the id is wrong, :925 alarms, and OPEN DECISIONS holds `runpod-watch-t1-pod-never-seen` naming pod-typo. (2) The operator checks the RunPod account, finds nothing billing under that id, and does exactly what the docs instruct - appends `resolved [key=runpod-watch-t1-pod-never-seen]: checked the account, nothing is billing`. bin/fm-classify-lib.sh:491-494 drops the key; the decision is correctly closed. The pod-typo LEDGER row is untouched, because the operator cannot edit it. (3) Later the same task runs a second GPU job: `arm --task t1 --pod pod-beta`. RunPod does not list pod-beta on the first poll (the normal request-then-appear delay the round-5 fix was built around), so :925 alarms for pod-beta and adds its row. (4) pod-beta appears; :907 calls `clear_alarm ... pod-never-seen "$REC_POD"`; the pod-beta rows drop, `standing` is set from the surviving pod-typo row at :433, and :446 appends `blocked [key=runpod-watch-t1-pod-never-seen]: ...has never seen pod pod-typo...`. The fold re-opens the decision the operator answered. Nothing in this watchdog can ever close it again - by design only sighting pod-typo does, and pod-typo is not being watched - so the operator must close it by hand a second time, and any further partial clear on this task re-opens it again. This is a behavior change the last fix round introduced: the prior code returned early on `remaining != 0` and appended nothing, so a hand-closed decision stayed closed. Marked ask-user rather than auto-fix because every honest remedy changes a design the author fixed deliberately in round 5: either give `pod-never-seen` a per-SUBJECT decision key (`runpod-watch-<task>-pod-never-seen-<pod>`), which the round-5 instruction explicitly kept per-condition, or make the restatement conditional on the decision still being open, which means reading `status_open_decisions` from bin/fm-classify-lib.sh and is a new dependency rather than a correction of what the line already does. It is the remedy, not the defect, that needs authorization here.
</details>

<details>

- ℹ️ `tests/fm-runpod-watchdog.test.sh` - Testing a deliberately detach-by-design tool leaves orphaned processes if the suite is interrupted. I found one watchdog from a fixture root that no longer exists (/tmp/fm-runpod-watchdog.hl9NJO/anchor-persists) still polling, started before any of my runs. It is harmless (its record is gone, its endpoint is https://fake.invalid) and I could not reproduce it: all four of my suite runs and 12 targeted arm/re-arm/disarm cycles left nothing behind. Worth knowing that a CI timeout killing this suite mid-run will strand watchdogs by design, since surviving process-group teardown is the property under test. No action needed on the change itself.
- <code>`bash tests/fm-runpod-watchdog.test.sh` — 23 cases, run 4× (deadline terminates and is verified by listing, unreadable record never terminates, never-sighted pod, pod-start ceiling anchor, stale carried anchor, status never quoting a retired arm's figures, detachment survival, credential containment)</code>
- <code>`bash tests/fm-documentation-audiences.test.sh` — validates the docs/runpod-pod-watchdog.md registration this change adds</code>
- <code>Manual end-to-end scene 1: `fm-runpod-watchdog.sh arm --task c6 --pod pod-c6-h100 --deadline <+6h> --ceiling-usd 20 --rate-usd-hr 3.49` then `status --task c6`, against a faked RunPod endpoint with the pod already 2h old</code>
- <code>Manual end-to-end scene 2: `setsid bash agent-session.sh` arms then exits; asserted the agent pid dead, watchdog alive in its own sid, then observed the unattended termination, listing-verified absence, wake-channel stop note, and self-exit</code>
- <code>Manual end-to-end scene 3: `fm-runpod-watchdog.sh run --task c6` with a truncated record — no podTerminate, `RECORD UNREADABLE` status, blocked alarm raised</code>
- <code>Manual end-to-end scene 4: `run --task c6` with a pod id never in the account, deadline long passed — no podTerminate, and `status_open_decisions` still shows the warning open after `disarm`</code>
- `Alarm rate limiting at the shipped default: 8 polls of an unreadable record appended exactly one alarm line`
- <code>Leak check: 12 arm/re-arm/arm-other-pod/disarm cycles asserting no watchdog process survives `disarm`</code>
</details>

<details>

✅ No issues found.
</details>

<details>

🔧 Fix: drop escaped apostrophe from watchdog test message
</details>

<details>

✅ No issues found.
</details>

---

## Known limitations accepted for this change

Two review findings were deliberately deferred rather than fixed here, and are filed as follow-ups. Neither can cause the watchdog to terminate a pod that should be running; both were judged not to warrant a further fix round on this change.

### 1. A carried-over ceiling anchor is treated as stale using a wall-clock proxy, when uptime could settle it

When a stored pod-start anchor is carried across a re-arm, the anchor is refused whenever the ceiling instant it produces is already in the past. That refusal is correct for a stopped-and-resumed pod, but it also fires for a pod that never stopped — including the exact flow this tool exists for, where the agent session is replaced and a new agent re-arms the same still-running pod. In that case the ceiling is not enforced and only the declared wall-clock deadline bounds the run.

The evidence that would settle it is already in hand: the same pod-list response reports `uptimeInSeconds`, so a freshly derived start that agrees with the carried anchor proves the pod never restarted. The narrow refinement is to treat a carried anchor as stale only when the freshly read start *disagrees* with it.

Failure mode is under-enforcement of the ceiling, never an unprompted termination. The declared deadline still bounds the meter. Note the effect is sticky: once a pod id's anchor is marked stale it is re-persisted, so no later watch of that same pod enforces its ceiling until a different pod is armed.

### 2. Restating a partially cleared `pod-never-seen` note can reopen a decision an operator closed by hand

The never-seen alarm keeps one per-task decision key while tracking per-pod subjects. When one subject clears and another remains, the note is restated so the open decision names a pod that is genuinely still unseen. If an operator had already closed that decision by hand — which the operator documentation tells them to do — the restatement reopens it, and only sighting the original pod would close it again.

Failure mode is noise on the decision channel, not incorrect termination or a missed alarm. Both honest remedies change keying or add a dependency on the status classifier, which is why this was not folded into the final fix round.


kunchenguid:
Speaking as Kun's firstmate:

**Outcome:** `waiting-ci` — tip `7920193ed4c848219c6582391f839ff9c3649c10` attestation **MATCH**. First-time fork CI/NM approved after diff review (runs `34709107351` CI, `34709107358`/`34710233018` NM now queued/in progress). No green CI yet — not merge-eligible this pass.

**contract-class:** **opt-in** — `arm`/`disarm`/`status`/`run` are operator- (or agent-) invoked; nothing always-on at session-start. Deferred follow-ups #4310/#4311 do not block.

**VISION:** One captain/interface aligns (explicit killswitch CLI). Authority explicit aligns (arm required; fail toward not killing). Scripts vs agents aligns (detached script owns meter stop). Restart non-event aligns (survives agent replacement). Delegation/spine aligns (deadline-only terminate + listing verify). Fleet/vendor: RunPod-specific but scoped. Scope aligns (new script + docs/tests only).

**Security FYI (tip only, not waiting-captain):** reads `RUNPOD_API_KEY` from gitignored `.env` via non-sourcing parse + curl stdin header (never argv/logs); can `podTerminate` after deadline. Fail-closed on unreadable key/record/API; never-sighted pods not terminated. Not a merge gate while CI pending.

Do not merge this pass.

Alberto-Codes:
Withdrawing this one. On reflection it is the wrong shape for this project.

It adds ~1,900 lines of RunPod-specific tooling — a vendor script, its docs and
its tests — to a shared codebase where almost nobody is renting RunPod GPUs.
That is infrastructure for one operator's setup rather than a fix to firstmate
itself, and it belongs in a fork, which is where it will live.

If the general problem underneath it is ever interesting here — a rented
resource that keeps billing after the agent session that acquired it has gone —
that would be worth raising separately as a small, provider-agnostic seam
rather than as one provider's implementation.

Thanks for the workflow approvals and the review time already spent on it.

PR #4314, opened 2026-09-12
fix(brief): make the no-mistakes pipeline step unmissable

kdrbrl:
## Why

Crewmates repeatedly finished their work, ran the local suite, saw it green, and appended a
`done:` status line without ever starting the no-mistakes pipeline. It happened five times in
two days on one project. Every time the work itself was fine and the correction was simply
"now run it", so this is a brief-comprehension failure rather than a competence one.

The pattern was always the same: the worker reports "all green: lint, typecheck, 378 unit and
contract tests, build, budget" and treats that as the definition of done. That is a
reasonable misreading - the worker HAS just watched a lot of things go green - which is why
the scaffold should make the mistake impossible instead of relying on careful reading.

## What changed

The generated brief now states the exact `done:` literal at the point in the status protocol
where the worker composes that line, rather than describing it earlier as background. The two
fixed `done:` forms are pinned to a single owner so they cannot drift apart, and the brief
test asserts the protocol literals.

## Worth knowing for review

The first review round caught a regression this change introduced: replacing the terminal
literal with a back-reference ("append form 2 above") would have invited approximations like
`done: PR <url> is green`, which `bin/fm-inactive-reconcile.sh` and `bin/fm-crew-state.sh`
both fail to parse - silently losing PR extraction and the CI-ready reading. That is the same
class of silent failure the change exists to close, so the literal is now spelled out exactly
where it is written. Two further findings were fixed in the same round: the duplicated
literals had no pinned owner, and the negative-mode test could pass vacuously because it never
asserted the brief existed before reading it.

Review, tests, documentation and lint all passed. The branch could not be pushed to the
upstream repository from the account that produced it, so this arrives from a fork.

### DIFF
--- a/bin/fm-brief.sh
+++ b/bin/fm-brief.sh
@@ -449,10 +449,12 @@ case "$MODE" in
     # two valid forms are restated at that exact moment. Five no-mistakes tasks
     # reported `done:` on a green LOCAL suite, never having started the
     # pipeline; the Definition of done owns the forms, this is the one
-    # deliberate reinforcement at the point of the mistake.
+    # deliberate reinforcement at the point of the mistake. The literals come
+    # from their single owner in bin/fm-dod-lib.sh so this restatement cannot
+    # drift out of step with the Definition of done rendered below it.
     RULE4_DONE="
    For this task \`done:\` has exactly two valid forms, both fixed by Definition of done:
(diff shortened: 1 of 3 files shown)


kdrbrl:
Withdrawing this - the captain has decided against sending it upstream. Apologies for the noise.

PR #4350, opened 2026-09-13
fix(bin): detect omp through its bun-global script path

linksucre:
## Problem

omp installed by `bun i -g @oh-my-pi/pi-coding-agent` runs as `bun .../bin/omp` (env-bun shebang into `dist/cli.js`), so `ps -o comm=` reports `bun` and the omp identity rides entirely in the script path at argv[1].

On that install form every primary session failed session-lock acquisition with `cannot locate harness process in ancestry` and rendered the supervision block as an unknown harness, stranding the session read-only.

The 18.1.11 verification covered the Bun-compiled single executable named `omp`; the bun-global script is a second real distribution that shares no process-name signal with it.

## Change

- `bin/fm-session-lock-lib.sh`: dedicated `bun` arm in `fm_harness_process_matches` accepting only an exact harness path component in the script argument (argv[1]), never the loose node-style args grep, so an ordinary bun script cannot claim an identity.
- `bin/fm-harness.sh`: the same arm in `detect_own`'s ancestry walk and in `ancestry_names_omp`, so `FM_OMP_HARNESS=omp` marker precedence also holds for the bun-scripted form.
- Tests: portable regressions in `tests/fm-omp-harness.test.sh` (real named-shell fixtures) and `tests/fm-session-lock-ancestry.test.sh` (fake ps shapes); negative fixtures run orphaned so a host whose test runner sits under a real omp cannot lend that ancestry to them.
- `harness-adapters` omp reference and `docs/verification/supervision.md` updated with the dated 18.1.15 bun-global evidence.

## Verification

Live on the bun-global primary itself (omp 18.1.15, process 71113, `ps -o pid,ppid,comm=,args=` printing `71113 7411 bun bun /Users/author/.bun/bin/omp`):

- `bash -c '. bin/fm-session-lock-lib.sh; fm_harness_ancestry_pid'` from a tool child prints `71113`.
- `env -u CLAUDECODE bin/fm-harness.sh` prints `omp`; a stale `CLAUDECODE=1` in the launching shell still loses to `FM_OMP_HARNESS=omp` under real omp ancestry.

Suites on this branch: `tests/fm-omp-harness.test.sh` 12 ok, `tests/fm-session-lock-ancestry.test.sh` 8 ok, `bin/fm-lint.sh` clean over the changed scripts, `bin/fm-doc-audience-check.sh` ok.


kustomrtr:
Same diagnosis here, so this may be useful rather than a competing PR: I implemented the bun-global case independently before checking the issue list, and my branch is `hotfix/omp-bun-shim-identity` on `kustomrtr/firstmate` (verified on omp 18.1.16, installed with `bun i -g @oh-my-pi/pi-coding-agent`, where `ps -o comm=` says `bun` and the identity only survives in `bun .../bin/omp`).

Two differences in how the identity is resolved, both about avoiding false positives:

1. **One owner for the interpreter case.** I put it in `fm_harness_interpreter_name()` inside `bin/fm-session-lock-lib.sh` and had `bin/fm-harness.sh` call that helper. Naming and the fleet session-lock identity then cannot disagree about the shape. Here, each site carries its own `case` arm, so the two can drift.
2. **Basename, not path component.** I require the script argument's own basename to be a known harness name. A component match (as in this PR) also accepts `bun /home/omp/tools/build.js`, where `omp` is just a directory. I tightened `fm_harness_path_name()` the same way, preferring the executable's basename over the nearest matching directory component; Claude's versioned native install (a path whose basename is only a version number) still resolves through the component walk.

Beyond that, my branch also covers the liveness side of the same shape: `bin/fm-agent-process-lib.sh` classifies a bun-shaped omp session as busy, and `bin/backends/tmux.sh` learns the interpreter form for endpoint and busy probes, with the two verification docs updated. Fixtures pin the cases that decide this: a bun option operand in the script position is rejected, and a script named `omp` inside an unrelated directory cannot claim the identity.

There is overlap in flight with #4599 (bun-script launcher shape for omp >= 18.1.22) as well. Happy to send a small patch with just the two semantic changes above rebased on top of this branch, so one version lands instead of three.

No pressure either way; use whatever is useful here.

PR #4403, opened 2026-09-14
fix(watch): bound stable-hash captain-call reminders

justin-belcourt:
## Intent
Stop repeated watcher reminders for stopped work the captain already holds when the pane hash is stable.

## What changed
Stable-hash captain-call holds stay on the long resurface cadence instead of wedge-escalating. Held / re-held / unheld / unreadable controls covered in the watcher suite; lint and docs checks passed in the isolated copy.

## Risk
Low. Watcher triage only; no project product code.

## Note
Published from fork `justin-belcourt/firstmate` because the operator account has pull-only on this upstream and no direct push.


justin-belcourt:
Closing: fix is landed locally on the captain machine. Prefer private local ownership; no need for public upstream merge for this install.

Issue #4414, opened 2026-09-14
Codex: supervision continuity lapses across turn endings and user interruptions


jcpoyser:
## Problem

When using Firstmate with Codex, a long-running review repeatedly triggered warnings that watcher supervision was missing after conversational turn endings.
This also happened around user approvals and Bearings requests.

The agent acknowledged that it ended turns without renewing the required foreground supervision checkpoint.
This is an observed coordination/integration failure, not a proven Codex runtime defect.

Related: https://github.com/kunchenguid/firstmate/issues/4412 describes related supervision/reporting symptoms with a different primary harness.

## Reproduction outline

1. Start Firstmate using Codex and dispatch a review that lasts longer than one foreground checkpoint.
2. While it runs, ask a status question, provide an approval, or invoke Bearings.
3. Let a checkpoint expire and observe whether supervision continues across the conversational turn ending.
4. Compare the Stop-hook warning with actual watcher liveness and worker progress.

## Actual behavior

- Repeated Stop-hook warnings said work was in flight without a live watcher.
- Recovery depended on the agent remembering to restart the foreground checkpoint loop.
- An attempted bundled drain/checkpoint command was correctly refused; running the checkpoint separately worked.
- A separate observation failure made a working review appear unreadable; a subsequent successful observation confirmed it was still working.
This should not be interpreted as proof that the worker stopped.

No lost work or worker crash was established.

## Expected behavior

- Supervision continuity is maintained across user replies, status requests, and checkpoint expiration, or the limitation is made clear before dispatch.
- Observation failures are distinguished from worker failure.
- Recovery guidance works through the supported command interface.
- Warnings reflect real coverage gaps rather than being merely suppressed.

Please investigate which guarantees should be enforced by the Codex integration and which failures are agent instruction compliance.
Include regression coverage for user interruptions and checkpoint expiration; any alternative wake mechanism needs end-to-end verification.

kunchenguid:
Speaking as Kun's firstmate:

[...]

PR #4422, opened 2026-09-14
fix(supervision): give a Codex primary hook-owned watcher continuity

andy-n:
## The problem

Supervision continuity is armed per-harness, and a Codex primary had no structural re-arm at all. `docs/supervision-protocols/codex.md` described a model-driven loop: the agent was expected to voluntarily re-launch `bin/fm-watch-checkpoint.sh` every turn. Any turn that ended with a final reply instead of a new checkpoint ended blind, and nothing recovered until the operator typed again.

Claude re-arms structurally via a `Stop` hook, Pi via an extension, Cursor via a parked stop hook. Codex relied on diligence, which is not a mechanism.

Observed in practice: completed workers sat ~20 minutes unreviewed, and the turn-end guard banner recurred. Switching a primary from Claude to Codex silently removed the only automatic re-arm; replacing the Herdr server then killed the live watcher (it is a child of the primary's process tree), and with nothing to re-arm, an ordinary event became a permanent outage.

## What was measured, not assumed

All hook behaviour was established in an isolated lab against Codex 0.154.0, not inferred:

- **`Stop` with `"async": true`** fires, and the turn is genuinely not held — but its `exit 2` and stderr are **silently discarded**. Claude's `asyncRewake` design therefore cannot be ported as-is.
- **Synchronous `Stop`** holds the turn open for the hook's duration and **does** deliver `exit 2` + stderr as a continuation; the follow-on payload carries `stop_hook_active: true`.

So this uses a hook-owned parked `Stop` plus exit-2 rewake, accepting that the pane reads busy while the hook parks. End to end in the lab, a worker's completion reached the Codex primary in 14s with no user prompt.

## What this adds

- `bin/fm-codex-stop-autoarm.sh`: a `Stop`-owned auto-arm for a Codex primary, mirroring the Claude sibling, including its refusal to publish a rewake row it cannot bind.
- `fm_supervision_model` maps Codex to `autoarm`, with `FM_AUTOARM_PREFIX` plumbed through the production callers so a Codex home reads its own ledger rather than Claude's.
- Prefix-aware cooperative mode in `bin/fm-turnend-guard.sh`, so the guard stops reporting `SUPERVISION IS OFF` on healthy hook-owned turns.

**It retires nothing.** `bin/fm-watch-checkpoint.sh` remains Codex supervision: the hook is the preferred path, and the checkpoint stays the fallback wherever the hook cannot be shown to be firing. An earlier iteration did retire it and was reverted — a Codex home must never be left with zero arm paths.

The turn-end guard's registered command is deliberately left **byte-identical**. Codex keys hook trust to a hash of the command, so appending a flag would have invalidated the already-approved guard entry and opened an upgrade window with no auto-arm *and* no blind-turn guard. Only the genuinely new entry needs approval.

## Known limitation, stated deliberately

Mapping Codex to `autoarm` means a fresh beacon counts as healthy without a live watcher. That is correct where the hook fires and the watcher only runs between turns. On the documented fallback path — the first-session approval window, or an operator who declines the hook — a watcher that dies mid-turn is reported healthy until the beacon ages past grace, rather than immediately.

This is a **delayed warning, never a blind turn**: `bin/fm-guard.sh` is the only caller and the turn-end guard stays PID-strict. Distinguishing "the hook fires in this home" from "this is a Codex primary" needs durable state beyond this change's scope, so it is documented at the mapping, beside the approval-window text, and here, rather than left for someone to discover.

## Operator note

Codex persists hook trust per entry, so the new `Stop` entry is untrusted on the first session after this change and needs a one-time approval. Until then the home supervises itself on the checkpoint fallback, and the ordinary-wake guidance now says so conditionally instead of claiming the hook already owns continuity.

## Testing

`tests/fm-codex-stop-autoarm.test.sh` is new (17 cases). Also green: `fm-turnend-guard` 95/95, `fm-claude-stop-autoarm` 41/41, `fm-guard-stale-banner` 33/33, `fm-supervision-instructions` 10/10, `fm-secondmate-harness` 50/50, plus the live end-to-end tier. `fm-lint` clean, doc-audience clean.

The live-e2e gate previously asserted on the vendor CLI's *error wording*; those assertions were dropped in favour of the behavioural one (non-zero exit for an impossible thread id), so a Codex release rewording a message cannot turn the lane red on working software.

One fixture flake was found and closed rather than retried: a `#!/usr/bin/env bash` script execs twice, and `ps` reports the interpreter in between, so an identity recorded in that window never matched later. The fixture now names the interpreter (one exec, no window) and waits on that condition, failing loudly if an identity never settles. Healthy-close case: 35/35, previously 1 failure in 6.

## Scope note

A Gram-intake half was developed alongside this and **deliberately removed** before submission: its deletion-purge inferred "the owner deleted this message" from absence in a listing, and a pane/audience change produces an identical empty listing — so it could destroy undeleted messages and drop durable wake rows. That needs an audience-aware, fail-safe design and is not part of this change.

The intermediate `no-mistakes(review): …` commits are validation-pipeline iterations from the authoring environment. **Squash on merge** if you prefer a single commit; the final state is what matters.


kunchenguid:
Speaking as Kun's firstmate:

Triage linked field report **#4414** (Codex: supervision continuity lapses across turn endings and user interruptions) to this PR as `existing-pr→#4422`. On current main Codex still relies on model-driven `fm-watch-checkpoint.sh` re-arm; this PR is the structural Stop-owned continuity repair. Prefer landing/helping here over a competing PR for #4414.

kunchenguid:
Speaking as Kun's firstmate:

First look on HEAD `5183bb92abf5bcef78fc7c16a6dfce22ce903c30`. MERGEABLE/UNSTABLE. Fork CI + NM approved this pass after safe file-list review (no `.github/workflows` changes; tracked `.codex/hooks.json` gains a Stop entry only): CI `34812341381` (in_progress), Require no-mistakes `34812341385` **FAIL**. Attestation **MISSING** (no `no-mistakes-pipeline-attestation:v1` binding the tip).

**Contract-class: new-default.** Replaces the previously specified Codex primary supervision default (documented foreground `fm-watch-checkpoint.sh` loop) with always-on Stop-owned `bin/fm-codex-stop-autoarm.sh` + `fm_supervision_model` → `autoarm`, registered in tracked `.codex/hooks.json` (async, timeout 28800). Checkpoint is retained as fallback, but unconfigured Codex primaries that pull this distro get a new default continuity path and cooperative turn-end guard behavior. Parity with Claude does not make replacing Codex's specified protocol into restore.

**VISION.md per-rule**
- One captain, one interface: **aligns** — fewer blind turns / permanent outages after Herdr restart; peace-of-mind for looking away.
- Authority is explicit and never inferred: **does not align (default)** — new always-on Codex Stop hook + model map without opt-in.
- Scripts own the mechanics, agents own the judgment: **aligns** — hook/scripts own re-arm; model diligence retired as primary path.
- A restart is a non-event: **aligns** — durable wake queue remains; delivery best-effort stated; checkpoint fallback kept.
- Delegation with a spine: **neutral**.
- The fleet outlives any vendor: **aligns** — harness-specific adapter; contracts to Codex Stop/async semantics measured on 0.154.0.
- Scope: **aligns** — supervision continuity at the command layer.

**Security FYI (tip only — not waiting-captain):** new async Stop hook can `codex app`-queue into the payload thread; one-time Codex hook-trust approval required (guard command left byte-identical to preserve existing trust hash). Scope/AFK/need/unbindable-rewake refusals mirror Claude sibling. Not malicious; residual is operator trust UX + delayed-healthy-beacon limitation the author documented. Do not Firstmate-flag on security alone while attestation/NM are red.

Merge-eligible **N** — attestation MISSING, NM FAIL, class new-default. outcome=waiting-author. Firstmate-flag=no. Please add a tip-bound no-mistakes attestation. Do not merge.

andy-n:
Withdrawing this — not pursuing it further. Thanks for the detailed review; your new-default objection was correct and I'm not going to leave it sitting open half-addressed.

PR #4448, opened 2026-09-14
fix(bin): restore config inherit primary catalog preflight

pedromuller-del:
Restores fm_config_inherit_primary_preflight removed during upstream sync while config push and remote inherit push still call it.


pedromuller-del:
Closing: branch was based on the pedromuller-del mirror, not upstream main; reopening with a scoped fix separately.

PR #4252, opened 2026-09-12
fix(composer): ignore Pi vim footer rows

anupamchugh:
## Summary
- Ignore Pi vim-mode status/footer rows inside the separated composer region.
- Preserve pending classification for real unsubmitted input.
- Add idle and done regression coverage.

## Validation
- `bin/fm-test-run.sh tests/fm-composer-lib.test.sh`
- `bin/fm-lint.sh` (ShellCheck passed; workflow lint could not run because actionlint is not installed.)

PR #4450, opened 2026-09-14
fix(fm-composer): reclassify idle Pi vimmode status row as empty, not pending

### LINKED ISSUE #3819: Pi doorbell silently suppressed: composer classifier returns pending for an idle agent with an empty composer, and the re-ring ladder retires the unrung record
## Symptom

A freshly seeded Pi secondmate on the herdr backend never received its routed work. The steer landed durably in the inbox, but the doorbell was never rung, the watcher's re-ring ladder never recovered, and the record escalated to `pending-reply-missed`. The mate sat idle with work waiting in its backlog and no indication anything was wrong on its own screen.

`fm-send` reported:

```
fm-send: doorbell skipped (composer visibly holds pending text); the steer is
durably recorded at state/learning-harness.inbox/001.msg and the watcher will re-ring
```

The composer was empty. The pane showed `─ INSERT 1:1 ─`, cursor at line 1 column 1.

## What was measured

With the pane idle and the composer empty:

```
$ (. bin/fm-backend.sh; fm_backend_composer_state herdr default:wPS:p2)
pending
```

The identity probe for the same pane returned `pi<TAB>idle`:

```
$ fm_backend_herdr_composer_identity default:wPS:p2
pi	idle
```

So `_fm_composer_pi_verdict` had an idle agent and still produced `pending`. Reading that function, the structural row scan short-circuits ahead of the status check:

```sh
state=$(_fm_composer_classify_pi_rows "$screen" "$styled")
if [ "$state" = pending ]; then
  printf 'pending'
  return 0
fi
case "$agent_status" in
  idle|done) printf 'empty' ;;
```

An idle or done Pi can therefore never reach the `empty` branch once the row scan claims pending, which is the documented intent ("Proven input remains pending for every live pi state") but leaves no floor when the row scan is wrong.

Verdict was stable across a full `fm-control.sh relaunch`, so this is not a transient render.

## Scope

Nine other idle Pi secondmate panes classified `unknown` rather than `pending`, and `unknown` does not suppress the ring, so only the affected pane was silenced. That makes this intermittent rather than fleet-wide, and correspondingly harder to notice: the failure is a mate that quietly does nothing.

Every idle Pi pane in this fleet renders an identical trailing `vim` row from the `pi-vimmode` extension, below the mode line and the stats row. That row is the obvious suspect, but stripping it from a captured screen did not flip the verdict in isolated re-tests, and re-running the classifier out of band produced `unknown` where the live path produced `pending`. The exact structural trigger is **not** pinned.

## Why it matters

The composer pre-check is documented as advisory, with the watcher re-ring as the safety net. In this case the net did not hold: the ladder escalated to `pending-reply-missed` and stopped, leaving routed work undelivered with no live-agent signal. A misread that is merely advisory on the first ring becomes terminal once the ladder gives up.

## Suggested direction

Treat a native identity of idle or done as a floor that the structural scan cannot override, or stop the re-ring ladder from retiring a record that was never actually rung. The second is the smaller change and fixes the class rather than this one shape.

## Environment

pi 0.84.4, herdr backend, `openai-codex/gpt-5.6-sol` at high, Calm on.
Extensions loaded: `@akepka/pi-cursor-cli-provider`, `context-mode`, `fm-branch-supervision.ts`, `fm-calm.ts`, `fm-primary-pi-watch.ts`, `fm-primary-turnend-guard.ts`, `pi-cursor-sdk`, `pi-vimmode`, `pi-web-access`.

## Workaround

`bin/fm-control.sh <id> relaunch` gets the mate moving, because a fresh launch reads its instructions from disk rather than waiting on the doorbell. It does not fix the next steer to that pane.

pablontiv:
## Intent

Captain's intent: convert the incident-driven Firstmate messaging-reliability backlog into shipped fixes now. Fix GitHub issue https://github.com/kunchenguid/firstmate/issues/3819 in this fork and open a PR through the normal contribution flow, with exact issue linkage (the PR must close #3819).

Issue #3819 substance: a freshly seeded Pi secondmate on the herdr backend never received its routed work. `fm_backend_composer_state` for an idle Pi pane (empty composer, `─ INSERT 1:1 ─` status line, agent status `idle`) returns `pending` instead of `empty`. Root cause in `_fm_composer_pi_verdict`: `state=$(_fm_composer_pi_classify_rows "$screen" "$styled"); if [ "$state" = pending ]; then printf 'pending'; return 0; fi` short-circuits ahead of the `idle|done` agent-status check, so an idle or done Pi can never reach the `empty` branch once the row scan claims pending — even though that row-scan behavior is documented intent for a genuinely pending composer. `fm-send` reports `doorbell skipped (composer visibly holds pending text)`, the steer lands durably in the inbox but is never rung, and the watcher's re-ring ladder escalates the record to `pending-reply-missed` with no recovery.

## What Changed

- Added `_fm_composer_pi_rows_are_idle_furniture` in `bin/fm-composer-lib.sh` to detect when the only non-blank content in a Pi separator pair is pi-vimmode's `─ INSERT y:x ─` status row.
- In `_fm_composer_pi_verdict`, when the row scan claims `pending` but the agent status is `idle` or `done` and the rows are only vimmode furniture, return `empty` instead of `pending`, fixing the short-circuit that previously prevented idle/done Pi panes from ever reaching the `empty` branch.
- Updated `docs/herdr-backend.md` to describe the new idle-furniture exception, and extended `tests/fm-backend-herdr.test.sh` to cover both `idle` and `done` native Pi status with a vimmode status row present, plus a regression case confirming real pending text alongside the status row still reports `pending`.

Closes #3819

## Risk Assessment

✅ Low: The fix is narrowly scoped (one new helper gated behind agent_status idle/done plus a strict furniture regex), matches the issue's stated root cause exactly, doesn't alter any other composer-state path, and is covered by tests that exercise the real composer_state entrypoint with realistic terminal output (both the idle/done-empty case and a real-pending-text regression case), not source-text matching.

## Testing

Drove fm_backend_herdr_composer_state live via the real code path (herdr.sh + fm-composer-lib.sh) with a scripted fake herdr CLI for the idle/done furniture-row case (now empty, was pending before fix), a genuine-pending-text adversarial case (stays pending), and a non-idle/done agent-status adversarial case (stays pending, guard doesn't over-fire); all 4 passed. Also ran the repository's own updated test file for this backend end-to-end with no failures.

- Live validation: ✅ go - 5 of 5 scenarios driven live against the product

| --- | --- | --- | --- |
| idle_pi_with_vimmode_status_row: idle Pi, empty composer with only the `─ INSERT 1:1 ─` furniture row reads empty (the reported bug) | ✅ pass | live | Standalone driver output: `PASS idle_pi_with_vimmode_status_row -&gt; empty`, plus single identity corroboration call (log grep count = 1) |
| done_pi_with_vimmode_status_row: done Pi with the same furniture row also reads empty | ✅ pass | live | Standalone driver output: `PASS done_pi_with_vimmode_status_row -&gt; empty` |
| idle_pi_with_real_pending_text_and_status_row (adversarial): a Pi with genuine unsent draft text alongside the status row must still read pending, not be swept into empty by the new furniture guard | ✅ pass | live | Standalone driver output: `PASS idle_pi_with_real_pending_text_and_status_row -&gt; pending`; mirrors updated tests/fm-backend-herdr.test.sh test_composer_state_pi_separator_real_text_is_pending which al… |
| working_pi_with_vimmode_status_row_stays_pending (adversarial): agent_status=working (neither idle nor done) with the furniture row must not be forced to empty, since the guard is scoped to idle\|done… | ✅ pass | live | Standalone driver output: `PASS working_pi_with_vimmode_status_row_stays_pending -&gt; pending` |
| Regression: full targeted test file for the herdr backend (composer-state, send-text-submit, transitions, wait-transition) still passes after the fix | ✅ pass | live | `bash tests/fm-backend-herdr.test.sh` completed with exit code 0, no `not ok` lines, including `ok - fm_backend_herdr_composer_state: native idle/done Pi ignores the pi-vimmode status row` |

## Pipeline

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
- Live validation: ✅ go - 5 of 5 scenarios driven live against the product

| --- | --- | --- | --- |
| idle_pi_with_vimmode_status_row: idle Pi, empty composer with only the `─ INSERT 1:1 ─` furniture row reads empty (the reported bug) | ✅ pass | live | Standalone driver output: `PASS idle_pi_with_vimmode_status_row -&gt; empty`, plus single identity corroboration call (log grep count = 1) |
| done_pi_with_vimmode_status_row: done Pi with the same furniture row also reads empty | ✅ pass | live | Standalone driver output: `PASS done_pi_with_vimmode_status_row -&gt; empty` |
| idle_pi_with_real_pending_text_and_status_row (adversarial): a Pi with genuine unsent draft text alongside the status row must still read pending, not be swept into empty by the new furniture guard | ✅ pass | live | Standalone driver output: `PASS idle_pi_with_real_pending_text_and_status_row -&gt; pending`; mirrors updated tests/fm-backend-herdr.test.sh test_composer_state_pi_separator_real_text_is_pending which al… |
| working_pi_with_vimmode_status_row_stays_pending (adversarial): agent_status=working (neither idle nor done) with the furniture row must not be forced to empty, since the guard is scoped to idle\|done… | ✅ pass | live | Standalone driver output: `PASS working_pi_with_vimmode_status_row_stays_pending -&gt; pending` |
| Regression: full targeted test file for the herdr backend (composer-state, send-text-submit, transitions, wait-transition) still passes after the fix | ✅ pass | live | `bash tests/fm-backend-herdr.test.sh` completed with exit code 0, no `not ok` lines, including `ok - fm_backend_herdr_composer_state: native idle/done Pi ignores the pi-vimmode status row` |

- `Standalone driver sourcing bin/backends/herdr.sh, calling fm_backend_herdr_composer_state against 4 scripted fake-herdr scenarios (idle+furniture, done+furniture, idle+real-pending-text, working+furniture)`
- <code>bash tests/fm-backend-herdr.test.sh (full targeted suite for this backend, includes test_composer_state_pi_separator_idle_is_empty and test_composer_state_pi_separator_real_text_is_pending) — exit 0, no `not ok` lines</code>
</details>

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>


pablontiv:
Superseded by #4252 - an existing external contribution that fixes the same root cause (idle Pi composer misclassified as pending on the herdr backend, issue #3819) with a narrower, more correct match (restricted to the literal origin caret position rather than any caret position, which correctly avoids a false-empty risk on a ghosted draft with a moved caret that this PR did not defend against). Closing this in favor of #4252; see the comment there. No further action needed on this PR.

PR #4463, opened 2026-09-14
fix(account-pin): refuse an empty Claude root on a skipped keychain answer

tiago-peixoto:
quota-axi 0.1.41 reports an empty Claude root as credential-present when its keychain presence check fails, and the account-pin preflight accepted it.

The preflight now accepts a skipped source with credentialPresent only when the root's .claude.json records a login (oauthAccount), which /login writes for file and keychain storage alike.

- New hermetic test tests/fm-account-pin-claude-preflight.test.sh replays the 0.1.41 answer: empty root and root without a login refuse, logged-in root passes. Fails without the fix.
- Live guard tests/fm-account-pin-preflight-live-e2e.test.sh failed before the fix and passes after (quota-axi 0.1.43).
- Real logged-in account roots on the verifying host still pass.
- docs/verification/dispatch-auth.md updated.


tiago-peixoto:
Opened against the wrong repository by mistake; moved to the fork.

PR #4471, opened 2026-09-14
fix(bin): make captain holds work on hosts with an older JSON::PP, and stop cleanup dropping accents from a held body

Marsjohn-11:
## Intent

The captain said "ship it" on a one-line fix a remote worker diagnosed and verified.

The defect: bin/fm-captain-hold.sh line 370 calls decode_json(<STDIN>). On hosts where JSON::PP is older (2.27202 was found on the Linux desk), allow_nonref defaults OFF, so a JSON-encoded bare scalar is rejected with "must be object or array".

The consequence is worse than one broken command. It blocks captain holds fleet-wide on any host with that older library: a remote worker there cannot formally record a decision for the captain at all - only mention it in passing in a status line, where it can be missed. That is how a real decision goes unrecorded.

The verified fix, from the worker that diagnosed it: JSON::PP->new->allow_nonref->decode instead of decode_json (or ->utf8->allow_nonref).

One correction that worker made to itself, so you do not over-scope: teardown is NOT blocked by this. Only hold on a task carrying a body is. Scout cleanup went through fine on that host.

The same one-line defect was then found at a second call site and folded into the same change deliberately, so both are in scope here rather than queued as follow-up: bin/fm-backlog-transition-lib.sh decodes a shown task body the same way when appending a finished work item's deliverable to a captain-held row. Its `/\A"/` guard reads as deliberate but a leading quote is exactly the bare-string case an older library rejects, so the guard selects for the failing input rather than protecting against it. Fixing only one of the two would leave captain holds broken on the same host through the cleanup path. Deliberately NOT in scope: any further hunt for decode_json call sites beyond these two. bin/fm-procevent-lavish.sh was checked and left alone because it matches a brace-delimited object before decoding.

Two things the captain asked to be stated plainly rather than left for a reader to assume. First, the verification limit: the mechanism is reproduced and tested, but neither fix is verified against a real JSON::PP 2.27202 host, because none is in the loop - this laptop runs 4.06, where the bug does not manifest at all. Second, the regression test must be written so it would have caught the original bug on any host: it forces allow_nonref explicitly off in the simulated case rather than passing only because this laptop's library is lenient. A check that passes because it could not detect the failure is the exact class of defect this fleet spent last night finding four of. Each half of the test was verified failing on its own unfixed call site with that site's real error message, and passing on the fix.

Nothing merges without the captain.

Later addition, decided at the review gate: review found a second, pre-existing defect on one of the lines this change already touches, and the call was to fix it here rather than queue it, because the sibling decoder in this very commit already carries the correct pattern and leaving one of two sibling decoders half-fixed would be worse. The cleanup decoder in bin/fm-backlog-transition-lib.sh printed decoded characters to a stream with no :raw layer, so a body containing a character at or below U+00FF emitted one latin-1 byte instead of the two UTF-8 ones. fm_backlog_retain writes that body straight back through --body-file, so a captain-held row containing an accented character lost it on cleanup with no error at all - silent data loss in the record path for captain decisions, which is the same failure mode this whole change is about. The fix applies the identical two lines the sibling already uses, binmode STDOUT ":raw" then utf8::encode. It was done TDD with a failing test first, asserting BYTES rather than decoded strings, and covering both rows the review measured separately, because any character above U+00FF makes perl print the whole string as UTF-8 and a single mixed body hides the latin-1 case entirely. Deliberately NOT widened: no sweep for other print-without-raw sites, no change to bin/fm-captain-hold.sh, and bin/fm-procevent-lavish.sh:471 stays untouched.

Note on how this arrived: the pipeline's own fix round for that finding reached its 30-minute wall-clock limit and the run ended with the branch released unchanged, so the fix was applied between runs at the same specification rather than by the fix round.

TEST STEP INSTRUCTIONS FOR THIS RUN, stated explicitly so they are visible rather than implicit.

Run exactly ONE suite and report its result, including how long it took:

bash tests/fm-backlog-atomicity.test.sh

Do NOT run tests/fm-captain-hold-lifecycle.test.sh in this run. Do NOT run any other suite. Do NOT build an independent adversarial before-and-after reproduction, a reverted-copy drive of the unfixed code, or a side-by-side comparison artifact.

WHY ONE SUITE, AND WHY THAT IS NOT REDUCED COVERAGE. This step has a 30-minute wall-clock cap. Measured in this step, tests/fm-captain-hold-lifecycle.test.sh takes 27.5 minutes and tests/fm-backlog-atomicity.test.sh takes 16.1 minutes (966.07s) - about 44 minutes together, so the pair can never fit one invocation in any order, parallel or serial. Four earlier runs of this branch died on that cap proving it. Both suites have already been executed in full by this pipeline's test step on this exact commit across earlier runs, at 52 cases and 99 cases with zero failures. This run re-executes the atomicity suite, which covers bin/fm-backlog-transition-lib.sh where one of the two fixes lives.

WHY DROPPING A SEPARATE REPRODUCTION WEAKENS NOTHING: the before-and-after verification is done and recorded, and it is no longer a simulation. Each fix was verified failing before and passing after against its OWN unfixed call site, both locally under a forced-off allow_nonref shim AND natively on a real JSON::PP 2.27202 host (perl 5.16.3), where all three defects reproduce with no shim at all and all three fixes hold. The simulation and the real library agree exactly, including error text. This pipeline's review step has also passed this exact head clean four times.

WHY THIS SUITE IS THE RIGHT ONE: bin/fm-backlog-transition-lib.sh carries the fm_backlog_retain body-decode fix, and tests/fm-backlog-atomicity.test.sh is the suite covering that library.

This is not a skip: the suite must actually run in full and must pass. If it does not pass, or the step cannot fit, report that plainly rather than narrowing further.

## What Changed

- `bin/fm-captain-hold.sh` and `bin/fm-backlog-transition-lib.sh` now decode a shown task body with `JSON::PP->new->utf8->allow_nonref->decode` instead of `decode_json`. A shown scalar field arrives as a JSON-encoded bare string, which `decode_json` rejects as "must be object or array" wherever the installed JSON::PP defaults `allow_nonref` off, so holding a task that carries a body and cleanup's retention of a captain-held row both failed on such a host.
- The cleanup decoder in `bin/fm-backlog-transition-lib.sh` also sets `binmode STDOUT, ":raw"` and `utf8::encode`s the decoded value, matching the sibling decoder. Without it a codepoint at or below U+00FF was printed as one latin-1 byte, so `fm_backlog_retain` wrote an invalid-UTF-8 body back through `--body-file` and silently lost accented characters.
- `tests/fm-captain-hold-lifecycle.test.sh` gains two cases: one forces `allow_nonref` off for every perl the command spawns (probing first that the simulation really does reject a bare scalar) and drives both body-reading paths, the other asserts a retained body's exact UTF-8 bytes using one row per character class, since any character above U+00FF would mask the latin-1 case in a mixed body. `docs/captain-hold-lifecycle.md` records what those cases pin and that both failures also reproduce natively on a fleet host carrying the older library.

## Risk Assessment

✅ Low: Two one-line decoder corrections plus a two-line byte fix, all matching an existing in-repo pattern, each backed by a regression case I traced as non-vacuous and pre-fix-failing, with no reachable same-class site left and no behavior change on modern JSON::PP.

## Testing

The single suite the intent named, tests/fm-backlog-atomicity.test.sh, ran in full and passed: 99 ok cases, zero failures, exit code 0, 16 minutes 37 seconds wall clock, transcript captured to the evidence directory. It drives the real backlog transition entrypoints against scratch homes, which is the surface bin/fm-backlog-transition-lib.sh serves and where one of the two decoder fixes lives, and it left the worktree clean. The captain-hold decode scenarios that pin the older-JSON::PP default and the retained body's UTF-8 bytes were not driven here, because their suite takes 27.5 minutes and the step has a 30-minute cap; they are reported untested with that reason. No visual artifact applies, since this change is entirely CLI and file-state behavior with no rendered surface.

- Live validation: ✅ go - 1 of 3 scenarios driven live against the product

| --- | --- | --- | --- |
| A user dispatches, completes, cleans up and reconciles backlog work through the real fm entrypoints, and every transition stays atomic after the body-decode change to fm_backlog_retain | ✅ pass | live | `bash tests/fm-backlog-atomicity.test.sh` - 99 ok, 0 not ok, exit 0, 16:36.56; transcript at ~/.no-mistakes/evidence/01M2HJWY18XRRT94VPFM3M9TY0/fm-backlog-atomicity.log |
| A remote worker records a captain hold on a task that carries a body where the installed JSON::PP defaults allow_nonref off, and the hold is formally recorded instead of being rejected as &#34;must be obj… | ⏸️ untested | no | The covering suite takes 27.5 minutes and this step has a 30-minute wall-clock cap, so it cannot run alongside the 16.1-minute atomicity suite the intent required; the author explicitly instructed thi… |
| Cleanup of a captain-held row whose body carries an accented character rewrites the body with the two UTF-8 bytes rather than one latin-1 byte, so the recorded decision text is not silently corrupted | ⏸️ untested | no | Same blocker as the captain-hold scenario: the case lives in the 27.5-minute lifecycle suite, which cannot fit this step&#39;s 30-minute cap next to the required atomicity suite, and the intent forbade ru… |

<details>
<summary>Evidence: Full fm-backlog-atomicity suite transcript with timing</summary>

Source: [Full fm-backlog-atomicity suite transcript with timing](https://github.com/Marsjohn-11/firstmate/blob/b072f665275759fe538abd7d0876e668ffb38820/.no-mistakes/evidence/fm/captain-hold-json-nonref/fm-backlog-atomicity.log)

```text
started_at=2026-09-15T05:22:38Z
ok - backend resolution preserves unreadable configuration errors for every caller
ok - backend resolution preserves environment, project, user, and default precedence
ok - backlog readers and mutations refuse unresolved backends without touching either backlog
ok - captain hold refuses backend errors and preserves relocated addressing for valid configuration
ok - dispatch publishes the record and moves the backlog item In flight in one run
ok - dispatch omits the markdown file when probing a Beads backlog
ok - a leftover markdown symlink does not brick a Beads home's dispatch
ok - completion applies and closes a Beads backlog without any markdown file
ok - dispatch refuses to supersede a pending authoritative close
ok - dispatch refuses held rows before creating resources
ok - dispatch refuses dependency-blocked rows before creating resources
ok - dispatch refuses held In-flight rows before relaunch
ok - dispatch reads backlog rows from the backlog addressing root
ok - recovery uses the parent of a trailing-slash data record
ok - completion targets nested relative data from the caller directory
ok - an immediate-child absolute data path keeps one paired backlog
ok - bare relative data addresses one backlog through dispatch and completion
ok - dispatch refuses symlinked backlogs without crossing homes
ok - automatic homes refuse lifecycle mutation without compatible tasks-axi
ok - dispatch refuses an unresolvable backlog data directory
ok - completion refuses before mutation when backlog data is unresolvable
ok - dispatch refuses, before creating anything, when the home has no item for the id
ok - dispatch distinguishes backlog read failures from missing items
ok - dispatch refuses a closed item instead of silently reopening it
ok - dispatch cannot commit without a verified task-record publication
ok - a failed backlog transition fails the dispatch loudly and leaves no record
ok - dispatch reports when failed-transition rollback cannot remove its record
ok - dispatch verifies both task and busy records during rollback
ok - dispatch commits neither record nor backlog state before launch delivery succeeds
ok - dispatch retries interrupted transitions before honoring termination
ok - a signal-deferred spawn verifies its preserved state and repairs a row that did not move
ok - a signal-deferred spawn reports attempted preservation as attempted when it cannot be verified
ok - a signal-deferred spawn bounds its verification so an unresponsive tasks-axi cannot hold the meta lock forever
~/.no-mistakes/worktrees/45849cd9dd00/01M2HJWY18XRRT94VPFM3M9TY0/bin/fm-backlog-transition-lib.sh: line 80: dirname: command not found
~/.no-mistakes/worktrees/45849cd9dd00/01M2HJWY18XRRT94VPFM3M9TY0/bin/fm-backlog-transition-lib.sh: line 80: ~/.no-mistakes/worktrees/45849cd9dd00/01M2HJWY18XRRT94VPFM3M9TY0/fm-timeout-lib.sh: No such file or directory
ok - fm_tasks_axi bounds the call through its perl watchdog when no timeout binary exists
~/.no-mistakes/worktrees/45849cd9dd00/01M2HJWY18XRRT94VPFM3M9TY0/bin/fm-backlog-transition-lib.sh: line 80: dirname: command not found
~/.no-mistakes/worktrees/45849cd9dd00/01M2HJWY18XRRT94VPFM3M9TY0/bin/fm-backlog-transition-lib.sh: line 80: ~/.no-mistakes/worktrees/45849cd9dd00/01M2HJWY18XRRT94VPFM3M9TY0/fm-timeout-lib.sh: No such file or directory
ok - fm_tasks_axi's perl watchdog passes the child status and output through unchanged
~/.no-mistakes/worktrees/45849cd9dd00/01M2HJWY18XRRT94VPFM3M9TY0/bin/fm-backlog-transition-lib.sh: line 80: dirname: command not found
~/.no-mistakes/worktrees/45849cd9dd00/01M2HJWY18XRRT94VPFM3M9TY0/bin/fm-backlog-transition-lib.sh: line 80: ~/.no-mistakes/worktrees/45849cd9dd00/01M2HJWY18XRRT94VPFM3M9TY0/fm-timeout-lib.sh: No such file or directory
ok - fm_tasks_axi fails closed rather than running unbounded when no bounding mechanism exists
~/.no-mistakes/worktrees/45849cd9dd00/01M2HJWY18XRRT94VPFM3M9TY0/bin/fm-backlog-transition-lib.sh: line 80: dirname: command not found
~/.no-mistakes/worktrees/45849cd9dd00/01M2HJWY18XRRT94VPFM3M9TY0/bin/fm-backlog-transition-lib.sh: line 80: ~/.no-mistakes/worktrees/45849cd9dd00/01M2HJWY18XRRT94VPFM3M9TY0/fm-timeout-lib.sh: No such file or directory
ok - fm_tasks_axi's GNU timeout kills a child that ignores SIGTERM after one further bound
ok - Kimi readiness interruptions fail before backlog commit
ok - dispatch does not resurrect a row closed after preflight
ok - dispatch fails when its backlog row vanishes after preflight
ok - completion closes a local-only ship, with its landing note, before reporting success
ok - completion closes a scout item against its report
ok - completion leaves legacy records open when no incarnation can be recorded
ok - completion refuses ambiguous task incarnations
ok - completion records relocated scout reports relative to the backlog root
ok - space-containing scout report paths round-trip through recovery
ok - control-byte data paths fail closed before paired transitions
ok - unreplayable data paths are refused before destructive cleanup
ok - completion preserves recovery state when task-record removal fails
ok - completion refuses to report success while its item is still open, and records what it owes
ok - restart recovers closes recorded before destructive cleanup
ok - completion refuses directory-symlink close targets
ok - completion reports failure until its durable close marker is removed
ok - session start retries a close whose marker could not be removed
ok - session start reports owned backlog rows it cannot read
ok - Orca cleanup recovery is excluded from backlog lifecycle transitions
ok - session start marks an item In flight when this home already owns a worker for it
ok - session start rejects internal worker-record symlinks
ok - session start rejects symlinked worker records
ok - session start finishes a close an interrupted cleanup recorded but never landed
ok - recovery backfills recorded links onto already Done items
ok - session start preserves a pending close across a transient backlog read failure
ok - recovery preserves incomplete-cleanup evidence across a failed replay
ok - session start finishes a close for the matching meta incarnation
ok - recovery preserves closes for ambiguous task incarnations
ok - recovery preserves both records when meta removal fails
ok - recovery preserves closes beside symlinked metadata
ok - recovery binds close-marker identity to its locked filename
ok - recovery rejects close markers targeting another home's data
ok - recovery validates an unterminated final marker field
ok - recovery rejects lexical traversal before resolving marker data
ok - recovery rejects marker control bytes before parsing
ok - recovery rejects malformed PR URL values
ok - a retained pending close is never started by reconciliation
ok - recovery rejects close-marker arguments outside its protocol
ok - recovery reports and rejects dangling symlink close markers
ok - session start drops a close recorded for an older meta incarnation
ok - session start rejects an unversioned close marker
ok - bootstrap rechecks worker-record containment after locking
ok - lifecycle files reject ancestor symlinks outside home roots
ok - same-home state overrides remain supported
ok - bootstrap refuses symlinked state before reconciliation
ok - bootstrap data-disappears fault injection is PATH-based; skipped on Darwin where stat is /usr/bin/stat
ok - bootstrap preserves secondmate, manual, and absent-backlog exemptions
ok - session start leaves a captain-held item where the captain put it
ok - no-backlog teardown refuses symlinked records before resource actions
ok - teardown rechecks record parents after locking
ok - teardown refuses symlinked state before resource actions
ok - a home with no backlog remains exempt from lifecycle transitions
ok - spawn refuses a special-file tasks-axi config instead of blocking on it
ok - spawn refuses an unsafe tasks-axi config before any exemption is derived from it
ok - spawn refuses a data directory symlinked outside the home
ok - a configured adapter refuses a data directory outside the home
ok - dispatch and completion transition structurally with evidence
ok - refused teardown leaves the backlog item in flight
ok - environment-selected adapters bypass the legacy markdown file override
ok - a manual-backlog home dispatches and completes without a hard failure
ok - a secondmate home keeps its own books paired through dispatch and completion
ok - dispatching a persistent secondmate needs no backlog item
bash tests/fm-backlog-atomicity.test.sh  197.84s user 250.49s system 44% cpu 16:36.56 total
exit_code=0
ended_at=2026-09-15T05:39:15Z
```
</details>
<details>
<summary>Evidence: Suite verdict and duration</summary>

```text
started_at=2026-09-15T05:22:38Z
... 99 ok cases ...
ok - a manual-backlog home dispatches and completes without a hard failure
ok - a secondmate home keeps its own books paired through dispatch and completion
ok - dispatching a persistent secondmate needs no backlog item
bash tests/fm-backlog-atomicity.test.sh 197.84s user 250.49s system 44% cpu 16:36.56 total
exit_code=0
ended_at=2026-09-15T05:39:15Z

ok: 99 not-ok: 0
```
</details>

## Pipeline

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>

- ℹ️ `bin/fm-captain-hold.sh:415` - Context worth recording, not a defect in this change. `show_field_value` assigns `value=$(decode_shown_value ...)` and never checks the status, so a decoder failure becomes an empty string that the function returns with success. That is exactly what happened on the older-JSON::PP host this change fixes: every quoted shown field (`hold_kind: &#34;-&#34;`, `hold_reason: &#34;-&#34;`) decoded to empty, so the guards at lines 1024/1459/1529 read hold_kind as empty rather than `captain` and would have let a still-held task through a merge refusal instead of blocking it. The fix at line 402 removes that trigger, so the blast radius was wider than the intent&#39;s &#34;blocks captain holds&#34; framing, in the fail-open direction. Sibling call sites that DO check (`write_hold_set_stamp` line 776, `write_resolution_record` line 924) fail loudly. Leaving this as-is is reasonable: hardening `show_field_value` would be pre-existing work outside this change&#39;s authorized scope, and no post-fix decode failure is reachable through intended usage (the `\&#34;*\&#34;` case only matches values tasks-axi quoted, which are always valid JSON).
- ⚠️ `tests/fm-captain-hold-lifecycle.test.sh:3984` - Simplification pass: the third assertion block in `test_retained_body_keeps_its_utf8_bytes` decodes all of data/backlog.md with `Encode::FB_CROAK()` to prove the whole file is valid UTF-8. The intent requires asserting the bytes of the two measured rows, and the two assertions above it already do that precisely - `grep -qF &#34;$accented&#34;` pins the C3 A9 pair and `! grep &#39;[\xe9]&#39;` pins the absence of the lone latin-1 byte, which is the exact corruption this fix removes (verified: unfixed output is `caf\xe9`). No intent requirement needs a file-wide UTF-8 validity claim, and it broadens what the case fails on beyond the defect under test. Recommend removing the perl block and keeping the two byte assertions.

🔧 Fix applied.
1 info still open:

- ℹ️ `docs/captain-hold-lifecycle.md:198` - The Verification record states &#34;No host carrying such a library is in the fleet, so that default is simulated rather than observed.&#34; The intent&#39;s later section contradicts that factual claim: it says each fix was verified &#34;natively on a real JSON::PP 2.27202 host (perl 5.16.3), where all three defects reproduce with no shim at all and all three fixes hold.&#34; The first half of the sentence is still true and worth keeping - the test case does force the older default rather than observe it - but the fleet claim now understates what was verified, so a future reader chasing this decoder would conclude the older-library behavior has never been seen on a real host and may redo that reproduction. Which statement is accurate is the author&#39;s call, hence ask-user rather than a silent edit; the remedy is a one-line wording change (for example, keep &#34;that default is simulated in the suite&#34; and note that the reproduction was confirmed natively on a JSON::PP 2.27202 / perl 5.16.3 host outside the fleet).

🔧 Fix applied.
1 info still open:

- ℹ️ `docs/captain-hold-lifecycle.md:198` - The sentence the last fix round wrote says &#34;all three defects reproduce on it natively with no shim&#34;, but at that point the section has named only two things - holding a task that carries a body, and cleanup&#39;s retention of a captain-held row. The third defect (the latin-1 byte loss) is introduced two lines later, so a reader reaching this sentence cannot resolve the count and may go looking for a defect the doc never names. The claim itself is right and matches the intent; only the number is unanchored where it appears. Smallest remedy is to say &#34;both&#34; here and let the byte paragraph stand on its own, or to move this sentence below the byte paragraph. Wording in the author&#39;s own doc, so it is their call rather than a silent edit.

🔧 Fix applied.
</details>

<details>

✅ No issues found.
- Live validation: ✅ go - 1 of 3 scenarios driven live against the product

| --- | --- | --- | --- |
| A user dispatches, completes, cleans up and reconciles backlog work through the real fm entrypoints, and every transition stays atomic after the body-decode change to fm_backlog_retain | ✅ pass | live | `bash tests/fm-backlog-atomicity.test.sh` - 99 ok, 0 not ok, exit 0, 16:36.56; transcript at ~/.no-mistakes/evidence/01M2HJWY18XRRT94VPFM3M9TY0/fm-backlog-atomicity.log |
| A remote worker records a captain hold on a task that carries a body where the installed JSON::PP defaults allow_nonref off, and the hold is formally recorded instead of being rejected as &#34;must be obj… | ⏸️ untested | no | The covering suite takes 27.5 minutes and this step has a 30-minute wall-clock cap, so it cannot run alongside the 16.1-minute atomicity suite the intent required; the author explicitly instructed thi… |
| Cleanup of a captain-held row whose body carries an accented character rewrites the body with the two UTF-8 bytes rather than one latin-1 byte, so the recorded decision text is not silently corrupted | ⏸️ untested | no | Same blocker as the captain-hold scenario: the case lives in the 27.5-minute lifecycle suite, which cannot fit this step&#39;s 30-minute cap next to the required atomicity suite, and the intent forbade ru… |

- <code>`bash tests/fm-backlog-atomicity.test.sh` - full suite, 99 ok / 0 not ok, exit 0, 16:36.56 wall clock (197.84s user, 250.49s system)</code>
- <code>`grep -c &#39;^not ok&#39; fm-backlog-atomicity.log` - confirmed zero failing cases in the captured transcript</code>
- <code>`git status --porcelain` - confirmed the suite left no transient files in the worktree</code>
</details>

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>


kunchenguid:
Speaking as Kun's firstmate: triage complete — this is a **restore** (captain-hold body decode on older JSON::PP + UTF-8 retain). Attestation MATCH on `8caaec7761806d13fea6934ed4fef135d40f0c82`, no-mistakes SUCCESS (run 34934229298), CI SUCCESS (run 34934181930 attempt 3; prior cancel by author re-run), MERGEABLE/CLEAN, safe review. First-time fork workflows approved after diff review. Auto-merging now.

kunchenguid:
Speaking as Kun's firstmate: this is merged. Thank you @Marsjohn-11 — really appreciate you taking the time on this.

PR #4500, opened 2026-09-15
fix(session-lock): find the harness across the Windows/POSIX process boundary

quansaritasa:
## Problem

Git Bash/MSYS/Cygwin's own `ps` only sees the POSIX-emulation process tree it manages itself. A harness that launches the shell as an ordinary Windows child process (for example Claude Code's own host process running bash.exe as a tool) sits outside that tree entirely, and `ps` reports a synthetic ppid of 1 the moment the ancestry walk would need to cross that boundary. Result: `fm-lock.sh` can never find the harness in the ancestry on Windows, and every session there is permanently stuck read-only (`error: cannot locate harness process in ancestry`).

## Fix

Add a Windows-native ancestry walk and liveness check backed by `Win32_Process` (via `powershell.exe`), keyed from the shell's own real Windows pid (`/proc/<pid>/winpid`, exposed under MSYS/Cygwin). It walks the real parent chain across the POSIX/Windows boundary that `ps` cannot see, normalizes the `.exe` suffix Windows process names carry before handing them to the existing classifier, and falls back to the previous `ps`-based walk only if the native path itself cannot run (own pid unreadable, or no `powershell.exe` on PATH) — never merely because it found no match.

`fm_harness_pid_alive` gets the same Windows branch, since a Windows pid stored in the lock file can't be resolved by `kill -0`/`ps -p` either (MSYS only recognizes its own POSIX-emulation pid numbering).

Both branches are gated on `uname -s` reporting `MINGW*|MSYS*|CYGWIN*`, so this is a no-op everywhere else (Linux/macOS CI unaffected).

## Verification

- Verified live against a real running Claude Code session on Windows: the walk now correctly resolves `claude.exe` in the ancestry, where it previously always failed.
- 4 new unit tests added to `tests/fm-session-lock-ancestry.test.sh` covering the Windows walk (finds the harness past `.exe` suffixes, stops at a non-harness gap same as the POSIX walk, liveness via `Win32_Process`, and fallback to the `ps`-based walk when `powershell.exe` is unavailable).
- Existing unit tests updated so they stay host-independent (a fake `uname` now pins them to the POSIX path even when run on an actual Windows/MSYS host, where they'd otherwise silently take the new Windows branch).
- Ran the full local test suite before and after this change on a real Windows/MSYS host: every pre-existing failure there (a set of e2e fixtures that rely on POSIX orphan-reparenting semantics Windows doesn't have) matches byte-for-byte between baseline and this branch, so nothing regressed.
- `shellcheck`/`bin/fm-lint.sh` could not be run locally (not installed in this environment); `bash -n` syntax-checked both changed files. CI's lint lane will need to confirm.

## Scope note

This does not attempt to fix the deeper, environment-specific issue where a script invoked as its own subprocess can have an already-exited immediate parent (Windows does not reparent orphans the way POSIX does, so that hop's identity is permanently unrecoverable once it exits) — that's a separate, narrower race tied to how a specific sandboxed tool spawns subprocesses, not a defect in this ancestry-detection logic itself.


quansaritasa:
Closing - not proceeding with this fix.

PR #4523, opened 2026-09-15
fix(spawn): strip agent co-author trailers in the task worktree

kkennethlau12393:
## What is wrong

Every supported worker harness injects its own commit attribution into the agent's session instructions, so the launch brief's rule against naming an agent as a co-author competes with the harness and loses.
Workers landed `Co-Authored-By: Claude Opus 5 <...>` and `Claude-Session: ...` trailers on commits whose brief forbade them, and one reached a project's default branch.
The existing per-launch `attribution` setting only disables Claude Code's automatic trailer; it does nothing about an agent that types the trailer itself.

## The fix

Enforcement moves below every harness, into git.

- `bin/fm-commit-trailer-lib.sh` (new) owns one constant list of agent products, generates the `commit-msg` hook, installs it, and removes it.
- The hook is bound to the task worktree alone, through the repository's per-worktree config (`extensions.worktreeConfig` + `git config --worktree core.hooksPath`), so it never reaches the captain's own checkout, the shared `.git/hooks`, or any tracked file in the project.
- An inherited hooks path or repository `commit-msg` hook is chained, never replaced: it runs first, its exit status propagates, and it is resolved at run time so a hook added later is still honoured.
- Only agent `Co-Authored-By:` lines and `<Agent>-Session:` lines are removed; a human co-author survives untouched.
- `bin/fm-spawn.sh` installs it once the task worktree is final, for every harness and backend at once, and refuses the spawn if it cannot.
- `bin/fm-teardown.sh` retires it while the local copy still exists and restores whatever hooks path that copy had.
- `bin/fm-brief.sh` keeps one line stating the rule, as the discovery layer.

A repository whose shared config sets `core.worktree` or `core.bare=true` without per-worktree config already enabled is refused with the two migration commands, rather than silently changing where those shared settings apply.

## Proof

`tests/fm-commit-trailer-lib.test.sh` spawns nothing: it builds a repository plus a linked worktree the way spawn does and asserts that agent trailers are stripped while a human co-author survives, that every listed agent product is caught case-insensitively, that an inherited repository hook still runs and still gates the commit, that a configured hooks path is chained and left unchanged, that the primary checkout is unaffected, that removal leaves no hook, no hooks path, and no rewriting behind, that a pre-existing worktree hooks path is restored, that reinstall is idempotent, and that a non-worktree path is refused.

`bin/fm-lint.sh` is green.
The changed-file test selection (60 scripts) has four failures, all reproduced unchanged on the base commit: the codex and muse process-ancestry checks, the herdr missing-adapter teardown preflight case, and `fm-backend`'s search for a historical permissive tmux kill adapter.

## Also in this branch

`tests/fm-lint.test.sh` wrote its temporary fixture into the repository as `tests/fm-lint-local-exclude-fixture.test.sh`.
The runner's coverage and lane-union checks glob `tests/*.test.sh`, so a concurrent `tests/fm-test-run.test.sh` saw a script that exists for a second and belongs to no lane and failed.
The fixture is now `tests/fm-lint-local-exclude-fixture.sh`, still inside `fm-lint.sh`'s canonical set.


kkennethlau12393:
Opened against the upstream by mistake; this change lands in our own fork.

PR #4567, opened 2026-09-15
Announce silenced open-decision replacements in the wake drain

prajwal-395:
PR 116 fixed the incident shape, not the class. Two decision keys at end-of-line now survive as their own records, but two keys mid-note still collapse in the default bucket with the survivor carrying the lost key as prose - and the listing shows one clean record with nothing anywhere saying a decision was discarded unread.

This change makes a genuine replacement of an open decision record announce itself. The fold itself is deliberately unchanged: two genuinely unkeyed decisions still collapse by documented design. The defect was the silence, so the fix is the announcement.

How it works: bin/fm-classify-lib.sh gains status_decision_replacements (plus scan_decision_replacements and a snapshot-bounded scan_decision_replacements_snapshot), which replays the stream through the exact same _fm_decision_fold_line rule the open-decisions folds use and fires when a step is about to return one record for a key that already had an open record and the two notes differ. bin/fm-wake-drain.sh prints those as a bounded DECISION REPLACEMENTS section beside OPEN DECISIONS on every presentation, including the empty-queue path.

Why this is not the shape PR 116 rejected: that worker rejected a generic misplaced-token detector because the existing prose fixture would flag on every drain as noise. This check keys on records being replaced, not on tokens being misplaced - no token position is inspected, so prose never creates a second open record to collide with and the objection cannot land on it.

Three-way proof in tests/fm-decision-replacement.test.sh (11 cases, all green): it fires on a genuine differing replacement (keyed and mid-note default), stays silent on a verbatim re-append, and stays silent on a prose mention of a key (working-line mention and resolved-line key quote). Also pinned: the marker clears when the survivor closes, reopen-after-close is a new decision, reserved-key foreign transitions the fold ignores announce nothing, and the snapshot scan agrees with the whole-file scan.

Open question from the filing, decided: two genuinely unkeyed decisions overwriting DOES announce. The mate's exact fire condition names no key exemption, the still-broken mid-note shape folds to default so exempting that bucket would leave the reported loss silent again, and one uniform rule with no carve-out is simpler to reason about. The fold still collapses them; only the announcement is new.

Validation: tests/fm-decision-replacement.test.sh green, plus neighboring suites green (fm-classify-decision-key, fm-classify-corr-token, fm-wake-drain-open-decisions, fm-wake-drain-open-decisions-cursor, fm-wake-drain-unread-status, fm-wake-drain-outcome-backstop) and bin/fm-lint.sh clean.


prajwal-395:
Closing: wrong base. This branch sits on the fork's port stack, so the diff against upstream main drags the whole stack along. The change itself is retargeted at fork main.

PR #4569, opened 2026-09-15
Add bin/fm-babysit.sh worker long-command completion wake helper

pikos-apikos:
Adds `bin/fm-babysit.sh`, a helper a worker wraps around a long-running command so its completion becomes a durable steering message instead of a polling sleep loop.

Behavior:

- Runs the child in the foreground of the invoking pane so the pane stays honestly busy, capturing stdout/stderr to a log file whose path is printed at start (fresh `/tmp` scratch file by default, `--log` for a caller-chosen path).
- On completion, kill, or timeout, appends one durable steering message through `bin/fm-send.sh` reporting command, status (completed, failed, signaled, timed-out, interrupted), exit code, elapsed time, log path, and a bounded tail (default 50 lines, byte-capped). The existing inbox doorbell then wakes the worker.
- Send home comes from `--fm-home` or `FM_HOME`; refused loudly when neither is set. One exact retry on delivery failure, then a single pane-visible fallback line with nothing else written.
- `--timeout` kills the child and its whole process group (SIGTERM, then SIGKILL after a grace) and exits 124 with no orphaned grandchildren; otherwise the child exit code is mirrored after the send attempt.

Also includes `tests/fm-babysit.test.sh` (11 cases, all green) and one worker-facing pointer in the ship brief scaffold. `bin/fm-lint.sh` clean.


pikos-apikos (commented):
Independent review: **blocking findings before merge**.

The direction is strong: turning long-command completion into a durable inbox event is a much better worker primitive than polling. I found three lifecycle/portability issues that should be fixed before merge.

1. **[P1] Default log creation is not portable to stock macOS `mktemp`.**

   The default path uses:

   ```sh
   mktemp "${TMPDIR:-/tmp}/fm-babysit-$TASK_ID.XXXXXX.log"
   ```

   BSD/macOS `mktemp` requires the replacement `X`s at the end of the template; the suffix-after-`XXXXXX` form is GNU-specific. Since Firstmate supports macOS and Linux, this can make the default invocation fail on macOS. Please use a portable form such as `.../fm-babysit-$TASK_ID.log.XXXXXX` and add a real macOS behavioral regression (the current stock-Bash lane parses the file but does not execute this suite).

2. **[P1] The timeout path does not actually guarantee “no orphaned grandchildren.”**

   After TERM + grace, the process-group KILL is conditional on `kill -0 "$CHILD"`. If the process-group leader exits on TERM but a descendant in that same group ignores TERM, `kill -0 "$CHILD"` fails and the group never receives SIGKILL. The main wrapper also kills the watchdog once the direct child exits, which can cancel the grace→KILL phase entirely.

   The existing orphan test uses a normal `sleep 60`, so it dies on the initial TERM and does not exercise this case. Please add a TERM-resistant descendant regression (for example, a grandchild with `trap '' TERM`) and make the KILL phase target the known process group independently of whether the group leader is still alive.

   There is also a contract/platform mismatch when `setsid` is unavailable: the implementation explicitly falls back to signaling only the direct child, while the PR description promises whole-process-group cleanup. Either provide an equivalent group-isolation path on supported platforms or narrow the documented guarantee.

3. **[P2] A single SIGINT/SIGTERM can leave the wrapper waiting indefinitely.**

   On the first interrupt, the trap records `INTERRUPTED` and sends TERM only. SIGKILL happens only after a second signal. If the direct child ignores TERM, the `wait` loop can remain blocked forever, so interruption is neither bounded nor guaranteed to produce the durable completion notification.

   I would reuse one bounded termination primitive for timeout and external interruption: TERM the owned group → wait grace → KILL the owned group → reap leader → send completion.

The 11 added tests are useful, but these failure modes sit just outside their current coverage. Also, the upstream CI / no-mistakes workflows are currently `action_required`, so there is not yet upstream CI evidence for this head.

With the portability fix plus adversarial process-lifecycle regressions, I think this is on the right track.

PR #4686, opened 2026-09-17
fm-pr-merge: give caller errors their own distinct messages

prajwal-395:
## Problem

`bin/fm-pr-merge.sh` reported three different caller errors - missing arguments, invalid task ID, malformed PR URL - with the same terse message used for genuine forge refusals: `error: invalid PR merge request`. The two are not the same event: a caller error means the arguments were wrong and should be retried correctly; a forge refusal is an external verdict that must never be retried around.

This cost real diagnostic time on 2026-09-16: a mate's first merge attempt failed on its own malformed arguments, that failure was indistinguishable from a refusal, and it fed a diagnosis that the harness was blocking merges. A real and separate defect existed alongside it, so the wrong reading was confirmed rather than caught.

### `bin/fm-pr-merge.sh`

Each caller-error path now emits `error (caller): ...` with the specific problem and the offending value, before any forge call is attempted:

- **Missing arguments**: names the expected usage
- **Invalid task ID**: echoes the bad value and names the expected pattern
- **Malformed PR URL**: echoes the bad value and names the expected URL formats
- **Unknown provider** (defensive catch): reclassified as `error (internal): ...` since it can only fire from a bug in the script itself

No refusal, network-failure, or authentication-failure path was changed. The four error categories - caller error, refusal, network failure, tool prerequisite - now carry distinct wording that can never be confused.

### `tests/fm-pr-merge.test.sh`

- Added `test_caller_errors_are_distinguishable_from_refusals`: regression test proving missing args, invalid task ID, and malformed URL each produce the `error (caller):` tag, never share wording with a forge refusal, and conversely that a real refusal (red checks) carries `refusing to merge` and never the caller-error tag.
- Updated the existing `test_malformed_url_refuses_before_merge` to expect the new self-describing message.

### `tests/fm-pr-check-security.test.sh`

- Updated the invalid-URL assertion to expect the new caller-error format.


prajwal-395:
Opened in error by automation against the wrong remote - this belongs on the prajwal-395 fork, not upstream. Closing immediately; apologies for the noise.

PR #4724, opened 2026-09-17
fix(bin): suppress false wedge alarms during active validation

bryceeppler:
## Intent

The away-mode supervisor keeps raising false wedge alarms on workers that are healthy.
During away windows on 2026-09-16 and 2026-09-17 it escalated "stale persisted ~245s (possible wedge)" for hzbook workers in panes wEF, wEG, wEH and again for wEK, while bin/fm-crew-state.sh reported "working, source: run-step, validating (running)" for each one and the pane showed a backgrounded no-mistakes axi respond or axi run shell.
Six separate escalations, every one a false alarm, each costing a firstmate turn that had nothing to do.
Stop the false alarms without losing genuine wedge detection.

## What Changed

- Add recent/quiet activity signals to working pipeline steps and classify only recent run steps as active validation.
- Defer stale-pane wedge alarms during active validation in both watcher modes, while resurfacing long-running validation on a bounded cadence.
- Preserve wedge detection for quiet, coarse, parked, failed, and overlong busy runs, with exact task marker tracking and expanded coverage.

## Risk Assessment

✅ Low: The change is well-bounded and matches the recorded user decisions exactly: the deferral is gated on a single-owner activity-recency fact with positive-evidence-only semantics, every marker lifecycle path drops the new marker alongside the stale one, the prior round's read/action task mismatch is fully corrected, and both directions (defers when moving, escalates on the ordinary schedule when quiet, parked/failed/busy-turn-bound unchanged) are pinned by executable regression tests.

## Testing

Following Prove It Works and Test Behavior, Not Implementation, I ran the three focused executable suites covering crew-state recency, away-daemon deferral, attended-watcher deferral, genuine wedge escalation, task-scoped markers, and shared re-surface cadence. All passed and the worktree remained clean. Live validation was blocked because the real no-mistakes CLI reported this worktree was not initialized, while this phase forbids initializing or starting a run. No visual artifact applies because the changed interface is shell supervision output and persisted state.

- Live validation: ⚠️ inconclusive - 0 of 5 scenarios driven live against the product

| --- | --- | --- | --- |
| A healthy validating crew remains quiet in away mode without raising a possible-wedge alarm | ⏸️ untested | no | A real validation run could not be started because this worktree is not initialized for no-mistakes and this phase forbids initialization and `axi run`. Provide a preinitialized isolated repository wi… |
| The attended watcher defers a quiet pane while its validation run reports recent activity | ⏸️ untested | no | The watcher requires a real attributable no-mistakes run to exercise this live. Provide a preinitialized isolated repository with an active validation run and a real supervised pane. |
| A recorded-running validation that stops reporting activity still escalates on the ordinary wedge schedule | ⏸️ untested | no | Driving this adversarial case live requires an active no-mistakes step whose activity can be suspended without ending its run record. This phase lacks that pre-provisioned run and forbids creating one… |
| A healthy task sharing a window identifier cannot suppress another task&#39;s genuine wedge alarm | ⏸️ untested | no | This requires two real supervised tasks with shared window metadata and distinct live-run states. No such isolated fleet was provided, and creating the required runs is forbidden in this phase. |
| Alternating liveness evidence produces only one bounded recheck for the same quiet window | ⏸️ untested | no | A live run would require controlling real validation activity and worktree writes across multiple cadence thresholds. The unavailable initialized pipeline prevents that product-level drive. |

<details>
<summary>Evidence: Live pipeline prerequisite check</summary>

```text
error: repo not initialized (run 'no-mistakes init' first)
```
</details>
<details>
<summary>Evidence: Focused behavioral test evidence</summary>

```text
ok - a live validation run step defers the wedge timer instead of escalating
ok - a run step recorded running whose own activity went quiet still escalates as a possible wedge
ok - the stale verdict reads the crew its own marker names, not a window sibling's
ok - a quiet window that alternates deferral kinds keeps one anchor and re-surfaces once per bounded cadence
```
</details>
- Outcome: ⚠️ 2 warnings across 1 run (18m44s)

## Pipeline

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>

- 🚨 `bin/fm-classify-lib.sh:1997` - The user approved this fix in review round 1 with explicit instructions, and it is NOT present at the target commit: crew_is_validating still accepts ANY `state: working · source: run-step` verdict as a live run with no activity-recency check. Concrete failing sequence: a no-mistakes step process is killed or hangs, but `axi status` keeps reporting `running`; bin/fm-crew-state.sh:811 maps `running|fixing` straight to `working · run-step · validating (running)`, and the coarse ledger path (bin/fm-crew-state.sh:757) maps a bare `running` row to `working · run-step · validating (background run)` with no steps table at all. Both new call sites then defer indefinitely. In bin/fm-supervise-daemon.sh:519 validating_defer rewrites `.subsuper-stale-&lt;key&gt;` on every tick, so the 240s wedge age never accrues again, and emits at most one non-wedge recheck per FM_PAUSE_RESURFACE_SECS (14400s) while resetting its own marker each time - the loop repeats forever and never reaches `stale persisted Ns (possible wedge)`. Genuine wedge detection for a crew whose run record is stale-but-`running` drops from ~240s to ~4h, indefinitely, and the FM_WEDGE_DEMAND_INSPECT_COUNT ladder is never reached even though its own message (bin/fm-watch.sh:1111) says &#34;do not re-absorb on the run-step/pane state alone&#34;. This directly contradicts the intent&#39;s required constraint &#34;Stop the false alarms without losing genuine wedge detection&#34;. Authorized remedy (user&#39;s round-1 instructions): emit bin/fm-crew-state.sh&#39;s existing nm_run_activity_is_recent (bin/fm-crew-state.sh:526, currently consumed only at :884) as a fact on the authoritative current-state line, require it in crew_is_validating, and do not re-derive recency anywhere else; keep the bounded resurface and the unchanged possible-wedge path for parked/failed/cancelled/terminal steps and panes with no live run step. Tests must cover both directions (recent activity defers; quiet activity escalates on the ordinary schedule).
- ⚠️ `bin/fm-watch.sh:1099` - The new live-run-step deferral is reachable from the busy-turn bound, which the change&#39;s own documentation says it must not touch. busy_turn_bound_check (bin/fm-watch.sh:1255) hands a genuinely busy pane past FM_BUSY_TURN_MAX_SECS to the same wedge_timer_check, so the new `crew_is_validating` branch here fires for it too. Concrete sequence: a crewmate backgrounds `axi run` (run record reads `running`), then its own harness hangs on a foreground tool call and renders a spinner for hours with no completed turn; previously busy_turn_bound_check escalated `possible wedge` once the bound plus FM_STALE_ESCALATE_SECS elapsed, now wedge_defer_validating absorbs it on the 14400s cadence with the wake text &#34;idle Ns, validating for Vs&#34;. This contradicts three stated contracts: crew_is_validating&#39;s own header (bin/fm-classify-lib.sh:1993) claims a busy pane &#34;keeps the unchanged wedge schedule&#34;; the watcher header (bin/fm-watch.sh:68) claims that past the bound &#34;every other pane goes through the same wedge timer and surfaces with the identical &#39;stale: ...&#39; reason&#34;; and docs/configuration.md:1121 documents that only a declared external wait or verified captain-held transfer takes the FM_PAUSE_RESURFACE_SECS recheck at that bound. It also weakens the exact detector whose rationale (bin/fm-watch.sh:1202) is &#34;a hung foreground call can hide behind a busy signature&#34;. No existing test covers busy-turn plus a live run step, because the busy-turn fixtures do not set a run-step verdict. This challenges deliberate scope: either gate the validating deferral to the non-busy stale path only, or, if the widening is intended, say so and update bin/fm-classify-lib.sh&#39;s narrowness claim, the bin/fm-watch.sh header, and docs/configuration.md:1121 to match, with a test pinning the new busy-turn behavior.
- ⚠️ `bin/fm-watch.sh:939` - Approved for fix in review round 1, still not applied. The watcher&#39;s two marker-anchored deferral chains are not mutually reset, so a quiet window that alternates deferral kinds reports a duration measured from the wrong chain. wedge_defer_wait calls clear_defer_tracking (clearing both), but wedge_defer_validating leaves `.writing-since-&lt;key&gt;` in place and wedge_defer_writing (bin/fm-watch.sh:912) leaves `.validating-since-&lt;key&gt;` in place. Concrete sequence inside one continuous quiet window with an unchanged pane hash (so no reset path fires): at T the threshold hits with a live run, creating `.validating-since` at T; the run ends and the pipeline&#39;s worktree writes take the next threshold down the write branch; a later threshold with a run live again re-enters wedge_defer_validating, reads the T-anchored marker, and computes vage as the whole elapsed window. The wake then reads &#34;validating for 14400s&#34; for a run that started minutes ago, and resurface_absorbed fires as soon as the `.validating-resurfaced` throttle is stale instead of after a fresh 4h. Smallest fix consistent with clear_defer_tracking&#39;s own comment (&#34;a long-finished one cannot make the next deferral resurface immediately&#34;): have each deferral drop the other chain&#39;s markers before anchoring its own.
- ℹ️ `bin/fm-classify-lib.sh:1982` - Approved for fix in review round 1, still not applied. The header&#39;s load-bearing narrowness claim is incomplete: it says fm-crew-state.sh reports `working · run-step` &#34;only for a running, fixing, or ci step&#34;, but the same block also emits it for an empty status (`run active`, bin/fm-crew-state.sh:819), for any UNRECOGNIZED status (`run active ($status)`, bin/fm-crew-state.sh:820), and for the coarse ledger fallback (`validating (background run)`, bin/fm-crew-state.sh:757). So an `axi status` reporting a status word this code does not know is silently absorbed onto the 4h cadence, which is the opposite of the file&#39;s stated fail-safe-to-escalate posture. The wider set is behaviourally defensible (all four are live-run shapes), so the mechanical fix is to correct the enumeration in the comment rather than narrow the predicate; leaving it as-is misrepresents the safety argument the next reader relies on.
- ℹ️ `bin/fm-supervise-daemon.sh:1121` - The busy (0) and exited (2) branches were changed from `rm -f &#34;$marker&#34;` to `stale_marker_remove &#34;$win&#34; &#34;$state&#34;` purely to also drop `.subsuper-validating-&lt;key&gt;`. That re-derives the key the loop already holds: stale_marker_remove recomputes it as `_stale_key(window_to_task(win))`, and window_to_task returns the FIRST meta whose window= matches, which need not be the task whose marker is being iterated when two metas share a window value - in that case neither marker is removed and the stale marker re-enters the loop on every housekeeping pass. The narrower, strictly-equivalent form is already two lines above in the window-gone branch (bin/fm-supervise-daemon.sh:1110): `rm -f &#34;$marker&#34; &#34;$state/.subsuper-validating-$key&#34;`. Using it here keeps the removal on the exact filename the loop iterated and drops the extra derivation. The same applies to validating_defer (bin/fm-supervise-daemon.sh:519), which recomputes `key=$(_stale_key &#34;$task&#34;)` although the caller already has `key` in scope.

🔧 Fix applied.
3 issues (1 error, 1 warning, 1 info) still open:

- 🚨 `indow_for_task() {  # &lt;task-key&gt; [state],+0p:1` - The fix-round commit 88054a9 added a tracked, empty file at the repo root literally named `indow_for_task() {  # &lt;task-key&gt; [state],+0p` - the debris of a malformed `sed`/`grep` invocation whose pattern text got consumed as a redirect target. `git ls-files` confirms it is tracked, and `git show --stat HEAD` lists it as an added path, so it ships in the PR. The name contains spaces, braces, parentheses, `#` and a comma, which is hostile to globbing, `find -exec`, and any tooling that walks the tree. Remedy: `git rm -- &#39;indow_for_task() {  # &lt;task-key&gt; [state],+0p&#39;` and amend/commit; nothing in the change references it.
- ⚠️ `bin/fm-watch.sh:918` - The new mutual reset between the two deferral chains removes each other&#39;s *anchor and throttle*, so a quiet window that alternates deferral kinds never re-surfaces at all - defeating the bound the change&#39;s own header promises at bin/fm-watch.sh:54 (&#34;Every one of those deferrals still re-surfaces once per PAUSE_RESURFACE_SECS&#34;) and at wedge_defer_validating&#39;s comment (&#34;a run that outlives any real one cannot stay invisible&#34;). Concrete sequence, one continuous quiet window, pane hash unchanged so no reset path fires and FM_PAUSE_RESURFACE_SECS=14400: at threshold T a live step reports `activity: recent`, so wedge_defer_validating (`:943`) anchors `.validating-since` at T. At T+240 the pipeline&#39;s step has gone quiet between steps, so crew_is_validating is false and control falls to crew_worktree_written_since (`:1131`), which is TRUE because the previous step wrote files inside the last window - wedge_defer_writing runs and its new `clear_validating_chain &#34;$key&#34;` (`:918`) deletes both `.validating-since` and `.validating-resurfaced`. At T+480 the next step starts and activity reads recent again: wedge_defer_validating&#39;s `clear_write_chain &#34;$key&#34;` (`:949`) deletes the write chain, and re-anchors `.validating-since` at T+480 with vage=0. Because resurface_absorbed gates on `[ &#34;$age&#34; -ge &#34;$min_age&#34; ]` using that per-chain age, reaching 14400s now requires 60 consecutive thresholds of one single deferral kind with zero intervening switches - which a real validation run that writes files essentially never produces. The pane is absorbed indefinitely with no bounded re-surface. This is NOT the pre-existing shape: before the change only the write chain existed and aged monotonically within a window, and the wait deferral&#39;s own bound is anchored on the worker&#39;s status-file mtime (which no deferral resets), so alternating with a wait still re-surfaces. Only the new validating&lt;-&gt;writing pair can starve both bounds. The smallest honest remedy is NOT to add hardening on top of the mutual clear: it is to anchor the bounded re-surface on the quiet window rather than on the deferral kind - one `.defer-since-&lt;key&gt;` marker created by whichever deferral takes the window and reset only where clear_defer_tracking already fires, with each deferral keeping its own throttle for wording. That changes the marker model the approved round-1 finding prescribed, so it needs your authorization: the remedy, not the defect, is what exceeds the settled decision. If you would rather keep the current shape, the alternative is to accept that an alternating window has no re-surface bound and say so in the two headers, which currently claim the opposite.
- ℹ️ `docs/architecture.md:12` - docs/architecture.md:12 enumerates what still counts as an actionable wake: &#34;provably-working stale panes that persist past `FM_STALE_ESCALATE_SECS` with neither a wait their own worker declared nor their own task worktree being written&#34;. That &#34;neither X nor Y&#34; is now a two-item list describing three exemptions - a crew on a live, still-moving run step is a third deferral and no longer produces an actionable wake on that path. AGENTS.md, the bin/fm-watch.sh header, and the bin/fm-supervise-daemon.sh header were all updated for the new deferral; this enumeration was missed, so a reader concludes a validating crew&#39;s quiet pane escalates when it does not. Mechanical sync: extend the clause to name the live-run-step deferral alongside the declared wait and the worktree write, matching the wording already used at docs/architecture.md:32 for the write deferral. No behavior change and no widening of the busy-turn-bound contract you asked to leave alone - docs/configuration.md:1120-1122 describe the busy path and stay correct as written.

🔧 Fix applied.

- ⚠️ `AGENTS.md:151` - AGENTS.md:151 is the &#34;watcher internals; never touch&#34; inventory, and after round 2 it no longer matches what the watcher writes. It still lists `.writing-*` and `.validating-*` - round 2 deleted both chains (`clear_write_tracking`/`.writing-since-*`/`.writing-resurfaced-*` and the short-lived `.validating-since-*`/`.validating-resurfaced-*`) - and it never gained `.defer-since-*` / `.defer-resurfaced-*`, the only per-window deferral markers the watcher now creates (bin/fm-watch.sh:1057, :1072, :919, :945). The daemon&#39;s `.subsuper-validating-*` is separately covered by the `.subsuper-*` entry, so nothing else covers the new files. Concretely: the anchor that carries the bounded re-surface is now a state file that no &#34;never touch&#34; glob names, so an agent tidying state/ by that list can delete it, restarting the quiet window&#39;s cadence from zero. bin/fm-watch.sh:404 has the same drift in the window_key enumeration (`.writing-*`, no `.defer-*`). Mechanical sync: replace `.writing-* .validating-*` with `.defer-*` in both places. No behavior change.
- ℹ️ `bin/fm-watch.sh:1078` - The `wedge_timer_check` header states &#34;Never re-reads the crew state (the costly check already ran once, at classification time)&#34; and then, eleven lines lower in the same block, &#34;the live run-step consult (crew_is_validating, one bounded crew-state read) ... run ONLY here.&#34; The second sentence is the truth: line 1114 calls `crew_is_validating &#34;$task&#34;`, which invokes `$FM_CREW_STATE_BIN` through `_crew_state_line` (bin/fm-classify-lib.sh:1929), a call the file elsewhere documents as &#34;NOT a pure read: fm-crew-state.sh may make a bounded no-mistakes call&#34;. On the declared-pause path at bin/fm-watch.sh:2619 this is the second fm-crew-state.sh invocation for the same window in the same poll, since `crew_absorb_class` already ran to produce the `working` case. The budget claim is still fine (at most once per window per STALE_ESCALATE_SECS), but the leading &#34;Never re-reads the crew state&#34; is now false and is exactly the sentence a future reader will rely on when deciding whether adding another consult here is free. Mechanical fix: drop or rewrite that clause so the header matches the consult list below it.
- ℹ️ `bin/fm-watch.sh:1072` - Round 2&#39;s instruction item 4 said &#34;Each deferral keeps its own -resurfaced-&lt;key&gt; throttle so the re-surface wording still names which evidence took the window.&#34; The implementation instead gives both deferrals one shared `.defer-resurfaced-&lt;key&gt;` (bin/fm-watch.sh:919 and :945), and `clear_defer_tracking` (:1072) removes it with the anchor. The comment at :1066-1069 argues explicitly against the instruction: &#34;which evidence took any given threshold is carried by the recheck&#39;s own reason, not by a second marker that would let an alternating window wake twice inside one cadence.&#34; That reading is defensible and resolves a real conflict inside your own instructions - item 8 asked for &#34;exactly one re-surface per FM_PAUSE_RESURFACE_SECS&#34; for an alternating window, which per-kind throttles cannot deliver (each kind would carry its own cadence, so an alternating window could wake twice per window). The stated purpose of item 4 (naming the evidence in the wording) is satisfied anyway by the reason strings &#34;validating right now&#34; / &#34;writing its worktree right now&#34;. Flagging only so you confirm the resolution you want, since it is a deliberate departure from an explicit build instruction rather than a defect.
- ℹ️ `bin/fm-supervise-daemon.sh:1127` - housekeeping&#39;s new deferral decision reads the crew state of a task it re-derived, while the comment added directly above it (bin/fm-supervise-daemon.sh:1121-1123) says exactly why that re-derivation is untrustworthy: &#34;two metas can share a window value, and window_to_task answers with the first of them&#34;. The loop iterates `.subsuper-stale-&lt;key&gt;` where `&lt;key&gt;` is task-scoped, resolves `win=window_for_task &#34;$key&#34;`, then sets `task=window_to_task &#34;$win&#34;` - which for two metas sharing a window returns the OTHER task. The marker removals were correctly switched to `$key`, but `crew_is_validating &#34;$task&#34;` at :1127 still uses the re-derived value. Concrete sequence: metas A and B both record `window=W`; the stale marker belongs to B; A is mid-run with `activity: recent`; B is genuinely wedged. `crew_is_validating` reads A, `validating_defer` fires for B&#39;s marker and resets `.subsuper-stale-B` every pass, so B&#39;s possible-wedge escalation never arrives - only a &#34;validating Ns&#34; recheck every FM_PAUSE_RESURFACE_SECS naming W. The same re-derivation already governs the `last_status_line`/pause routing two lines up, so the mis-attribution is pre-existing, but this change makes it govern wedge suppression for the first time. `window_for_task` (:1259-1266) already has the correct task in hand in its meta loop, so the mechanical fix is to have it yield the task alongside the window rather than round-tripping through `window_to_task`.

🔧 Fix applied.

- ⚠️ `bin/fm-supervise-daemon.sh:1117` - Round 3 moved the housekeeping loops&#39; READ onto the marker&#39;s own task but left the ACTION re-deriving from the window, so the two loops now decide about one task and act on another. At :1115 `last` is read from `$task` (the pair from `window_and_task_for_key`), but :1117 hands `$win` to `reconcile_pause_tracking`, which at :573 does `task=$(window_to_task &#34;$win&#34;)` - the exact re-derivation the comment added at :1121-1124 says is untrustworthy. Concrete failure under the premise this change itself asserts (two metas sharing a `window=` value W): task B has `.subsuper-stale-B` and appends `paused: waiting on upstream`; task A is genuinely wedged with `.subsuper-stale-A` aged past FM_STALE_ESCALATE_SECS. The loop reads B&#39;s paused line, calls `reconcile_pause_tracking W`, which resolves to A and runs `stale_marker_remove W` (deleting `.subsuper-stale-A` and `.subsuper-validating-A`) plus `pause_marker_record W` (writing `.subsuper-paused-A`). A&#39;s wedge timer is wiped and A is recorded as a declared wait it never declared, so A&#39;s possible-wedge escalation never fires - a loss of genuine wedge detection, which the intent marks as required. Pre-change both read and action used A, so they were at least consistent; the new harm is introduced by moving only the read. `.subsuper-stale-B` also stays and re-enters every pass. Line 1165 in the pause re-surface loop has the identical split (B&#39;s marker cleared through A). Smallest remedy is the pattern `validating_defer` already uses: pass the loop&#39;s own task/key into `reconcile_pause_tracking` instead of letting it re-derive, so the decision and the mutation name the same task.
- ℹ️ `bin/fm-supervise-daemon.sh:1130` - `crew_is_validating &#34;$task&#34;` is the first fm-crew-state.sh invocation housekeeping has ever made; the daemon&#39;s own classification comment at :439 calls out &#34;no fm-crew-state.sh call&#34; as a deliberate property. Each call can chain several no-mistakes invocations bounded at FM_CREW_STATE_NM_TIMEOUT (10s each: `axi status`, then `nm_runs_list`), and housekeeping runs them serially inside `fm_super_main`&#39;s loop at :1824. Concrete sequence matching the incident scenario: away mode with four-plus quiet crews (wEF, wEG, wEH, wEK) all crossing the 240s threshold in one tick while the no-mistakes daemon is unresponsive - each crew costs ~20s, so the tick stalls for over a minute before returning to the watcher-stdout read and the next tick&#39;s escalate flush at :1092. Durable wakes are queued to .wake-queue so nothing is lost, and a failed read reads not-validating and removes the marker so the stall is not repeated for those crews, which is why this is info rather than a blocker. Flagging because the remedy (bounding this call separately, or moving it off the main loop) would EXTEND the change rather than correct it, so it needs your call on whether the tradeoff is acceptable as-is.
- ℹ️ `bin/fm-watch.sh:1287` - Leftover wording from the two-chain design that round 2 replaced. The header says `clear_stale_hash_tracking` clears &#34;the shared quiet-window anchor and both re-surface throttles (clear_defer_tracking) plus the wait-deferral throttle&#34;, but `clear_defer_tracking` (:1072) removes exactly two files - `.defer-since-&lt;key&gt;` and the single `.defer-resurfaced-&lt;key&gt;` throttle - and the wait throttle `.waiting-resurfaced-&lt;key&gt;` is removed by the `rm -f` below it. There is no second deferral throttle any more; that was the point of the shared-throttle decision recorded at :1066-1069. As written, a future reader looking for the per-kind throttle this sentence promises will not find it. Mechanical: say &#34;the shared quiet-window anchor and its one re-surface throttle&#34;.

🔧 Fix applied.
1 info still open:

- ℹ️ `bin/fm-watch.sh:1073` - The write-deferral markers were renamed (.writing-since-&lt;key&gt;/.writing-resurfaced-&lt;key&gt; -&gt; .defer-since-&lt;key&gt;/.defer-resurfaced-&lt;key&gt;) with no path that removes the old names, so on any already-running home the legacy pair leaks permanently. Concretely: a home upgraded mid-quiet-window has .writing-since-&lt;key&gt; on disk from a prior write deferral; after this change clear_defer_tracking (bin/fm-watch.sh:1073) and the daemon&#39;s clear_pause_tracking (bin/fm-supervise-daemon.sh:565-570) both name only the new files, nothing else greps the old ones, and AGENTS.md:151 no longer lists .writing-* in the state inventory, so the &#34;never touch&#34; guidance no longer covers them either. No correctness impact - the files are never read, and the first deferral of the next window just re-anchors on .defer-since - but fm-watch.sh&#39;s own window_key header (:401) calls out orphaning markers rather than clearing them as exactly the hazard its single-derivation rule exists to prevent. Mechanical remedy: add &#34;$STATE/.writing-since-$1&#34; &#34;$STATE/.writing-resurfaced-$1&#34; to clear_defer_tracking&#39;s rm -f (and the watcher_key equivalents to the daemon&#39;s clear_pause_tracking) so the legacy pair is swept the next time the window&#39;s stale bookkeeping resets.
</details>

<details>
<summary>⚠️ **Test** - 2 warnings</summary>

- ⚠️ Live product validation was blocked because this gate worktree is not initialized for no-mistakes, while the assigned phase explicitly forbids `no-mistakes init` and `axi run`. Re-run with a preinitialized isolated repository and an already-running validation step, including authority to suspend its activity for the adversarial case.
- ⚠️ live validation verdict: inconclusive (0 of 5 scenarios were driven live against the product); untested: A healthy validating crew remains quiet in away mode without raising a possible-wedge alarm, The attended watcher defers a quiet pane while its validation run reports recent activity, A recorded-running validation that stops reporting activity still escalates on the ordinary wedge schedule, A healthy task sharing a window identifier cannot suppress another task&#39;s genuine wedge alarm, Alternating liveness evidence produces only one bounded recheck for the same quiet window
- Live validation: ⚠️ inconclusive - 0 of 5 scenarios driven live against the product

| --- | --- | --- | --- |
| A healthy validating crew remains quiet in away mode without raising a possible-wedge alarm | ⏸️ untested | no | A real validation run could not be started because this worktree is not initialized for no-mistakes and this phase forbids initialization and `axi run`. Provide a preinitialized isolated repository wi… |
| The attended watcher defers a quiet pane while its validation run reports recent activity | ⏸️ untested | no | The watcher requires a real attributable no-mistakes run to exercise this live. Provide a preinitialized isolated repository with an active validation run and a real supervised pane. |
| A recorded-running validation that stops reporting activity still escalates on the ordinary wedge schedule | ⏸️ untested | no | Driving this adversarial case live requires an active no-mistakes step whose activity can be suspended without ending its run record. This phase lacks that pre-provisioned run and forbids creating one… |
| A healthy task sharing a window identifier cannot suppress another task&#39;s genuine wedge alarm | ⏸️ untested | no | This requires two real supervised tasks with shared window metadata and distinct live-run states. No such isolated fleet was provided, and creating the required runs is forbidden in this phase. |
| Alternating liveness evidence produces only one bounded recheck for the same quiet window | ⏸️ untested | no | A live run would require controlling real validation activity and worktree writes across multiple cadence thresholds. The unavailable initialized pipeline prevents that product-level drive. |

- `no-mistakes axi status`
- `bash tests/fm-watch-triage.test.sh`
- `git status --short && find . -maxdepth 2 -type d \( -name '.playwright-cli' -o -name '.tmp*' -o -name 'test-results' \) -print`
</details>

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>


bryceeppler:
Hi maintainers - this PR is ready for review but its workflow runs are sitting in the "awaiting approval" state for a first-time contributor, so CI has never run on it.

Could someone approve the workflow runs and take a look when you get a chance?

Context: `fm-watch.sh` raised wedge alarms on tasks that were parked in an active no-mistakes validation run. The pane is legitimately quiet while the pipeline works, so the stale detector was firing on healthy work. This change teaches the detector to recognise an active validation run as a declared wait rather than a wedge.

Happy to rebase or adjust anything you'd like changed.

bryceeppler:
Closing unmerged: withdrawing this change on the author's decision. No further action needed.

PR #4799, opened 2026-09-18
fix(bin): answer Kimi 2.0.0 folder-trust dialog during spawn

Shazellb:
## Intent

fix the kimi trust stuff

Context for that ask: Kimi Code CLI 2.0.0 now shows a "Trust this folder?" dialog on every fresh worktree. Firstmate's spawn path does not answer it, so every Kimi worker launch fails before it ever receives its instructions. Kimi is currently unusable for dispatched work without a human pressing a key.

## What Changed
- `fm-spawn.sh` now answers Kimi 2.0.0's `Trust this folder?` dialog while waiting for Kimi to be ready. It reads only the visible pane, not the scrollback history. It presses Enter on the pre-selected `Trust this folder` option on every poll while the full dialog is still on screen. Spawn counts Kimi as ready only after two captures in a row show a ready signal and no dialog text. A blank capture proves nothing, so it restarts that count. A failed viewport read, or a dialog that never clears, now fails the spawn with a specific error message instead of a generic one.
- Adds `fm_backend_visible_capture` and `fm_backend_visible_capture_supported` to `fm-backend.sh`. The list of backends that support it is `FM_BACKEND_VISIBLE_CAPTURE`, and each has its own viewport read: tmux (`capture-pane -S -0`), herdr (`pane read --source visible`) and zellij (`dump-screen` without `--full`). cmux and Orca have no verified viewport read, so a Kimi spawn on either is refused at preflight, before the worktree or pane is created.
- Updates the Kimi harness reference (now verified against CLI 2.0.0: trust dialog, autonomy tiers, effort catalog), `docs/configuration.md` (the cmux/Orca refusal) and the herdr viewport-capture evidence in `docs/verification/runtime-backends.md`. Extends `tests/fm-kimi-harness.test.sh` to cover the trust-dialog handling.

## Risk Assessment

✅ Low: The round-5 fixes check out against the code. A viewport read that fails now fails readiness at once and names the backend. A blank frame costs one poll and restarts the two-capture ready count without clearing the trust diagnostics. The navigation hint's two parts are matched separately, so a wrapped hint in a narrow pane still counts. The list of supported backends lives in one place. The matching uses shell patterns instead of starting grep processes. The cmux and herdr notes now agree with the stated verification. Each new regression test drives the spawn command end to end, and the blank-between-banners test would fail if the ready-count reset were removed. The only behavior narrowing is that Kimi refuses to spawn on cmux and orca, which the user explicitly authorized.

## Testing

I ran the Kimi harness suite on the target commit and all 29 cases pass. It drives the real fm-spawn.sh against a fake tmux that shows the Kimi trust dialog. The spawn selects Trust (sending Enter again if a keypress is swallowed), confirms the dialog has cleared, and only then delivers the brief. It refuses up front on a backend that cannot read just the visible screen. The base commit's scripts fail the same fresh-worktree trust case, so the test reproduces the reported bug and the change fixes it. No real Kimi binary was available, so there was no live run. The temporary base copy was removed, and the worktree is clean.

<details>
<summary>Evidence: Kimi harness test transcript, target commit (all pass)</summary>

Source: [Kimi harness test transcript, target commit (all pass)](https://github.com/Shazellb/firstmate/blob/a70ee31c769bdd28f4b43342feeb090cb8b6d9c9/.no-mistakes/evidence/fm/fm-kimi-trust/kimi-harness-test.log)

```text
ok - Kimi hook install is idempotent and removal restores every foreign config byte
ok - Kimi hook removal preserves owned newline boundaries and pristine bytes
ok - Kimi hook install refuses missing, malformed, and surprising config without writing
ok - Kimi hook install refuses without jq before any config write
ok - fm-spawn: kimi launches, delivers its brief, and registers a guarded turn-end token
ok - Kimi hook stays silent and inert without a Firstmate registry token
ok - fm-spawn: unsafe Kimi global config refuses before pane creation
ok - fm-teardown: Kimi task pointer and registry token are removed
ok - fm-spawn: Kimi fallback expands the active HOME
ok - fm-spawn: missing Kimi executable refuses before pane creation
ok - fm-spawn: kimi treats a silent pointer drop as a failed spawn
ok - fm-spawn: kimi never sends the brief pointer before an observable ready signal
ok - fm-spawn: a fresh Kimi worktree answers the exact trust dialog once and verifies advancement
ok - fm-spawn: a swallowed Kimi trust keypress is re-sent until the dialog clears
ok - fm-spawn: a Kimi banner captured before the trust dialog paints does not read as ready
ok - fm-spawn: an answered Kimi trust dialog left in scrollback neither re-answers nor fails the spawn
ok - fm-spawn: a blank Kimi viewport frame costs its poll and nothing else
ok - fm-spawn: Kimi refuses a backend that cannot read the viewport, before launching
ok - fm-spawn: a Kimi trust dialog with its hint wrapped across rows is answered normally
ok - fm-spawn: a blank Kimi frame between banners restarts the two-capture ready count
ok - fm-spawn: a failed Kimi viewport read fails readiness with the backend named
ok - fm-spawn: Kimi refuses the ready verdict while trust dialog markers remain
ok - fm-spawn: a Kimi trust dialog must visibly clear before brief delivery
ok - fm-spawn: Kimi trust detection requires every observed dialog signal
ok - fm-harness: markerless kimi keeps its ancestry identity under an inherited marker
lock acquired: harness pid 1801404
ok - fm-lock recognizes Kimi ancestry and live lock holders
ok - busy detection: real Kimi moon-plus-middot captures require its harness while idle labels stay idle
ok - fm-watch classifies Kimi as unknown rather than from its spinner, and Grok's fallback stays isolated
ok - composer classifier: kimi's existing bordered > shape is already safe without an override
```
</details>
<details>
<summary>Evidence: Same trust test against base-commit fm-spawn.sh (reproduces the failure)</summary>

Source: [Same trust test against base-commit fm-spawn.sh (reproduces the failure)](https://github.com/Shazellb/firstmate/blob/a70ee31c769bdd28f4b43342feeb090cb8b6d9c9/.no-mistakes/evidence/fm/fm-kimi-trust/kimi-harness-test-BASE-before-fix.log)

`not ok - fresh Kimi trust dialog should advance into verified delivery: expected exit 0, got 1`

```text
ok - Kimi hook install is idempotent and removal restores every foreign config byte
ok - Kimi hook removal preserves owned newline boundaries and pristine bytes
ok - Kimi hook install refuses missing, malformed, and surprising config without writing
ok - Kimi hook install refuses without jq before any config write
ok - fm-spawn: kimi launches, delivers its brief, and registers a guarded turn-end token
ok - Kimi hook stays silent and inert without a Firstmate registry token
ok - fm-spawn: unsafe Kimi global config refuses before pane creation
ok - fm-teardown: Kimi task pointer and registry token are removed
ok - fm-spawn: Kimi fallback expands the active HOME
ok - fm-spawn: missing Kimi executable refuses before pane creation
ok - fm-spawn: kimi treats a silent pointer drop as a failed spawn
ok - fm-spawn: kimi never sends the brief pointer before an observable ready signal
not ok - fresh Kimi trust dialog should advance into verified delivery: expected exit 0, got 1
```
</details>
- Outcome: ⚠️ 1 info across 1 run (2m48s)

## Pipeline

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>
<summary>🔧 **Review** - 3 issues found → auto-fixed (5) ✅</summary>

- ⚠️ `bin/fm-spawn.sh:3293` - bin/fm-spawn.sh:3293 answers the Kimi trust dialog with exactly one Enter and never re-sends it, even though this same file models Kimi as swallowing early Enters: the brief pointer is submitted through fm_backend_send_text_submit with FM_KIMI_SUBMIT_RETRIES=3 (bin/fm-spawn.sh:4436) and the canonical success test runs with FM_FAKE_KIMI_SWALLOW_FIRST=yes (tests/fm-kimi-harness.test.sh:221), and the harness doc states that no rendering signal proves Kimi will accept input during the startup window. The trust Enter is sent inside that same startup window with no retry, so the authorized failure (Kimi worker launch fails before it receives instructions, needing a human keypress) stays reachable. Recommended boundary: re-send Enter on subsequent polls while kimi_trust_dialog_is_visible still reports the complete dialog - that is postcondition-gated, matches the fleet&#39;s submit-then-verify posture, and cannot fire once the dialog has cleared. This contradicts the deliberate, documented and test-asserted &#39;one-shot&#39; design (kimi.md &#39;The trust answer is one-shot&#39;, tests/fm-kimi-harness.test.sh:553 asserting exactly one Enter), so it needs the author&#39;s decision rather than a silent fix.
- ⚠️ `bin/fm-spawn.sh:3304` - bin/fm-spawn.sh:3304 accepts &#39;Welcome to Kimi Code!&#39; as a ready signal in the else branch, which is entered whenever kimi_trust_dialog_is_visible fails for ANY reason - including a mid-redraw capture that has painted only part of the dialog, or a narrow pane in which the exact hint row &#39;↑↓ navigate · Enter select · Esc exit&#39; (bin/fm-spawn.sh:3279) is wrapped or truncated. If Kimi&#39;s startup banner is on screen above the dialog (the normal TUI shape, and not excluded by the recorded verification), such a capture returns ready and the brief pointer is typed into the still-open dialog; the pointer is then lost and the spawn fails at the delivery gate. The agy sibling gate at bin/fm-spawn.sh:3476 already encodes this class of hazard (&#39;the native working verdict is known to coexist with an unanswered dialog&#39;) and refuses a ready-ish verdict until trust was pre-registered or the dialog was seen and answered; kimi has no pre-registration, so it accepts a pre-trust ready signal unconditionally. Cheap close: keep the strict four-signal test for deciding to press Enter, but refuse the ready verdict while any trust-dialog marker (e.g. &#39;Trust this folder?&#39; with &#34;Don&#39;t trust&#34;) is present. That leaves the decoy test&#39;s expected outcome unchanged.
- ℹ️ `tests/fm-kimi-harness.test.sh:596` - tests/fm-kimi-harness.test.sh:596 asserts that trust-enter.log is empty (no Enter was sent at an incomplete trust lookalike) but reports the failure as &#39;Kimi trust detection answered output missing the exact navigation hint&#39;, which describes a different condition and would mislead whoever debugs a future regression. Message should say that an incomplete trust lookalike was answered with Enter.

🔧 Fix: Retry Kimi trust Enter and gate ready on dialog markers
- ⚠️ `bin/fm-spawn.sh:3290` - The new ready gate is a same-capture test that requires BOTH &#39;Trust this folder?&#39; and &#34;Don&#39;t trust&#34; (kimi_trust_dialog_markers_present, bin/fm-spawn.sh:3290-3294), so the authorized failure stays reachable on a narrow window. Concrete path: kimi_capture runs `tmux capture-pane -p -S -120` (bin/backends/tmux.sh:42), which includes scrollback, and Kimi&#39;s static &#39;Welcome to Kimi Code!&#39; banner sits in scrollback ABOVE the Ink-managed dialog region - the change&#39;s own trust-partial fake (tests/fm-kimi-harness.test.sh:50) encodes exactly that banner-above-dialog shape. A poll at i=0 that lands after the banner has printed but before the dialog&#39;s first frame is painted (or on an Ink erase/redraw, or on a capture where only the box title row `╭─ Trust this folder? ─╮` has painted and the option rows have not) sees no marker at all, falls into the elif at bin/fm-spawn.sh:3322, matches the banner, and returns 0. fm_backend_send_text_submit then types the brief pointer into the still-open dialog; the pointer is lost and the spawn dies at kimi_wait_for_delivery. Two cheap tightenings: (a) block ready on ANY single trust marker rather than both - this is literally what the accepted fix instruction said (&#34;refuse the ready verdict while any trust-dialog marker is present&#34;), and the AND was chosen instead; (b) require two consecutive marker-free captures before returning ready, which closes the banner-before-paint race that no single-capture predicate can close. The durable boundary is pre-registering trust before launch, the shape this same file already uses for the two sibling harnesses with this exact problem (bin/fm-agy-trust.sh at bin/fm-spawn.sh:3669 and bin/fm-claude-trust.sh at bin/fm-spawn.sh:3662); Kimi already has a firstmate-owned surgical region in $HOME/.kimi-code/config.toml. This contradicts the author&#39;s documented AND choice (kimi.md: &#34;While `Trust this folder?` and the negative `Don&#39;t trust` option are BOTH still on screen&#34;), and the AND does buy real protection against dialog text lingering in scrollback after the answer, so it is the author&#39;s call rather than a silent fix.
- ℹ️ `bin/fm-spawn.sh:3319` - trust_markers_pending is set to 1 at bin/fm-spawn.sh:3319 and only ever cleared inside the complete-dialog branch (bin/fm-spawn.sh:3305), never in the else-branch path where the markers have gone away. Failure scenario: poll 1 captures a partially painted dialog (markers present, pending=1), then Kimi crashes and every later poll shows a bare shell prompt with neither markers nor a ready signal. The loop exhausts and bin/fm-spawn.sh:3335 emits &#34;&#39;Trust this folder?&#39; and the negative Don&#39;t trust option stayed on screen&#34; - a diagnostic that is simply false, pointing whoever debugs it at a trust problem instead of a crashed binary. Reset trust_markers_pending=0 in the else branch when kimi_trust_dialog_markers_present fails, so the flag describes the last capture rather than any capture.
- ℹ️ `bin/fm-spawn.sh:292` - The file-header contract comment at bin/fm-spawn.sh:291-293 still states that the launch-readiness loop &#34;selects the already highlighted affirmative option once&#34;. The fix round changed that to a re-send on every poll the complete dialog is visible (bin/fm-spawn.sh:3302-3315) and updated kimi.md accordingly, but left this header stale. A reader who trusts the header - it is the file&#39;s stated per-harness contract, the same place the agy and grok trust behaviours are recorded - will believe the trust Enter is one-shot and misjudge the Enter-storm blast radius of a stuck dialog. Update it to the retried-until-cleared wording.
- ℹ️ `tests/fm-kimi-harness.test.sh:638` - tests/fm-kimi-harness.test.sh:637-638 asserts the decoy output contains &#34;kimi did not show a verified ready signal&#34; and reports failure as &#34;incomplete Kimi trust lookalike did not retain the ordinary readiness failure&#34;. The trust-decoy pane carries both markers, so after the fix round it takes the trust_markers_pending branch (bin/fm-spawn.sh:3335), not the ordinary readiness failure the message names; the assertion only passes because the markers message happens to begin with the same prefix. Concretely, the assertion would keep passing even if the ordinary-readiness path were deleted outright, and the test now exercises exactly the same branch as test_kimi_partial_trust_dialog_blocks_the_ready_verdict. Either assert the markers-specific text and rename the message, or give the decoy a pane that lacks &#34;Don&#39;t trust&#34; so it genuinely reaches the ordinary readiness failure it claims to cover.

🔧 Fix: Gate Kimi ready on any trust marker and clean captures
- ⚠️ `bin/fm-spawn.sh:3340` - The new two-consecutive-clean-captures rule does not close the banner-before-paint race it was added for, because dialog_free_captures (bin/fm-spawn.sh:3340) counts every capture without trust markers, not consecutive captures that actually show the ready signal. Concrete sequence on a real fresh worktree: fm-spawn submits the launch command, then kimi_wait_for_ready polls while Kimi is still booting. Those early captures are the bare shell prompt - no &#39;Trust this folder&#39; text, and kimi_ready_signal_is_present is false (the existing not-ready test at tests/fm-kimi-harness.test.sh:544 proves the composer classifier does not call a shell prompt empty). So the else branch at bin/fm-spawn.sh:3336-3341 increments dialog_free_captures to 1, 2, 3... with no reset. When the first capture containing &#39;Welcome to Kimi Code!&#39; arrives, dialog_free_captures is already &gt;= 2, so the guard passes on that single capture and returns ready immediately - which is precisely the state the change&#39;s own comment (bin/fm-spawn.sh:3336-3339) says cannot be told apart from a pane whose dialog is one redraw away. The brief pointer is then typed into the dialog that paints a moment later and the spawn dies at kimi_wait_for_delivery, the exact authorized failure. The new test does not catch this: FM_FAKE_KIMI_TRUST=late (tests/fm-kimi-harness.test.sh:605) has the launch Enter set the fake state straight to banner-only, so poll 1 already carries the banner and the counter starts at 0 - the fake never emits the pre-launch shell captures that defeat the counter in production. Fix: reset dialog_free_captures whenever kimi_ready_signal_is_present is false, so the gate means two consecutive ready-and-clean captures, and give the late test a fake that emits at least one bare-shell capture before the banner.
- ⚠️ `bin/fm-spawn.sh:3314` - The change makes two assumptions about the same capture that cannot both be true. kimi_capture (bin/fm-spawn.sh:3262) is fm_backend_capture ... 120, i.e. `tmux capture-pane -p -S -120`, which includes scrollback. kimi.md&#39;s new text and the comment at bin/fm-spawn.sh:3292-3294 justify the trust_seen bypass by asserting that after the dialog is answered, &#39;dialog text left behind in scrollback&#39; is expected. But the four strings kimi_trust_dialog_is_visible requires (bin/fm-spawn.sh:3279-3283) all live in that one dialog block: if the block lingers in scrollback, all four linger together, kimi_trust_dialog_is_visible stays true on every subsequent poll, and the retry branch at bin/fm-spawn.sh:3314-3328 fires up to FM_KIMI_READY_POLLS (default 60) times - 60 stray Enters into an already-trusted, live Kimi composer - and then hard-fails the spawn with the false diagnostic &#34;kimi trust dialog did not clear after selecting &#39;Trust this folder&#39; on 60 poll(s)&#34; even though trust was accepted. The sibling agy gate (bin/fm-spawn.sh:3509-3513) avoids this by answering exactly once under AGY_TRUST_ANSWERED, reading the same 120-line capture; that one-shot guard is what makes scrollback persistence survivable there, and the kimi path deliberately removed it. No test covers a post-answer capture that still contains dialog text (the fake&#39;s `ready` screen has none), so the assumption is asserted in docs and untested in code. Either the residual-scrollback premise is wrong - in which case the trust_seen bypass at bin/fm-spawn.sh:3331 is unnecessary and only widens the window - or it is right and the retry loop storms. Earliest boundary that settles both: have the trust predicates read a scrollback-free capture (fm_backend_capture ... 0) so &#39;still on screen&#39; means the visible pane, and drop the trust_seen bypass. This contradicts the author&#39;s documented scrollback rationale, so it needs their call.
- ℹ️ `bin/fm-spawn.sh:3331` - bin/fm-spawn.sh:3331 disables the any-marker ready refusal as soon as trust_seen is 1, and trust_seen is set at bin/fm-spawn.sh:3315 the moment one Enter is sent - not when the pane is observed to have advanced. The accepted instruction was to lift the refusal only &#39;once the answer is confirmed by the pane advancing&#39;. Reachable gap: poll N sees the complete dialog and sends Enter, Kimi swallows it (the documented startup-window hazard this very change cites as the reason for retrying), the dialog stays up; polls N+1 and N+2 catch it mid-redraw so kimi_trust_dialog_is_visible fails; with trust_seen=1 the marker check at 3331 is skipped, dialog_free_captures reaches 2, and the banner above the dialog satisfies kimi_ready_signal_is_present - ready is returned with the dialog still open and the pointer is lost. Note this is the same predicate question as the scrollback finding: if residual dialog text does not in fact linger in the capture, the trust_seen condition can simply be dropped, which closes this path at no cost.

🔧 Fix: Read visible pane for Kimi trust and ready gates
- ⚠️ `bin/fm-spawn.sh:3274` - kimi_visible_capture asks fm_backend_capture for 0 lines, but 0 is a brand-new argument value that only the tmux adapter interprets as &#34;visible pane&#34; (bin/backends/tmux.sh:42 → `capture-pane -p -S -0`). Every other adapter ends with `printf &#39;%s&#39; &#34;$out&#34; | tail -n &#34;$lines&#34;`, and `tail -n 0` prints nothing: zellij (bin/backends/zellij.sh:484-493 — it even fetches the correct viewport via `dump-screen` for lines&lt;=40, then discards it), herdr (bin/backends/herdr.sh:3009-3017), cmux (bin/backends/cmux.sh:521-530); orca&#39;s `terminal read --limit 0` (bin/backends/orca.sh:199) likewise yields an empty tail. So on all four, bin/fm-spawn.sh:3275 falls back to kimi_capture — the 120-line scrollback read that kimi.md:26 and the comment at bin/fm-spawn.sh:3266-3271 say must never back a trust predicate. Concrete consequence on a zellij spawn: after Enter is accepted, the answered dialog block is still inside the 120-line history, all four strings kimi_trust_dialog_is_visible requires are still present, so the retry branch (bin/fm-spawn.sh:3326-3341) re-sends Enter into the live composer on every poll up to FM_KIMI_READY_POLLS (60) and then hard-fails the spawn with &#34;kimi trust dialog did not clear after selecting &#39;Trust this folder&#39;&#34; even though trust was granted — the exact regression this commit was written to prevent, and the exact user-visible failure the intent names (&#34;every Kimi worker launch fails before it ever receives its instructions&#34;). The comment&#39;s stated rationale is also factually wrong for zellij: its primitive can bound to the viewport; it is this repo&#39;s own `tail -n 0` that empties the result. Nothing in the new tests reaches this path — the fake tmux always returns non-empty for `-S -0`, so the fallback is never exercised. Fix needs a decision: either add an explicit visible-pane read per backend (zellij&#39;s `dump-screen` without --full is already it; herdr/cmux/orca genuinely cannot bound and would need honest per-backend documentation plus a different guard), or gate the kimi trust handling to backends that can prove a visible-only capture. Both change behavior on those backends, so it is the author&#39;s call.
- ℹ️ `bin/fm-spawn.sh:3275` - The `[ -n &#34;$visible&#34; ] || visible=$(kimi_capture)` fallback is unconditional, so it also fires on tmux whenever the visible pane happens to capture blank — an Ink clear-then-repaint frame between the dialog being torn down and the next screen being painted is the realistic case, and `$(...)` strips the trailing blank lines that would otherwise make the capture non-empty. That single poll then reads 120 lines of scrollback, where the just-answered dialog still sits, and either re-sends a stray Enter into a live composer or, on the last poll, leaves trust_still_visible=1 and fails an already-trusted spawn with the false &#34;did not clear&#34; diagnostic. An empty visible capture is absence of evidence, not evidence of a dialog: treating it as a skipped poll (advance i, sleep, continue) would keep the predicate honest on tmux. Note the remedy is coupled to the backend question above — making empty mean &#34;skip&#34; would leave the non-tmux backends never seeing the dialog at all.

🔧 Fix: Add per-backend visible-pane capture for Kimi trust gate
7 issues (5 warnings, 2 infos) still open:
- ⚠️ `bin/fm-backend.sh:756` - The new capability model asserts that cmux has no viewport-bounded read (`bin/fm-backend.sh:751-753` comment, `:756` refusal, `:738` supported-list), and `bin/fm-spawn.sh:2289` turns that into a hard preflight refusal of every Kimi spawn on cmux. The cmux adapter&#39;s own live-verified notes contradict the premise: `bin/backends/cmux.sh:44-49` records finding #3 as `read-screen --lines N` verified for N=1..10 with correctly clamped output, and `bin/backends/cmux.sh:515-520` states that &#34;a single read-screen call is still bounded by the surface&#39;s actual current viewport height regardless of the requested --lines value&#34;. `--scrollback` (passed at `bin/backends/cmux.sh:527`) is this adapter&#39;s explicit opt-in to history, not the CLI&#39;s only mode - so `read-screen --workspace W --surface S --json` without `--scrollback` is precisely the viewport-bounded primitive the refusal says does not exist. Failure scenario: a fleet on BACKEND=cmux that spawns Kimi today (worktree trust already remembered, readiness reached through kimi_capture) now exits 1 at preflight with &#34;backend &#39;cmux&#39; has no viewport-bounded capture&#34; and can never dispatch Kimi again, even though the backend can satisfy the new predicate. The refusal itself was authorized in the prior round, but it was authorized on this premise; correcting it restores a capability the intent wants (&#34;Kimi is currently unusable for dispatched work&#34;) rather than removing one, so it needs the author&#39;s call. Suggested resolution: verify `cmux read-screen` without `--scrollback` live, add `fm_backend_cmux_visible_capture` on that shape, and move cmux out of the unsupported arm and into the supported list - or, if the live check shows it is history-backed after all, correct `bin/backends/cmux.sh:515-520` so the two comments stop contradicting each other.
- ⚠️ `bin/backends/herdr.sh:3025` - `fm_backend_herdr_visible_capture` invokes `herdr pane read &lt;pane&gt; --source visible`, and `fm_backend_visible_capture_supported` (bin/fm-backend.sh:738) declares herdr supported on the strength of it, but nothing in this repo attests that `visible` is an accepted `--source` value. The only `pane read --source` shape used or verified anywhere is `recent`: `bin/backends/herdr.sh:3015` and `:3034`, `tests/fm-herdr-submit-confirm-live-e2e.test.sh:109`, and the CLI matrix in `docs/verification/runtime-backends.md:904` (&#34;`herdr pane read &lt;pane&gt; --source recent --lines N`&#34;) - which this change does not extend. kimi.md nonetheless states the primitive as verified fact under &#34;Verified on 2026-09-17&#34;. Failure scenario if the value is rejected (or the flag is spelled differently on the installed herdr): every poll&#39;s `fm_backend_herdr_cli` exits nonzero with stderr suppressed, `kimi_visible_capture` (bin/fm-spawn.sh:3280) swallows the status with `|| true`, the empty branch at `bin/fm-spawn.sh:3330` treats it as a skipped poll, and all 60 polls are spent before the spawn dies with the generic &#34;kimi did not show a verified ready signal before brief delivery&#34;. That breaks Kimi on herdr even for worktrees whose trust is already remembered and previously spawned fine - the exact user-visible failure the intent forbids, on a backend the repo exercises heavily. No test reaches this path; the fake tmux is the only backend exercised. Suggested resolution: verify `--source visible` against a live herdr and record it in the `docs/verification/runtime-backends.md` CLI matrix alongside `--source recent`, or drop herdr from the supported list until it is verified.
- ⚠️ `bin/fm-spawn.sh:3280` - `kimi_visible_capture` ends in `2&gt;/dev/null || true`, collapsing three different outcomes into one empty string: a transient blank viewport frame (rc 0, no output), a dead or unready pane, and a viewport read that failed outright (unsupported flag, missing session, target-ready failure in the herdr/zellij wrappers). The poll loop&#39;s empty branch (bin/fm-spawn.sh:3330-3336) is written only for the first case - &#34;absence of evidence&#34; that costs one poll. Failure scenario: any permanent read failure silently consumes all FM_KIMI_READY_POLLS (60 x 0.5s) and then reports &#34;kimi did not show a verified ready signal before brief delivery&#34;, pointing the operator at Kimi&#39;s startup rather than at the capture primitive that never returned anything - and the stderr that would have named it was discarded. Fix: keep the exit status (`local out rc=0; out=$(fm_backend_visible_capture ... ) || rc=$?`) and treat rc!=0 as a hard failure with a diagnostic naming the backend and the failed viewport read, reserving the skipped-poll path for a successful-but-empty read. That also converts the herdr risk above from a 30-second mystery timeout into an immediate, accurate error.
- ⚠️ `bin/fm-spawn.sh:3330` - The blank-capture branch resets `trust_still_visible` and `trust_markers_pending` but leaves `ready_captures` untouched, which contradicts the rule the code and kimi.md both state one branch below: &#34;any capture that is not ready restarts that count&#34; (bin/fm-spawn.sh:3362). Failure scenario - exactly the banner-before-paint race the two-capture guard was added for: poll N captures Kimi&#39;s banner before the dialog paints (ready_captures=1); poll N+1 lands on the Ink clear frame that precedes the dialog&#39;s first paint, so the viewport reads blank and the count is preserved; poll N+2 lands on the same banner still without the dialog (ready_captures=2) and `kimi_wait_for_ready` returns 0. The dialog paints a moment later, `fm_backend_send_text_submit` types the brief pointer into it, the pointer is lost, and the spawn dies at `kimi_wait_for_delivery` - the authorized failure. A clear-then-repaint frame is the realistic blank case, and `$(...)` strips the trailing blank lines that would otherwise make such a capture non-empty, so blank frames cluster exactly around redraws. Secondary effect: clearing `trust_still_visible` on a blank final poll downgrades a genuinely stuck dialog to &#34;kimi trust dialog was answered but the pane never advanced&#34; (bin/fm-spawn.sh:3377), dropping the answer count from the diagnostic. Fix: set `ready_captures=0` in the blank branch so the two captures must be adjacent, and leave `trust_still_visible`/`trust_markers_pending` as they were so a blank read really does &#34;leave the verdict where it was&#34; as kimi.md promises. `test_kimi_blank_viewport_frame_costs_only_its_poll` still passes under that change (its blank frame is followed by two ready captures).
- ⚠️ `bin/fm-spawn.sh:3298` - `kimi_trust_dialog_is_visible` requires the whole navigation hint `↑↓ navigate · Enter select · Esc exit` as one unbroken line. In a pane narrower than that row (~42 columns - a routine tiled/split pane width), Ink wraps the hint and the plain-text capture contains it as two lines, so `grep -Fq` on the full string fails. The dialog is then never answered: `kimi_trust_marker_is_present` still matches, the loop takes the markers branch on every poll, and the spawn fails with &#34;trust dialog text stayed on screen without the complete dialog&#34; - i.e. the authorized failure (&#34;every Kimi worker launch fails before it ever receives its instructions&#34;, needing a human keypress) stays fully reachable on a narrow pane. The code comment at bin/fm-spawn.sh:3303-3310 names this exact pane-too-narrow case, but only as a reason to withhold the ready verdict; it does not make the dialog answerable. No test covers a wrapped hint - the `trust` fake screen (tests/fm-kimi-harness.test.sh:47) always renders it on one row. Cheap close that keeps the predicate strict: require the hint&#39;s two distinctive tokens independently (`↑↓ navigate` and `Enter select`) instead of the whole row, so a wrapped hint still qualifies as the complete dialog while a banner or a title-only frame still does not. This narrows a detection rule the author deliberately made exact, so it is their call.
- ℹ️ `bin/fm-backend.sh:738` - `fm_backend_visible_capture_supported` hardcodes `tmux|herdr|zellij` while `fm_backend_visible_capture` (bin/fm-backend.sh:748-757) carries its own per-backend arms, so the two lists must be kept in sync by hand. Failure scenario: a sixth backend gains a `*_visible_capture` implementation and is added to the dispatcher but not to the predicate - `bin/fm-spawn.sh:2289` then refuses every Kimi spawn on a backend that can in fact read the viewport; the reverse omission passes preflight and fails on every poll with the generic ready timeout (see the `|| true` finding). Derive the predicate from a single list - e.g. one `FM_BACKEND_VISIBLE_CAPTURE_BACKENDS` string both read, or have the predicate test for the `fm_backend_&lt;backend&gt;_visible_capture` function after `fm_backend_source` - so the capability answer and the dispatch arm cannot disagree.
- ℹ️ `bin/fm-spawn.sh:3295` - `kimi_trust_dialog_is_visible` forks four `printf | grep` pipelines and `kimi_trust_marker_is_present` two more, so a poll costs up to six process pairs and the 60-poll loop up to ~360. The pane is already in a shell variable and every pattern is a fixed substring, so bash pattern matching does the same job with no subprocesses and no quoting hazard (the patterns are quoted, hence literal): `case &#34;$pane&#34; in *&#39;Trust this folder?&#39;*) ;; *) return 1 ;; esac` chained per marker, or a single `case` with the `|` alternation for the any-marker predicate. Non-functional; it also removes the `printf &#39;%s\n&#39;` re-serialization of the capture on every check.

🔧 Fix: Harden Kimi viewport capture and trust dialog detection

</details>

<details>

- ℹ️ Kimi Code CLI is not installed in this environment, so no live Kimi 2.0.0 spawn was run. The tests use a fake tmux that draws the dialog text the adapter expects: &#39;Trust this folder?&#39;, the &#39;↑↓ navigate · Enter select&#39; hint, &#39;❯ Trust this folder&#39; and &#34;Don&#39;t trust&#34;. The fix depends on those strings matching the real Kimi 2.0.0 dialog.
- <code>`bash tests/fm-kimi-harness.test.sh` on the target commit: all 29 cases pass, including the 12 new trust-dialog spawn cases (fresh dialog, swallowed Enter retried, banner shown before the dialog, answered dialog left in scrollback, blank frame, wrapped hint, partial markers, failed viewport read, refusal on a backend with no viewport read)</code>
- <code>Regression check: exported the base commit&#39;s `bin/` next to the new tests in a temporary copy and ran `bash tests/fm-kimi-harness.test.sh`. It fails with &#39;fresh Kimi trust dialog should advance into verified delivery: expected exit 0, got 1&#39;, which reproduces the reported bug</code>
- <code>Checked whether `kimi` is installed (it is not), so no live spawn was possible</code>

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>


kunchenguid (approved):
Speaking as Kun's firstmate: approving after tip security review (fork ruleset require_extra_approval_for_unattributed_changes). Tip c949cb9630d7227953d00695a7420199d45e8385 — CI/NM green; attestation MATCH; class restore; no .github/**; viewport-only trust Enter is bounded to complete dialog frames.

kunchenguid:
Speaking as Kun's firstmate:

First look on HEAD `c949cb9630d7227953d00695a7420199d45e8385`. Whole thread read (body + pipeline; no prior comments/reviews). Tip vs main `888871de` inspected (`fm-spawn.sh` Kimi trust readiness, `fm-backend.sh` viewport capture + tmux/herdr/zellij adapters, docs, `tests/fm-kimi-harness.test.sh`). No `.github/**`.

**Attestation:** MATCH — body `head_sha=c949cb9630d7227953d00695a7420199d45e8385` equals tip.
**CI/NM:** CI `35292976757` all SUCCESS; NM raise PASS. mergeState CLEAN.

**Contract-class:** restore — Kimi 2.0.0 folder-trust dialog broke the existing spawn readiness path on fresh worktrees; answering the pre-selected Trust option from a viewport-only read restores that concrete spawn contract. cmux/orca Kimi preflight refusal is fail-closed for backends without verified viewport capture (same readiness cannot complete there). Not new-default (no always-on observer/wake/Bearings; non-Kimi spawn unchanged).

**VISION.md (per rule):**
- One captain, one interface: **aligns** — spawn succeeds or fails with a named reason; no new captain noise.
- Authority is explicit and never inferred: **aligns** — only answers the already-highlighted Trust option; refuses backends that cannot confirm the dialog cleared.
- Scripts own the mechanics, agents own the judgment: **aligns** — dialog recognition and keypress are scripted; no agent judgment mixed in.
- A restart is a non-event: **aligns** — no durable-state change beyond spawn succeeding.
- Delegation with a spine: **aligns** — restores worker launch so the brief can be delivered.
- The fleet outlives any vendor: **aligns** — Kimi adapter tracks CLI 2.0.0; backend capability list stays explicit.
- Scope: **aligns** — harness/backend spawn mechanics; forge/CI untouched.

**Closes issues:** none linked.
**Outcome:** auto-merge (restore + MATCH + green CI/NM + safe tip). No Firstmate flag. No security FYI.

kunchenguid:
Speaking as Kun's firstmate: this is merged. Thank you @Shazellb — really appreciate you taking the time on this.

PR #4826, opened 2026-09-18
fix(bin): bound fm_lock_try_acquire's steal-chain recursion depth

data-driven-decision:
## Summary
- `fm_lock_try_acquire()` recursed into `fm_lock_try_acquire "$lockdir.steal"` with no depth bound; each `.steal` mutex only exists because a prior holder died mid-transaction (e.g. a `fm-control.sh` interrupt hitting the lock critical section), so under rapid relaunch/interrupt cycles the `.steal` suffix could grow across recursive calls until the path exceeded the OS filename-length limit and the shell segfaulted (exit 139). Reproduced twice live via `fm-captain-hold.sh complete jev-implementation-research --none` segfaulting, tied to `state/.meta-jev-implementation-research.lock` contention.
- Added `FM_LOCK_STEAL_MAX_DEPTH` (default 2, env-overridable) and an internal `depth` parameter to `fm_lock_try_acquire`; once the bound is hit, the function fails closed (`return 1`) instead of recursing into a deeper steal mutex, reporting the blocking pid via `FM_LOCK_HELD_PID` exactly as the other failure paths already do.
- No caller changes were needed: both call sites that matter (`fm-captain-hold.sh`) already use `fm_lock_acquire_wait`, whose retry loop handles a failed acquire gracefully.
- Existing self-held-reclaim, staleness, and steal-ownership-verification logic is unchanged; only the recursion depth is bounded.

## Test plan
- [x] Added 3 regression tests to `tests/fm-watcher-lock.test.sh`: a dead-owner steal chain exactly at the configured depth still reclaims (no regression), a chain one level deeper than the bound fails closed without creating a new steal artifact past the limit, and 10 concurrent processes racing an over-depth chain all fail cleanly with none crashing.
- [x] `bash tests/fm-watcher-lock.test.sh` - 35/35 pass
- [x] `bash tests/fm-wake-queue.test.sh` - 42/42 pass (no regressions)
- [x] `bash tests/fm-captain-hold-lifecycle.test.sh` - 52/52 pass (no regressions)
- [x] Manually verified against the pre-fix code that a 5-level dead-pid chain succeeds (rc=0, proving the prior unbounded recursion), while the fixed code fails closed (rc=1) on the same chain
- [x] `bin/fm-lint.sh` clean (ShellCheck 0.11.0)


kunchenguid:
Speaking as Kun's firstmate: triage of this PR.

**Outcome: waiting-author** (attestation MISSING; CI just approved and is pending)

**contract-class: restore** — `fm_lock_try_acquire` steal-chain recursion was unbounded and could grow `.steal` paths until OS filename limits / shell segfault (exit 139). Bounds depth (`FM_LOCK_STEAL_MAX_DEPTH` default 2) and fails closed. Corrective of broken lock reclaim; default steal behavior within bound unchanged.

**VISION.md**
- One captain one interface: pass — fewer silent lock crashes under relaunch contention.
- Authority explicit never inferred: pass — fail-closed; no new autonomy.
- Scripts own mechanics: pass — lock depth in wake-lib.
- Restart non-event: pass — lock reclaim still works within bound; over-depth fails closed for retry loops.
- Delegation with spine: pass.
- Fleet outlives vendor: pass.
- Scope: pass.

**Attestation:** MISSING — please add no-mistakes attestation for HEAD `d781fe2a91ca85f6ed949022f5ac79412cb961b1` (or re-run pipeline).
**CI/NM:** pending after first-time fork approve (35330699303, 35330699330) · **Draft:** no · **Mergeable:** MERGEABLE/UNSTABLE · **Firstmate-flag:** no

data-driven-decision:
Heads up on the required "PR must be raised via no-mistakes" check failing here.

This PR was opened from a fork (`data-driven-decision/firstmate`) because the `data-driven-decision` account that ran the fix doesn't have push access to `kunchenguid/firstmate`. We manually opened this cross-repo PR via `gh pr create` as a workaround.

That workaround structurally can't pass the required check: the check verifies a `## Pipeline` attestation section that only the real `no-mistakes` push remote writes into the PR body, and `no-mistakes status` on our side confirms its configured push target for this repo is `https://github.com/kunchenguid/firstmate` directly (not a fork-aware flow). So running `no-mistakes` from our side would hit the exact same missing-write-access wall, just inside the daemon instead of via a manual PR.

The fix (bounding `fm_lock_try_acquire`'s `.steal` recursion depth to prevent an unbounded-filename segfault under real lock contention) is implemented, tested, and committed on `fm/fix-lock-steal-recursion` - we're separately still investigating two CI test regressions (`fm-pr-check-security.test.sh`, `fm-secondmate-reconcile.test.sh`) that surfaced under the depth bound before this is fully ready anyway.

For the attestation/access gate specifically: the clean fix is giving `data-driven-decision` genuine write/collaborator access to `kunchenguid/firstmate`, so `no-mistakes` can push and open the PR directly with no fork involved. Let us know if that's workable, or if there's a different intended contribution path for this repo.

PR #4844, opened 2026-09-18
omp watch: drop stale replacement handoff records and close on first delivery

rafaelreis-r:
Fixes the stale-replay and unbounded-growth bug in `session-replacement-actionable.json`.

## Problem
The omp watcher extension re-delivered stale wakes of already-closed tasks at every session start. The session-replacement handoff store `state/extensions/omp-primary-watch/session-replacement-actionable.json` accumulated pending records (`stale` of panes that no longer exist, `signal` of tasks migrated to other homes) that were never consumed nor cleaned, so each new session replayed them.

## Solution
- Added `loadAndFilterReplacementHandoff()` that filters out handoff records whose backing task no longer exists (no state files, no queued wakes) and rewrites the handoff file.
- Added `isTaskAlive()` and `extractTaskId()` helpers to determine if a task is still live by checking for state files (.meta, .status, .inbox, .progress, .turn-ended) and the durable wake queue.
- Modified `activateOwnedWatch()` to use `loadAndFilterReplacementHandoff()` instead of `loadReplacementHandoff()`.
- `consumeWake()` already calls `finishPendingActionable()` which clears the handoff record on first delivery at `before_agent_start`/`message_start`, preventing replay on the next session start.

## Tests
- `test_watch_extension_drops_stale_replacement_records`: verifies stale records for ghost tasks (no state files, no queued wakes) are dropped at load time, while valid records with state files are preserved.
- `test_watch_extension_consumption_clears_handoff`: verifies the handoff is cleared after consumption at `before_agent_start`, and a subsequent actionable does not ride the replacement handoff.


rafaelreis-r:
Opened against upstream by mistake; captain instructed fork-only fix. Reopening on fork rafaelreis-r/firstmate.

PR #4849, opened 2026-09-18
feat(bin): carry a project's untracked local material into task worktrees

wesleykam:
A project's untracked local material now travels into every fresh task worktree, so a worker that needs the app running finds it already there.

## The problem

Firstmate clones a project from its forge, so the clone holds only what the forge tracks.
Every pooled worktree cut from that clone inherits the same absence, which means a worker that needs the application running finds no environment file, no stored browser session, and no local tool configuration.

The cost of that absence is not inconvenience. Observed on a live fleet, within a single day:

- A worker that found no environment file copied one from a sibling checkout, got the production one, and wrote rows into live data.
- A second task lost every one of its live-validation scenarios, because no server could start.
- A third spent half an hour of its step budget on setup and timed out twice.
- Three worktrees were unblocked by hand.

The live worktrees also showed the material is not uniformly absent but inconsistently hand-patched, which is worse: across the pool slots for one application no two held the same set of files and none held all of it. A worker cannot tell which slot it drew, so the same task succeeds or fails depending on which worktree the pool hands it.

Treehouse has no hook for this - its configuration is a pool size and a root, with no per-repo setup step and no untracked-file list - so the placement belongs to the spawn.

## What changed

A per-project manifest of untracked local material, placed into every fresh ship or scout worktree after the worktree reaches its clean base and before the worker launches.

- `bin/fm-local-material.sh` is the single owner of the schema, the placement, the refusals, and the operating rules the receiving worker is held to.
- `bin/fm-spawn.sh` calls it once the worktree exists, and puts those rules in the launch brief of every worker that receives material.
- `config/project-local-material.json` is local and gitignored, and secondmate homes inherit it.
- `docs/configuration.md` documents the schema, `docs/examples/project-local-material.json` is a worked example, and `docs/scripts.md` carries the toolbelt row.

A project the manifest does not name is a no-op, and no manifest exists in this change: it is gitignored local configuration that a commit cannot supply. An unconfigured fleet therefore behaves exactly as it did before.

## Link versus copy, and why both exist

The two modes are not interchangeable, and choosing the wrong one fails silently.

- **`link`** is a symlink to the source. One source of truth: a rotated key reaches every live worktree at once, and the secret exists in one place on disk rather than one copy per worktree. This is the right default for environment files.
- **`copy`** is a real copy with its **timestamp preserved**. Required wherever the consumer decides freshness by modification time or re-mints the file in place. A stored browser session is usually both: a Playwright global-setup re-mints the saved session once it is older than a max-age constant and decides that by mtime, so a timestamp-resetting copy would present an expired session as current, and a symlink shared across concurrent workers would race when one of them re-mints it.

## What is deliberately not carried

Anything a worktree must own separately.
A per-worktree dev-server port file is the standard example: ports are derived per worktree precisely so concurrent workers do not collide, so carrying one everywhere would reintroduce the collision.
Its absence is correct rather than a gap.

## The operating rules travel with the credentials

Credentials that arrive without rules for handling them are what produced the production write, so the rules are emitted into the launch brief of every worker that receives material, and only those workers:

- a dev server defaults to the configured staging environment;
- the production environment is used only when the captain names it, for the feature-shaped reason the manifest records;
- end-to-end tests and screenshots always run in staging, with no exception;
- these files are inputs to tools, never reading material - a worker may let a process consume one, but may not open, print, grep, or copy a value into a status line, commit, pull request, or report.

## Refusals

The manifest is explicit and per-project rather than a sweep of whatever sits beside the project.
A listed source that is missing refuses the spawn before placing anything, because a worktree that looks ready and cannot test is the state each of those incidents started from.
Also refused by name: an absolute path, a `..` component, a path resolving to the project root, a path git tracks, and a path the project does not gitignore - that last one would otherwise leave every worktree reading as having uncommitted changes, which blocks cleanup and invites a worker to commit a credential.

## Verification, and its limit

End-to-end proof that a freshly spawned worktree arrives equipped requires spawning a worker, which a test cannot do. That proof is still outstanding and should be done after this lands.

What is proven here, in two layers:

- `tests/fm-local-material.test.sh` pins the placement contract against temporary directories, including that a link tracks its source through a rotation, that a copy keeps its timestamp and stays independent, that a reused worktree has stale material replaced, and every refusal above.
- `tests/fm-spawn-local-material.test.sh` drives the real `fm-spawn.sh` against a fake terminal and a real pooled worktree, asserting what lands on disk, what the launch brief says, that the equipped worktree still reads as clean, and that a missing source refuses the spawn without publishing a task record.

Both suites are registered in `bin/fm-test-run.sh`, so the existing lanes pick them up.

## What was actually executed

**No repository checks have run on this head yet.** The check suites attached to it carry zero check runs between them. For a fork contribution every push re-arms GitHub's manual approve-and-run gate, so the checks stay queued until someone with write access releases them. An earlier head of this branch did run the full suite, but those results describe a commit that is no longer the tip, so they are not evidence about this one.

That earlier run is worth recording, because it found something. CI's canonical lint partition runs ShellCheck with extended dataflow analysis; the pre-push validation runs a changed-file mode that does not. `Lint 1` reported SC2034 in `tests/fm-spawn-local-material.test.sh`, where `read_case_record` bound the record's first field to a name nothing in the suite ever read. Every other job on that run passed. The last commit here fixes it, discarding the field with the `read -r _` form already used throughout `bin/`, and that fix is verified locally under the same analysis CI runs.

**The pre-push validation** ran review, test, document and lint against this head. Review, test and document reported no findings.

**Its workflow linter could not run.** `bin/fm-lint.sh` also invokes `actionlint`, and that binary is absent from that environment's `PATH`, so the step exited non-zero for that reason alone. Installing the pinned actionlint 1.7.12 with the repository's own `bin/fm-install-actionlint.sh <directory>` and putting that directory on `PATH` reproduces a complete local lint.

## Known limitation: a manifest edit followed by a relaunch

A worker can be told a file is the staging environment while the file actually sitting in its worktree is the production one.

It happens like this. Material is placed once, when a task's worktree is first created. A relaunch of that same task deliberately reuses its existing worktree and places nothing, because that worktree was already equipped. The instructions handed to the worker, however, are rendered fresh from the manifest every time, including on relaunch. So if the manifest is edited between a task's first spawn and a later relaunch of it, the worker is told about the manifest as it reads now, while the files on disk are the ones placed under the manifest as it read then.

Two things can go wrong from that, and they are the two this feature exists to prevent. If the edit changed which path is the default environment and which is production, the worker is told a file is staging when the copy in its worktree is the production placement, and it will act on that belief. If the edit added an entry, the worker is told a file is present that was never placed, and the instructions then send it looking for the file elsewhere rather than reporting a blocker.

Both need an operator to edit the manifest and then relaunch into an existing worktree. Neither is covered by a test.

Closing it properly means recording, at spawn, what was actually placed, and rendering the instructions from that record instead of from the current manifest. That is a change in what the feature does rather than a correction to it, so it is called out here rather than folded into this change. The alternative of skipping the instructions on relaunch is worse and is deliberately not taken: it would leave a relaunched worker holding live credentials with no rules attached to them at all.

## One unrelated commit

`test(secondmate): raise the concurrent config-push wait above its measured cost` is not part of this feature.
`tests/fm-secondmate-harness.test.sh` bounded a wait for a backgrounded config push at 2s; instrumenting that loop on an idle machine measured 4257 ms, 4920 ms and 6651 ms across three runs, so the bound sat below every honest case rather than merely tight under load, and the case failed here in a way that read as a lost delivery rather than an expired wait.
Only the ceiling moved. It is called out separately because it surfaced while validating this branch and is easier to review, or to drop, on its own.

## Pipeline

<details>

✅ No issues found.
</details>

<details>

- ⚠️ `.agents/skills/project-management/SKILL.md` - branch carries 2 commit(s) that exist on your local main branch but were never pushed to origin/main; these may be unintended bundled work (proposed PR changes 13 file(s)):
- 0eda4d5 feat(bin): carry a project&#39;s untracked local material into task worktrees
- b96965b test(secondmate): raise the concurrent config-push wait above its measured cost

Confirm these commits belong in this PR before approving, or manually separate the intended work onto origin/main before gating.
</details>

<details>
<summary>⚠️ **Review** - 1 warning</summary>

- ⚠️ `bin/fm-spawn.sh:3701` - `cmd_apply` (material placement) is skipped on relaunch (`[ &#34;$RELAUNCH&#34; -eq 0 ]` at bin/fm-spawn.sh:3701), but `cmd_brief_section` (bin/fm-spawn.sh:2602, inside the `[ &#34;$KIND&#34; = ship ] || [ &#34;$KIND&#34; = scout ]` block that is NOT gated on RELAUNCH) always re-reads the CURRENT config/project-local-material.json. If an operator edits the manifest (adds/removes an entry, or changes which path is `environments.default` vs `environments.production`) between a task&#39;s original spawn and a later relaunch of that same task - both ordinary operations on a hand-edited, gitignored, local file and a routine relaunch flow already exercised throughout fm-spawn.sh - the relaunched worker&#39;s brief will describe the NEW manifest (e.g. list a file as &#39;placed here for this task&#39; that was never placed, or label a path as STAGING/PRODUCTION differently than what was actually symlinked/copied into that worktree at its original spawn). This can cause the exact class of incident the feature exists to prevent: a worker told a credential file is staging when the on-disk file is actually the untouched production placement from the original spawn, or a worker told a file is present when it is not (so it goes looking for it elsewhere instead of reporting a blocker, per the brief&#39;s own instruction). Neither test file (tests/fm-spawn-local-material.test.sh, tests/fm-local-material.test.sh) exercises a relaunch after a manifest edit. The smallest honest remedy either contradicts a decision stated three times as deliberate (fm-spawn.sh:3710 &#39;A relaunch is deliberately excluded&#39;, docs/configuration.md:486 &#39;A relaunch reuses its recorded worktree as-is and places nothing&#39;), by re-running apply() on relaunch too, or requires new durable state (e.g. persisting the manifest snapshot/hash used at original spawn so brief-section can render what was ACTUALLY placed rather than the current manifest) - both are product-behavior decisions beyond a mechanical fix.
</details>

<details>

✅ No issues found.
- Live validation: ✅ go - 9 of 10 scenarios driven live against the product

| --- | --- | --- | --- |
| Fresh worktree receives configured material in the right mode (link vs copy) | ✅ pass | live | tests/fm-spawn-local-material.test.sh::test_spawn_places_the_configured_material_in_the_task_worktree (drives real fm-spawn.sh --scout against a real pooled worktree) plus manual CLI run showing wt/.e… |
| Launch brief carries the operating rules for placed material | ✅ pass | live | tests/fm-spawn-local-material.test.sh::test_the_launch_brief_carries_the_operating_rules asserts the real launch-brief.md written by fm-spawn.sh contains the staging-default, production-naming, and to… |
| Adversarial: production-only manifest (no configured default) still emits the PRODUCTION warning | ✅ pass | live | tests/fm-local-material.test.sh::test_production_only_environment_still_warns, plus manual brief-section run against a production-only manifest showing the PRODUCTION warning line and no STAGING claim… |
| Adversarial: a missing listed source refuses the entire spawn/apply and places nothing | ✅ pass | live | tests/fm-spawn-local-material.test.sh::test_a_missing_source_refuses_the_spawn (real fm-spawn.sh exits 1, no task record published) and tests/fm-local-material.test.sh::test_a_missing_source_refuses_w… |
| Unconfigured project spawns exactly as before (no-op) | ✅ pass | live | tests/fm-spawn-local-material.test.sh::test_an_unconfigured_project_spawns_unchanged - real spawn with no manifest produces no placement output and no local-material section in the launch brief |
| Adversarial: refusal-by-name boundaries (absolute path, .. escape, project-root path, git-tracked path, ungitignored path) | ✅ pass | live | tests/fm-local-material.test.sh::test_malformed_configuration_is_refused_by_name, ::test_a_git_tracked_path_is_refused, ::test_material_the_project_does_not_ignore_is_refused - all drive the real vali… |
| Reused pooled worktree has stale material replaced rather than merged with the previous task&#39;s leftovers | ✅ pass | live | tests/fm-local-material.test.sh::test_apply_replaces_material_left_by_a_previous_task - drives real apply against a worktree pre-seeded with a stale .env and a stale file inside a copied directory, co… |
| Equipped worktree still reads as git-clean (does not block teardown) | ✅ pass | live | tests/fm-spawn-local-material.test.sh::test_spawn_places_the_configured_material_in_the_task_worktree asserts `git status --porcelain` is empty after real placement; manual run corroborated this (only… |
| Shipped example manifest (docs/examples/project-local-material.json) actually validates | ✅ pass | live | tests/fm-local-material.test.sh::test_documented_example_validates_against_real_source_directories drives the real validate subcommand against the shipped example with its placeholder source paths rew… |
| End-to-end proof that a live-spawned worker actually finds the app running using placed material | ⏸️ untested | no | Requires spawning a real agent worker into a pool slot and having it start a dev server / run e2e tests, which this test phase cannot do (no worker-spawn authority here) - the user intent itself state… |

- `bash tests/fm-local-material.test.sh (19 cases, all pass)`
- `bash tests/fm-spawn-local-material.test.sh (4 cases, all pass, drives real fm-spawn.sh + real git worktree + fake terminal)`
- `Manual CLI drive: fm-local-material.sh apply/brief-section against a hand-built proj+wt fixture with a production-only manifest (no default env)`
- `Manual CLI drive: fm-local-material.sh apply against a fixture with a missing listed source, confirming refusal and zero placement`
</details>

<details>

✅ No issues found.
</details>

<details>

</details>

<details>

✅ No issues found.
</details>


wesleykam:
Related to #4333, which solves the other half of the same problem: that pull request carries environment **variables** from firstmate's own process into a spawned worker, while this one carries the untracked **files** a worktree needs. The two are complementary rather than competing - they touch different parts of `bin/fm-spawn.sh` and neither subsumes the other.

kunchenguid:
Speaking as Kun's firstmate: triage of this PR.

**Outcome: waiting-ci** (attestation MATCH; first-time fork CI approved this pass and is pending)

**contract-class: opt-in** — `config/project-local-material.json` is local/gitignored; unconfigured projects are a silent no-op on spawn (verified: apply/brief only run from manifest; no shipped fleet manifest). Not a new always-on default path.

**VISION.md**
- One captain, one interface: pass — equipping worktrees is below-deck; captain still talks to firstmate.
- Authority explicit never inferred: pass — captain must write the per-project manifest; no inferred sweep of sibling checkouts.
- Scripts own mechanics: pass — `fm-local-material.sh` owns schema/placement/refusals; spawn only calls it.
- Restart non-event: pass — material re-applied on fresh worktrees; reuse replaces stale.
- Delegation with spine: pass — missing source refuses spawn before publish; brief carries credential rules.
- Fleet outlives vendor: pass — forge-agnostic local material.
- Scope: pass — spawn/worktree equipping, not merge policy.

**Security FYI (cron):** Opt-in symlink/copy of env/session material into task worktrees; refusals for tracked/non-ignored/abs/`..`; brief forbids printing secrets. Diff reviewed; no withhold. Not Firstmate-flagged (CI pending).
**Attestation:** MATCH `0eda4d57f305a4894e03865f43b5b36785a9b981` · **CI/NM:** pending after approve · **Draft:** no · **Mergeable:** MERGEABLE/UNSTABLE · **workflow-zero:** yes · **Firstmate-flag:** no · **14d-stale:** no

PR #4868, opened 2026-09-18
fix(bin): bound concurrent ShellCheck processes across the whole machine

matthewstrud:
## Problem

An 8-core, 15 GiB host hung under a ShellCheck storm while several checkouts of this repo ran their suites at once.
The cause is multiplication, not any one number: `bin/fm-test-run.sh` runs up to 8 test files, `tests/fm-lint.test.sh`'s completeness sweep backgrounds 4 roots at a time, and `bin/fm-lint.sh` runs 2 workers.
Nothing at any layer knows how many other checkouts are doing the same on the same host.
Measured here: one ShellCheck of `bin/fm-spawn.sh` peaks at 1.06 GiB RSS, so the product exhausts memory well before it exhausts patience.

## Change

`bin/fm-lint-slot.sh` is the single owner of a whole-machine bound: it runs a command only while holding one of N per-user, per-host slots.
Both ShellCheck launch sites (the `fm-lint.sh` worker and the lint test's sweep) now launch through it.
No layer's own parallelism number changes.

Coordination available between independent processes in separate worktrees, started at different times, under one user: the filesystem and the kernel.
A slot is a non-blocking `flock(2)` on `/tmp/fm-lint-slots.<uid>/slot.<i>`, taken via perl (already a lint requirement; `flock(1)` is absent on macOS).
The helper then `exec`s the command with the locked descriptor inherited, so the command keeps the helper's pid (existing signal handling is unchanged) and the kernel frees the slot the instant that process dies, including on SIGKILL.
There is no lock file to go stale, no daemon, and no cleanup.

Default slots: half the online CPUs, at least 4 (one checkout's own peak, so a lone run never waits), capped at half the memory in GiB, never below 2.
`FM_LINT_SLOTS` overrides; `0` disables.

## Degradation, all toward running rather than hanging

- A slot holder is always the exec'd command and never waits for a slot; a nested call runs directly. Deadlock is structurally impossible.
- Coordination unavailable (no perl; slot directory missing and uncreatable, a symlink, not a directory, or not owned by this user; slot file unopenable): the command runs immediately, unbounded, with one loud `fm-lint-slot: ... running without the machine-wide bound` line.
- No slot frees within `FM_LINT_SLOT_WAIT_SECS` (default 600): the command runs anyway with the same loud line.
- Known limit, stated in the header: the bound covers one user on one host. Another user, or a sandbox with a private `/tmp`, holds separate slots.

## Measurements

Simulated worker = one full `bash tests/fm-lint.test.sh` in its own clone (real ShellCheck 0.11.0, includes the 408-root sweep and full-analysis partitions). 8 cores, 15 GiB. Sampled every 0.5 s: `pgrep -x shellcheck` count, summed ShellCheck RSS, `/proc/loadavg`. Ambient load1 on the host was 3-5 throughout from unrelated processes, so load figures carry that floor. The harness killed a run if load1 passed 11.

| run | workers | slots | peak concurrent shellcheck | peak shellcheck RSS | peak load1 | wall | result |
|---|---|---|---|---|---|---|---|
| alone, unbounded | 1 | off | 4 | 1.9 GiB | 4.3 | 293 s | pass |
| alone, default | 1 | 4 | 4 | 1.9 GiB | 7.0 | 297 s | pass |
| two, unbounded | 2 | off | **8** | 3.7 GiB | 6.4 | 312 s | pass |
| two, bounded | 2 | 4 | **4** | 2.9 GiB | 8.1 | 344 s | pass |
| three, bounded (first default tried) | 3 | 7 | 7 | 2.7 GiB | 11.1 | - | killed by the load guard at 260 s |
| three, bounded (shipped default) | 3 | 4 | **4** | 3.2 GiB | 9.2 | 402 s | pass, all three |

Reading:
- Unbounded, concurrent ShellCheck scales with worker count (4 -> 8). Bounded, it stays at 4 for one, two, and three workers.
- A lone worker is not slower: 297 s vs 293 s, inside run-to-run noise (a repeat lone bounded run took 299 s), because its own peak equals the bound and it never waits.
- The first default I tried (min of CPUs and memory/2 = 7) held its bound but still let three workers push load past the core count, so I tightened it to half the CPUs. That failed run is in the table on purpose.
- I did not run three or four workers unbounded. The 7-slot run already tripped the safety guard, and reproducing the hang was not worth the evidence.

## What this does not bound

In the three-worker runs load1 reached ~10 while only 3 ShellChecks were alive; the rest was the test scripts themselves plus ambient load.
This PR bounds lint, which is what the card scopes and what holds the memory.
`bin/fm-test-run.sh`'s up-to-8 test files per checkout is still multiplied by the number of checkouts, and the same slot helper pattern would apply there. That is a follow-up, not part of this change.

## Verification

- `tests/fm-lint.test.sh` gains seven behavior tests: the total is bounded across independently started processes (and both slots are actually used), status and output pass through unchanged, a SIGKILLed holder frees its slot immediately, a full slot set runs loudly after the bounded wait, unavailable coordination (non-directory, symlink) runs loudly and keeps the command's stderr byte-identical, a holder never waits for a second slot, and `fm-lint.sh` really launches ShellCheck through the bound.
- `bin/fm-test-run.sh --changed` (39 scripts): 37 pass. `tests/fm-calm-pi-extension.test.sh` fails because this host has no Chrome. `tests/fm-test-run.test.sh` fails identically on the unmodified base commit (same 24 passing assertions, same `comm: input is not in sorted order`), so it predates this change.
- `bin/fm-lint.sh` and `bin/fm-doc-audience-check.sh` pass.


matthewstrud:

PR #4891, opened 2026-09-19
fix(bin): archive unsupported PR links and harden watcher identity

yetihan:
Backlog close: tasks-axi records a --pr link structurally only for the
GitHub /pull/<n> shape, so a GitLab merge-request URL failed teardown's
close and every crash-recovery replay of it forever. An unsupported link
is now adapted to a --note that still archives the URL as row evidence,
keeping close and replay idempotent; supported links pass through
unchanged.

Claude Stop auto-arm: the failure-episode reset now rides out a transient
budget-lock publication from the synchronous turn-end guard with a
bounded retry (the same window the auto-arm ledger write uses) instead of
losing that race, failing the reset, and feeding the Stop-owned auto-arm
into an endless exit-2 rewake loop on an otherwise healthy home.

Watcher: the pid-identity publication into state/.watch.lock is
fail-closed (bounded identity-read retry, read-back verification, typed
refusal to supervise instead of an empty identity file), and the
singleton path replaces a live holder whose empty identity makes the
lock unauditable - TERM only after the holder's live command line proves
this watcher script, loud refusal otherwise - so a corrupt lock is
repaired by one resurface wake instead of yielding to an unauditable
supervisor forever.

Validated through the no-mistakes pipeline (intent, rebase, review,
test, document, lint); the review round added the Linux cmdline-hex
identity-form proof for the takeover guard, and the document step
recorded the retained-delivery gap for non-GitHub links in
docs/captain-hold-lifecycle.md.

## CI follow-up: lint fix and the portable-serial-4 watcher failure

Lint 2 (SC2034): tests/fm-backlog-atomicity.test.sh:2110 declared a dead
local `run`; the run that test calls is the function defined inside the
same test. The unused local is dropped.

Behavior portable serial 4, "taking-over watcher never recorded itself
as the lock holder": the failed run's log (35414428775) shows both
takeover announcements in the second watcher's captured output, so the
takeover verdict itself worked on the failing run - including the
cmdline-hex identity form this PR's review round added, which is the
form a Linux runner produces. The holder check then read the lock as
absent, and the log shape fits the second watcher exiting between
acquiring the lock and publishing its identity, its EXIT trap releasing
the lock on the way out. The only silent exits on that path are this
PR's fail-closed identity refusals, and the publish leg (write plus
read-back verify) was single-shot while the identity read already
retried on a bounded 5x0.1s budget - one transient publish failure would
refuse supervision where the next attempt would succeed. The publish leg
now uses the same bounded 5x0.1s retry as the read leg.

Re-run evidence on this branch (98fe91b8):

- tests/fm-watcher-lock.test.sh full suite: 10/10 green before the fix,
  10/11 green after it. The single miss is the simultaneous-start race
  below, not the takeover case.
- The CI-failing takeover case alone: 30 runs with the real-ps identity
  modality plus 30 runs with FM_PROC_ROOT_OVERRIDE forcing the Linux
  /proc hex-identity modality, all green before the fix; 10 more of each
  after the fix, all green. 80 runs, zero failures.
- Separately observed while re-running: test_singleton_start failed once
  in 21 full-suite passes across pre- and post-fix runs. The shape is
  consistent with a pre-existing timing race - a loser starting inside
  the winner's lock-creation window can read the lock directory before
  the pid file lands and emit the pid-less "already running" line, which
  the test's grep for 'already running pid ' cannot match. This PR's
  changes all run after acquisition and cannot influence that window;
  40 isolated runs of that test against this branch were 40/40 green.
  Left as follow-up work rather than expanded into this PR.
- bin/fm-lint.sh full run: clean.

The "PR must be raised via no-mistakes" check stays red by construction
for this fork-delivered PR; that merge path is decided outside this
branch.


kunchenguid:
Speaking as Kun's firstmate:

**HEAD** `da6448c6f05326e99c5b1ddbaf255c58a37989f8`. MERGEABLE/UNSTABLE vs main `1bb72cc5f880`. Author yetihan is not blocked. No `.github/workflows/*`.

**Attestation: missing** (PR body has no `<!-- no-mistakes-pipeline-attestation:v1 … -->` / `head_sha`). Tip CI/NM were `action_required`; **workflow-approved this pass** after diff review — runs `35414428775` / `35414428782` queued. Author still needs a current-head attestation before merge eligibility.

**Contract-class: restore.** Tip vs main: (1) unsupported non-GitHub PR links adapt to `--note` so backlog close/replay stops infinite failure — restores existing teardown close path; (2) Claude Stop auto-arm rides out transient budget-lock with bounded retry — restores existing auto-arm path; (3) watcher lock pid-identity fail-closed + verified-identity corrupt-lock takeover — hardens existing singleton watcher. No new default-on observer/wake/Bearings. Unconfigured GitHub PR links unchanged.

**VISION.md (per rule)**
- One captain, one interface: **aligns** — fewer silent forever-fail closes / endless rewake loops.
- Authority explicit never inferred: **aligns** — no new autonomy; refusal loud when unauditable.
- Scripts own mechanics: **aligns** — lock/identity and backlog adapt are scripted.
- A restart is a non-event: **aligns** — crash-recovery replay stays idempotent; lock healable.
- Delegation with a spine: **aligns** — retained URL evidence on unsupported links.
- Fleet outlives vendors: **aligns** — non-GitHub link retained as note; contracts not pixel-bound.
- Scope: **aligns** — command-layer reliability.

**Outcome: waiting-author** (attestation missing). Firstmate flag: no. Security FYI: corrupt-lock takeover TERMs only after live cmdline proves this watcher script + empty identity; unprovable holders refused loudly (not a captain choice while attestation still missing).

PR #4901, opened 2026-09-19
Fix opencode ladder free-cap detection for bare plugin model ids

prajwal-395:
Detection fix only: the free-capped gate now normalises the provider prefix on both sides of the model comparison.

Why: the retry sidecar's model comes from OpenCode's own event stream (model.id via reportModel in bin/fm-spawn.sh, passed through by recordRetry to bin/fm-opencode-retry.sh record), which carries no provider prefix. The gate compared it against the prefixed constant opencode/muse-spark-1.3-contributor-free, so the free-capped branch could never fire and the cap was silently counted as another rung's. Proven live on 2026-09-19: two lanes showed status=blocked horizon_s=69412 model=muse-spark-1.3-contributor-free with the real Free-usage banner while fm_opencode_ladder_free_capped answered not-capped.

What changed:
- bin/fm-opencode-ladder-lib.sh: new fm_opencode_ladder_bare_model helper (strips provider prefix); both evidence comparisons in fm_opencode_ladder_free_capped (sidecar model and idle-shape lane meta) normalise both sides through it. No second constant added. Fall-through bias, FM_OPENCODE_LADDER_OVERRIDE, and the off kill-switch are untouched.
- tests/fm-opencode-ladder.test.sh: two regressions speaking the plugin's bare vocabulary through the real record helper with run-time-relative horizons, asserted via the public gate: bare free cap falls through to Go (fails on the old code), bare Go cap leaves free alone (guards against collapsing the rungs, since only the -free suffix separates them).

Agy ladder checked, not changed: bin/fm-agy-ladder-lib.sh has no equivalent producer/consumer mismatch. Its evidence producer (fm_agy_quota_observe via fm_agy_footer_fields in bin/fm-agy-quota-lib.sh) records the pane-footer's display name under md5(display) key, and the consumer (fm_agy_ladder_state via fm_agy_quota_read, plus fm_agy_ladder_rung accepting kebab or display) reads the same vocabulary - one spelling family throughout, no prefix to strip.

Follow-up (out of scope, not changed): none noticed beyond the brief.

Verification: bin/fm-lint.sh clean; tests/fm-opencode-ladder.test.sh 16/16 pass; the new bare-free test fails on the pre-fix lib (stays on free) and passes with the fix.


prajwal-395:
Opened against the wrong base by mistake - this fork never sends PRs upstream. Re-opened on the fork; closing here. Sorry for the noise.

PR #4932, opened 2026-09-19
fix(bin): give each firstmate home its own treehouse worktree pool root

mehulbhagwani:
## Problem

Treehouse keys a worktree pool by the repository's resolved origin, not by the clone. Every firstmate home holding its own clone of one repository therefore allocated from a single shared pool: homes competed for the same numbered slots, and a returned slot could read "available" while its checkout was still a linked worktree of another home's clone. `bin/fm-spawn.sh`'s isolation assertion correctly refuses such a slot, so the task stops until someone releases it by hand from the home that actually owns it.

Reproduced directly before changing anything: two clones of one origin both landed in pool `proj-x-f23145`; after the first returned slot 1, the second was handed that same slot while its `.git` still read `gitdir: .../home-a/projects/proj/.git/worktrees/proj`.

## Fix

Give each home its own worktree root, derived from the home's own resolved path, and pass it through treehouse's documented global `--root` flag at both acquisition sites. `fm_treehouse_home_root` in `bin/fm-wake-lib.sh` is the single owner of the derivation.

Alternatives rejected, with reasons:

- **In-project pool (`--root .`)** - treehouse's own documented mode, zero derivation, but it puts worktrees inside the project clone where `git clean -xdf` would delete them, and for the secondmate seed the "repo" is the shared firstmate code root, so every home's worktrees would land in the primary checkout.
- **Root inside `$FM_HOME`** - a secondmate home is itself a treehouse worktree, so this nests a pool inside a git worktree and dirties the copy teardown's landed-work checks inspect.
- **Keying off the `.fm-secondmate-home` marker** - better names, but a marker can be rewritten or restored, and a root that moves orphans the pool it already allocated. Identity is the resolved path.
- **Making the shared project lock per-home** - now possible, but legacy pools stay reachable while they drain, and narrowing the lock would stop serializing exactly those returns.

## Existing pools

Nothing is stranded and nothing needs migrating. `treehouse return <path>` resolves its pool from the path it is handed and ignores the configured root, so worktrees already leased under the previously shared root stay returnable. Old pools simply stop being allocated from and drain as their worktrees return.

## Treehouse version floor

CI pinned treehouse 2.0.1, which has **no `--root` flag at all** - every spawn in the required Herdr lane would have failed. The pin moves to v2.3.0 with re-derived SHA-256 checksums for all four platform assets, and `bin/fm-bootstrap.sh` now probes the `--root` capability alongside the durable lease.

## Validation

Pipeline run `01M2WW2AD5DTBE2KK8S4TM42PW`, head `3dcbcf7371ea21e0d9b6e5f35596ac4a6b327d9c`.

| Step | Result |
|---|---|
| intent, rebase | completed |
| review | completed, **0 findings** across all 24 changed files |
| test | completed **with an approved exception** - see below |
| document | completed, **0 findings** |
| lint | completed, **0 findings** |

### The test step did not complete, and that is recorded honestly

The pipeline's Test agent **timed out at its 30-minute invocation budget without finishing**. Its own finding states this is "a budget or provider-slowness cut, not a code failure", and the step record reads: *"The Test agent exceeded its invocation budget before live validation completed; no evidence was gathered for this head."* The exception was approved deliberately rather than silently; re-running the same request costs another full budget and was judged unlikely to buy new information.

Because that step produced no evidence, the testing account below is **local evidence gathered directly**, not a pipeline attestation:

- `tests/fm-treehouse-home-root.test.sh` - 5 portable regressions, all passing. Each was verified non-vacuous by reverting the fix and confirming the assertion fails.
- `tests/fm-treehouse-pool-isolation-live-e2e.test.sh` - 5 scenarios against the real treehouse v2.3.0 binary, all passing: the shared-root collision, the foreign-slot handout, the interactive acquire form `fm-spawn.sh` actually sends, concurrent acquisition from per-home roots with no contention, and a legacy-root worktree still returning.
- `tests/fm-bootstrap.test.sh` - 28 tests passing, including new coverage that the lease and worktree-root probes gate independently.
- Real end-to-end: `bin/fm-spawn.sh` -> real treehouse -> real herdr -> `bin/fm-teardown.sh`, which created exactly the predicted per-home root.

A broader `--changed` suite run was **stopped at 87 of 138 tests** and is not a complete result. It had four failures, every one confirmed unrelated to this change:

| Suite | Cause |
|---|---|
| `fm-backlog-atomicity` | `timeout: command not found` - GNU coreutils absent; aborts that suite at case 33 of ~99 |
| `fm-secondmate-liveness` | pre-existing; reproduced on a pristine `origin/main` clone |
| `fm-calm-pi-extension` | no Chrome/Chromium installed, `FM_CHROME_BIN` unset |
| `fm-watch-triage` | exceeded its 900s bound under load (recorded CI weight ~698s) |

### The lint step could not verify twice before it did

Lint first reported "ShellCheck not found", then "actionlint not found" - both absent machine-wide, neither a code issue. Rather than approving an unverified gate, each missing tool was installed at its pinned version and lint was genuinely re-run. The final result is real: ShellCheck 0.11.0 (pinned) and actionlint 1.7.12 (pinned), **0 findings**.

## Out of scope

Changing treehouse itself; redesigning the pool layout beyond per-home separation; migrating existing worktrees. No migration is required - see "Existing pools" above.

## Unrelated pre-existing issues found, not fixed here

- `fm_backend_herdr_agent_state` falls through on an `unknown` pane state to `fm_backend_herdr_server_running_state`, which `tests/fm-secondmate-liveness.test.sh` does not stub - so it reaches the real `herdr` binary and returns `missing` instead of `unreadable`. That test therefore only passes on machines without herdr installed.
- `tests/fm-calm-pi-extension.test.sh` hard-fails on a missing browser rather than capability-skipping like the other tool-gated suites.


mehulbhagwani:
Bug report with reproduction: kunchenguid/firstmate#4977

mehulbhagwani:
Rebased onto current main (`e2afb57381eff11d99a956aa14d253513aef7c7f`). Unit tests pass (`tests/fm-treehouse-home-root.test.sh`: 5/5 ok). Ready for CI run and triage.

mehulbhagwani:
Attestation is bound to head `e2afb57381eff11d99a956aa14d253513aef7c7f` (`no-mistakes-pipeline-attestation:v1`).

- **Tip vs main:** Fixes per-home treehouse worktree pool root derivation (`fm_treehouse_home_root`), preventing slot collisions across sibling Firstmate homes.
- **Contract-class:** fix / restore
- **Tests:** `tests/fm-treehouse-home-root.test.sh` and `tests/fm-treehouse-pool-isolation-live-e2e.test.sh` pass cleanly.

Linked issue with reproduction: #4977

PR #4937, opened 2026-09-19
feat(bearings): push the board from fleet events instead of rebuilding it

BenjaminLu:
## What this is

The bearings board was exactly as fresh as the last time firstmate rebuilt it - minutes at best, much worse while firstmate was busy. The transport was never the slow part. Nothing wrote.

This makes the fleet write. The scripts that already record a fact publish it at the moment they record it, and every open board repaints from that push. Measured end to end on this machine, an append reaches a subscriber in single-digit milliseconds.

## The three decisions that are actually the design

**The websocket is Node's standard library, not a package.** There is no `package.json` in this repository and no `node_modules`, so `npm install ws` is not a step a fresh clone can take offline, and a board that needs one is a board most clones do not get. RFC 6455's server half is a SHA-1 handshake and a length-prefixed frame. Writing it was cheaper than the dependency. This was judged against the self-contained principle before anything else, as the brief required.

**The wire carries the whole board, never a delta.** A subscriber that misses a message cannot be left behind by it: the next one it receives is complete, and a reconnecting page is sent the current state before anything else. That is the correctness guarantee, stated plainly rather than implied - it is not best effort, and it is the reason a delta protocol was not built. The cost is bandwidth on a loopback socket, which is not a cost.

**An event may never invent words.** It can change what `fm-bearings-board.v1` already defines, with values the publisher already held - a state word, a row, a pull request URL. It cannot compose the question a captain-facing card is made of, because no script knows it. So a new captain's call, or a pull request on work still listed as underway, marks the board behind and names what changed, and the page says a rebuild is owed. A board quietly missing a question is the failure this whole branch exists to stop. A board that says it is behind is not that failure.

## What publishes, and where

Six sites, each one already the place that records the fact - no watcher was added, and nothing polls for changes:

| site | event |
| --- | --- |
| `bin/fm-spawn.sh` | a task went underway |
| `bin/fm-teardown.sh` | work landed |
| `bin/fm-captain-hold.sh hold` | a captain's call was opened |
| `bin/fm-captain-hold.sh answer` | it was answered |
| `bin/fm-busy-event.sh apply` | a worker started or stopped working |
| `bin/fm-pr-check.sh` | a pull request was registered |

Publishing is one append to `state/board-live.jsonl` and nothing else: no port, no process to reach, nothing that can fail inside a teardown. The log is the durability and the socket is only the latency, so an event published while the server is down is read at the next start rather than lost. `bin/fm-board-live.sh event` exits 0 on every failure by design, and says what it refused.

The server follows that log with a filesystem watch. There is a 30-second interval beside it; it is a backstop for a watch the platform drops, deliberately too slow to be the mechanism, and it is named as such in the source rather than left to look like a poller.

## Out of the box

`bin/fm-bearings-board.sh build` starts the server and derives the subscribing board. Nothing is configured, nothing is installed, nothing is a step to remember. The port is derived from the home's own path - stable across restarts so a board built yesterday reconnects today, distinct per home so a secondmate or a second clone never contends. A board opened with no server running renders from the payload built into it and says on the page that it is not updating; when the build cannot start a server it says so on stderr rather than leaving that to be discovered at the surface the captain reads.

## The page

The shipped template with one script added above its data slot - derived mechanically, never re-authored - so every card type, badge, picker and packet reaches it with no list to maintain. Both rules paid for on the remote copy are carried here and proved by executing the page: an update is held while the captain is writing an answer, and a payload the board cannot render is undone with the last good one redrawn.

Answers are untouched. The board still sends them through the surface that serves it.

## Verification

- `tests/fm-board-live.test.sh` - 26 cases, all passing. The publisher, the merge, a real websocket client speaking to the real server, the derivation, and the transport executed under a DOM shim.
- Not vacuous: two guarantees were deliberately broken and the suite went red on each - removing the undo of an unrenderable payload, and letting a state-only event overwrite what a worker is doing. Both restored.
- The timestamp boundary bug that first run found is fixed: an event stamped to the second and a page mtime in milliseconds meant an event published in the build's own second was dropped.
- `tests/fm-bearings-board-render.test.sh` - 27 cases, passing against the derived board, which is the parity check.
- `bin/fm-lint.sh` clean. `bin/fm-doc-audience-check.sh` clean.

## Still open, and stated as open

- **`tests/fm-bearings-board.test.sh` had not finished when this was pushed.** It is the build suite and it is slow; the build path changed, so it is the one result that matters most and it is not yet in hand. I will report it on this PR.
- **`fm/remote-board-tracked` carries a sibling page harness** (`tests/assets/remote-board-harness.mjs`) with a near-identical DOM shim. That branch is unmerged, so nothing was taken from it; the two should be unified into one shim when it lands rather than kept as two.
- **The `Require no-mistakes` check on this repository will fail.** The captain moved the workflow to direct-PR with a reviewer agent today; `CONTRIBUTING.md` and that CI job still describe the old gate. Not changed here because it is not this branch's scope.
- **The per-home server is real per-home cost.** A test run creates many homes and therefore many servers; they exit on their own within 30 seconds of their home being removed, which was verified, but a suite that holds its homes open holds the servers open too.
- **Not measured:** behavior on a network filesystem, where `fs.watch` is least reliable and the backstop is what would carry it.

Out of scope, as briefed: the remote artifact copy, the answer path back to firstmate, and any hosting service.


BenjaminLu:
Opened against this repository by mistake - this is fleet-internal work that belongs on the fork it came from. Closing; nothing here needs your attention. Apologies for the noise.

Issue #4928, opened 2026-09-19
Away daemon can run for hours delivering nothing; wedge is visible only in its own log


marcospaterson:
## What happens

The away-mode supervisor daemon started and stayed alive, but delivered no escalation for the entire away window. Three crewmates finished their work, parked at ask-user gates, and waited. The captain returned ten and a half hours later to find the fleet idle.

The daemon knew. `bin/fm-afk-return.sh` reported at return:

```
delivery wedged: fm away-mode inject WEDGED: 37935s undelivered as of 2026-09-19T10:04:25-0400
```

So the wedge was detected and recorded, and the durable marker did its job. What did not happen is anyone finding out while it mattered. The one place the operator would look, the primary pane, showed nothing, because the thing that is broken is precisely the channel that would tell them.

## Evidence, 2026-09-18/19, firstmate main at 269f8fe

- Away entered through `bin/fm-afk-launch.sh propose` then `confirm`, then `start-native` and `FM_AFK_STATE_PREPARED=1 bin/fm-afk-start.sh` in the harness background tool, per the `/afk` skill.
- Daemon process alive throughout (`bin/fm-supervise-daemon.sh`).
- Three tasks reached `needs-decision` within about 20 minutes of entry; none was delivered.
- Return brief also reported `GAP: watcher downtime was detected during the away window` and `away-mode shutdown failed; lifecycle state preserved for retry`.

## Why it matters

Away mode's whole promise is that work continues unattended. A daemon that is alive but silent is worse than one that has crashed, because every health signal an operator can cheaply check says it is running.

## Suggested direction

Two separable things:

1. A wedge that persists past `FM_MAX_DEFER_SECS` by orders of magnitude should reach the operator through a channel that does not depend on the wedged one. `docs/wedge-alarm.md` owns an active alert channel; ten hours of undelivered escalations should have triggered it, and on this install nothing did. Worth checking whether the alarm fired and failed, or never fired.
2. Entering away mode should verify delivery once, end to end, before reporting success. A single test injection proven to land would have failed loudly at entry instead of silently at hour ten.

Related: #4927, the turn-end guard deadlock that made away mode look like the safe workaround in the first place.

kunchenguid:
Speaking as Kun's firstmate: **ready-for-pr** (contract-class: **restore**). First stamp (why=unstamped).

[...]

PR #4943, opened 2026-09-19
fix: let the Stop auto-arm claim an unowned session lock; verify the wedge alarm at away entry

### LINKED ISSUE #4927: Turn-end guard blocks the same-group Stop hook that restores the watcher, so missing supervision cannot self-heal
## What happens

On the claude primary, `.claude/settings.json` registers both Stop hooks in the SAME hook group: `bin/fm-turnend-guard.sh --claude` first, then `bin/fm-claude-stop-autoarm.sh` with `asyncRewake: true`.

When no live watcher holds the home lock, the guard blocks the stop (exit 2). Observed on this install, the auto-arm in the same group then does not run at all, so the watcher is never armed, so the next turn end blocks again. A home whose watcher has exited cannot recover on its own; every turn ends with the same banner and the operator is pushed into the manual-arm loop the supervision instructions explicitly forbid.

## Evidence, 2026-09-18/19, firstmate main at 269f8fe

- Guard fired on five consecutive turn ends with `no live watcher holds this home lock`.
- `state/.claude-autoarm-epoch` stayed frozen at `epoch=300 owner_pid=1567976 outcome=rewake updated_at=1789788317` across those turn ends, while wall clock advanced past 1789788577. The auto-arm did not run.
- No failure epoch, no `state/.claude-autoarm-failure-*` marker, registration intact, no stale `state/.watch.lock`.
- `bin/fm-watch-arm.sh` run by hand succeeds immediately every time (`watcher: started pid=... (beacon fresh)`), so the arm path itself is healthy.

## Why it matters

The guard exists to stop a turn ending blind. In this state it guarantees the condition it is guarding against: supervision stays off, and the only way forward is the manual loop the contract tells the agent not to enter.

## Suggested direction

The guard and the arm must not be able to deadlock. Either the arm runs before the guard can block, or the guard treats a successful in-band arm as its own remedy, or they are separated so a blocking guard cannot suppress the recovery hook. Whichever shape, the invariant to test is: a home with in-flight tasks and no live watcher recovers automatically within one turn end, with no operator action.

marcospaterson:
Fixes #4927. Addresses the entry half of #4928.

## The problem

A home whose watcher has exited cannot recover on its own. The turn-end guard reports `TURN WOULD END BLIND` every turn, the auto-arm runs but does nothing, and the operator is pushed into the manual arm loop the supervision instructions explicitly forbid.

Observed on a live home with in-flight work over several hours: `state/.claude-autoarm-epoch` stayed frozen at one record across many turn ends while the guard fired each time, and the arm wrapper run by hand succeeded immediately every time.

## The cause, which is not what the issue first guessed

The issue speculated that the guard's blocking exit prevented the second same-group Stop hook from running. That is wrong, and worth correcting for anyone reading #4927: on Claude 2.1.278 both same-group Stop hooks run in parallel, verified while investigating. The auto-arm does run.

The real cause is that the arm exits early when the session lock is absent or malformed, so it runs, claims nothing, and returns without arming, while the guard keeps blocking on the resulting absence of a watcher.

## The change

- The Stop auto-arm may claim an unowned session lock rather than exiting early, so a home with no live watcher recovers automatically within one turn end.
- At away-mode entry, the wedge alarm is verified rather than assumed, so a broken escalation channel fails loudly at entry instead of silently hours later. This is the entry half of #4928; it does not address a channel that wedges mid-window, and that issue should stay open for it.

## Verification

- Regression test reproduces the pre-fix state, in-flight work with no live watcher and an unowned lock, and fails on the unfixed code.
- The original symptom was reproduced on a live home before anything was changed, rather than working from the report alone.
- ShellCheck could not run in the pipeline on this machine because the binary is not installed there; it was run locally and is clean.

## Note for reviewers

This was found while the fix's own home was losing supervision, so the manual arm workaround was in use throughout. The guard itself is unchanged and still blocks a genuinely blind stop; only the arm's lock handling changed.


kunchenguid:
Speaking as Kun's firstmate: **waiting-author** (contract-class: **new-default**). **Prefer sibling #4955** for the same #4927 fix.

**Tip check** (`b6930db737c382533120d6005ac515478eb6f083`): same unowned-lock inertness bug as #4955. This HEAD claims absent/malformed locks and adds away-entry wedge verify, but **does not** add the guard `FM_CLAUDE_AUTOARM_LOCK_CLAIM_WAIT_MS` allowance that #4955 needs for the integrated self-heal regression under load. Based on older tip than #4955.

**Does it close #4927?** **Would, if preferred** — same lock-claim restore invariant. Body correctly says it only covers the **entry half** of #4928; leave #4928 open. We are **not** landing this while #4955 is open and strictly stronger.

**Contract-class: new-default** — same mix as #4955 (restore lock-claim + new default-on away-entry verify). Cannot auto-merge.

**VISION.md (each rule)** — same verdicts as #4955 (aligns on peace-of-mind / restart / scripts / spine / fleet / scope; **tension** on Authority for the new away-start refusal).

**Checks:** MERGEABLE / UNSTABLE. Attestation **MISSING**. First-time fork CI: **workflow-approved**. If #4955 merges, this will be **closed-overlap** with a thank-you naming #4955 — please do not keep both open after one lands.

**Security FYI:** none material.

No Firstmate escalation. Help #4955 (rebase/attest there) rather than advancing this duplicate.

PR #4955, opened 2026-09-19
fix: let the Stop auto-arm recover an unowned home; alert on wedged away daemon

### LINKED ISSUE #4927: Turn-end guard blocks the same-group Stop hook that restores the watcher, so missing supervision cannot self-heal

marcospaterson:
Fixes #4927. Addresses #4928 (scoped).

## Change
- 4927: the Stop auto-arm now claims an absent or malformed session lock through `bin/fm-lock.sh`, so a home with in-flight work and no live watcher recovers within one turn end. The guard grants the arm a bounded extra wait (`FM_CLAUDE_AUTOARM_LOCK_CLAIM_WAIT_MS`, default 2000) when the lock is unowned, because the claim precedes the generation claim. The guard is not made fail-open. Ownership order is documented in `docs/turnend-guard.md`.
- 4928: the away daemon raises an out-of-band alert when a wedge exceeds its bound, and away entry verifies delivery once (`docs/wedge-alarm.md`).

## Maintainer verification
- New regression test `test_hook_claude_mode_home_with_no_watcher_and_no_lock_self_heals_in_one_turn_end` runs the real guard and real arm together on a home with no lock and no watcher; it fails on the pre-fix arm.
- The first version of that test was flaky (about 1 in 3 runs, always under load) because `fm-lock.sh` alone takes ~0.4s against the 800ms wait; after the lock-claim allowance it passed 6/6 under 6 CPU-burner processes.
- `tests/fm-turnend-guard.test.sh` and `tests/fm-claude-stop-autoarm.test.sh` pass; `bin/fm-lint.sh` (ShellCheck 0.11.0) and `bin/fm-doc-audience-check.sh` pass.
- `tests/fm-afk-launch.test.sh` passes except the herdr e2e case, which cannot prepare a lab session on this machine (herdr unavailable).


kunchenguid:
Speaking as Kun's firstmate: **waiting-author** (contract-class: **new-default**). Preferred over sibling #4943 for #4927.

**Tip check** (`b6930db737c382533120d6005ac515478eb6f083`): `bin/fm-claude-stop-autoarm.sh` still exits 0 on absent/malformed `state/.lock` (inert), while `fm-turnend-guard.sh --claude` blocks when no live watcher holds the home — so an unowned home cannot self-heal. Stop hooks remain same-group in `.claude/settings.json`. Not fixed on main. This HEAD claims the unowned lock through `fm-lock.sh`, adds `FM_CLAUDE_AUTOARM_LOCK_CLAIM_WAIT_MS` (default 2000) to the guard sync wait, and adds a regression that runs real guard+arm together.

**Does it close #4927?** **Yes (invariant).** Author body correctly rejects the issue's same-group-suppression guess (parallel Stop hooks) and names the real tip bug: arm inert on absent/malformed lock. The stated invariant — home with in-flight work and no live watcher recovers within one turn end — is what the new integrated test asserts. GitHub `Fixes #4927` holds for that invariant. Does **not** fully close #4928 (entry-half only; mid-window wedge delivery already partially exists via `inject_wedge_alarm`; leave #4928 open).

**Contract-class: new-default** (mixed; new-default wins). The lock-claim + guard wait half is **restore** of the existing Stop auto-arm recovery path. Bundled away-entry `wedge_alarm_verify` is a **new default-on gate**: `fm-afk-launch.sh` start/start-native now refuses when no out-of-band channel delivers (Linux `auto` → empty channel fails until captain writes `off` or `command:`). Mixed restore + new default-on surface → **new-default**. Cannot auto-merge.

**VISION.md (each rule)**
- One captain, one interface / peace of mind: **aligns** (self-heal restores looking-away; entry alarm check deepens honesty) — with tension: new away-start refusal is louder machinery on the unconfigured path.
- Authority is explicit and never inferred: **tension** — entry verify assumes consent to a new refusal unless `off`/`command:` is set; not a quiet widen of agent autonomy, but it does change default away-start.
- Scripts own the mechanics, agents own the judgment: **aligns** — lock claim and wait are script mechanics.
- A restart is a non-event: **aligns** — watcher death recovers without operator ritual.
- Delegation with a spine: **aligns** — in-flight tasks need live supervision.
- The fleet outlives any vendor: **aligns** — Claude Stop adapter contract.
- Scope: **aligns** — primary supervision / away daemon.

**Checks:** MERGEABLE / UNSTABLE. Attestation **MISSING** (no `no-mistakes-pipeline-attestation:v1` for `7a9dd593eb23529931b27f171a8a94e6b2ba2e4f`). First-time fork CI: **workflow-approved** (runs for HEAD). Need matching NM attestation + green CI before any Firstmate flag for the new-default merge gate.

**Overlap:** Prefer this over #4943 (same author; #4943 lacks the lock-claim wait that the regression needs under load). Open #4026 (CONFLICTING) takes a different approach (`--ensure-watcher` prime on refusal); keep it until one lands, then close leftovers. Sibling #4943 will be closed-overlap if this merges.

**Security FYI:** none material — reuses existing `fm-lock.sh` reclaim path; wedge verify sends a test alert through already-configured channels only.

No Firstmate escalation yet (attestation/CI not green). Author: please raise via no-mistakes so attestation matches HEAD, then CI can finish.

PR #5014, opened 2026-09-20
fix(bin): patch lavish-axi's client to release SSE on hidden tabs

talktejas:
## Summary
- Chrome allows six simultaneous connections per origin, and every open Lavish review page permanently holds one for its live-update stream, so the sixth open page never loads and Send-to-agent on an already-open page silently queues (data/fm-lavish-pages-dont-load/report.md).
- The installed lavish-axi client (0.1.63) still ships this design unchanged; adds `bin/fm-lavish-client-patch.sh`, which patches the installed `dist/chrome-client.js` to release the stream when a tab is hidden and reopen it when shown again, is idempotent on a re-run, and refuses without touching the file if the targeted upstream code has moved.
- The installed client is already patched by this run. A restart of the shared Lavish server is required for the patch to take effect on already-open pages; this task deliberately did not restart it.
- Adds `docs/lavish-connection-limit.md` documenting the problem, the patch, that upgrading lavish-axi reverts it, and the two no-patch workarounds (cap at four or five open pages, or split pages across `127.0.0.1:4387` and `localhost:4387`).
- Adds `tests/fm-lavish-client-patch.test.sh` covering apply-once, no-op-on-repeat, refuse-without-corrupting when the target code has drifted, and a real hide/show behavioral check that the patched stream actually closes and reopens.

## Verification
- Live multi-tab browser verification was not performed: this machine's only browsers are the captain's own, and an earlier attempt to use an automation tool for this turned out to drive his real Edge session (no tabs of his were closed and nothing else was touched).
- Verified statically instead: `node --check` on the patched installed client, presence of the new code with the original event listeners intact, three idempotency cases (unpatched, already-patched, and upstream-code-moved) against throwaway copies of the exact installed version, and an HTTP check against a disposable Lavish instance on a private test port confirming the served client is the patched one.

## Test plan
- [x] `bin/fm-lint.sh`
- [x] `bin/fm-doc-audience-check.sh`
- [x] `tests/fm-lavish-client-patch.test.sh`


talktejas:
Opened against the wrong repo by mistake (gh defaulted to the fork parent instead of talktejas/firstmate). Correct PR: https://github.com/talktejas/firstmate/pull/6

PR #5035, opened 2026-09-20
feat(bin): give task worktrees the clone's graphify graph

salahmwahsh:
## Give task workers the clone's graphify graph

A repo's graphify graph lives at `<clone>/graphify-out/`, but crewmates and scouts work in
disposable task worktrees (`~/.treehouse/<repo>-<hash>/N/<repo>`, or any other pool shape), which
have no `graphify-out` at all. So the agents that do the most reading have never been able to query
a graph, even when the clone has a current, watcher-maintained one sitting next to them. They grep
instead, which is slower and markedly more expensive in tokens.

This links the clone's graph into each task worktree at spawn, and removes the link again at
teardown.

### What changed

- **`bin/fm-spawn.sh`** — when the source clone has a `graphify-out` directory, the task worktree
  receives an absolute symlink to it before launch. An absent source is the normal case and is
  skipped silently; an existing entry in the worktree is preserved untouched; every failure after
  the presence check is reduced to a warning, so graph availability can never block a spawn. The
  link happens at the single pre-launch point every backend and pool shape passes through, not in
  one backend's branch. Secondmate spawns are excluded.
- **`bin/fm-teardown.sh`** — removes the worktree's `graphify-out` symlink before any branch or
  worktree cleanup, guarded on `-L` so only a symlink is ever removed and the clone's real graph
  directory is never a target. A removal failure preserves the task rather than proceeding.
- **`bin/fm-brief.sh`** — one line in the generated ship and scout briefs: when `graphify-out/`
  exists, answer codebase questions with `graphify query` before reaching for grep or bulk reads.

No copying and no rebuilding: the graph for one repo can be tens of megabytes, and the clone's
existing watcher keeps the single copy fresh for every worktree at once. `graphify-out/` is
conventionally covered by a global gitignore, so the link does not appear in `git status`.

### Tests

- `tests/fm-teardown.test.sh` — **the important one**: a worktree whose `graphify-out` is a symlink
  into the source clone is torn down, and the clone's real graph must still be there afterwards.
  Also asserts the symlink is gone before the worktree is returned.
- `tests/fm-spawn-pool-base-freshen.test.sh` — a clone with a graph produces a working link in the
  worktree; a clone without one spawns cleanly and silently.
- `tests/fm-brief.test.sh` — the generated brief carries the graph line.

### Verification

Run on macOS 15.5, bash 3.2:

- `tests/fm-brief.test.sh` — pass, 26 checks.
- `tests/fm-spawn-pool-base-freshen.test.sh` — pass, 25 checks.
- `tests/fm-teardown.test.sh` — the new graph checks pass, including
  `teardown removes only the worktree graphify-out symlink before cleanup`. One unrelated check,
  `herdr-preflight-missing-adapter`, fails in this environment — **it fails identically on
  `origin/main` without this change** (verified in a clean worktree at 9bc051ff: same single
  `not ok`), so it is pre-existing here and not introduced by this branch.
- The full suite was not run to completion locally; it exceeded an hour and hit further
  environment-specific failures on this machine. CI is the gate for the rest.


kunchenguid:



**HEAD** `1c55fb25a8439b21068c1e0b2ed35ce41ec96318` (salahmwahsh/firstmate). MERGEABLE/UNSTABLE vs main `a09090d13ef24ce3cf71d171ade119896d6db301`. Workflow-zero (no `.github/workflows/*`). Author not blocked.

**Attestation: MISSING** — PR body has no bound no-mistakes attestation SHA. Require no-mistakes run `35510640197` FAILURE (body compliance). CI run `35510640113` queued after fork approval.

**Contract-class: new-default** — inspected tip vs main on `bin/fm-spawn.sh`, `bin/fm-teardown.sh`, `bin/fm-brief.sh` + tests. Unconfigured spawn now always-on links the clone's `graphify-out/` into every ship/scout worktree when the source has that directory, and every generated ship/scout brief gains a graphify-first instruction. Main does not specify that worktrees already receive the clone graph; this is a new default path (not restore of a broken promise, not opt-in-gated). Claims that workers “should” have had the graph do not change the class.

**VISION (per rule):**
1. One captain, one interface — **aligns** in intent (token-cheaper reads stay below deck) but the brief line is a new always-on worker instruction surface.
2. Authority explicit — **does not align / product ambiguity**: new default behavior ships without an explicit captain opt-in; consent is inferred from “clone has graphify-out”.
3. Scripts own mechanics — **aligns**: symlink + teardown guards are deterministic scripts; brief text is static.
4. Restart is a non-event — **aligns**: link is recreated on spawn; teardown removes only `-L` before worktree return.
5. Delegation with a spine — **cannot tell / resist pressure**: always-on brief guidance changes how workers explore without a new explicit contract knob.
6. Fleet outlives any vendor — **aligns** if graphify stays optional tooling; **cannot tell** whether coupling worktrees to a graphify convention hardens a vendor-shaped path.
7. Scope — **aligns** as command-layer spawn/brief wiring, not forge/CI.

Closing: token savings are welcome, but always-on default linking + brief instruction should stay opt-in or wait for an explicit captain default-behavior decision once otherwise ready. Resist assuming consent.

**First-time fork CI:** after full diff review I approved workflow runs `35510640113` (CI) and `35510640197` (Require no-mistakes) for this HEAD. No workflow security risk.

**Security FYI (not a captain gate alone):** spawn creates an absolute symlink from the disposable worktree into the clone's `graphify-out/`; teardown only removes when `-L`. Failures warn without blocking spawn; removal failure preserves the task. Tip-only FYI — symlink into the primary clone tree is intentional sharing, not a new credential path.

**Waiting on you:** rebind no-mistakes attestation to this exact HEAD, get green CI/NM. Even then this will **not** auto-merge (new-default); it needs a captain default-behavior decision when otherwise ready. Prefer gating the link/brief behind an explicit opt-in if you want an auto-merge path. Not waiting on the captain yet (attestation + CI still open). Do not merge.

salahmwahsh:
Thanks for the thorough triage — the new-default classification is correct, and I'm not going to argue it.

Context you should have: this change was written for a single local fleet and opened here without its author intending an upstream contribution yet. I'm not pursuing a merge on it right now, so please don't spend further review time on it.

Two accurate notes for anyone who picks it up later:

- **The new-default objection is right.** Linking the graph and adding the brief line are both unconditional whenever the source clone happens to have `graphify-out/`. Consent is genuinely inferred from a directory's existence, which is not consent. A proper version gates both behind an explicit opt-in — a `config/` flag in the same shape as the other per-home knobs — and defaults to off.
- **The branch is 14 commits behind `main`** (base `9bc051ff`, `main` now at `a09090d1`). The change itself is a clean +154 across `fm-spawn.sh`, `fm-teardown.sh`, `fm-brief.sh` and their tests, and cherry-picks onto current `main` with no conflicts, but the branch as pushed is stale.

On the attestation: correct that it's unbound, and I'm not going to rebind it for a PR I'm not pursuing.

One test result that may be worth your time independently of this PR: `tests/fm-teardown.test.sh` has `herdr-preflight-missing-adapter` failing on unmodified `a09090d1` in this environment (macOS 15.5, bash 3.2), verified in a clean worktree with this change absent. Pre-existing and unrelated to this branch, but it does fail on your default branch here.

Happy for this to be closed if it's cluttering the queue — no objection from me.

PR #5116, opened 2026-09-21
feat(bin): bound channel intake reads to an optional scanning window

Amplify-Logic:
## What

Adds two optional keys to private `config/channel-intake`: `active_start` and `active_end` (HH:MM, both required together).
When set, `source_due()` returns not-due for every source outside the window, so `due_source_ids()`, `tick`, the registered watcher check shim and `claim` all report nothing until the window opens.

Previously `quiet_start`/`quiet_end` only suppressed alert payloads; the reads themselves still became due and woke the orchestrator around the clock.

## Behavior

- Inside the window the existing interval and backoff logic is unchanged.
- The first tick after the window opens reads immediately: the cadence is measured from each source's `last_attempt`, so a laptop opened at 10:15 into an 09:00 window reads at 10:15, not at the boundary.
- The window wraps midnight when the end is at or before the start, the same rule quiet hours use.
- Both keys unset keeps the historical all-day behavior.
- Half a window, or equal bounds, is refused rather than silently scanning all day.
- `status` reports `active_window` and `in_active_window`.
- The notify path and quiet hours are untouched.

## Verification

- `tests/fm-channel-intake.test.sh` extended with `test_scan_window_bounds_the_reads`, driving the executable at a fixed epoch across inside, outside, wrap-around, unset, half-configured and equal-bounds cases; full suite green.
- `bin/fm-doc-audience-check.sh` ok.


Amplify-Logic:
opened against the wrong repository, re-raised on the fork

PR #5121, opened 2026-09-21
fix(bin): let teardown reconcile pre-binding legacy task records

Amplify-Logic:
## Problem

A task record written before the endpoint-binding (`endpoint_task_id`) and incarnation (`spawn_gen`) fields existed can name neither an isolated copy nor an endpoint it can prove is its own. Endpoint validation refused both shapes outright:

- `legacy Herdr endpoint metadata for task <id> lacks an exact task binding; preserving task state`
- `task <id> has a missing, empty, or ambiguous worktree identity; preserving task state`

Those refusals fire before `--legacy-record`'s own endpoint gate, so such a record could not be retired by any supported path. It stays in the agent pane and status bar forever as a terminal doing nothing.

## Change

`fm_backend_validate_task_endpoint` gains an opt-in `--reconcile-legacy` third argument that *reports* exactly two degradations instead of refusing them:

- an absent or empty `worktree=` field -> `FM_BACKEND_VALIDATED_WORKTREE_ABSENT=1`
- an absent `endpoint_task_id` on a Herdr record -> `FM_BACKEND_VALIDATED_ENDPOINT_UNBOUND=1`

`fm-teardown.sh` passes that flag only under `--legacy-record`, and then:

- **No worktree identity** means nothing to return to the pool. The existing slot machinery already skips a record with no `worktree=`, so every isolated-copy step falls away. A worktree that *is* present still passes the ordinary dirty and landed-work checks before anything is returned.
- **An unbound endpoint** must be proved gone by the backend's own liveness read, never by name-matching another home's endpoints. Only `missing` (absent) qualifies; `dead` means the pane is still there without an agent, which this record cannot claim, so it refuses. Once absent, the Herdr preflight, the close, and the confirmed-gone gate are all skipped as "no endpoint to stop", so no endpoint is ever closed on another task's behalf.

The backlog transition and status-log retirement are byte-for-byte the ordinary teardown path.

### Still refuses, naming the failing condition

- uncommitted or unlanded work (both degradations, `--force` remains the only discard authority)
- a record naming more than one `worktree=`
- an `endpoint_task_id` naming a different task
- inconsistent or malformed Herdr endpoint metadata
- another backend's missing binding (zellij, orca, cmux unchanged)
- an endpoint the backend still reports present
- any of the above without `--legacy-record`

## Tests

Six new end-to-end cases in `tests/fm-teardown.test.sh` driving the real executable with fixture records, plus direct validation coverage in `tests/fm-teardown-endpoint-safety.test.sh` (relaxations, and each preserved refusal including the unknown-option guard).

## Verification

- `tests/fm-teardown.test.sh`: 91 passing, exit 0
- `tests/fm-teardown-endpoint-safety.test.sh`: clean
- `tests/fm-backend.test.sh`, `tests/fm-backlog-atomicity.test.sh`: clean
- `shellcheck`, `bin/fm-lint.sh`, `bin/fm-doc-audience-check.sh`: clean

Note: `test_herdr_flat_teardown_preflight_refuses_before_changes` (missing-adapter mode) fails on `main` as well - confirmed against a clean clone of the default branch, so it is pre-existing and unrelated to this change.

## Docs

`bin/fm-teardown.sh`'s header remains the owner of the reconciliation rules; `--help` was added and `docs/architecture.md` carries a pointer.

## Out of scope

`fm-spawn`'s record format, in-place record migration, and any change to existing records.


Amplify-Logic:
opened against the wrong repository, re-raised on the fork

PR #5147, opened 2026-09-21
fix(bin): pin watcher guard lock reads and stage .watch.lock publication

slee029:
## Intent

Fix the watcher turn-end guard false-BLIND race caused by non-atomic .watch.lock publication vs unpinned guard reads (ticket HHE-1805).

## What Changed

- `bin/fm-wake-lib.sh`: guard readers (`fm_watcher_healthy`, lock-matches-pid, lock-unheld) now resolve the `.watch.lock` symlink once, read pid/fm-home/watcher-path/pid-identity from that pinned owner generation, and re-verify the lock still names the identical owner before deciding; mid-read owner flips are retried up to `FM_WATCHER_PIN_ATTEMPTS` (default 10) and a briefly absent lock is retried `FM_WATCHER_PIN_ABSENT_ATTEMPTS` (default 3) times instead of being reported as down. `fm_lock_claim` gains `fm_lock_prepare_owner`, which writes the full owner file set into the owner directory before the symlink is published when `FM_LOCK_OWNER_FOR` names that lock, and `FM_WATCHER_HEALTH_REASON` now records which predicate failed (`lock-absent`, `pid-dead`, `home-mismatch`, `torn-generation`, etc.).
- `bin/fm-watch.sh`: the watcher sets `WATCHER_PID`, installs its EXIT/HUP/INT/TERM cleanup traps, and stages the owner-generation values before `fm_lock_try_acquire`, then unsets the staging variables so unrelated locks publish bare owner directories; the old post-acquire `printf` writes of fm-home/watcher-path/pid-identity are removed, and the retain-evidence exit paths set `WATCHER_RETAIN_LOCK_EVIDENCE` so cleanup leaves the lock and recovery marker untouched. `bin/fm-turnend-guard.sh` prints a `Lock evidence:` line with the failed predicate when it blocks.
- Added `tests/fm-watcher-guard-race.test.sh` covering staged vs. legacy staggered publication, generation flips during a guard read, the health-reason matrix, verified-not-rewritten pid, watch-lock-only staging, retain-evidence exits, and absent-lock retry tolerance vs. sustained absence; documented the new env vars and guard behavior in `docs/configuration.md` and `docs/turnend-guard.md`.

## Risk Assessment

⚠️ Medium: The change is correct as far as I can verify — staged publication is properly scoped to the watch lock, every guard reader funnels through one pinned reader, the claim-verify change does not regress the two pre-existing direct callers in tests/fm-watcher-lock.test.sh, and the retain-evidence exits leave both the lock and the marker untouched — but it rewrites a lock primitive shared repo-wide and two of its headline behavioral proofs self-skip under root, so it warrants attention before merge rather than being trivially safe.

## Testing

Ran tests/fm-watcher-guard-race.test.sh (9 cases, including the real fm-watch.sh retain-evidence exit, absent-lock retry in both directions, staging scope, and pid verify) and tests/fm-turnend-guard.test.sh (all pass), then drove the real watcher and real Stop hook live in isolated homes: rapid restart hammer on fixed vs base at the same cadence (0 vs 399 false BLINDs while the lock was published throughout), base steady state confirmed allowing so the contrast isolates the publication window, dead and tampered watcher adversarial checks block with named reasons, and a 150-iteration TERM-during-startup stress on each commit (fixed leaves no stale lock, base leaves 61). All scenarios passed; evidence logs saved to the evidence directory.

- Live validation: ✅ go - 8 of 8 scenarios driven live against the product

| --- | --- | --- | --- |
| Turn-end hook never fires TURN WOULD END BLIND while a live watcher&#39;s lock is published, even under continuous watcher restarts | ✅ pass | live | live-fixed-hammer.log: hook evaluations=400 allowed=277 blocked-while-lock-held-throughout(FALSE-BLIND)=0 across 248 real watcher restarts |
| Regression reproduction: the same restart hammer against the base commit produces false BLINDs | ✅ pass | live | live-base-hammer.log: 399 of 399 lock-held-throughout evaluations blocked on base 43bf6d3 at the same cadence; live-base-hammer-1s-hold.log shows base allows once the partial window has passed |
| Adversarial: a genuinely dead watcher (SIGKILL, lock left behind) still blocks with a named predicate reason | ✅ pass | live | live-dead-and-tampered-watcher.log: hook rc=2 with &#39;Lock evidence: pid-dead&#39; |
| Adversarial: tampered lock metadata (fm-home rewritten on a live lock) blocks with a named reason | ✅ pass | live | live-dead-and-tampered-watcher.log: hook rc=2 with &#39;Lock evidence: home-mismatch&#39; |
| Published lock generation is complete at publication (no partial owner directory ever observable) while legacy staggered publication is still detectable by the same observer | ✅ pass | live | guard-race-suite.log: staged publication 0 partials in 139 samples; legacy staggered 103 partials in 124 samples |
| A generation flip forced at every read position inside one guard evaluation yields zero false BLINDs; absent-lock retry rides a republish within budget and blocks a sustained absence | ✅ pass | live | guard-race-suite.log: generation-flip and absent-retry cases pass |
| Retain-evidence exit of the real watcher leaves the held lock and a malformed marker byte-untouched | ✅ pass | live | guard-race-suite.log: test_retain_evidence_exit_leaves_lock_and_marker_untouched drives bin/fm-watch.sh and passes (run as uid 1000, not skipped) |
| A TERM arriving at any point during watcher startup, including mid-acquire, never strands a stale lock | ✅ pass | live | live-term-stress-fixed.log: iterations=150 survivors=0 stale-locks=0; base comparison live-term-stress-base.log: stale-locks=61 |

<details>
<summary>Evidence: Live hammer on fixed code: real hook vs continuously restarting real watcher</summary>

Source: [Live hammer on fixed code: real hook vs continuously restarting real watcher](https://github.com/slee029/firstmate/blob/7484d746a73635bd56f6a02ddf58e7f3579b8aa5/.no-mistakes/evidence/fix/hhe1805-watcher-guard-race/live-fixed-hammer.log)

<code>hook evaluations=400 allowed=277 blocked-during-genuine-gap=123 blocked-while-lock-held-throughout(FALSE-BLIND)=0 other=0 evaluations-with-lock-held-throughout=165&#10;watcher restarts performed=248</code>

```text
first watcher pid=1872037 lock=/tmp/hhe1805-fixed/state/.watch.lock.owner.PbwEol pid-file=1872037
watcher restarts performed=248
hook evaluations=400 allowed=277 blocked-during-genuine-gap=123 blocked-while-lock-held-throughout(FALSE-BLIND)=0 other=0 evaluations-with-lock-held-throughout=165
watcher stdout/stderr (uniq):
      1 check: rearm-resurface
      2 /tmp/hhe1805-fixed/bin/fm-wake-lib.sh: /tmp/hhe1805-fixed/bin/fm-watch.sh: line 39: unexpected EOF while looking for matching `)'
      2 /tmp/hhe1805-fixed/bin/fm-wake-lib.sh: trap: line 2: unexpected EOF while looking for matching `)'
```
</details>
<details>
<summary>Evidence: Live hammer on base commit 43bf6d3 at identical cadence (bug reproduced)</summary>

Source: [Live hammer on base commit 43bf6d3 at identical cadence (bug reproduced)](https://github.com/slee029/firstmate/blob/7484d746a73635bd56f6a02ddf58e7f3579b8aa5/.no-mistakes/evidence/fix/hhe1805-watcher-guard-race/live-base-hammer.log)

<code>hook evaluations=400 allowed=0 blocked-during-genuine-gap=1 blocked-while-lock-held-throughout(FALSE-BLIND)=399 other=0 evaluations-with-lock-held-throughout=399&#10;● TURN WOULD END BLIND - SUPERVISION IS OFF&#10;● 1 task(s) in flight, but no live watcher holds this home lock (last beat: 1s ago).</code>

```text
first watcher pid=2036947 lock=/tmp/hhe1805-basehome/state/.watch.lock.owner.giGCJL pid-file=2036947
cycle 2 FALSE-BLIND (lock /tmp/hhe1805-basehome/state/.watch.lock.owner.MSINrk held throughout):
●  1 task(s) in flight, but no live watcher holds this home lock (last beat: 1s ago).
cycle 3 FALSE-BLIND (lock /tmp/hhe1805-basehome/state/.watch.lock.owner.MSINrk held throughout):
●  1 task(s) in flight, but no live watcher holds this home lock (last beat: 3s ago).
cycle 4 FALSE-BLIND (lock /tmp/hhe1805-basehome/state/.watch.lock.owner.MSINrk held throughout):
●  1 task(s) in flight, but no live watcher holds this home lock (last beat: 4s ago).
cycle 5 FALSE-BLIND (lock /tmp/hhe1805-basehome/state/.watch.lock.owner.MSINrk held throughout):
●  1 task(s) in flight, but no live watcher holds this home lock (last beat: 5s ago).
cycle 6 FALSE-BLIND (lock /tmp/hhe1805-basehome/state/.watch.lock.owner.MSINrk held throughout):
●  1 task(s) in flight, but no live watcher holds this home lock (last beat: 7s ago).
cycle 7 FALSE-BLIND (lock /tmp/hhe1805-basehome/state/.watch.lock.owner.MSINrk held throughout):
●  1 task(s) in flight, but no live watcher holds this home lock (last beat: 8s ago).
cycle 8 FALSE-BLIND (lock /tmp/hhe1805-basehome/state/.watch.lock.owner.MSINrk held throughout):
●  1 task(s) in flight, but no live watcher holds this home lock (last beat: 9s ago).
cycle 9 FALSE-BLIND (lock /tmp/hhe1805-basehome/state/.watch.lock.owner.MSINrk held throughout):
●  1 task(s) in flight, but no live watcher holds this home lock (last beat: 10s ago).
cycle 10 FALSE-BLIND (lock /tmp/hhe1805-basehome/state/.watch.lock.owner.MSINrk held throughout):
●  1 task(s) in flight, but no live watcher holds this home lock (last beat: 12s ago).
cycle 11 FALSE-BLIND (lock /tmp/hhe1805-basehome/state/.watch.lock.owner.MSINrk held throughout):
●  1 task(s) in flight, but no live watcher holds this home lock (last beat: 13s ago).
cycle 12 FALSE-BLIND (lock /tmp/hhe1805-basehome/state/.watch.lock.owner.MSI

... [304942 bytes truncated] ...

held throughout):
●  1 task(s) in flight, but no live watcher holds this home lock (last beat: 483s ago).
cycle 392 FALSE-BLIND (lock /tmp/hhe1805-basehome/state/.watch.lock.owner.MSINrk held throughout):
●  1 task(s) in flight, but no live watcher holds this home lock (last beat: 485s ago).
cycle 393 FALSE-BLIND (lock /tmp/hhe1805-basehome/state/.watch.lock.owner.MSINrk held throughout):
●  1 task(s) in flight, but no live watcher holds this home lock (last beat: 486s ago).
cycle 394 FALSE-BLIND (lock /tmp/hhe1805-basehome/state/.watch.lock.owner.MSINrk held throughout):
●  1 task(s) in flight, but no live watcher holds this home lock (last beat: 487s ago).
cycle 395 FALSE-BLIND (lock /tmp/hhe1805-basehome/state/.watch.lock.owner.MSINrk held throughout):
●  1 task(s) in flight, but no live watcher holds this home lock (last beat: 488s ago).
cycle 396 FALSE-BLIND (lock /tmp/hhe1805-basehome/state/.watch.lock.owner.MSINrk held throughout):
●  1 task(s) in flight, but no live watcher holds this home lock (last beat: 490s ago).
cycle 397 FALSE-BLIND (lock /tmp/hhe1805-basehome/state/.watch.lock.owner.MSINrk held throughout):
●  1 task(s) in flight, but no live watcher holds this home lock (last beat: 491s ago).
cycle 398 FALSE-BLIND (lock /tmp/hhe1805-basehome/state/.watch.lock.owner.MSINrk held throughout):
●  1 task(s) in flight, but no live watcher holds this home lock (last beat: 492s ago).
cycle 399 FALSE-BLIND (lock /tmp/hhe1805-basehome/state/.watch.lock.owner.MSINrk held throughout):
●  1 task(s) in flight, but no live watcher holds this home lock (last beat: 493s ago).
cycle 400 FALSE-BLIND (lock /tmp/hhe1805-basehome/state/.watch.lock.owner.MSINrk held throughout):
●  1 task(s) in flight, but no live watcher holds this home lock (last beat: 495s ago).
watcher restarts performed=127
hook evaluations=400 allowed=0 blocked-during-genuine-gap=1 blocked-while-lock-held-throughout(FALSE-BLIND)=399 other=0 evaluations-with-lock-held-throughout=399
watcher stdout/stderr (uniq):
```
</details>
<details>
<summary>Evidence: Live dead and tampered watcher checks (named reasons)</summary>

Source: [Live dead and tampered watcher checks (named reasons)](https://github.com/slee029/firstmate/blob/7484d746a73635bd56f6a02ddf58e7f3579b8aa5/.no-mistakes/evidence/fix/hhe1805-watcher-guard-race/live-dead-and-tampered-watcher.log)

<code>hook rc (live)=0&#10;● Lock evidence: pid-dead.&#10;hook rc (dead)=2&#10;● Lock evidence: home-mismatch.&#10;hook rc (fm-home tampered)=2</code>

```text
live watcher pid=2452610
hook rc (live)=0
after SIGKILL: lock -> /tmp/hhe1805-dead/state/.watch.lock.owner.n14gTC pid=2452610
●  1 task(s) in flight, but no live watcher holds this home lock (last beat: 1s ago).
●  Lock evidence: pid-dead.
hook rc (dead)=2
--- corrupt evidence: blank out pid-identity in the live lock
hook rc (new live watcher 2454357)=0
●  1 task(s) in flight, but no live watcher holds this home lock (last beat: 0s ago).
●  Lock evidence: home-mismatch.
hook rc (fm-home tampered)=2
```
</details>
<details>
<summary>Evidence: TERM-during-startup stress, fixed vs base</summary>

Source: [TERM-during-startup stress, fixed vs base](https://github.com/slee029/firstmate/blob/7484d746a73635bd56f6a02ddf58e7f3579b8aa5/.no-mistakes/evidence/fix/hhe1805-watcher-guard-race/live-term-stress-fixed.log)

<code>FIXED: iterations=150 survivors=0 stale-locks=0 parse-errors=0&#10;BASE: iterations=150 survivors=0 stale-locks=61 parse-errors=0</code>

```text
iterations=150 survivors=0 stale-locks=0 leaked-marker-lock-owner-dirs=9 parse-errors=0
watcher output uniq:
     49 check: rearm-resurface
```
</details>
<details>
<summary>Evidence: Base TERM stress log</summary>

Source: [Base TERM stress log](https://github.com/slee029/firstmate/blob/7484d746a73635bd56f6a02ddf58e7f3579b8aa5/.no-mistakes/evidence/fix/hhe1805-watcher-guard-race/live-term-stress-base.log)

```text
iter 1 delay=0.506 pid=2513654 STALE LOCK left (pid file 2513654)
iter 9 delay=0.734 pid=2518105 STALE LOCK left (pid file 2518105)
iter 11 delay=0.777 pid=2518734 STALE LOCK left (pid file 2518734)
iter 12 delay=1.077 pid=2519280 STALE LOCK left (pid file 2519280)
iter 17 delay=0.801 pid=2522293 STALE LOCK left (pid file 2522293)
iter 18 delay=0.540 pid=2522901 STALE LOCK left (pid file 2522901)
iter 19 delay=1.191 pid=2523398 STALE LOCK left (pid file 2523398)
iter 20 delay=0.543 pid=2524157 STALE LOCK left (pid file 2524157)
iter 21 delay=0.748 pid=2524510 STALE LOCK left (pid file 2524510)
iter 22 delay=1.090 pid=2524926 STALE LOCK left (pid file 2524926)
iter 23 delay=0.680 pid=2525880 STALE LOCK left (pid file 2525880)
iter 26 delay=0.691 pid=2528323 STALE LOCK left (pid file 2528323)
iter 30 delay=0.527 pid=2531991 STALE LOCK left (pid file 2531991)
iter 31 delay=0.503 pid=2532398 STALE LOCK left (pid file 2532398)
iter 32 delay=0.570 pid=2532839 STALE LOCK left (pid file 2532839)
iter 33 delay=0.785 pid=2533338 STALE LOCK left (pid file 2533338)
iter 36 delay=0.645 pid=2535446 STALE LOCK left (pid file 2535446)
iter 38 delay=0.493 pid=2536884 STALE LOCK left (pid file 2536884)
iter 40 delay=0.514 pid=2537534 STALE LOCK left (pid file 2537534)
iter 48 delay=0.780 pid=2542467 STALE LOCK left (pid file 2542467)
iter 53 delay=0.490 pid=2545125 STALE LOCK left (pid file 2545125)
iter 55 delay=0.809 pid=2546320 STALE LOCK left (pid file 2546320)
iter 56 delay=0.865 pid=2546680 STALE LOCK left (pid file 2546680)
iter 64 delay=0.728 pid=2550486 STALE LOCK left (pid file 2550486)
iter 65 delay=0.620 pid=2550864 STALE LOCK left (pid file 2550864)
iter 67 delay=0.806 pid=2551210 STALE LOCK left (pid file 2551210)
iter 68 delay=1.110 pid=2551478 STALE LOCK left (pid file 2551478)
iter 69 delay=0.624 pid=2552396 STALE LOCK left (pid file 2552396)
iter 71 delay=0.711 pid=2553774 STALE LOCK left (pid file 2553774)
iter 77 delay=0.673 pid=2556655 STALE LOCK left (pid file 2556655)
iter 78 delay=1.166 pid=2557138 STALE LOCK left (pid file 2557138)
iter 80 delay=0.563 pid=2558601 STALE LOCK left (pid file 2558601)
iter 81 delay=1.180 pid=2559042 STALE LOCK left (pid file 2559042)
iter 98 delay=0.864 pid=2570655 STALE LOCK left (pid file 2570655)
iter 106 delay=0.795 pid=2575842 STALE LOCK left (pid file 2575842)
iter 107 delay=0.910 pid=2576540 STALE LOCK left (pid file 2576540)
iter 108 delay=0.881 pid=2577253 STALE LOCK left (pid file 2577253)
iter 109 delay=0.728 pid=2577974 STALE LOCK left (pid file 2577974)
iter 110 delay=0.919 pid=2578481 STALE LOCK left (pid file 2578481)
iter 111 delay=1.082 pid=2579289 STALE LOCK left (pid file 2579289)
iter 112 delay=1.285 pid=2580235 STALE LOCK left (pid file 2580235)
iter 113 delay=1.222 pid=2581210 STALE LOCK left (pid file 2581210)
iter 114 delay=0.591 pid=2582209 STALE LOCK left (pid file 2582209)
iter 117 delay=1.137 pid=2583032 STALE LOCK left (pid file 2583032)
iter 120 delay=0.629 pid=2586106 STALE LOCK left (pid file 2586106)
iter 121 delay=0.862 pid=2586592 STALE LOCK left (pid file 2586592)
iter 122 delay=1.348 pid=2587211 STALE LOCK left (pid file 2587211)
iter 124 delay=0.713 pid=2588441 STALE LOCK left (pid file 2588441)
iter 125 delay=1.043 pid=2588956 STALE LOCK left (pid file 2588956)
iter 126 delay=0.677 pid=2589843 STALE LOCK left (pid file 2589843)
iter 127 delay=0.914 pid=2590330 STALE LOCK left (pid file 2590330)
iter 128 delay=0.516 pid=2590908 STALE LOCK left (pid file 2590908)
iter 131 delay=0.740 pid=2592993 STALE LOCK left (pid file 2592993)
iter 132 delay=0.691 pid=2593369 STALE LOCK left (pid file 2593369)
iter 133 delay=1.061 pid=2593642 STALE LOCK left (pid file 2593642)
iter 134 delay=1.278 pid=2594023 STALE LOCK left (pid file 2594023)
iter 135 delay=1.265 pid=2594670 STALE LOCK left (pid file 2594670)
iter 136 delay=0.563 pid=2595183 STALE LOCK left (pid file 2595183)
iter 139 delay=0.649 pid=2595741 STALE LOCK left (pid file 2595741)
iter 140 delay=0.592 pid=2596040 STALE LOCK left (pid file 2596040)
iter 141 delay=1.213 pid=2596302 STALE LOCK left (pid file 2596302)
iterations=150 survivors=0 stale-locks=61 leaked-marker-lock-owner-dirs=25 parse-errors=0
watcher output uniq:
     46 check: rearm-resurface
```
</details>
<details>
<summary>Evidence: Guard-race suite output</summary>

Source: [Guard-race suite output](https://github.com/slee029/firstmate/blob/7484d746a73635bd56f6a02ddf58e7f3579b8aa5/.no-mistakes/evidence/fix/hhe1805-watcher-guard-race/guard-race-suite.log)

```text
ok - staged publication shows zero partial generations in 139 samples
ok - legacy staggered publication is observable (103 partials in 124 samples)
ok - a generation flip at every read position inside a guard evaluation yields zero false-BLINDs
ok - dead watcher still blocks with a named predicate reason
ok - staged acquire publishes a complete lock generation
ok - the published pid is verified, never truncated and rewritten
ok - staged owner files reach only the published watch lock, not the nested steal or marker locks
ok - a retain-evidence exit leaves the held lock and the malformed marker untouched
ok - absent-lock retry rides a republish inside its 3-retry budget and blocks an absence that outlasts it
```
</details>
<details>
<summary>Evidence: Turn-end guard hook suite output</summary>

Source: [Turn-end guard hook suite output](https://github.com/slee029/firstmate/blob/7484d746a73635bd56f6a02ddf58e7f3579b8aa5/.no-mistakes/evidence/fix/hhe1805-watcher-guard-race/turnend-guard-suite.log)

```text
ok - fm_supervision_unhealthy: false with no state/*.meta at all
ok - fm_supervision_unhealthy: true with in-flight task and no beacon ever
ok - fm_supervision_unhealthy: true with in-flight task and a beacon far outside the grace window
ok - fm_supervision_unhealthy: false with in-flight task and a fresh beacon
ok - fm_supervision_status: FM_SUP_QUEUE_PENDING tracks state/.wake-queue
ok - fm_supervision_needed: X-mode relay poll needs supervision
ok - fm_supervision_unhealthy: source-only home needs supervision
ok - fm_supervision_needed: a registered custom check needs supervision with no task in flight
ok - fm_supervision_needed: a registered check whose bytes drifted still needs supervision
ok - fm_supervision_needed: false for a check.sh with no registration binding
ok - fm_supervision_needed: a task PR poll without a trust binding is not a registered custom check
ok - fm_supervision_status: the relay shim is not counted as a registered custom check
ok - fm-turnend-guard: silent no-op with nothing in flight
ok - fm-turnend-guard: blocks when a fresh beacon has no live watcher lock
ok - fm-turnend-guard: non-Claude path blocks a source-only home
ok - fm-turnend-guard: blocks on a dead watcher lock even when the beacon is fresh
ok - fm-turnend-guard: silent no-op with a live watcher lock and fresh beacon
ok - fm-turnend-guard: healthy non-Claude harness paths ignore Claude episode contention
ok - fm-turnend-guard: blocks on a live watcher lock with an ancient beacon
ok - fm-turnend-guard: blocks with the exact required reason in the primary when unhealthy
ok - fm-turnend-guard: blocks from active FM_HOME state, not only repo-root state
ok - fm-turnend-guard: X-mode repair reason sources the cadence config
ok - fm-turnend-guard: X-mode-only supervision remains guarded in default mode
ok - fm-turnend-guard: registered-check-only supervision is named in the block banner
ok - fm-turnend-guard: ignores stale repo-root state when FM_HOME is set
ok - fm-turnend-guard: uses FM_STATE_OVERRIDE ahead of FM_HOME/state
ok - fm-turnend-guard: stop_hook_active=true always allows the stop (never blocks twice in one turn)
ok - fm-turnend-guard: blocks a blind turn end in a secondmate's own home (.fm-secondmate-home no longer excludes it)
ok - fm-turnend-guard: idle-by-default - silent in a secondmate home with nothing in flight
ok - fm-turnend-guard: stop_hook_active=true allows the stop in a secondmate home (never blocks twice in one turn)
ok - fm-turnend-guard: secondmate deferred-death recovery - silent while watched, forces re-arm once the watcher exits
ok - fm-turnend-guard: inert in a secondmate's own child worktree (linked git worktree) even when unhealthy
ok - fm-turnend-guard: blocks a blind turn end in a treehouse-leased LINKED secondmate home (marker force-include)
ok - fm-turnend-guard: an invalid (empty) marker cannot spoof inclusion; linked worktree stays exempt
ok - fm-turnend-guard: a non-ASCII marker cannot spoof inclusion; linked worktree stays exempt
ok - fm-turnend-guard: inert in a crewmate/scout task worktree (linked git worktree) even when unhealthy
ok - fm-turnend-guard: fails open (never blocks) when jq is missing
ok - fm-turnend-guard: silent no-op on empty stdin
ok - fm-turnend-guard: runs well under the generous timing margin (0s)
ok - fm-turnend-guard-grok: forces one explicitly marked same-session resume when the shared predicate blocks
ok - fm-turnend-guard-grok: legacy environment loop guard prevents a nested resume loop
ok - fm-turnend-guard-grok: native false delegates blocking feedback with zero resume processes
ok - fm-turnend-guard-grok: native true remains bounded and starts no resume process
ok - fm-turnend-guard-grok: both spellings are typed and camelCase has deterministic precedence
ok - fm-turnend-guard-grok: malformed, invalidly typed, and missing-prerequisite payloads start neither path
ok - fm-turnend-guard-grok: missing jq and no-supervision-needed stops stay silent and bounded
ok - tracked .claude/settings.json entries: 5 inert under grok, the documented subagent exception still armed, all live under Claude
ok - .codex/hooks.json: Stop hook uses hook process root when payload cwd is outside
ok - .codex/hooks.json: Stop hook ignores nested git root guard scripts
ok - .opencode primary plugin: guard path is anchored to worktree, not directory
ok - .pi primary extension: no-tool and multi-tool runs each inject exactly one guard follow-up
ok - .pi primary extension: delivery failure resets the logical-run latch
ok - fm-turnend-guard --claude: re-blocks a loop-guarded stop while unhealthy and unclaimed (incident regression)
ok - fm-turnend-guard --claude: X-mode-only homes re-block when auto-arm recovery is absent
ok - fm-turnend-guard --claude: a live arming epoch advances once and repeated observation is idempotent
ok - fm-turnend-guard --claude: repeated failed-to-arming races make bounded monotonic progress
ok - fm-turnend-guard --claude: terminal owner boundary excludes a concurrent start without deadlock
ok - fm-turnend-guard --claude: fresh rewake epoch prevents a duplicate continuation for the same event
ok - fm-turnend-guard --claude: an abandoned auto-arm claim no longer allows a blind stop (incident regression)
ok - fm-turnend-guard --claude: a claim whose pid was reused stops counting as recovery even while its entry reads arming
ok - fm-turnend-guard --claude: a hung owner frozen at arming with no watcher beat no longer allows a blind stop
ok - fm-turnend-guard --claude: a live open generation claim owns recovery with no lock held
ok - fm-turnend-guard --claude: a stuck generation claim no longer allows a blind stop
ok - fm-turnend-guard --claude: the terminal path clears an abandoned claim instead of stepping aside silently
ok - fm-turnend-guard --claude: fresh failed epochs preserve and advance monotonic fail-open progression
ok - fm-turnend-guard --claude: integrated fresh failures reach one bounded fail-open, stop continuation, and reset on recovery
ok - fm-turnend-guard --claude: an inert auto-arm's frozen epoch reaches one bounded fail-open and resets on recovery
ok - fm-turnend-guard --claude: a frozen unverified epoch spends the budget yet still blocks
ok - fm-turnend-guard --claude: reset contention preserves all episode state until retry
ok - fm-turnend-guard --claude: concurrent auto-arm and guard resets are idempotent and deadlock-free
ok - fm-turnend-guard --claude: stale rewake epoch does not allow a blind stop
ok - fm-turnend-guard --claude: budget exhaustion alone cannot permit a blind stop
ok - fm-turnend-guard --claude: verified fail-open is loud, bounded, attended, and non-repeating
ok - fm-turnend-guard --claude: fail-open requires both exhausted retries and consumed notice
ok - fm-turnend-guard --claude: away ownership excludes the Stop-autoarm fail-open
ok - fm-turnend-guard --claude: positive watcher recovery resets failure episode state
ok - fm-turnend-guard --claude: bounded claim wait avoids a token-consuming forced continuation
ok - fm-turnend-guard --claude: secondmate home re-blocks unclaimed and allows auto-arm-claimed stops
ok - fm-turnend-guard: a live away-mode daemon satisfies supervision with no watcher holding the lock
ok - fm-turnend-guard: away-mode daemon ownership survives a leftover dead watcher lock
ok - fm-turnend-guard: away mode with no daemon and no watcher still blocks
ok - fm-turnend-guard: away mode blocks on a dead away-mode daemon
ok - fm-turnend-guard: away mode blocks on a pid-reused away-mode daemon lock
ok - fm-turnend-guard: away-mode daemon ownership never substitutes for a fresh beacon
ok - fm-turnend-guard: a daemon lock proves nothing while away mode is off
ok - fm-turnend-guard: away-mode beacon freshness uses the poll-derived grace, not the flat default
ok - fm-turnend-guard: a dead away-mode daemon still blocks under the poll-derived grace
ok - fm-turnend-guard: the poll-derived grace is bounded, not unlimited
ok - fm-turnend-guard: with away mode off, the poll-derived grace never applies
```
</details>
<details>
<summary>Evidence: Base hammer with 1s hold (steady state allows)</summary>

Source: [Base hammer with 1s hold (steady state allows)](https://github.com/slee029/firstmate/blob/7484d746a73635bd56f6a02ddf58e7f3579b8aa5/.no-mistakes/evidence/fix/hhe1805-watcher-guard-race/live-base-hammer-1s-hold.log)

```text
first watcher pid=2324539 lock=/tmp/hhe1805-basehome/state/.watch.lock.owner.VouiMT pid-file=2324539
watcher restarts performed=156
hook evaluations=300 allowed=132 blocked-during-genuine-gap=168 blocked-while-lock-held-throughout(FALSE-BLIND)=0 other=0 evaluations-with-lock-held-throughout=107
watcher stdout/stderr (uniq):
    156 check: rearm-resurface
```
</details>

- Evidence: [Early-TERM timing sweep on fixed watcher](https://github.com/slee029/firstmate/blob/7484d746a73635bd56f6a02ddf58e7f3579b8aa5/.no-mistakes/evidence/fix/hhe1805-watcher-guard-race/live-early-term-sweep.log)
- Outcome: ⚠️ 1 info across 1 run (34m20s)

## Pipeline

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>

- ⚠️ `bin/fm-wake-lib.sh:713` - fm_lock_claim truncates and rewrites $ownerdir/pid AFTER fm_lock_try_create has already published the symlink (fm-wake-lib.sh:793 does `ln -s` then calls fm_lock_claim). The value is byte-identical to what fm_lock_prepare_owner wrote at :661, so the write is pure redundancy, but it is a truncate-then-write on an already-published lock - exactly the window this change exists to close, and the same class the author is removing from bin/fm-watch.sh&#39;s post-acquire refresh writes. Concrete sequence: the watcher publishes the lock; a turn-end guard evaluation calls fm_watcher_lock_read_pinned, whose `pid` read lands after open(O_TRUNC) and whose `pid_after` read (4 cat forks plus one readlink later) still lands before the writer&#39;s write() completes because the publisher was descheduled. The read is then STABLE (link unchanged, all four files byte-identical, pid empty), so fm_watcher_healthy takes the `pid-missing` branch at :299 and returns 1 with no retry, and fm_watcher_lock_unheld returns 0 (unheld) - a live, fresh watcher is reported as absent and the guard prints TURN WOULD END BLIND. The pinned-read retry cannot save this case because nothing looks torn. Remedy at the shared boundary rather than in the reader: in fm_lock_claim, read $ownerdir/pid and confirm it already equals $mypid (failing the claim if not) instead of rewriting it; fm_lock_claim is called only from fm_lock_try_create, whose owner dir is a fresh mktemp directory no other process can write, so the rewrite is always redundant.
- ⚠️ `bin/fm-wake-lib.sh:664` - The staging contract is keyed to nothing but &#34;the next lock this shell acquires&#34;, so the comment at :644-:649 (&#34;Empty values are skipped, so every other lock using this primitive is unaffected&#34;) is inaccurate for nested acquisitions. Concrete path: bin/fm-watch.sh:2141-2143 sets FM_LOCK_OWNER_FM_HOME/WATCHER_PATH/PID_IDENTITY and calls fm_lock_try_acquire &#34;$WATCH_LOCK&#34;; on the stale-recovery branch that function recurses into fm_lock_try_acquire &#34;$lockdir.steal&#34; (:1153) and calls _fm_recovery_marker_publish, which does fm_lock_acquire_wait on &#34;$STATE/.watcher-down.lock&#34; (:1193 -&gt; :855). Both of those owner directories are therefore created carrying the watcher&#39;s fm-home, watcher-path and pid-identity. No current reader consumes those files on those two locks and fm_lock_clean_known_files removes them on release, so there is no live defect today - but other locks in this repo DO read pid-identity from their owner dir (fm-wake-lib.sh:1908, :1957; fm-afk-launch.sh:128; fm-afk-start.sh:86), so the unscoped global is a live foot-gun for any future caller that stages before a lock sequence. The intent needs staging only for $WATCH_LOCK; the strictly narrower form is to key the staged values to the intended lock path (stage only when the owner dir belongs to that lock) rather than to shell-global state. Flagging rather than auto-fixing because the narrower form changes the primitive&#39;s public contract.
- ℹ️ `bin/fm-wake-lib.sh:305` - fm_watcher_healthy no longer calls fm_watcher_lock_matches_pid; it inlines the same four predicates (home equality, watcher-path equality, non-empty recorded identity, live-pid identity equality) at :305-:325, while fm_watcher_lock_matches_pid keeps its own copy at :255-:262. Inlining is justified - calling matches_pid would perform a second, independent pinned read of a possibly different generation - but the matching rule is now defined twice and the two copies can drift (for example if identity comparison gains a normalization step, a watcher could pass one reader and fail the other, which is precisely the mixed-verdict shape this ticket is about). The narrower form is one helper that evaluates the predicates against the already-pinned FM_WATCHER_PIN_* values, called by both readers after their own pinned read; that keeps a single owner of the rule without adding a second lock read.
- ℹ️ `bin/fm-watch.sh:2131` - Informational, for the executor&#39;s awareness only - not a re-litigation of settled decisions. The reviewed head 66ae06b is byte-identical to the commit reviewed in round 1, so none of the five round-1 findings the user chose to fix are present yet: the relocated EXIT trap still releases the lock and re-publishes the downtime marker on the two retain-evidence exits (bin/fm-watch.sh:2172, :2177 via watcher_cleanup at :2107-:2128); FM_WATCHER_VERDICT_DETAIL still exists with no production consumer (bin/fm-wake-lib.sh:555, read only by tests/fm-watcher-guard-race.test.sh:300); FM_WATCHER_PIN_TORN is still incremented and never read (bin/fm-wake-lib.sh:140, :200, :203, :212); the three post-acquire refresh writes remain (bin/fm-watch.sh:2245-2247); and test_rapid_cycling_yields_no_false_blind still races the 60ms absent window (tests/fm-watcher-guard-race.test.sh:205). I have not re-reported these as new findings; the risk level below accounts for them being outstanding.

🔧 Fix applied.

- ⚠️ `tests/fm-watcher-guard-race.test.sh:354` - test_published_pid_is_verified_not_rewritten detects a post-publication pid rewrite only through `chmod a-w &#34;$2/pid&#34;` inside the `ln` override. Root bypasses DAC on file writes, so if fm_lock_claim still did `printf &#39;%s\n&#39; &#34;$mypid&#34; &gt; &#34;$ownerdir/pid&#34;` the write would succeed, the acquisition would succeed, and both assertions (`pinned rc=0 pid=$me` and `acquired pid=$me`) would match — the test passes with the defect present. This matters because the test is the only evidence for the claim-verify fix, and CI containers commonly run as root. The suite already root-guards permission-dependent cases (tests/fm-procevent.test.sh:2930, tests/fm-spawn-dispatch-profile.test.sh:1159, tests/fm-extension-binding.test.sh:301); add the same `[ &#34;$(id -u)&#34; = 0 ]` skip here, or use a detector independent of permission bits. Mechanical test hardening, no product behavior change.
- ℹ️ `bin/fm-wake-lib.sh:563` - fm_watcher_supervision_verdict no longer resets FM_WATCHER_HEALTH_REASON=&#39;unknown&#39; at entry (removed alongside FM_WATCHER_VERDICT_DETAIL). On the autoarm early return at :576 the verdict never consults fm_watcher_healthy, so the variable retains a reason from an unrelated earlier evaluation in the same shell, contradicting its documented contract at :237 (&#34;names which predicate failed the last guard evaluation&#34;). Not a live defect: the only reader, bin/fm-turnend-guard.sh:252, always calls fm_watcher_healthy itself first, and no caller reads it after a verdict. Noted for awareness only.

🔧 Fix applied.
2 warnings still open:

- ⚠️ `tests/fm-watcher-guard-race.test.sh:445` - test_retain_evidence_exit_leaves_lock_and_marker_untouched forces the retain-evidence exit purely through `chmod 0500 &#34;$marker&#34;`. The failure it depends on is the quarantine step in _fm_recovery_marker_arm_check (bin/fm-wake-lib.sh:987), `mv -- &#34;$marker&#34; &#34;$quarantine/marker&#34;`: renaming a directory requires write permission on the directory itself, so for an ordinary user the mv fails, arm_check returns 1, and the watcher takes the `could not be consumed safely` exit the test asserts (I confirmed the EACCES behavior directly). Root holds CAP_DAC_OVERRIDE and bypasses that check, so under root the mv succeeds, _fm_recovery_marker_write_locked publishes a fresh marker, arm_check returns 0, and the watcher proceeds into its normal poll loop with FM_CHECK_INTERVAL=999999. The 10s wait loop at :450 then expires and the test fails at :458 with &#34;watcher kept running instead of taking the retain-evidence exit&#34; — a false CI failure with no product defect, in root containers. This is the same root-fragility class the previous fix round just closed at :353 for test_published_pid_is_verified_not_rewritten, and there is no permission-independent way to make this arm-check fail (the mktemp -d quarantine and the marker write are both DAC-gated too). Remedy: add the same `[ &#34;$(id -u)&#34; = 0 ]` skip used at :353 and in tests/fm-extension-binding.test.sh:301. Mechanical test hardening, no product behavior change.
- ⚠️ `bin/fm-wake-lib.sh:147` - The absent-lock retry component — fm_watcher_absent_attempts (:147-158), the FM_WATCHER_PIN_ABSENT_ATTEMPTS override, and the rc=1 retry branch in fm_watcher_healthy (:314-322) — is separate from the pinning mechanism and is not needed by the stated intent. The intent names &#34;non-atomic .watch.lock publication vs unpinned guard reads&#34;; an absent lock is not a torn read and not a partial publication, it is a genuinely unheld lock, and after the staging fix the only way a reader lands there is the release-plus-republish gap of a watcher restart. No test in the change exercises it: test_generation_flip_mid_read_yields_no_false_blind blocks the reader inside a `cat` interception and waits on flip.done, so the reader never evaluates during the absent window, and test_health_reason_matrix reaches lock-absent only after the retries are exhausted. It is not free either: on the genuinely-down path fm_watcher_healthy now sleeps 3 x 0.02s before returning, and bin/fm-turnend-guard.sh calls it at :205, inside autoarm_owns_recovery for each of the 8 sync-wait iterations at :493, and again at :452/:495/:504 — roughly half a second added to every real TURN WOULD END BLIND decision. Recommend removing the component so rc=1 returns lock-absent immediately as before; if the release-gap tolerance is deliberate, it needs a test that actually evaluates the guard inside the gap. Flagging rather than auto-fixing because removing it changes guard timing behavior the author may have chosen on purpose.

🔧 Fix applied.
2 warnings still open:

- ⚠️ `bin/fm-watch-arm.sh:412` - The `--restart` path is now the one remaining watch-lock reader that mixes generations, and the change widened its window. At :412 it reads `lock_pid` with a bare `cat &#34;$WATCH_LOCK/pid&#34;`, then hands that pid to `fm_watcher_lock_matches_pid`, which (as of this change, bin/fm-wake-lib.sh:271-288) performs its OWN pinned read of a possibly different generation and, on a detected tear, sleeps up to 10 x 0.02s before settling. Concrete sequence in the autoarm model, where the watcher exits on every wake and the arm starts a replacement: W1 (pid 100) is tearing down while W2 (pid 200) has just published its complete generation. `--restart` reads pid=100 (gen A); `fm_pid_alive 100` is still true; `fm_watcher_lock_matches_pid` pins gen B and compares `fm_pid_identity 100` against W2&#39;s identity -&gt; identity-mismatch -&gt; returns 1. Control falls to `clear_stale_recorded_watcher_lock` (:225-234), which re-reads fm-home/watcher-path/pid-identity unpinned from gen B - all three match this home - and runs `fm_recovery_transition ... clear-stale-lock`, i.e. `fm_lock_remove_path &#34;$WATCH_LOCK&#34;` (bin/fm-wake-lib.sh:1086-1089). That deletes the LIVE W2&#39;s lock; the arm then starts W3, leaving two watchers in one home, and W2&#39;s own EXIT trap will no longer own the lock so it also skips its marker transition. The cross-generation read is pre-existing, but before this change the metadata reads followed the pid read within three syscalls; the retrying pinned reader makes the mismatch materially more likely precisely during a flip. Earliest shared boundary for the remedy: have the restart path take the pid from a single pinned read (expose the already-pinned `FM_WATCHER_PIN_PID` alongside the match verdict) instead of reading `$WATCH_LOCK/pid` independently. Flagged ask-user because that extends the change beyond the stated turn-end-guard intent and touches the arm layer&#39;s restart semantics.
- ⚠️ `bin/fm-wake-lib.sh:716` - The fix round introduced a new branch in `fm_lock_claim`: on `back != mypid` it now removes the published symlink (`:716-718`) before discarding the owner. The finding that prompted this round required only that the claim VERIFY the staged pid instead of rewriting it; this extra lock-removal branch is not required and is redundant for every reachable caller. `fm_lock_try_create` - the only production caller - already performs exactly this cleanup at :753-755 after any `fm_lock_claim` failure, and the direct test caller in tests/fm-watcher-lock.test.sh:465 reaches the identical removal through the steal branch at :726-729. There is no caller in the repo that creates an owner directory with a foreign pid and no steal lock, so :716-718 can never be the branch that does the work. Smallest honest remedy is to remove the component (drop lines :716-718, leaving `fm_lock_discard_owner` + `return 1`), not to document or harden it.

🔧 Fix applied.
2 infos still open:

- ℹ️ `bin/fm-watch.sh:2088` - The comment says this watcher&#39;s pid is &#34;recorded in the lock by fm_lock_claim (which writes ${BASHPID:-$$} from this same main shell)&#34;. As of this change fm_lock_claim no longer writes the pid at all — it only reads it back and compares (bin/fm-wake-lib.sh:711-720). The pid is now written by fm_lock_prepare_owner (bin/fm-wake-lib.sh:653-656) into the staged owner directory before the symlink is published. The invariant the comment is defending (WATCHER_PID must equal the stored holder pid, read without a command substitution) still holds, because fm_lock_prepare_owner also runs in the main shell via fm_current_pid; only the named owner is wrong. In a file whose whole point is that nothing rewrites the published generation, a comment pointing at the function that used to do the rewrite is the one most likely to be trusted by the next reader. Mechanical comment correction, no behavior change.
- ℹ️ `tests/fm-watcher-guard-race.test.sh:352` - Both behaviors that prior fix rounds introduced as their headline repairs are now unprovable under a root CI container: test_published_pid_is_verified_not_rewritten (:352) and test_retain_evidence_exit_leaves_lock_and_marker_untouched (:440) return pass() immediately when id -u is 0, because each detector depends on DAC denying a write (chmod a-w on the pid file) or a rename (chmod 0500 on the marker directory), and root bypasses both. The skips themselves are correct — a false failure would be worse, the repo already uses this pattern (tests/fm-extension-binding.test.sh:301), and the prior round established there is no permission-independent way to force the arm-check quarantine to fail. Recording it only so the merge decision accounts for it: if CI runs as root, the claim-verify change and the retain-evidence regression repair land with their behavioral proof skipped, and the remaining seven cases in this file do not cover either. No action requested.
</details>

<details>

- ℹ️ `bin/fm-wake-lib.sh:1230` - During the 50ms-cadence restart hammer, 2 of 248 watcher startups printed a bash parse error (&#39;trap: line 2: unexpected EOF while looking for matching )&#39; from the fm_lock_acquire_wait TERM trap in fm-wake-lib.sh) when TERM landed while bash was parsing, and one watcher ignored that TERM until a later TERM stopped it cleanly. The trap string is unchanged from base and the 150-iteration TERM stress on the fixed code reproduced no survivor, no stale lock, and no parse error, so this is a pre-existing bash trap/parser interaction rather than a defect introduced here. Noted for awareness only.
- Live validation: ✅ go - 8 of 8 scenarios driven live against the product

| --- | --- | --- | --- |
| Turn-end hook never fires TURN WOULD END BLIND while a live watcher&#39;s lock is published, even under continuous watcher restarts | ✅ pass | live | live-fixed-hammer.log: hook evaluations=400 allowed=277 blocked-while-lock-held-throughout(FALSE-BLIND)=0 across 248 real watcher restarts |
| Regression reproduction: the same restart hammer against the base commit produces false BLINDs | ✅ pass | live | live-base-hammer.log: 399 of 399 lock-held-throughout evaluations blocked on base 43bf6d3 at the same cadence; live-base-hammer-1s-hold.log shows base allows once the partial window has passed |
| Adversarial: a genuinely dead watcher (SIGKILL, lock left behind) still blocks with a named predicate reason | ✅ pass | live | live-dead-and-tampered-watcher.log: hook rc=2 with &#39;Lock evidence: pid-dead&#39; |
| Adversarial: tampered lock metadata (fm-home rewritten on a live lock) blocks with a named reason | ✅ pass | live | live-dead-and-tampered-watcher.log: hook rc=2 with &#39;Lock evidence: home-mismatch&#39; |
| Published lock generation is complete at publication (no partial owner directory ever observable) while legacy staggered publication is still detectable by the same observer | ✅ pass | live | guard-race-suite.log: staged publication 0 partials in 139 samples; legacy staggered 103 partials in 124 samples |
| A generation flip forced at every read position inside one guard evaluation yields zero false BLINDs; absent-lock retry rides a republish within budget and blocks a sustained absence | ✅ pass | live | guard-race-suite.log: generation-flip and absent-retry cases pass |
| Retain-evidence exit of the real watcher leaves the held lock and a malformed marker byte-untouched | ✅ pass | live | guard-race-suite.log: test_retain_evidence_exit_leaves_lock_and_marker_untouched drives bin/fm-watch.sh and passes (run as uid 1000, not skipped) |
| A TERM arriving at any point during watcher startup, including mid-acquire, never strands a stale lock | ✅ pass | live | live-term-stress-fixed.log: iterations=150 survivors=0 stale-locks=0; base comparison live-term-stress-base.log: stale-locks=61 |

- <code>`bash tests/fm-watcher-guard-race.test.sh` (9/9 pass, evidence guard-race-suite.log)</code>
- <code>`bash tests/fm-turnend-guard.test.sh` (all pass, evidence turnend-guard-suite.log)</code>
- <code>Live hammer: real `bin/fm-watch.sh` killed/relaunched continuously (248 restarts) while real `bin/fm-turnend-guard.sh` evaluated 400 times with a task in flight; classified each evaluation by whether the same lock generation was published before and after (live-fixed-hammer.log)</code>
- `Same live hammer against base commit 43bf6d3 bin/ at identical cadence: 399/399 lock-held evaluations blocked (live-base-hammer.log); at 1s hold base steady state allows (live-base-hammer-1s-hold.log)`
- <code>Live adversarial: SIGKILL the live watcher -&gt; hook exit 2 with `Lock evidence: pid-dead`; overwrite fm-home on a live lock -&gt; exit 2 with `home-mismatch` (live-dead-and-tampered-watcher.log)</code>
- `Live TERM-during-startup sweep (fixed delays 5ms..1s) and 150-iteration random-delay stress on fixed and base: fixed 0 stale locks/0 survivors, base 61 stale locks (live-early-term-sweep.log, live-term-stress-fixed.log, live-term-stress-base.log)`
</details>

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>


slee029:
Hi — friendly nudge from a fork contributor: this PR is waiting on a maintainer approving the workflow runs, so CI has not been able to start (`action_required`). Whenever you have a moment, would you mind approving the runs so the checks can execute? Happy to rebase or split the change if that makes review easier, and no rush from our side. Thanks for maintaining this.

PR #5153, opened 2026-09-21
fix(bin): treat a slow contribution read as unmeasured, not unavailable

sdivanl:
## What changed

- Distinguish a forge read killed at its own five-second slice while the poll budget still had room (slow/unmeasured) from a genuinely unreachable forge. A slow read no longer records an error or emits the `contributions: observation unavailable for <url>` alarm; its prior record and verdict are preserved, and the once-per-episode wake for a genuinely unreachable forge is unchanged.
- Preserve the poll's rotate-after-every-attempt scheduling by tracking a durable per-URL last-attempt recency sequence: a global monotonic `seq` plus each URL's last sequence. An unseen URL is seeded at the current sequence so it joins the rotation, takes one turn, then yields. This keeps scheduling separate from observation freshness, so a slow read stays unmeasured (`checked_at` frozen) without pinning the head of the queue.
- Add multi-poll regressions: a slow read with budget remaining stays unmeasured and silent; a slow URL is retried after a healthy budget-consuming read; a newly linked URL joins the rotation without starving established ones.

Fixes the repeated `contributions: observation unavailable for <url>` wakes for contributions that read fine by hand moments later.

## Pipeline


sdivanl:
Rollup note: this head commit counts 20 check runs for 19 distinct checks. `PR must be raised via no-mistakes` was reported twice on `5087c0488f3a40e6d89d6deae0a0f28c3b8e45ba` — run `106398448360` (started 15:30:44Z) failed, and run `106400051167` (started 15:34:56Z) passed on the identical commit. Every other check is green, so that stale duplicate is the only red in the rollup.

Re-running or dismissing that single instance should clear it; no code change is needed.

